Issue #44, opened 2021-06-11
add whatsapp phone number for support in the mobile wallet


nicolasburtey:
**Is your feature request related to a problem? Please describe.**
People who want to reach out for any question on the wallet don't know how to do it.

**Describe the solution you'd like**
There is no easy way to contact the support from the app currently. The easiest way to get started because we think of a broader approach is to add a button from the settings to open the whatsapp app when tapping on a button.

The number can be hardcoded for now. We could provide the number with an API through the backend in later update. 

**Describe alternatives you've considered**
In app messaging, but would take a long time to implement

**Additional context**

cryptosalomao:
Addressed in #45

PR #45, opened 2021-06-12
add whatsapp contact button

cryptosalomao:
This PR addresses #44


app/constants/support.ts
```diff
@@ -0,0 +1,2 @@
+export const WHATSAPP_CONTACT_NUMBER = '+5599985049935'
```
nicolasburtey, line 1: should be +503 6983 5117

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -1,25 +1,36 @@
...
+
 
 const styles = EStyleSheet.create({
   screenStyle: {
     backgroundColor: palette.lighterGrey
   },
 })
 
-export const SettingsScreen = ({navigation}) => {
+export const SettingsScreen: any = ({ navigation }) => {
```
nicolasburtey, line 33: we're trying to move forward a strongly type database. could you add the correct type here?

app/constants/support.ts
cryptosalomao: Oh yeah, of course. I just added mine as an example

app/screens/settings-screen/settings-screen.tsx
cryptosalomao: Sure.

app/constants/support.ts
nicolasburtey: > Oh yeah, of course. I just added mine as an example
you'll risk getting a lot of spam ;)

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -32,7 +32,11 @@ const styles = EStyleSheet.create({
   },
 })
 
-export const SettingsScreen: ScreenType = ({ navigation }) => {
+type Props = {
+  navigation: any,
```
nicolasburtey, line 36: we're just pushing the `any` one depth further here :) but that'll do it for now. thanks.

nicolasburtey:
merged. didn't pay attention but we use yarn not npm. so I deleted package-json.lock in main

Issue #46, opened 2021-06-13
improve phone number formatting on confirmation screen


nicolasburtey:
**Is your feature request related to a problem? Please describe.**
When confirming the code sent by text, there is no mention of which phone the code was sent to.
This is from the prior screen. So if the phone was not entered correctly the only way to figure out is typically to wait for a few minutes, and if nothing happen go the previous screen

**Describe the solution you'd like**
Show the phone code in the code confirmation page.

nicolasburtey:
https://github.com/GaloyMoney/galoy-mobile/pull/48 is a first step. the downside is that the phone number is presented with the international prefix. many people who have not lived abroad might not be aware that +503 is supposed to be the international prefix and could be confuse by that.

I'll confirm this is the case when I'll be physically in El Salvador.

nicolasburtey:
@daviroo

secondl1ght:
It looks like the phone number is displayed now so I think we can close this issue.

PR #48, opened 2021-06-13
Feature/show phone number

cryptosalomao:
Addresses #46

### DIFF
--- a/app/screens/phone-auth-screen/phone-auth.tsx
+++ b/app/screens/phone-auth-screen/phone-auth.tsx
@@ -255,7 +255,7 @@ export const WelcomePhoneValidationScreen = ({
         <ScrollView>
           <View style={{ flex: 1, minHeight: 32 }} />
           <BadgerPhone style={styles.image} />
-          <Text style={styles.text}>{translate("WelcomePhoneValidationScreen.header")}</Text>
+          <Text style={styles.text}>{translate("WelcomePhoneValidationScreen.header", { phone })}</Text>
           <KeyboardAvoidingView
             keyboardVerticalOffset={-110}
             behavior={Platform.OS === "ios" ? "padding" : undefined}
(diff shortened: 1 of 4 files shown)


app/types/screen.ts
```diff
@@ -1 +1 @@
-export type ScreenType = (navigation: any) => void
+export type ScreenType = (navigation: any) => any
```
nicolasburtey, line 1: a type `(any) => any` is not really helpful. it makes the syntax more boiler-platy without adding any type safety. Typescript would tread no type information as any IIRC

app/screens/phone-auth-screen/phone-auth.tsx
```diff
@@ -108,15 +112,19 @@ const styles = EStyleSheet.create({
...
+}
+
+export const WelcomePhoneInputScreen: ScreenType = ({ navigation }: WelcomePhoneInputScreenProps) => {
   const [requestPhoneCode, { loading }] = useMutation(REQUEST_PHONE_CODE, {
     fetchPolicy: "no-cache"
-  });
+  })
 
-  const inputRef = useRef()
+  const inputRef: any = useRef()
```
nicolasburtey, line 124: here are some doc to type a mutable useRef the correct way: https://github.com/typescript-cheatsheets/react#option-2-mutable-value-ref

nicolasburtey:
The core PR (adding the phone number in the confirmation page) looks good.

On the typescript aspect, I know this repo is barely covered by types, and we should seriously add type from now on to have easier time refactoring, onboarding new developer, and limit risk of crash due to type issue. I don't have a strong opinion yet on what would be the best approach to scale this, but I have a feeling adding `any` to most of the type to not really helpful to getting there. I'll be more thinking: either add the correct type, or do not type at all. 

I'm sure there are best practices on how best going from javascript to typescript. I have not dive into them.

Issue #49, opened 2021-06-14
icon and text not showing on latest build


nicolasburtey:
**Describe the bug**

The home page should show 3 buttons for the payment and some text, on the last build it just show a white button. 

**To Reproduce**
Build with current master.
Problem is present both on Android and iOS

**Screenshots**
![IMG_C1EE302CB75B-1](https://user-images.githubusercontent.com/11156370/121922343-80872780-ccff-11eb-83e4-18b70de6868f.jpeg)

**Additional context**
I have not looked into the issue but I'm not sure why this is happening, is it a regression related to some previous (I had a look quickly and see nothing obvious related) or related to some third party dependancy.

The button are based on the [react native element](https://reactnativeelements.com/) and we use an old way to describe the syntax, there has been warning on it on the console for some time now. maybe it's related.

jrbuhl93:
The problem seems to be related to the ListItem used to make the large button component. https://reactnativeelements.com/docs/listitem/ It's possible it no longer supports setting title, left Avatar, etc in the properties of ListItem. A workaround is to set everything up more explicitly like in the contacts screen with nested elements. I will continue to look into this, but wanted to post this in case you plan on fixing it immediately.

nicolasburtey:
fix by @jrbuhl93 thanks!

PR #50, opened 2021-06-14
nest avatar and title in ListItem

jrbuhl93:
Possible workaround for issue #49


nicolasburtey:
tried it locally and it seems to work, thanks for the fix!

Issue #53, opened 2021-06-16
integration of the translation of the quiz


nicolasburtey:
**Describe the bug**

most of the app is being translate in Spanish but not the quiz. 

the quizz is quite complex so the translation file is (mostly) being generated programmatically from a google sheet. the script is in utils/parseCSV.js

PR #54, opened 2021-06-16
Spanish quiz translation integration

jrbuhl93:
Implements #53 . You may want to double check some of my translations such as "Usted Ganó" for "You Earned." I also changed the font size in textStyleBox from 18 to 16 because the Spanish translations were larger and hitting the borders of "BoxAdding."


nicolasburtey:
thanks for doing that @jrbuhl93 !

Issue #15, opened 2020-03-30
use biometrics if available when opening the application


nicolasburtey:
We are adding a 2fa to validating large payment, but it would be a good features that would add security easily if we could ask for the biometrics at the launch of the app.

we have a lot more user on Android, so we could start with Android first.

jrbuhl93:
#58 addressed this issue, but we still want authentication when returning from the background state after a period of time.

nicolasburtey:
has been integrated

PR #58, opened 2021-06-20
Authentication screen

jrbuhl93:
Feature for #15 

I have added biometric authentication using this library: https://github.com/hieuvp/react-native-fingerprint-scanner. The library should work with all devices according to the documentation including those that use face id instead of fingerprint and ios. I have only thoroughly tested with my device: an Android Pixel 2 that uses fingerprint authentication. If the library does not detect a biometry sensor available, the app should go into the Primary screen (untested).

The authentication in this pull request is mandatory and cannot be turned off in the settings. If you think this is something users would want, Opt-in/out authentication would require a boolean property added to the user object (on servers too) and a another item in the settings.

I hope we can thoroughly test this feature before merging it into the main codebase.


app/screens/authentication-screen/biometry-check-screen.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type Props = {
+  navigation: any,
+}
+
+export const BiometryCheckScreen: ScreenType = ({ navigation }: Props) => {
+
+  React.useEffect(() => {
+    checkIfBiometryIsAvailable()
+  })
```
nicolasburtey, line 41: this can probbaly be written as `React.useEffect(() => checkIfBiometryIsAvailable)`

app/screens/authentication-screen/biometry-check-screen.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const checkIfBiometryIsAvailable = async () => {
+    const biometryIsAvailable = await isSensorAvailable()
```
nicolasburtey, line 44: are there some odds that `isSensorAvailable` could throw?

app/screens/move-money-screen/money-money-screen.tsx
```diff
@@ -173,6 +173,12 @@ export const MoveMoneyScreenDataInjected = ({ navigation }) => {
 export const MoveMoneyScreen = (
   ({ walletIsActive, navigation, loading, error, transactions,
     refetch, amount, amountOtherCurrency, isUpdateAvailable }) => {
+  
+  useEffect(() => {
+    StatusBar.setTranslucent(false)
```
nicolasburtey, line 178: if this supposed to be set for the whole app as a permanent settings?
In this case we could set that earlier than `money-money-screen.tsx`

I remember trying to play with the StatusBar bar some time ago. I think there are some bug when going to the quiz tab and coming back to the main page. 

I think a useful issue would be a large refactoring the all the style/css. right now there are a lot of quick fixes on the style without a coherent structure. this will get more important as other project wants to fork this project to have their own wallet.

I've created a ticket for that: https://github.com/GaloyMoney/galoy-mobile/issues/59

app/utils/biometricAuthentication.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const authenticate = async (handleSuccess, handleFailure) => {
+  await FingerprintScanner.release()
+  FingerprintScanner
+    .authenticate({ 
+      description: translate("AuthenticationScreen.authenticationDescription"),
+      fallbackEnabled: true
+    })
+    .then(() => {
+      FingerprintScanner.release()
```
nicolasburtey, line 18: could `FingerprintScanner.release()` be in a `finally` branch? or it needs to be called before `handleSuccess`/`handleFailure` function?

app/screens/authentication-screen/biometry-check-screen.tsx
jrbuhl93: Reworked it to be then/catch

app/screens/move-money-screen/money-money-screen.tsx
jrbuhl93: The "StatusBar.setTranslucent(false)" line was unneeded, so I deleted it. I had copied all three lines together from another screen.

app/utils/biometricAuthentication.ts
jrbuhl93: Yes, this was me being unfamiliar/rusty with some parts of javascript.

jrbuhl93:
I should mention https://github.com/hieuvp/react-native-fingerprint-scanner is no longer being actively maintained by the owner: https://github.com/hieuvp/react-native-fingerprint-scanner/issues/178. I'm wasn't able to find an actively maintained library that seemed like a good fit. https://github.com/oblador/react-native-keychain seems maintained and GaloyMoney already has it installed, but I was running into issues with it on my Android Pixel 2.

Also, how do you think the user experience of this feature will be? I could see it being a problem if people have dirt on their hands from working, but I haven't been to El Zonte, so it is hard for me to say. An alternate way to authenticate besides logging out and then back in might be needed.

app/screens/authentication-screen/biometry-check-screen.tsx
jrbuhl93: Is there a reason you prefer it setup like this? The rest of the app seems to use React.useEffect(() => {}).

nicolasburtey:
A quick test.

I have tried it on an iPhone X with FaceId (have not tried android yet) and the flow is a bit awkward: after login in, nothing really changes as of now.

IMG_5494

but then, when you restart the app, it's asking if you want to set up FaceId at the second launch. this is not the place to ask for FaceId setup. this should be either after a succesful login, or in the settings tab.

I have not tried the flow of clicking on "don't allow" here. something to give a look at.

also, if you log out, the faceId registration is not being reset. need to re-install the app for that.

But otherwise the faceId works as expected after it has been set up. if you don't faceId you can't log in. I think there could be a fallback with a pin number of the iOS. that would be nice to have.

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -49,6 +50,42 @@ export const SettingsScreen: ScreenType = ({ navigation }: Props) => {
...
+        }
+      }, (err) => {
+        if (isBiometricsEnabled !== null) {
+          navigation.navigate("security", { mIsBiometricsEnabled: isBiometricsEnabled, mIsPinEnabled: false })
+        } else {
+          isPinEnabled = false
+        }
+      })
+
+  }
```
cryptosalomao, line 87: Let's make this an util function in case we need this check somewhere else in the future

app/screens/authentication-screen/pin-screen.tsx
```diff
@@ -0,0 +1,303 @@
...
+          <View style={enteredPIN.length > 3 ? styles.filledCircle : styles.emptyCircle} />
+        </View>
+      </View>
+      <View style={styles.helperTextContainer}>
+        <Text style={styles.helperText}>{helperText}</Text>
+      </View>
+      <View style={styles.pinPad}>
+        <View style={styles.pinPadRow}>
+          <View style={styles.pinPadButtonContainer}>
+            <Button buttonStyle={styles.pinPadButton} titleStyle={styles.pinPadButtonTitle} title="1" onPress={() => addDigit("1")} />
```
bodymindarts, line 246: Can this repetition be de-duplicated? If you wanted to change something on the `<Button>` tag it looks like you'd have to change every line.
Can't you have a function or component that just takes the number as input and returns the correct `<Button>`

app/screens/authentication-screen/pin-screen.tsx
```diff
@@ -0,0 +1,303 @@
...
+            <Button buttonStyle={styles.pinPadButton} titleStyle={styles.pinPadButtonTitle} title="9" onPress={() => addDigit("9")} />
+          </View>
+        </View>
+        <View style={styles.pinPadRow}>
+        <View style={styles.pinPadButtonContainer}/ >
+        <View style={styles.pinPadButtonContainer}>
+          <Button buttonStyle={styles.pinPadButton} titleStyle={styles.pinPadButtonTitle} title="0" onPress={() => addDigit("0")} />
+        </View>
+        <View style={styles.pinPadButtonContainer}>
+          {/* <Icon.Button
```
bodymindarts, line 283: Can we get rid of the comment before merge?

app/screens/settings-screen/security-screen.tsx
```diff
@@ -0,0 +1,196 @@
+import * as React from "react"
+import { useState } from "react"
+import { useFocusEffect } from '@react-navigation/native'
+import { Text, View } from "react-native"
+import { Button, Switch } from "react-native-elements"
+import EStyleSheet from 'react-native-extended-stylesheet'
+import RNSecureKeyStore, {ACCESSIBLE} from "react-native-secure-key-store"
+
+// Components
```
bodymindarts, line 9: Are these comments necessary?

app/screens/settings-screen/security-screen.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+export const SecurityScreen: ScreenType = ({ route, navigation }: Props) => {
+
+  const { mIsBiometricsEnabled, mIsPinEnabled } = route.params
+
+  const [isBiometricsEnabled, setIsBiometricsEnabled] = useState(mIsBiometricsEnabled)
+  const [isPinEnabled, setIsPinEnabled] = useState(mIsPinEnabled)
+
+  useFocusEffect(() => {
+    RNSecureKeyStore.get("isBiometricsEnabled")
```
bodymindarts, line 95: Can these strings `"PIN"` and `"isBiometricsEnabled"` be extracted into `const` variables?

app/screens/settings-screen/security-screen.tsx
```diff
@@ -0,0 +1,196 @@
...
+      }, (err) => {
+        setIsPinEnabled(false)
+      })
+  })
+
+  const onBiometricsValueChanged = (value) => {
+    if (value) {
+      isSensorAvailable(handleBiometryAvailabilitySuccess, handleBiometryAvailabilityFailure)
+    } else {
+      RNSecureKeyStore.remove("isBiometricsEnabled")
```
bodymindarts, line 114: Duplication of `magic` value... definitely needs extracting into `constant`

app/screens/settings-screen/security-screen.tsx
```diff
@@ -0,0 +1,196 @@
...
+          // unable to remove isBiometricsEnabled
+        })
+    }
+  }
+
+  const handleBiometryAvailabilitySuccess = (isBiometryAvailable: Boolean) => {
+    if (isBiometryAvailable) {
+      authenticate(translate("AuthenticationScreen.setUpAuthenticationDescription"), handleAuthenticationSuccess, handleAuthenticationFailure)      
+    } else {
+      toastShow("Biometric sensor is not available.")
```
bodymindarts, line 127: Are these strings for internal logging / debugging or are they user facing and need need translation via `a18n`?

app/screens/settings-screen/security-screen.tsx
```diff
@@ -0,0 +1,196 @@
...
+        <Button
+          buttonStyle={styles.button}
+          titleStyle={styles.buttonTitle}
+          title={translate("SecurityScreen.setPin")}
+          onPress={() => navigation.navigate("pin", { screenPurpose: "setPIN" })}
+        />
+      </View>
+    </Screen>
+  )
+}
```
bodymindarts, line 196: Line break warning. Hopefully linting will find this when we've integrated `eslint`

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -49,6 +50,42 @@ export const SettingsScreen: ScreenType = ({ navigation }: Props) => {
...
+        }
+      }, (err) => {
+        if (isPinEnabled !== null) {
+          navigation.navigate("security", { mIsBiometricsEnabled: false, mIsPinEnabled: isPinEnabled })
+        } else {
+          isBiometricsEnabled = false
+        }
+      })
+
+    RNSecureKeyStore.get("PIN")
```
bodymindarts, line 72: Again these magic strings

app/screens/settings-screen/settings-screen.tsx
bodymindarts: This sounds like pre-mature optimisation.
> _in case_ we need

I think its fine to keep it in the this file as long as its the only place we're using it.

Not a big deal though.

app/utils/biometricAuthentication.ts
```diff
@@ -0,0 +1,29 @@
...
+    .then(() => {
+      handleSuccess()
+    })
+    .catch((error) => {
+      handleFailure()
+    })
+    .finally(() => {
+      FingerprintScanner.release()
+    })
+}
```
bodymindarts, line 29: again EOF... perhaps you can set your editor to always include it

app/utils/logout.ts
```diff
@@ -9,7 +10,10 @@ export const resetDataStore = async (client) => {
     await (new Token()).remove()
     // await AsyncStorage.clear()
     await AsyncStorage.multiRemove([NETWORK_STRING, TOKEN_KEY, BUILD_VERSION]) // use storage.ts wrapper
+    await RNSecureKeyStore.remove("isBiometricsEnabled")
+    await RNSecureKeyStore.remove("PIN")
+    await RNSecureKeyStore.remove("pinAttempts")
```
bodymindarts, line 15: Another magic string.

app/utils/logout.ts
```diff
@@ -9,7 +10,10 @@ export const resetDataStore = async (client) => {
...
   } catch(err) {
     console.log({err}, `error resetting RootStore`)
   }
-}
+}
```
bodymindarts, line 19: eof

bodymindarts (commented):
Looks good over all.... see comments for small cleanup suggestions.

app/screens/settings-screen/security-screen.tsx
jrbuhl93: You're referring to the comments for the imports? I'm not sure what style we will end up going with for importing files. I'm agnostic on it. I was trying to copy the import styling I saw cryptosalomao using in some of his PRs.

bodymindarts:
Actually had a thought regarding the accesses with magic strings ie: ` RNSecureKeyStore.get("PIN")`. Perhaps you could wrap access to `RNSecureKeyStore`. In a dedicated object and use it via:
```
KeyStoreWrapper.getPin()
```
I think that would make it cleaner all together. Group all functions regarding that `KeyStore` in a wrapper.

app/screens/settings-screen/security-screen.tsx
bodymindarts: Yes that is what I'm referring too.

I would advocate to use as few comments as possible and instead try to make the code as self-explanitory as possible.
In these cases there is often a 1 word comment where the 1 word is contained in the path of what is being imported.
Like `themes` I don't need to read it in a comment. I can see it in the import path.

Communicating via code layout / structure (ie. dir and file names) is the way to go here! cc @cryptosalomao

app/screens/authentication-screen/authentication-check-screen.tsx
```diff
@@ -0,0 +1,61 @@
...
+})
+
+type Props = {
+  navigation: any,
+}
+
+export const AuthenticationCheckScreen: ScreenType = ({ navigation }: Props) => {
+  useFocusEffect(() => {
+    checkForAuthentication()
+  })
```
nicolasburtey, line 35:
```
  useEffect(() => {
    checkForAuthentication()
  }, [])
```

app/screens/authentication-screen/authentication-check-screen.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const checkForAuthentication = async () => {
```
nicolasburtey, line 37: put function within useEffect scope

app/screens/authentication-screen/authentication-screen.tsx
```diff
@@ -0,0 +1,180 @@
...
+    marginTop: 24,
+  }
+})
+
+type Props = {
+  route: any,
+  navigation: any,
+}
+
+export const AuthenticationScreen: ScreenType = ({ route, navigation }: Props) => {
```
nicolasburtey, line 70: add type related to route.params

app/screens/authentication-screen/authentication-screen.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const client = useApolloClient()
+
+  const { screenPurpose, pin, mPinAttempts } = route.params
+
+  useFocusEffect(() => {
+    attemptAuthentication()
+  })
+
+  const attemptAuthentication = () => {
```
nicolasburtey, line 79: async?

app/screens/authentication-screen/authentication-screen.tsx
```diff
@@ -0,0 +1,180 @@
...
+    let description
+    if (screenPurpose === "authenticate") {
+      description = translate("AuthenticationScreen.authenticationDescription")
+    } else {
+      description = translate("AuthenticationScreen.setUpAuthenticationDescription")
+    }
+    BiometricWrapper.authenticate(description, handleAuthenticationSuccess, handleAuthenticationFailure)
```
nicolasburtey, line 86: add 1 line comment on what this does

app/screens/authentication-screen/pin-screen.tsx
```diff
@@ -0,0 +1,273 @@
...
+    if (newEnteredPIN === pin) {
+      KeyStoreWrapper.setPinAttempts("0")
+      navigation.reset({
+        index: 0,
+        routes: [
+          { name: 'Primary' }
+        ]
+      })
+    } else {
+      if (pinAttempts < 2) {
```
nicolasburtey, line 132: should be const outside of the loop

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -139,6 +148,14 @@ export const SettingsScreenJSX = (params) => {
     enabled: walletIsActive && notificationsEnabled,
     greyed: !walletIsActive,
   },
+  {
+    category: translate('common.security'),
+    icon: 'lock-closed-outline',
+    id: 'security',
+    action: () => securityAction(),
```
nicolasburtey, line 155: `action: securityAction`

app/screens/authentication-screen/authentication-screen.tsx
jrbuhl93: I don't think `attemptAuthentication` needs to be async even though `BiometricWrapper.authenticate` is asnyc. Unless we want to await a return value `BiometricWrapper.authenticate` instead of using callbacks. Is there another reason to declare `attemptAuthentication` async?

nicolasburtey:
some text cropped on iOS:

image

nicolasburtey:
App running in the background. When coming back to the app, not asking for faceId again. But I think it ask back FaceId because the app can stay in the background for a long time.

nicolasburtey:
If I go to setPin after a pin has already being set, I am being stuck on this screen. I can put 4 digits and nothing happens.

IMG_5A7B11B86475-1

nicolasburtey:
after too many attempt of bad pin, the "too many attempt to login, log out" text appeared maybe only for 0.2 seconds. I could barely see it before the screen had reset.

jrbuhl93:
Would you like me to add authentication when the device goes from background -> active. I have not implemented this yet, only when the app starts from terminated. Also do you think the best way to handle the "too many attempt to login, log out" is to add an `await sleep(1000)`?

nicolasburtey:
> Would you like me to add authentication when the device goes from background -> active. I have not implemented this yet, only when the app starts from terminated

yes. this could be done as a follow up PR.

nicolasburtey:
> Also do you think the best way to handle the "too many attempt to login, log out" is to add an await sleep(1000)?

can be a good option yes. or it could be a popup. and when you click on ok you're being redirected at this stage.

nicolasburtey:
> If I go to setPin, I am being stuck on this screen. I can put 4 digits and nothing happens.

I have this issue on Android as well. this is the biggest issue I see so far on my end.

PR #65, opened 2021-06-25
feature/linting

cryptosalomao:
- added proper configuration for typescript and react native linting
- added proper prettier configurations through eslint
- formatted project according to the already existing rules on the back-end repository


bodymindarts:
There are many differences between `.eslintrc.js` and `.eslintrc.json` in https://github.com/GaloyMoney/galoy. Can we make them more similar?

Also for the commit messages lets try to be consistent over all repos. In the [main galley repo](https://github.com/GaloyMoney/galoy/blob/main/DEV.md#contributing) we have a pointer to [this guide](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) as a good starting point.

nicolasburtey:
> There are many differences between .eslintrc.js and .eslintrc.json in https://github.com/GaloyMoney/galoy. Can we make them more similar?

I actually think they should be the same unless there is a strong reason for them to be different

cryptosalomao:
> > There are many differences between .eslintrc.js and .eslintrc.json in https://github.com/GaloyMoney/galoy. Can we make them more similar?
> 
> I actually think they should be the same unless there is a strong reason for them to be different

They can't be the same because one has a general typescript project configuration, and the other a react native typescript configuration. Some rules are different, such as JSX syntax, styling format, etc. We can get closer to the configuration, but this one already includes everything we use on the back-end repo.

nicolasburtey:
what is different that could closer to our galoy repo, given the react-native constraint?

PR #66, opened 2021-06-27
[WIP] lnurl implementation

cryptosalomao:
Work in progress branch for lnurl protocol implementation

Progress:

- [x] implement lnurl-pay
- [ ] implement lnurl-withdraw
- [ ] integrate lnurl-pay into payments flow
- [ ] integrate lnurl-withdraw into invoices flow


cryptosalomao:
I'm still trying to understand the invoice/payment flow, since things under `send-bitcoin-screen.tsx` file are messy, such as the state management.

nicolasburtey:
@cryptosalomao I don't think there is a need for a lnurl-withdraw for our wallet (at least for now, I don't see any obvious use case). I would skip this part from the PR.

bodymindarts:
Can we merge this in steps and hide the features behind a feature flag? Waiting until all 4 boxes are checked could result in merge conflicts.

Also I have an issue with this PR putting so much business / domain logic into the `utils` folder.

So far this repo lacks structure outside of the UI screens and components.

I advocate having a domain revealing folder structure which enables ppl browsing the code to quickly assess what the repo is about. In this case that would mean something like:

```
src/lnurl/{constants, types,utils}.ts
```

This improves locality of information and reduces comprehension time.

Also we should reduce the usage of `utils` files. Relying heavily on `helpers` or `utils` type of files is an anti-pattern. I would only use it for _non_- domain relevant functions.

jrbuhl93:
It looks like you are checking if the amount is valid before building the invoice request where you convert sats to mSats. But the request response will have the min and max measured in MilliSatoshi. So I think right now your call to isAmountValid is comparing a user entered value in sats to bounds in MilliSatoshi.

app/utils/lnurl.ts
```diff
@@ -0,0 +1,57 @@
...
+export const isProtocolSupported = (tag: string): boolean => !!SUPPORTED_PROTOCOLS[tag]
+
+export const isAmountValid = (
+  amount: number,
+  minSendable: number,
+  maxSendable: number,
+): boolean => minSendable <= amount && amount <= maxSendable
+
+export const parseUrl = async (invoice: string): Promise<LNUrlPay> => {
+  const { words } = bech32.decode(invoice, 2000)
```
jrbuhl93, line 21: Why was the number 2000 chosen? Is there a standard? I notice Blue Wallet uses 10000.

app/screens/send-bitcoin-screen/send-lnurl-screen.tsx
```diff
@@ -0,0 +1,299 @@
...
+            editable={true}
+            onUpdateAmount={(input) => {
+              setAmount(input)
+            }}
+            forceKeyboard
+            price={price}
+          />
+          <View style={styles.errorContainer}>
+            {!hasEnoughBalance && <Text>You do not have enough balance</Text>}
+            {!minSendable !== null && !isAmountValid(amount, minSendable, maxSendable) && (
```
jrbuhl93, line 229: Is the `!` before `minSendable` needed?

jrbuhl93 (approved):

bodymindarts:
Check code is failing. And when you rebase please remove the commit that adds the Apollo typegen stuff. Its not working and is not related to this PR. `git rebase -i` is your friend here.

app/utils/lnurl.ts
cryptosalomao: This is not standard. The recommendation is to not exceed 1.000 characters because of error correction, and most of the lnurl-pay codes will not exceed 2.000 characters, so it seems reasonable to use this value.

PR #69, opened 2021-06-30
Enforce linting

bodymindarts:
Add a GitHub action to enforce the linting as we do in the backend repo.

This is currently failing and needs fixing before we merge.


cryptosalomao:
We should consider using a cache step for the dependencies install to speedup the process

https://github.com/actions/cache/blob/main/examples.md#node---yarn

PR #75, opened 2021-07-06
Convert QR data to lowercase after scan

juandelcid:
Some apps encode bitcoin addresses and lightning invoices in UPPERCASE, lightningPayReq from bolt11 needs it in lowercase to work so we need to convert to lowercase after scan.


jrbuhl93 (approved):

bodymindarts:
Can we clean up the commit history?
This commit
> Merge branch 'main' into main
Is noise

Issue #37, opened 2021-05-11
make sure phone only have + and number


nicolasburtey:
phone number form when logging in case have non number characters like ",". the query is still send to the backend but is being rejected. there should be a check before sending the query that the phone number has the same regex check than the backend has.

dannyjmac:
[PR 76 ](https://github.com/GaloyMoney/galoy-mobile/pull/76)

daviroo:
Looks like this was fixed in #76.

PR #76, opened 2021-07-07
Feat/37 make sure phone number only number and plus

dannyjmac:
[Issue 37](https://github.com/GaloyMoney/galoy-mobile/issues/37)

Tests the phone number against the same regexp in the backend's graphql schema before requesting a phone code.

And some minor refactoring.


nicolasburtey:
thank you!

PR #87, opened 2021-07-12
Refactor username screen

jrbuhl93:
I was looking to add validation to prevent the username exists errors on the server for invalid usernames, but it looks like this is already implemented. But I noticed some things like the init boolean weren't working, so I cleaned up the screen a bit. There are still the todos for memoization and debouncing, but those changes probably aren't priorities for now. 

Edit: The username exists query errors are coming from the send bitcoin screen. I'll fix that in a different PR.


app/screens/settings-screen/username-screen.tsx
```diff
@@ -34,14 +36,14 @@ type Props = {
...
+export const UsernameScreen: ScreenType = ({ navigation }: Props) => {
+  const [loading, setLoading] = useState(false)
+  const [input, setInput] = useState("")
+  const [message, setMessage] = useState("")
+  const [messageIsError, setMessageIsError] = useState(false)
 
-  // we don't show a error message before the user had a change to add 3 caracters
-  const [init, setInit] = React.useState(true)
+  // we don't show a error message before the user had a change to add 3 characters
+  const [init, setInit] = useState(true)
```
bodymindarts, line 46: Since you're also cleaning up in this PR. Can we rename this to something more explicit like `shouldShowErrorMessage` or something like that?

app/screens/settings-screen/username-screen.tsx
```diff
@@ -202,18 +186,24 @@ export const UsernameScreen = ({ navigation }: Props) => {
         leftIcon={{ type: "ionicon", name: "ios-person-circle" }}
         onChangeText={onChangeText}
         errorStyle={styles(messageIsError).error}
-        errorMessage={message}
+        errorMessage={(() => {
+          // Wait for everything to download and sync through useEffect before displaying
+          if (
+            !loadingUserNameExist &&
+            (messageIsError || messageIsError === usernameExists)
```
bodymindarts, line 193: This boolean expression is hard to parse. How do we ever get to `messageIsError === usernameExists`? If `messageIsError` is defined the first part of the `||` will already be `true` right?

app/screens/settings-screen/username-screen.tsx
```diff
@@ -130,8 +118,10 @@ export const UsernameScreen = ({ navigation }: Props) => {
 
   const assertValidInput = (): boolean => {
     if (input.length <= 2) {
-      setMessage(translate("UsernameScreen.3CaractersMinimum"))
-      setMessageIsError(true)
+      if (!init) {
```
bodymindarts, line 121: I wouldn't expect this function to have a side effect. Can we decouple the logic of `validness` from the resulting behaviour (error message + focus)?

app/screens/settings-screen/username-screen.tsx
jrbuhl93: The problem is messageIsError is updated by a useEffect watching data (data is what gets parsed into usernameExists). So there was a flicker between react renderings where usernameExists could be true and messageIsError could be false or vice versa.

I could give this boolean expression a name or we could reduce the use of useEffect.

app/screens/settings-screen/username-screen.tsx
```diff
@@ -202,18 +187,14 @@ export const UsernameScreen = ({ navigation }: Props) => {
         leftIcon={{ type: "ionicon", name: "ios-person-circle" }}
         onChangeText={onChangeText}
         errorStyle={styles(messageIsError).error}
-        errorMessage={message}
+        errorMessage={loadingUserNameExist ? "" : message}
```
bodymindarts, line 190: I'm surprised not to see the `shouldShowErrorMessage` variable used here. Perhaps `loadingUserNameExists` needs to be taken into account in the logic that determines the value of that variable.

app/screens/settings-screen/username-screen.tsx
```diff
@@ -62,24 +62,27 @@ export const UsernameScreen = ({ navigation }: Props) => {
...
+      }
+      inputForm.current.focus()
       return
     }
 
-    const isValid = assertValidInput()
-
-    if (!isValid) {
+    if (!assertValidInputCharacters()) {
+      setMessage(translate("UsernameScreen.letterAndNumber"))
```
bodymindarts, line 84: Instead of showing an error message can't we just disallow all non-accepted characters at the input?
So if I type:
```
abc%12e
```
the input box would just not accept `%` as input and read:
```
abc12e
```

bodymindarts (commented):
I think we can get rid of the whole `valid character` errors by changing:
```
  const onChangeText = (value) => {
    setInput(value)

    if (!shouldShowErrorMessage && value.length > 2) {
      setShouldShowErrorMessage(true)
    }
  }
```
To strip illegal chars before the call to `setInput`.

app/screens/settings-screen/username-screen.tsx
jrbuhl93: The one problem with this approach is a [limitation of React Native.](https://github.com/facebook/react-native/issues/28535#issuecomment-611196270) The text will flicker when adding one of the disallowed characters. Say my input is "Justin" and i type '?'. The screen will briefly show "Justin?" before rendering back to "Justin".

bodymindarts: I see, didn't know that about react-native. In web it works fine.

But is a Flickr worse than a warning? Is there another way to improve UX here.... perhaps limit the characters on the keyboard?

app/screens/settings-screen/username-screen.tsx
jrbuhl93: It's a little clunky to update `shouldShowErrorMessage` with `loadingUserNameExists`. I think it would be cleanest to rename `shouldShowErrorMessage` to `shouldShowCharacterMinimumErrorMessage`.

app/screens/settings-screen/username-screen.tsx
jrbuhl93: I looked into the keyboard types available for inputs and none match what we want. I personally prefer the warning, but I it's not by much and subjective.

bodymindarts: Do what you prefer

Issue #85, opened 2021-07-12
Cache dismiss address in clipboard


nicolasburtey:
If we keep coming back to the wallet after switching to some other app, we should only ask once. If the user dismiss the address once, we should not ask again if the same address is in the clipboard

PR #89, opened 2021-07-13
Cache last valid payment in clipboard

jrbuhl93:
Fixes issue #85

### DIFF
--- a/app/components/modal-clipboard/modal-clipboard.tsx
+++ b/app/components/modal-clipboard/modal-clipboard.tsx
@@ -14,8 +14,8 @@ import { color } from "../../theme"
 import { palette } from "../../theme/palette"
 import { validPayment } from "../../utils/parsing"
 import { Token } from "../../utils/token"
-import { saveString } from "../../utils/storage"
-import { LAST_CLIPBOARD_PAYMENT } from "../../utils/clipboard"
+import { LAST_CLIPBOARD_PAYMENT } from "../../graphql/client-only-query"
+import { cache } from "../../graphql/cache"
 
 const styles = StyleSheet.create({
   buttonContainer: {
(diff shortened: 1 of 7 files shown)


app/components/modal-clipboard/modal-clipboard.tsx
```diff
@@ -72,12 +76,13 @@ export const ModalClipboard = () => {
...
   const open = async () => {
-    dismiss()
+    modalClipboardVisibleVar(false)
     navigation.navigate("sendBitcoin", { payment: await Clipboard.getString() })
   }
 
-  const dismiss = () => {
+  const dismiss = async () => {
     modalClipboardVisibleVar(false)
+    saveString(LAST_CLIPBOARD_PAYMENT, await Clipboard.getString())
```
bodymindarts, line 85: Why are we not using Apollo for storing this?

app/utils/clipboard.ts
```diff
@@ -1,18 +1,26 @@
 import Clipboard from "@react-native-community/clipboard"
+import { ApolloClient } from "@apollo/client"
 
 import { validPayment } from "./parsing"
 import { modalClipboardVisibleVar, walletIsActive } from "../graphql/query"
 import { Token } from "./token"
-import { ApolloClient } from "@apollo/client"
+import { loadString } from "./storage"
+import { LAST_CLIPBOARD_PAYMENT } from "../components/modal-clipboard"
```
bodymindarts, line 8: This creates a circular dependency. Better to define the const here so that dependencies can be a DAG.

app/components/modal-clipboard/modal-clipboard.tsx
jrbuhl93: After thinking about this more I believe continuing to use AsyncStorage directly with saveString and loadString does not seem bad for client only state. In order to persist the cache we are using [apollo3-cache-persist](https://github.com/apollographql/apollo-cache-persist) and [Async Storage](https://github.com/react-native-async-storage/async-storage). Thoughts?

jrbuhl93: The alternative is trying to get something like this to work: https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies/#storing-local-state-in-the-cache

PR #93, opened 2021-07-16
Testing and prettier more similar to backend

jrbuhl93:
My first pass at making prettier more similar to the backend and adding testing to github actions. Do we want prettier to also format json files on this repo?


__tests__/addresses.ts
```diff
@@ -10,118 +10,149 @@ beforeAll(() => {
...
+      client,
+    )
     expect(valid).toBeTruthy()
     expect(paymentType).toBe("onchain")
     expect(amount).toBeFalsy()
     expect(amountless).toBe(true)
-  }
+  })
 
+  it("prefix", () => {
```
bodymindarts, line 99: As far as I can tell this is a new test and not just a new formatting.

I'm going to approve to move on but in general can we keep PRs focused to 1 thing. A formatting PR should not add a test.

bodymindarts (approved):

PR #98, opened 2021-07-17
proof of concept for billpay

nicolasburtey:
The main idea here is to help integration with utilities companies to settle their bill through the wallet. just a poc for now.

### DIFF
--- a/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
+++ b/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
@@ -81,7 +81,7 @@ const getBillpayData = (client) => {
         query: {
           amount: data.balance,
           memo: `billpay ${data.customer}-${client}`,
-          currency: data.currency
+          currency: data.currency,
         },
       }
 
--- a/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
+++ b/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
@@ -7,11 +7,13 @@ import EStyleSheet from "react-native-extended-stylesheet"
 import { launchImageLibrary } from "react-native-image-picker"
 import Svg, { Circle } from "react-native-svg"
 import Icon from "react-native-vector-icons/Ionicons"
+import url from "url"
 import { Screen } from "../../components/screen"
 import { translate } from "../../i18n"
 import { palette } from "../../theme/palette"
 import { validPayment } from "../../utils/parsing"
 import { Token } from "../../utils/token"
+import { LNPAY_DOMAIN } from "../../utils/helper"
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const LocalQRCode = require("@remobile/react-native-qrcode-local-image")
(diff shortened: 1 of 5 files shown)


app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -72,6 +99,26 @@ export const ScanningQRCodeScreen = () => {
...
+      setPending(true)
+      try {
+        Alert.alert(`fetching billpay ${data}`)
+        const billpayEncoded = await getBillpayData(data)
+
+        if (billpayEncoded) {
+          Alert.alert(JSON.stringify(billpayEncoded))
+          // const d = { username: "nicolas", paymentType: "username" }
+          navigate("sendBitcoin", { payment: billpayEncoded })
+          return
```
jrbuhl93, line 112: If billpayEncoded is false or errors the function won't return yet and run the validPayment function. Is this intended behavior?

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -62,6 +64,31 @@ const styles = EStyleSheet.create({
...
+        host: LNPAY_DOMAIN,
+        pathname: data.username,
+        query: {
+          amount: data.balance,
+          memo: `billpay ${data.customer}-${client}`,
+          currency: data.currency,
+        },
+      }
+
+      return url.format(urlFormat)
```
jrbuhl93, line 88: It seems a little clunky to format all the information into a url string and then decode it again down in the parsing function without using it to make a network request. Would passing the data in a typed object instead of a string be a worthwhile refactoring in the future? It makes sense to leave it as a string for now because it's how all the functions on the send bitcoin screen are set up.

app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
```diff
@@ -236,10 +236,23 @@ export const SendBitcoinScreen: React.FC<SendBitcoinScreenProps> = ({
...
+    const { valid, username, amount, memo, currency } = validPayment(
+      route.params?.payment,
+      network,
+      client,
+    )
     if (route.params?.username || username) {
       setInteractive(false)
       setDestination(route.params?.username || username)
+      if (currency === "USD") {
+        const sats = Math.floor(amount / satusd)
```
jrbuhl93, line 248: should this be amount * satusd?

app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
```diff
@@ -236,10 +236,23 @@ export const SendBitcoinScreen: React.FC<SendBitcoinScreenProps> = ({
...
-    const { valid, username } = validPayment(route.params?.payment, network, client)
+    const { valid, username, amount, memo, currency } = validPayment(
+      route.params?.payment,
+      network,
+      client,
+    )
     if (route.params?.username || username) {
       setInteractive(false)
       setDestination(route.params?.username || username)
+      if (currency === "USD") {
```
jrbuhl93, line 247: Do you need this conversion for the new billpay? I don't think it will be called since you only pass payment and not username when navigating to route params.

jrbuhl93 (commented):
I haven't tested out the bill pay feature, only looked at the code. Would you like me to test it?

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -72,6 +99,26 @@ export const ScanningQRCodeScreen = () => {
       return
     }
 
+    if (data.startsWith("200")) {
```
nicolasburtey, line 102: beginning of the scan/billpay

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -72,6 +99,26 @@ export const ScanningQRCodeScreen = () => {
...
+      setPending(true)
+      try {
+        Alert.alert(`fetching billpay ${data}`)
+        const billpayEncoded = await getBillpayData(data)
+
+        if (billpayEncoded) {
+          Alert.alert(JSON.stringify(billpayEncoded))
+          // const d = { username: "nicolas", paymentType: "username" }
+          navigate("sendBitcoin", { payment: billpayEncoded })
```
nicolasburtey, line 111: field should be written as read only here

PR #102, opened 2021-07-19
Eslint check

jrbuhl93:
First pass at fixing the warnings from eslint. Added a lot of typing including to navigation. 
Edit: The react hooks warnings fixes are now present in the PR.


bodymindarts (approved):

app/screens/earns-map-screen/earns-map-screen.tsx
```diff
@@ -89,7 +92,7 @@ interface ISectionData {
 }
 
 interface IEarnMapScreen {
-  navigation: Record<string, any> // FIXME
+  navigation: StackNavigationProp<PrimaryStackParamList, "Earn"> // FIXME
```
cem2ran, line 95: `FIXME` can be removed right? :)

Issue #99, opened 2021-07-18
Size of the text for amount for the Send Bitcoin should be bigger


nicolasburtey:
(look at video from galoy slack#mobile for more context)

Issue #100, opened 2021-07-18
amounts should be set in cts, not dollar


nicolasburtey:
Instead of having to type "2" "." "5" "0" to receive $2.50, we should have the amount in cents, such that typing 250 would have an amount of $2.50

video from galoy slack#mobile would give more content around this

PR #106, opened 2021-07-22
feat: amounts should be set in cts, not dollar

cem2ran:
## Summary
Implements enhancement proposed in #100.

~As far as I can tell it is not possible to avoid the flickering issue when adding values on the right hand side of the decimal point without overlaying a Text field that covers/masks the input field thus fixing the cursor to the end and sacrificing the ability to do selections and interact with the input field in ways that one may expect. Let me know if you think we should go with the masked approach, then I'll revise this PR. @nicolasburtey @jrbuhl93~ **edit:** New approach below

~More details about this issue here: https://github.com/CaioQuirinoMedeiros/react-native-currency-input#fakecurrencyinput~ 

<details>
  <summary>Prior approach demo</summary>
 
https://user-images.githubusercontent.com/959142/126685362-42163139-fe99-4b14-bf3a-4f9af28d27a4.mp4
</details>

### Changes

1. Utilizes a masked input field approach to give the user a flicker free experience: A Text component is overlayed on top of an invisible TextInput. The TextInput is technically still selectable. I couldn't find any trivial way of disabling this, but I think this is a minor issue.
2. Input field is no longer empty. Either is 0.00 for USD or 0 for sats. I have removed the placeholder text for the empty string/empty as this no longer is a valid state.
3. Formatting improvement for USD. As an example _fifty thousand_ is no longer formatted as `50000.00` but rather `50,000.00`. To enable this on Android a polyfill was necessary. See: https://stackoverflow.com/a/55861894. It is possible to have it formatted as `$50,000.00` by changing the style option of `toLocaleString` to currency and [specifying the additionally required properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#using_options). 

**New approach** (Masked input):

https://user-images.githubusercontent.com/959142/126820798-182f9ad8-74d6-4220-8e7a-4e013cdc37de.mp4

Tested on:
- [x] iOS
- [x] Android

### DIFF
--- a/app/components/input-payment/input-payment.tsx
+++ b/app/components/input-payment/input-payment.tsx
@@ -9,10 +9,13 @@ import { TouchableOpacity } from "react-native-gesture-handler"
 import Icon from "react-native-vector-icons/Ionicons"
 import { btc_price } from "../../graphql/query"
 import { usePrefCurrency } from "../../hooks/usePrefCurrency"
-import { translate } from "../../i18n"
 import { palette } from "../../theme/palette"
 import type { ComponentType } from "../../types/jsx"
-import { CurrencyConversion, toCurrency } from "../../utils/currencyConversion"
+import {
+  CurrencyConversion,
+  currencyToText,
+  textToCurrency,
+} from "../../utils/currencyConversion"
 import { CurrencyType } from "../../utils/enum"
 import { TextCurrency } from "../text-currency/text-currency"
 
@@ -29,11 +32,19 @@ const styles = EStyleSheet.create({
     width: "100%",
   },
 
+  inputMaskPositioning: {
+    position: "absolute",
+    width: "100%",
+  },
+
+  inputText: {
+    opacity: 0,
+  },
+
   main: {
     alignItems: "center",
     flexDirection: "row",
     marginTop: "8rem",
-    width: "100%",
   },
 
   subCurrencyText: {
@@ -94,13 +105,13 @@ export const InputPayment: ComponentType = ({
 }: InputPaymentDataInjectedProps) => {
   const [amount, setAmount] = React.useState(initAmount)
   const [input, setInput] = React.useState("")
-
-  const handleChange = (text) => {
-    setInput(currency === CurrencyType.USD || currency === CurrencyType.BTC ? toCurrency(text) : text)
-  }
   const mapping = CurrencyConversion(price)
   const amountInput = mapping[prefCurrency].conversion(amount)
   const currency = mapping[prefCurrency].primary
+
+  const handleTextInputChange = (text) => {
+    setInput(textToCurrency(text, currency))
+  }
 
   React.useEffect(() => {
     setAmount(initAmount)
@@ -112,19 +123,19 @@ export const InputPayment: ComponentType = ({
       setAmount(newAmount)
       onUpdateAmount(toInteger(newAmount))
     }
-  // note: adding additional dependencies here breaks the input field
-  // when switching between usd & sats
-  // eslint-disable-next-line react-hooks/exhaustive-deps
+    // note: adding additional dependencies here breaks the input field
+    // when switching between usd & sats
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [input])
 
   // is Focused part
 
   React.useEffect(() => {
     // TODO: show "an amount is needed" in red
-    if (forceKeyboard && (amountInput == "" || amountInput == "." || +amountInput == 0)) {
+    if (forceKeyboard && +amountInput == 0) {
       inputRef?.current.focus()
     }
-  }, [editable, forceKeyboard, amountInput])
+  }, [forceKeyboard, amountInput])
 
   const inputRef = React.useRef<TextInput>()
 
@@ -140,15 +151,18 @@ export const InputPayment: ComponentType = ({
   const _keyboardDidHide = () => {
     inputRef?.current?.blur()
   }
+  const displayValue = currencyToText(amountInput, currency)
 
   return (
     <View style={styles.container}>
       <View style={styles.main}>
+        <Text style={[styles.textStyle, styles.inputMaskPositioning]}>
+          {displayValue}
+        </Text>
         <Input
           ref={inputRef}
-          placeholder={translate("common.setAnAmount")}
           autoFocus={forceKeyboard}
-          value={amountInput}
+          value={displayValue}
           leftIcon={
             currency === CurrencyType.USD ? (
               <Text
@@ -175,7 +189,7 @@ export const InputPayment: ComponentType = ({
               <Text
                 style={[
                   styles.textStyle,
-                  { color: input === "" ? palette.midGrey : palette.darkGrey },
+                  { color: amount === 0 ? palette.midGrey : palette.darkGrey },
                 ]}
               >
                 sats
@@ -183,9 +197,10 @@ export const InputPayment: ComponentType = ({
             ) : null
           }
           inputContainerStyle={styles.inputContainer}
-          inputStyle={[styles.textStyle]}
-          onChangeText={handleChange}
-          keyboardType={currency === "sats" ? "number-pad" : "decimal-pad"}
+          inputStyle={[styles.textStyle, styles.inputText]}
+          contextMenuHidden
+          onChangeText={handleTextInputChange}
+          keyboardType={"number-pad"}
           onBlur={onBlur}
           enablesReturnKeyAutomatically
           returnKeyLabel="Update"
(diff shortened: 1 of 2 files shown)


cem2ran:
I think the masked input would go great with a UI where the input is more prominent and the center of attention, like this example: https://www.figma.com/file/aLAa3OuBW74EcSWxjjY82m/Bitcoin-Wallet-UI-Kit-Community?node-id=831%3A18002

app/components/input-payment/input-payment.tsx
```diff
@@ -88,25 +88,27 @@ export const InputPayment = ({
   prefCurrency,
   nextPrefCurrency,
 }: InputPaymentDataInjectedProps) => {
-  const endByDot = (s: string) => s.match(/^[0-9]*\.{1}$/)
-
   const [amount, setAmount] = React.useState(initAmount)
   const [input, setInput] = React.useState("")
-  const [appendDot, setAppendDot] = React.useState(false)
 
+  const handleChange = (text) => {
```
bodymindarts, line 94: can we be more explicit? `handleTextInputChange` perhaps?

app/utils/currencyConversion.ts
```diff
@@ -41,3 +41,37 @@ export const textCurrencyFormatting = (sats, price, currency) => {
...
+export const toCurrency = (value: string , separator = '.'): string => {
+  const digits = getDigitsFromValue(value)
+  const digitsWithPadding = padDigits(digits)
+  return addDecimalToNumber(digitsWithPadding, separator);
+}
+
+const getDigitsFromValue = (value = '') => value.replace(/(-(?!\d))|[^0-9|-]/g, '') || ''
+
+const padDigits = digits => {
+  const desiredLength = 3
```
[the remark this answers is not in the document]
jrbuhl93: He needs three digits to show the value in dollars (eg $ 0.09 instead of $ 9). But the constant could be named more explicitly to show it's purpose.

cem2ran:
Here is an example of how the masked input would look and behave like: ([source](https://codeburst.io/currency-input-for-react-de1c1f1a5877))

image

app/components/input-payment/input-payment.tsx
```diff
@@ -92,12 +92,12 @@ export const InputPayment: ComponentType = ({
...
+    setInput(currency === CurrencyType.USD || currency === CurrencyType.BTC ? toCurrency(text) : text)
```
jrbuhl93, line 99: I think instead of doing a check here you could pass `currency` to the function and keep more logic in the `toCurrency` function.

jrbuhl93 (commented):
I think we should do the mask as well. The flickering is bad aesthetically, and it is also confusing.

app/utils/currencyConversion.ts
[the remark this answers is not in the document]
bodymindarts: Perhaps we can create a type like “USDDisplayConfig”. With those fields set. Something that we can easily create more of four different currencies

nicolasburtey:
without looking in the PR (yet) I can say that I spent some time on this (maybe 2 or 3 days) and struggled to come up with something that really worked well. glad this is being worked out. 

I think the flickering is indeed not great. it give an impression of a "cheap" app.

cem2ran:
> without having a look into the PR (yet) I can say that I spent some time on this (maybe 2 or 3 days) and struggled to come up with something that really worked well. glad this is being worked out.

Was quite surprised with the amount of complexity here. May be able to simplify it a bit if we make the suggested changes below:

> I think the flickering is indeed not great. it give an impression of a "cheap" app.

Agree. Will make this change. Do we want this to only apply to the USD view or also sats for consistency?

Furthermore, should we remove the decimal pad and only use number pad?

jrbuhl93:
> Agree. Will make this change. Do we want this to only apply to the USD view or also sats for consistency?

I think it's best to give a consistent feel between currencies.
> Furthermore, should we remove the decimal pad and only use number pad?
I say use number pad for usd after this change.

app/utils/currencyConversion.ts
```diff
@@ -41,3 +41,37 @@ export const textCurrencyFormatting = (sats, price, currency) => {
...
+  const padding = '0'.repeat(amountToAdd)
+
+  return padding + digits
+}
+
+const removeLeadingZeros = number => number.replace(/^0+([0-9]+)/, '$1')
+
+const addDecimalToNumber = (number: string, separator: string) => {
+  const centsStartingPosition = number.length - 2
+  const dollars = removeLeadingZeros(
```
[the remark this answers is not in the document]
cem2ran: Went with Integers and Fractions. Appears to also be the terminology used in Intl.NumberFormat API.

app/utils/currencyConversion.ts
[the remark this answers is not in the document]
cem2ran: padDigits appears to no longer be needed, or at least I couldn't find a case where it was needed to format the values with the masked input approach.

cem2ran:
@nicolasburtey I think this is ready for you to try out/review!

app/utils/polyfill.ts
```diff
@@ -0,0 +1,10 @@
+import { Platform } from "react-native"
+
+// Polyfill for Intl.NumberFormat. 
+// See: https://stackoverflow.com/a/55861894
+if(Platform.OS === 'android') { // only android needs polyfill
+  // eslint-disable-next-line import/no-unassigned-import
+  require('intl'); // import intl object
+  // eslint-disable-next-line import/no-unassigned-import
+  require('intl/locale-data/jsonp/en-US'); // load the required locale details
+}
```
jrbuhl93, line 10: add automatic new line at EOF in your text editor to not have this problem

app/utils/polyfill.ts
```diff
@@ -0,0 +1,10 @@
+import { Platform } from "react-native"
+
+// Polyfill for Intl.NumberFormat. 
+// See: https://stackoverflow.com/a/55861894
+if(Platform.OS === 'android') { // only android needs polyfill
+  // eslint-disable-next-line import/no-unassigned-import
```
jrbuhl93, line 6: @bodymindarts @samerbuna Is there another syntax for this import eslint will be happy with? Otherwise can we remove [this](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unassigned-import.md) rule?

jrbuhl93 (commented):
I like the masked input a lot! A great improvement.

Without the cursor to act as a guide for the user, should we have another visual cue that the field is selected like changing the color of the underline of the input?

When I changed between sats and usd the keyboard was dismissed. I think we should keep it up after switching.

cem2ran:
> Without the cursor to act as a guide for the user, should we have another visual cue that the field is selected like changing the color of the underline of the input?

Sounds good to me. Maybe just a darker tint of the existing color or should we use something more vibrant?

> When I changed between sats and usd the keyboard was dismissed. I think we should keep it up after switching.

Made this change by essentially [removing some of the code](https://github.com/GaloyMoney/galoy-mobile/pull/106/commits/298df2133c70cf04d35b95cf68babc142afe32b1). @nicolasburtey is it safe to remove this, I don't see a need for it?

https://github.com/GaloyMoney/galoy-mobile/blob/2e8b839e6e2222e713aa8415d96512e3a7f6d349/app/components/input-payment/input-payment.tsx#L130-L141

nicolasburtey:
I had gave it a try. I was a bit confuse at first because there is no more cursor so it's unclear we are editing the field. but after a few try I got used to it. 

I feel overall this is much better but I'm wondering if there are some insights to give to the user to suggest him the focus is on the price and not the memo for instance.

There is some misalignement on my iPhone (X). maybe this could be the address at the same time: https://github.com/GaloyMoney/galoy-mobile/issues/99

IMG_6C69E60F0B25-1

jrbuhl93:
> Sounds good to me. Maybe just a darker tint of the existing color or should we use something more vibrant?

I think we should do something a little more vibrant and use one of the the app's blue colors. We might need it on the other input fields that aren't masked in order to maintain consistency, but that could be out of scope for this PR.

cem2ran:
Could also make amount text larger when input field is focused. Initially it would be large as input field is autofocused and when you go to the next field it would shrink (nicely animated of course). Maybe this would also resolve #99?

Should I give this one a shot or do we want to go with the colored underline on focus despite the inconsistency?

Could also add a faux blinking thing at the end of the masked input, but would prefer the larger input text to this.

app/utils/polyfill.ts
samerbuna: I'd rather keep the rule (which is good for our own code) and specify an `allow` list for things we don't have control over.

jrbuhl93:
> Should I give this one a shot or do we want to go with the colored underline on focus despite the inconsistency?

I personally prefer the underline, but if you or @nicolasburtey prefer growing and shrinking the text you should do that.

bodymindarts:
Input from Nicolas:
- good enough for merge since its already better than what we have
- potentially highlight text that is being edited via "color bubble" but can be addressed in additional PR
- need to increase font size as its hard to read

PR #112, opened 2021-07-25
Fix es translations

dolcalmi:
@jrbuhl93 apologies for the messy PR, you just need to check the latest commit changes

- first commit: only sort of the keys in the translation files (no content changes) https://github.com/GaloyMoney/galoy-mobile/commit/d1029f41e5fb01d4f995d963a921c175ffd0f5be 
- Add missing translations, fix some translations https://github.com/GaloyMoney/galoy-mobile/pull/112/commits/029bfb4ba8005569eb9205e810dabb556c1d0b2a

### DIFF
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -1,62 +1,636 @@
 {
-  "common": {
-    "ok": "OK",
-    "confirm": "Confirm",
-    "cancel": "Cancel",
-    "back": "Back",
-    "next": "Next",
-    "error": "Error",
-    "close": "Close",
-    "bankAccount": "Cash Account",
-    "firstName": "First Name",
-    "lastName": "Last Name",
-    "username": "Username",
-    "logout": "Log Out",
-    "loggedOut": "You have been logged out.",
-    "email": "Email",
-    "sats": "sats",
-    "to": "To",
-    "bank": "Bank",
-    "bitcoin": "Bitcoin",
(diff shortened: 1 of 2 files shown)


nicolasburtey:
@dolcalmi I just realized that the earns section with the quiz can't be re-ordered. because the questions/answers in the `"earns": []` array are in the order they are displayed to the user. so going to an alphabetical order here would just completely messed up the education story line

dolcalmi:
> @dolcalmi I just realized that the earns section with the quiz can't be re-ordered. because the questions/answers in the `"earns": []` array are in the order they are displayed to the user. so going to an alphabetical order here would just completely messed up the education story line

are there other sections that needs to be in specific order?

jrbuhl93:
> are there other sections that needs to be in specific order?

I am pretty sure there aren't any other parts of the translations where order matters. As far as I know every other call to translation will just return a string instead of a nested object.

dolcalmi:
ok, just checked the files and arrays are not sorted (there are not keys to sort them)

Current Order of "earns" (meta id): bitcoinWhatIsIt, WhatIsMoney, HowDoesMoneyWork and BitcoinWhySpecial

Also did the check of "content" arrays and they also keep the same order.

jrbuhl93 (approved):
The translations look good to me. This PR will improve UX a lot.

PR #128, opened 2021-07-31
Add underline when focused to text inputs

jrbuhl93:
Change the underline color to blue when a textinput is focused to provide visual feedback after #106. I added the underlines to the other textinputs in the send and receives screens to maintain consistency.

https://user-images.githubusercontent.com/35902932/127748768-3edfc6f2-b908-49c7-8409-329ce02e1327.mp4


app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
```diff
@@ -600,188 +604,209 @@ export const SendBitcoinScreenJSX: ScreenType = ({
...
-            !potentialBitcoinOrLightning &&
-            paymentType === "username" ? (
-              loadingUserNameExist ? (
-                <ActivityIndicator size="small" />
-              ) : UsernameValidation.isValid(destination) && usernameExists ? (
-                <Text>✅</Text>
+}: SendBitcoinScreenJSXProps) => {
+  const [isDestinationFocused, setIsDestinationFocused] = useState(false)
+  const [isMemoFocused, setIsMemoFocused] = useState(false)
+  const [isFeeFocused, setIsFeeFocused] = useState(false)
```
cem2ran, line 610: I would consider creating a Galoy specific Input component and encapsulate this type of logic within it. `input-payment.tsx` could also use that as a base.

jrbuhl93: I agree. Do you like the design of the underline? If so, I can work on encapsulating the input.

PR #129, opened 2021-07-31
Add polyfill

jrbuhl93:
This got lost when I rebased #106. Adding it back in here for another immediate release.


nicolasburtey:
why is this (now) needed?

jrbuhl93:
Without this the digits in the new text input aren't working, so zeros are being removed. You can't type $9.00 for instance because of the trailing zeros.

jrbuhl93:
I added some unit tests for `currencyToText`, but they won't catch this bug. The bug is due to the version of Javascript on Android not having the needed dependencies, and a local environment's laptop will have those dependencies.

PR #130, opened 2021-07-31
Increase input payment font size to 28rem

jrbuhl93:
Fixes issue #99. Thoughts on the new size?
Screenshot_20210731-161639
Screenshot_20210731-162016


nicolasburtey:
I think it could be even bigger. like 25% to 50% bigger.

nicolasburtey:
maybe use the typo in bold?

jrbuhl93:
The next iteration:
Screenshot_20210801-112958

nicolasburtey:
ok let's go for it for now. I think this is the type of things where we would benefits from the help of the bitcoin design group.

PR #138, opened 2021-08-02
Check for null when reading clipboard cache

jrbuhl93:
Fixes a bug when reading `lastClipboardPayment` from apollo cache when the value is null.

### DIFF
--- a/app/utils/clipboard.ts
+++ b/app/utils/clipboard.ts
@@ -19,8 +19,7 @@ export const showModalClipboardIfValidPayment = async (
   }
 
   const data = cache.readQuery<LastClipboardPayment>({ query: LAST_CLIPBOARD_PAYMENT })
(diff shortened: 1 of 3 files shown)


app/utils/clipboard.ts
```diff
@@ -17,7 +18,8 @@ export const showModalClipboardIfValidPayment = async (
     return
   }
 
-  const { lastClipboardPayment } = cache.readQuery({ query: LAST_CLIPBOARD_PAYMENT })
+  const data = cache.readQuery<LastClipboardPayment>({ query: LAST_CLIPBOARD_PAYMENT })
+  const lastClipboardPayment = data?.lastClipboardPayment ?? ""
```
dolcalmi, line 22: maybe we just need
```js
if (clipboard === data?.lastClipboardPayment)
...
```

dolcalmi (approved):

PR #139, opened 2021-08-06
Unit tests for SendBitcoinScreen

jrbuhl93:
- Set up the project for unit testing with the [react native testing library](https://github.com/callstack/react-native-testing-library)
- Added unit tests for the SendBitcoinScreen including one that would have caught the bug that necessitated #124 
- Added unit tests for username validation
- Fixed typescript errors that prevented the component (SendBitcoinScreen) unit test from compiling


__mocks__/@react-native-async-storage/async-storage.js
```diff
@@ -0,0 +1,2 @@
+import AsyncStorageMock from "@react-native-async-storage/async-storage/jest/async-storage-mock"
+export default AsyncStorageMock
```
dolcalmi, line 2: you can also use 

```js
export { AsyncStorageMock as default } from "@react-native-async-storage/async-storage/jest/async-storage-mock"
```

__tests__/addresses.ts
```diff
@@ -119,9 +119,10 @@ describe("Lightning", () => {
     const address =
```
dolcalmi, line 119: I guess we need to create some helpers

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,352 @@
...
+import "@testing-library/jest-native/extend-expect"
+import "../../node_modules/react-native-gesture-handler/jestSetup.js"
+
+import { translate } from "../../app/i18n/translate"
+import "../../__mocks__/react-native-firebase"
+import {
+  LIGHTNING_PAY,
+  PAY_KEYSEND_USERNAME,
+  SendBitcoinScreen,
+} from "../../app/screens/send-bitcoin-screen"
```
dolcalmi, line 15: we can use moduleNameMapper to avoid this :)

https://github.com/GaloyMoney/galoy/blob/main/jest.config.js#L12

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,352 @@
...
+import { QUERY_PRICE, WALLET } from "../../app/graphql/query"
+
+jest.mock("@react-navigation/native", () => {
```
dolcalmi, line 18: is this a common mock to multiple test suites? if so, please add it in a setup or in its own mock file

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,352 @@
...
+    pending: false,
+    text: "$0.0004",
+    type: "on_us",
+    usd: 0.0003693000390625,
+    username: "BitcoinBeachMarketing",
+  },
+]
+
+const cache = new InMemoryCache()
+cache.writeQuery({
```
dolcalmi, line 90: we need some helpers here but I think we can do it after test structure refactor

__tests__/username-validation.ts
```diff
@@ -0,0 +1,73 @@
+import { UsernameValidation } from "../app/utils/validation"
+
+it("empty", () => {
```
dolcalmi, line 3: please use another description, I think is a general request for the tests in this file

app/screens/earns-screen/earns-section.tsx
```diff
@@ -176,7 +176,7 @@ export const EarnSection: ScreenType = ({ route, navigation }: Props) => {
 
   const open = async (card) => {
     // FIXME quick fix for apollo client refactoring
```
dolcalmi, line 178: is this comment needed?

app/utils/token.ts
```diff
@@ -76,8 +84,8 @@ export class Token {
 
 export const getNetwork = async (): Promise<INetwork> => {
   let network
-  if (new Token().has()) {
-    network = new Token().network
+  if (Token.getInstance().has()) {
+    network = Token.getInstance().network
   } else {
     network = await loadNetwork()
   }
```
dolcalmi, line 91: in general else blocks can be avoided (cyclomatic complexity mentioned by Justin C in code review)
```js
if (Token.getInstance().has()) {
  network = Token.getInstance().network
}

network = network || await loadNetwork()
```
It is not a request for change but is a good example for future code

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "start": "react-native start",
-    "test": "jest --runInBand",
+    "test": "jest --runInBand --silent",
```
dolcalmi, line 7: I think is better remove --silent, if you want to reduce the messages just change the loglevel

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -74,7 +74,11 @@ export const ScanningQRCodeScreen: ScreenType = () => {
     }
 
     try {
-      const { valid, errorMessage } = validPayment(data, new Token().network, client)
+      const { valid, errorMessage } = validPayment(
+        data,
+        Token.getInstance().network,
+        client,
+      )
       console.log({ valid, errorMessage, data }, "result")
```
dolcalmi, line 82: please remove console logs (at least in the code you modify)

app/components/text-currency/text-currency.tsx
```diff
@@ -45,7 +45,6 @@ export const TextCurrency: ComponentType = ({ amount, currency, style }: Props)
     <Text style={style}>
       {currency_fmt
         .default(amount, {
-          formatWithSymbol: false,
```
dolcalmi, line 48: dont we need the symbol? or is related to the "format" issue?

dolcalmi (changes requested):

app/components/text-currency/text-currency.tsx
jrbuhl93: `formatWithSymbol` is not accepted by this library. Perhaps it was deprecated? We don't want the symbol here as can be seen by `symbol: ""` below.

app/screens/earns-screen/earns-section.tsx
jrbuhl93: We will refactor token storage in the mobile app and probably use apollo client. I think (?) this comment is referring to that. There are a lot of fixmes (26 in 18 files) throughout the mobile codebase, and I don't want to remove them before addressing them.

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,342 @@
...
+      }
+    }
+  `,
+  data: {
+    me: {
+      username: "BitcoinBeach",
+    },
+  },
+})
+
```
samerbuna, line 138: I think you can combine the multiple `writeQuery` calls into a single one with a combined query and a combined data object. Not sure that's better (in a test) though.

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,342 @@
...
+afterEach(cleanup)
+
+describe("SendBitcoinScreen", () => {
+  it("can render", () => {
+    render(
+      <MockedProvider cache={cache}>
+        <SendBitcoinScreen route={{ params: null }} />
+      </MockedProvider>,
+    )
```
samerbuna, line 168: How about some snapshot testing here

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,342 @@
...
+    const { queryByPlaceholderText, queryByText } = render(
+      <MockedProvider cache={cache}>
+        <SendBitcoinScreen route={{ params: null }} />
+      </MockedProvider>,
+    )
+
+    expect(queryByText("0.00")).not.toBeNull()
+    expect(queryByPlaceholderText(translate("SendBitcoinScreen.input"))).not.toBeNull()
+    expect(queryByPlaceholderText(translate("SendBitcoinScreen.note"))).not.toBeNull()
+    expect(queryByPlaceholderText(translate("SendBitcoinScreen.fee"))).not.toBeNull()
```
samerbuna, line 181: Should we assert values are `HTMLElement` here instead?

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,342 @@
...
+    const amountInput = getByText("0.00")
+    const destinationInput = getByPlaceholderText(translate("SendBitcoinScreen.input"))
+
+    fireEvent.changeText(destinationInput, "Bitcoin")
+    fireEvent(amountInput, "onUpdateAmount", 252119)
+
+    const sendButton = getByText(translate("common.send"))
+    fireEvent.press(sendButton)
+
+    await act(async () => await new Promise((resolve) => setTimeout(resolve, 0)))
```
samerbuna, line 253: Move this into a helper?

__tests__/screens/send-bitcoin-screen.tsx
```diff
@@ -0,0 +1,342 @@
...
+              __typename: "Invoice",
+            },
+          },
+        },
+      },
+    ]
+
+    moment.now = function () {
+      return 1598110996000 // Aug 22 2020 10:43
+    }
```
samerbuna, line 315: Not very familiar with `moment` (which I think we should git rid of, really), but is this time traveling for something that uses `moment` in the code? Can we maybe wrap that with a clear function and mock the function instead?

app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
```diff
@@ -709,14 +719,12 @@ export const SendBitcoinScreenJSX: ScreenType = ({
...
-              <Text>{translate("SendBitcoinScreen.feeCalculationUnsuccessful")}</Text> // todo: same calculation as backend
-            ) : (
-              <TextInput {...props} />
-            )
+          InputComponent={
+            fee === undefined
+              ? FeeActivityIndicator
+              : fee === -1
+              ? FeeCalculationUnsuccessfulText
+              : TextInput
```
samerbuna, line 727: This double ternary here is a bit hard to digest

__tests__/screens/send-bitcoin-screen.tsx
jrbuhl93: This gets a little tricky due to only being able to mock [the exported function and not internal functions](https://stackoverflow.com/a/55193363). I am going to go with importing the module into itself for now, but I am not sure if that is a good practice.

__tests__/screens/send-bitcoin-screen.tsx
jrbuhl93: The correct way to check type with jest seems to be to use [toBeInstanceOf](https://stackoverflow.com/a/62564851), but this returns the error `'HTMLElement' only refers to a type, but is being used as a value here.` It seems that these type errors should be caught at [compile time](https://stackoverflow.com/a/64113496), so are not needed in the unit test.

app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
jrbuhl93: I only refactored the `feeInputComponent`, but I'll do the others over in another PR if you approve of the refactoring.

__tests__/screens/send-bitcoin-screen.tsx
jrbuhl93: I left this as is for now. But as dolcami mentioned, eventually these will go into helpers.

samerbuna (approved):

__tests__/helpers/wait.ts
```diff
@@ -0,0 +1,5 @@
+import { act } from "@testing-library/react-native"
+
+export const waitForNextRender = async () => {
+  await new Promise((resolve) => setTimeout(resolve, 0))
+}
```
dolcalmi, line 5: you dont have to use async await because you are not using a returned value from an internal promise. Just use

```js
export const waitForNextRender = () => {
  return new Promise((resolve) => setTimeout(resolve, 0))
}
```

dolcalmi (changes requested):

dolcalmi (approved):

PR #142, opened 2021-08-08
fetch wallet updates separate from useLazyQuery

### LINKED ISSUE #141: Balance doesn’t refresh automatically
It seems the problem occurs both when sending a payment and receiving a payment.

For sending the problem doesn’t seems fully deterministic. Salim says the balance would refresh “some times”. Maybe 20% of the time.

Refreshing the wallet by scrolling down/triggering the “refresh” wheel solve the problem. But this is not intuitive for everyone.

the problem is occurring on Android phones. It may matter because the notification system is OS dependant

jrbuhl93:
Fixes #141. I am not sure exactly how useQuery and useLazyQuery work, but it seems if you leave the component before they complete, the data retrieved from them will not update in the cache. By separating the updateWallet query from the component we ensure that the cache is updated when the query completes.

### DIFF
--- a/app/graphql/query.ts
+++ b/app/graphql/query.ts
@@ -97,6 +97,15 @@ export const balanceUsd = (client: ApolloClient<unknown>): number =>
 export const balanceBtc = (client: ApolloClient<unknown>): number =>
   _.find(getWallet(client), { id: "BTC" }).balance
 
+export const getWalletNetworkOnly = async (
+  client: ApolloClient<unknown>,
+): Promise<void> => {
+  await client.query({
+    query: WALLET,
+    fetchPolicy: "network-only",
+  })
+}
+
 export const getPubKey = (client: MockableApolloClient): string => {
   const { nodeStats } = client.readQuery({
     query: gql`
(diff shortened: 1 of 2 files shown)


bodymindarts (approved):

app/graphql/query.ts
```diff
@@ -97,6 +97,15 @@ export const balanceUsd = (client: ApolloClient<unknown>): number =>
 export const balanceBtc = (client: ApolloClient<unknown>): number =>
   _.find(getWallet(client), { id: "BTC" }).balance
 
+export const getWalletNetworkOnly = async (
```
nicolasburtey, line 100: instead of having a `getWalletNetworkOnly`, I'll recommend a `getWallet` function where you can pass a `fetchPolicy` to override the default settings.

nicolasburtey:
> Fixes #141. I am not sure exactly how useQuery and useLazyQuery work, but it seems if you leave the component before they complete, the data retrieved from them will not update in the cache. By separating the updateWallet query from the component we ensure that the cache is updated when the query completes.

useLazyQuery works like a mutation and only execute when you use the function returned. otherwise the query would execute automatically when the screen is being initialized for display

Issue #126, opened 2021-07-30
Have a way to hide the balance


nicolasburtey:
when paying an invoice in person, in many scenarios, people around the payer/payee can see the balance of both the payer/payee. this is not ideal when someone has a large balance. 

we should have a way to hide the balance in the settings, and only show it after clicking on the balance/an eye icon, or something along those lines.

I think this can be a setting only saved in the mobile app.

jrbuhl93:
Added in #181

PR #144, opened 2021-08-08
Hide balance security setting

danwebb1:
This is a potential feature addition for [Issue 126](https://github.com/GaloyMoney/galoy-mobile/issues/126)

In this PR:
-Added storage keys `HIDE_BALANCE` and `WALKTHROUGH_TOOLTIP`
-Added a toggle switch to the `security-settings` screen to switch on to hide balance details by default
-Added language to the `en.json` for descriptions, titles, etc
-Refactored `TextCurrency` component. Now checks the `HIDE_BALANCE` settings
-Added `TouchableHighlight` to reveal/hide balance details
-Added a React Native Walkthrough ToolTip which alerts the user via automatic tooltip popup over the eye icon the first time they navigate to a screen with a concealed balance then deletes the key to avoid subsequent, annoying pop ups.

Tested on iOS and Android locally successfully. 

I tried to follow styles and patterns as closely as I could based on my understanding. I am not sure about your workflows or any specific steps I may have skipped. Happy to make necessary changes as requested.


jrbuhl93:
Thank you for the pull request! I'll look into it in more detail later, but one thing to work on is we are moving away from using async storage directly. The plan is to use apollo client through gql queries like in [this file](https://github.com/GaloyMoney/galoy-mobile/blob/main/app/graphql/client-only-query.ts). Here is [an example of reading](https://github.com/GaloyMoney/galoy-mobile/blob/0185f9e263eeeb6e488509f48470311400032576/app/utils/clipboard.ts#L21) from storage, and here is [an example of writing ](https://github.com/GaloyMoney/galoy-mobile/blob/0185f9e263eeeb6e488509f48470311400032576/app/components/modal-clipboard/modal-clipboard.tsx#L85) to storage. Another part of our workflow is to run `yarn prettier-write` for code formatting.

jrbuhl93:
What do you think about having only one eye icon to tap here to reveal both balances since one is just a conversion of the other? This would also prevent the tooltip from duplicating.
Screenshot_20210809-092311

danwebb1:
> What do you think about having only one eye icon to tap here to reveal both balances since one is just a conversion of the other? This would also prevent the tooltip from duplicating.

I personally think that is preferable and that bugged me but I wanted to see what you guys thought. I will update that when I get a chance later today.

jrbuhl93:
I also noticed that by modifying TextCurrency you have caused the eye to appear other places the TextCurrency component is used. BalanceHeader seems like the component you want to modify instead.
Screenshot_20210809-092318
Screenshot_20210809-092325

jrbuhl93:
Two UI glitches are present:

https://user-images.githubusercontent.com/35902932/128809367-a27640b8-10fe-4e56-97f2-d9db6ddddc27.mp4

One is the black square when toggling the eye. The other is the price screen displaying the balance temporarily before switching to the eye when first rendered. Let me know if you need any help with these.

danwebb1:
Thanks for the feedback, Justin. I have addressed the black square as well as the shifting buttons when the eye is toggled. 
A few things: 
1. I am learning Apollo client and trying to modify the interaction with asyncstorage through Apollo client. I am working out some of the issues with that at the moment.
2. Regarding the price screen displaying for a split second when first rendering, I am hoping to resolve that once I have a better grasp of managing state with Apollo. Alternatively, would a loader icon placeholder while fetching settings be desirable to solve this?

jrbuhl93:
> Alternatively, would a loader icon placeholder while fetching settings be desirable to solve this?

I don't think we should add a second spinner to that screen, because the graph already uses one. Leaving the area blank until the variable loads might not look bad. You could also load the variable prior to navigating to price screen like [here](https://github.com/GaloyMoney/galoy-mobile/blob/529c0aa5170f499a51582598d2395dbad3562ac3/app/screens/settings-screen/settings-screen.tsx#L63).

danwebb1:
1. I have addressed the issue of the balance displaying for a split second when first rendering the home/move-money screen after enabling balance hiding settings.
2. I have added a delay for the tool tip popup alerting the user they can click the eye icon to display their balance
3. I have created functions which handle savings settings now through functions that take inputs and interface with apollo instead of asyncstorage directly.

Open to feedback especially as it regards to my solving for saving settings and handling state.

jrbuhl93:
Some ui problems are still present:

1. The price and settings buttons still move when the balance is loaded.
2. The balance appears for a split second before hiding when going to the price screen.
3. I personally don't like how the eye or balance loads after everything else. What do you think?

https://user-images.githubusercontent.com/35902932/129382973-1d1b7c19-9963-47f1-b3e4-135914cc5f10.mp4

jrbuhl93:
Looks good to me if you fix the check code errors and warnings.

PR #149, opened 2021-08-11
Always set InputPayment cursor position to displayValue length

### LINKED ISSUE #148: cursor on send bitcoin not always start from the right
if when I'm on the send bitcoin screen and I tap on the left side of the field (between the dollar sign and the 0 on the left) to select the cursor, then the cursor is implicitly on the left side and adding a 2 would make a 20.00 instead of 0.02.

we need to make sure that the cursor always start from the right when someone is tapping on the amount field to have a deterministic behavior

https://user-images.githubusercontent.com/11156370/128941683-bbcf69e2-c193-4283-b035-148dc0f5d610.MP4

jrbuhl93:
Fixes #148. Also refactored the icons into functions to make the JSX more readable.

### DIFF
--- a/__tests__/screens/__snapshots__/send-bitcoin-screen.spec.tsx.snap
+++ b/__tests__/screens/__snapshots__/send-bitcoin-screen.spec.tsx.snap
@@ -463,6 +463,12 @@ exports[`SendBitcoinScreen render matches snapshot 1`] = `
                     rejectResponderTermination={true}
                     returnKeyLabel="Update"
                     returnKeyType="done"
+                    selection={
+                      Object {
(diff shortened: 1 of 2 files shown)


jrbuhl93:
A few problems are still present.

1. When going to price (upper left button) the balance will still flicker for a moment before hiding.
2. The price and settings buttons still move to the side once the balance is loaded.
3. I think a good solution for a loading placeholder maybe be the white box used when pulling down to refresh the home (move-money) screen.


https://user-images.githubusercontent.com/35902932/129382218-b34cc8a5-7b7e-4398-9480-7137ebc8739e.mp4

Issue #151, opened 2021-08-11
The amount field on the send and receive screens should not use '...'


jrbuhl93:
We want the exact amount here for the best UX. Limit the number of characters? This may be tricky with differing screen sizes.

PR #154, opened 2021-08-12
Cap amount digits and custom margins for currencies

jrbuhl93:
Fixes issue #151. But there is a catch. This PR causes the text to center differently betweens `sats` and `$`.

https://user-images.githubusercontent.com/35902932/129268102-889b30b7-b4b1-40ab-b597-72761720f846.mp4

This is because the margins to avoid the unit symbols are different.

If the change of text positioning is unacceptable we can either:

1. Leave the margins the same between currencies. This requires using the strictest margins for all currencies. This leaves unused white space (`$` space for an amount in sats and `sats` space for an amount in dollars). This is what the current version of the app does.
2. Dynamically alter the margins as more text is added so that the amount starts in the center, but can fill in the blank space as more digits are added. The downside is that this is complex, and I am not sure if the end product will look good.

### DIFF
--- a/__tests__/screens/__snapshots__/send-bitcoin-screen.spec.tsx.snap
+++ b/__tests__/screens/__snapshots__/send-bitcoin-screen.spec.tsx.snap
@@ -385,6 +385,7 @@ exports[`SendBitcoinScreen render matches snapshot 1`] = `
           "justifyContent": "space-between",
         }
       }
+      keyboardShouldPersistTaps="always"
     >
       <View>
         <View>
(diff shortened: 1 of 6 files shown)


nicolasburtey:
this looks better to me than what we have. 
agree that the jumping is not ideal, but maybe we can still go with this until we improve it?
bitcoin design would provide some help here I believe.

one think that is getting my attention thought is that we don't have any comma for sats (but we do for usd). this would help for the readability.

PR #157, opened 2021-08-13
fix: Add param to get full uri without prefix

### LINKED ISSUE #155: Onchain bitcoin address copy with bitcoin: prefix
**Describe the bug**
when you copy the bitcoin address it copies with `bitcoin:` prefix

**To Reproduce**
1. Go to Receive screen
2. Swipe to get an onchain address
3. Tap on QR code

**Expected behavior**
copy the address without `bitcoin:` prefix

**Smartphone:**
 - Device: iPhoneXs
 - OS: 14.6
 - Version 1.4.9 build 259

mfferreira:
Fixes #155

Added an extra param to be able to get the full uri with (default) or without the `bitcoin:` prefix.

### DIFF
--- a/app/screens/receive-bitcoin-screen/ReceiveBitcoinScreenTypes.d.ts
+++ b/app/screens/receive-bitcoin-screen/ReceiveBitcoinScreenTypes.d.ts
@@ -1,0 +1,5 @@
+interface IGetFullUri {
+  input: string
+  uppercase?: boolean
+  prefix?: boolean
+}
--- a/app/screens/receive-bitcoin-screen/ReceiveBitcoinScreenTypes.d.ts
+++ b/app/screens/receive-bitcoin-screen/ReceiveBitcoinScreenTypes.d.ts
@@ -1,4 +1,4 @@
-interface IGetFullUri {
+type GetFullUriInput = {
   input: string
   uppercase?: boolean
   prefix?: boolean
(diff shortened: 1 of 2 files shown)


app/screens/receive-bitcoin-screen/ReceiveBitcoinScreenTypes.d.ts
```diff
@@ -0,0 +1,5 @@
+interface IGetFullUri {
```
jrbuhl93, line 1: Maybe this should just be a type  instead of an interface since we aren't extending it? `GetFullUriInput`

mfferreira: Sure, I can update it to type.
Would that be the rule of thumb for using types or interfaces?

jrbuhl93: It's a complicated topic [it seems](https://stackoverflow.com/questions/37233735/typescript-interfaces-vs-types), and we shouldn't take my word as gospel. @samerbuna

jrbuhl93: It's possible I was wrong because further down in [the link](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) it says: `For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type.`

mfferreira: Indeed, for the most part it's personal preference, so it's fine either way. I'll just follow the project setup 👌

My force of habit is to use `interface` when it's literally an interface to talk to some code.

jrbuhl93 (approved):

PR #164, opened 2021-08-19
Use hooks for btc price and usd balance

jrbuhl93:
Using hooks and useQuery causes the components to re-render with the updated price when the price is updated in the cache. It will be important to add having usd as a possible reference price (instead of always using sats for the amount variable) for the best user experience. This will be done in another pr.


jrbuhl93:
#170 builds on this with another commit

PR #165, opened 2021-08-20
Issue 135

MikeMwambia-TrojanSystem:
Change pin color


README.md
```diff
@@ -5,8 +5,17 @@
 This repository is the mobile application of Galoy. It is build with [React Native](https://reactnative.dev/), and compatible on iOS and Android.
 
 ## Start
+Prerequisite -- Follow the instruction [getting started](https://reactnative.dev/docs/getting-started) to have the correct setup.
 
-type `react-native run-ios` or `react-native run-android` to run locally. Follow the instruction [getting started](https://reactnative.dev/docs/getting-started) to have the correct setup.
+cd into project folder
+
+type `npm install`
```
jrbuhl93, line 12: We use yarn internally for the project, so I think we should recommend `yarn install` instead. Or have both: `yarn install` or `npm install` .

README.md
```diff
@@ -5,8 +5,17 @@
...
+
+`npx react-native start` to start metro.
```
jrbuhl93, line 14: `react-native start` or `npx react-native start` to be consistent with other lines.

jrbuhl93 (commented):
The pull request makes 27,404 changes to the `package-lock.json` file because you used `npm install` instead of `yarn install`. We'll want those changes removed before merging.

app/screens/map-screen/map-screen.tsx
```diff
@@ -131,7 +131,7 @@ export const MapScreen: ScreenType = ({ navigation }: Props) => {
         ? navigation.navigate("sendBitcoin", { username: item.username })
         : navigation.navigate("phoneValidation")
     markers.push(
-      <Marker coordinate={item.coordinate} key={item.title}>
+      <Marker coordinate={item.coordinate} key={item.title} pinColor={'#FF8000'}>
```
jrbuhl93, line 134: We may also want to extract this into https://github.com/GaloyMoney/galoy-mobile/blob/main/app/theme/palette.ts . Or use the orange from that file.

PR #174, opened 2021-08-29
V2 schema for phone auth

jrbuhl93:
Added a split to ApolloClient links so that we use `userRequestAuthCode` and `userLogin` from the new graphql api.

### DIFF
--- a/app/screens/phone-auth-screen/phone-auth.tsx
+++ b/app/screens/phone-auth-screen/phone-auth.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react-native/no-inline-styles */
 import * as React from "react"
-import { useEffect, useRef, useState } from "react"
+import { useCallback, useEffect, useRef, useState } from "react"
 import {
   ActivityIndicator,
   Alert,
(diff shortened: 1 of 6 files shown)


app/screens/phone-auth-screen/phone-auth.tsx
```diff
@@ -93,15 +98,14 @@ type WelcomePhoneInputScreenProps = {
...
   navigation,
 }: WelcomePhoneInputScreenProps) => {
-  const [requestPhoneCode, { loading }] = useMutation(REQUEST_PHONE_CODE, {
+  const [requestPhoneCode, { loading }] = useMutation(REQUEST_AUTH_CODE, {
     fetchPolicy: "no-cache",
   })
 
   const inputRef = useRef<PhoneInput | null>()
 
   const send = async () => {
```
mfferreira, line 107: What do you think about including `useCallback` hook here? Can be part of a separate code cleanup PR.

jrbuhl93: Is it not already? https://github.com/apollographql/react-apollo/issues/3260

mfferreira: I'm referring to the `send` function.

```suggestion
  const send = useCallback(async () => {
```

jrbuhl93: Oh, then yes. In general we aren't using enough useCallbacks throughout the app. I can add it now.

mfferreira (approved):

samerbuna (approved):

Issue #94, opened 2021-07-16
Fix Lint/TS issues on Mobile


samerbuna:

jrbuhl93:
#102 fixed the lint warnings. But typescript still has errors that need to be fixed in another pr.

mfferreira:
#177 will clean most or all of the remaining issues.

PR #177, opened 2021-08-31
#94: Fix a bunch of TS errors

mfferreira:
When `yarn compile`:
- Before: 80 errors
- Now: 0 errors


app/screens/map-screen/map-screen.tsx
```diff
@@ -91,8 +91,8 @@ export const MapScreen: ScreenType = ({ navigation }: Props) => {
     useCallback(() => {
       requestLocationPermission()
 
-      if (!grantedPermission) {
-        return
+      if (grantedPermission) {
```
jrbuhl93, line 94: I think `!grantedPermission` is correct.

mfferreira: Not sure how that happened 🤔  Thanks for spotting!

jrbuhl93:
Could you also add `yarn compile` to `check-code` in the Makefile

app/screens/earns-map-screen/earns-map-screen.story.tsx
```diff
@@ -1,3 +1,4 @@
+import { action } from "@storybook/addon-actions"
```
jrbuhl93, line 1: Can be removed

mfferreira:
> Could you also add `yarn compile` to `check-code` in the Makefile. Maybe rename `yarn compile` to `yarn build` for back end [consistency](https://github.com/GaloyMoney/galoy/blob/26d9e44c25f1d5f4629dca71c4439f5738f9dbdf/Makefile#L55).

Updated 👌

PR #187, opened 2021-09-07
Confirmation screen

jrbuhl93:
- Added a confirm payment screen
- Refactored some of the functions for the send screens to use callbacks and memos
- Refactored how amounts are handled into their own hook
- Updated tests to match the new UX

There are a lot of edge cases and errors that need to be handled, so if reviewers could manually test the app I would appreciate it.

Screenshot_20210907-143955
Screenshot_20210907-144340


app/components/input-payment/input-payment.tsx
```diff
@@ -64,131 +52,58 @@ const styles = EStyleSheet.create({
...
-}
-
 type InputPaymentProps = {
   editable: boolean
   forceKeyboard: boolean
-  nextPrefCurrency: () => void
+  toggleCurrency: () => void
   onBlur?: () => void
-  onUpdateAmount: (amount: number) => void
+  onUpdateAmount: ({ moneyAmount: MoneyAmount }) => void
```
samerbuna, line 60: onUpdateAmount in this component takes a single number argument

app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
```diff
@@ -31,251 +21,63 @@ import { textCurrencyFormatting } from "../../utils/currencyConversion"
...
 
-  const satMoneyAmount = React.useCallback((): MoneyAmount => {
-    return {
-      value: satAmount,
-      currency: "BTC",
-    }
-  }, [satAmount])
+  const satAmount = useMemo(() => {
+    return primaryCurrency === "BTC" ? primaryAmount.value : secondaryAmount.value
+  }, [primaryAmount.value, primaryCurrency, secondaryAmount.value])
```
samerbuna, line 55: No need for useMemo here

PR #188, opened 2021-09-09
Fix too many auth code requests error

### LINKED ISSUE #186: Login screen is not showing errors
**Describe the bug**
If we return an error from backend the user is not receiving feedback

**To Reproduce**
1. Go to login screen
2. Request phone code (do a lot of requests to hit `TOO_MANY_REQUEST` error)

**Expected behavior**
show an error message.  (something similar to "connection issue. Verify your internet connection")

**Additional context**
Backend PR [555](https://github.com/GaloyMoney/galoy/pull/555)

daviroo:
Fixes #186 

**Changes**
- Now shows toast error message to user when too many requests error is received for `userRequestAuthCode` call is made

image

### DIFF
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -665,7 +665,8 @@
   "errors": {
     "generic": "There was an error. Please try again later",
     "invalidEmail": "Invalid email",
-    "invalidPhoneNumber": "is not a valid phone number"
+    "invalidPhoneNumber": "is not a valid phone number",
+    "tooManyRequestsPhoneCode": "Please wait 60 minutes before requesting another text message"
   },
   "notifications": {
     "payment": {
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -666,7 +666,7 @@
     "generic": "There was an error. Please try again later",
     "invalidEmail": "Invalid email",
     "invalidPhoneNumber": "is not a valid phone number",
-    "tooManyRequestsPhoneCode": "Please wait 60 minutes before requesting another text message"
+    "tooManyRequestsPhoneCode": "Too many requests. Please wait before requesting another text message."
   },
   "notifications": {
     "payment": {
--- a/app/i18n/es.json
+++ b/app/i18n/es.json
@@ -701,7 +701,8 @@
   "errors": {
     "generic": "Hubo un error inesperado. Por favor intente más tarde\n",
     "invalidEmail": "Correo inválido",
-    "invalidPhoneNumber": "Esto no es un número de teléfono válido"
+    "invalidPhoneNumber": "Esto no es un número de teléfono válido",
+    "tooManyRequestsPhoneCode": "Espere 60 minutos antes de solicitar otro mensaje de texto"
   },
   "notifications": {
     "payment": {
--- a/app/i18n/es.json
+++ b/app/i18n/es.json
@@ -702,7 +702,7 @@
     "generic": "Hubo un error inesperado. Por favor intente más tarde\n",
     "invalidEmail": "Correo inválido",
     "invalidPhoneNumber": "Esto no es un número de teléfono válido",
-    "tooManyRequestsPhoneCode": "Espere 60 minutos antes de solicitar otro mensaje de texto"
+    "tooManyRequestsPhoneCode": "Demasiadas peticiones. Por favor, espere antes de solicitar otro mensaje de texto."
   },
   "notifications": {
     "payment": {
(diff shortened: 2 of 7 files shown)


daviroo:
The block duration is currently different depending on whether the user has hit the IP block or the phone number block.  You can see the current durations [here](https://github.com/GaloyMoney/galoy/blob/3bec75f77dc02ba4170afc3dbab1322f92bedf2d/default.yaml#L78).  They both seem pretty extreme to me - especially the IP based one as I've seen this trigger before on mobile networks where users sometimes share IP addresses.

app/i18n/es.json
```diff
@@ -682,7 +682,8 @@
   "errors": {
     "generic": "Hubo un error inesperado. Por favor intente más tarde\n",
     "invalidEmail": "Correo inválido",
-    "invalidPhoneNumber": "Esto no es un número de teléfono válido"
+    "invalidPhoneNumber": "Esto no es un número de teléfono válido",
+    "tooManyRequestsPhoneCode": "Espere 60 minutos antes de solicitar otro mensaje de texto"
```
daviroo: Would be good to get a native Spanish speaker to review this before merge.

daviroo:
Also I've yet to test this change on android because I couldn't get the emulator running on my local computer (complaining about missing `adb`).  Still working on it.

app/i18n/es.json
jrbuhl93: @dolcalmi

app/i18n/en.json
```diff
@@ -651,7 +651,8 @@
   "errors": {
     "generic": "There was an error. Please try again later",
     "invalidEmail": "Invalid email",
-    "invalidPhoneNumber": "is not a valid phone number"
+    "invalidPhoneNumber": "is not a valid phone number",
+    "tooManyRequestsPhoneCode": "Please wait 3 hours before requesting another text message"
```
jrbuhl93, line 655: I would add "Too many requests. Please wait...", so that they know why they have to wait.

jrbuhl93: This will also be different depending on whether the block is for an ip (1 hour) or phone number (3 hours). We should either just do "Pleas wait before..." and not include the number or pass a different number to the message depending on which error code was hit.

daviroo: I think a user will most likely be using the same phone number and same IP address with each request, in which case both will trigger.  I'm assuming the rate limiting defaults to the longest block duration if 2 are triggered (maybe I'm wrong?), in which case it will **almost** always be 3 hours.  As I said in my comment on the PR itself I think we should change these values anyway (the block durations) as they seem a bit extreme.

Also regarding the error code I couldn't find the error code on the object which we get in the client code.  The object we are dealing with only seems to have a `name` and `message`.  I think it's one of [these](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error).

jrbuhl93: Ah yes, we don't have codes for these mutations. I mean we may need different messages we can check against. You're right that it will probably default to three hours for now unless the user switches ip or we change the values on the backend.

daviroo:
I finally got the Android emulator running locally.  I had to update the `compileSdkVersion` to 31 to get it to work. 

image

app/i18n/en.json
daviroo: I've updated the error message with the "Please wait before..." party you suggested.  I have changed it to say 3 hours instead of 60 minutes as I think it will almost always hit the 3 hour limit.

nicolasburtey: the time mention here (1h or 3h) is a default settings. it can be overriden with a yaml file so we should not have any statics time set in the translation file (we may already have different value for the Bitcoin Beach wallet)

nicolasburtey:
> I finally got the Android emulator running locally. I had to update the compileSdkVersion to 31 to get it to work.

does it have any drawback? like no longer being compatible with old android version?

jrbuhl93:
> does it have any drawback? like no longer being compatible with old android version?

It should be okay since the minSDKVersion is still the same (21). I made the same change in my [pr](https://github.com/GaloyMoney/galoy-mobile/pull/187/commits/7d205267526f7d3166d237e25faf227baffe74ce).

app/i18n/en.json
jrbuhl93: I agree with Nicolas. Until we have a standard way of passing back how long the the timeout is in the error, we should just leave the message as "Too many requests. Please wait before requesting another text message."

jrbuhl93 (approved):

jrbuhl93 (approved):

PR #191, opened 2021-09-13
Don't lowercase addresses that might be p2pkh or p2sh

jrbuhl93:
p2pkh and p2sh are case sensitive, so we should only lowercase lightning invoices and bech 32 paths.


dolcalmi (approved):
I have not tested in phone, before release please test QRs with `lightning:` and some p2pkh and p2sh addresses

Issue #182, opened 2021-09-04
We should advertise the tipping/link to the user for lnpage


nicolasburtey:
https://twitter.com/theguyswann/status/1433934882221273091?s=21

maybe have a link for it in the settings page so people are aware of this option

danwebb1:
I have a working prototype, so to speak, on this [branch] (https://github.com/danwebb1/galoy-mobile/tree/feature_advertise_copy_tip_link). I am having environment issues with Android locally so I have to deal with that when I get free time. Needs to be tested on Android and polished but its close. If anyone else can finish this off quickly, feel free to use or lose what I started!

PR #193, opened 2021-09-15
Feature advertise copy tip link

danwebb1:
Added the settings menu item to advertise sharing of user's tip links. Copies the tip link to their clipboard and uses Toast to display that they copied their link so they can share. This PR is an attempt to satisfy [Issue #182](https://github.com/GaloyMoney/galoy-mobile/issues/182)

### DIFF
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -664,6 +664,10 @@
     "other": "{{formatted_number}} sats",
     "zero": "0 sats"
   },
+  "tippingLink": {
+    "title": "Want to receive tips? Share your tipping link!",
+    "copied": "{{data}} saved to clipboard"
+  },
   "whatsapp": {
     "contactUs": "Need help? Talk to us on WhatsApp!",
     "defaultSupportMessage": "Hey there! I need some help with Bitcoin Beach Wallet"
--- a/app/screens/settings-screen/settings-screen.tsx
+++ b/app/screens/settings-screen/settings-screen.tsx
@@ -222,7 +222,7 @@ export const SettingsScreenJSX: ScreenType = (params: SettingsScreenProps) => {
       id: "tippingLink",
       action: () => copyToClipBoard(username),
       enabled: walletIsActive && username !== null,
-      greyed: !walletIsActive || !username,
+      greyed: !walletIsActive || username === null,
     },
     {
       category: translate("whatsapp.contactUs"),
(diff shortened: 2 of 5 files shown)


app/screens/settings-screen/settings-screen.tsx
```diff
@@ -208,6 +217,14 @@ export const SettingsScreenJSX: ScreenType = (params: SettingsScreenProps) => {
       enabled: walletIsActive,
       greyed: !walletIsActive,
     },
+    {
+      category: translate("tippingLink.title"),
+      icon: "cash-outline",
+      id: "tippingLink",
+      action: () => copyToClipBoard(username),
+      enabled: walletIsActive,
+      greyed: !walletIsActive,
```
dolcalmi, line 226: this should also validate if the username has been set

Screenshot_1631682778

app/i18n/en.json
```diff
@@ -664,6 +664,10 @@
     "other": "{{formatted_number}} sats",
     "zero": "0 sats"
   },
+  "tippingLink": {
+    "title": "Want to receive tips? Share your tipping link!",
+    "copied": "{{data}} saved to clipboard"
+  },
```
dolcalmi, line 670: please add the next to `app/i18n/es.json`

```json
"tippingLink": {
  "title": "¿Quiere recibir propinas? ¡Comparta su enlace!",
  "copied": "{{data}} se ha guardado al portapapeles"
},
```

dolcalmi (changes requested):

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -156,6 +158,12 @@ export const SettingsScreenJSX: ScreenType = (params: SettingsScreenProps) => {
     securityAction,
     resetDataStore,
   } = params
+  const copyToClipBoard = (username) => {
+    Clipboard.setString(LN_PAGE_DOMAIN + username)
+    Clipboard.getString().then((data) =>
+      toastShow(translate("tippingLink.copied", { data })),
```
jrbuhl93, line 164: Could you add an optional parameter to [toastShow ](https://github.com/GaloyMoney/galoy-mobile/blob/5d0ac53558c3655fddfea9642e17fd9c6d813f23/app/utils/toast.ts#L4) for the backgroundColor, so that the toast can be a more friendly color than red. Maybe green?

danwebb1: I changed it to the lightBlue options because I personally liked it better/matches the app's branding a bit better but happy to make it green if preferable.

dolcalmi (approved):

jrbuhl93 (approved):

Issue #122, opened 2021-07-30
Improve pin for current location


nicolasburtey:
Currently the current location is shown as a pin that is the same for the merchant, just with a different color (blue instead or red). we want to use something similar to Google Maps and use the compass.

MikeMwambia-TrojanSystem:
More details please

jrbuhl93:
I think this is referring to the icon that shows your current location on the map screen. Instead of a blue colored pin, it should be what is used in google and apple maps.

Our app:

![Blue pin from iOS](https://user-images.githubusercontent.com/35902932/129904599-58d9831a-f88b-4ba3-a7a9-5848e204dcdf.png)

Apple maps:

![Apple map from iOS](https://user-images.githubusercontent.com/35902932/129904632-02de5b6e-9693-471e-869a-0978f77c018c.png)

MikeMwambia-TrojanSystem:
Got it

MikeMwambia-TrojanSystem:
![PinLocation](https://user-images.githubusercontent.com/5917824/130427651-7ffc1c1a-ddfd-4931-9f16-2606239f5024.png)

This is the best image I can get for the icon better suggestions are welcome..

MikeMwambia-TrojanSystem:
@jrbuhl93 according to [this](https://github.com/react-native-maps/react-native-maps/blob/master/docs/marker.md) , I think there are 2 approaches to use `image` or `icon` , what do you think ?

jrbuhl93:
I think we should use `showsUserLocation` in [MapView](https://github.com/react-native-maps/react-native-maps/blob/master/docs/mapview.md).

MikeMwambia-TrojanSystem:
Did I miss something because to my understanding [on this code](https://github.com/GaloyMoney/galoy-mobile/blob/73b96b738855fceccacb3580923ad10b9f42c63c/app/screens/map-screen/map-screen.tsx#L108) it gives the current location and to have a custom pin we need to reference an image , is that so ?

jrbuhl93:
You are correct. But I think Nicolas may have done this because he didn't see the `showsUserLocation` in`MapView`. This makes all the code to create the blue pin unnecessary (assuming `showsUserLocation` works well).

MikeMwambia-TrojanSystem:
Okay got it.
Thank you.

PR #195, opened 2021-09-16
Improve pin for current location

krsjoseph:
- Removed old pin for user location
- Use showsUserLocation within MapView comp

Fixes https://github.com/GaloyMoney/galoy-mobile/issues/122

Simulator Screen Shot - iPhone 12 - 2021-09-16 at 21 27 22

### DIFF
--- a/app/screens/map-screen/map-screen.tsx
+++ b/app/screens/map-screen/map-screen.tsx
@@ -87,35 +87,6 @@ export const MapScreen: ScreenType = ({ navigation }: Props) => {
     }
   }
 
-  useFocusEffect(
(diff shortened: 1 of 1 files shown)


krsjoseph:
Hey folks not sure if this is the right way to do a PR, if there's a template let me know. This was a small fix I wanted to get in to test my setup was working.

jrbuhl93:
> Hey folks not sure if this is the right way to do a PR, if there's a template let me know. This was a small fix I wanted to get in to test my setup was working.

It looks good to me. I'll test it soon.

PR #199, opened 2021-09-19
Add 2FA to payments above the sat threshold (300,000 sats)

jrbuhl93:
- [X] Add turning on and off 2FA to settings screen
- [] Add 2FA to confirmation screen if twoFAEnabled
Image from iOS
Screenshot 2021-09-19 at 11 45 35 AM
Screenshot 2021-09-19 at 11 45 07 AM


app/screens/security-screen/two-factor-authentication-selection-screen.tsx
```diff
@@ -0,0 +1,171 @@
...
+//       errors {
+//         message
+//       }
+//       twoFASecret {
+//         secret
+//         uri
+//       }
+//     }
+//   }
+// `
```
daviroo, line 37: Do we need this?

app/screens/security-screen/two-factor-authentication-verification-screen.tsx
```diff
@@ -0,0 +1,221 @@
...
+        if (success) {
+          fetchTwoFAEnabled()
+          navigation.navigate("security")
+        } else {
+          toastShow(translate("TwoFAVerificationScreen.errorVerifyingCode"))
+        }
+      }
+    } catch (err) {
+      console.warn({ err })
+      toastShow(`${err}`)
```
daviroo, line 104: Can we be sure that the error surfaced to the user here will be friendly?

app/screens/security-screen/two-factor-authentication-verification-screen.tsx
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  // useEffect(() => {
+  //   ;(async () => {
+  //     if (setupMethod === "Linking") {
+  //       Linking.openURL(twoFAUri)
+  //     }
+  //   })()
+  // }, [setupMethod, twoFAUri])
```
daviroo, line 114: Is this required?

daviroo, line 114: Is this required?

daviroo (commented):
I'm going to check this code out and give it a run locally.  I only had a few small comments/questions regarding commented out code and a few other things.

app/screens/security-screen/two-factor-authentication-selection-screen.tsx
jrbuhl93: This is for switching things over to api v2. I was having some trouble with some of the mutations not being ready when I was working on it.

app/screens/security-screen/two-factor-authentication-verification-screen.tsx
jrbuhl93: No, I'll remove it in the next commit.

jrbuhl93: No, I'll remove it in the next commit.

app/screens/security-screen/two-factor-authentication-verification-screen.tsx
jrbuhl93: The problem is we might need the more descriptive error to debug issues for the app. I think for common errors we except we should have checks and translations and then this as a fallback. Perhaps rap it in a translation that says "There was an unknown error: ".

Issue #196, opened 2021-09-17
Username Validation Extension


jrbuhl93:
We avoid a username starting with 1, 3, or bc1 because these could be bitcoin addresses. This should be added to client side username validation and a custom error message that says username can't start with 1, 3, or bc1

PR #205, opened 2021-09-23
Add restriction on username start characters

RonanJackson:
Fixes https://github.com/GaloyMoney/galoy-mobile/issues/196

Worth noting someone might need a proper Spanish translation as this was just a quick Google Translate, thanks!

**Changes**
Now shows an error message when trying to create a username beginning with 'bc1', '1', or '3'.
Screenshot 2021-09-23 at 4 31 57 pm

### DIFF
--- a/app/i18n/es.json
+++ b/app/i18n/es.json
@@ -626,6 +626,7 @@
     "confirmSubtext": "Este nombre de usuario es permanente y no se podrá cambiar más adelante.",
     "confirmTitle": "¿Desea establecer %{input} como su nombre de usuario?",
     "letterAndNumber": "Solo se aceptan letras en minúscula, números y guiones bajos (_)",
+    "forbiddenStart": "No se puede empezar con bc1, 1, o 3 ",
     "notAvailable": "❌  %{input} no está disponible",
     "success": "¡%{input} es ahora su nombre de usuario!",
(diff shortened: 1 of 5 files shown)


app/i18n/es.json
```diff
@@ -626,6 +626,7 @@
     "confirmSubtext": "Este nombre de usuario es permanente y no se podrá cambiar más adelante.",
     "confirmTitle": "¿Desea establecer %{input} como su nombre de usuario?",
     "letterAndNumber": "Solo se aceptan letras en minúscula, números y guiones bajos (_)",
+    "forbiddenStart": "No se puede empezar con bc1, 1, o 3 ",
```
dolcalmi, line 629: maybe is better to use "No debe empezar con bc1, 1 o 3 "

dolcalmi (changes requested):

app/i18n/es.json
jrbuhl93: Also list "forbiddenStart" before "letterAndNumber" to maintain alphabetical order.

jrbuhl93:
Thanks for the pr!

app/i18n/es.json
RonanJackson: Done!

PR #208, opened 2021-09-26
Fix storybook

daviroo:
I've fixed storybook so it now runs for me on device.  The PR contains the following changes:
- Added `mock-apollo-client` lib so we can mock the graphql responses as shown [here](https://www.npmjs.com/package/mock-apollo-client)
- Updated the start command for storybook
- Fixed up a few of the stories so they actually render

If there are any visual issues in any of the stories then I suggest we fix those in a separate PR.

image


jrbuhl93 (approved):

jrbuhl93:
Looks good

PR #209, opened 2021-09-27
Add stop and startPolling when app comes to foreground

jrbuhl93:
I believe apollo client stops polling in the background: https://github.com/apollographql/apollo-client/issues/5666. This causes users who make payments after bringing the app to the foreground from the background rather than terminated to not receive fresh btc prices. By starting polling (and stopping for safety) when the app comes to the foreground, we might fix this issue.

### DIFF
--- a/app/navigation/root-navigator.tsx
+++ b/app/navigation/root-navigator.tsx
@@ -62,6 +62,8 @@ import type { NavigatorType } from "../types/jsx"
 
 import PushNotification from "react-native-push-notification"
 
+const PRICE_POLLING_INTERVAL = 30000
+
 // Must be outside of any component LifeCycle (such as `componentDidMount`).
 PushNotification.configure({
   // (optional) Called when Token is generated (iOS and Android)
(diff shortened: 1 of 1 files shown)


nicolasburtey:
The questions comes back to: is the polling start again automatically when the app is going from background to foreground? 
the link article doesn't mention anything about it

jrbuhl93:
> The questions comes back to: is the polling start again automatically when the app is going from background to foreground?
> the link article doesn't mention anything about it

Based on my testing, no.

jrbuhl93:
I just tested again without this update and it was updating after closing and reopening the app. I'll test more.

jrbuhl93:
Polling updates starts up again automatically, so this must not be the problem.

jrbuhl93:
Usually the price continues to update fine after entering the background and then coming in. But I have noticed sometimes price updates stop working (it may be unrelated to going into the background). When this happens, going from background to foreground doesn't restart the polling. I think this pr might be a good way to mitigate this problem even though I am not sure exactly what triggers the polling to stop.

daviroo:
> Usually the price continues to update after entering the background and then entering foreground. But I have noticed sometimes price polling stops working (it may be unrelated to going into the background). When this happens, going from background to foreground doesn't restart the polling. I think this pr might be a good way to mitigate this problem by regularly manually restarting polling even though I don't know what is triggering the polling to stop.

Do we see the same behaviour on both ios and android?  I know ios is quite aggressive in stopping background activities.

jrbuhl93:
> Do we see the same behaviour on both ios and android? I know ios is quite aggressive in stopping background activities.

I mostly have tested on Android, so I know the polling stops on android sometimes. But I believe Mike from Bitcoin Beach has an iphone, so it probably happens on iOS too. At this point, I am not sure what is causing the polling to stop and not restart. It could be unrelated to being in the background.

nicolasburtey:
yeah there are 2 states of "backgroundness" I think. one if the app is in the background but still running, and the other is the app is paused and not running any more. both can go back to a running state. 

what we want to check is whether the "paused" state to "running" state still run the query. I would guess it would but we should test.

having visibility on the backend with a local running instance could be an easy way to test this. wait until the query stops and then switch back to the app in the fontend

nicolasburtey:
jorge has this issue also, and AFAIK he's on android. so seems to affect both platform.

Issue #210, opened 2021-09-27
automatically suggest retrying send text after 60 sec


nicolasburtey:
Sometimes, text doesn't arrive to the mobile phone. One way to improve this is to suggest trying to send the text another time after a delay. 

If the phone was not accessible by the carrier initially, 1 min after this could work better.

We could suggest sending back the text after 1 minutes, by having a timer on the text confirmation screen.

maxghafori:
[PR 212](https://github.com/GaloyMoney/galoy-mobile/pull/212)

PR #212, opened 2021-09-27
Add timer and retry button in phone verification screen

maxghafori:
[issue 210](https://github.com/GaloyMoney/galoy-mobile/issues/210)

Suggest retrying send text after 60 sec in phone verification screen.

### DIFF
--- a/app/components/timer/utils.ts
+++ b/app/components/timer/utils.ts
@@ -1,0 +1,25 @@
+export function parseTimer(seconds: number): string {
+  if (!seconds) {
+    return "00:00"
+  }
+  const min = parseInt(String(seconds / 60)).toString()
+  const sec = parseInt(String(seconds % 60)).toString()
+  const minute = parseInt(min) < 10 ? `0${min}` : min
+  const second = parseInt(sec) < 10 ? `0${sec}` : sec
+  return `${minute}:${second}`
+}
+
+// this function is used to compute correct time spent
+export function computeCount(startTime: Date): number {
+  const oldMin = startTime.getMinutes()
+  const oldSec = startTime.getSeconds()
+  const newMin = new Date().getMinutes()
+  const newSec = new Date().getSeconds()
+  if (oldMin === newMin) {
+    return 60 - (newSec - oldSec)
+  } else if (newMin - oldMin === 1) {
+    return 61 - (60 - oldSec + newSec)
+  } else {
+    return 0
+  }
+}
--- a/app/components/timer/utils.ts
+++ b/app/components/timer/utils.ts
@@ -2,11 +2,9 @@ export function parseTimer(seconds: number): string {
   if (!seconds) {
     return "00:00"
   }
-  const min = parseInt(String(seconds / 60)).toString()
-  const sec = parseInt(String(seconds % 60)).toString()
-  const minute = parseInt(min) < 10 ? `0${min}` : min
-  const second = parseInt(sec) < 10 ? `0${sec}` : sec
-  return `${minute}:${second}`
+  const minute = parseInt(String(seconds / 60)).toString()
+  const second = parseInt(String(seconds % 60)).toString()
+  return `${minute.padStart(2, '0')}:${second.padStart(2, '0')}`
 }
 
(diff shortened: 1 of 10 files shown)


app/components/timer/utils.ts
```diff
@@ -0,0 +1,25 @@
+export function parseTimer(seconds: number): string {
+  if (!seconds) {
+    return "00:00"
+  }
+  const min = parseInt(String(seconds / 60)).toString()
+  const sec = parseInt(String(seconds % 60)).toString()
+  const minute = parseInt(min) < 10 ? `0${min}` : min
```
jrbuhl93, line 7: Cleaner to use [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) here.

app/screens/phone-auth-screen/phone-auth.tsx
```diff
@@ -317,6 +347,24 @@ export const WelcomePhoneValidationScreen: ScreenType = ({
...
+              >
+                <Text
+                  style={{
+                    color: timer === 0 ? color.palette.black : color.palette.midGrey,
+                  }}
+                >
+                  {translate("WelcomePhoneValidationScreen.sendAgain")}
+                </Text>
+              </Pressable>
+              <Timer onTimerFinish={() => setTimer(0)} count={timer} />
```
jrbuhl93, line 366: I think hiding the timer might look good once it reaches 0.

jrbuhl93 (commented):
Let's also remove the PhoneBadger component on line 325. There isn't enough space on most screen size for it, and the image will cause the keyboard to cover up the timer so users won't know it is there.

This is a great pr! Thank you.

app/screens/phone-auth-screen/phone-auth.tsx
maxghafori: I changed it to show timer when it's more than 0 and also show `Send Again` button when timer is 0.

maxghafori:
Thank you for your comments , I fixed them.

jrbuhl93:
> Thank you for your comments , I fixed them.

Looks like you might have forgotten to add the changes to your new commit.

maxghafori:
Sorry for that, now it should be ok .

maxghafori:
I couldn't use alias path `@app` for import component because i got error.

If you agree, I will add this plugin to babel config 
`plugins: [['module-resolver',{ alias: {'@app': './app'}}]]`

and also I should add this dev dependency
`yarn add --dev babel-plugin-module-resolver`

jrbuhl93:
> I couldn't use alias path `@app` for import component because i got error.
> 
> If you agree, I will add this plugin to babel config `plugins: [['module-resolver',{ alias: {'@app': './app'}}]]`
> 
> and also I should add this dev dependency `yarn add --dev babel-plugin-module-resolver`

You are right that for now we aren't set up to resolve @app (except in the tests folder). Hold off on adding the plugin for now. Samer is going to look into alternatives.

app/components/timer/utils.ts
```diff
@@ -0,0 +1,23 @@
+export function parseTimer(seconds: number): string {
```
jrbuhl93, line 1: Could you add some unit tests for parseTimer?

app/components/timer/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (!seconds) {
+    return "00:00"
+  }
+  const minute = parseInt(String(seconds / 60)).toString()
+  const second = parseInt(String(seconds % 60)).toString()
+  return `${minute.padStart(2, '0')}:${second.padStart(2, '0')}`
+}
+
+// this function is used to compute correct time spent
+export function computeCount(startTime: Date): number {
```
jrbuhl93, line 11: And unit tests for computeCount

jrbuhl93 (commented):
Thanks for making the change to hide the timer at 0. But I think we should keep the disabled "send again" while the timer is active, so that users know the purpose of the timer and aren't confused.

maxghafori:
Now the Send Again button is shown when disabled.

PR #223, opened 2021-09-30
Query price in send screens if stale during render

jrbuhl93:
I'd like input before merging this. Is this a good approach to ensure price is up to date? 

One possibility for improvement: is there a better way to call the check for a stale price rather than using an interval? One alternative is only when the screen loads.


package.json
```diff
@@ -13,7 +13,7 @@
     "prettier:fix": "prettier '**/*.{js,ts,tsx,json}' --write",
     "check-code": "yarn tsc:check && yarn eslint:check && yarn prettier:check",
     "codegen": "apollo client:codegen --endpoint http://127.0.0.1:4000/graphql --target typescript --globalTypesFile=app/__generated__/globalTypes.ts -c ./apollo.client.js --watch",
-    "test": "LOGLEVEL=warn jest --runInBand",
+    "test": "LOGLEVEL=warn jest --runInBand --forceExit",
```
dolcalmi, line 16: did you check with --detectOpenHandles before use force exit? is better detect and fix the open handles

samerbuna (approved):

PR #225, opened 2021-10-01
Alert user not to leave while payment is processing

jrbuhl93:
When a user makes a payment we don't want them to leave the screen while the mutation hasn't concluded because they may try to make the payment again and end up sending two payments.


daviroo:
Closing as PR is stale.

Issue #171, opened 2021-08-24
Remove lodash


jrbuhl93:
To minimize external imports we should use javascript instead of lodash where possible.

danwebb1:
Would it be appropriate to have a subdirectory under utils for functions written to replace lodash and then just work through the project as able?

jrbuhl93:
In this commit, not yet merged, Samer replaced importing all of lodash with just importing the functions we needed. What would be preferred is to use native javascript functions instead of lodash. If needed we can make our own functions under utils, but I think a lot of them may be doable in javascript.

PR #236, opened 2021-10-07
Remove some lodash use

RonanJackson:
Addresses some of: https://github.com/GaloyMoney/galoy-mobile/issues/171

I have removed some of the `lodash` usage around the project. The remaining packages seem non trivial to change, bar `sumby`, but that is a little involved too. These are just the basic straight swaps into vanilla JS.


app/screens/move-money-screen/move-money-screen.tsx
```diff
@@ -204,10 +203,9 @@ export const MoveMoneyScreenDataInjected: ScreenType = ({
     }
   }
 
-  const lastTransactions = find(data?.wallet, { id: "BTC" })?.transactions?.slice(
-    undefined,
-    3,
-  )
+  const lastTransactions = data?.wallet
```
jrbuhl93, line 206: A `?` after wallet is needed or else the app crashes on this screen when logged out.

daviroo:
This is an old PR with no recent activity so I'm going to close it.

PR #260, opened 2021-11-04
v2 fixes

samerbuna:
- Fix quiz screen to render wihtout a session
- Fix the auth errors on login


daviroo (approved):

daviroo:
Me and @samerbuna paired on this yesterday.  There's better ways of doing it we think but we want to solve the problem for now as we will probably be doing larger refactoring at some point in the future.

PR #269, opened 2021-11-04
Install react native in README.md

trigger67:
New link + info about the tab to follow.
Because using "Expo CLI Quickstart" will lead to many issues.


daviroo (approved):
Thanks 👍

Issue #190, opened 2021-09-10
Logout when token is invalid


dolcalmi:
**Describe the bug**
Support has received multiple cases with the recent problems in backend related to invalid token (in context), if the user can't request its data then the balance and transactions appears empty(zero) so we need to logout the user if we have this problem 

**To Reproduce**
1.  Using local environment, log in with a test user
2. After being logged in reset the local test enviroment `TEST="01|02" make reset-integration` (this will reset user ids and current token will be invalid)
3. Refresh balance/transactions

**Expected behavior**
Logout if invalid token

daviroo:
Fixed in #281

PR #281, opened 2021-11-06
Fix to Logout when token is invalid

rimbener:
Fixes Issue [#190](https://github.com/GaloyMoney/galoy-mobile/issues/190)


samerbuna:
This is great @rimbener. Thanks for working on it!

Let's move the `removeToken` logic into `resetDataStore`, I think the line `authTokenVar(null)` is what's needed as the storage key is removed in `resetDataStore` already.

Maybe also have the logout method part of useToken as well? We'll probably eventually rename that hook to something more accurate, but it is the hook for the user's authentication status.

app/hooks/use-logout.ts
```diff
@@ -0,0 +1,17 @@
+import { useApolloClient } from "@apollo/client"
+import { resetDataStore } from "../utils/logout"
+import useToken from "../utils/use-token"
+
+const useLogout = () => {
+  const client = useApolloClient()
+  const { removeToken } = useToken()
+
+  // TODO a login function could be created in this hook
+  const logout = () => resetDataStore({ client, removeToken })
```
daviroo, line 10: We could bring in the navigator using the [useNavigation hook](https://reactnavigation.org/docs/5.x/use-navigation) and always navigate the user to the Home Screen on logout.

app/components/global-error/global-error.tsx
```diff
@@ -27,8 +32,20 @@ export const GlobalErrorToast: ComponentType = () => {
   }
 
   if (networkError.statusCode >= 400 && networkError.statusCode < 500) {
-    // TODO translation
-    toastShow("Request issue.\nContact support if the problem persists")
+    const errorCode = (networkError as ServerError).result?.errors?.[0]?.code
+    switch (errorCode) {
+      case NetworkErrorCode.InvalidAuthentication:
+        toastShow(translate("common.logoutOnInvalidToken"), {
```
daviroo, line 38: I think the error message displayed should just say "You have been logged out.".  Saying the token has expired probably means nothing to most users and so best to keep the language as plain as possible.  We already have an error message we use when a user logs themselves out [here](https://github.com/GaloyMoney/galoy-mobile/blob/52efe8baf99d0831dac98cbb3a26d0dd70eb95a5/app/screens/authentication-screen/authentication-screen.tsx#L115).

daviroo (changes requested):
Hi @rimbener.  First of all thanks very much for your contribution.  It's a really good first PR.  I've made a couple of small comments.  We would be happy to expand this to include your suggested TODO's as well if you are interested?

app/hooks/use-logout.ts
rimbener: I tried but there is on problem to do this, because `GlobalError` is not under any navigator so you can't redirect to any screen from GlobalError. 
**BUT** anyways this code in `RootNavigator` is redirecting to getStarted:
```
 <RootNavigator.Navigator
      screenOptions={{ gestureEnabled: false }}
      initialRouteName={token ? "authenticationCheck" : "getStarted"}
    >
```

app/components/global-error/global-error.tsx
rimbener: > Hi @rimbener. First of all thanks very much for your contribution. It's a really good first PR. I've made a couple of small comments. We would be happy to expand this to include your suggested TODO's as well if you are interested?

Sure!, I'll do the suggested changes and also will work on those TODO's.

app/components/global-error/global-error.tsx
```diff
@@ -27,8 +32,20 @@ export const GlobalErrorToast: ComponentType = () => {
...
+      case NetworkErrorCode.InvalidAuthentication:
+        toastShow(translate("common.loggedOut"), {
+          duration: Toast.durations.SHORT,
+          onHidden: () => logout(),
+        })
+        break
+
+      default:
+        // TODO translation
+        toastShow("Request issue.\nContact support if the problem persists")
```
daviroo, line 46: Sorry I know this isn't your code and you have just moved it - but do you mind moving this error into the translation files?

daviroo (changes requested):
One minor suggestion @rimbener on some code you have refactored.

app/components/global-error/global-error.tsx
rimbener: Sure, np, I just didn't know how far I could go with the changes, I just pushed a commit with all the GlobalError messages translated

daviroo (approved):

PR #282, opened 2021-11-07
Implement Set/Unset notifications setting

samerbuna:
This PR depends on https://github.com/GaloyMoney/galoy/pull/724


samerbuna:
For now, we'll not include a Notifications menu until we support notifications in V2.

The notifications row was removed in #296

PR #291, opened 2021-11-09
Upgrade all dependencies

samerbuna:
Including:

- React Navigation (major versions upgrade, some API changes)
- Firebase libraries (major versions upgrade)
- Other packages like graphql, subscription transport, vector icons (major versions upgrade)
- Other minor/patch upgrades
- All pods

Please test on Android


app/app.tsx
```diff
@@ -230,30 +230,33 @@ export const App = (): JSX.Element => {
     return null
   }
 
+  // TODO: Fix this. It does not work currently
```
daviroo, line 233: Could you add more context to the TODO? If I saw this I wouldn't know what isn't working (is it the type?). Or even better link it to an issue in GitHub.

daviroo (changes requested):
Other than the TODO it looks good. Also assuming that deep linking has been tested after the upgrade to the navigation library.

samerbuna:
Tests pass for me locally but fail here 🤔

daviroo:
This was broken due to jest not being able to transform some of our dependency code.  I told it to ignore the `react-navigation` library and now the tests are passing. The reason the tests were failing here but not locally is that the tests here run on the pull request and there were changes from main to be rebased.

daviroo (approved):

PR #295, opened 2021-11-11
Fix duplicate transactions problem

samerbuna:


daviroo (approved):
We paired on this so looks good 👍

PR #313, opened 2021-11-27
v2 fixes

samerbuna:
- Don't display price in debug screen if it's not available
- Make default OS default in the UI when there's no session
- Fix transactions screen not displaying latest updates


samerbuna:
@daviroo merging this but please review. This should fix the transactions screen issue. I don't see a value of caching this list between screen renders.

PR #318, opened 2021-12-01
Fix the transactions data problem

samerbuna:


daviroo (approved):

vindard:
We're also working on a fix for this at https://github.com/GaloyMoney/galoy/pull/771 that may make these changes redundant

PR #326, opened 2021-12-05
Change default currency for transactions

bityogi:


bityogi:
Sorry, wrong repository. This was meant as an internal PR for our branch.

nicolasburtey:
@bityogi we would welcome PR that would make it easy to replace "BitcoinBeach" to string set from an environment variable or configurable yaml/json file. this would limit the risk of having a fork that is diverging from the upstream

PR #329, opened 2021-12-07
Fix Contacts Not Updating

### LINKED ISSUE #327: Contacts not updating
**Describe the bug**
Contacts are not updating in the app after a successful payment has been made using a username

**To Reproduce**

1. Send a payment to someone who isn't already a contact using their username
2. Navigate to contacts screen and notice that their username hasn't been added as a contact

**Expected behavior**
When successful payments are made to a username that username should be added to your list of contacts

adam-stover:
Fixed by refetching contacts on intra-ledger payment mutation. We could add check to see if contact already exists to avoid unnecessary refetch, wasn't sure if that was worth pulling in the contacts query to this component.

Resolves #327

### DIFF
--- a/app/screens/contacts-screen/contacts.tsx
+++ b/app/screens/contacts-screen/contacts.tsx
@@ -99,9 +99,8 @@ export const ContactsScreen: ScreenType = ({ navigation }: Props) => {
   const { hasToken } = useToken()
   const [matchingContacts, setMatchingContacts] = useState([])
   const [searchText, setSearchText] = useState("")
-  const [isRefreshed, setIsRefreshed] = useState(false)
 
-  const { loading, data, error, refetch } = useQuery(
+  const { loading, data, error, startPolling, stopPolling } = useQuery(
     gql`
       query contacts {
         me {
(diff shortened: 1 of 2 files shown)


samerbuna (dismissed):

app/screens/contacts-screen/contacts.tsx
```diff
@@ -113,7 +113,7 @@ export const ContactsScreen: ScreenType = ({ navigation }: Props) => {
         }
       }
     `,
-    { skip: !hasToken },
+    { fetchPolicy: 'cache-and-network', skip: !hasToken },
```
daviroo, line 116: Does this work?  When I tried this policy on another query it didn't seem to change anything.

adam-stover: It's meant to fix the case where the recipient of a payment from a new contact has `contacts` in their cache. When they mount this component, it will use the `contacts` in the cache while refetching, then use the updated `contacts` if they have changed.

So it's similar to `network-only` in what it solves for us here.

app/screens/contacts-screen/contacts.tsx
```diff
@@ -116,12 +115,13 @@ export const ContactsScreen: ScreenType = ({ navigation }: Props) => {
...
 
-  useFocusEffect(() => {
-    if (!isRefreshed) {
-      setIsRefreshed(true)
-      refetch()
-    }
-  })
+  useFocusEffect(
+    useCallback(() => {
+      startPolling(10 * 1000)
```
daviroo, line 120: As we spoke about on the call I'm not convinced on the polling.  We don't do it anywhere else in the app.  Why not just use a subscription if it's important to keep the data up to date while on the screen?

daviroo (changes requested):
As we spoke about on our call we should probably remove the polling.

PR #330, opened 2021-12-07
add masking to transaction items if balances are hidden in user settings

### LINKED ISSUE #319: transaction/amount should be masked to the home page when security setting is on

adam-stover:
User can click on amount to toggle visibility for that specific tx.

image

image

image

Resolves #319


daviroo (dismissed):

daviroo (approved):
Reviewed on a call with @adam-stover

PR #335, opened 2021-12-08
Replace anies with types

adam-stover:
Adds some types to take care of the check code complaints.

### DIFF
--- a/app/navigation/stack-param-lists.ts
+++ b/app/navigation/stack-param-lists.ts
@@ -86,10 +86,11 @@ export type MoveMoneyStackParamList = {
 }
 
 export type PhoneValidationStackParamList = {
+  Primary: undefined
   welcomePhoneInput: undefined
   welcomePhoneValidation: {
     phone: string
-    setPhone: (string) => void
+    setPhone: (str: string) => void
   }
   authentication: {
     screenPurpose: AuthenticationScreenPurpose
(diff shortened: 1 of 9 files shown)


app/navigation/stack-param-lists.ts
```diff
@@ -16,7 +16,7 @@ export type RootStackParamList = {
   authenticationCheck: undefined
   authentication: {
     screenPurpose: AuthenticationScreenPurpose
-    isPinEnabled: boolean
+    isPinEnabled: boolean | undefined
```
nicolasburtey, line 19: doesn't feel right for this to be undefined.

app/navigation/stack-param-lists.ts
```diff
@@ -86,10 +86,11 @@ export type MoveMoneyStackParamList = {
 }
 
 export type PhoneValidationStackParamList = {
+  Primary: undefined
```
nicolasburtey, line 89: does it means `Primary` can only been `undefined` here? is `Primary` needed then? same question for `welcomePhoneInput` below I guess

adam-stover: These types are used with `react-navigation` to define the routes and any params these routes expect -- setting up `Primary: undefined` allows us to do, for example, `navigator.navigate("Primary")` and TS will auto-suggest `"Primary"`. See https://reactnavigation.org/docs/typescript#type-checking-the-navigator

app/navigation/stack-param-lists.ts
adam-stover: yes this is unnecessary, good catch -- the `PhoneValidationStackParamList['authentication']` has the definition we need (to use in `phone-auth.tsx`).

daviroo (approved):
Reviewed on a call with @adam-stover

PR #350, opened 2021-12-13
fix: use a reactive-var to reset cache after mutations

samerbuna:
Include refactoring of BalanceHeader to not invoke its hooks
while the main query is loading, and a few other minor fixes

This commit also fixes the Balance 0 problem after navigating to
contacts screen. This was an Apollo cache missing field problem
because the contacts screen was querying the me field without and id

### DIFF
--- a/app/screens/contacts-screen/contacts.tsx
+++ b/app/screens/contacts-screen/contacts.tsx
@@ -1,5 +1,4 @@
 import { gql, useQuery } from "@apollo/client"
-import { useFocusEffect } from "@react-navigation/native"
 import { StackNavigationProp } from "@react-navigation/stack"
 import * as React from "react"
 import { useCallback, useMemo, useState } from "react"
@@ -99,12 +98,12 @@ export const ContactsScreen: ScreenType = ({ navigation }: Props) => {
   const { hasToken } = useToken()
   const [matchingContacts, setMatchingContacts] = useState([])
   const [searchText, setSearchText] = useState("")
-  const [isRefreshed, setIsRefreshed] = useState(false)
 
-  const { loading, data, error, refetch } = useQuery(
+  const { loading, data, error } = useQuery(
     gql`
       query contacts {
         me {
+          id
           contacts {
             username
             alias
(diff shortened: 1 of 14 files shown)


app/screens/receive-bitcoin-screen/qr-view.tsx
```diff
@@ -247,7 +247,6 @@ const styles = EStyleSheet.create({
   },
 
   lottie: {
-    color: palette.darkGrey,
```
samerbuna, line 250: @daviroo Lottie was complaining about this property stating that it's invalid.

app/screens/contacts-screen/contacts.tsx
```diff
@@ -99,12 +98,12 @@ export const ContactsScreen: ScreenType = ({ navigation }: Props) => {
...
   const [matchingContacts, setMatchingContacts] = useState([])
   const [searchText, setSearchText] = useState("")
-  const [isRefreshed, setIsRefreshed] = useState(false)
 
-  const { loading, data, error, refetch } = useQuery(
+  const { loading, data, error } = useQuery(
     gql`
       query contacts {
         me {
+          id
```
samerbuna, line 106: This is what fixed the balance 0 problem. Was not easy to find.

app/screens/contacts-screen/contacts.tsx
```diff
@@ -113,16 +112,12 @@ export const ContactsScreen: ScreenType = ({ navigation }: Props) => {
         }
       }
     `,
-    { skip: !hasToken },
+    {
+      skip: !hasToken,
+      fetchPolicy: "cache-and-network",
```
samerbuna, line 117: Without `cache-and-network`, once the data is persisted, Apollo does not go back to the server (even when the component re-renders fresh!). 

I think we need `cache-and-network` in most cases.

samerbuna:
@daviroo a snapshot change is failing the tests. I'm not sure why. Can you please take a look?

app/screens/contacts-screen/contacts.tsx
daviroo: Is this because the cache is stored using this `id` as the key?  So if you don't provide one then it will become `null` on every refresh?

app/screens/receive-bitcoin-screen/qr-view.tsx
daviroo: We were adding colors to fonts because some devices were defaulting to white text for some reason.  Perhaps this got added by mistake.

daviroo (approved):
I've fixed the tests and tested locally and it seems to be working.  This is definitely a hot fix and when we come to refactor we should come up with a better way of doing this.  I'm going to merge this now as it fixes a pretty important bug in the app.

PR #360, opened 2021-12-15
fix: change where to send mutation to avoid race condititon

samerbuna:


daviroo (approved):
We paired up on this 👍

PR #361, opened 2021-12-15
fix: refactor price cache to fix the ... problem

samerbuna:


samerbuna:
@daviroo this should fix the `...` problem. Note, it'll still show up if on initial load there is no price data from server or cache, but it was actually showing up when there's price data from server, because of how to convert amount was depending on a ref. I killed the ref and made the `cachedPrice` reactive instead.

daviroo (approved):
I checked this branch out and tested it and as you said it doesn't fix the actual issue.  If it improves things then happy to merge 👍 .

PR #363, opened 2021-12-16
Fix: Remove thrown errors in code.

### LINKED ISSUE #362: Errors from server cause app to not render
**Describe the bug**
There was a recent regression brought into the staging environment which caused an error in MainQuery. @nicolasburtey noticed that this caused the app to render a blank screen.

**To Reproduce**

1. Point app towards local setup.
2. Cause a server side error to occur during the response for MainQuery
3. See that the app shows a blank screen

**Expected behavior**
The app should render and display an error.

daviroo:
Fixes #362

### DIFF
--- a/app/hooks/use-wallet-balance.ts
+++ b/app/hooks/use-wallet-balance.ts
@@ -33,8 +33,12 @@ export const useWalletBalance = (): {
   const wallet = data?.me?.defaultAccount?.wallets?.[0]
 
   if (!wallet) {
-    // If this is thrown, it means Apollo Cache is not able to read the MAIN_QUERY from the cache (because of missing fields)
-    throw new Error("Something went wrong")
+    // If we get here, it means Apollo Cache is not able to read the MAIN_QUERY from the cache (because of missing fields)
+    return {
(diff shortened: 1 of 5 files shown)


app/hooks/use-wallet-balance.ts
```diff
@@ -33,8 +33,12 @@ export const useWalletBalance = (): {
   const wallet = data?.me?.defaultAccount?.wallets?.[0]
 
   if (!wallet) {
-    // If this is thrown, it means Apollo Cache is not able to read the MAIN_QUERY from the cache (because of missing fields)
-    throw new Error("Something went wrong")
+    // If we get here, it means Apollo Cache is not able to read the MAIN_QUERY from the cache (because of missing fields)
+    return {
+      satBalance: 0,
```
nicolasburtey, line 38: could we use NaN instead of 0?

it may be better to show that something is wrong, instead of showing the balance of the user to 0 (which could make a user think his balance is gone)

samerbuna (approved):

daviroo:
This was replaced by #370

PR #367, opened 2021-12-20
Add a button to blockchain explorer for onchain txns

leesalminen:
This PR adds a button to the Transaction Details screen for on-chain transactions. The button links to a blockchain explorer for the given hash. This allows the user to view the status of their on-chain transaction (say, if it's currently pending, they can estimate how many blocks out they are until confirmed).


daviroo (changes requested):
Thanks @leesalminen.  This would be a good addition.  Would it be possible to make the entire row a link, including the "Hash" row title and the hash value?  I think this would make the touchable area of the button larger which helps our accessibility score on the google play store.  Also the tests will need to be fixed and the code needs formatted to comply with `check-code` command.

leesalminen:
Hi @daviroo, thanks for the feedback. I've made the requested change to the touchable area and fixed the tests and formatting.

daviroo (approved):

daviroo:
Thanks for your contribution @leesalminen 🎉

PR #368, opened 2021-12-21
fix: Add global error handling.

daviroo:
I think this PR can replace #363.  It adds global error handling which will catch any fatal errors in the code and display an alert to the user instead of a white screen which was the previous behaviour.  I've also added crashlytics reporting of the errors - it remains to be seen how useful this will be and we may want to use a different service for this in the future.  I also added the ability for the user to restart the app and contact support in the dialogs.  Tested in both ios and android in my local environment.  Screenshots below.  Ignore the dev debug screens behind the alerts - these won't show in production.

### iOS

**English**
image

**Spanish**
image
@dolcalmi please check the translation

### Android

**English**
image

**Spanish**
image
@dolcalmi please check the translation


package.json
```diff
@@ -89,6 +90,7 @@
     "react-native-push-notification": "^8.1.1",
     "react-native-qrcode-svg": "^6.0.6",
     "react-native-reanimated": "^2.2.3",
+    "react-native-restart": "daviroo/react-native-restart.git#master",
```
samerbuna, line 93: Why do we need this fork?

daviroo: The released version of this wasn't working for me and there have been further changes on their master branch since.  I tested their main branch and it works but jest was unable to resolve the module without the lib directory which is generated as part of the build (not stored on github).  I branched the project and added the lib directory.

samerbuna (dismissed):

nicolasburtey:
@daviroo please use rebase instead of merge

PR #370, opened 2021-12-23
Refactor main query usage into a single hook

daviroo:
This is a WIP PR to refactor all `mainQuery` usage into a single hook.  It should mean that state is managed a lot better.  I also fixed the `@app` module importing.  This PR will need a lot of testing so I'm going to need to pick it back up tomorrow but I wanted to raise this draft so you can see where I'm going with it.


daviroo:
I've noticed an issue with the error handling code I added last week.  During testing for this change I encountered a react state change error (actually a warning) which is non fatal for the app but it actually triggered the [error handling function](https://github.com/GaloyMoney/galoy-mobile/blob/5b504dde79f4417d91a626bec9245fb898fb30d4/app/components/global-error/global-error.tsx#L16).  There is a boolean which is passed to the error handling called `isFatal` which should let you know if the error is fatal or not, but during my testing this does not accurately represent actual fatal errors.

TLDR as part of this change I'm going to have to rip out the error handling I added previously and come up with a new strategy.

daviroo:
> TLDR as part of this change I'm going to have to rip out the error handling I added previously and come up with a new strategy.

I decided to keep this change clean and do the change for error handling in a separate PR.

samerbuna (approved):
Looks good. I tested a few operations locally against production and everything seems ok.

The changes however block running the app with no backend, so locally you'll have to start the backend to run the app, in order to switch it to mainnet since the app defaults to regtest. That's not ideal but let's maybe tackle it in a different PR.

daviroo:
> Looks good. I tested a few operations locally against production and everything seems ok.
> 
> The changes however block running the app with no backend, so locally you'll have to start the backend to run the app, in order to switch it to mainnet since the app defaults to regtest. That's not ideal but let's maybe tackle it in a different PR.

Yeah let's discuss the best way to solve this on slack.  We definitely would want to fix this before release as I guess a user could kill their app by accessing the dev menu and pointing towards local by mistake.

PR #380, opened 2022-01-07
refactor: Refactored transactions screen

daviroo:
I'm refactoring the transaction screen.  I think I found out why we saw really strange cache behaviour, especially in our paginated endpoints.  Apollo caching takes into account the variables passed to graphql functions when creating cache keys.  This makes sense if, for example, you are querying for a user using a user ID - you would want the cache to store different users under different locations in the cache.  But this totally breaks when it comes to paginated endpoints as it stores the query results for different pages in different cache keys.  In the apollo docs it includes a section on [how to use the relay style of pagination](https://www.apollographql.com/docs/react/pagination/cursor-based/#relay-style-cursor-pagination) (which we use) in the cache.  I think this is a pattern we should bring across to the web wallet @samerbuna.  It's not working fully yet, the `fetchMore()` function runs but then the original query for the first 20 runs again and causes a refresh loop.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,3 +5,4 @@ android
 ignite/ignite.json
 package.json
 __generated__
+app/components/price-graph/data-tst.ts
--- a/app/graphql/cache.ts
+++ b/app/graphql/cache.ts
@@ -1,7 +1,17 @@
-import { InMemoryCache } from "@apollo/client"
+import { InMemoryCache, PossibleTypesMap } from "@apollo/client"
 import "moment/locale/es"
+import possibleTypes from "@generated/possibleTypes.json"
+import { relayStylePagination } from "@apollo/client/utilities"
+import * as tslib from "tslib"
+
+const notExtras = ["edges", "pageInfo"]
+const getExtras = function (obj) {
+  return tslib.__rest(obj, notExtras)
+}
 
 export const cache = new InMemoryCache({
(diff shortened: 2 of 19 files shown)


.prettierignore
```diff
@@ -5,3 +5,4 @@ android
 ignite/ignite.json
 package.json
 __generated__
+app/components/price-graph/data-tst.ts
```
daviroo, line 8: This file causes prettier to take much longer to run.  It's just a massive json blob so I've ignored it.

app/graphql/cache.ts
```diff
@@ -19,5 +29,40 @@ export const cache = new InMemoryCache({
         },
       },
     },
+    BTCWallet: {
+      fields: {
+        // transactions: relayStylePagination(),
+        transactions: {
+          ...relayStylePagination(),
+          read: (existing, _a) => {
```
daviroo, line 37: Overriding the read function provided by `relayStylePagination()`.

app/graphql/query.ts
```diff
@@ -88,9 +88,6 @@ export const MAIN_QUERY = gql`
           id
           balance
           walletCurrency
-          transactions(first: 3) {
```
daviroo, line 91: Now only querying the transactions from the transaction query.

daviroo:
Closed this as it's pretty stale now.

PR #381, opened 2022-01-08
Resolve USD to BTC issue with lnurl-pay

leesalminen:
Our beta testers identified an issue with the lnurl-pay implementation. The app was not converting the USD amount to BTC properly when hitting the lnurl-pay callback URL. This resulted in an invoice of the incorrect amount being created on the back-end.


daviroo (approved):
Good catch from the beta testers.  I checked out main and confirmed the issue then I checked out your branch and confirmed the fix.  Thanks again for the contribution.

PR #383, opened 2022-01-10
Add better error handling for 401 status codes

leesalminen:
This PR adds better handling for 401 status codes coming from the graphql endpoint. This is helpful for jwt secret rotation. When the mobile app receives a 401 response while logged in, we should log the user out and prompt them to log back in.

### DIFF
--- a/app/screens/error-screen/error-screen.tsx
+++ b/app/screens/error-screen/error-screen.tsx
@@ -66,14 +66,14 @@ export const ErrorScreen = ({ error, resetError }) => {
 
   const logoutAction = async () => {
     try {
-        await logout()
-        resetError()
-        Alert.alert(translate("common.reauth"), "", [
-          {
-            text: translate("common.ok"),
-            onPress: () => console.log("OK pressed"),
-          },
-        ])
+      await logout()
+      resetError()
+      Alert.alert(translate("common.reauth"), "", [
+        {
+          text: translate("common.ok"),
+          onPress: () => console.log("OK pressed"),
+        },
+      ])
     } catch (err) {
       // TODO: figure out why ListItem onPress is swallowing errors
       console.error(err)
@@ -85,7 +85,7 @@ export const ErrorScreen = ({ error, resetError }) => {
       setLogoutShown(true)
       logoutAction()
     }
-    
+
     return null
   }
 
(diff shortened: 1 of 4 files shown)


app/screens/error-screen/error-screen.tsx
```diff
@@ -59,7 +60,35 @@ const styles = EStyleSheet.create({
...
+      await logout()
+      resetError()
+      Alert.alert(translate("common.reauth"), "", [
+        {
+          text: translate("common.ok"),
+          onPress: () => console.log("OK pressed"),
+        },
+      ])
+    } catch (err) {
+      // TODO: figure out why ListItem onPress is swallowing errors
```
daviroo, line 78: Not sure what this TODO means.  Can you give more detail?

app/screens/error-screen/error-screen.tsx
```diff
@@ -59,7 +60,35 @@ const styles = EStyleSheet.create({
...
   useEffect(() => crashlytics().recordError(error), [error])
+
+  const logoutAction = async () => {
+    try {
+      await logout()
+      resetError()
+      Alert.alert(translate("common.reauth"), "", [
+        {
+          text: translate("common.ok"),
+          onPress: () => console.log("OK pressed"),
```
daviroo, line 74: Please remove the `console.log()` statement.

daviroo (changes requested):
I'm not sure about dealing with this in the `error-screen`.  It's a little bit of a mixing of concerns.  The `error-screen` should be for errors which break the functionality of the app, not a generic error handler.  I think the proper way to handle this would be to use [apollo link](https://www.apollographql.com/docs/react/data/error-handling/#on-graphql-errors) and have a global handler for all calls which respond with 401.  Let me know if you agree/disagree.

app/screens/error-screen/error-screen.tsx
leesalminen: That's a fair point. I took a stab at implementing this as an Apollo Link from `app/app.tsx`, but the problem I was running into was that I couldn't use the `useLogout` hook from outside of a component. So, that's when I tried it this way from `error-screen`.

One thing I noticed after submitting this PR was https://github.com/GaloyMoney/galoy-mobile/blob/main/app/components/global-error/global-error.tsx. It looks like there was code that handled 401 status codes at some point in there, but it doesn't seem to get called from any place anymore. Maybe it would make more sense to get this code working again?

leesalminen: > Not sure what this TODO means. Can you give more detail?

This was copy-pasted from https://github.com/GaloyMoney/galoy-mobile/blob/main/app/screens/settings-screen/settings-screen.tsx#L45 and doesn't belong here. If we continue down this path, I will remove.

leesalminen:
I'm withdrawing this PR. Everything is good as-is currently.

PR #384, opened 2022-01-11
fix: Handle 401 Unauthorized Responses from GraphQL

leesalminen:
This PR has the mobile app respond to 401 Unauthorized responses from the GraphQL API by notifying the user and  This is a small change that fixes existing code that wasn't being triggered as expected.


daviroo (approved):
We discussed this change on slack.  LGTM.

PR #388, opened 2022-01-17
fix: Fix error messages generated by useMainQuery

daviroo:


app/hooks/use-main-query.ts
```diff
@@ -23,7 +23,9 @@ const useMainQuery = (): useMainQueryOutput => {
         crashlytics().recordError(e)
         console.log(e)
       })
-      throw new Error(translate("errors.fatalError"))
+      throw new Error(
+        "Error occurred during main query execution prior to cache population.  Unable to confirm if data necessary for proper execution of the app is present.",
```
nicolasburtey, line 27: why removing the translation?

daviroo: This is for developers.  There is a user friendly error with translation on the screen itself [here](https://github.com/GaloyMoney/galoy-mobile/blob/a141893529d0883804b8e31fd9a51457bba7b810/app/screens/error-screen/error-screen.tsx#L74).

PR #391, opened 2022-01-26
fix: Fix translation in username validation

daviroo:
The translation wasn't working for username validation error messages.  After some investigation I found out that this is because we were setting up the error messages as part of a class which was instantiated before the i18n context had been setup - therefore always defaulted to English.  I've changed this to be functional rather than a class and it works.


app/utils/validation.ts
```diff
@@ -7,61 +7,53 @@ export class InvalidUsernameError extends Error {
...
-        case this.hasNoRestictedStartCharacters(username):
-          return new InvalidUsernameError(this.InvalidStartCharacters)
-        default:
-          return new InvalidUsernameError(this.InvalidUnexpected)
-      }
+export const validateUsername = (username: string) => {
+  const usernameRegex = /(?!^(1|3|bc1|lnbc1))^[0-9a-z_]{3,50}$/i
+
+  if (!username.match(usernameRegex)) {
+    switch (false) {
```
nicolasburtey, line 14: `switch(false)`? what is the goal of this statement?

daviroo: To be clear I didn't add this code I was just moving it into a function.  Looks like the switch is looking for cases which return false (i.e. fail validation in some way).

Issue #316, opened 2021-11-30
Add LNURL integration


krtk6160:
## Title
Add [LNURL](https://github.com/fiatjaf/lnurl-rfc/blob/legacy/lnurl-pay.md) and [Lightning address](https://github.com/andrerfneves/lightning-address) integration

## Description
From #63 :

> currently payment in the form of [lnurl-pay](https://github.com/fiatjaf/lnurl-rfc/blob/master/lnurl-pay.md) are not accepted by the mobile application.
> 
> lnurl-pay is very useful in the context of printed QR code, such that a merchant could have a printed QR code and their customer could pay by simply scanning the QR code, whether they use the Galoy wallet or [any other wallets](https://github.com/fiatjaf/awesome-lnurl#wallets) that accept lnurl-pay.
> 
> Eventually we want to work on lnurl with our own platform, but on the mobile side, the task can be developped independantly by using other platforms. a good way to test this is to use [this messaging platform](https://chat.blixtwallet.com/).
> 
> on scanning the QR code from https://chat.blixtwallet.com/, our wallet should ask for the invoice, have a screen for confirmation and then pay it.
> 
> BlueWallet has a [complete integration of lnurl](https://github.com/BlueWallet/BlueWallet/search?q=lnurl) so looking at their repo is a good way to understand the whole flow.

| Date added    | Bounty Amount |
|---------------|---------------|
| 23/11/2021    | 0.03 BTC |

pseudozach:
hi, I'm attempting this bounty, will update here with progress.

pseudozach:
I pushed the code, pending review/feedback: https://github.com/GaloyMoney/galoy-mobile/pull/321

krsjoseph:
Is this bounty still open, or is it closed?

pretyflaco:
Can this issue be closed? Afaik, LNURL-Pay has been merged. Or we keeping this open for LNURL-withdraw etc?

PR #392, opened 2022-01-27
refactor: Enable paste lnurl

danwebb1:
Addresses issue #316 

Send Bitcoin screen was able to handle lnurl from scanned QR code but was not able to handle user pasting lnurl directly.

- added ability to paste lnurl from send bitcoin screen 
- added clipboard icon to bottom of QR code, if user has copied it to clipboard and taps the clipboard it will paste and transition to send bitcoin screen to confirm, similar to Wallet of Satoshi and Strik

https://user-images.githubusercontent.com/17483038/151304524-8c7487ab-3a25-4be0-8e32-1bac5f83c890.mov

### DIFF
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -522,7 +522,8 @@
     "invalidContent": "We found:\n\n{{found}}\n\nThis is not a valid Bitcoin address or Lightning invoice",
     "invalidTitle": "Invalid QR Code",
     "noQrCode": "We could not find a QR code in the image",
-    "title": "Scan QR Code"
+    "title": "Scan QR Code",
+    "invalidContentLnurl": "We found:\n\n{{found}}\n\n is not currently supported"
   },
   "SecurityScreen": {
     "biometricDescription": "Unlock with fingerprint or facial recognition.",
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -565,8 +565,7 @@
     "note": "optional note",
     "success": "Payment has been sent successfully",
     "title": "Send Bitcoin",
-    "usernameNotFound": "A user matching the entered username could not be found.",
-    "invalidLnurl": "Invalid lnurl"
+    "usernameNotFound": "A user matching the entered username could not be found."
   },
   "SettingsScreen": {
     "activate": "Tap to activate",
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -564,7 +564,8 @@
     "note": "optional note",
     "success": "Payment has been sent successfully",
     "title": "Send Bitcoin",
-    "usernameNotFound": "A user matching the entered username could not be found."
+    "usernameNotFound": "A user matching the entered username could not be found.",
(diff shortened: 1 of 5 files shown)


daviroo:
Hey @danwebb1 .  I think this PR looks good.  I haven't had time to check it out to test yet unfortunately.  My only question would be does the clipboard modal pickup that a lnurl is in the clipboard and prompt the user?  Code is here https://github.com/GaloyMoney/galoy-mobile/blob/b626cd4e757e9da7fb45699b98a2444f67f3095b/app/utils/clipboard.ts#L19.

danwebb1:
> Hey @danwebb1 . I think this PR looks good. I haven't had time to check it out to test yet unfortunately. My only question would be does the clipboard modal pickup that a lnurl is in the clipboard and prompt the user? Code is here
> 
> https://github.com/GaloyMoney/galoy-mobile/blob/b626cd4e757e9da7fb45699b98a2444f67f3095b/app/utils/clipboard.ts#L19
> 
> .

Yes, it does pick up and alert the user. Additionally, I added slightly more specific error message to Alert on QR screen. If a lnurl-withdraw/auth/channel request is pasted instead of lnurl-pay it says that the its not currently supported instead of sayings its not a valid address.

app/i18n/en.json
```diff
@@ -522,7 +522,8 @@
     "invalidContent": "We found:\n\n{{found}}\n\nThis is not a valid Bitcoin address or Lightning invoice",
     "invalidTitle": "Invalid QR Code",
     "noQrCode": "We could not find a QR code in the image",
-    "title": "Scan QR Code"
+    "title": "Scan QR Code",
+    "invalidContentLnurl": "We found:\n\n{{found}}\n\n is not currently supported"
```
daviroo, line 526: The Spanish translation for this is missing.

app/i18n/en.json
```diff
@@ -564,7 +565,8 @@
     "note": "optional note",
     "success": "Payment has been sent successfully",
     "title": "Send Bitcoin",
-    "usernameNotFound": "A user matching the entered username could not be found."
+    "usernameNotFound": "A user matching the entered username could not be found.",
+    "invalidLnurl": "Invalid lnurl"
```
daviroo, line 569: I don't see this being used in the actual code anywhere.

app/i18n/es.json
```diff
@@ -603,7 +604,8 @@
     "payFromUSD": "Pago desde $USD",
     "success": "El pago se ha enviado correctamente.",
     "title": "Enviar Bitcoin",
-    "totalExceed": "total excede su saldo"
+    "totalExceed": "total excede su saldo",
```
daviroo, line 607: Not part of your change but I just happened to notice that this is only used in the `SendBitcoinConfirmationScreen` so I think this is an unnecessary key - could be removed.

daviroo (changes requested):
I checked it out and tested and it works ok.  Just a few minor comments RE translations.

daviroo (approved):

PR #394, opened 2022-02-05
fix: Fix username input

daviroo:
We had an issue where a user paid to the wrong username because the `defaultWalletId` forwarded to the `SendBitcoinConfirmation` screen was incorrect.  This was because the debounced function which fetches the `defaultWalletId` had returned for a previous valid username and the user was then able to navigate to the next page which cancelled the most recent invocation of the debounced function.  I've added logic to block the user from continuing to the next page while we are executing checks for the current username value.

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-screen.tsx
@@ -299,7 +299,8 @@ export const SendBitcoinScreen: ScreenType = ({
         userDefaultWalletIdQueryDebounced()
       }
     }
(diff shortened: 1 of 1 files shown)


samerbuna (approved):
Looks good.

Debounce is used in a few other places, we should probably check these as well.

daviroo:
@samerbuna On further testing it looks like cancel is actually the proper way to do it since that will execute the function reference at the trailing edge.  I think what might have happened is that the `useMemo` function which contains the debounced function is returning a new debounced function every time the destination changes.  On my testing it looks like you need to use the same debounced ref to get the expected behaviour so I'm going to change this a little. Using the loading indicators in the send payment button will also help to prevent this from happening.

PR #397, opened 2022-02-16
ci: releasing, trigger on tag push, bump on both android and ios build completion

sandipndev:
Upon merging this and using this CircleCI configuration, if you commit a build won't trigger but it would trigger if you push a tag.


daviroo (dismissed):

sandipndev:
Builds are failing because the rollout has already been done it looks like @daviroo 
Is it possible to test this PR out?
Do a code level whitespace change create a new build?

sandipndev:
would add auto versioning in this PR itself, so making it draft

sandipndev:
Will be making a fresh PR with the newer trigger from Concourse model.

Issue #390, opened 2022-01-24
Add a continue button on the phone number input screen on phone auth


daviroo:

secondl1ght:
PR submitted #399

PR #399, opened 2022-02-25
feat: Add continue button to phone auth screen

secondl1ght:
As requested in issue #390 I have added a Continue button to the phone auth screen. Now users can either use the submit button on the keyboard or they can dismiss the keyboard and hit the Continue button when they are ready. I can adjust the styling if you had something different in mind for the design.

Cheers!

Screenshot:
image

### DIFF
--- a/app/i18n/en.json
+++ b/app/i18n/en.json
@@ -621,7 +621,8 @@
     "header": "Enter your phone number, and we'll text you an access code.",
     "headerVerify": "Verify you are human",
     "placeholder": "Phone Number",
-    "verify": "Click to Verify"
+    "verify": "Click to Verify",
+    "continue": "Continue"
   },
   "WelcomePhoneValidationScreen": {
     "errorLoggingIn": "Error logging in. Did you use the right code?",
--- a/__tests__/screens/__snapshots__/phone-auth.spec.tsx.snap
+++ b/__tests__/screens/__snapshots__/phone-auth.spec.tsx.snap
@@ -710,6 +710,75 @@ exports[`WelcomePhoneInputScreen render matches snapshot 1`] = `
               }
             />
           </View>
+          <View
+            style={
+              Array [
+                Object {
+                  "borderRadius": 3,
+                  "overflow": "hidden",
+                },
+                Object {
+                  "borderRadius": 3,
+                },
+                undefined,
+                false,
+              ]
+            }
+          >
+            <View
+              accessibilityRole="button"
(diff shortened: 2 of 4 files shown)


secondl1ght:
I made a few updates: I added some padding to make the button a little bit bigger, I added some vertical margin so that it has more spacing during the loading screen and I also made the button disable when the phone number is set.

I want to do a bit more additional testing but I have been blocked from submitting due to too many attempts, so I will try again later.

Updated screenshot:
image

app/i18n/en.json
```diff
@@ -621,7 +621,8 @@
     "header": "Enter your phone number, and we'll text you an access code.",
     "headerVerify": "Verify you are human",
     "placeholder": "Phone Number",
-    "verify": "Click to Verify"
+    "verify": "Click to Verify",
+    "continue": "Continue"
```
daviroo, line 625: Please add this key in the `es.json` file as well.

daviroo (changes requested):

daviroo:
Hey @secondl1ght.  We will need to add the `continue` key to the Spanish translation file otherwise the app will crash for users in El Salvador.  Also the `yarn check-code` and `yarn test` commands are failing.  Those commands will need to run successfully before we can merge.

secondl1ght:
Thanks for the review! I added the Spanish translation (I just looked it up online so hopefully it is correct) and I made sure the 2 yarn commands pass now. I just had to refactor the disabled attribute on the new button to be a boolean value in order to pass the type check.

daviroo:
> Thanks for the review! I added the Spanish translation (I just looked it up online so hopefully it is correct) and I made sure the 2 yarn commands pass now. I just had to refactor the disabled attribute on the new button to be a boolean value in order to pass the type check.

The `yarn test` command is still failing with a broken snapshot.  Did you push the changes after updating the snapshots?

secondl1ght:
> > Thanks for the review! I added the Spanish translation (I just looked it up online so hopefully it is correct) and I made sure the 2 yarn commands pass now. I just had to refactor the disabled attribute on the new button to be a boolean value in order to pass the type check.
> 
> The `yarn test` command is still failing with a broken snapshot. Did you push the changes after updating the snapshots?

After updating the snapshots there was nothing to commit so I think that file is git-ignored. It tells you in the terminal output how to fix it though - if I remember correctly you need to run the command with a flag, I believe it was `yarn test -u`.

PR #401, opened 2022-02-28
Fix: Change `LN_PAGE_DOMAIN` to `GALOY_PAY_DOMAIN`, Use `GALOY_PAY_DOMAIN` constant in LNURL screen.

leesalminen:
This PR changes the new LNURL screen to use the `LN_PAGE_DOMAIN` constant instead of the hardcoded string of `ln.bitcoinbeach.com`. This makes it easier for other instances of Galoy to incorporate this change into their repo.

### DIFF
--- a/app/screens/settings-screen/lnurl-screen.tsx
+++ b/app/screens/settings-screen/lnurl-screen.tsx
@@ -79,7 +79,10 @@ export const LnurlScreen: ScreenType = ({ route }: Props) => {
     ),
     1500,
   )
-  const lnurlAddress = `${username}@${LN_PAGE_DOMAIN.replace('https://', '').replace('/', '')}`
+  const lnurlAddress = `${username}@${LN_PAGE_DOMAIN.replace("https://", "").replace(
+    "/",
+    "",
+  )}`
   const viewPrintableVersion = (): Promise<Linking> =>
     Linking.openURL(`${LN_PAGE_DOMAIN}${username}/print`)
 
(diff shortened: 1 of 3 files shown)


app/screens/settings-screen/lnurl-screen.tsx
```diff
@@ -10,6 +10,7 @@ import { palette } from "../../theme/palette"
 
 import type { ScreenType } from "../../types/jsx"
 import type { RootStackParamList } from "../../navigation/stack-param-lists"
+import { LN_PAGE_DOMAIN } from "../../constants/support"
```
krtk6160, line 13:
```suggestion
import { GALOY_PAY_DOMAIN } from "../../constants/support"
```

We renamed lnpage to galoy-pay sometime ago

krtk6160 (changes requested):
Galoy-pay should be used in place of lnpage

app/screens/settings-screen/lnurl-screen.tsx
krtk6160: > Are you saying you would like me to rename the constant across the whole app in this PR?

That is correct. It's only used in two more places -> https://github.com/GaloyMoney/galoy-mobile/search?q=ln_page_domain so I think it makes sense to include that change in this PR since it touches that code anyway

leesalminen: I've pushed the requested change.

daviroo (approved):

PR #413, opened 2022-03-20
test: add browserstack remote testing and streamlines release to happen only on tag push, also release to huawei store

sandipndev:
Adds Appium Browserstack setup to run local webdriver tests in debug mode on Real Devices.


sandipndev:
okay I guess if we use Development Profile, only then it's gonna build for ios. Doing that via a search and replace.

daviroo (dismissed):
I've sanity checked this @sandipndev.  I can see the tests running in browserstack.  As long as it's working I think we can merge it and iterate improvements later.

daviroo (approved):

PR #426, opened 2022-04-03
feat: use galoy-client for message translations

samerbuna:


samerbuna:
@daviroo I tested the translation move to galoy-client in this branch and it's all working. However:

- One test is failing, not sure why but probably related to updates. I've updated React and Apollo to match galoy-client.
- I had to make a few changes to error handling and I commented out one error that was overtaking the invalid authentication logic.

daviroo (approved):
I had to remove the failing test.  I actually think the test has been broken since it was written but that a bug in the testing library was causing it to pass.  I'll create an issue in the repo to recreate the test as I didn't want to spend too long figuring it out right now.

PR #440, opened 2022-05-16
fix: workaround for useSubscription multiple use

samerbuna:


daviroo (approved):
Code looks fine but I don't know why the title mentions useSubscription?

PR #453, opened 2022-05-23
feat: Add fee probe mutations to no retry list

daviroo:


daviroo:
@dolcalmi is it appropriate to add all these fee probe endpoints into the no retry list?

Issue #411, opened 2022-03-16
Add support for recognizing `lightning` param in BIP 21 QR codes


sbddesign:
## Background
An issue more broadly in the Bitcoin ecosystem is situations where the user must choose whether to generate an on-chain Bitcoin address or a lightning invoice. Bitcoin Beach wallet falls into this category since it supports both on-chain and lightning.

While Bitcoin Beach will scan a QR code of a BIP 21 payment URI, it will only recognize the on-chain address and not a `lightning` parameter containing a BOLT 11 invoice. I would love it if Bitcoin Beach wallet could support scanning a BIP-21 QR with a `lightning` param.

## Steps to Replicate

![image](https://user-images.githubusercontent.com/2414177/158688145-7fe10237-1e40-4608-9337-88fbe14a5f3f.png)

```
bitcoin:bc1qylh3u67j673h6y6alv70m0pl2yz53tzhvxgg7u?amount=0.00001&label=sbddesign%3A%20For%20lunch%20Tuesday&message=For%20lunch%20Tuesday&lightning=lnbc10u1p3pj257pp5yztkwjcz5ftl5laxkav23zmzekaw37zk6kmv80pk4xaev5qhtz7qdpdwd3xger9wd5kwm36yprx7u3qd36kucmgyp282etnv3shjcqzpgxqyz5vqsp5usyc4lk9chsfp53kvcnvq456ganh60d89reykdngsmtj6yw3nhvq9qyyssqjcewm5cjwz4a6rfjx77c490yced6pemk0upkxhy89cmm7sct66k8gneanwykzgdrwrfje69h9u5u0w57rrcsysas7gadwmzxc8c6t0spjazup6
```

- Open Bitcoin Beach wallet
- Scan the above QR code
- Results: it takes me to the payment screen and fills the `To` field with the on-chain address.
- Expected results: it should instead fill the `To` field with the BOLT 11 invoice from the `lightning` param in the QR code (or, it could prompt the user to choose which to use, but I think it would be easier on the user if it defaulted to lightning).

[See this website](https://bitcoinqr.dev/) for more background information.

## Screenshots

![_bitcoin-beach_1](https://user-images.githubusercontent.com/2414177/158688878-3de7771f-9e23-458e-8404-3bedb4c85613.png)

![_bitcoin-beach_2](https://user-images.githubusercontent.com/2414177/158688882-bdfc38a8-111a-42af-bdb3-6065c8898e93.png)

[...]

PR #454, opened 2022-05-23
fix: detect lightning param in bip 21 QR codes

allenwhite:
This PR addresses [this issue](https://github.com/GaloyMoney/galoy-mobile/issues/411), where we do not recognize the lightning param in BIP 21 QR codes. 

In completing this work, I refactored the `validPayment` method, which was getting a little long. The existing logic seemed to initially determine the `paymentType`, and then create the payment response based on this type. I attempted to reflect this approach through some new functions.

To detect the lightning param, I leveraged the `query-string` node module which was already present in the project as a dependency. 

All existing test cases pass, and I added a case for this specific piece of functionality, as well as a couple other cases we were missing.


daviroo:
Hi @allenwhite.  Thanks for raising the PR.  I tried to scan the example QR code on the linked issue but received the following error.

image

allenwhite:
Good call @bodymindarts - I didn't see any guidelines the the contributing readme, but now that I know, I will clean up the commits.

@daviroo I got that as well, but when I debugged it, it seemed the root cause was because the invoice was expired. Running locally also gave me that error due to the network being testnet. I commented out those checks in `getLightningPayResponse` and saw the expected behavior. I also included the QR in the tests I added to confirm it was working there (since we override `moment.now()`). Let me know if you are seeing something different tho.

daviroo:
> @daviroo I got that as well, but when I debugged it, it seemed the root cause was because the invoice was expired. Running locally also gave me that error due to the network being testnet. I commented out those checks in `getLightningPayResponse` and saw the expected behavior. I also included the QR in the tests I added to confirm it was working there (since we override `moment.now()`). Let me know if you are seeing something different tho.

I think in the case that the lightning invoice has expired we should instead pay to the onchain address.  Can you incorporate this into your change?  Also the commits still don't seem right, did you use rebase?

allenwhite:
Closing in favor of [galoy-client/pull/42](https://github.com/GaloyMoney/galoy-client/pull/42)

PR #456, opened 2022-05-24
fix: probe usd invoices

UncleSamtoshi:


daviroo (changes requested):
Just a couple of things I noticed.  Also there are some code check errors to fix before we can merge.

samerbuna (approved):
@daviroo @UncleSamtoshi: Merging this but please review as I made a few changes to handle loading/error states for fees.

PR #463, opened 2022-05-26
Fix car phone validation

daviroo:


daviroo:
There's something wrong with our `yarn.lock` file but I will look into it in a separate issue.

PR #477, opened 2022-06-02
feat: input and memo input in receive screens

samerbuna:


daviroo (approved):
I'm going to merge this and fix the tests and check-code on usd-integration branch so that we can release.

PR #482, opened 2022-06-05
Feat: show errors on the quiz screen

### LINKED ISSUE #467: Error handling for quiz mutation
I've noticed that there is no error handling in the quiz section. For example, if you use a phone that's not allowed to earn sats, answering a quiz question should fail in the UI (right now, the mutation fails but the UI shows no indicators).

TimoSolo:
fixes #467
image


samerbuna (changes requested):
This looks great. Thanks!

Do you mind redoing this on the `usd-integration` branch? On that branch, the mutations is done through the galoy-client package, and the error message mapping is done there as well, so you can use the `errorsMessage` value returned from the galoy-client operation.

TimoSolo:
Created a new PR at https://github.com/GaloyMoney/galoy-mobile/pull/483
Closing this PR

PR #483, opened 2022-06-05
Show errors on quiz screens

### LINKED ISSUE #467: Error handling for quiz mutation

TimoSolo:
fixes #467 on usd-integration branch


TimoSolo:
the toast is quite small.. is there a better way to show these errors perhaps?
image

PR #485, opened 2022-06-07
feat: BTC first and single wallet fallback

samerbuna:


daviroo (approved):
This looks ok to me.  I know there are dicussions going on regarding the design but these things can be improved later.

PR #492, opened 2022-06-10
Fixes prep for release

daviroo:


app/hooks/user-hooks.ts
```diff
@@ -95,10 +95,13 @@ const usePriceCache = () => {
 
   const [cachedPrice, setCachedPrice] = React.useState(() => {
     if (initialBtcPrice) {
-      return initialBtcPrice
+      return initialBtcPrice?.formattedAmount
```
daviroo, line 98: I think this is the fix @samerbuna.  It was returning a full price object returned from main query instead of a number.

samerbuna (approved):

PR #495, opened 2022-06-16
fix: multiple property value on object

vindard:
## Description

I haven't tested this but I suspect it's a bug


daviroo:
I think this has been fixed in another PR @vindard.  Going to close this.

PR #499, opened 2022-06-21
fix: submit while fee is unknown

daviroo:


samerbuna:
The idea is to have [the backend return max fees when probe fails](https://github.com/GaloyMoney/galoy/pull/1343), and have the frontend use that to always enforce the greater-than-balance check.

daviroo:
> The idea is to have [the backend return max fees when probe fails](https://github.com/GaloyMoney/galoy/pull/1343), and have the frontend use that to always enforce the greater-than-balance check.

I agree that this is the best way to fix the issue.  Closing this for now.

Issue #562, opened 2022-07-13
Main balance unequal to total balance


UncleSamtoshi:
The main wallet balance is less than the sum of the usd wallet balance and the bitcoin wallet balance at times.

PR #563, opened 2022-07-14
fix: inconsistent wallet balances

UncleSamtoshi:
- Removes use of useWalletBalance from the component level and returns it to the screen level to prevent inconsistent balances across components
- addresses issue #562


daviroo:
@UncleSamtoshi did you manage to reproduce the issue locally?

daviroo (approved):

Issue #430, opened 2022-04-18
Recreate failing phone auth screen test


daviroo:
We had to remove a test which was failing after we upgraded a testing library.  I suspect this test has been broken since it was written and that upgrading the library just uncovered this fact.  I tried to fix the test at the time but I was unable to select the country picker button on the phone auth screen in order to fire the on press event.  The following commit shows the removed test https://github.com/GaloyMoney/galoy-mobile/pull/426/commits/9bfc220b109a43d0786d5a6af25ec937ee9d9f24.  We need to recreate the unit test for this screen and find a way to select the country picker.

PR #566, opened 2022-07-14
fix: test to check if country picker is visible was recreated

eimolina:
I recreated a test that had problems. I extracted the same code that uses the "PhoneInput" component, to have a component with the TestID field and be able to use it in the test. 
Ref https://github.com/GaloyMoney/galoy-mobile/issues/430


daviroo (changes requested):
Hey @eimolina.  Thanks for the contribution.  Can you remove the merge commits and instead use rebase to bring changes in from other branches?

eimolina:
@daviroo, I had to open a new PR to remove the merge commits. the new PR is https://github.com/GaloyMoney/galoy-mobile/pull/568. thanks for your review.

PR #567, opened 2022-07-15
fix: Fix pending onchain payment date

daviroo:
This PR ended up growing a bit of arms and legs because of the refactoring required.  I also added some new testing patterns for testing components.  One of the patterns I introduced was to use ts-auto-mock to create typed mocks.  This allows you to create a mock objects which matches the properties of a type (in this case a transaction).  However you only need to define the properties which are actually used in the function, but avoid the type warnings of having missing properties.  This should make it easier to test functions which use typed objects as inputs.  More details here:
https://typescript-tdd.github.io/ts-auto-mock/


jest.config.js
```diff
@@ -1,12 +1,14 @@
 module.exports = {
   preset: "react-native",
+  setupFiles: ["<rootDir>jest-ts-auto-mock-config.ts"],
```
UncleSamtoshi, line 3: Does this need to have the ```/```  between the dir and file name  so it would look like ```setupFiles: ["<rootDir>/jest-ts-auto-mock-config.ts"],```?

UncleSamtoshi (changes requested):
Looks good. Nice job adding in the testing! I just requested a few changes.

__tests__/components/transaction-date.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+describe("Display the createdAt date for a transaction", () => {
+  it("Displays pending for a pending onchain transaction", () => {
+    const mockedTransaction = createMock<GaloyGQL.Transaction>({
+      status: "PENDING",
+      createdAt: new Date().getDate(),
+    })
+
+    const { queryAllByText } = render(
+      <AppConfigurationContextNamespace.Provider
```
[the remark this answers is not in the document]
daviroo: I've removed this change.

jest.config.js
daviroo: When I googled it it does look like the leading `/` is required.  I've added it in.

yarn.lock
```diff
@@ -1880,10 +1894,10 @@
   resolved "https://registry.yarnpkg.com/@expo/sdk-runtime-versions/-/sdk-runtime-versions-1.0.0.tgz#d7ebd21b19f1c6b0395e50d78da4416941c57f7c"
   integrity sha512-Doz2bfiPndXYFPMRwPyGa1k5QaKDVpY806UJj570epIiMzWaYyCtobasyfC++qfIXVb5Ocy7r3tP9d62hAQ7IQ==
 
-"@galoymoney/client@^0.1.33":
-  version "0.1.33"
-  resolved "https://registry.yarnpkg.com/@galoymoney/client/-/client-0.1.33.tgz#3695000029eaf7931b6a180a979b6d8bec798796"
-  integrity sha512-qFinuoH5B4jnonaUn7F+xhltFJw2VQarOGKc0EAulGLHqCbypyIB0w/Ra3Q6SQig5ozEkpQFEuGMi/TipE/7zg==
+"@galoymoney/client@^0.1.36":
+  version "0.1.36"
```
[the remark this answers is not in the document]
daviroo: Rebased against main which is using the latest version.

app/screens/contacts-screen/contact-transactions.tsx
```diff
@@ -119,7 +119,9 @@ export const ContactTransactionsDataInjected: ScreenType = ({
 
   for (const txEdge of transactionsRef.current) {
     const tx = txEdge.node
-    if (isToday(tx)) {
+    if (tx.status === "PENDING") {
```
UncleSamtoshi, line 122: Change to ```if (tx.status === "PENDING || isToday(tx))```?

__tests__/components/transaction-date.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    })
+
+    const { queryAllByText } = render(
+      <AppConfigurationContext.Provider
+        value={{
+          appConfig: defaultConfiguration,
+          setAppConfig: (config: AppConfiguration) => {},
+        }}
+      >
+        <TransactionDate tx={mockedTransaction} />
```
UncleSamtoshi, line 28: ```TransactionDate``` component doesn't rely on AppConfigurationContext, so the provider is not necessary for these tests.

UncleSamtoshi (approved):
Left a couple of comments, but this can be merged after this is completed..

app/screens/contacts-screen/contact-transactions.tsx
daviroo: Done

__tests__/components/transaction-date.spec.tsx
daviroo: Yeah I think I copied this pattern from another test and didn't consider that it wasn't needed.  Removed now.

PR #570, opened 2022-07-19
fix: Fixed bug where price cache wouldn't populate

daviroo:


daviroo:
@samerbuna @UncleSamtoshi I'm going to merge this without review for now to move forward with the release.  I won't release the app until you have reviewed but I want to at least kick off the build.

PR #573, opened 2022-07-20
fix: Fixed price cache initialising with 0

daviroo:
On some screens the price cache was sending down 0 because the values in the `useEffect` call were not updating in between the screen rendering and attempting to utilise the price cache in some operation.


daviroo:
@UncleSamtoshi @samerbuna I'm going to merge this without approval.  It's a small change and it would be good to include it in the release to avoid any bugs from cropping up (we already found one which @UncleSamtoshi fixed but there might be others we haven't found yet).

PR #574, opened 2022-07-20
chore: fixed github release uploads

daviroo:


sandipndev:
Ahh so it was a path issue, I see.

PR #575, opened 2022-07-20
fix: removed query all packages permission on android

daviroo:


daviroo:
This is also blocking the release on android @UncleSamtoshi @samerbuna so going to merge it in.

nicolasburtey:
@daviroo AFAIR changing the permission will require explicit approval for the user to upgrade. you should not take this slightly.

PR #582, opened 2022-07-25
fix: Fixed lnurl payments issue

daviroo:
lnurl payments were broken since the refactor since the `useFee` hook was being passed a null value.  The fix was to ensure that lnurl invoices are being treated as fixed amount invoices.

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
@@ -415,7 +415,7 @@ const SendBitcoinDetailsScreen = ({
     }
     navigation.navigate("sendBitcoinConfirmation", {
       lnurlInvoice: invoice,
-      fixedAmount,
+      fixedAmount: paymentType === "lnurl" ? paymentAmountInBtc : fixedAmount,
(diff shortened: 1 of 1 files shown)


daviroo:
@samerbuna @UncleSamtoshi can one of you give this a quick look?  It fixes lnurl payments in the app.

UncleSamtoshi (dismissed):

UncleSamtoshi:
I think this should be fixed in the useFee hook. I'm not quite sure I understand what the issue is, but an lnurl invoice is not a fixed amount invoice, and I think it would be better to change useFee to handle this case rather than incorrectly use the fixedAmount variable so that it works with the existing useFee.

daviroo:
> I think this should be fixed in the useFee hook. I'm not quite sure I understand what the issue is, but an lnurl invoice is not a fixed amount invoice, and I think it would be better to change useFee to handle this case rather than incorrectly use the fixedAmount parameter so that it works with the existing useFee.

An lnurl invoice is a fixed amount invoice. Once the user has chosen the amount we generate an invoice for that specific amount for them to pay using the lnurl url (at least that's how I understand it).

UncleSamtoshi (approved):

PR #585, opened 2022-07-26
fix: improve transfer flow

UncleSamtoshi:
- rename transfer to convert
- separate conversion process into seperate steps
- redesign conversion confirmation
- create useUsdBtcAmount hook that isolates logic for multiple currency inputs


UncleSamtoshi:
This pr is waiting on an update to @galoymoney/client

daviroo (approved):
Looks good 👍 .  I think we should try and add unit tests for all our future PRs though.  Not going to hold up this one though since it fixes a current issue.

daviroo:
> This pr is waiting on an update to @galoymoney/client

Woops I missed this comment.  Should I not have merged this?

PR #586, opened 2022-07-27
chore: Fixed end to end testing framework

daviroo:
This PR enables end to end testing on browserstack.  There is only one very simple test right now - will add more tests tomorrow.


UncleSamtoshi:
How does this work with version numbers? what versions will the test builds get and will we accidentally increment twice?

UncleSamtoshi:
Apart from my question on version numbers, I think this looks good.

daviroo:
> How does this work with version numbers? what versions will the test builds get and will we accidentally increment twice?

If you check out the Fastfiles you can see the steps involved in the build. There's a specific step for incrementing the build number which I'm not running in the end to end testing lanes. So should be good 👍🏻

UncleSamtoshi (approved):

PR #592, opened 2022-08-07
fix: refactor hooks

UncleSamtoshi:
- add price context and theme context
- remove unused hooks
- refactor hooks to remove deeply nested graphql dependencies


app/hooks/use-price-conversion.ts
```diff
@@ -0,0 +1,69 @@
...
+    [priceData],
+  )
+
+  return {
+    convertCurrencyAmount,
+    convertPaymentAmount,
+    usdPerBtc: {
+      currency: WalletCurrency.USD,
+      amount: priceData.initialized ? priceData.price * 100000000 : NaN,
+    } as PaymentAmount<WalletCurrency.USD>,
```
daviroo, line 66: I don't understand why the compiler isn't complaining here.  `NaN` is a possible return value for amount but it isn't included in the `PaymentAmount` type.

app/hooks/use-price-conversion.ts
```diff
@@ -0,0 +1,69 @@
+import { usePriceContext } from "@app/store/price-context"
+import { PaymentAmount, WalletCurrency } from "@app/types/amounts"
+import * as React from "react"
+
+export const usePriceConversion = () => {
+  const priceData = usePriceContext()
+  const convertCurrencyAmount = React.useCallback(
```
daviroo, line 7: I know this isn't your code but could we add types for args and return value for this function?

app/hooks/use-price-conversion.ts
```diff
@@ -0,0 +1,69 @@
...
+      if (from === "USD" && to === "BTC") {
+        return (100 * amount) / priceData.price
+      }
+      return amount
+    },
+    [priceData],
+  )
+
+  const convertPaymentAmount = React.useCallback(
+    (paymentAmount, toCurrency: WalletCurrency) => {
```
daviroo, line 24: Same here - I know this isn't your code but could we add types for args and return value for this function?

app/store/price-context.tsx
```diff
@@ -0,0 +1,97 @@
...
+      initialized: true
+    }
+
+type PriceContextType = {
+  priceData: PriceData
+  setPrice: (price: number, priceDate?: Date) => void
+}
+
+type PriceReducerAction = {
+  type: "set_price"
```
daviroo, line 22: Not a big deal but this string should be a const which can be exported so that when we are firing actions from other components we avoid typos which could cause the event not to fire.

daviroo (changes requested):
When I first saw how many files had changed I thought this was going to be harder to review :sweat_smile:. While testing the app locally I noticed that the bug with the sat balance showing as `...` has returned.

UncleSamtoshi:
> 😅. While testing the app locally I noticed that the bug with the sat balance showing as `...` has returned.

What are the steps to reproduce the "..."? I haven't had that issue myself.

app/hooks/use-price-conversion.ts
UncleSamtoshi: Yeah I'm not sure why this works either, however it is the expected behavior in several parts of the app currently.

daviroo (approved):

PR #598, opened 2022-08-10
fix: fix transaction formatting issues

UncleSamtoshi:
- fix strange borders on move money screen on ios
- fix section headers on transactions page on ios
- clean up contacts page


daviroo (approved):
This looks a lot better than what we had :+1:

PR #599, opened 2022-08-11
fix: remove rendering quirks

UncleSamtoshi:
- prevent login causing double render flow
- default to cache on main query
- fix bug with navigating to earn screen

### DIFF
--- a/app/hooks/use-main-query.ts
+++ b/app/hooks/use-main-query.ts
@@ -13,7 +13,6 @@ const useMainQuery = (): useMainQueryOutput => {
   const { convertCurrencyAmount } = usePriceConversion()
   const { data, previousData, error, loading, refetch } = useQuery(MAIN_QUERY, {
     variables: { hasToken },
-    fetchPolicy: "cache-and-network",
   })
   let errors = []
   if (error) {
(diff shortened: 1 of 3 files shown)


app/hooks/use-main-query.ts
```diff
@@ -13,7 +13,6 @@ const useMainQuery = (): useMainQueryOutput => {
   const { convertCurrencyAmount } = usePriceConversion()
   const { data, previousData, error, loading, refetch } = useQuery(MAIN_QUERY, {
     variables: { hasToken },
-    fetchPolicy: "cache-and-network",
```
daviroo, line 16: How much testing did you do for this change? Last time we changed the fetch policy for main query we ended up with lots of bugs.

daviroo (changes requested):
Just one comment regarding testing for the mainquery change.  Looks good otherwise.

app/hooks/use-main-query.ts
UncleSamtoshi: Not a lot of testing on this one. When was the last time we changed the network policy?

daviroo: It was when we did the work to migrate to API v2, so quite a while ago.  I remember we had lots of issues with transactions not appearing in the transaction list after a call to send a new transaction for instance.  I guess we could merge and test this on `main` but it would need quite a bit of edge case testing before making it's way to prod IMO.

daviroo (approved):
Approved but needs tested on `main`

Issue #552, opened 2022-07-07
expire invoice should no longer show QR code/invoice data


nicolasburtey:
see screenshot below that says the invoice has expired. we should no longer show the content because it's irrelevant now. we could blur the QR code but maybe there are some other ways to do this. 

![image](https://user-images.githubusercontent.com/11156370/177830689-e3f45af5-f71d-4bcb-97ce-3dc5627a4b3a.png)

Issue #540, opened 2022-07-06
Use expiration from invoice for UI


UncleSamtoshi:
USD invoice expired after 5 min (based on the content of the invoice) but the interface say they expire after 2 minutes

PR #600, opened 2022-08-11
fix: Fixed receive screen bugs

### LINKED ISSUE #526: Paid invoice falsely says expired
A2D07E35-1BF0-43F2-A470-29E7523DA79F-4529-00001774D346640C

daviroo:
Fixes #526 , #552 and #540


app/screens/receive-bitcoin-screen/receive-usd.tsx
```diff
@@ -145,6 +146,45 @@ const ReceiveUsd = () => {
...
+            () => setStatus("expired"),
+          )
+        }
+      }
+      appState.current = nextAppState
+    })
+    return () => {
+      subscription.remove()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
UncleSamtoshi, line 176: Why not include ```setStatus``` as one of the dependencies?

app/hooks/use-countdown-timer.ts
```diff
@@ -0,0 +1,85 @@
...
+    newCallback?: () => void,
+  ) => {
+    setTimeLeft(seconds)
+    if (completedCallback) {
+      completedCallback()
+    }
+    setCompletedCallback(() => newCallback)
+  }
+
+  const stopCountdownTimerAndExecuteCallback = (newCallback?: () => void) => {
```
UncleSamtoshi, line 69: I think this should be called ```clearCountdown...``` and I dont think it makes sense to pass in a new callback for stopping the countdown timer because it will just be executed immediately anyway. The function that calls this function can just execute the callback itself if it wants to.

app/hooks/use-countdown-timer.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  const resetCountdownTimer = (seconds: number, newCallback?: () => void) => {
+    setTimeLeft(seconds)
+    if (newCallback) {
+      setCompletedCallback(() => newCallback)
+    }
+  }
+
+  const resetCountdownTimerAndExecuteCallback = (
```
UncleSamtoshi, line 58: Maybe rename this to ```resetCountdownTimerAndExecuteExistingCallback```?

app/hooks/use-countdown-timer.ts
```diff
@@ -0,0 +1,85 @@
...
+  }, [timeLeft, completedCallback])
+
+  const startCountdownTimer = (seconds: number, callback?: () => void) => {
+    setTimeLeft(seconds)
+    if (callback) {
+      setCompletedCallback(() => callback)
+    }
+  }
+
+  const stopCountdownTimer = () => {
```
UncleSamtoshi, line 47: This will trigger the callback which I dont think is the expected behavior. ```stopCountdownTimer``` to me implies that you are pausing the time at its current value.

app/screens/receive-bitcoin-screen/receive-usd.tsx
```diff
@@ -145,6 +146,45 @@ const ReceiveUsd = () => {
   const { lnUpdate } = useSubscriptionUpdates()
   const [showMemoInput, setShowMemoInput] = useState(false)
   const [showAmountInput, setShowAmountInput] = useState(false)
+  const { timeLeft, startCountdownTimer, resetCountdownTimer } = useCountdownTimer()
+  const timeUntilInvoiceExpires = useMemo(() => {
+    if (invoice) {
+      return (
+        getLightningInvoiceExpiryTime(decodeInvoiceString(invoice.paymentRequest)) -
```
UncleSamtoshi, line 153: This value could be negative, and your hook doesnt currently handle negative values.

app/hooks/use-countdown-timer.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+      }
+
+      return () => {
+        clearInterval(interval)
+      }
+    }
+    // We need to return from all code paths but we only need a callback if an interval is set so returning undefined is fine
+    return undefined
+  }, [timeLeft, completedCallback])
```
UncleSamtoshi, line 38: Should ```setTimeLeft``` be a part of the dependency list?

app/screens/receive-bitcoin-screen/receive-usd.tsx
```diff
@@ -145,6 +146,45 @@ const ReceiveUsd = () => {
   const { lnUpdate } = useSubscriptionUpdates()
   const [showMemoInput, setShowMemoInput] = useState(false)
   const [showAmountInput, setShowAmountInput] = useState(false)
+  const { timeLeft, startCountdownTimer, resetCountdownTimer } = useCountdownTimer()
+  const timeUntilInvoiceExpires = useMemo(() => {
```
UncleSamtoshi, line 150: I think ```timeUntilInvoiceExpires``` is a misleading name, because on subsequent renders it is no longer actually the time until invoice expires. You can get rid of lines 150-158 all together and add that logic to the effect in line 179 while changing line 187 to just depend on the invoice.

app/screens/receive-bitcoin-screen/receive-usd.tsx
```diff
@@ -145,6 +146,45 @@ const ReceiveUsd = () => {
...
+    if (invoice) {
+      return (
+        getLightningInvoiceExpiryTime(decodeInvoiceString(invoice.paymentRequest)) -
+        Math.round(Date.now() / 1000)
+      )
+    }
+    return NaN
+  }, [invoice])
+
+  useEffect(() => {
```
UncleSamtoshi, line 160: You don't have to make this change, but one way to avoid having to keep the timer in sync like this would be to implement the hook by storing the expiration of the timer, and then each second calculating how much time is left, vs the current implementation where you store how much time is left but then it can get out of sync if you leave the app.

app/screens/receive-bitcoin-screen/receive-usd.tsx
daviroo: The intention is for this `useEffect` to only run once when the component mounts as it is setting a callback for when the appstate changes and doesn't need to be rerun when any of the dependencies change.

app/screens/receive-bitcoin-screen/receive-usd.tsx
daviroo: Yeah I agree and I have made the change.

app/screens/receive-bitcoin-screen/receive-usd.tsx
daviroo: I guess in practice this could only happen if the app came back from background and the invoice had already expired - but you're right we should just set the invoice as expired if the value is negative and not start the timer.

app/hooks/use-countdown-timer.ts
daviroo: Done

app/screens/receive-bitcoin-screen/receive-usd.tsx
daviroo: I've changed the way it works so now the hook utilises the dependencies.

app/hooks/use-countdown-timer.ts
daviroo: Yup changed

app/hooks/use-countdown-timer.ts
daviroo: Yup agreed, changed

app/hooks/use-countdown-timer.ts
daviroo: Yeah agreed - have added it to the dependency list

app/screens/receive-bitcoin-screen/receive-usd.tsx
daviroo: I made your suggested change.

daviroo:
@UncleSamtoshi I've made your suggested change and responded to your comments.  I'll leave the comments for you to resolve if you feel like I've addressed them appropriately.

UncleSamtoshi (approved):

PR #601, opened 2022-08-16
fix: dont trigger clipboard on own invoice

UncleSamtoshi:
Currently if a user creates an invoice, leaves the app, and then returns, the clipboard modal will ask if they want to open the invoice. This PR prevents the clipboard modal from being triggered for invoices created by the application.

### DIFF
--- a/app/utils/clipboard.ts
+++ b/app/utils/clipboard.ts
@@ -46,3 +46,13 @@ export const showModalClipboardIfValidPayment = async ({
 
   modalClipboardVisibleVar(true)
 }
+
+export const copyPaymentInfoToClipboard = (paymentInfo: string): void => {
+  cache.writeQuery({
+    query: LAST_CLIPBOARD_PAYMENT,
+    data: {
+      lastClipboardPayment: paymentInfo,
+    },
+  })
+  Clipboard.setString(paymentInfo)
+}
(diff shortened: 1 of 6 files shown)


app/utils/clipboard.ts
```diff
@@ -34,15 +34,25 @@ export const showModalClipboardIfValidPayment = async ({
...
 
-  const { valid } = parsePaymentDestination({
+  const { valid, paymentType } = parsePaymentDestination({
     destination: clipboard,
     network,
     pubKey: myPubKey,
   })
 
-  if (!valid) {
+  if (!valid || paymentType === "intraledger") {
```
daviroo, line 43: Am I understanding this right - this means that if the user has a pay url copied e.g. https://ln.bitcoinbeach.com/unclesamtoshi then it won't trigger the modal?

UncleSamtoshi: Yes. I will remove this line. Originally I thought it could be worth removing because the way that the galoy-client parses the string will return intraledger by default, so even if something is clearly not a username (such as your phone auth code) it will still trigger an invoice. However the better solution is to change the galoy-client to better parse the strings.

daviroo (dismissed):

daviroo (approved):

daviroo (approved):

PR #605, opened 2022-08-18
fix: Fixed deep linking issue

daviroo:
Fixes issue when scanning a printable QR code for BBW pay link from external camera app.


UncleSamtoshi:
Migrated to PR #607

PR #608, opened 2022-08-24
feat: add stablesats-modal

UncleSamtoshi:
- Introduce persistent storage provider. This storage provider is capable of data migrations and is loaded before the rest of the app. Several migrations have been demonstrated as a pattern.
- Add Stablesats modal that appears one time when a user has access to the stablesats feature. Acknowledging the notification is stored in the persistent state, so that it is not shown again. In the debug screen you can reset the persistent state to see the modal once again.


UncleSamtoshi:
This PR is waiting on translations for the Stablesats modal.

daviroo (approved):

PR #618, opened 2022-09-12
feat: username validation

UncleSamtoshi:
- build out the destination validation to enhance error messages and prevent sending to incorrect users


Extheoisah (approved):
Fine with me. @daviroo should probably review again

daviroo (approved):

PR #637, opened 2022-09-21
chore: Added generic input component

daviroo:


app/components/galoy-text-input/galoy-text-input.tsx
```diff
@@ -0,0 +1,110 @@
...
+export const GaloyTextInput = ({
+  placeholder,
+  label,
+  labelStyle,
+  append = "",
+  prepend = "",
+  onChangeText,
+}: GaloyTextInputProps) => {
+  const [value, setValue] = React.useState("")
+  const [cleanValue, setCleanValue] = useState("")
```
UncleSamtoshi, line 57: I think this may be an unnecessary duplication of state, because the two values are deterministically related. The ```value``` can be functionally derived by simply adding the prepend and/or append to the ```cleanValue```.  The risk would be that any time state is duplicated it could introduce bugs and complexity surrounding keeping them in sync.

app/components/galoy-text-input/galoy-text-input.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  useEffect(() => {
```
UncleSamtoshi, line 59: I think by using ```useEffect``` in this component you are introducing unnecessary rerenders. For example user calls setValue which triggers re-render with new value which triggers useEffect in line 68 which sets cleanedValue which causes a rerender which finally triggers useEffect in line 80. These could all happen in the same callback suchas:
```
handleChangeText( text) => {
    // filter text to cleanValue
    setCleanValue(cleanValue)
    onChangeText(cleanValue)
}
```

ios/Podfile.lock
```diff
@@ -856,7 +856,7 @@ EXTERNAL SOURCES:
 SPEC CHECKSUMS:
   boost: a7c83b31436843459a1961bfd74b96033dc77234
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
```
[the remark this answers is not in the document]
daviroo: Can you run the following commands and compare them to my output:
```
➜  ios git:(bbw-address-settings-screen) ✗ pod --version
Resolving dependencies...
1.11.3
➜  ios git:(bbw-address-settings-screen) ✗ ruby -v
ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [arm64-darwin21]
```

UncleSamtoshi (approved):

PR #647, opened 2022-10-11
Initial logic for lnurl-withdraw support

kngako:
Ability to scan and redeem lnurl withdrawRequest.


app/screens/redeem-lnurl-withdrawal-screen/redeem-btc.tsx
```diff
@@ -0,0 +1,556 @@
...
+            styles.container,
+            styles.bigText,
+            styles.textCenter,
+            styles.textPaddedBottom,
+          ]}
+        >
+          Redeeming Bitcoin from {domain}
+        </Text>
+        {/* TODO: Replace this with a spinning wheel waiting for payment */}
+        <QRView
```
daviroo, line 541: Why do we need a QR view here?

daviroo (changes requested):
First of all thanks for the contribution so far!

We had a look through the code and at a high level we think the following changes should be made:

We think the logic should be split between 3 different screens :
- RedeemBitcoinDetails (Where we would display the amount and ask the user which wallet they want to redeem the value into)
- RedeemBitcoinConfirmation (where we would display all the details of the payment alongside a confirm button)
- RedeemBitcoinSuccess (We use lottie in other flows to show a green tick when things are successful and navigate the user back to the move money screen)

For inspiration please see the [conversation success screen](https://github.com/GaloyMoney/galoy-mobile/tree/main/app/screens/conversion-flow)

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.tsx
```diff
@@ -0,0 +1,388 @@
...
+    if (receiveCurrency === WalletCurrency.Usd) {
+      navigation.setOptions({ title: LL.RedeemBitcoinScreen.usdTitle() })
+    }
+
+    if (receiveCurrency === WalletCurrency.Btc) {
+      navigation.setOptions({ title: LL.RedeemBitcoinScreen.title() })
+    }
+  }, [receiveCurrency, navigation, LL])
+
+  // useEffect(() => {
```
dolcalmi, line 105: please remove unused code

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.tsx
```diff
@@ -0,0 +1,388 @@
...
+
+    if (result.ok) {
+      const lnurlResponse = await result.json()
+      if (lnurlResponse?.status?.toLowerCase() === "ok") {
+        // TODO: Set processing payment
+      } else {
+        console.error(lnurlResponse, "error with redeeming")
+        // TODO: Set failed payment
+        setErr(LL.RedeemBitcoinScreen.redeemingError())
+      }
```
dolcalmi, line 248: is this still in progress? else please update the condition and remove the else block

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
```diff
@@ -0,0 +1,445 @@
...
+    if (receiveCurrency === WalletCurrency.Usd) {
+      navigation.setOptions({ title: LL.RedeemBitcoinScreen.usdTitle() })
+    }
+
+    if (receiveCurrency === WalletCurrency.Btc) {
+      navigation.setOptions({ title: LL.RedeemBitcoinScreen.title() })
+    }
+  }, [receiveCurrency, navigation, LL])
+
+  // useEffect(() => {
```
dolcalmi, line 156: please remove

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
```diff
@@ -0,0 +1,445 @@
...
+  })
+  const [usdAmount, setUsdAmount] = useState(satAmountInUsd)
+
+  const [amountCurrency, setAmountCurrency] = useState("BTC")
+
+  // const { btcWalletId } = useMainQuery()
+  const { data } = useReceiveBtcQuery({ fetchPolicy: "cache-first" })
+  const btcWalletId = data?.me?.defaultAccount?.btcWallet?.id
+
+  const usdWalletId = null // TODO: when usd wallet ln invoices can be generated providing the sats amount as in put we can have the usdWalletId from useMainQuery as follows: const { usdWalletId } = useMainQuery()
```
dolcalmi, line 217: @daviroo  can this be done using price val?

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.tsx
```diff
@@ -0,0 +1,388 @@
...
+      return (
+        <View style={styles.container}>
+          <View style={styles.errorContainer}>
+            <Text style={styles.errorText} selectable>
+              {err}
+            </Text>
+          </View>
+        </View>
+      )
+    } else if (!invoicePaid) {
```
dolcalmi, line 287: else is not necessary

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.tsx
UncleSamtoshi, line 105: Can you remove this unused code

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.tsx
```diff
@@ -0,0 +1,388 @@
...
+    }
+  }
+
+  useEffect((): void | (() => void) => {
+    if (withdrawalInvoice) {
+      submitLNURLWithdrawRequest(withdrawalInvoice)
+    } else {
+      createWithdrawRequestInvoice({ satAmount, memo })
+    }
+  }, [walletId, withdrawalInvoice, memo, type, satAmount, createWithdrawRequestInvoice])
```
UncleSamtoshi, line 261: In order to prevent missing dependencies in hooks, can you add submitLNURLWithdrawRequest (you might have to memoize it or use callback so that it doesnt keep rerendering). `type` and `walletId` aren't used in this hook, so I dont believe they should be dependencies here, but should be instead added as dependencies to `createWithdrawRequestInvoice`.

UncleSamtoshi (changes requested):
Getting really close! Tested it out locally and it was awesome to see it working. Left some comments and you will have to rebase to bring in some new type checking rules that we have.

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
UncleSamtoshi: We could try and create an invoice that was the correct sat amount using the usd wallet, but we can only specify usd amounts so we potentially could never get the correct amount of sats.

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.tsx
kngako: This actually shows the loading animation before we either get the invoice paid, or an error.

I should actually add it to it's own thing.

app/i18n/en/index.ts
```diff
@@ -519,6 +519,17 @@ const en: BaseTranslation = {
...
+  RedeemBitcoinScreen: {
+    title: "Redeem Bitcoin",
+    usdTitle: "Redeem for USD",
+    error: "Failed to generate invoice. Please contact support if this problem persists.",
+    redeemingError: "Failed to redeem Bitcoin. Please contact support if this problem persists.",
+    submissionError: "Failed to submit withdrawal request. Please contact support if this problem persists.",
+    minMaxRange: "Min: {minimumAmount: string} {currencyTicker: string}, Max: {maximumAmount: string} {currencyTicker: string}",
+    redeemBitcoin: "Redeem Bitcoin",
+    amountToRedeemFrom: "Amount to redeem from {domain: string}",
+    redeemAmountFrom: "redeem {amountToRedeem: string} {currencyTicker: string} from {domain: string}",
```
UncleSamtoshi, line 531: Can you capitalize `redeem` at the start of 531?

UncleSamtoshi:
Spoke with the team and the only other feedback would be that if the min and max value are the same for the lnurl withdraw, we shouldnt display min and max and the amount field shouldnt be editable.

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.tsx
```diff
@@ -0,0 +1,395 @@
...
+  }, [err, lnServiceErrorReason])
+
+  const renderActivityStatusView = useMemo(() => {
+    if (err === "" && !invoicePaid) {
+      return (
+        <View style={styles.container}>
+          <View>
+            <ActivityIndicator
+              size="large"
+              color={processingPayment ? palette.green : palette.blue}
```
dolcalmi, line 296: this should only be green when invoice is paid... with the current behavior is when lnurl withdraw service return a response

Extheoisah: that check is in line 224

UncleSamtoshi: The check mark doesnt come until it is paid in line 256

kngako: Yes, this is just to change the loading indicator colour when we get an okay response from the ln service processing the withdrawal.

UncleSamtoshi (approved):

PR #650, opened 2022-10-14
feat: migrate token to kratos

UncleSamtoshi:
- prepares the app for authentication changes in galoy-backend
- removes dependency on fields of jwt
- automatically migrates jwt to kratos-session-token


app/app.tsx
```diff
@@ -136,41 +134,35 @@ export const App = (): JSX.Element => {
...
 
-      if (token) {
-        currentNetwork = tokenNetwork
-      } else if (networkReactiveVar) {
-        currentNetwork = networkReactiveVar
-      } else {
-        currentNetwork = await loadNetwork()
-        networkVar(currentNetwork)
-      }
+      // used to migrate from jwt to kratos token, remove after a few releases
```
nicolasburtey, line 144:
```suggestion
      // TODO: used to migrate from jwt to kratos token, remove after a few releases
```

app/config/galoy-instances.ts
```diff
@@ -0,0 +1,30 @@
...
+import { scriptHostname } from "@app/utils/helper"
+
+export type GaloyInstance = {
+  name: "BBW" | "Staging" | "Local"
+  network: INetwork
+  graphqlUri: string
+  graphqlWsUri: string
+}
+
+export const GALOY_INSTANCES: GaloyInstance[] = [
```
nicolasburtey, line 11: maybe a follow up PR, but would be a good idea to be able to add new instance, even if just from the developer screen initially. there are things related to the federation that we could test out that way.

app/screens/settings-screen/language-screen.tsx
```diff
@@ -43,7 +42,7 @@ export const LanguageScreen: ScreenType = () => {
                     errors: [],
                     user: {
                       __typename: "User",
-                      id: tokenUid,
+                      id: me?.id,
```
nicolasburtey, line 45: fyi (I don't think there is anything to change):
- me.id will be returning the Account.id 
- user id will eventually be be migrated to be the kratosUserId

app/components/modal-clipboard/modal-clipboard.tsx
```diff
@@ -120,18 +121,16 @@ export const ModalClipboard: ComponentType = () => {
       if (clipboard) {
         const { paymentType } = parsePaymentDestination({
           destination: clipboard,
-          network: tokenNetwork,
+          network: appConfig.galoyInstance.network,
```
nicolasburtey, line 124: shouldn't be fetched from the graphql query instead?

app/config/galoy-instances.ts
```diff
@@ -0,0 +1,30 @@
...
+  {
+    name: "BBW",
+    network: "mainnet",
```
nicolasburtey, line 14: instead of hardcoding this, shouldn't this be fetch dynamically from the graphql endpoint?

daviroo (approved):

app/config/galoy-instances.ts
UncleSamtoshi: I will create an issue for a follow up PR.

PR #667, opened 2022-11-04
feat: e2e testing payments flow

ntheile:


daviroo (approved):

.circleci/config.yml
```diff
@@ -199,10 +199,8 @@ workflows:
     jobs:
       - test_ios:
           filters:
-            tags:
-              only: /^[0-9]+\.[0-9]+\.[0-9]+/
             branches:
-              ignore: /.*/
+              only: main
```
daviroo, line 203: I think this will break our tagged builds (releases).  Since the jobs `build_ios` and `build_android` depend on the testing jobs but the testing jobs no longer build on tags.  3 options I can think of:
1. Remove the test job dependencies from the build jobs (since every commit to main is getting tested anyway) - this is a little dangerous though because someone could theoretically create a tag before the tests have completed for a commit and introduce a breaking change.
2. Figure out a way to have the tests run on every commit and in the release builds (but seems wasteful).
3. Release every commit to main? (I kind of like the sound of this one)

ios/fastlane/Fastfile
```diff
@@ -46,7 +46,8 @@ platform :ios do
       browserstack_username: ENV["BROWSERSTACK_USER"],
       browserstack_access_key: ENV["BROWSERSTACK_ACCESS_KEY"],
     )
-    sh("npx wdio ../browserstack.js")
+    # sh("npx wdio ../browserstack.js")
```
daviroo, line 49:
```suggestion
```

android/fastlane/Fastfile
```diff
@@ -78,7 +78,8 @@ platform :android do
       browserstack_access_key: ENV["BROWSERSTACK_ACCESS_KEY"],
       file_path: lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH],
     )
-    sh("npx wdio ../browserstack.js")
+    # sh("npx wdio ../browserstack.js")
```
daviroo, line 81:
```suggestion
```

daviroo (changes requested):
Hey @ntheile sorry for the delayed review.  Just a few suggestions and one comment regarding the setup of the circleci config.

daviroo (approved):

PR #669, opened 2022-11-07
Fix android build issue

daviroo:
For more info on why this is necessary see -> https://github.com/facebook/react-native/issues/35210#issue-1436785719


daviroo:
I'm going to merge this without approval since the android build is currently broken and it's blocking the release.

PR #676, opened 2022-11-29
fix: fee amount error

Extheoisah:


app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -231,7 +233,19 @@ const SendBitcoinConfirmationScreen = ({
     paymentAmount: paymentAmountInWalletCurrency,
   })
 
-  const feeDisplayText = fee.amount && paymentAmountToTextWithUnits(fee.amount)
+  useEffect(() => {
+    if (fee.amount) {
+      try {
+        if (paymentAmountToTextWithUnits(fee.amount) === null) {
+          throw Error("Payment amount to text errored with null value")
```
nicolasburtey, line 240: `null` won't speak to non dev

```suggestion
          throw Error("Payment amount to text errored with invalid value")
```

daviroo (approved):

PR #677, opened 2022-11-29
chore: refactor paymentAmountToText function

Extheoisah:
This PR is a draft of my refactor of the paymentAmountToText function. It takes off from this PR [here](https://github.com/GaloyMoney/galoy-mobile/pull/676). It adds extra validation and catches ay errors before it gets to the component level in the send bitcoin screen.
However, it is not robust enough to handle some edge cases present in the parent function `paymentAmountToTextWithUnit` [here](https://github.com/GaloyMoney/galoy-mobile/blob/525cf4f0d6ded12ee4e308583853302be3fbdb17/app/utils/currencyConversion.ts#L45).
It still needs some work in my opinion.

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
@@ -1,5 +1,5 @@
 import DestinationIcon from "@app/assets/icons/destination.svg"
-import React, { useState } from "react"
+import React, { useState, useEffect } from "react"
 import { ActivityIndicator, ScrollView, StyleSheet, Text, View } from "react-native"
 import { palette } from "@app/theme"
 import { WalletCurrency } from "@app/types/amounts"
@@ -22,6 +22,7 @@ import { CommonActions } from "@react-navigation/native"
 import useMainQuery from "@app/hooks/use-main-query"
 import { useI18nContext } from "@app/i18n/i18n-react"
 import { logPaymentAttempt, logPaymentResult } from "@app/utils/analytics"
+import crashlytics from "@react-native-firebase/crashlytics"
 
 const styles = StyleSheet.create({
   scrollView: {
@@ -191,6 +192,7 @@ const SendBitcoinConfirmationScreen = ({
   const { usdWalletBalance, btcWalletBalance, btcWalletValueInUsd } = useMainQuery()
   const isNoAmountInvoice = fixedAmount === undefined
   const [, setStatus] = useState<Status>(Status.IDLE)
+  const [feeDisplayText, setFeeDisplayText] = useState<string>("")
 
   const paymentAmountInWalletCurrency =
     payerWalletDescriptor.currency === WalletCurrency.BTC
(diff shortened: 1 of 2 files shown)


app/utils/currencyConversion.ts
```diff
@@ -64,19 +64,33 @@ export const paymentAmountToText = (
...
 ): string => {
   if (paymentAmount.currency === WalletCurrency.USD) {
-    return (paymentAmount.amount / 100).toLocaleString(locale, {
-      style: "decimal",
-      maximumFractionDigits: 2,
-      minimumFractionDigits: 2,
-    })
+    try {
+      const amount = paymentAmount.amount / 100
+      if (amount) {
```
UncleSamtoshi, line 69: It seems that this `if` is unnecessary and so is line 76. Since the whole statement is inside of a try block, you can just do the calculation as normal and then catch any errors that are thrown from the function.

app/utils/currencyConversion.ts
```diff
@@ -64,19 +64,33 @@ export const paymentAmountToText = (
...
+        return amount.toLocaleString(locale, {
+          style: "decimal",
+          maximumFractionDigits: 2,
+          minimumFractionDigits: 2,
+        })
+      }
+      throw Error("unable to parse USD amount to text")
+    } catch (error) {
+      console.error(error)
```
UncleSamtoshi, line 78: Can we record the error in crashlytics as well.

app/utils/currencyConversion.ts
```diff
@@ -64,19 +64,33 @@ export const paymentAmountToText = (
...
+      return "USD amount parsing error"
```
UncleSamtoshi, line 79: I would return "USD formatting error"

app/utils/currencyConversion.ts
```diff
@@ -64,19 +64,33 @@ export const paymentAmountToText = (
...
-    })
+    try {
+      return paymentAmount.amount.toLocaleString(locale, {
+        style: "decimal",
+        maximumFractionDigits: 0,
+        minimumFractionDigits: 0,
+      })
+    } catch (e) {
+      console.error(e)
+      return "Unable to parse amount to"
```
UncleSamtoshi, line 92: Here you can return "BTC formatting error"

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -231,7 +233,19 @@ const SendBitcoinConfirmationScreen = ({
     paymentAmount: paymentAmountInWalletCurrency,
   })
 
-  const feeDisplayText = fee.amount && paymentAmountToTextWithUnits(fee.amount)
+  useEffect(() => {
+    if (fee.amount) {
+      try {
+        if (paymentAmountToTextWithUnits(fee.amount) === null) {
```
UncleSamtoshi, line 239: paymentAmountToTextWithUnits will never be null, so this try-catch block is unnecessary.

UncleSamtoshi (changes requested):

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
Extheoisah: > paymentAmountToTextWithUnits will never be null, so this try-catch block is unnecessary.

In the case that `fee.amount` returns an invalid value, the try catch block handles the error and records it in crashlytics. But I agree that `paymentAmountToTextWithUnits` will not return a null value, hence I'll remove the if block since its unnecessary

app/utils/currencyConversion.ts
Extheoisah: > Here you can return "BTC formatting error"
I think the issue here is  how this interpolates with the return statement in the parent function `paymentAmountToTextWithUnit` in [line 52](https://github.com/GaloyMoney/galoy-mobile/blob/046d285fc46dd4c3009ffa933984a93845607ff7/app/utils/currencyConversion.ts#L52)

app/utils/currencyConversion.ts
Extheoisah: > Can we record the error in crashlytics as well.

Not sure if it's okay to import the crashlytics callback in this file. If that's fine, then I can add it.

PR #693, opened 2022-12-05
fix: usd amount js extra decimals

ntheile:


app/screens/receive-bitcoin-screen/receive-usd.tsx
```diff
@@ -227,13 +227,14 @@ const ReceiveUsd = () => {
             hasAmount: true,
             receivingWallet: WalletCurrency.USD,
           })
+          const amountInCents = Number((parseFloat(usdAmount) * 100).toFixed(2))
```
UncleSamtoshi, line 230: amountInCents is still a float here but it should be an integer

UncleSamtoshi (changes requested):

UncleSamtoshi (approved):

PR #727, opened 2022-12-16
feat: add account limits screen

Extheoisah:
**Description**

This PR adds account limits screen to users settings in the app and should be merged after the corresponding [account limits PR](https://github.com/GaloyMoney/galoy/pull/1953) from the Galoy backend repo has been merged.

To test the PR, 
- run the Galoy backend locally and switch to [this](https://github.com/GaloyMoney/galoy/pull/1953) branch
- next, checkout to this commit [`65c649a`](https://github.com/GaloyMoney/galoy/commit/65c649a8ce617f06b4765bed77a4680eeb2accb7) and start the main API. 
- next, in the debugging section of the mobile app, change your backend development server to local and connect your android/iOS simulator/device to the local server (for android run `yarn adb` to sync ports)
- You should find the limits screen under `Settings -> Accounts -> Account Limits`

Update:
- updated the screen to not differentiate USD and BTC transactions.
Screenshot 2022-12-20 at 12 53 15
Screenshot 2022-12-18 at 23 04 56


daviroo:
Reviewing now.  Here are some screenshots for others to look at.

image

image

app/hooks/use-account-limits.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  if (error) {
+    crashlytics().recordError(error)
+  }
+
+  React.useEffect(() => {
+    if (data) setAccountLimits([limits])
+  }, [data, limits])
+
+  React.useMemo(
```
daviroo, line 76: Not sure what the purpose of this is?

app/screens/settings-screen/account-screen.tsx
```diff
@@ -59,6 +59,15 @@ export const AccountScreen = ({ navigation }: Props) => {
   }
 
   const accountSettingsList: SettingRow[] = [
+    {
+      category: LL.common.accountLimits(),
+      id: "limits",
+      icon: "",
```
daviroo, line 65: What about some type of exclamation icon?  e.g. `exclamationcircleo`.  You can see the choices here https://oblador.github.io/react-native-vector-icons/.

daviroo (changes requested):
Just a few comments.  There is a bit of margin at the top which seems out of place.  I've included a screenshot to show what I mean.

image

app/hooks/use-account-limits.ts
Extheoisah: > Not sure what the purpose of this is?

This persists the limits data and checks for changes in the user account limits. So this ensures real-time change as opposed to using a if statement

app/hooks/use-account-limits.ts
```diff
@@ -0,0 +1,108 @@
+import * as React from "react"
+
+import { ApolloError, ApolloQueryResult, gql, useQuery } from "@apollo/client"
+import crashlytics from "@react-native-firebase/crashlytics"
+
+const ACCOUNT_LIMITS_QUERY = gql`
```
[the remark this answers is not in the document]
Extheoisah: > Can you move this into galoy-client?

Not at the moment. Till the limits PR on the backend is merged. In my opinion, I'd suggest we hold off adding it to the galoy-client until we finalise the KYC mutations and query to avoid continuously making changes to the galoy client code

app/screens/settings-screen/account-screen.tsx
Extheoisah: > What about some type of exclamation icon? e.g. `exclamationcircleo`. You can see the choices here https://oblador.github.io/react-native-vector-icons/.

👍

app/hooks/use-account-limits.ts
```diff
@@ -0,0 +1,108 @@
...
+  const { data, loading, error, refetch } = useQuery(ACCOUNT_LIMITS_QUERY)
+  const limits: accountLimitsData = data?.me?.defaultAccount?.limits
+
+  const [accountLimits, setAccountLimits] = React.useState<accountLimitsData[]>(null)
+
+  const withdrawalLimitsRef = React.useRef<accountLimitPeriod>(null)
+  const internalSendLimitsRef = React.useRef<accountLimitPeriod>(null)
+  const convertLimitsRef = React.useRef<accountLimitPeriod>(null)
+
+  if (error) {
```
UncleSamtoshi, line 68: I think this needs to be wrapped in a useEffect so that on every rerender you dont record the same error.

app/hooks/use-account-limits.ts
UncleSamtoshi: I'm still having trouble understanding why you cant just use the limits directly. It seems that the state is unnecessarily duplicated  which also results in extra renders.

app/hooks/use-account-limits.ts
```diff
@@ -0,0 +1,108 @@
...
+    crashlytics().recordError(error)
+  }
+
+  React.useEffect(() => {
+    if (data) setAccountLimits([limits])
+  }, [data, limits])
+
+  React.useMemo(
+    () =>
+      accountLimits?.map((data) => {
```
UncleSamtoshi, line 78: What is the point in using references vs just return the memoized value?

app/hooks/use-account-limits.ts
```diff
@@ -0,0 +1,108 @@
...
+  convert: limitValue[]
+}
+
+type accountLimitPeriod = {
+  DailyAccountLimit?: limitValue
+  WeeklyAccountLimit?: limitValue
+}
+
+type useAccountLimitsOutput = {
+  withdrawalLimits: Readonly<accountLimitPeriod>
```
UncleSamtoshi, line 50: These types might be null right?

UncleSamtoshi (changes requested):

app/hooks/use-account-limits.ts
Extheoisah: Yes

app/hooks/use-account-limits.ts
Extheoisah: I'm pushing a refactor. Its much simpler and I eliminated the use of memoised values

app/screens/settings-screen/account-limits-screen.tsx
```diff
@@ -111,136 +100,77 @@ export const AccountLimitsScreen = () => {
...
+          </View>
+        </View>
+
+        <View style={styles.divider}></View>
+
+        <View style={styles.limitWrapper}>
+          <Text adjustsFontSizeToFit style={styles.valueFieldType}>
+            {LL.AccountLimitsScreen.withdraw()}
+          </Text>
+          {limits?.withdrawal.map((data, index: number) => {
```
UncleSamtoshi, line 124: You should check with @vindard if there will be any guarantee that they are sorted from shortest to longest duration. Otherwise we will have to find a way to sort it on the frontend.

app/screens/settings-screen/account-limits-screen.tsx
```diff
@@ -111,136 +100,77 @@ export const AccountLimitsScreen = () => {
...
+      <View style={styles.contentTextBox}>
+        <Text adjustsFontSizeToFit style={styles.valueRemaining}>
+          {`${usdAmountDisplay(
+            Number(data.remainingLimit),
+            0,
+          )} ${LL.AccountLimitsScreen.remaining().toLocaleLowerCase()}`}
+        </Text>
+        <Text adjustsFontSizeToFit style={styles.valueTotal}>
+          {`${usdAmountDisplay(Number(data.totalLimit), 0)} ${
+            data.__typename === "DailyAccountLimit"
```
UncleSamtoshi, line 168: This will also need to handle monthly limit, so it may be best to use a map or function that maps typename to a string representing the duration. As it is written currently, if a new month limit was added, it would display as a week limit.

UncleSamtoshi (changes requested):
I think this is the right idea for the current schema of the API!

app/screens/settings-screen/account-limits-screen.tsx
Extheoisah: waiting on a reply from him but it seems so from here src/graphql/types/object/account-limits.ts

daviroo: I think that even if the order is coming back from the API in the right way at the moment we should probably have code in the front end which orders it anyway.  If someone changes the order of an array on the backend they wouldn't expect that to change the rendering on the mobile app.

daviroo:
I'm happy for @UncleSamtoshi to approve once his concerns have been addressed.  I've only left 1 comment regarding ordering.

app/screens/settings-screen/account-limits-screen.tsx
vindard: Added this as a consideration under `point 3.` here: https://github.com/GaloyMoney/galoy/pull/1953#issuecomment-1357689726

daviroo:
@UncleSamtoshi I'm going to approve this one since it seems @Extheoisah addressed all the concerns.

daviroo (approved):

PR #728, opened 2022-12-16
Create currency display

daviroo:


app/hooks/use-display-currency.ts
```diff
@@ -0,0 +1,10 @@
+import { useState } from "react"
+
+export const useDisplayCurrency = () => {
+  const [displayCurrency, setDisplayCurrency] = useState("USD")
```
UncleSamtoshi, line 4: It doesn't appear this hook is used, do you have future plans for it? Would you add additional functionality to it because as it stands now it seems to be an unnecessary wrapper of useState.

app/components/text-currency/text-currency.tsx
```diff
@@ -30,8 +31,11 @@ export const TextCurrencyForAmount: ComponentType = ({
   iconColor = palette.black,
 }: Props) => {
   if (currency === "USD") {
-    const amountDisplay = Number.isNaN(amount) ? "..." : usdAmountDisplay(amount)
-    return <Text style={style}>{amountDisplay}</Text>
+    return (
+      <Text style={style}>
```
UncleSamtoshi, line 35: Here you are nesting DisplayCurrency in a Text tag, even though it is already a Text element. Would it make sense to have DisplayCurrency accept the text style, or have it not be a component at all but instead be a hook that returns the raw text value?

app/store/localization-context.tsx
```diff
@@ -1,11 +1,36 @@
...
+
+export const LocalizationContext = createContext<LocalizationContextType>(null)
 
 export const LocalizationContextProvider = ({ children }) => {
   const { userPreferredLanguage } = useMainQuery()
   const { locale, setLocale } = useI18nContext()
+  const [displayCurrency, setDisplayCurrency] = useState("USD")
+
+  const convertUsdToDisplayCurrency = useCallback(
+    (usdAmount: number) => {
```
UncleSamtoshi, line 20: Can we use `UsdPaymentAmount` here for the type? In general, I think we want to be using the types in this [file](app/types/amounts.ts) because as we start to work with multiple currencies, it will be really important for us to know what currencies we are dealing with.

app/components/transaction-item/transaction-item.tsx
```diff
@@ -168,7 +169,7 @@ export const TransactionItem: React.FC<TransactionItemProps> = ({
           >
             {primaryCurrency === "BTC" && tx.settlementCurrency === WalletCurrency.BTC
               ? satAmountDisplay(tx.settlementAmount)
-              : usdAmountDisplay(usdAmount)}
+              : convertUsdToDisplayCurrency(usdAmount)}
```
UncleSamtoshi, line 172: From my understanding, `convertUsdToDisplayCurrency` is not relevant in transation-items or transaction-detail-screens because we will be using what the display currency was at the time as that will be the only thing we have saved with the tx in the ledger. For example if you make some transactions with euros as your display currency and then you switch to another display currency, we will still show euros for that past transaction, because in the ledger we have recorded the euro display amount.  We are not able to convert to the new display currency because we would also need the historical exchange rate for that tx.

UncleSamtoshi: Instead for any component involving past transactions, they just need to be capable of taking a `PaymentAmount<Curreny>` and be able to properly display that.

UncleSamtoshi (changes requested):
Good first iteration!

app/components/transaction-item/transaction-item.tsx
daviroo: Yeah agreed but for now this is to test the currency set in debug screen.

app/store/localization-context.tsx
daviroo: I've moved the function into the custom hook as I think that is a better pattern to follow.  Can discuss in front end sync.

app/hooks/use-display-currency.ts
daviroo: The updated version of this hook now has additional features, but it still relies on the LocalizationContext. Once we receive the display currency from the backend, we can replace the Context with the Apollo client and the hook should function properly.

app/components/text-currency/text-currency.tsx
daviroo: I've removed this and it's now using the function to generate the display currency.

UncleSamtoshi (approved):

daviroo:
Had to rebase to main - going to merge based on @UncleSamtoshi approval.

PR #742, opened 2022-12-22
chore: Lowering jitpack priority

daviroo:


android/build.gradle
```diff
@@ -49,7 +49,6 @@ allprojects {
             url("$rootDir/../node_modules/jsc-android/dist")
         }
       maven { url 'https://maven.google.com' }
-      maven { url "https://www.jitpack.io" }
         mavenCentral {
```
UncleSamtoshi, line 52: does this break the schema? It almost looks like mavenCentral is some kind of child to maven

daviroo: No it shouldn't break the schema.

UncleSamtoshi (approved):

PR #764, opened 2022-12-30
feat: add slider component

Extheoisah:
Not a priority
open todo:
- [ ] add animation transition effect to slider


app/components/atomic/galoy-slider-button/galoy-slider-button.tsx
```diff
@@ -0,0 +1,158 @@
...
+  send = "slide to send",
+  sending = "send",
+  sent = "sent!",
+  failed = "failed",
+}
+
+const sliderStartValue = 6
+const sliderMaxValue = 100
+
+export const Sliders: React.FunctionComponent<GaloySliderComponentProps> = (props) => {
```
UncleSamtoshi, line 24: Right now the slider is able to go backwards a little bit, can we prevent this?

app/components/atomic/galoy-slider-button/galoy-slider-button.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+import { GaloyIcon } from "../galoy-icon"
+
+type GaloySliderComponentProps = {
+  disabled?: boolean
+  callback?: () => void
+  text?: string
+}
+
+enum sliderTextValues {
```
UncleSamtoshi, line 14: The slider may be used for other things besides sending (such as swapping). I would make it more generic by having the following props.
`initialText` `disabledText` `slidingText` `completedText` and use these instead of the sliderTextValues enum.

app/components/atomic/galoy-slider-button/galoy-slider-button.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const [value, setValue] = useState(sliderStartValue)
+  const [showIcon, setShowIcon] = useState<boolean>(true)
+  const [sliderText, setSliderText] = useState<sliderTextValues>(sliderTextValues.send)
+
+  React.useEffect(() => {
+    if (value === sliderStartValue) {
+      setSliderText(sliderTextValues.send)
+    }
+    if (value >= sliderMaxValue / 2) {
```
UncleSamtoshi, line 36: Can we have it auto-complete at a little more than halfway (maybe 2/3)?

UncleSamtoshi (changes requested):

app/components/atomic/galoy-slider-button/galoy-slider-button.tsx
UncleSamtoshi: Additionally we cannot have text hardcoded in components as they will need to be translated to the relevant languages in the parent component.

app/components/atomic/galoy-slider-button/galoy-slider-button.tsx
Extheoisah: I haven't found a way around this. but it snaps back to its original position if taken backwards.

daviroo:
@UncleSamtoshi @Extheoisah I tested this on a low end android phone (Samung Galaxy A10) and it performs really bad (unusable).

daviroo (changes requested):
This doesn't work on low end phones.

nicolasburtey:
closing this PR as we have another open PR for the slider to send

PR #774, opened 2023-01-02
chore: upgrade rn 0.70

nicolasburtey:


nicolasburtey:
> @nicolasburtey I see that you have removed lottie library and replaced the lottie component with a string. Is the idea that this would be merged after the redesign is complete? We use lottie in both send and receive flow so it will probably be quite a while before we are in a state to merge this.

I was considering to replace with the non interactive confirmed logo from the new design

.node-version
```diff
@@ -0,0 +1 @@
+16
```
daviroo, line 1: Should this be 18?

daviroo (approved):

PR #789, opened 2023-01-04
feat: add checkbox and radio button

Extheoisah:


app/components/atomic/galoy-checkbox-button/galoy-checkbox-button.tsx
```diff
@@ -0,0 +1,69 @@
...
+  buttonType: ButtonTypeOptions
+  disabled?: boolean
+  callback?: () => void
+}
+
+export const GaloyCheckboxButton = (props: GaloyCheckboxButtonProps & CheckBoxProps) => {
+  const { theme } = useTheme()
+  const styles = useStyles(theme)
+
+  const [checked, setChecked] = useState<boolean>(props.checked)
```
UncleSamtoshi, line 20: I think line 20 and line 22 are unnecessary. The component that is using the checkbox will be keeping track of the state of whether it is checked or not, so there is no point in duplicating it as well within the checkbox component. For example right now, if the parent component wanted to change whether or not the check box was checked, they would not be able to because the state is being set once at line 20. You can instead just pass the `checked` and `onPress` props directly through to the checkbox.

UncleSamtoshi (changes requested):

PR #790, opened 2023-01-04
feat: using graphql-ws subscription library

nicolasburtey:


daviroo (dismissed):

app/app.tsx
```diff
@@ -161,16 +162,17 @@ export const App = (): JSX.Element => {
...
-        },
-      })
+      const wsLink = new GraphQLWsLink(
+        createClient({
+          url: appConfig.galoyInstance.graphqlWsUri,
+          connectionParams: hasToken ? { Authorization: `Bearer ${token}` } : undefined,
+          // Voluntary not using: webSocketImpl: WebSocket
+          // seems react native already have an implement of the websocket?
+          //
+          // TODO: implement keepAlive and reconnection?
```
daviroo, line 172: Just wondering what the implications of not having these implemented is?  Will the websocket reconnect if it gets disconnected?

nicolasburtey: I don't think the behavior would be different than before.

if you look at the old library argument: https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L61-L71, you'll see that there is no option to do ping automatically.

the new library already have a `retryAttempt` with a default value of 5: https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/client.ClientOptions.md#retryattempts

I think the new library has the option to do PING/PONG in additional to actively detect a connection that might has been stale. but this is a new behavior afaiu, this can be implemented in another PR.

daviroo (approved):

PR #791, opened 2023-01-04
chore: test on every branch

nicolasburtey:


UncleSamtoshi (approved):

Extheoisah (approved):

ntheile:
this could lead to hitting rate limits in our backend if too many instances are running at once. for example, if i check in code to a feature branch, 2 instances kick off, 1 on ios and 1 on android. Then i commit again in under a few minutes, or somebody else on the team commits...we could have many instances running at once hitting the api limits, since we are using the same token (this already has happened in the past).

another approach could be to either run the test on a tag or regex in the branch name. maybe something like this:

```
  feature-branch:
    jobs:
      - test_android:
          filters:
            branches:
              only: /.*testme/
      - test_ios:
          filters:
            branches:
              only: /.*testme/
```

nicolasburtey:
> also there are still a few flakes in the login-flow tests...i.e when switching to staging and clicking the input on ios sometimes the keyboard does not show up properly. the test waits for the "save changes" button to show up, but sometimes it's hidden and needs to scroll down the screen. we can probably remove the logic to wait for the save changes button to appear.

yes we should strive to make the test more deterministic

nicolasburtey:
> we could have many instances running at once hitting the api limits, since we are using the same token (this already has happened in the past).

rate limit for which service? circleci? browserstack?

ntheile:
> rate limit for which service? circleci? browserstack?

for our graphql backend.

ntheile (approved):

PR #794, opened 2023-01-05
refactor: add types for use-main query

Extheoisah:


app/hooks/hooks.types.ts
```diff
@@ -8,6 +9,45 @@ type mobileVersions = {
   platform: string
 }
 
+type Wallet = {
```
nicolasburtey, line 12: type should be written manually. they should be imported from codegen 

ie: https://github.com/GaloyMoney/galoy-client/blob/36929772cec9c12c796a37b07b6c3dc61c2dfca9/package.json#L11

nicolasburtey (changes requested):
type should not be imported manually as they can be auto generated from graphql

app/hooks/hooks.types.ts
```diff
@@ -1,53 +1,7 @@
...
-    readonly wallets: Wallet[]
-  }
-  readonly id: string
-  readonly language: string
-  readonly phone: string
-  readonly quizQuestions: object[]
-  readonly username: string
-}
-
 export type useMainQueryOutput = {
```
nicolasburtey, line 5: the whole query response is given to you fully typed by codegen. you don't need to redefine it here.

Extheoisah: Yeah. I have removed the types I manually added. These types are needed for the useMainQuery output because we modify the data response in the hook here https://github.com/GaloyMoney/galoy-mobile/blob/4705306b1e9f24ae67c00506b610f358031a2e39/app/hooks/use-main-query.ts#L51

Extheoisah:
Closing this as we need better context on how to approach this topic

PR #797, opened 2023-01-05
fix: add missing useEffect dependencies

UncleSamtoshi:
-  fixes repeated logging out with expired session token
- evaluates the use of disable linting rules and reenables them where neccessary


nicolasburtey:
I have been away from the react hooks for too long to really review this PR effectively. will let @daviroo or @Extheoisah to approve. maybe @samerbuna also would have a better opinion than mine

Extheoisah (approved):
All good on my end.

PR #798, opened 2023-01-05
feat: more design components

UncleSamtoshi:
- add tertiary button
- add SourceSansPro font


app/components/atomic/galoy-tertiary-button/galoy-tertiary-button.tsx
```diff
@@ -0,0 +1,69 @@
...
+  )
+}
+
+const useStyles = makeStyles((theme, disabled) => ({
+  buttonTitleStyle: {
+    lineHeight: 20,
+    fontSize: 14,
+    fontWeight: "600",
+    color: theme.colors.primary,
+    opacity: disabled === true ? 0.7 : 1,
```
Extheoisah, line 67: no need to equate disabled to true since disabled is a always a boolean. 
```suggestion
    opacity: disabled ? 0.7 : 1,
```

Extheoisah (changes requested):

app/components/atomic/galoy-tertiary-button/galoy-tertiary-button.tsx
UncleSamtoshi: `disabled` is actually sometimes an empty object because of how `makeStyles` is implemented, however I will refactor it so that disabled is a property in the object.

Extheoisah (approved):

Extheoisah (approved):

PR #802, opened 2023-01-06
feat: revise input component design

Extheoisah:


app/components/atomic/galoy-input/galoy-redesigned-input.tsx
```diff
@@ -1,57 +1,110 @@
...
   ref: React.Ref<TextInput> & React.Ref<React.PropsWithChildren<InputProps>>,
 ) => {
   const { theme } = useTheme()
-  const styles = useStyles(theme)
   const [isFocused, setIsFocused] = React.useState(props.initIsFocused ?? false)
+  const styles = useStyles(props)
+
+  return (
+    <View>
+      <LabelComponent
```
UncleSamtoshi, line 42: What is the thinking behind using a custom LabelComponent instead of using the `labelStyle` and `labelProps` that are contained in the `Input` component? The same questions goes for the caption component. If it is necessary to use a CustomComponent, you can still pass that directly into the `label` prop of Input.

UncleSamtoshi (changes requested):

app/components/atomic/galoy-input/galoy-redesigned-input.tsx
```diff
@@ -1,57 +1,119 @@
...
+}
+
+const LabelComponent = ({ props, labelStyle, isFocused, theme }) => {
+  if (!props.label) return null
+  const styles = (): StyleProp<ViewStyle> => {
+    const labelStyles = {
+      marginLeft: isFocused ? 2 : 10,
+      marginBottom: 9,
+      fontWeight: "400",
+      fontSize: 16,
```
UncleSamtoshi, line 93: `fontWeight fontSize lineHeight` can be removed and replaced with the prop `type={'p1'}` on the Text component

app/components/atomic/galoy-input/galoy-redesigned-input.tsx
```diff
@@ -1,57 +1,119 @@
...
+        theme={theme}
+        errorStyles={remainingProps.errorStyle}
+      />
+    </View>
+  )
+}
+
+const LabelComponent = ({ props, labelStyle, isFocused, theme }) => {
+  if (!props.label) return null
+  const styles = (): StyleProp<ViewStyle> => {
```
UncleSamtoshi, line 88: We should use makeStyles here

app/components/atomic/galoy-input/galoy-redesigned-input.tsx
```diff
@@ -1,57 +1,119 @@
...
 
+const CaptionComponent = ({ props, errorStyles, isFocused, theme }) => {
+  if (!props.caption && !props.errorMessage) return null
+  const styles = (): StyleProp<ViewStyle> => {
+    const errorMessageStyles = {
+      color: props.caption ? theme.colors.grey5 : theme.colors.error5,
+      textTransform: "capitalize",
+      fontSize: 14,
+      marginTop: 9,
+      lineHeight: 24,
```
UncleSamtoshi, line 110: Here lineHeight and fontSize can be removed and replaced with `type={'p2'}`

app/components/atomic/galoy-input/galoy-redesigned-input.tsx
```diff
@@ -1,57 +1,119 @@
...
+      color: theme.colors.grey5,
+    }
+    return labelStyles
+  }
+  return <Text style={[styles(), labelStyle]}>{props.label}</Text>
+}
 
+const CaptionComponent = ({ props, errorStyles, isFocused, theme }) => {
+  if (!props.caption && !props.errorMessage) return null
+  const styles = (): StyleProp<ViewStyle> => {
```
UncleSamtoshi, line 104: makeStyles can be used here as well

UncleSamtoshi (changes requested):

UncleSamtoshi (approved):

PR #803, opened 2023-01-06
chore: build a daily version to testflight internal

nicolasburtey:
the idea is to always have main available on testflight on a daily basis to the team internally. we should all be onboarded into it so we can detect early regression that the tests would have not catched.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -24,6 +24,13 @@ source:
   repository: #@ release_pipeline_image()
 #@ end
 
+resources:
+- name: daily-release
+  type: cron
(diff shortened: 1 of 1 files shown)


sandipndev (approved):
By default when a release is cut it releases to testflight, and from there is a manual task to roll out to public, is that correct?

The Pipeline looks correct.

nicolasburtey:
That is correct. We can choose to do the process to make it public automatic also. Fastlane can do that. 

We should probably increase the automation to make a release process easier, but it should still be gated to make sure we don't release a version with known regression

Issue #799, opened 2023-01-05
Update text in the Set BBW Address (username) screen 


agbegin:
**Describe the bug**
When a user goes to set their BBW Address, the example text reads ____ @pay.galoy.io.  This should instead read @pay.bbw.sv. 

**To Reproduce**

1. Use a wallet that doesn't yet have a username
2. Click in settings > "Receive payments using your BBW..." > Set your BBW Address

**Expected behavior**
Placeholder text in the set BBW Address screen should read *@pay.bbw.sv*

**Screenshots**
<img width="254" alt="Screenshot 2023-01-05 at 5 09 28 PM" src="https://user-images.githubusercontent.com/90297599/210889661-0ef9e420-c66d-4d3e-a1ce-417c425d5d0f.png">

Extheoisah:
this issue has been addressed https://github.com/GaloyMoney/galoy-mobile/pull/805

PR #805, opened 2023-01-07
fix: fix username suffix based on app env

Extheoisah:
This should fix and close this issue https://github.com/GaloyMoney/galoy-mobile/issues/799

### DIFF
--- a/app/screens/galoy-address-screen/set-address-modal.tsx
+++ b/app/screens/galoy-address-screen/set-address-modal.tsx
@@ -162,10 +162,7 @@ export const SetAddressModal = ({ modalVisible, toggleModal }: SetAddressModalPr
     setAddress(value)
   }
 
-  const usernameSuffix =
-    appConfig.galoyInstance.name.toLowerCase() === "staging"
-      ? "@pay.staging.galoy.io"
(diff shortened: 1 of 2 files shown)


app/screens/galoy-address-screen/set-address-modal.tsx
```diff
@@ -159,6 +161,12 @@ export const SetAddressModal = ({ modalVisible, toggleModal }: SetAddressModalPr
     setError("")
     setAddress(value)
   }
+
+  const usernameSuffix =
+    appConfig.galoyInstance.name.toLowerCase() === "staging"
```
nicolasburtey, line 166: this value should be passed from the graphql layer. I think it's already available

nicolasburtey (dismissed):
domain name should not be hardcoded

app/screens/galoy-address-screen/set-address-modal.tsx
Extheoisah: fixed here https://github.com/GaloyMoney/galoy-mobile/pull/805/commits/128c613d54fe5a0cc604ec914f430f7fc44137b5

daviroo (approved):

PR #828, opened 2023-01-10
Test and prs workflows

daviroo:
FYI I've changed some settings in circleci.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -218,7 +218,7 @@ jobs:
           path: ios/output
 
 workflows:
-  mobile:
+  release:
     jobs:
       - test_ios:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -248,19 +248,13 @@ workflows:
...
             branches:
               ignore: /.*/
-  main-branch:
+  in-pr:
     jobs:
       - test_android:
           filters:
             branches:
-              only: main
+                ignore: main
```
nicolasburtey, line 256: we also want the tests to be run on `main`, to make sure we catch any merge regression

nicolasburtey (approved):

PR #836, opened 2023-01-10
chore: verify transactions are showing up

nicolasburtey:


nicolasburtey:
@daviroo not passing for Android

nicolasburtey:
@daviroo it doesn't work for android

Extheoisah:
> @daviroo it doesn't work for android

From my investigation, the issue seems to be because for there's a different accessibility ID for each list item when they're rendered on browser stack. And this is because the elements are only set in the DOM when we fetch the transaction list. 
I think I have a work around for this and I'm working on it.

nicolasburtey:
approved

Extheoisah (approved):

PR #840, opened 2023-01-11
chore: first pass importing GQL queries

nicolasburtey:


codegen.yml
```diff
@@ -0,0 +1,40 @@
...
+# schema: "https://raw.githubusercontent.com/GaloyMoney/galoy/main/src/graphql/main/schema.graphql"
+schema: "https://raw.githubusercontent.com/GaloyMoney/galoy/feb351eb62943bef80f8694868a0bf27596b4951/src/graphql/main/schema.graphql"
+documents: 
+  - "app/graphql/*/*.ts"
+generates:
+  app/graphql/generated.ts:
+    plugins:
+      - typescript
+      - typescript-operations
+      - typescript-react-apollo
```
nicolasburtey, line 11: @samerbuna fyi, by adding this directive, all the queries/mutations are automatically typed. ie: 

```
import { useCaptchaCreateChallengeMutation } from "@app/graphql/generated"
```

used here https://github.com/GaloyMoney/galoy-mobile/pull/840/files#diff-8b5390c865ce474872401927ca1d6026fb74fc6a313a02f6a5fedea92006e908R25

Makefile
```diff
@@ -0,0 +1,11 @@
+check-code:
```
daviroo, line 1: Why can't these commands be incorporated into the package.json file?

nicolasburtey: they already are. the idea is to have the same command in other repo, including backend, so that one don't have to remember if `make` or `yarn` has to be used

UncleSamtoshi (changes requested):

Makefile
```diff
@@ -0,0 +1,11 @@
...
+	yarn check-code
+
+unit:
+	yarn test
+
+test: unit check-code
+
+gen:
```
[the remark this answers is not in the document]
nicolasburtey: `codegen`? `gql-codegen`? 

I think as long as there is only one `codegen` command keeping it short is better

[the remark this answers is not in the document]
nicolasburtey: going for codegen for now

UncleSamtoshi (approved):

PR #843, opened 2023-01-11
chore: add langague change test

UncleSamtoshi:
- add test that switches from english to spanish to english
- fix device detection script in testing
- add the ability for single tests


.gitignore
```diff
@@ -73,3 +73,5 @@ ios/GoogleService-Info.plist
 yalc.lock
 
 *.log
+
+.env
```
daviroo, line 77: Why do we need this?

daviroo (dismissed):

e2e/02-login-flow.e2e.spec.ts
```diff
@@ -32,7 +32,7 @@ describe("Login Flow", async () => {
     await browser.pause(2000)
     await scrollDown()
 
-    await browser.pause(2000)
+    await browser.pause(1000)
```
nicolasburtey, line 35: probably because we don't have yet some check on some screen, having some wait can be useful.

video from https://app.circleci.com/pipelines/github/GaloyMoney/galoy-mobile/2265/workflows/974b0d49-a2ca-4290-b457-65c629a28e49/jobs/1563 shows the issue: at the beginning we don't see if the home page correctly load

e2e/02-login-flow.e2e.spec.ts
```diff
@@ -72,31 +72,42 @@ describe("Login Flow", async () => {
...
+    await accountButton.click()
+    const logoutButton = await $(selector(LL.common.logout(), "StaticText"))
+    await logoutButton.waitForDisplayed({ timeout })
+    expect(logoutButton.isDisplayed()).toBeTruthy()
+    const backButton = await $(goBack())
+    await backButton.waitForDisplayed({ timeout })
+    await backButton.click()
     await browser.pause(1000)
+
+    // let loginSelector =
```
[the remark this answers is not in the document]
UncleSamtoshi: Yup good catch.

e2e/02-login-flow.e2e.spec.ts
UncleSamtoshi: I believe it will wait up to 30 seconds in many places until the desired component to press is displayed.

nicolasburtey: > Are you suggesting I increase the pauses back to their original value?

not too sure yet. just thinking out loud as I'm learning the e2e flow and seeing the pro/cons or longer wait time

ntheile: i found 2 seconds to be the most consistent wait time between screen transitions/animations. in theory waitForDisplayed should wait for the element, but in practice it does not always work. Not sure it was just related to v1 of appium, maybe they cleaned up the flakes in v2

PR #846, opened 2023-01-11
chore: moving local types into codegen

nicolasburtey:


Extheoisah (approved):
Looks okay to me. Except the test_android is failing

PR #858, opened 2023-01-12
chore: add strict null checks

UncleSamtoshi:


UncleSamtoshi:
Going to write some observations as I come across them in the application

UncleSamtoshi:
- Current ScreenType and NavigationType are masking types on a lot of components, removing these types should occur although it will increase the scope of this change

UncleSamtoshi:
Is there a way to type local graphql queries so that they are always defined?

nicolasburtey (approved):

PR #868, opened 2023-01-13
chore: removing dependance on mainQuery from conversion screen

nicolasburtey:
also removing some duplicated states

### DIFF
--- a/app/graphql/cache.ts
+++ b/app/graphql/cache.ts
@@ -1,11 +1,8 @@
-import { ApolloClient, InMemoryCache, gql } from "@apollo/client"
+import { InMemoryCache, gql } from "@apollo/client"
 import { WalletCurrency } from "./generated"
 
-// FIXME is there a way to avoid the circular dependency cache/client?
 export const createCache = () => {
-  let client: ApolloClient<unknown> | undefined
-
-  const cache = new InMemoryCache({
+  return new InMemoryCache({
     typePolicies: {
       Contact: {
         fields: {
@@ -37,12 +34,12 @@ export const createCache = () => {
       ConsumerAccount: {
         fields: {
           usdWallet: {
-            read: (_, { readField }) => {
+            read: (_, { readField, cache }) => {
               const wallets = readField("wallets")
               if (Array.isArray(wallets)) {
                 for (const w of wallets) {
                   try {
-                    const wallet = client?.readFragment({
+                    const wallet: Wallet = cache.readFragment({
                       id: w.__ref,
                       fragment: gql`
                         fragment MyWallet on UsdWallet {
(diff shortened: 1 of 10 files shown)


app/graphql/generated.ts
```diff
@@ -61,12 +61,14 @@ export type Scalars = {
 }
 
 export type Account = {
+  readonly btcWallet?: Maybe<Wallet>
```
bodymindarts, line 64: isn't `?` redundant when using `Maybe` type?

nicolasburtey: Don't know for sure. But it's a generated files

app/graphql/cache.ts
```diff
@@ -29,5 +30,33 @@ export const cache = new InMemoryCache({
         },
       },
     },
+    ConsumerAccount: {
```
daviroo, line 33: When I attempted this last time I wrote a script to generate the possible types from the schema on the server https://github.com/GaloyMoney/galoy-mobile/compare/main...refactor-transactions-screen#diff-d3e97a63b7fd33227d3577ef3a18ec8793bf860730677e9f7f3382be13d0cc8f.

I was then able to pass in the generated file to the possibleTypes config - https://github.com/GaloyMoney/galoy-mobile/compare/main...refactor-transactions-screen#diff-05ff1f30a72721bd2ff7481a4e51b12592d8d7ed89c5ac459c25c9c9d2505314

daviroo (changes requested):

daviroo:
Since we have polymorphic relationships in our graphql schema I think we need to give the client a possible types map (I left a comment in the code as well).  There are more details in the docs https://www.apollographql.com/docs/react/data/fragments/#using-fragments-with-unions-and-interfaces

daviroo (approved):

app/graphql/cache.ts
```diff
@@ -1,33 +1,104 @@
...
+            read: (value) => value ?? false,
+          },
         },
-        hiddenBalanceToolTip: {
-          read: (value) => value ?? false,
+      },
+      ConsumerAccount: {
+        fields: {
+          usdWallet: {
+            read: (_, { readField }) => {
```
UncleSamtoshi, line 40:
```suggestion
            read: (_, { readField, cache }) => {
```

app/graphql/cache.ts
```diff
@@ -1,33 +1,104 @@
...
+              const wallets = readField("wallets")
+              if (Array.isArray(wallets)) {
+                for (const w of wallets) {
+                  try {
+                    const wallet = client?.readFragment({
```
UncleSamtoshi, line 45:
```suggestion
                    const wallet = cache.readFragment({
```

UncleSamtoshi (changes requested):
You can use the `cache` argument instead of having to set the client.

app/graphql/cache.ts
nicolasburtey: oh awesome. thanks! I'll refactor

UncleSamtoshi (approved):

PR #869, opened 2023-01-13
chore: Localized gql

nicolasburtey:


app/screens/galoy-address-screen/set-default-wallet.tsx
```diff
@@ -41,9 +44,47 @@ const styles = EStyleSheet.create({
...
+        __typename
+        id
+        defaultWalletId
+      }
+    }
+  }
+
+  query setDefaultWallet {
+    me {
+      defaultAccount {
```
Extheoisah, line 64: Since we query for `defaultAccount` in several places (address-screen, receive-etc, set-default-wallet, etc), would it be helpful to create a fragment of it and perhaps that'll make it easier to get wallets data instead of always writing out the query for each screen that needs it🤔. What do you think?

nicolasburtey: > Since we query for `defaultAccount` in several places (address-screen, receive-btc, set-default-wallet, etc), would it be helpful to create a fragment of it and perhaps that'll make it easier to get wallets data

so far the queries are pretty different, so I'm not sure there would be a need to do that. even if we need defaultAccount for many queries, most of the time we need different field under it.

I have not started with the easier screen/flow, ie the non send/receive payment screen, so maybe there would be more duplication there

> instead of always writing out the query for each screen that needs it🤔. What do you think?

I think we should 
1/ finish the refactoring and remove the dependency on useMainQuery for all screens 

2/ try to integrate storybooks for building screen and mocking screens / gql queries (as opposed to just using story book for components) and see of that works. then we can reassess refactoring with fragment once we've more understand on how to mock things also, because this is an additional element to consider. but because starting on 2/ I'd like to complete 1.

UncleSamtoshi (approved):

PR #874, opened 2023-01-14
chore: align more styles from backend and frontend

nicolasburtey:
run prettier to eslint to make check faster


Extheoisah (approved):
I'll approve but we might need to upgrade the typescript version because this warning shows up after running `yarn check-code`

`WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.`

PR #889, opened 2023-01-16
chore: add e2e test for username payment

Extheoisah:
This PR adds e2e test for payment to username. Temporarily, it uses `extheo` as the username it pays to. The goal is to create another test account on staging for this test.


e2e/04-username-payments-flow.e2e.spec.ts
```diff
@@ -0,0 +1,79 @@
+import { i18nObject } from "../app/i18n/i18n-util"
+import { loadLocale } from "../app/i18n/i18n-util.sync"
+import { selector, enter } from "./utils"
+import { getInvoice } from "./utils/graphql"
+
+describe("Username Payment Flow", async () => {
+  loadLocale("en")
+  const LL = i18nObject("en")
+  const timeout = 30000
+  const username: string = "extheo"
```
Extheoisah, line 10: @nicolasburtey I couldn't create another test account because I don't have a phone number to use for it. How I can I go about this?
But temporarily this works.

Extheoisah (commented):
If we create another test account on staging, I can test USD invoice generation and payment which we currently do not do (just BTC invoice we test for now). We can set the new test account default wallet to USD then make payment to it. Then we can generate a USD invoice on the test1 account and pay it using the new test account.

e2e/04-username-payments-flow.e2e.spec.ts
nicolasburtey: > @nicolasburtey I couldn't create another test account because I don't have a phone number to use for it. How I can I go about this? But temporarily this works.

you can sync with anyone on the backend/infra they can get you onboarded

nicolasburtey (approved):

PR #903, opened 2023-01-17
chore: continue null checks

UncleSamtoshi:


app/components/price-graph/price-graph.tsx
```diff
@@ -101,7 +101,9 @@ export const PriceGraphDataInjected = () => {
 
   return (
     <PriceGraph
-      prices={data.btcPriceList.slice()} // FIXME: PricePoint should be non-nullable
+      prices={data.btcPriceList
+        .filter((price) => price !== null)
+        .map((price) => price as PricePoint)} // FIXME: PricePoint should be non-nullable
```
daviroo, line 106: I think you have fixed the FIXME :)

app/screens/conversion-flow/conversion-confirmation-screen.tsx
```diff
@@ -118,14 +123,18 @@ export const ConversionConfirmationScreen = ({
           },
         })
 
-        const status = data.intraLedgerPaymentSend.status
+        const status = data?.intraLedgerPaymentSend.status
+
+        if (!status) {
+          throw new Error("Conversion failed")
```
daviroo, line 129: Is this caught somewhere or does it show the error screen?

app/screens/receive-bitcoin-screen/receive-btc.tsx
```diff
@@ -245,46 +245,56 @@ const ReceiveBtc = () => {
...
-            data: {
-              lnNoAmountInvoiceCreate: { invoice, errors },
-            },
-          } = await lnNoAmountInvoiceCreate({
+          const { data } = await lnNoAmountInvoiceCreate({
             variables: { input: { walletId, memo } },
           })
+
+          if (!data) {
+            throw new Error("No data returned from lnNoAmountInvoiceCreate")
```
daviroo, line 253: Same question here - is this caught or does it bubble up to the error screen?

app/screens/receive-bitcoin-screen/receive-usd.tsx
```diff
@@ -343,6 +354,8 @@ const ReceiveUsd = () => {
   }, [invoice?.paymentRequest])
 
   const copyToClipboard = () => {
+    if (!invoice?.paymentRequest) return
```
daviroo, line 357: Maybe show an error toast here?

daviroo (changes requested):

app/components/price-graph/price-graph.tsx
UncleSamtoshi: Maybe I could be more clear, I think the backend should make it non nullable, because right now we receive an array will some of the values possibly being null

app/screens/conversion-flow/conversion-confirmation-screen.tsx
UncleSamtoshi: Caught in line 171

app/screens/receive-bitcoin-screen/receive-btc.tsx
UncleSamtoshi: Caught in 294

app/components/price-graph/price-graph.tsx
nicolasburtey: yup would be relevant to precise the backend should be updated here. 

a nullable to non nullable is not a breaking change afair so should be possible easily

app/screens/receive-bitcoin-screen/receive-usd.tsx
UncleSamtoshi: Yeah I was debating what to do here, lets discuss it in mobile sync. It should technically not happen, but we dont have the type safety.

nicolasburtey: maybe add a `// TODO` so someone eventually looks at it?

app/screens/receive-bitcoin-screen/receive-btc.tsx
daviroo: Right enough.  Should have looked at the whole file I thought this was part of the render method 😅

app/screens/receive-bitcoin-screen/receive-usd.tsx
UncleSamtoshi: I think the correct behavior is to make a function null if the function is not able to perform its core tasks because of null | undefined dependencies. This way we dont have functions floating around that don't do what their name says they do.

UncleSamtoshi: @nb1 @daviroo thoughts on this? I just pushed the new changes

app/screens/conversion-flow/conversion-confirmation-screen.tsx
```diff
@@ -57,8 +57,13 @@ export const ConversionConfirmationScreen = ({
 
   const { data } = useConversionScreenQuery({ fetchPolicy: "cache-only" })
 
-  const usdWallet = data.me.defaultAccount.usdWallet
-  const btcWallet = data.me.defaultAccount.btcWallet
+  const usdWallet = data?.me?.defaultAccount.usdWallet
+  const btcWallet = data?.me?.defaultAccount.btcWallet
+
+  if (!data?.me || !usdWallet || !btcWallet) {
+    // TODO: handle errors and or provide some loading state
```
nicolasburtey, line 64: note: because `fetchPolicy` is `cache-only`, I don't think there is any loading to consider.

nicolasburtey (approved):

app/screens/conversion-flow/conversion-confirmation-screen.tsx
UncleSamtoshi: Gotcha, from the perspective of types I dont think that is represented

PR #907, opened 2023-01-18
chore: more null checks

UncleSamtoshi:
- refactor useFee


nicolasburtey:
> * refactor useFee

I've tried to untangled the current type error that made the build fail but I think there could be quite a bit of refactoring to do.

first PaymentType now refers to 2 differents things:
`PaymentType` from graphql codegen
`PaymentType` from galoy-client

those things are confusing so we should rename one of them

I also think we should stop using `enum` from js, and rely on `const object` instead. there is less type quirks IMO with plan js const object. declared with: 

```
const obj = {
   LnUrl: "lnurl"
  Onchain: "onchain"
} as const
```

the key take away from the code below is the `as const` at the end. otherwise typescript would not infer properly that the set of keys are fixed

overall, I think we could extend the `graphql.PaymentType` with the additional "unified" and other property. The only thing I'm not sure how to really handle now is the fact that we're have the galoy-client library, but this library doesn't rely on codegen. so we can't really extend the `PaymentType` that graphql give us and have more internal only mobile type. 

anyway, just some food for thought

PR #914, opened 2023-01-18
chore: set proper fetchPolicy type

nicolasburtey:


daviroo (approved):

nicolasburtey:
the tests are passing but they should not pass. there is an issue both with the test (@Extheoisah: test should fail here because the user already exist) and from the code

Extheoisah:
> the tests are passing but they should not pass. there is an issue both with the test (@Extheoisah: test should fail here because the user already exist) and from the code

You're correct. I'll fix it. We should check if the username exists in the contacts. If it does exist, no need to check for the modal in the test but if not, we should check for the modal.

PR #916, opened 2023-01-18
chore: refactor notification with a wrapper

nicolasburtey:


app/navigation/root-navigator.tsx
```diff
@@ -271,18 +270,16 @@ export const RootStack: NavigatorType = () => {
       })
   }, [])
 
+  useAddDeviceToken()
```
nicolasburtey, line 273: not needed?

app/navigation/root-navigator.tsx
```diff
@@ -271,18 +270,16 @@ export const RootStack: NavigatorType = () => {
...
     const setupNotifications = async () => {
-      if (hasToken && client) {
-        const hasPermission = await hasNotificationPermission()
-        if (hasPermission) {
-          addDeviceToken(client)
-          messaging().onTokenRefresh(() => addDeviceToken(client))
-        }
+      const hasPermission = await hasNotificationPermission()
+      if (hasPermission) {
+        messaging().onTokenRefresh(useAddDeviceToken)
```
UncleSamtoshi, line 278: Hooks need to run on every render, I dont believe they can be used in a useEffect which happens outside of a render cycle and only when its dependencies change.

app/navigation/root-navigator.tsx
UncleSamtoshi, line 273: I believe this will call addDeviceToken on every render.

UncleSamtoshi (changes requested):

daviroo:
@nicolasburtey this needs rebased

PR #930, opened 2023-01-20
chore: enable strict null checks

UncleSamtoshi:
- enables strict null checks
- disables ts checking on send flow files that are being refactored


Extheoisah (commented):
I think you need to rebase on main for the tests to pass.

nicolasburtey (approved):

PR #935, opened 2023-01-21
chore: include ids in queries

nicolasburtey:
this should help with caching

I don't know what is the effect of object which don't have an id, like:
```
globals {
  network
}
me {
...
}
```

does the globals request force a network request because `globals.networks` is not cache properly because there is no `globals.id` ? I will dive in this later.

### DIFF
--- a/app/screens/map-screen/map-screen.tsx
+++ b/app/screens/map-screen/map-screen.tsx
@@ -41,6 +41,7 @@ type Props = {
 }
 
 gql`
+  # TODO: caching
   query businessMapMarkers {
     businessMapMarkers {
       username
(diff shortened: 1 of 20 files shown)


UncleSamtoshi (approved):

app/screens/map-screen/map-screen.tsx
```diff
@@ -41,6 +41,7 @@ type Props = {
 }
 
 gql`
+  # TODO: caching
```
Extheoisah, line 44: We can try defining a type policy for this in the cache using the coordinates as a key identifier since they're unique.
Something like
`typePolicies: {
   (typeof buniessMapMarkers): {
           keyFields: [ "coordinates", ["longitude", "latitude"] ]
       }
}`

PR #936, opened 2023-01-21
chore: remove price context

nicolasburtey:
- setting cache policy to `[]` for price to behave as a singleton
- the price context is very convoluted, probably because of cache/client issue. I think now that is fixed we don't need any of the related code.
- I think `useContext` really only make sense for non Apollo related data. translation is a good example. displayCurrency is probably not a good one @daviroo


app/hooks/use-price-conversion.ts
```diff
@@ -1,10 +1,21 @@
...
 export const usePriceConversion = () => {
-  const priceData = usePriceContext()
+  const { data } = useCurrentPriceQuery({
+    query: CurrentPriceDocument,
+    fetchPolicy: "cache-only",
+  })
+
+  const price = Number(data?.btcPrice?.formattedAmount ?? NaN)
+
+  console.log("priceData", { price })
```
UncleSamtoshi, line 17: Console.log here that needs to be removed

UncleSamtoshi (changes requested):

app/hooks/use-price-conversion.ts
```diff
@@ -1,10 +1,21 @@
...
+  useCurrentPriceQuery,
+} from "@app/graphql/generated"
 import { PaymentAmount, UsdPaymentAmount } from "@app/types/amounts"
 import * as React from "react"
 
 export const usePriceConversion = () => {
-  const priceData = usePriceContext()
+  const { data } = useCurrentPriceQuery({
+    query: CurrentPriceDocument,
+    fetchPolicy: "cache-only",
```
Extheoisah, line 12: Apollo recommends using `fetchPolicy: cache-and-network` for data that frequently updates. Maybe we can try that. https://www.apollographql.com/docs/react/data/queries/#cache-and-network

nicolasburtey: we already get the update from the subscription.

PR #937, opened 2023-01-22
chore: add contact validation

Extheoisah:


nicolasburtey (approved):

e2e/03-validate-username-flow.e2e.spec.ts
```diff
@@ -41,10 +42,14 @@ describe("Validate Username Flow", async () => {
     const confirmButton = await $(
       selector(LL.SendBitcoinDestinationScreen.confirmModal.confirmButton(), "Button"),
     )
-    await checkBoxButton.waitForDisplayed({ timeout })
-    await checkBoxButton.click()
-    await confirmButton.isEnabled()
-    await confirmButton.click()
+    const isContactAvailable = await checkContact(username)
```
nicolasburtey, line 45: approved and deployed to fix the build. but the PR could be improved:

name can be improved. I believe this is not meant for `isContactAvailable` but `isContactPresent`.

similarly, the mutation checkContact is not very descriptive of the goal of the mutation.

Extheoisah: I will resolve this in a follow-up PR

PR #938, opened 2023-01-22
chore: merging global error toast with client lib + some refactoring

nicolasburtey:


app/graphql/client.tsx
```diff
@@ -102,7 +129,70 @@ const GaloyClient: React.FC<{ children: React.ReactNode }> = ({ children }) => {
         }),
       )
 
-      const splitLink = split(
+      const errorLink = onError(({ graphQLErrors, networkError: networkErrorRaw }) => {
```
UncleSamtoshi, line 132: Has this been tested at all? My concern is that since the error link is executed for each query, if we make multiple queries at the same time, we will see multiple instances of the same toast issue. In the past, it seems the checks for network errors were psuedo-batched by being processed all at once.

UncleSamtoshi (changes requested):

app/graphql/client.tsx
nicolasburtey: I need to make some testing around this. 

Where is batching done by the previous code?

UncleSamtoshi: Previously the check for errors only ran once after every render with the `useEffect`, so if multiple errors came back in between renders it would only run once.

UncleSamtoshi: I'm not sure how common that is however. Also not sure if the retry logic here will trigger the error link foreach retry or just once.

nicolasburtey: so I've tested it locally.

basically from my test: the old code didn't work anymore, or at least even if the server was stopped, there was no error showing up.

with the new code, the toast is showing up. It' not reappearing in an annoying way. if you move screen, it may reappear if you deeper in the flow and are loader new graphql queries that are not watched yet. if you go "back there" is no error message.

UncleSamtoshi: Where does `setNetworkError` get called to clear the error after a toast has been shown?

app/graphql/client.tsx
```diff
@@ -316,6 +333,67 @@ const MyUpdateSub = ({ children }: PropsWithChildren) => {
   return <LnUpdateHashPaidProvider value={lastHash}>{children}</LnUpdateHashPaidProvider>
 }
 
+const NetworkErrorToast: React.FC<{ networkError: ServerError | undefined }> = ({
```
UncleSamtoshi, line 336: Is there a reason to include this component in this file? If not I think it would be cleaner to split it out.

UncleSamtoshi (changes requested):

app/graphql/client.tsx
```diff
@@ -316,6 +333,67 @@ const MyUpdateSub = ({ children }: PropsWithChildren) => {
...
+    }
+
+    if (networkError.message === "Network request failed") {
+      // TODO translation
+      toastShow({
+        message: (translations) => translations.errors.network.connection(),
+        currentTranslation: LL,
+      })
+    }
+  }, [networkError, LL, logout])
```
UncleSamtoshi, line 392: Where does setNetworkError get called to clear the error after a toast has been shown?

UncleSamtoshi (changes requested):

app/graphql/client.tsx
nicolasburtey: no particular reason. I have just put all the Apollo only thing into the client for code colocation convenience. but can move into a dedicated component.

app/graphql/client.tsx
nicolasburtey: why would we need to clear it? 

if there are no error (ie: at startup), this rountine exit early
if there is a new instance of the network, then it display an error message via toast. it won't re-enter the function until a new error happens, at which point it would display a new error message via toast.

app/graphql/client.tsx
nicolasburtey: this has been update. could make sense to split up client more in different individual component but can do that in another PR

app/graphql/client.tsx
UncleSamtoshi: Makes sense

UncleSamtoshi (approved):

PR #939, opened 2023-01-23
chore: rewrite send logic

UncleSamtoshi:
- introduce payment details type
- create functions for building separate payment details
- extract sendPayment logic into useSendPayment hook
- rewrite useFee hook


app/screens/send-bitcoin-screen/use-send-payment.ts
```diff
@@ -0,0 +1,67 @@
...
+  const [lnNoAmountUsdInvoicePaymentSend] = useLnNoAmountUsdInvoicePaymentSendMutation()
+  const [onChainPaymentSend] = useOnChainPaymentSendMutation()
+  const [loading, setLoading] = useState(false)
+
+  const sendPayment = useMemo(() => {
+    return (
+      sendPaymentFn &&
+      (async () => {
+        setLoading(true)
+        const response = await sendPaymentFn({
```
nicolasburtey, line 40: doesn't `sendPaymentFn` already have `loading`?  this might be better than managing the state explicitly. ie if throw for whatever reason, I guess loading would be correctly update if fetch from mutation, but not as it's currently implemented.

app/screens/send-bitcoin-screen/use-fee.ts
```diff
@@ -13,20 +12,21 @@ import {
...
-  address: string
-  isNoAmountInvoice: boolean
-  invoice: string
-  paymentType: string
-  paymentAmount: PaymentAmount<WalletCurrency>
-}
+type FeeType =
+  | {
+      status: "loading" | "error" | "unset"
+      amount?: undefined | null
```
nicolasburtey, line 20: no need for `amount`?

app/screens/send-bitcoin-screen/payment-details/index.types.ts
```diff
@@ -0,0 +1,116 @@
...
+import { LnUrlPayServiceResponse } from "lnurl-pay/dist/types/types"
+
+export type ConvertPaymentAmount = <T extends WalletCurrency>(
+  paymentAmount: PaymentAmount<WalletCurrency>,
+  toCurrency: T,
+) => PaymentAmount<T>
+
+export type BaseCreatePaymentDetailsParams<T extends WalletCurrency> = {
+  convertPaymentAmount: ConvertPaymentAmount
+  sendingWalletDescriptor?: WalletDescriptor<T>
```
nicolasburtey, line 28: in which scenario is `sendingWalletDescriptor` optional?

app/screens/send-bitcoin-screen/payment-details/index.types.ts
```diff
@@ -0,0 +1,116 @@
...
+  onChainTxFee: ReturnType<typeof useOnChainTxFeeLazyQuery>["0"]
+}
+
+export type GetFee<T extends WalletCurrency> = (getFeeFns: GetFeeParams) => Promise<{
+  amount?: PaymentAmount<T> | null | undefined
+  errors?: readonly GraphQlApplicationError[]
+}>
+
+type SendPaymentParams = {
+  lnInvoicePaymentSend: ReturnType<typeof useLnInvoicePaymentSendMutation>["0"]
```
nicolasburtey, line 57: can't you use this type directly? `LnInvoicePaymentSendMutationResult` or `LnInvoicePaymentSendMutationHookResult` (same for all the other types here)

app/screens/send-bitcoin-screen/payment-details/intraledger-payment-details.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  const memo = senderSpecifiedMemo
+  const settlementAmount =
+    unitOfAccountAmount &&
+    sendingWalletDescriptor &&
+    convertPaymentAmount(unitOfAccountAmount, sendingWalletDescriptor.currency)
+
+  const getFee: GetFee<T> | undefined =
+    sendingWalletDescriptor &&
+    (async (_) => {
```
nicolasburtey, line 41: async needed?

app/screens/send-bitcoin-screen/payment-details/onchain-payment-details.ts
```diff
@@ -0,0 +1,291 @@
...
+    destinationSpecifiedAmount,
+    convertPaymentAmount,
+    sendingWalletDescriptor,
+    destinationSpecifiedMemo,
+    senderSpecifiedMemo,
+    address,
+  } = params
+
+  if (sendingWalletDescriptor && sendingWalletDescriptor.currency !== "BTC") {
+    throw new Error("Onchain payments are only supported for BTC wallets")
```
nicolasburtey, line 176: it's no longer the case. we could use this PR to add onchain USD payment as well?

nicolasburtey (commented):
some comments skimming through the PR. I didn't really dive into it. 

we haven't made any or much test on hooks until now, but maybe this could be a good place to start?

app/screens/send-bitcoin-screen/payment-details/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      if (validPaymentDestination.amount) {
+        return CreateAmountOnchainPaymentDetails({
+          address: validPaymentDestination.address,
+          destinationSpecifiedAmount: {
+            amount: validPaymentDestination.amount,
+            currency: "BTC",
+          },
+          convertPaymentAmount,
+          destinationSpecifiedMemo: validPaymentDestination.memo,
+          unitOfAccount: "USD",
```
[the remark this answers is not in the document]
UncleSamtoshi: They create the same type (PaymentDetails which has a unitOfAccountAmount) but they are constructor functions that take in different arguments because an invoice with an amount is not capable of specifying the desired unitOfAccountAmount, the amount comes from the invoice amount and then it is converted to the desired unit of account. Whereas an invoice without an amount will determine its settlement amount from the desired unit of account amount entered.

app/screens/send-bitcoin-screen/payment-details/index.types.ts
UncleSamtoshi: UseFee and SendPayment will not be defined untill sendingWalletDescriptor is defined, however you could imagine a scenario where you would create the PaymentDetail, then set the amount, then set the sendingWalletDescriptor.

app/screens/send-bitcoin-screen/payment-details/index.types.ts
UncleSamtoshi: You could, although we just need the function, so it is not necessary to depend on the entire return result. It also makes mocking a little simpler for the test by just having to mock the function.

app/screens/send-bitcoin-screen/use-send-payment.ts
UncleSamtoshi: > some comments skimming through the PR. I didn't really dive into it.
> 
> we haven't made any or much test on hooks until now, but maybe this could be a good place to start?

Yeah if the team feels that this is a good approach for the refactor, it will be really easy to unit test,  and I plan on writing them.

UncleSamtoshi:
> 

Yeah if the team feels that this is a good approach for the refactor, it will be really easy to unit test, and I plan on writing them.

app/screens/send-bitcoin-screen/payment-details/onchain-payment-details.ts
UncleSamtoshi: Yeah we can do that

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -262,301 +251,110 @@ const SendBitcoinConfirmationScreen = ({
...
-      }
+        if (status === "ALREADY_PAID") {
+          setPaymentError("Invoice is already paid")
+          return
+        }
 
-      if (status === "ALREADY_PAID") {
-        setPaymentError("Invoice is already paid")
-        return
+        setPaymentError(errorsMessage || "Something went wrong")
```
nicolasburtey, line 318: string need translation

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -239,60 +235,73 @@ const SendBitcoinDetailsScreen = ({
     returnPartialData: true,
   })
 
-  const defaultWallet = data?.me?.defaultAccount?.defaultWallet
+  const defaultWallet = data?.me?.defaultAccount?.defaultWallet // FIXME: Backend should not allow default wallet to be nullable
```
nicolasburtey, line 238: defaultWallet is a @client property. it doesn't exist on the backend.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -239,60 +235,73 @@ const SendBitcoinDetailsScreen = ({
...
   const usdWalletId = data?.me?.defaultAccount?.usdWallet?.id
-  const btcWallet = data?.me?.defaultAccount?.btcWallet
+  const btcWallet = data?.me?.defaultAccount?.btcWallet // FIXME: Backend should not allow btc wallet to be nullable
```
nicolasburtey, line 240: btcWallet is a @client property. it doesn't exist on the backend.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -538,26 +600,28 @@ const SendBitcoinDetailsScreen = ({
...
+                      )}
                       onChangeValue={setAmountsWithBtc}
                       prefix=""
                       delimiter=","
                       separator="."
                       precision={0}
                       suffix=" sats"
                       minValue={0}
-                      editable={!isFixedAmountInvoice}
+                      editable={Boolean(paymentDetail.setAmount)}
```
nicolasburtey, line 618: this line feels weird. setAmount is a function and it's been converted to a boolean?

btw, I'm not a huge fan of a the Boolean requirement, maybe we could drop the eslint rules? (I guess here it's triggering a relevant question)

app/screens/send-bitcoin-screen/use-fee.ts
UncleSamtoshi: If you want to be able to access `fee.amount` without going `if (amount in fee) fee.amount`, you have to set the property type as being undefined.

app/screens/send-bitcoin-screen/payment-details/onchain-payment-details.ts
UncleSamtoshi: I will hold off on USD for now, as there is work to be done on the backend to properly support amount usd invoices.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
UncleSamtoshi: Replacing it with `paymentDetail.canSetAmount`

app/screens/send-bitcoin-screen/use-send-payment.ts
UncleSamtoshi: Switched to using the `loading` from the mutation

daviroo (approved):

Extheoisah:
Just one comment. I think the convention is to use double quotes and not single quotes? @UncleSamtoshi

UncleSamtoshi:
Switched to using double quotes!

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
UncleSamtoshi: This PR doesnt change the UI, it only refactors the logic for constructing and sending the payments. I think it makes sense to address the untranslated strings when we revisit the UI aspect of this flow in the redesigned screens.

Extheoisah (approved):

PR #944, opened 2023-01-24
chore: fix price changing bug

UncleSamtoshi:
- remove price cache
- update root query cache on price from subscription


nicolasburtey (approved):

nicolasburtey:
I believe there can be a better design with better graphql types but in the meantime let's merge this for now @UncleSamtoshi

PR #945, opened 2023-01-24
chore: fix authentication flow

UncleSamtoshi:
- fixes authentication by reverting changes to phone auth in PR #922


nicolasburtey:
What is the issue? Is there a way to do a fix without relying on onComplete?

onComplete make the code more complex versus a async flow because it put the logic in different places

Also we use async typically so the callback model with onComplete is not really the pattern we have in our code generally

UncleSamtoshi:
The function using `userLogin` expects a return result of the token but `userLogin` doensnt return anything and doesnt save the token. Additonaly, you didn't introduce this bug, but it is strange behavior to navigate to a new screen before the token has been saved. I can refactor and repush.

PR #953, opened 2023-01-25
chore: strict true for tsconfig

nicolasburtey:


app/screens/receive-bitcoin-screen/receive-btc.tsx
```diff
@@ -325,10 +325,12 @@ const ReceiveBtc = () => {
...
         }
-      } catch (err) {
-        console.error(err, "error with AddInvoice")
-        crashlytics().recordError(err)
-        setErr(`${err}`)
+      } catch (err: unknown) {
+        if (err instanceof Error) {
+          console.error(err, "error with AddInvoice")
+          crashlytics().recordError(err)
+          setErr(`${err}`)
```
UncleSamtoshi, line 332: setErr should probably be outside of the error check, so that we can still show a string to the screen

UncleSamtoshi: There are several other places where some methods may have been unintentionally included in the conditional blocks.

nicolasburtey: I think a PR reviewing comprehensibly the error management is probably a good way to think about it

nicolasburtey: Typescript return either `any` or `unknown` in the catch block but I think it should typically (always?) return an Error object?

app/i18n/formatters.ts
```diff
@@ -6,11 +6,14 @@ export const initFormatters: FormattersInitializer<Locales, Formatters> = (
   locale: Locales,
 ) => {
   const formatters: Formatters = {
-    sats: (value: number): string => {
+    sats: (value: unknown): unknown => {
```
daviroo, line 9: Why is this unknown?

app/screens/settings-screen/types.d.ts
```diff
@@ -1,16 +1,13 @@
 type SettingsScreenProps = {
   hasToken: boolean
   navigation: StackNavigationProp<RootStackParamList, "settings">
-  username: string
-  phone: string
+  username: string | unudefined
```
daviroo, line 4:
```suggestion
  username: string | undefined
```

daviroo (changes requested):

app/i18n/formatters.ts
nicolasburtey: this is what the library i18n expects AFAIU

daviroo (approved):

app/screens/move-money-screen/move-money-screen.tsx
```diff
@@ -225,16 +224,20 @@ export const MoveMoneyScreenDataInjected: ScreenType = ({
   })
 
   const mobileVersions = data?.mobileVersions ? data.mobileVersions[0] : undefined // FIXME array/item mismatch
-  const mergedTransactions = data?.me?.defaultAccount?.transactions?.edges
+  const transactionsEdges = data?.me?.defaultAccount?.transactions?.edges ?? undefined
   const usdWalletId = data?.me?.defaultAccount?.usdWallet?.id
 
   const btcWalletValueInUsd = hasToken
-    ? data?.me?.defaultAccount?.btcWallet?.usdBalance
+    ? data?.me?.defaultAccount?.btcWallet?.usdBalance ?? NaN
```
Extheoisah, line 231: I think this code can be simplified.
In a sense adding `: 0` does not change the final value of `btcWalletValueInUsd` when `hasToken` is false because it'll always fallback to `NaN` not 0. we could just write it as:
```js
const btcWalletValueInUsd = hasToken ? data?.me?.defaultAccount?.btcWallet?.usdBalance : NaN

The same applies for `usdWalletBalance` and `btcWalletBalance`

nicolasburtey: if `hasToken` is `true` but the value is undefined (should not be, but is required for type) then the value will be `NaN`

nicolasburtey: if `hasToken` is `false`, then the value will be `0`

PR #962, opened 2023-01-27
chore: Remove pauses from login test

daviroo:


app/config/galoy-instances.ts
```diff
@@ -6,9 +6,11 @@ const scriptHostname = (): string => {
   return scriptHostname
 }
 
-export type GaloyInstanceNames = "BBW" | "Staging" | "Local" | "Custom"
+export const possibleGaloyInstanceNames = ["BBW", "Staging", "Local", "Custom"]
```
nicolasburtey, line 9:
```suggestion
export const possibleGaloyInstanceNames = ["BBW", "Staging", "Local", "Custom"] as const
```

nicolasburtey: type won't be narrowed as properly if you don't add the `as const`.  it would be a string instead of the individual values

e2e/02-login-flow.e2e.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-await-in-loop */
```
nicolasburtey, line 1: let's remove this rule instead?

nicolasburtey (commented):
some comments

e2e/config/browserstack.conf.js
```diff
@@ -9,7 +9,6 @@ let capabilities = {
   "device": "Google Pixel 3",
   "os_version": "9.0",
   "app": process.env.BROWSERSTACK_APP_ID,
-  "browserstack.local": true,
```
nicolasburtey, line 12: what does this do?

nicolasburtey (approved):

e2e/config/browserstack.conf.js
daviroo: It allows you to connect to processes running on the machine which is running the tests (in this case circleci). We aren't doing that anymore and it's just one more thing which can go wrong on the test when the setup code fails so I removed it.

PR #979, opened 2023-01-31
fix: fix price graph

Extheoisah:


app/components/price-graph/price-graph.tsx
```diff
@@ -49,6 +49,7 @@ export const PriceGraphDataInjected = () => {
 
   const { error, loading, data, refetch } = useBtcPriceListQuery({
     fetchPolicy: "no-cache",
+    variables: { range: graphRange },
```
nicolasburtey, line 52: the value is passed there already: https://github.com/GaloyMoney/galoy-mobile/pull/979/files#diff-d75c8934e23c87b556deda51f102e3d3045a1da88266c8a2391ba51c8a130873R68

if there is an issue with this function, line 68, 72, ... probably needs some refactoring. ie: update the state of `graphRange` with `setGraphRange` instead of calling refetch directly?

nicolasburtey (changes requested):
probably something is missing in the refactoring.

also title doesn't need `fix: fix`

app/components/price-graph/price-graph.tsx
```diff
@@ -48,46 +48,39 @@ export const PriceGraphDataInjected = () => {
...
-        refetch({ range: GraphRange.ONE_YEAR })
-      }
-    } else if (graphRange === GraphRange.FIVE_YEARS) {
-      if (unixTime - lastPrice.timestamp > 86400) {
-        refetch({ range: GraphRange.FIVE_YEARS })
-      }
+    const index = ranges.indexOf(graphRange)
+    if (unixTime - lastPrice.timestamp > timestamps[index]) {
+      setGraphRange(ranges)
+      refetch({ range: ranges })
```
nicolasburtey, line 77: are we refetching twice right now? here and https://github.com/GaloyMoney/galoy-mobile/pull/979/files#diff-d75c8934e23c87b556deda51f102e3d3045a1da88266c8a2391ba51c8a130873R57 ?

Extheoisah: Oops! You're right. The useEffect is unnecessary.

Extheoisah:
@nicolasburtey I refactored the logic for the refetch here https://github.com/GaloyMoney/galoy-mobile/pull/979/commits/b9db58db72e1adedc41635406c70620b97dd9e98

nicolasburtey (approved):

PR #981, opened 2023-01-31
fix: Fixed language parsing issue on settings page

daviroo:
The server responds with an [empty string](https://github.com/GaloyMoney/galoy/blob/6bbbbb46ed3e196ab362c7c875207dbff7059ba2/src/graphql/types/scalar/language.ts#L30) if the language is set as default and the code doesn't deal with this at the moment.  I was seeing the `FIXME: error missing language` message in the settings page when my language was set as default.

### DIFF
--- a/app/screens/settings-screen/settings-screen.tsx
+++ b/app/screens/settings-screen/settings-screen.tsx
@@ -69,21 +69,13 @@ export const SettingsScreen: React.FC<Props> = ({ navigation }) => {
 
   const username = data?.me?.username ?? undefined
   const phone = data?.me?.phone ?? undefined
-  const languageQuery = data?.me?.language
-
-  // The server responds with "" if the user has set their language to DEFAULT
-  let parsedLanguageFromQuery
-  if (languageQuery) {
-    parsedLanguageFromQuery = languageQuery
-  } else {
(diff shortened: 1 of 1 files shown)


app/screens/settings-screen/settings-screen.tsx
```diff
@@ -69,13 +69,21 @@ export const SettingsScreen: React.FC<Props> = ({ navigation }) => {
 
   const username = data?.me?.username ?? undefined
   const phone = data?.me?.phone ?? undefined
-  const languageQuery = data?.me?.language ?? "DEFAULT"
+  const languageQuery = data?.me?.language
```
nicolasburtey, line 72:
```suggestion
  const languageQuery = data?.me?.language || "DEFAULT"
```

should do the trick.

But it's the type of use case where `fixing` the backend would be better imo?

nicolasburtey: yup that works @daviroo 

```
$ node
Welcome to Node.js v18.10.0.
Type ".help" for more information.
> let language = undefined
undefined
> language || "Default"
'Default'
> language ?? "Default"
'Default'
> language = ""
''
> language || "Default"
'Default'
> language ?? "Default"
''
```

nicolasburtey (approved):

PR #982, opened 2023-01-31
chore: use setValue for improved input key send

Extheoisah:


daviroo (commented):
Did you run the ios tests locally?

Extheoisah:
> Did you run the ios tests locally?

Yeah. I ran the tests on both ios and android locally and it passed. The experience was better than before.
I used this doc as reference https://webdriver.io/docs/api/element/addValue

nicolasburtey (approved):

PR #984, opened 2023-01-31
fix: remove unrecognized character

UncleSamtoshi:


app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
```diff
@@ -135,8 +135,8 @@ const RedeemBitcoinDetailScreen = ({
 }: StackScreenProps<RootStackParamList, "redeemBitcoinDetail">) => {
   const { callback, domain, defaultDescription, k1, minWithdrawable, maxWithdrawable } =
     route.params
-  const minWithdrawableSatoshis = minWithdrawable / 1_000
-  const maxWithdrawableSatoshis = maxWithdrawable / 1_000
+  const minWithdrawableSatoshis = minWithdrawable / 1000
```
nicolasburtey, line 138: weird. this is valid `ts` afaik? probably a wrong setting somewhere in our tsconfig or somewhere else?

UncleSamtoshi: Yeah I thought it was strange as well. It works locally.

nicolasburtey: would be worth understand what settings are different but approving for now probably not worth the effort

nicolasburtey (approved):

PR #993, opened 2023-02-01
chore: refactor setting languages

UncleSamtoshi:
- add thai language support
- improve typing around languages
- fix bug with setting languages

### DIFF
--- a/__tests__/utils/locale-detector.test.ts
+++ b/__tests__/utils/locale-detector.test.ts
@@ -1,0 +1,28 @@
+import { matchOsLocaleToSupportedLocale } from "../../app/utils/locale-detector"
+
+describe("matchOsLocaleToSupportedLocale", () => {
+    it("exactly matches a supported locale", () => {
+        const supportedCountyAndLang = [
+            { countryCode: "CA", languageTag: "fr-CA", languageCode: "fr", isRTL: false },
+        ]
+        const locale = matchOsLocaleToSupportedLocale(supportedCountyAndLang)
+        expect(locale).toEqual("fr-CA")
+
+    })
+
+    it("approximately matches a supported locale", () => {
+        const unsupportedCountrySupportedLang = [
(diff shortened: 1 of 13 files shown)


app/utils/locale-detector.ts
```diff
@@ -1,38 +1,60 @@
...
-    RNLocalize.getLocales() &&
-    RNLocalize.getLocales().some(
-      (locale) =>
-        locale.languageCode?.startsWith("en") ||
-        locale.languageCode?.startsWith("es") ||
-        locale.languageCode?.startsWith("fr") ||
-        locale.languageCode?.startsWith("pt") ||
-        locale.languageCode?.startsWith("de") ||
-        locale.languageCode?.startsWith("cs"),
+export const detectDefaultLocale = (): Locales => {
```
nicolasburtey, line 5: some tests on this could be useful

UncleSamtoshi: Added some tests

__tests__/utils/locale-detector.test.ts
```diff
@@ -0,0 +1,28 @@
+import { matchOsLocaleToSupportedLocale } from "../../app/utils/locale-detector"
```
nicolasburtey, line 1: side note: I have been confused as some test are "in line", ie in the same folder as the other one, and other tests are in `__tests__`.

I dpn't really care which way we go, but having consistency here would be useful. feel free to take this PR if you'd like probably a small one

test.js
```diff
@@ -0,0 +1,7 @@
+const test = function(a, b) {
```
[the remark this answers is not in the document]
UncleSamtoshi: Yes i will remove that

__tests__/utils/locale-detector.test.ts
UncleSamtoshi: I will create a follow up PR to bring consistency to the testing.

nicolasburtey (approved):

PR #997, opened 2023-02-02
fix: Fixed issue with btc wallet balance

daviroo:
BTC wallet sat balance was showing as `NaN` on the send flow.  I also added a new test to look out for this in future.

image

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
@@ -575,12 +575,20 @@ const SendBitcoinDetailsScreen = ({
                 <View style={Styles.walletSelectorBalanceContainer}>
                   {sendingWalletDescriptor.currency === WalletCurrency.Btc ? (
                     <>
-                      <Text style={Styles.walletBalanceText}>
+                      <Text
+                        style={Styles.walletBalanceText}
+                        {...testProps("BTC Wallet Balance in USD")}
+                      >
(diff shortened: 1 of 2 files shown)


app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -575,12 +575,20 @@ const SendBitcoinDetailsScreen = ({
...
                           ? formatToDisplayCurrency(btcBalanceInUsd) + " - "
                           : ""}
+                      </Text>
+                      <Text
+                        style={Styles.walletBalanceText}
+                        {...testProps("BTC Wallet Balance in sats")}
+                      >
                         {typeof btcWalletBalance === "number"
-                          ? Number(satAmountDisplay(btcWalletBalance))
+                          ? satAmountDisplay(btcWalletBalance)
```
nicolasburtey, line 591: why is this fixing the issue?

nicolasburtey (approved):

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
daviroo: `satAmountDisplay` returns a string which can't be converted to a number because it returns the value + "sats" https://github.com/GaloyMoney/galoy-mobile/blob/23b92256af045d7862da6667e6aaf73d12e606dd/app/utils/currencyConversion.ts#L117

PR #1005, opened 2023-02-03
chore: Added function to silence logs

daviroo:


app/app.tsx
```diff
@@ -28,6 +28,9 @@ import { NavigationContainerWrapper } from "./navigation/navigation-container-wr
 import theme from "./rne-theme/theme"
 import { PersistentStateProvider } from "./store/persistent-state"
 import { detectDefaultLocale } from "./utils/locale-detector"
+import silenceAlerts from "./config/silence-alerts"
+
+silenceAlerts()
```
nicolasburtey, line 33: we should fix the underlying issue instead of silence them

so many of issues/crash we had was because we disabled checks here and there

daviroo: I don't think this is a real issue.  The warning is related to a function we are passing in the react-navigation params.  The reason they warn is because this could break persistence or deep linking but neither of those apply in this part of the flow.  You can read more here about why they recommend against it https://reactnavigation.org/docs/troubleshooting/#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state

nicolasburtey: Feels like passing function around instead of id can be an anti pattern that make testing more complex because more mocking is needed

PR #1006, opened 2023-02-03
chore: RN 0.71.2 upgrade

nicolasburtey:
- hermes is now activated on iOS. this should make the app much smaller and faster to boot up
- I had to update all the package in react native. not ideal but I ran into issue and was not sure what it was. upgrading all fixed it
- I clean up some ios/android files like Podfile or build.gradle a little bit. for instance we were download a package for youtube which we were not using. not sure if this had an impact on the build size as we had no reference in our code to it.
- remove some files that were not used android/browserstack.js and ios/browserstack.js
- will need to dive on whether there is some potential regression with notifications (we probably need test on that) after PR is been merged. https://github.com/wix/react-native-notifications/issues/938


nicolasburtey:
maybe an issue: https://github.com/wix/react-native-notifications/issues/938

nicolasburtey:
this commit still feels necessary to be merged: https://github.com/mrousavy/react-native-vision-camera/pull/1419

could do a patch in the meantime

UncleSamtoshi:
Built it on both ios and android and it was able to open.

UncleSamtoshi (approved):

PR #1027, opened 2023-02-06
fix: settings / contact support appearance modal screen

nicolasburtey:
also:
- remove `mock-apollo-client`. a better version is already included with apollo
- make settings screen works back with storybook
- add i18n wrapper for storybook so that component that depend on LL are working property within storybooks

### DIFF
--- a/app/app.tsx
+++ b/app/app.tsx
@@ -1,7 +1,14 @@
 // Welcome to the main entry point of the app.
 //
-// In this file, we'll be kicking off our app or storybook.
+// In this file, we'll be kicking off our app
+
+// language related import
 import "intl-pluralrules"
+import "./utils/polyfill"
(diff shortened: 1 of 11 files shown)


app/app.tsx
```diff
@@ -35,29 +37,33 @@ EStyleSheet.build({
   // $textColor: '#0275d8'
 })
 
+// FIXME should we only load the currently used local?
+// this would help to make the app load faster
+// this will become more important when we add more languages
+// and when the earn section will be added
+//
+// alternatively, could try loadAllLocalesAsync()
```
daviroo, line 45: On my local file system each locale takes up ~60kb.  I think we could load it async without taking up too much user bandwidth.

daviroo (approved):

PR #1041, opened 2023-02-07
fix(i18n): Fixed some spelling mistakes for german language

fmaul:
While using the iOS Beta I recognized some spelling mistakes in the german translation. This PR contains some fixes.


nicolasburtey:
> While using the iOS Beta I recognized some spelling mistakes in the german translation. This PR contains some fixes.

Hi fmaul, thanks for your contribution!

I'm not sure if editing the json directly is synchronizing directly with transifex, our source of truth for the translation (ie: will the get erased on the next rollout from transifex?)

I'll let @UncleSamtoshi answer as he has more context on the integration with transifex. 

cc @pretyflaco @agbegin

fmaul:
OK, I see you are using transifex. Let me know if there is a better way to contribute translations.

nicolasburtey:
> OK, I see you are using transifex. Let me know if there is a better way to contribute translations.

easiest would probably be to onboard you to transifex! I'm not up to date on my end on transifex but someone on the team with more knowledge will respond shortly

nicolasburtey:
merging this now so this is part of the next beta but eventually should be done in transifex otherwise the changes might be reverted

nicolasburtey (approved):

PR #1046, opened 2023-02-07
fix: refresh on send

nicolasburtey:
I don't quite get yet how all the send flow is working from a type perspective and the `sendPayment`, `sendPaymentFn` and `useSendPayment` and how it's flowing between screen and where argument are been set

in essence, the goal of this PR is to add `refetchQueries` as an argument to the payment mutation, but I don't think this is where it should be set because I would guess the argument are been passed somewhere else. 

I think my approach may work but @UncleSamtoshi can you double check?

### DIFF
--- a/app/screens/send-bitcoin-screen/use-send-payment.ts
+++ b/app/screens/send-bitcoin-screen/use-send-payment.ts
@@ -1,4 +1,5 @@
 import {
+  MainAuthedDocument,
   PaymentSendResult,
   useIntraLedgerPaymentSendMutation,
   useIntraLedgerUsdPaymentSendMutation,
@@ -11,7 +12,7 @@ import { useMemo } from "react"
 import { SendPayment } from "./payment-details/index.types"
 import { gql } from "@apollo/client"
 
-export type UseSendPaymentResult = {
+type UseSendPaymentResult = {
   loading: boolean
   sendPayment:
     | (() => Promise<{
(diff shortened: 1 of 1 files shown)


app/screens/send-bitcoin-screen/use-send-payment.ts
```diff
@@ -26,18 +27,24 @@ export const useSendPayment = (
 ): UseSendPaymentResult => {
   const [intraLedgerPaymentSend, { loading: intraLedgerPaymentSendLoading }] =
     useIntraLedgerPaymentSendMutation()
```
UncleSamtoshi, line 29: Although the current way may work, I think it is cleaner to pass the refetch queries to the hook like 
```
useIntraLedgerPaymentSendMutation({
    refetchQueries: [MainAuthedDocument]
})
```

UncleSamtoshi (dismissed):

app/screens/send-bitcoin-screen/use-send-payment.ts
nicolasburtey: I've tried that, but you can't pass this variable here

nicolasburtey: maybe give it a try, I may have had the type wrong

UncleSamtoshi: What wasnt working for you? I dont have any type errors when I do that

nicolasburtey: was not working but working now. probably ts cache issue :/

UncleSamtoshi (approved):

PR #1048, opened 2023-02-07
chore: refactor check for self intraledger payment

Extheoisah:
Currently if you try to pay yourself using your username in the app, you get a toast UI with the message `Error checking username`. We instead want to show the user the message telling them they can't pay themselves. This PR fixes that.


UncleSamtoshi (approved):

UncleSamtoshi:
I'm currently rewriting most of that function, but this can be in place in the mean time.

nicolasburtey:
> I'm currently rewriting most of that function, but this can be in place in the mean time.

we're trying to ship a beta update this week so good to have this fix

PR #1052, opened 2023-02-07
chore: storybook improvment

nicolasburtey:
just a simple refactoring for now:
- making sure all storybook load
- fixing a few types
- using storybook v6 beta instead of v5
- proof of concept of the new format [app/screens/settings-screen/settings-screen.stories.tsx](https://github.com/GaloyMoney/galoy-mobile/blob/8cf7bf0f9da3da12334f497735365351f2660e47/app/screens/settings-screen/settings-screen.stories.tsx)
- moving some files around
-- storybook/ --> .storybook
-- using the stories.tsx pattern that is now suggested as best practice from documentation

note: stories.tsx still excludes both from eslint and tsc currently. will slowly fix that.


daviroo:
> making sure all storybook load

How are you achieving this?

nicolasburtey:
> > making sure all storybook load
> 
> How are you achieving this?

manual operation for now. had to fix a few things on some screens to make all the stories work,

running this https://storybook.js.org/docs/react/writing-tests/test-runner in CI would be the right solution. 

but right now server doesn't work yet, there is some incompatibilities issues to look at.

daviroo (approved):

Issue #1060, opened 2023-02-08
Cached wallet issue?


nicolasburtey:
Alice: Set usd as default wallet
Bob: send intraledger payment through mobile app. Payment appears in usd
Alice: change default wallet to btc
Bob: send intraledger to Bob. Payment still appear in usd.

PR #1077, opened 2023-02-08
chore: prevent stale defaultWalletId

UncleSamtoshi:


UncleSamtoshi:
Fixes issue #1060

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -218,7 +218,9 @@ const SendBitcoinDestinationScreen = ({
   const contacts = useMemo(() => data?.me?.contacts ?? [], [data?.me?.contacts])
 
   const { LL } = useI18nContext()
-  const [userDefaultWalletIdQuery] = useUserDefaultWalletIdLazyQuery()
+  const [userDefaultWalletIdQuery] = useUserDefaultWalletIdLazyQuery({
+    fetchPolicy: "network-only",
```
nicolasburtey, line 222: do we still want to cache it? maybe we could go for `no-cache`?

UncleSamtoshi: Switching to that

nicolasburtey (approved):

Issue #1089, opened 2023-02-09
Scanning an expired invoice shows: This is not a valid Bitcoin address or Lightning invoice


openoms:
Scanning an expired invoice shows: This is not a valid Bitcoin address or Lightning invoice while showing the correct (but expired) invoice.

Reproduce:
Scan the invoice from snort after two minutes (expiry 117)
https://snort.social/p/npub14tq8m9ggnnn2muytj9tdg0q6f26ef3snpd7ukyhvrxgq33vpnghs8shy62

![image](https://user-images.githubusercontent.com/43343391/217816446-da6848a0-9ed3-478f-8e1f-0bc92d4c94ad.png)

Would be better if the user would be informed of the problem which is the expiration.

Aditi-Singh16:
In this, I would like to test for expired invoices, how does one have their invoice expire?

openoms:
> In this, I would like to test for expired invoices, how does one have their invoice expire?

One way is to create an invoice with a short expiry and wait.

Aditi-Singh16:
So, I debugged and this snort qr code is not expired but an unknown destination error, so I will add for that condition

PR #1097, opened 2023-02-09
chore: destination logic refactor

UncleSamtoshi:
- refactor destination logic so that it can be shared across screens
- enable parsing of lnurlwithdraw strings from the SendBitcoinDestinationScreen


app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -256,7 +254,9 @@ const SendBitcoinDetailsScreen = ({
   const [isModalVisible, setIsModalVisible] = useState(false)
   const [errorMessage, setErrorMessage] = useState("")
   const [validAmount, setValidAmount] = useState(false)
-  const usdDisabled = paymentType === "onchain" || usdWalletId === undefined
+  const usdDisabled =
```
nicolasburtey, line 257: I feel we could remove this assumption for now. I don't think we need this complexity of managing having or not having usd enabled. can be done as part of another PR.

app/screens/send-bitcoin-screen/destination-information.tsx
```diff
@@ -6,21 +6,22 @@ import { useAppConfig } from "@app/hooks"
...
 import { IntraledgerPaymentDestination } from "@galoymoney/client/dist/parsing-v2"
+import { InvalidDestinationReason } from "./payment-destination/index.types"
 
 const toLnAddress = (handle: string, lnDomain: string) => {
   return `${handle}@${lnDomain}`
 }
 
 const destinationStateToInformation = (
-  destinationState: SendBitcoinDestinationState,
+  sendBitcoinReducerState: SendBitcoinDestinationState,
```
nicolasburtey, line 18: I think we should also keep, for consistency, the pattern we have in the backend of using object are argument, instead of positional argument. 

we can do those changes in another PR

app/screens/send-bitcoin-screen/destination-information.tsx
```diff
@@ -6,21 +6,22 @@ import { useAppConfig } from "@app/hooks"
...
   translate: TranslationFunctions,
```
nicolasburtey, line 19: you could just pass `translate.SendBitcoinDestinationScreen`. I think otherwise we typically call the translate function `TT` in the other part of the app (assuming this is the same thing). we should keep the same function name here

app/screens/send-bitcoin-screen/destination-information.tsx
```diff
@@ -6,21 +6,22 @@ import { useAppConfig } from "@app/hooks"
...
 import { Text, View } from "react-native"
 import EStyleSheet from "react-native-extended-stylesheet"
-import { SendBitcoinDestinationState } from "./send-bitcoin-reducer"
+import { DestinationState, SendBitcoinDestinationState } from "./send-bitcoin-reducer"
 import { IntraledgerPaymentDestination } from "@galoymoney/client/dist/parsing-v2"
+import { InvalidDestinationReason } from "./payment-destination/index.types"
 
 const toLnAddress = (handle: string, lnDomain: string) => {
   return `${handle}@${lnDomain}`
 }
```
nicolasburtey, line 15: lnDomain doesn't change. this could be memoized, such that when it's called only the handle needs to be passed 

```suggestion
const toLnAddress = (lnDomain: string)(handle: string)  => {
  return `${handle}@${lnDomain}`
}
```

nicolasburtey: means `toLnAddress ` should be in `destinationStateToInformation`

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -93,33 +87,29 @@ type ScanningQRCodeScreenProps = {
...
+      }
+      contacts {
+        id
+        username
+      }
+    }
+  }
+
+  # TODO replace with AccountDefaultWallet?
+  query userDefaultWalletId($username: Username!) {
```
nicolasburtey, line 111: I didn't know we could copy the query multiple times

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -213,232 +161,87 @@ const SendBitcoinDestinationScreen = ({
...
-    return async (username: string) =>
-      getUsernameStatus({
-        username,
-        contacts: lowercaseContacts,
-        myUsername: lowercaseMyUsername,
-        getWalletIdForUsername,
-      })
-  }, [contacts, userDefaultWalletIdQuery, myUsername])
+  const [userDefaultWalletIdQuery] = useUserDefaultWalletIdLazyQuery({
+    fetchPolicy: "cache-and-network",
```
nicolasburtey, line 170: it's a lazy query. I don't think we want to use `cache and network`:
- because its a lazy query, it acts like a mutation and will only be called when the function is called instead of when the screen is loaded. 

so I'm not sure how we would get the network query backed from this function if the value is cached. I'll assume we don't

also, we may not want to cache this at all, so maybe `no-cache` is the appropriate solution.

app/screens/send-bitcoin-screen/payment-destination/lnurl.ts
```diff
@@ -0,0 +1,208 @@
...
+
+export type ResolveLnurlDestinationParams = {
+  parsedLnurlDestination: LnurlPaymentDestination
+  lnurlDomains: string[]
+  userDefaultWalletIdQuery: LazyQueryExecFunction<
+    UserDefaultWalletIdQuery,
+    Exact<{
+      username: string
+    }>
+  >
```
nicolasburtey, line 27:
```suggestion
  userDefaultWalletIdQuery: UserDefaultWalletIdLazyQueryHookResult
```

nicolasburtey: there are many instance where the type that could be simplify with the diff above

app/screens/send-bitcoin-screen/payment-destination/index.types.ts
```diff
@@ -0,0 +1,113 @@
...
+  rawInput: string
+  myWalletIds: string[]
+  bitcoinNetwork: Network
+  lnurlDomains: string[]
+  userDefaultWalletIdQuery: LazyQueryExecFunction<
+    UserDefaultWalletIdQuery,
+    Exact<{
+      username: string
+    }>
+  >
```
nicolasburtey, line 31:
```suggestion
  userDefaultWalletIdQuery: UserDefaultWalletIdLazyQueryHookResult[0]
```

app/screens/send-bitcoin-screen/payment-destination/index.types.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export type ValidParsedReceiveDestination = LnurlWithdrawDestination
+
+export type LnurlWithdrawDestination = {
+  valid: true
+  lnurl: string
+  callback: string
+  domain: string
+  k1: string
```
nicolasburtey, line 109: what is `k1`?

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
UncleSamtoshi: Yeah my one thinking for cache and network was that the call would always succeed instantly if it was in the cache, and then the next time it was called it would be the up to date value. The trade off would be that if someone changed default wallets, there would be one payment to them using the old wallet id. I suppose however that if the getWaleltId query fails there will probably be other things failing and it wont be worth it to try and prevent that situation.

nicolasburtey: >  I suppose however that if the getWaleltId query fails there will probably be other things failing and it wont be worth it to try and prevent that situation.

exactly. 

I think this is like the getFee where we try to preemptively calculate them locally if this is from our node instead of calling the server. the benefits do not outweigh the added complexity

app/screens/send-bitcoin-screen/destination-information.tsx
UncleSamtoshi: I think this should be handled in when creating the fix for issue #1089

app/screens/send-bitcoin-screen/payment-destination/index.types.ts
UncleSamtoshi: It is part of the lnurl withdraw spec that the lnurl server uses to identify the calling wallet https://github.com/lnurl/luds/blob/luds/03.md

app/screens/send-bitcoin-screen/payment-destination/index.types.ts
UncleSamtoshi: Changed it

app/screens/send-bitcoin-screen/destination-information.tsx
UncleSamtoshi: This can be addressed with issue #1089

nicolasburtey (approved):

nicolasburtey (approved):

PR #1100, opened 2023-02-10
chore: remove failing usd e2e test

Extheoisah:


nicolasburtey (changes requested):
you should use `it.skip()` instead

nicolasburtey (approved):

PR #1101, opened 2023-02-10
fix: correct mempool.space link for signet

### LINKED ISSUE #684: wrong mempool explorer when not on mainnet
when I click on a transaction detail/hash, there is a hyperlink opening to mempool
it seems to always assume the hash is for mainnet

nicolasburtey:
fix #684

### DIFF
--- a/app/config/galoy-instances.ts
+++ b/app/config/galoy-instances.ts
@@ -15,6 +15,7 @@ export type GaloyInstance = {
   graphqlWsUri: string
   posUrl: string
   lnAddressHostname: string
+  blockExplorer: string
 }
 
 export const GALOY_INSTANCES: GaloyInstance[] = [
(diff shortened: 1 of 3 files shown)


app/config/galoy-instances.ts
```diff
@@ -24,19 +25,22 @@ export const GALOY_INSTANCES: GaloyInstance[] = [
     graphqlWsUri: "wss://api.mainnet.galoy.io/graphql",
     posUrl: "https://pay.bbw.sv",
     lnAddressHostname: "pay.bbw.sv",
+    blockExplorer: "https://mempool.space/tx/",
```
UncleSamtoshi, line 28: The GaloyInstance is stored in the PersistentStorageProvider and needs to be migrated, otherwise the user will be stuck with their old version of the GaloyInstance. The PersistentStorageProvider should probably be refactored to only store the type of the GaloyInstance and then at runtime it grabs the correct GaloyInstance.

nicolasburtey: not sure to fully understand. I need to create a migration for that PR?

UncleSamtoshi: Right now there is a serialized GaloyInstance stored in local storage https://github.com/GaloyMoney/galoy-mobile/blob/2e209bd15c34abed160c84c69e2ecf71fb2a5c6e/app/store/persistent-state/state-migrations.ts#L40. The app config uses the GaloyInstance from localstorage, so if you update this GaloyInstance without manually updating what is stored in local storage it wont receive the changes.

PR #1107, opened 2023-02-10
chore: fix usd payment e2e test

Extheoisah:


e2e/03-validate-username-flow.e2e.spec.ts
```diff
@@ -42,7 +42,7 @@ describe("Validate Username Flow", () => {
     )
     const { isContactAvailable } = await checkContact(username)
     if (!isContactAvailable) {
-      if ((await checkBoxButton.isDisplayed()) || (await confirmButton.isEnabled())) {
+      if (await checkBoxButton.isDisplayed()) {
```
nicolasburtey, line 45: if in test is an antipattern btw

nicolasburtey (changes requested):
instead of renaming the file each time a test is added or removed, can you refactor them logically in a small group of files?

nicolasburtey (approved):

PR #1111, opened 2023-02-10
fix: inconsistent e2e token

UncleSamtoshi:
- currently we grab a random token every time we are trying to grab the same token
- this pr chooses a single random token and uses that for the remainder of the test


nicolasburtey (approved):
good catch

PR #1113, opened 2023-02-11
chore: make it work on m2 with rosetta

nicolasburtey:


Extheoisah (approved):
For M1 users with existing code, this might trigger an update. So we'd have to run this command `pod install --repo-update`
Asides, that its all good.

PR #1114, opened 2023-02-11
docs: update token key

nicolasburtey:


UncleSamtoshi:
Are you making the changes to circle ci env at the same time that you release this?

UncleSamtoshi (approved):

nicolasburtey:
> Are you making the changes to circle ci env at the same time that you release this?

no because it was already changed in the code some times ago (when we went from 1 token to many to avoid conflicting response with multiple tests running at the same time)

doc was not up to date. not sure why Fastlane actually worked. maybe it's not been used? just did a search and replaced to be on par with the code.

PR #1116, opened 2023-02-11
chore: home+earn storybooked

nicolasburtey:


app/screens/earns-screen/earns-quiz.tsx
```diff
@@ -178,15 +178,48 @@ gql`
...
-  const [isCompleted, setIsCompleted] = useState(completed)
+  const [isCompleted, setIsCompleted] = useState(false)
   const [quizVisible, setQuizVisible] = useState(false)
   const [recordedAnswer, setRecordedAnswer] = useState<number[]>([])
-  const [permutation] = useState<ZeroTo2[]>(shuffle([0, 1, 2]))
-  const { LL } = useI18nContext()
+
+  // when apollo loading = true, completed value will be false
+  // so we have to wait for the loading to complete
+  React.useEffect(() => {
```
UncleSamtoshi, line 218: Not sure I understand why this is necessary.

nicolasburtey: because if we have that:
`const [isCompleted, setIsCompleted] = useState(completed)`

and completed is derived from:
`data?.me?.defaultAccount?.quiz?.slice() ?? []`

and is `false` if data is null, then `completed` will always be initialized to false. 

because on the first run, apollo client will return `{loading: true, data: undefined}` (even if the data is cached)

UncleSamtoshi: That makes sense for that change, but is `isCompleted` necessary at all or could we just use `completed`

nicolasburtey: I think you are right actually. I just re-use `  const [isCompleted, setIsCompleted] = useState(completed)` because it was here already but didn't think about it. 

maybe let's merge this and I'll do that in a follow up PR?

UncleSamtoshi: Sounds good

UncleSamtoshi (approved):

PR #1118, opened 2023-02-12
chore: only subscribe to myUpdates once

### LINKED ISSUE #1062: Success icon disappearing on payment success for receive
Tested with wos on same phone 

It make the invoice that have been paid will been displayed in the app

nicolasburtey:
also may fix #1062

### DIFF
--- a/app/graphql/ln-update-context.ts
+++ b/app/graphql/ln-update-context.ts
@@ -1,0 +1,7 @@
+import { createContext, useContext } from "react"
+
+const LnUpdateHash = createContext<string>("")
+
+export const LnUpdateHashProvider = LnUpdateHash.Provider
+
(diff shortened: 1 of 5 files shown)


app/graphql/ln-update-context.ts
```diff
@@ -0,0 +1,7 @@
+import { createContext, useContext } from "react"
+
+const LnUpdateHash = createContext<string>("")
```
UncleSamtoshi, line 3: I would consider changing the naming to indicate that the LnUpdate is Paid, something like `LnUpdateHashPaid`

nicolasburtey: updated

Extheoisah (approved):

PR #1120, opened 2023-02-12
chore: make it possible to parse build number more easily

nicolasburtey:
not sure why implication this may have on the build if new build on a same arch have lower version


UncleSamtoshi (dismissed):

UncleSamtoshi:
> not sure why implication this may have on the build if new build on a same arch have lower version

I've had problems in the past locally downgrading versions, but I dont recall the specifics of it. To be safe you could round up to 10,000,000 instead of down to 1,000,000

UncleSamtoshi (approved):

PR #1122, opened 2023-02-12
fix: notification library

nicolasburtey:


app/navigation/local-notification-wrapper.tsx
```diff
@@ -0,0 +1,40 @@
+import { useApolloClient } from "@apollo/client"
+import { useIsAuthed } from "@app/graphql/is-authed-context"
+import { addDeviceToken, hasNotificationPermission } from "@app/utils/notifications"
+import messaging, { FirebaseMessagingTypes } from "@react-native-firebase/messaging"
+import { useEffect } from "react"
+
+export const NotificationWrapper = (): JSX.Element => {
```
UncleSamtoshi, line 7: This doesn't appear to be a wrapper, but just a component that handles the notification logic.

app/navigation/app-state.tsx
```diff
@@ -0,0 +1,37 @@
...
+    },
+    [client, isAuthed],
+  )
+
+  useEffect(() => {
+    const subscription = AppState.addEventListener("change", handleAppStateChange)
+    return () => subscription.remove()
+  }, [handleAppStateChange])
+
+  return <>{children}</>
```
UncleSamtoshi, line 36: It it necessary for AppStateWrapper to pass through `<>{children}</>` and be a parent component of them, or could it just be a sibling component in `app.tsx`.

UncleSamtoshi (dismissed):

app/navigation/local-notification-wrapper.tsx
nicolasburtey: any suggestion of name we should give to such component?

app/navigation/app-state.tsx
nicolasburtey: yeah good can be just a sibling

Extheoisah (approved):

PR #1124, opened 2023-02-13
chore: using the reducer directly

nicolasburtey:
I would recommend not wrapping the reducer. I think it should be like a useState where it's easier to read if it's paste directly into the Component itself

this is especially true as this component is not re-use elsewhere


UncleSamtoshi:
I don't have a strong opinion on this style, so I'm happy to switch it. To avoid a merge conflict, I think it may just be easier for me to add it to the parsing destination refactor and close this pr

nicolasburtey:
> I don't have a strong opinion on this style, so I'm happy to switch it. To avoid a merge conflict, I think it may just be easier for me to add it to the parsing destination refactor and close this pr

great let's do that

PR #1125, opened 2023-02-13
chore: unused code?

nicolasburtey:


nicolasburtey:
ok this doens't work but I'm confused by this function. maybe something to discuss during a daily dev.

PR #1141, opened 2023-02-14
chore: wrapping every call to the server using me throught a isAuthed()

nicolasburtey:


UncleSamtoshi:
It would probably be ideal if we could have the isAuthed variable as part of the state of the client or added to some kind of middleware so that this behavior is automatic. My concern would be that it would be easy to forget when adding a new query to not add the `skip` argument.

UncleSamtoshi (dismissed):

nicolasburtey:
> It would probably be ideal if we could have the isAuthed variable as part of the state of the client or added to some kind of middleware so that this behavior is automatic. My concern would be that it would be easy to forget when adding a new query to not add the `skip` argument.

agree, I also thought about it. not straightforward because we would need to list extensively (and have a check on the extensivity) on every queries and whether they're in one bucket or another one

currently the main issue is server side. it's returning a 401 error. so it's not a big deal if we forget one: it will appear server wise and we would then fix it.

UncleSamtoshi:
> > It would probably be ideal if we could have the isAuthed variable as part of the state of the client or added to some kind of middleware so that this behavior is automatic. My concern would be that it would be easy to forget when adding a new query to not add the `skip` argument.
> 
> agree, I also thought about it. not straightforward because we would need to list extensively (and have a check on the extensivity) on every queries and whether they're in one bucket or another one
> 
> currently the main issue is server side. it's returning a 401 error. so it's not a big deal if we forget one: it will appear server wise and we would then fix it.

That makes sense to me.

app/screens/receive-bitcoin-screen/receive-btc.tsx
```diff
@@ -227,7 +228,10 @@ const ReceiveBtc = () => {
   )
   const { convertCurrencyAmount } = usePriceConversion()
 
-  const { data } = useReceiveBtcQuery({ fetchPolicy: "cache-first" })
+  const { data } = useReceiveBtcQuery({
+    fetchPolicy: "cache-first",
+    skip: !useIsAuthed(),
```
Extheoisah, line 233: Not critical but we should probably follow one pattern of using the hook. either ` const isAuth = useIsAuth()` before using the variable `isAuth` or using the hook directly like is used here. I prefer the later though.

PR #1142, opened 2023-02-14
fix: 401 from e2e

nicolasburtey:
I have introduce the bug back a day or 2 ago when refactoring auth.

I've combined useToken and useAppConfig because I needed a method to both set the apollo instance and the token simulatenously, otherwise one client at a time would be created, which would be creating 401 from the the main server (ie: set the token first, then there would already be called been initiated to the server, then the galoy instance would be set correctly, but there would still be 401 error been returned)


app/hooks/use-logout.ts
```diff
@@ -11,23 +11,28 @@ const useLogout = () => {
...
-    try {
-      await client.cache.reset()
-      await AsyncStorage.multiRemove([BUILD_VERSION])
-      await KeyStoreWrapper.removeIsBiometricsEnabled()
-      await KeyStoreWrapper.removePin()
-      await KeyStoreWrapper.removePinAttempts()
+  const logout = useCallback(
+    async (stateToDefault = true): Promise<void> => {
+      try {
+        await client.cache.reset()
```
UncleSamtoshi, line 17: I wonder if `client.cache.reset()` should be moved into a cleanup function in the useEffect that creates the client. My concern is that the cache could be reset in line 17 and then throughout the remainder of the function whenever the calls are awaited, there is a possibility for a rerender where the client cache would be repopulated again.

UncleSamtoshi (dismissed):

PR #1152, opened 2023-02-15
chore: misc queries improvments

nicolasburtey:
chore: separate tx query from the rest of the query
this allow to not reset the cache for tx for now on home refresh. this is a tradeoff because now all but the tx are been refetch from the home screen on refresh (ie: scroll down to show the "refresh" wheel)

I also increase the number of tx to 12 so that when someone move to tx they can see the first screen fully populated while the other transactions are been fetched in the background

### DIFF
--- a/app/graphql/analytics.tsx
+++ b/app/graphql/analytics.tsx
@@ -1,11 +1,11 @@
 import analytics from "@react-native-firebase/analytics"
-import { useRootStackQuery } from "@app/graphql/generated"
+import { useAnalyticsQuery } from "@app/graphql/generated"
 import { gql } from "@apollo/client"
 import { useEffect } from "react"
 import { useIsAuthed } from "./is-authed-context"
 
 gql`
-  query rootStack {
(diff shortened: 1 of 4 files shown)


UncleSamtoshi:
I'm concerned that this could result in a frustrated user repeatedly trying to refetch but nothing happening when they are expecting a tx to show up. Usually when im refetching, its to try and get the latest transactions.

nicolasburtey:
> I'm concerned that this could result in a frustrated user repeatedly trying to refetch but nothing happening when they are expecting a tx to show up. Usually when im refetching, its to try and get the latest transactions.

fair point. this also make me think that the mainAuthed query is the forced-refreshed when there are payments made. my PR would not make those transaction appear either I believe.

I need to rethink this a bit more. won't merge it for now.

UncleSamtoshi:
Looks like you may have accidentally added a `galoy` folder.

UncleSamtoshi (approved):

PR #1153, opened 2023-02-15
fix: lnurl parsing

UncleSamtoshi:
Fixes ln address issue #1148

### DIFF
--- a/__tests__/payment-destination/onchain.spec.ts
+++ b/__tests__/payment-destination/onchain.spec.ts
@@ -8,7 +8,6 @@ import {
   CreateAmountOnchainPaymentDetailsParams,
   CreateIntraledgerPaymentDetailsParams,
 } from "@app/screens/send-bitcoin-screen/payment-details"
-import { createMock } from "ts-auto-mock"
 
 const mockCreateAmountLightningPaymentDetail = jest.fn<
   PaymentDetail<WalletCurrency>,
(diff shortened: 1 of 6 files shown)


__tests__/payment-destination/onchain.spec.ts
```diff
@@ -45,101 +44,9 @@ jest.mock("@app/screens/send-bitcoin-screen/payment-details", () => {
...
-  createLnurlPaymentDestination,
-  createLightningDestination,
-  createIntraLedgerDestination,
-  createOnchainDestination,
-} from "@app/screens/send-bitcoin-screen/payment-destination"
-import { LnUrlPayServiceResponse } from "lnurl-pay/dist/types/types"
+import { createOnchainDestination } from "@app/screens/send-bitcoin-screen/payment-destination"
 import { defaultPaymentDetailParams } from "./helpers"
 
-describe("create lightning destination", () => {
```
UncleSamtoshi, line 57: deleting these tests because they were accidentally copied over from other files

nicolasburtey:
> Fixes ln address issue #1148

do we have some unit tests around this issue? 
it seems there are some tests but they were not be used to find or fix the issue?

app/screens/send-bitcoin-screen/payment-destination/lnurl.ts
```diff
@@ -31,62 +31,56 @@ export const resolveLnurlDestination = async ({
   userDefaultWalletIdQuery,
   myWalletIds,
 }: ResolveLnurlDestinationParams): Promise<ParseDestinationResult> => {
-  if (parsedLnurlDestination.valid) {
-    try {
-      const lnurlParams = await getParams(parsedLnurlDestination.lnurl)
-      if ("reason" in lnurlParams) {
-        throw lnurlParams.reason
-      }
+  // TODO: Move all logic to galoy client or out of galoy client, currently lnurl pay is handled by galoy client
```
nicolasburtey, line 34: I think all in client eventually make most sense. could be after we have done all the tests/iteration for it and we want to work on refactoring the web wallet

UncleSamtoshi: Yeah I think that makes sense

UncleSamtoshi:
> > Fixes ln address issue #1148
> 
> do we have some unit tests around this issue? it seems there are some tests but they were not be used to find or fix the issue?

This was not currently tested, I will to add a unit test around it.

nicolasburtey (approved):

nicolasburtey:
> > > Fixes ln address issue #1148
> > 
> > 
> > do we have some unit tests around this issue? it seems there are some tests but they were not be used to find or fix the issue?
> 
> This was not currently tested, I will to add a unit test around it.

ok. let's merge it now and roll this fix out tomorrow in the beta track, but also let's add the test in a separate PR to make sure we don't break this again in the future

PR #1156, opened 2023-02-16
fix: crash on android when going to brackground

nicolasburtey:


daviroo (approved):

daviroo:
Can you give some context on why this is necessary?

nicolasburtey:
> Can you give some context on why this is necessary?

part of the install requirement: https://github.com/software-mansion/react-native-screens#android

I guess it was lost on some RN upgrade

nicolasburtey:
> > Can you give some context on why this is necessary?
> 
> part of the install requirement: https://github.com/software-mansion/react-native-screens#android
> 
> I guess it was lost on some RN upgrade

it's also why there has been some crash on Android

PR #1163, opened 2023-02-16
ci(shared): bump vendored ci files

galoybot:
# Bump Shared Tasks

This PR syncs in this repository, shared CI tasks from [concourse-shared](https://github.com/GaloyMoney/concourse-shared).


daviroo:
I've spoken to @krtk6160 about this and we should ignore this PR for now.

Issue #873, opened 2023-01-14
show username/BBW address in settings


nicolasburtey:
probably below phone.

Ayush0054:
hi can i take this issue

nicolasburtey:
> hi can i take this issue

let me know if you have any questions on this. I would start by making sure storybook works, and work with storybook on this.

Ayush0054:
![image](https://user-images.githubusercontent.com/97244608/219442735-34f3cf83-7983-4e3f-926b-e8c28c2ce55c.png)

feedback please

nicolasburtey:
Looks good to me. How would it look like when it's not authed?

nicolasburtey:
Feel free to post a PR if you want feedback on the code

Ayush0054:
![image](https://user-images.githubusercontent.com/97244608/219665479-5271a02f-5136-47f2-823c-294580311b4c.png)
![image](https://user-images.githubusercontent.com/97244608/219665514-b53abb55-863b-40bd-b9d1-18b553f82d14.png)

auth disabled
![image](https://user-images.githubusercontent.com/97244608/219665751-54ac2afb-2bb4-4e49-b023-4857f8ff7771.png)
![image](https://user-images.githubusercontent.com/97244608/219665885-e82b6c2c-367e-4db5-a830-18b28ee640ed.png)

Ayush0054:
![image](https://user-images.githubusercontent.com/97244608/219666289-d249c1bd-a965-493d-b22d-2a10f1405a7c.png)
![image](https://user-images.githubusercontent.com/97244608/219666351-56831d2c-7e8d-4a76-a9e8-ff17513534f8.png)
when not authed
![image](https://user-images.githubusercontent.com/97244608/219666421-91a58e17-18ce-4e59-b248-66c4a0da8bd0.png)
![image](https://user-images.githubusercontent.com/97244608/219666469-372a5a20-4619-42ef-9806-bf327683532d.png)

Ayush0054:
@nicolasburtey  do i need to remove changes like in eslint  , i18n files (due to yarn update-translation) while making pr

PR #1169, opened 2023-02-17
Feature: added username in settings 

Ayush0054:
fixed issue no #873 [show username/BBW address in settings]


android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Wed Apr 29 14:45:34 PDT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-all.zip
```
nicolasburtey, line 4: we probably don't want to upgrade gradle for now. this may break our CI/CD pipeline. can you revert this file?

package.json
```diff
@@ -82,6 +82,7 @@
     "lodash.merge": "^4.6.2",
     "moment": "^2.29.3",
     "node-libs-react-native": "^1.2.1",
+    "pod-install": "^0.1.38",
```
nicolasburtey, line 85: this should not be needed? what platform are you using?

in any case, if needed, it should be a dev-dependancies not a dependancy

package.json
```diff
@@ -94,6 +95,7 @@
     "react-native-extended-stylesheet": "^0.12.0",
     "react-native-fingerprint-scanner": "git+https://github.com/hieuvp/react-native-fingerprint-scanner.git#9cecc0db326471c571553ea85f7c016fee2f803d",
     "react-native-gesture-handler": "^2.9.0",
+    "react-native-gradle-plugin": "^0.71.15",
```
nicolasburtey, line 98: this should not be needed? what platform are you using?

in any case, if needed, it should be a dev-dependancies not a dependancy

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -162,6 +163,15 @@ export const SettingsScreenJSX: React.FC<SettingsScreenProps> = (params) => {
       enabled: !isAuthed,
       greyed: isAuthed,
     },
+    {
+      category: LL.common.username(),
+      icon: "person",
+      id: "username",
+      subTitleDefaultValue: LL.SettingsScreen.tapUserName(),
+      subTitleText: username,
```
nicolasburtey, line 171: this will be undefined if the `username` is not set. instead we want the button to still be enabled and send to the set username . I think all the logic should be in `addressScreen`

nicolasburtey (commented):
Some comments: 
- some environment/setup files should not be committed here
- need to think more about the logic when username is undefined

android/gradle/wrapper/gradle-wrapper.properties
Ayush0054: ok i will do it

package.json
Ayush0054: ok i will remove it

app/screens/settings-screen/settings-screen.tsx
Ayush0054: do i need to make a screen just like addressScreen in which i can edit my username  or just need to navigate to adressScreen on clicking username

Ayush0054: hey i added these line
   action: () => navigation.navigate("addressScreen"),
      enabled: isAuthed && username === undefined,
      greyed: !isAuthed || username !== undefined,

package.json
Ayush0054: removed this dependency

nicolasburtey:
@Ayush0054 the yarn.lock file should also not be part of the pull request. 

doesn't really matter for this PR but for your information, using git rebase can help make the git commit history cleaner: https://about.gitlab.com/blog/2020/11/23/keep-git-history-clean-with-interactive-rebase/

nicolasburtey:
@Ayush0054 yarn.lock should not be deleted. there should no changes to it

Ayush0054:
Ok i will fix this

Ayush0054:
i have deleted this repo and re forked it , since it was too messy to fix issue

PR #1175, opened 2023-02-20
feat: adding whatsapp login method

nicolasburtey:
- integrating Whatsapp as an alternative to SMS
- creating stories for the phone screen (input and validation)
- splitting tsx into 2 to keep 1 component per screen
- adding a set of stories. 
-- one for the seconds screen
-- a "flow" that starts with the first screen and go to the next when phone/catcha is validated
- done some quirky mocking of geetest within the hook. this could be improve in another PR.


app/i18n/en/index.ts
```diff
@@ -705,14 +704,15 @@ const en: BaseTranslation = {
...
     learnToEarn: "Learn to Earn",
   },
-  WelcomePhoneInputScreen: {
+  PhoneInputScreen: {
     header: "Enter your phone number, and we'll text you an access code.",
     headerVerify: "Verify you are human",
     placeholder: "Phone Number",
     verify: "Click to Verify",
-    continue: "Continue",
+    sms: "send via SMS",
```
UncleSamtoshi, line 712: Should these be capitalized?

UncleSamtoshi (approved):
One nitpick otherwise looks good to me.

app/i18n/en/index.ts
nicolasburtey: done

Extheoisah (approved):

PR #1184, opened 2023-02-21
fix: Fixed icon issues

daviroo:


nicolasburtey (changes requested):
just add the relevant one instead of the including all. this increase the build size by multiple MB otherwise.

daviroo:
> just add the relevant one instead of the including all. this increase the build size by multiple MB otherwise.

Because the component library we use is using this icon library I think the only future proof way forward is to add them all back in.  Even if we searched their entire codebase now for what icons they are using there's nothing stopping them from using an icon from a different set in a future release.

nicolasburtey:
> > 
> 
> > > just add the relevant one instead of the including all. this increase the build size by multiple MB otherwise.
> > 
> > 
> > Because the component library we use is using this icon library I think the only future proof way forward is to add them all back in. Even if we searched their entire codebase now for what icons they are using there's nothing stopping them from using an icon from a different set in a future release.
> 
> then we should look for another component library. but it feels the development has stopped on this library anyhow.
> 
> this doesn't justify the non negligeable bundle size overhead.

actually no the development hasn't really stopped. it seem the v4.0 will eventually comes out. 

we'll have to be careful if we update this library then

daviroo:
Closing to go in favour of a workaround where we take the risk that some components might not work properly when imported from react native elements.

PR #1201, opened 2023-02-21
chore: receive money refactor

UncleSamtoshi:
- introduce InvoiceDetails and Invoice objects
- create useReceiveBitcoin hook


__tests__/invoices/helpers.ts
```diff
@@ -0,0 +1,77 @@
...
+  "lnbc1p3l2qmfpp5t2ne20k97f3n24el9a792fte4q6n7jqr6x8qjjnklgktrdvpqq2sdqqcqzpuxqyz5vqsp5n23d3as4jxvpaemnsnvyynlpsg6pzsmxhn3tcwxealcyh6566nys9qyyssqce802uft9d44llekxqedzufkeaq7anldzpf64s4hmskwd9h5ppe4xrgq4dpq8rc3ph048066wgexjtgw4fs8032xwuazw9kdjcq8ujgpdk07ht"
+export const btcAmountInvoice =
+  "lnbc23690n1p3l2qugpp5jeflfqjpxhe0hg3tzttc325j5l6czs9vq9zqx5edpt0yf7k6cypsdqqcqzpuxqyz5vqsp5lteanmnwddszwut839etrgjenfr3dv5tnvz2d2ww2mvggq7zn46q9qyyssqzcz0rvt7r30q7jul79xqqwpr4k2e8mgd23fkjm422sdgpndwql93d4wh3lap9yfwahue9n7ju80ynkqly0lrqqd2978dr8srkrlrjvcq2v5s6k"
+export const mockOnChainAddress = "tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx"
+
+const mockLnInvoice = createMock<LnInvoice>({
+  paymentRequest: btcAmountInvoice,
+})
+
+const mockLnInvoiceCreate = jest.fn().mockResolvedValue({
```
nicolasburtey, line 17: have you tried using `MockedProvider` instead?

app/screens/receive-bitcoin-screen/invoices/index.types.ts
```diff
@@ -0,0 +1,103 @@
...
+export type Invoice = {
+  expiration?: Date
+  getFullUri: GetFullUriFn
+  invoiceDisplay: string // what is displayed to the user
+  invoiceData: LightningInvoiceData | OnChainInvoiceData | undefined
+}
+
+export type GetFullUriFn = (params: { uppercase?: boolean; prefix?: boolean }) => string
+
+export type InvoiceDetails<V extends WalletCurrency> = {
```
nicolasburtey, line 30: why not `W` instead of `V`?

app/screens/receive-bitcoin-screen/receive-btc.tsx
```diff
@@ -453,53 +253,69 @@ const ReceiveBtc = () => {
...
+    const validAmount = Boolean(invoiceDetails.unitOfAccountAmount.amount)
 
     return (
       <View style={[styles.inputForm, styles.container]}>
         <View style={styles.currencyInputContainer}>
           <View style={styles.currencyInput}>
-            {amountCurrency === "BTC" && (
+            {unitOfAccountAmount.currency === WalletCurrency.Btc && (
               <>
                 <FakeCurrencyInput
```
nicolasburtey, line 316: why do we have this name for the input?

app/screens/receive-bitcoin-screen/use-receive-bitcoin.ts
```diff
@@ -0,0 +1,249 @@
...
+        paymentHash
+        paymentRequest
+        paymentSecret
+        satoshis
+      }
+    }
+  }
+`
+
+export const useReceiveBitcoin = ({
```
nicolasburtey, line 76: when I see `useReceiveBitcoin`, I'm expecting it's a bitcoin only wallet flow. I think this hook can also be used to receive stablesats, so maybe a better name would just be `useReceive`, or `useReceiveFunds`. something along those lines that will convey that it's not only for bitcoin wallets

__tests__/invoices/helpers.ts
```diff
@@ -0,0 +1,77 @@
+import { LnInvoice } from "@app/graphql/generated"
```
nicolasburtey, line 1: not sure having the helper.ts/test fixtures separate from the tests to be useful. I would suggest to put them in the same file?

__tests__/invoices/invoice-details.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+    })
+
+    it("creates an onchain address", async () => {
+      const btcOnchainParams: CreateInvoiceDetailsParams<typeof WalletCurrency.Btc> = {
+        ...btcWalletParams,
+        invoiceType: InvoiceType.OnChain,
+        unitOfAccountAmount: { amount: 1, currency: WalletCurrency.Usd },
+      }
+
+      const invoiceDetails = createInvoiceDetails(btcOnchainParams)
```
nicolasburtey, line 171: is `createInvoiceDetails` a good fn name if it's also use for onchain?

__tests__/invoices/invoice-details.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      expect(invoice).toEqual(
+        expect.objectContaining({
+          invoice: expect.objectContaining({
+            invoiceDisplay: btcAmountInvoice,
+          }),
+        }),
+      )
+    })
+
+    it("creates an onchain address", async () => {
```
nicolasburtey, line 164:
```suggestion
    it("creates an btc onchain address", async () => {
```

we will probably have usd as well soon enough.

__tests__/invoices/invoice-details.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+      const btcOnchainParams: CreateInvoiceDetailsParams<typeof WalletCurrency.Btc> = {
+        ...btcWalletParams,
+        invoiceType: InvoiceType.OnChain,
+        unitOfAccountAmount: { amount: 1, currency: WalletCurrency.Usd },
```
nicolasburtey, line 168: we should test when no amount is needed as well for onchain

__tests__/invoices/helpers.ts
UncleSamtoshi: I believe mockProvider would only work in a situation where you are rendering react components. For these tests, there is no react, it is just simple js unit tests.

app/screens/receive-bitcoin-screen/invoices/index.types.ts
```diff
@@ -0,0 +1,103 @@
...
+  unitOfAccountAmount?: PaymentAmount<WalletCurrency> // amount the user wants to receive in the currency that they think in
```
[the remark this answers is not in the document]
UncleSamtoshi: Maybe I should have a better comment description. It is not always display currency. Say you wanted to receive 1000 sats, 1000 sats would be the unitOfAccountAmount but there would be a separate display currency amount in something like dollars.

app/screens/receive-bitcoin-screen/invoices/index.types.ts
UncleSamtoshi: I can make this change

app/screens/receive-bitcoin-screen/receive-btc.tsx
UncleSamtoshi: Its a poorly named component from a library that we use that we want to get rid of. `Fake` in this context means that it is not a real React TextInput component but a simulated or fake TextInput component.

app/screens/receive-bitcoin-screen/use-receive-bitcoin.ts
UncleSamtoshi: Sounds good!

__tests__/invoices/helpers.ts
UncleSamtoshi: I can move them to be in the same file.

__tests__/invoices/invoice-details.spec.ts
UncleSamtoshi: Yeah I struggled finding the terminology for it, lets discuss it in the mobile sync. I thought potentially it made sense because invoice could be a catch all for either an address or a payment request.

__tests__/invoices/invoice-details.spec.ts
UncleSamtoshi: Can do

__tests__/invoices/invoice-details.spec.ts
UncleSamtoshi: `btc` is already in the name in the enclosing `describe` function.

PR #1236, opened 2023-02-22
chore: adding back DMSans-Bold.ttf used in SVGs

nicolasburtey:


app/rne-theme/theme.ts
```diff
@@ -38,7 +38,9 @@ const theme = createTheme({
     Text: (props, theme) => {
       const universalStyle = {
         color: props.color || theme.colors.grey5,
-        fontFamily: "SourceSansPro",
```
UncleSamtoshi, line 41: Why was this removed?

UncleSamtoshi (changes requested):

app/rne-theme/theme.ts
nicolasburtey: red screen of death complaining the font doesn't exist. tried adding `-Regular` but got same message.

nicolasburtey: here is the issue, with the previous code: 

```
        fontWeight: props.bold ? "600" : "400",
        fontFamily: "SourceSansPro-Regular",
```

image

nicolasburtey: > here is the issue, with the previous code:
> 
> ```
>         fontWeight: props.bold ? "600" : "400",
>         fontFamily: "SourceSansPro-Regular",
> ```
> 
> image

@UncleSamtoshi if it's not been used in the app but only storybook I propose you to approve so we can unlock the release and can dive into this whenever we use the component

UncleSamtoshi: Works for me

UncleSamtoshi (approved):

PR #1247, opened 2023-02-23
chore: first version of storybook for receive screen

nicolasburtey:
- adding storybook for receive screen. only query on loaded up been mocked right now. 
- removing the backward compatibility with pre-USD wallet (when I see them). I don't think we want to maintain the complexity of having BTC only wallet going forward. it adds boilerplate. if someone wants a BTC only wallet then they can fork and adapt the code.


app/screens/receive-bitcoin-screen/receive-btc.tsx
```diff
@@ -311,6 +326,12 @@ const ReceiveBtc = () => {
   if (showAmountInput && unitOfAccountAmount && btcAmount && usdAmount) {
     const validAmount = Boolean(paymentRequestDetails.unitOfAccountAmount.amount)
 
+    console.log({
```
UncleSamtoshi, line 329: lingering console.log

UncleSamtoshi (changes requested):

UncleSamtoshi (approved):

PR #1248, opened 2023-02-23
feat: Improved address display in settings screen

### LINKED ISSUE #1185: Look again at username on setting screen
A pr was submitted by an external contributor.  Things we would probably want to change:
- tap to copy username
- display username in address format (e.g. user@bbw.sv@)
https://github.com/GaloyMoney/galoy-mobile/pull/1172

daviroo:
Fixes #1185 

image


Extheoisah:
> > LGTM. Just one thing. How do users know this button enables them to copy their username? Because its currently greyed out and looks like its a button that's not clickable/pressable
> 
> It's really just there so that any user who tries to highlight the text to copy it will have the ability to just copy it. It was a suggestion from @agbegin on one of the mobile sync calls.

Sounds good

Extheoisah (approved):

PR #1255, opened 2023-02-24
fix: set username issue in address screen

Extheoisah:
This fix ensures that when the user sets his/her username, we query for an update is that when the user presses the back or home button, the `set address` modal is not rendered as we currently have it.

### DIFF
--- a/app/screens/galoy-address-screen/galoy-address-screen.tsx
+++ b/app/screens/galoy-address-screen/galoy-address-screen.tsx
@@ -92,7 +92,10 @@ gql`
 export const GaloyAddressScreen = () => {
   const { LL } = useI18nContext()
   const isAuthed = useIsAuthed()
-  const { data } = useAddressScreenQuery({ fetchPolicy: "cache-first", skip: !isAuthed })
+  const { data } = useAddressScreenQuery({
+    fetchPolicy: "cache-and-network",
+    skip: !isAuthed,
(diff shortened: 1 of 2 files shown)


app/screens/galoy-address-screen/galoy-address-screen.tsx
```diff
@@ -92,7 +92,10 @@ gql`
 export const GaloyAddressScreen = () => {
   const { LL } = useI18nContext()
   const isAuthed = useIsAuthed()
-  const { data } = useAddressScreenQuery({ fetchPolicy: "cache-first", skip: !isAuthed })
+  const { data } = useAddressScreenQuery({
+    fetchPolicy: "cache-and-network",
```
daviroo, line 96: If we refetch the query do we need this change?

nicolasburtey: probably not a bad idea to avoid potential stale data

Extheoisah: Oops! It was the first option I tried. forgot to delete it. I've updated it now

Extheoisah: we don't need `cache-and-network` since we refetch the query

daviroo (approved):

PR #1256, opened 2023-02-24
chore: send confirmations screen display currency aware

nicolasburtey:


app/utils/currencyConversion.ts
```diff
@@ -29,37 +29,6 @@ export const textToCurrency = (
...
-    if (paymentAmount.amount === 1) {
-      return "1 sat"
-    }
-    return paymentAmountToText(paymentAmount) + " sats"
-  }
-
-  return "$" + paymentAmountToText(paymentAmount)
-}
-
-export const paymentAmountToText = (
```
UncleSamtoshi, line 45: It may be easier for unit testing if you keep these functions outside of a hook, and then just make a call to it from within the hook.

UncleSamtoshi (changes requested):

app/utils/currencyConversion.ts
nicolasburtey: I don't know. not sure if it's ideal to extract the function just for testing because it create a level of indirection, and it won't really test the hook for things like caching/potential cache been stale for instance.

a better approach would be imo to use https://react-hooks-testing-library.com/ for more complex hooks.

UncleSamtoshi: I think a combination of the two would probably be good. Use the hook testing library to ensure that renders are happening properly but use functional unit tests for more of the logic of how the underlying functions work. It seems similar to our backend where a hook is almost our application-layer/imperative-shell that makes calls to a functional-core/domain.

UncleSamtoshi: Although we haven't formalized or enforced that mental model at all.

UncleSamtoshi (approved):

app/utils/currencyConversion.ts
nicolasburtey: Yup that make sense. We could try to formalize it eventually but should do some tests/example first to get a better feeling of the development flow

PR #1260, opened 2023-02-24
chore: simplify username in settings row

Extheoisah:


app/screens/settings-screen/settings-screen.tsx
```diff
@@ -137,6 +140,27 @@ export const SettingsScreen: React.FC = () => {
...
+          message: (translations) =>
+            translations.GaloyAddressScreen.copiedAddressToClipboard({
+              bankName,
+            }),
+          type: "success",
+          currentTranslation: LL,
+        })
+      },
+      enabled: isAuthed,
+      greyed: true,
```
nicolasburtey, line 161: why is greyed always true? shouldn't it only be grey when it's already set?

nicolasburtey: I know you're copying this around, but this doens't make much sense to me to grey it out if the button is a navigation action.

nicolasburtey: oh also why was the action to set the username removed?

Extheoisah: I agree we should make the button grey. I'll fix that.
But we don't show this button if the user hasn't set the username, we only show it if they're logged in and when a username is present. They set the username from the next button `Receive payment using your bbw address`. The action to set username was not for this button in the first place.

nicolasburtey: > But we don't show this button if the user hasn't set the username, we only show it if they're logged in and when a username is present. 

I think we should show it back. it's very intention to have a clear place to set your username. it's a key part of the onboarding, and currently it's otherwise "hidden" in a "merchants" feature

Extheoisah: > > But we don't show this button if the user hasn't set the username, we only show it if they're logged in and when a username is present.
> 
> I think we should show it back. it's very intentional to have a clear place to set your username. it's a key part of the onboarding, and currently it's otherwise "hidden" in a "merchants" feature

Yeah, that makes sense. I'll work on it.

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -141,13 +144,38 @@ export const SettingsScreen: React.FC = () => {
...
+            translations.GaloyAddressScreen.copiedAddressToClipboard({
+              bankName,
+            }),
+          type: "success",
+          currentTranslation: LL,
+        })
+      },
+      enabled: isAuthed,
+      greyed: !isAuthed,
+      hidden: !isAuthed,
```
nicolasburtey, line 170:
```suggestion
```

nicolasburtey: I don't think we should have this hidden?

Extheoisah: fixed

nicolasburtey (approved):

PR #1265, opened 2023-02-24
feat: display currency in receive flow

UncleSamtoshi:


nicolasburtey (commented):
some first thoughts skimming through the PR.

maybe we should merge `useDisplayCurrency`, `usePriceConversion` in a single hook?

we also have `useUsdBtcAmount`

seems a lot of the logics is shared/overlapping here. without clear understand of what is depending on what.

UncleSamtoshi:
Was thinking the same things that `useDisplayCurrency` and `usePriceConversion` should be merged because its essential that the display currency stays in sync with the price feed.

> we also have useUsdBtcAmount

Yup this should go away.

nicolasburtey:
Ok. I think you have more context than I do so I'll let you do those changes, unless you want me to look into it

nicolasburtey (approved):

PR #1266, opened 2023-02-25
chore: moving fr ca to fr and pt br to pt

nicolasburtey:
- also fixing italian language


Extheoisah (approved):
LGTM. works okay locally too.

PR #1267, opened 2023-02-25
feat: expose display currency (beta)

nicolasburtey:


Extheoisah:
LGTM... Just some observations, we can tackle in another PR ->
1. when you enable beta in debug mode and change your currency to say Euro, and you go back to disable beta features, we should automatically reset the app to the default features. Currently, even after changing beta to false, if I don't change my currency manually to dollars, it remains in Euro.
2. we might want to have the display currency in the price graph as users check this page a lot.

I'll approve nonetheless.

Extheoisah (approved):

nicolasburtey:
1. yup I know but people who will use this feature are probably only the team so they can understand what's going on
2. yup, but the api is still not exposing price history for non USD

PR #1270, opened 2023-02-25
chore: iteration on real time price fetching

nicolasburtey:


app/hooks/use-realtime-price.ts
```diff
@@ -0,0 +1,22 @@
+import { WatchQueryFetchPolicy } from "@apollo/client"
+import { useDisplayCurrencyQuery, useRealtimePriceQuery } from "@app/graphql/generated"
+import { useIsAuthed } from "@app/graphql/is-authed-context"
+
+export const useRealtimePriceWrapper = (
+  { fetchPolicy }: { fetchPolicy: WatchQueryFetchPolicy } = {
+    fetchPolicy: "network-only",
```
UncleSamtoshi, line 7: I don't think the default here should be `network-only` as this hook will render multiple times on every single screen and if we ever use `network-only` itll result in a ton of network calls.

app/graphql/cache.ts
```diff
@@ -19,8 +21,8 @@ gql`
     }
   }
 
-  query realtimePrice {
-    realtimePrice {
+  query realtimePrice($currency: DisplayCurrency!) {
+    realtimePrice(currency: $currency) {
```
UncleSamtoshi, line 25: Will we ever use a price feed without passing in a display currency? If not, we should talk to the backend about whether or not even allowing the query to not take a display currency should be a possibillity. The alternative to passing in the display currency, is using the pricefeed as our source of truth for the display currency rather than using the display currency query.

UncleSamtoshi (changes requested):

app/graphql/cache.ts
nicolasburtey: Price feed is default USD in the backend. Can use useful for galoy-pay for instance.

For the mobile I don't think we want to use the default, we always want to pass a value

app/hooks/use-realtime-price.ts
nicolasburtey: Maybe there should be no default?

It's used in 4 places, and 3 of them should be network only. The one which is inside another hook and can be called with high frequency is cache-first. So logically this should be good but maybe this is not obvious from the function signature. Maybe I can have no default and let the caller explicitly say which network policy to apply?

nicolasburtey:
superseeded by https://github.com/GaloyMoney/galoy-mobile/pull/1275

PR #1275, opened 2023-02-26
chore: more display currency

nicolasburtey:
- passing txid instead of full tx between screens with transactions info. will make it much easier to do mocking.
- introducing the use of `useFragment_experimental` to get information from transaction within screen
- removing currency.js, using Intl package (already present) instead
- making txLimitScreen, conversionScreen and homeScreen and dependent component (balanceHeader, walletSummary) display currency aware
- moving some values from usdBalance to displayBalance across send screens (work still in progress, still ok to merge as it should be a no-op for when displayCurrency = USD)
- removing useless routes param for conversion screen
- exposing `displayBalance` both from UsdWallet and BtcWallet
- creating a useRealtimePriceWrapper
- removed useReactiveVar for prefCurrency. we'll be using the displayCurrency now instead


app/components/transaction-item/transaction-item.tsx
```diff
@@ -110,28 +92,55 @@ const amountDisplayStyle = ({
...
+  const isPending = tx.status === "PENDING"
+  const description = descriptionDisplay(tx)
+  const walletCurrency = tx.settlementCurrency as WalletCurrency
   const pressTxAmount = () => setTxHideBalance((prev) => !prev)
 
+  const amountWithCurrency = moneyAmountToTextWithUnits({
+    amount: tx.settlementAmount,
+
+    // FIXME: will be the wrong number if non USD currency
+    currency: tx.settlementCurrency,
```
UncleSamtoshi, line 141: Why will this be the wrong number?

app/hooks/use-display-currency.ts
```diff
@@ -51,6 +53,13 @@ export const useDisplayCurrency = () => {
     [displayCurrency],
   )
 
+  const formatToUsd = useCallback((amount: number) => {
```
UncleSamtoshi, line 56: What is the reason for including this in a hook? It has no dependencies and could a pure utility function.

app/hooks/use-display-currency.ts
```diff
@@ -81,12 +90,23 @@ export const useDisplayCurrency = () => {
     [fiatSymbol],
   )
 
+  // TODO: remove
+  const computeUsdAmount = (tx: Transaction) => {
```
UncleSamtoshi, line 94: This does not seem to belong to in `useDisplayCurrency`. It will also not always be the UsdAmount. It is actually computing the amount in display currency.

app/hooks/use-realtime-price.ts
```diff
@@ -0,0 +1,25 @@
+import { WatchQueryFetchPolicy } from "@apollo/client"
+import { useDisplayCurrencyQuery, useRealtimePriceQuery } from "@app/graphql/generated"
+import { useIsAuthed } from "@app/graphql/is-authed-context"
+
+export const useRealtimePriceWrapper = (
+  { fetchPolicy }: { fetchPolicy: WatchQueryFetchPolicy } = {
+    fetchPolicy: "network-only",
```
UncleSamtoshi, line 7: I dont think "network-only" is the desired fetch policy here. I believe this will cause a network request to be made on each rerender. You could instead use a lazy query or just refetch when desired.

app/screens/conversion-flow/conversion-details-screen.tsx
```diff
@@ -349,16 +369,34 @@ export const ConversionDetailsScreen = ({
...
+                    convertMoneyAmount
+                      ? moneyAmountToMajorUnitOrSats(
+                          convertMoneyAmount(usdAmount, DisplayCurrency),
+                        )
+                      : NaN
+                  }
+                  prefix={fiatSymbol}
+                  delimiter=","
+                  separator="."
+                  precision={2}
```
UncleSamtoshi, line 394: Precision should come from the `useDisplayCurrency`.

UncleSamtoshi (changes requested):

app/components/transaction-item/transaction-item.tsx
nicolasburtey: I want to say `if non WalletCurrency`. I'm assuming settlementCurrency in the WalletCurrency.

app/hooks/use-display-currency.ts
nicolasburtey: I put it there as something some query depends on that should be refactored. it could be living outside of it but if we put all the functions that will be in charged of either do a conversion or displaying a currency to a string, then it's easier to understand the requirement.

app/hooks/use-display-currency.ts
nicolasburtey: same comment as here: https://github.com/GaloyMoney/galoy-mobile/pull/1275#discussion_r1119101823

app/hooks/use-realtime-price.ts
nicolasburtey: earlier comment on this here: https://github.com/GaloyMoney/galoy-mobile/pull/1270#discussion_r1117984415

app/graphql/cache.ts
```diff
@@ -190,5 +204,42 @@ export const createCache = () =>
           },
         },
       },
+      UsdWallet: {
```
UncleSamtoshi, line 207: Line 188 will have a bug if the display currency does not have two decimals. We need to use the decimal places of the display currency. Same for line 239.

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -173,10 +173,7 @@ const SendBitcoinDestinationScreen = ({
     skip: !useIsAuthed(),
   })
 
-  useRealtimePriceQuery({
-    fetchPolicy: "network-only",
-    skip: !useIsAuthed(),
-  })
+  useRealtimePriceWrapper()
```
UncleSamtoshi, line 176: Can you check to see if this makes a network request with each letter you type?

nicolasburtey: sure let me check

app/graphql/cache.ts
nicolasburtey: this `minorUnitToMajorUnitOffset`?

app/screens/receive-bitcoin-screen/receive-wrapper.tsx
```diff
@@ -87,10 +84,7 @@ const ReceiveWrapperScreen = () => {
     skip: !isAuthed,
   })
 
-  useRealtimePriceQuery({
-    fetchPolicy: "network-only",
-    skip: !useIsAuthed(),
-  })
+  useRealtimePriceWrapper()
```
UncleSamtoshi, line 87: Do you want this to refresh every time a switch between btc and usd happends?

app/graphql/cache.ts
UncleSamtoshi: Yes

app/components/transaction-item/transaction-item.tsx
UncleSamtoshi: Yeah I think they should always be the same

app/screens/conversion-flow/conversion-details-screen.tsx
UncleSamtoshi: yes

nicolasburtey: ok. I'll fix that once it's available from the gql api. because it's in the cache so there is limited information available and don't want to import a hooks here (not sure I even can)

nicolasburtey: done

app/graphql/cache.ts
nicolasburtey: have added a TODO for it

app/hooks/use-realtime-price.ts
nicolasburtey: only if the component unmount and remount. ie screen closed and reopen

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
nicolasburtey: nope not seen such issue

app/screens/receive-bitcoin-screen/receive-wrapper.tsx
nicolasburtey: > Do you want this to refresh every time a switch between btc and usd happends?

is that a suggestion or a question is that is currently the case?

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
UncleSamtoshi: Sounds good

app/hooks/use-realtime-price.ts
UncleSamtoshi: I misunderstood how this hook worked. You can ignore my comments.

PR #1291, opened 2023-02-28
feat: enable send flow display currency

UncleSamtoshi:
- create MoneyAmountInput component
- modify PaymentDetails types to handle WalletOrDisplayCurrency
- update send flow to use display currency

### DIFF
--- a/app/components/money-amount-input/index.tsx
+++ b/app/components/money-amount-input/index.tsx
@@ -1,0 +1,69 @@
+import { WalletCurrency } from "@app/graphql/generated"
+import { useDisplayCurrency } from "@app/hooks/use-display-currency"
+import { DisplayCurrency, MoneyAmount, WalletOrDisplayCurrency } from "@app/types/amounts"
+import { StyleProp, TextStyle } from "react-native"
+import { FakeCurrencyInput } from "react-native-currency-input"
+
+type MoneyAmountPaymentInputProps = {
+  moneyAmount: MoneyAmount<WalletOrDisplayCurrency>
+  setAmount?: (moneyAmount: MoneyAmount<WalletOrDisplayCurrency>) => void
+  style?: StyleProp<TextStyle>
+  editable?: boolean
+}
+
+export const MoneyAmountInput = ({
+  moneyAmount,
+  setAmount,
+  style,
+  editable,
+  ...props
+}: MoneyAmountPaymentInputProps) => {
+  const {
+    moneyAmountToMajorUnitOrSats,
+    amountInMajorUnitOrSatsToMoneyAmount,
+    minorUnitToMajorUnitOffset,
+    fiatSymbol,
+  } = useDisplayCurrency()
+
+  let prefix: string | undefined
+  let suffix: string | undefined
+  let precision: number | undefined
+  switch (moneyAmount.currency) {
+    case WalletCurrency.Btc:
+      prefix = ""
+      suffix = moneyAmount.amount === 1 ? " sat" : " sats"
+      precision = 0
+      break
+    case WalletCurrency.Usd:
+      prefix = "$"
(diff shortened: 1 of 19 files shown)


app/components/money-amount-input/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  editable?: boolean
+}
+
+export const MoneyAmountInput = ({
+  moneyAmount,
+  setAmount,
+  style,
+  editable,
+  ...props
+}: MoneyAmountPaymentInputProps) => {
```
nicolasburtey, line 20: eventually there should be a dependency on language also. 

in European language, currency is after the number.

running this here gives a good explanation: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

```
> "123.456,79 €"
> "￥123,457"
> "1,23,000"
```

I don't think we should do this for now. just giving this for context.

nicolasburtey: actually right now we put sats the European way, and dollars/other fiat American way. not sure why this has emerged naturally. probably me been confused by the 2 standards.

UncleSamtoshi: I think since sats is not a symbol it made sense to put it after the currency

UncleSamtoshi:
A cool aspect of the recent refactoring is that I was able to easily implement that if the display currency is equal to the wallet currency it wont show both. So in the case of dollars being your display currency it will just show dollars for stablesats, whereas if you have euros as your display currency it will show euros and dollars.

app/hooks/use-display-currency.ts
```diff
@@ -60,34 +69,73 @@ export const useDisplayCurrency = () => {
...
+    },
+    [minorUnitToMajorUnitOffset],
+  )
 
-  const moneyAmountToTextWithUnits = useCallback(
+  const amountInMajorUnitOrSatsToMoneyAmount = useCallback(
+    (
+      amount: number,
+      currency: WalletOrDisplayCurrency,
+    ): MoneyAmount<WalletOrDisplayCurrency> => {
```
nicolasburtey, line 93: in the backend we have a rule to use an object argument when there is more than one input, ie

```suggestion
    ({amount, currency}: {
      amount: number,
      currency: WalletOrDisplayCurrency,
    }): MoneyAmount<WalletOrDisplayCurrency> => {
```

nicolasburtey: do we want to enforce this rule here? I think that make sense to have the same rules to have less context switch when working on both code base simultaneously

UncleSamtoshi: Yeah I think it makes sense to use that rule

nicolasburtey (commented):
some comments but looks good to me overall. does this fixes any bug to your knowledge or is it just a type improvement? are you intending to do more refactoring after that? ie merge useConversion and useDisplayCurrency hooks?

UncleSamtoshi:
> some comments but looks good to me overall. does this fixes any bug to your knowledge or is it just a type improvement? 

No bug fixes to my knowledge.

> are you intending to do more refactoring after that? ie merge useConversion and useDisplayCurrency hooks?

I can look at the conversion screen after this as well as cleanup some of the tangled hooks.

nicolasburtey:
> > some comments but looks good to me overall. does this fixes any bug to your knowledge or is it just a type improvement?
> 
> No bug fixes to my knowledge.
> 
> > are you intending to do more refactoring after that? ie merge useConversion and useDisplayCurrency hooks?
> 
> I can look at the conversion screen after this as well as cleanup some of the tangled hooks.

needs rebasing, probably following your onchain usd fix

nicolasburtey (approved):

Issue #1290, opened 2023-02-28
Unable to scan/paste bitcoin address when default wallet set to USD


agbegin:
**Describe the bug**
Bug found in iOS, v2.1.56 build 427
With default wallet set to USD, an error screen is thrown when trying to pay to a bitcoin address: 

**To Reproduce**

1. Set default wallet to USD (within merchant features in settings)
3. Click on Send
4. Scan or paste a bitcoin address 
5. See error

**Expected behavior**
Expect to continue with payment flow, with bitcoin account selected as the send from account if that's the only account i can send onchain payments from.

**Screenshots**
![IMG_2824](https://user-images.githubusercontent.com/90297599/221958397-d4da2666-0dd3-48f0-a242-747829f85064.PNG)

**Smartphone (please complete the following information):**

- Device: iphone 11
- OS: iOS 15.6.1
- Version:  Bug found in iOS, v2.1.56 build 427

nicolasburtey:
issue is here @UncleSamtoshi : https://github.com/GaloyMoney/galoy-mobile/blob/main/app/screens/send-bitcoin-screen/payment-details/onchain.ts#L40

related to the recent refactoring. wondering if there is an easy fix or if we should wait for doing the onchain send from stablesats

UncleSamtoshi:
@nicolasburtey I think there should be an easy fix. We should probably halt the most recent rollout till this fix is put through. The scope is anyone scanning an onchain address with usd as their default wallet.

Extheoisah:
closed by https://github.com/GaloyMoney/galoy-mobile/pull/1293

PR #1293, opened 2023-02-28
fix: prevent usd wallet for onchain send

UncleSamtoshi:
- use btcwallet as default for onchain payments
- alert user when selecting a usd wallet that it is not supported for onchain

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
@@ -9,15 +9,17 @@ import { useI18nContext } from "@app/i18n/i18n-react"
 import { RootStackParamList } from "@app/navigation/stack-param-lists"
 import { palette } from "@app/theme"
 import { DisplayCurrency } from "@app/types/amounts"
+import { WalletDescriptor } from "@app/types/wallets"
 import { satAmountDisplay } from "@app/utils/currencyConversion"
 import { Network as NetworkLibGaloy, fetchLnurlInvoice } from "@galoymoney/client"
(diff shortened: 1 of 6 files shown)


UncleSamtoshi:
Fixes issue #1290

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -364,6 +374,20 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
...
   }
 
+  const setSendingWalletDescriptor = (
+    walletDescriptor: WalletDescriptor<WalletCurrency>,
+  ) => {
+    if (
+      walletDescriptor.currency === WalletCurrency.Usd &&
+      paymentDetail.paymentType === PaymentType.Onchain
+    ) {
+      Alert.alert(LL.SendBitcoinScreen.walletDoesNotSupportOnchain())
```
nicolasburtey, line 384: when would that be triggered? as we're setting btc wallet if this is an onchain address above by default

UncleSamtoshi: It would be triggered by `click on wallet icon -> modal pops up allowing user to choose wallet -> user clicks on usd wallet -> alert is displayed`.

UncleSamtoshi: I figured this would help the user understand better what is going on rather than just not having the option to switch wallets.

nicolasburtey (approved):

PR #1294, opened 2023-02-28
chore: updating currency list screen design

nicolasburtey:


app/screens/settings-screen/language-screen.tsx
```diff
@@ -76,13 +76,17 @@ export const LanguageScreen: React.FC = () => {
               }
             }}
           >
+            <View style={styles.viewSelectedIcon}>
+              {(newLanguage === language && loading && <ActivityIndicator />) ||
+                (languageFromServer === language && !loading && (
+                  <Icon name="ios-checkmark-circle" size={18} color={palette.green} />
+                )) || (
+                  <Icon name="ios-checkmark-circle" size={18} color={palette.white} />
```
Extheoisah, line 84: This chain of "OR" statements is quite confusing. On what condition is the last icon displayed?

nicolasburtey: yeah let me refactor it. it's kind of a hack currently: because `color = palette.white`, basically the icon would not really appear on the screen as the screen background is white also.

Extheoisah (approved):

PR #1300, opened 2023-03-01
chore: fetch number of fraction for currency dynamically

nicolasburtey:


app/graphql/cache.ts
```diff
@@ -61,6 +63,33 @@ const getWallets = ({
...
+  const resCurrency = cache.readQuery<DisplayCurrencyQuery>({
+    query: DisplayCurrencyDocument,
+  })
+  const displayCurrency = resCurrency?.me?.defaultAccount?.displayCurrency
+
+  if (!displayCurrency) {
+    return { displayCurrency: null, fractionDigits: null }
+  }
+
+  const res2 = cache.readQuery<CurrencyListQuery>({
```
UncleSamtoshi, line 78: `res2` should be renamed to something like `resCurrencyList`

app/graphql/cache.ts
```diff
@@ -61,6 +63,33 @@ const getWallets = ({
...
+    query: CurrencyListDocument,
+    variables: { currency: displayCurrency },
+  })
+  const currencyList = res2?.currencyList ?? []
+  if (!res2?.currencyList) {
+    return { displayCurrency: null, fractionDigits: null }
+  }
+
+  const fractionDigits =
```
UncleSamtoshi, line 87: Does it make sense to have a default to 2  here, in other places if we can't find the value we return null. Something is wrong with the currency list if a value can't be found.

app/graphql/cache.ts
```diff
@@ -61,6 +63,33 @@ const getWallets = ({
...
+
+  if (!displayCurrency) {
+    return { displayCurrency: null, fractionDigits: null }
+  }
+
+  const res2 = cache.readQuery<CurrencyListQuery>({
+    query: CurrencyListDocument,
+    variables: { currency: displayCurrency },
+  })
+  const currencyList = res2?.currencyList ?? []
```
UncleSamtoshi, line 82:
```suggestion
  const currencyList = res2?.currencyList
  if (!currencyList) ....
```

app/hooks/use-display-currency.ts
```diff
@@ -53,6 +54,10 @@ export const useDisplayCurrency = () => {
     )
   }, [dataCurrencyList, displayCurrency])
 
+  const fractionDigits = useMemo(() => {
```
UncleSamtoshi, line 57: It takes longer to memoize this then it does to just have `const fractionDigits = displayCurrencyInfo.fractionDigits` which is a simple assignment vs a function call to useMemo every render where it has to check for equality of dependencies.

UncleSamtoshi (changes requested):

app/graphql/cache.ts
```diff
@@ -61,6 +63,35 @@ const getWallets = ({
...
+    variables: { currency: displayCurrency },
+  })
+  const currencyList = resCurrencyList?.currencyList
+
+  if (!currencyList) {
+    return { displayCurrency: null, fractionDigits: null }
+  }
+
+  const fractionDigits =
+    currencyList.find((currency) => currency.id === displayCurrency)?.fractionDigits ??
```
UncleSamtoshi, line 89: If fraction digits is 0 here it will be falsy and then the value will be null.

app/graphql/cache.ts
```diff
@@ -171,20 +202,16 @@ export const createCache = () =>
...
-              // could we have a way to not fetch DisplayCurrencyQuery?
-              const resCurrency = cache.readQuery<DisplayCurrencyQuery>({
-                query: DisplayCurrencyDocument,
-              })
-              const currency = resCurrency?.me?.defaultAccount?.displayCurrency
+              const { displayCurrency, fractionDigits } =
+                getDisplayCurrencyAndFractionDigits(cache)
 
-              if (!currency) {
+              if (!displayCurrency || !fractionDigits) {
```
UncleSamtoshi, line 208: FractionDigits could be 0 here so the check should `typeof fractionDigits !== "number"`

app/graphql/cache.ts
```diff
@@ -210,20 +235,16 @@ export const createCache = () =>
...
-              // could we have a way to not fetch DisplayCurrencyQuery?
-              const resCurrency = cache.readQuery<DisplayCurrencyQuery>({
-                query: DisplayCurrencyDocument,
-              })
-              const currency = resCurrency?.me?.defaultAccount.displayCurrency
+              const { displayCurrency, fractionDigits } =
+                getDisplayCurrencyAndFractionDigits(cache)
 
-              if (!currency) {
+              if (!displayCurrency || !fractionDigits) {
```
UncleSamtoshi, line 241: same here

UncleSamtoshi (changes requested):

app/graphql/cache.ts
UncleSamtoshi: You can just remove the `?? null`

nicolasburtey: not with ??:

```
% node
Welcome to Node.js v18.14.0.
Type ".help" for more information.
> 0 ?? "a"
0
> 0 || "a"
'a'
```

UncleSamtoshi: I thought it behaved the same as `||`. My mistake.

nicolasburtey: if I remove `?? null` then the return type of undefined. then I have to check both for `null` and `undefined` in the calling function, which I don't want to do

PR #1303, opened 2023-03-01
chore: clean up conversion display currency

UncleSamtoshi:
- remove useUsdBtc hook
- introduce useConvertMoneyDetails
- improve display currency in the conversion flow


app/screens/conversion-flow/conversion-confirmation-screen.tsx
```diff
@@ -269,7 +276,17 @@ export const ConversionConfirmationScreen: React.FC<Props> = ({ route }) => {
         <View style={styles.conversionInfoField}>
           <Text style={styles.conversionInfoFieldTitle}>{LL.common.rate()}</Text>
           <Text style={styles.conversionInfoFieldValue}>
-            ~ {formatMoneyAmount(usdPerBtc)} / 1 BTC
+            ~{" "}
+            {formatMoneyAmount(
+              convertMoneyAmount(
+                {
+                  amount: 1 * 100000000,
```
nicolasburtey, line 283:
```suggestion
                  amount: 1 * 100_000_000,
```

app/screens/conversion-flow/conversion-details-screen.tsx
```diff
@@ -52,52 +52,53 @@ export const ConversionDetailsScreen = () => {
...
+
   const btcWalletBalance = btcWallet?.balance ?? NaN
   const usdWalletBalance = usdWallet?.balance ?? NaN
 
   const usdWalletDisplayBalance = data?.me?.defaultAccount?.usdWallet?.displayBalance
 
-  const usdWalletDisplayBalanceText = usdWalletDisplayBalance
-    ? formatToDisplayCurrency(usdWalletDisplayBalance)
-    : ""
+  console.log("usdWalletDisplayBalance", usdWalletDisplayBalance)
```
nicolasburtey, line 96: dandling console.log

app/screens/conversion-flow/use-convert-money-details.ts
```diff
@@ -0,0 +1,95 @@
+import { DisplayCurrency, MoneyAmount, WalletOrDisplayCurrency } from "@app/types/amounts"
+import React from "react"
+import { usePriceConversion } from "../../hooks/use-price-conversion"
+import { WalletCurrency } from "@app/graphql/generated"
+
+type Wallet = {
+  balance: number
```
nicolasburtey, line 7: I'm wondering whether we should introduce branded type on the mobile end

UncleSamtoshi: For balance?

app/screens/conversion-flow/use-convert-money-details.ts
```diff
@@ -0,0 +1,95 @@
+import { DisplayCurrency, MoneyAmount, WalletOrDisplayCurrency } from "@app/types/amounts"
+import React from "react"
+import { usePriceConversion } from "../../hooks/use-price-conversion"
+import { WalletCurrency } from "@app/graphql/generated"
+
+type Wallet = {
```
nicolasburtey, line 6: shouldn't this be from generated.ts?

you can trim down the type as needed, something like `Pick<Wallet, ["id", "balance", "walletCurrency"]>` 

not sure about the syntax but you'll get the idea: referencing the typescript help understand that those contexts are tied together

app/screens/conversion-flow/use-convert-money-details.ts
```diff
@@ -0,0 +1,95 @@
...
+  balance: number
+  walletCurrency: WalletCurrency
+  id: string
+}
+
+export const useConvertMoneyDetails = () => {
+  const [wallets, _setWallets] = React.useState<{
+    fromWallet: Wallet
+    toWallet: Wallet
+  }>()
```
nicolasburtey, line 16: shouldn't this have a default?

app/screens/conversion-flow/use-convert-money-details.ts
UncleSamtoshi: Didn't realize you could do the pick. I will make that change.

app/screens/conversion-flow/use-convert-money-details.ts
```diff
@@ -0,0 +1,95 @@
...
+  const [moneyAmount, setMoneyAmount] = React.useState<
+    MoneyAmount<WalletOrDisplayCurrency>
+  >({ amount: 0, currency: DisplayCurrency })
+
+  const setWallets = (wallets: { fromWallet: Wallet; toWallet: Wallet }) => {
+    if (wallets.fromWallet.balance === 0) {
```
nicolasburtey, line 22: maybe a comment explaining the intent of this line can be helpful? I guess we want to have the fromWallet having the funds but this requires me to read the line several time to understand the idea.

what happens if both wallets have 0 balance?

UncleSamtoshi: I can add a comment. If both wallets have 0, it will still show them both but you wont be able to do anything with them.

app/screens/conversion-flow/use-convert-money-details.ts
UncleSamtoshi: It gets set right away as soon as the component calling the hook knows the wallets

app/screens/conversion-flow/use-convert-money-details.ts
```diff
@@ -0,0 +1,95 @@
...
+      convertMoneyAmount(
+        moneyAmount,
+        moneyAmount.currency === DisplayCurrency
+          ? fromWallet.walletCurrency
+          : DisplayCurrency,
+      ),
+    )
+  }
+
+  const toggleWallet = toWallet.balance
```
nicolasburtey, line 64:
```suggestion
  const toggleWallet = toWallet.balance > 0
```

maybe that would be more explicit about the intent?

app/screens/conversion-flow/use-convert-money-details.ts
UncleSamtoshi: I can add an initial argument that will try to set it, but it will still be potentially undefined.

nicolasburtey (commented):
some feedbacks. looks good overall.

app/screens/conversion-flow/use-convert-money-details.ts
nicolasburtey: yeah for instance. maybe those should be branded at the graphql layer. just speaking out loud here not a strong opinion on this

app/screens/conversion-flow/conversion-confirmation-screen.tsx
nicolasburtey: still have the issue today? a lot of things were not necessarily properly set up on ci. like we were using node12 not long enough. I would give it another try. I think it really helps on readability.

nicolasburtey: seems it was introduce in ES12

image

app/screens/conversion-flow/use-convert-money-details.ts
UncleSamtoshi: I think that would be awesome, but I'm not sure how that works. They are currently branded scalars coming from the graphql api and then converted to primitive types.

UncleSamtoshi:
Feedback addressed

nicolasburtey (approved):

PR #1309, opened 2023-03-02
chore: fixing some storybook regressions

nicolasburtey:


UncleSamtoshi (approved):
Did you look into why the android e2e tests were failing?

nicolasburtey:
> Did you look into why the android e2e tests were failing?

backend 403 error :/

maybe one we should look with @bodymindarts to look at now that the e2e tracing works

Issue #1304, opened 2023-03-02
choosing USD wallet when sending onchain crashes the app


Extheoisah:
**Describe the bug**
When you scan a BTC enchain address and change your sending wallet to USD, the app crashes.

**Expected behavior**
We should show the user a modal or toast message stating that unchain payments can only be done from the BTC wallet instead of crashing the app.

**Screenshots**

https://user-images.githubusercontent.com/60826700/222412835-64dedce6-fb1a-4c61-9d13-d5344efa9dd2.mov

**Smartphone (please complete the following information):**

- Device: iPhone XS Max
- OS: 16.3.1
- Version 2.1.59 build 433

PR #1311, opened 2023-03-02
fix: error crash for USD wallet onchain payment

Extheoisah:
fixes issue https://github.com/GaloyMoney/galoy-mobile/issues/1304


app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -385,15 +413,7 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
...
-                setPaymentDetail(
-                  (paymentDetail) =>
-                    paymentDetail &&
-                    paymentDetail.setSendingWalletDescriptor({
-                      id: wallet.id,
-                      currency: wallet.walletCurrency,
-                    }),
-                )
-                toggleModal()
+                chooseWallet(wallet as unknown as Wallet)
```
nicolasburtey, line 416: this type of casting is not good. you should find the root cause and fix the type instead of remove type safeguard.

nicolasburtey (dismissed):
type is important, we should not remove the type safety

UncleSamtoshi (approved):

PR #1312, opened 2023-03-03
Puravida main

phaoga54:
Marketplace functionality


nicolasburtey:
hi @phaoga54,

this PR is much too big for us to consider merging it. 

a preferred way would be to make small PR so we understand the different functionality you want to add and we can both test them and verify the code work well with the rest of the app. 

also it will be much easier to review if there is no merge but instead you rely on rebase. we only use rebase in our repository, the history tree is much cleaner as a result and easier to understand for anyone that wants to follow along. I believe rebase is the standard for most/all open source repositories nowadays.

phaoga54:
@nicolasburtey the reason for so much changes is because we are out sync with Galoy latest changes.

Probably will me much smaller PR when I finish the sync.

Will open another PR after

PR #1314, opened 2023-03-03
test: get tests running for component in js

nicolasburtey:
- wip to have the local stack running (need to be able to bypass captcha)
- working on home screen to be loaded from javascript only
- remove some component from home that didn't seems really relevant anymore (but still need to test this, see comment below)

TODO: more testing on how the app behave when the server is on and off (which is why I tried to get the backend working to easily been able to switch it on and off, but running into issue there)


Extheoisah (approved):
LGTM... quite a big one.
@UncleSamtoshi can also have a look

app/screens/phone-auth-screen/phone-input.tsx
```diff
@@ -332,3 +338,9 @@ export const PhoneInputScreen: React.FC = () => {
     </Screen>
   )
 }
+
+const phoneNumberIsLocalTestAccount = (phone: string) => {
+  // TODO: extract the list from https://github.com/GaloyMoney/galoy/blob/main/galoy.yaml#L38
+  if (phone === "+16505554321") return true
```
UncleSamtoshi, line 344: This is a real phone number and it will prevent them from signing in on prod (although very small chance they have an account). Its probably cleaner to bypass based on the GaloyInstace from config rather than the phone number.

UncleSamtoshi (changes requested):

__tests__/screens/helper.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { createStackNavigator } from "@react-navigation/stack"
+
+const Stack = createStackNavigator()
+
+export const Wrapping: React.FC<PropsWithChildren> = ({ children }) => (
+  <ThemeProvider theme={theme}>
+    <NavigationContainer>
+      <Stack.Navigator>
+        <Stack.Screen name="Home">
+          {() => (
```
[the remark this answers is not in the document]
nicolasburtey: don't know this is why `Stack.Screen` is expecting afaict

__tests__/screens/send-confirmation.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+    </Wrapping>,
+  )
+  // it seems we need multiple act because the component re-render multiple times
+  // probably this could be debug with why-did-you-render
+  await act(async () => {})
+  await act(async () => {})
+
+  // console.log(JSON.stringify(toJSON(), null, 2))
+  const { children } = await findByLabelText("Successful Fee")
+  expect(children).toEqual(["₦0.00 - $0.00"])
```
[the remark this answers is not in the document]
nicolasburtey: correct. I'm showing how this can be used eventually, this is not meant to be a "real test" that will catch bug. but eventually we should build on it.

app/screens/phone-auth-screen/phone-input.tsx
nicolasburtey: I think 555 are dedicated number for special use case (including for TV shows) so I think the assumption to get a user from that account is... null? 

https://en.wikipedia.org/wiki/555_(telephone_number)

nicolasburtey: I'm not sure to understand how you would do it otherwise? the thing is that if it's not a real number, it's been considered invalid by the component and we can't move to the next screen

app/screens/phone-auth-screen/phone-input.tsx
```diff
@@ -138,9 +138,15 @@ export const PhoneInputScreen: React.FC = () => {
 
   useEffect(() => {
     if (phoneNumber) {
+      if (phoneNumberIsLocalTestAccount(phoneNumber)) {
```
UncleSamtoshi, line 141: Could you do `if (galoyInstance.name === 'Local')` or am I missing something?

app/screens/phone-auth-screen/phone-input.tsx
UncleSamtoshi: See above.

UncleSamtoshi (changes requested):

app/screens/phone-auth-screen/phone-input.tsx
nicolasburtey: oh yup indeed. will do that.

UncleSamtoshi (approved):

PR #1316, opened 2023-03-03
feat: display currency in lnurlw

UncleSamtoshi:
- added in display currency
- deleted unused code for usd wallet use case as usd wallet is not supported until we can denominate usd invoices in sats

### DIFF
--- a/app/navigation/stack-param-lists.ts
+++ b/app/navigation/stack-param-lists.ts
@@ -7,6 +7,7 @@ import {
   PaymentDestination,
   ReceiveDestination,
 } from "@app/screens/send-bitcoin-screen/payment-destination/index.types"
+import { WalletDescriptor } from "@app/types/wallets"
 
 export type RootStackParamList = {
   getStarted: undefined
@@ -59,13 +60,13 @@ export type RootStackParamList = {
     domain: string
     k1: string
     defaultDescription: string
-    minWithdrawableSatoshis: number
-    maxWithdrawableSatoshis: number
+    minWithdrawableSatoshis: MoneyAmount<typeof WalletCurrency.Btc>
+    maxWithdrawableSatoshis: MoneyAmount<typeof WalletCurrency.Btc>
     receiveCurrency: WalletCurrency
-    walletId: string
-    satAmount: number
-    satAmountInUsd: number
-    amountCurrency: string
+    receivingWalletDescriptor: WalletDescriptor<typeof WalletCurrency.Btc>
+    unitOfAccountAmount: MoneyAmount<WalletOrDisplayCurrency>
+    settlementAmount: MoneyAmount<typeof WalletCurrency.Btc>
+    secondaryAmount: MoneyAmount<WalletOrDisplayCurrency> | undefined
   }
   phoneFlow: undefined
   transactionDetail: { txid: string }
--- a/app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
+++ b/app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
@@ -1,9 +1,7 @@
 import React, { useEffect, useState } from "react"
-import { Pressable, View } from "react-native"
-import { FakeCurrencyInput } from "react-native-currency-input"
+import { Pressable, View, ScrollView } from "react-native"
 import EStyleSheet from "react-native-extended-stylesheet"
 import { TouchableWithoutFeedback } from "react-native-gesture-handler"
-
 import SwitchIcon from "@app/assets/icons/switch.svg"
 import { useReceiveBtcQuery, WalletCurrency } from "@app/graphql/generated"
 import { usePriceConversion } from "@app/hooks"
@@ -14,6 +12,13 @@ import { StackScreenProps } from "@react-navigation/stack"
 import { Button, Text } from "@rneui/base"
 import { testProps } from "@app/utils/testProps"
 import { useDisplayCurrency } from "@app/hooks/use-display-currency"
+import {
+  BtcPaymentAmount,
+  DisplayCurrency,
+  MoneyAmount,
+  WalletOrDisplayCurrency,
+} from "@app/types/amounts"
+import { MoneyAmountInput } from "@app/components/money-amount-input"
 
 const styles = EStyleSheet.create({
   tabRow: {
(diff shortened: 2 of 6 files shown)


app/navigation/stack-param-lists.ts
```diff
@@ -59,13 +60,13 @@ export type RootStackParamList = {
...
-    minWithdrawableSatoshis: number
-    maxWithdrawableSatoshis: number
+    minWithdrawableSatoshis: MoneyAmount<typeof WalletCurrency.Btc>
+    maxWithdrawableSatoshis: MoneyAmount<typeof WalletCurrency.Btc>
     receiveCurrency: WalletCurrency
-    walletId: string
-    satAmount: number
-    satAmountInUsd: number
-    amountCurrency: string
+    receivingWalletDescriptor: WalletDescriptor<typeof WalletCurrency.Btc>
```
nicolasburtey, line 66: lnurlWithdraw is always connected to the BTC wallet?

UncleSamtoshi: Until we can create usd invoices using sat amounts

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
```diff
@@ -135,17 +140,28 @@ const RedeemBitcoinDetailScreen = ({
...
 }: StackScreenProps<RootStackParamList, "redeemBitcoinDetail">) => {
-  const { fiatSymbol } = useDisplayCurrency()
+  const { formatMoneyAmount, displayCurrency } = useDisplayCurrency()
 
   const { callback, domain, defaultDescription, k1, minWithdrawable, maxWithdrawable } =
     route.params.receiveDestination.validDestination
-  const minWithdrawableSatoshis = minWithdrawable / 1000
-  const maxWithdrawableSatoshis = maxWithdrawable / 1000
+  const minWithdrawableSatoshis: BtcPaymentAmount = {
+    amount: Math.round(minWithdrawable / 1000),
```
nicolasburtey, line 148: is `minWithdrawable` in msats? maybe this should be explained as why we're divided by 1000

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
```diff
@@ -135,17 +140,28 @@ const RedeemBitcoinDetailScreen = ({
...
+    currency: WalletCurrency.Btc,
+  }
+  const maxWithdrawableSatoshis: BtcPaymentAmount = {
+    amount: Math.round(maxWithdrawable / 1000),
+    currency: WalletCurrency.Btc,
+  }
+  const canSetAmount = minWithdrawableSatoshis.amount !== maxWithdrawableSatoshis.amount
```
nicolasburtey, line 155:
```suggestion
  const amountIsDefined = minWithdrawableSatoshis.amount !== maxWithdrawableSatoshis.amount
```

maybe a better variable name?

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
UncleSamtoshi: I can add a comment

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
```diff
@@ -135,17 +140,28 @@ const RedeemBitcoinDetailScreen = ({
...
 
   const [receiveCurrency, setReceiveCurrency] = useState<WalletCurrency>(
     WalletCurrency.Btc,
   )
   const { LL } = useI18nContext()
+  const { data } = useReceiveBtcQuery({ fetchPolicy: "cache-first" })
+  const btcWalletId = data?.me?.defaultAccount?.btcWallet?.id
+
+  const usdWalletId = null // TODO: when usd wallet ln invoices can be generated providing the sats amount as in put we can have the usdWalletId from useMainQuery as follows: const { usdWalletId } = useMainQuery()
```
nicolasburtey, line 164: we should not use the mainQuery on this screen. I know the comment has just been copied but it's no longer relevant

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
UncleSamtoshi: What about `amountIsFlexible`

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
```diff
@@ -307,104 +291,46 @@ const RedeemBitcoinDetailScreen = ({
...
-              </>
+            <MoneyAmountInput
+              moneyAmount={unitOfAccountAmount}
+              setAmount={(amount) => {
+                setUnitOfAccountAmount(amount)
+              }}
+              style={styles.walletBalanceInput}
+              editable={canSetAmount}
+            />
+            {canSetAmount && (
```
nicolasburtey, line 302: I think this could be another PR, but the message here is not very clear imo. we got some feedback on this.

I think typically the min max is either:
- equal
- very broad (ie: 1 sats to 1 btc)

if this is equal, then we don't need to display it which is what this branch is about I think

if this is not equal, then I think we should do like when handle the balance. we only show it when we're out of bound.

nicolasburtey: ohhh. this is for lnurl withdrawal. I thought this was for lnurl-pay. not sure if my comment is still relevant.

nicolasburtey (commented):
few comments but looks good overall

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
nicolasburtey: fine with that. I think that will be better than canSetAmount. the fact it can be interpreted as verb and not as a statement is where is it's confusing to me, because it's given state.

app/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.tsx
UncleSamtoshi: I think it may be more relevant to lnurlpay. Especially if you want to try and withdraw the max amount you need to be able to see what it is.

UncleSamtoshi:
Feedback addressed

nicolasburtey (approved):

PR #1317, opened 2023-03-03
chore: better use display hook

nicolasburtey:
- do not rely on useDisplayCurrencyQuery internally. instead displayCurrency value returned by useRealtimePriceQuery
- use the authenticated endpoint. set the right refrech so we should never have stealth data.


app/hooks/use-price-conversion.ts
```diff
@@ -9,21 +9,22 @@ import {
 } from "@app/types/amounts"
 import * as React from "react"
 import { useMemo } from "react"
-import { useRealtimePriceWrapper } from "./use-realtime-price"
 
 export const SATS_PER_BTC = 100000000
 
 export const usePriceConversion = () => {
-  const { data } = useRealtimePriceWrapper({ fetchPolicy: "cache-first" })
+  const { data } = useRealtimePriceQuery({ fetchPolicy: "cache-first" })
```
UncleSamtoshi, line 16: Why does `useRealtimePriceQuery` not have to be skipped here but skipped in the `useDisplayCurrencyHook`

nicolasburtey: that's an oversight.

UncleSamtoshi (approved):

PR #1320, opened 2023-03-04
chore: format currency to always show symbol

Extheoisah:


nicolasburtey (commented):
It feels like a hack:

Intl should solve those internationalization for us automatically. If we're using it wrong we should understand why

We're using Intl in few other places. So it is that your workaround is the cleaner solution we should use it other place we're calling Intl.NumberFormat

nicolasburtey (changes requested):
this would not work with hermes

nicolasburtey:
ok I got to understand why we have this discrepency. 

`Intl.NumberFormat` doesn't have the full capability on hermes than the normal Javascript engine provide. you can see a number of open issues there: https://github.com/facebook/hermes/search?q=NumberFormat&type=issues

so we probably need to use a polyfill until this is implemented in hermes

Extheoisah:
I didn't know about the hermes issue.
But the intl.formatNumber method doesn't automatically return some symbols by default so you need to explicitly handle it. See the MDN docs here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat
You can try putting

nicolasburtey:
> I didn't know about the hermes issue. But the intl.formatNumber method doesn't automatically return some symbols by default so you need to explicitly handle it. See the MDN docs here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat You can try putting NGN in the playground to replicate

it does if you add narrowSymbol:

```
console.log(new Intl.NumberFormat('us', { style: 'currency', currency: 'NGN',         currencyDisplay: "narrowSymbol"
 }).format(number));
```

return "₦123,456.79"

but this is not available on Hermes

Issue #1264, opened 2023-02-24
bug: Dust transaction onchain


nicolasburtey:
Showing unable to calculate fees where it should be saying another error message like impossible to send tx because amount is too low (below dust limit).

This probably have to do with the error parsing. 

![image](https://user-images.githubusercontent.com/11156370/221298551-6e801e17-f2b3-4044-a619-f221dc46169c.png)

designsats:
Still present

grimen:
**Problem**
Sending a dust/below-fee-threshold on-chain amount shows a generic "unable to calculate fees" error instead of a message that explains the actual cause (amount too low to be economically sendable / below dust limit).

**Context**
- Confirmed still reproducible.
- Likely an error-parsing/mapping issue: the underlying fee-estimation failure (amount too low to cover fees, or below the network dust limit) is not distinguished from other fee-calculation failures before being surfaced to the user.
- Screenshot in the original report shows the generic error on an on-chain send.

**Acceptance criteria**
- [ ] Identify the fee-estimation error case(s) that correspond to "amount too low to send" (below dust limit and/or amount insufficient to cover fees), and distinguish them from other "unable to calculate fees" failures at the point they're parsed.
- [ ] User sees a specific, actionable message (e.g. "amount is too low to send — increase the amount") instead of the generic fee-calculation error, for this case.
- [ ] Other genuine fee-calculation failures (e.g. network/node errors) continue to show the existing generic error — no regression there.
- [ ] Covered by a test that triggers a below-dust/insufficient-amount on-chain send and asserts the specific error message is shown.

PR #1333, opened 2023-03-04
fix: added checks for onChain trasactions

Aditi-Singh16:
Fixes for #1264

image


nicolasburtey:
the problem is not as easy as changing the string. the error needs to be parse to see if it arises from a fee too low or some other user/server errors.

Aditi-Singh16:
here I can see that appropriate error msg is shown in the onerror function in client.tsx when I tried to pay myself
so, I should show that error here?

Aditi-Singh16:
So now it does two things:
1. Shows the error msg "User tried to pay themselves" when I send sats to myself
2. Shows error when the amount is < dust limit (I googled it and have set it to 546 sats)

nicolasburtey:
> So now it does two things:
> 
> 1. Shows the error msg "User tried to pay themselves" when I send sats to myself
> 2. Shows error when the amount is < dust limit (I googled it and have set it to 546 sats)

setting a "global variable" from client.tsx is not the correct approach. queries and mutation return errors, they should be use directly.

I think this is a good reading to understand more the capability of apollo: https://www.apollographql.com/docs/react/data/error-handling

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -330,6 +340,10 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
     DisplayCurrency,
   )
 
+  if (paymentDetail.settlementAmount.amount < 546) {
```
nicolasburtey, line 343: @dolcalmi @vindard dust threshold is a setting on the galoy.yaml https://github.com/GaloyMoney/galoy/blob/main/galoy.yaml#L168. should we expose that to make the mobile app more dyanmic on what threshold it should accept for sending onchain transactions?

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -330,6 +340,10 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
...
+    feeDisplayText = "Impossible to send transaction because the amount is too low"
```
nicolasburtey, line 344: @Aditi-Singh16 string needs to be translated

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -219,6 +220,13 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
 
   const [paymentError, setPaymentError] = useState<string | undefined>(undefined)
   const { LL } = useI18nContext()
+  const { error } = useQuery(OnChainTxFeeDocument, {
```
nicolasburtey, line 223: - we don't want to query `useOnChainTxFeeLazyQuery` again, this will overload the server for now reason. we should be using the result from the existing `useOnChainTxFeeLazyQuery` query
- we should not use useQuery but always use the associated types query, ie: `useOnChainTxFeeLazyQuery` or `useOnChainTxFeeQuery`

Aditi-Singh16: Okay, got it!

Aditi-Singh16:
This is weird, I added only the app folder but I see so many changes 💀

Aditi-Singh16:
image

if I click on contribute on my branch(forked version) it only shows 3 files changed, any idea why is it 164 here?
I think I will close this and create a new PR, sorry for the inconvenience,

PR #1338, opened 2023-03-05
chore: cleaning up dependency on non JS component

nicolasburtey:
this commit is an extract from https://github.com/GaloyMoney/galoy-mobile/pull/1314, which aims at introducing js only tests for flows. 

splitting the PR to make it easier to understand what can be breaking (this PR). the other part is only new code or test/stories edit only.


app/graphql/messaging.tsx
```diff
@@ -0,0 +1,19 @@
...
+// refetch when we receive an OS notification
+export const MessagingContainer = () => {
+  const client = useApolloClient()
+
+  useEffect(() => {
+    const unsubscribe = messaging().onMessage(async (_remoteMessage) => {
+      client.refetchQueries({ include: [MainAuthedDocument] })
+    })
+
+    return unsubscribe
```
UncleSamtoshi, line 15: Is unsubscribe a function here?

UncleSamtoshi (approved):
Approved but one question.

app/graphql/messaging.tsx
nicolasburtey: that is correct

image

why?

UncleSamtoshi: Sorry I could have phrased the question better. I was just confirming it was a function because I wasn't sure what `.onMessage` returned.

PR #1343, opened 2023-03-06
Marketplace Functionality

phaoga54:


nicolasburtey:
hey @phaoga54 I'm closing this PR as it seems it aims at pulling from the upstream on our end but with logo/name change and other things that we will not want to merge. when you have the marketplace functionality ready you can re-open a Pr with it and we can get guidance on whether it's something we could merge on our end

PR #1346, opened 2023-03-06
chore: parse send payment error messages

Extheoisah:


app/screens/send-bitcoin-screen/use-send-payment.ts
```diff
@@ -114,15 +115,19 @@ export const useSendPayment = (
     return (
       sendPaymentFn &&
       (async () => {
-        const response = await sendPaymentFn({
+        const { status, errors } = await sendPaymentFn({
```
nicolasburtey, line 118: why did the type didn't catch this one?

Extheoisah: Not sure myself, but here's chatGPT's answer
`the inferred types of the sendPayment function returned by the hook and the sendPaymentFn function passed into the hook are structurally compatible, even though their explicit types do not match.
This means that TypeScript does not throw a type error in this case, even though the explicit types of the two functions do not match. However, it is still important to ensure that the actual types of the functions match the intended types, to avoid potential runtime errors or unexpected behavior.`

UncleSamtoshi: Very strange the type checker didn't catch this

UncleSamtoshi (approved):

PR #1372, opened 2023-03-06
feat: display currency in tx history

UncleSamtoshi:
- adds display currency in tx history


UncleSamtoshi:
This PR still requires two things:

- Fee to returned from the api in a display amount
- Regenerate the storybook mock data with the new tx fields

app/components/transaction-item/transaction-item.tsx
```diff
@@ -75,8 +77,8 @@ export const descriptionDisplay = (tx: Transaction) => {
       return "Invoice"
     case "SettlementViaIntraLedger":
       return isReceive
-        ? `From ${settlementVia.counterPartyUsername || "BitcoinBeach Wallet"}`
-        : `To ${settlementVia.counterPartyUsername || "BitcoinBeach Wallet"}`
+        ? `From ${settlementVia.counterPartyUsername || "BitcoinBeach"}`
```
nicolasburtey, line 80: we should use a const for `"walletName"`

UncleSamtoshi:
image

I also updated the background colors to match other branding in the app, although I am now realizing it may not be enough contrast with the white text.

app/components/transaction-item/transaction-item.tsx
```diff
@@ -136,13 +138,21 @@ export const TransactionItem: React.FC<Props> = ({
...
     currency: tx.settlementCurrency,
   })
 
+  const formattedDisplayAmount = formatCurrency({
+    amountInMajorUnits: tx.settlementDisplayAmount,
+    currency: tx.settlementDisplayCurrency,
+  })
+
+  const formattedSecondaryAmount =
+    tx.settlementDisplayCurrency === tx.settlementCurrency
```
nicolasburtey, line 152: in what case would this be true or false?

app/components/transaction-item/transaction-item.tsx
```diff
@@ -177,12 +193,19 @@ export const TransactionItem: React.FC<Props> = ({
...
-          <Text
-            style={amountDisplayStyle({ isReceive, isPending })}
-            onPress={hideBalance ? pressTxAmount : undefined}
-          >
-            {amountWithCurrency}
-          </Text>
+          <View>
+            <Text
+              style={amountDisplayStyle({ isReceive, isPending })}
+              onPress={hideBalance ? pressTxAmount : undefined}
```
nicolasburtey, line 199: onPress should maybe be in the parent View?

app/components/transaction-item/transaction-item.tsx
UncleSamtoshi: Usd display currency in a usd wallet or sat display currency in a btc wallet would both be true. Everything else will be false.

app/hooks/use-display-currency.ts
```diff
@@ -32,54 +31,67 @@ gql`
...
 
-  const displayCurrencyInfo = useMemo(() => {
+  const displayCurrencyDictionary = useMemo(() => {
     const currencyList = dataCurrencyList?.currencyList || []
-    return (
-      currencyList.find((currency) => currency.id === displayCurrency) ||
-      defaultDisplayCurrency
-    )
-  }, [dataCurrencyList, displayCurrency])
+    return currencyList.reduce((acc, currency) => {
```
nicolasburtey, line 74: why using a reducer here / creating a dictionnary? There'll be at most 150 currencies, the loop should run super quickly. not sure it's worth the code complexity to create mapping imo. useMemo already provide some optimization.

UncleSamtoshi: With tx history you need to be able to display any currency not just the current display currency, so without the dictionary you need to loop through the list for every amount displayed on every render. It still may not be worth the optimization, but we have optimized for less in the past.

nicolasburtey: oh right ok. 

we're still doing N * M, N been transaction list and M been currency list. so that make more sense. didn't had the N in mind.

nicolasburtey (approved):

PR #1374, opened 2023-03-07
ci: store build logs as artifacts

krtk6160:


sandipndev (dismissed):

sandipndev:
Don't we need logs from other steps too?
And does tee take into account stderr?

krtk6160:
> Don't we need logs from other steps too? 
We might, can add whichever we identify as important in another PR. I'm trying to get this flow working end to end first.

> And does tee take into account stderr?
Don't think so, so I've added redirection of stderr to stdout in [f299be1](https://github.com/GaloyMoney/galoy-mobile/pull/1374/commits/f299be1c0ff024f72fecc7b26fc8d1d041b78b8c)

sandipndev (approved):

PR #1380, opened 2023-03-07
chore: put menu in the right place

nicolasburtey:


app/screens/settings-screen/settings-screen.tsx
```diff
@@ -195,6 +195,17 @@ export const SettingsScreen: React.FC = () => {
       enabled: isAuthed,
       greyed: !isAuthed,
     },
+    beta
```
Extheoisah, line 198: This ternary operation might not be necessary. We can just hide the button using `hidden` property if beta is false like so: 
```js
   {
      category: LL.common.currency(),
      icon: "ios-cash",
      id: "currency",
      action: () => navigation.navigate("currency"),
      subTitleText: displayCurrency,
      enabled: isAuthed,
      greyed: !isAuthed,
      hidden: !beta,
    }

PR #1403, opened 2023-03-08
fix: display amounts in fixed amount invoices

UncleSamtoshi:
- show display amount instead of btc amount for fixed amount invoices
- prevent minor field from being editable

### DIFF
--- a/app/components/money-amount-input/index.tsx
+++ b/app/components/money-amount-input/index.tsx
@@ -50,11 +50,16 @@ export const MoneyAmountInput = ({
   return (
     <FakeCurrencyInput
       value={moneyAmountToMajorUnitOrSats(moneyAmount)}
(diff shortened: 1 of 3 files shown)


app/components/money-amount-input/index.tsx
```diff
@@ -50,11 +50,16 @@ export const MoneyAmountInput = ({
...
+      onChangeValue={
+        setAmount
+          ? (amount) =>
+              setAmount(
+                amountInMajorUnitOrSatsToMoneyAmount(
+                  Number(amount),
+                  moneyAmount.currency,
+                ),
+              )
+          : undefined
```
nicolasburtey, line 62: imo, it starts to be complex enough that I would not inline that in the JSX but have this in the code before the return

nicolasburtey (approved):

PR #1407, opened 2023-03-08
chore: reduce android build size using android app bundles

Extheoisah:
This PR is an attempt to reduce the bundle size for android. The last android update (to v2.1.67) was about `89mb` which is too large in my opinion. 
the google docs recommend using Android app bundles instead of Apk's to reduce bundle size and it is new reommended method.
see docs here       https://developer.android.com/topic/performance/reduce-apk-size#app_bundle

One thing we need to do is to update the fastlane commands for uploading to playstore to use the .aab files instead of apk's https://docs.fastlane.tools/actions/upload_to_play_store/

I couldn't test this locally cox i needed to sign the new build (compulsory for aab) for which I dont have the signing certificate.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -103,6 +103,22 @@ android {
         versionName "2.1.55"
         missingDimensionStrategy 'react-native-camera', 'general' // React native camera
     }
+    buildFeatures {
+        viewBinding = true
+    }
+    bundle {
+        // use android app bundle instead of apk (https://developer.android.com/guide/app-bundle)
+        // https://developer.android.com/topic/performance/reduce-apk-size#app_bundle
+        language {
+            enableSplit = false
+        }
+        density {
+            enableSplit = true
+        }
+        abi {
+            enableSplit = true
+        }
+    }
     splits {
         abi {
             reset()
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -831,7 +831,7 @@ SPEC CHECKSUMS:
   FirebaseInstallations: 40bd9054049b2eae9a2c38ef1c3dd213df3605cd
   FirebaseMessaging: 5e5118a2383b3531e730d974680954c679ca0a13
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 04b94705f318337d7ead9e6d17c019bd9b1f6b1b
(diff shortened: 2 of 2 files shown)


krtk6160 (approved):

nicolasburtey (changes requested):
we are already splitting the .apk into several runtime

nicolasburtey:
> we are already splitting the .apk into several runtime

I don't think we should do this. I can provide more context later.

nicolasburtey:
> This PR is an attempt to reduce the bundle size for android. The last android update (to v2.1.67) was about `89mb` which is too large in my opinion. the google docs recommend using Android app bundles instead of Apk's to reduce bundle size and it is new reommended method. see docs here https://developer.android.com/topic/performance/reduce-apk-size#app_bundle
> 
> One thing we need to do is to update the fastlane commands for uploading to playstore to use the .aab files instead of apk's https://docs.fastlane.tools/actions/upload_to_play_store/
> 
> I couldn't test this locally cox i needed to sign the new build (compulsory for aab) for which I dont have the signing certificate.

we have some regression. last week we were shipping one apk/runtime: 
https://github.com/GaloyMoney/galoy-mobile/releases?page=2

we should look into why we have this regression, instead of going for Android app bundle for now.

Extheoisah:
> > This PR is an attempt to reduce the bundle size for android. The last android update (to v2.1.67) was about `89mb` which is too large in my opinion. the google docs recommend using Android app bundles instead of Apk's to reduce bundle size and it is new reommended method. see docs here https://developer.android.com/topic/performance/reduce-apk-size#app_bundle
> > One thing we need to do is to update the fastlane commands for uploading to playstore to use the .aab files instead of apk's https://docs.fastlane.tools/actions/upload_to_play_store/
> > I couldn't test this locally cox i needed to sign the new build (compulsory for aab) for which I dont have the signing certificate.
> 
> we have some regression. last week we were shipping one apk/runtime: https://github.com/GaloyMoney/galoy-mobile/releases/tag/2.1.55
> 
> we should look into why we have this regression, instead of going for Android app bundle for now.

But don't you think we should move towards using app bundles since google is pushing it as the new default way of uploads?
I agree we should look into if this was a regression though

Extheoisah:
closing this PR, as there some critical side effects these might have that we aren't ready for at the moment.

PR #1414, opened 2023-03-08
fix: contact list not reinitialized correctly when tapping the cross …

nicolasburtey:
…icon

### DIFF
--- a/app/screens/contacts-screen/contacts.tsx
+++ b/app/screens/contacts-screen/contacts.tsx
@@ -123,7 +123,9 @@ export const ContactsScreen: React.FC = () => {
 
   React.useEffect(() => {
     reset()
-  }, [reset])
+    // only call on mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
 
   // This implementation of search will cause a match if any word in the search text
   // matches the contacts name or prettyName.


app/screens/contacts-screen/contacts.tsx
```diff
@@ -116,10 +116,15 @@ export const ContactsScreen: React.FC = () => {
     return data?.me?.contacts.slice() ?? []
   }, [data])
 
-  React.useEffect(() => {
+  const reset = useCallback(() => {
+    setSearchText("")
```
UncleSamtoshi, line 120: As more contacts load  or reload I dont think we want to reset the search text.

UncleSamtoshi (changes requested):

app/screens/contacts-screen/contacts.tsx
nicolasburtey: not sure to understand your inputs

reset only get called:
- on initial load (when searchText is "" already)
- when tapping on the cross, at the end of the search bar.

UncleSamtoshi: It gets called whenever reset changes in line 126. Reset changes whenever contacts change in line 115. Which changes whenever data changes. Which right now should only happen potentially a second after initial load when the 'network' part of 'cache-and-network' happens, but it could also happen if we add in a refetch or pagination.

nicolasburtey: ok. not sure it would be an issue, if network were to update the app, it would most likely before the user has start to use the search, but I get the point. PR updated

UncleSamtoshi (approved):

PR #1416, opened 2023-03-08
fix: send flow when switching wallets

UncleSamtoshi:
- fixes issue when switching between usd and btc wallets in the send flow where the currency was stuck in the settlement amount for the previous wallet

### DIFF
--- a/app/graphql/generated.ts
+++ b/app/graphql/generated.ts
@@ -1165,6 +1165,7 @@ export type Transaction = {
   readonly settlementCurrency: WalletCurrency;
   readonly settlementDisplayAmount: Scalars['SignedDisplayMajorAmount'];
   readonly settlementDisplayCurrency: Scalars['DisplayCurrency'];
+  readonly settlementDisplayFee: Scalars['SignedDisplayMajorAmount'];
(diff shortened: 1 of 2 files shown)


nicolasburtey (approved):

nicolasburtey:
> * fixes issue when switching between usd and btc wallets in the send flow where the currency was stuck in the settlement amount for the previous wallet

I'm confused from the description on what is happening and how this is fixing it, but approving nonethelss, this is a small PR. 

I think only unit/component tests will be able to explain more of those nuances behavior in detail

UncleSamtoshi:
> > * fixes issue when switching between usd and btc wallets in the send flow where the currency was stuck in the settlement amount for the previous wallet
> 
> I'm confused from the description on what is happening and how this is fixing it, but approving nonethelss, this is a small PR.
> 
> I think only unit/component tests will be able to explain more of those nuanced behavior/expectation in detail

If the wallet is set to usd and the amount is being entered in usd, when you switch to the btc wallet you need to convert the amount to being entered in either the display currency or in btc otherwise the amount is still denominated in usd and the input is configured to show the amount that the payment is denominated it.

PR #1424, opened 2023-03-09
fix: scanning an expired invoice shows new alert

### LINKED ISSUE #1089: Scanning an expired invoice shows: This is not a valid Bitcoin address or Lightning invoice

siiddhantt:
fixes #1089 
image


siiddhantt:
happy to contribute 🚀

PR #1430, opened 2023-03-09
fix: Error msg fix for onchain tx

Aditi-Singh16:
Fixes for #1264

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
@@ -318,7 +318,7 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
           return
         }
 
-        setPaymentError(errorsMessage || "Something went wrong")
+        setPaymentError(paymentError || "Something went wrong")
         ReactNativeHapticFeedback.trigger("notificationError", {
           ignoreAndroidSystemSettings: true,
         })
--- a/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
@@ -1,13 +1,13 @@
-import { gql } from "@apollo/client"
+import { gql, useQuery } from "@apollo/client"
 import DestinationIcon from "@app/assets/icons/destination.svg"
 import NoteIcon from "@app/assets/icons/note.svg"
 import { MoneyAmountInput } from "@app/components/money-amount-input"
 import { PaymentDestinationDisplay } from "@app/components/payment-destination-display"
 import {
+  OnChainTxFeeDocument,
   useSendBitcoinConfirmationScreenQuery,
   WalletCurrency,
 } from "@app/graphql/generated"
-import { onChainError } from "@app/graphql/client"
 import { useIsAuthed } from "@app/graphql/is-authed-context"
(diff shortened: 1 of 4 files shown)


Aditi-Singh16:
@nicolasburtey , @Extheoisah could you please review this PR too?

nicolasburtey:
> @nicolasburtey , @Extheoisah could you please review this PR too?

I provided feedback on the previous PR that you have closed. we don't want to call the method just to get the error. this would overload the server for now reason.

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -219,6 +220,14 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
 
   const [paymentError, setPaymentError] = useState<string | undefined>(undefined)
   const { LL } = useI18nContext()
+  const { error } = useOnChainTxFeeQuery({
```
nicolasburtey, line 223: this is a call to the server. we don;t want that just to get the error

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -327,6 +338,18 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
...
+  }
+
+  if (paymentDetail.settlementAmount.amount < 546) {
+    feeDisplayText = LL.SendBitcoinConfirmationScreen.dustError()
+  }
+
+  if (paymentDetail.settlementAmount.amount < 546) {
+    feeDisplayText = LL.SendBitcoinConfirmationScreen.dustError()
+  }
+
```
nicolasburtey, line 352: why is this copy/paste 3 times?

Aditi-Singh16: I think this occurred when I was resolving merge conflicts, I will change it

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
Aditi-Singh16: I was debugging some variables and the `sendPayment `  function returns the correct gql error, like : 
1. Unable to calculate fee because Amount is too low
2. User tried to pay themselves
These errors are shown when the user clicks on confirm payment button.
So, this may come across as redundant to show the same message again, in the Fee container, and when the user clicks on confirm payment...
Should we disable the confirm payment button on such errors and then just show the error once?

Aditi-Singh16: |||
| :-------------: | :-------------: |
| <video src="https://user-images.githubusercontent.com/74586135/226171020-b5ab66c0-aadf-4534-9307-e5b99e4256e7.mp4" > | <video src="https://user-images.githubusercontent.com/74586135/226171023-d95e28c2-89bb-4b3f-a91e-a768ba4f1f8d.mp4">

Aditi-Singh16:
I did update the branch, but why isn't it reflecting here?

image

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
Aditi-Singh16: This is how it works now

PR #1432, opened 2023-03-09
chore: first iteration on refactoring WalletOverview

nicolasburtey:


app/components/text-currency/text-currency.tsx
```diff
@@ -31,17 +31,17 @@ export const TextCurrencyForAmount: React.FC<Props> = ({
   iconColor,
 }) => {
   const { formatToDisplayCurrency } = useDisplayCurrency()
```
UncleSamtoshi, line 33: Is this component still used anywhere?

app/screens/send-bitcoin-screen/payment-details/onchain.ts
```diff
@@ -22,6 +22,8 @@ export const createNoAmountOnchainPaymentDetails = <T extends WalletCurrency>(
   params: CreateNoAmountOnchainPaymentDetailsParams<T>,
 ): PaymentDetail<T> => {
   const {
+    // FIXME: is that coming from usePriceConversion()
+    // in that case we should not pass it as a param
```
UncleSamtoshi, line 26: Whats your thinking behind this?

UncleSamtoshi (changes requested):

UncleSamtoshi (changes requested):

app/components/text-currency/text-currency.tsx
nicolasburtey: Yeah still in balance header
Plan to removed it when I'm doing a PR for it next

app/screens/send-bitcoin-screen/payment-details/onchain.ts
nicolasburtey: Simplifying the Params passed via route. It makes mocking/testing screen easier/ less boilerplat-y. 

But the question here is: is it the same function as the one returned by the hook or do you wrap it somehow? Didn't check into details?

UncleSamtoshi: It is the same function from the hook, but it is important that the function is part of the object as it is used for the conversion to the settlement amount and it ensures that we can freeze the price at a particular point which is important when the user has confirmed a settlment amount and validation has been performed on that settlement amount.

app/components/text-currency/text-currency.tsx
UncleSamtoshi: Sounds good!

app/screens/send-bitcoin-screen/payment-details/onchain.ts
nicolasburtey: ok got it.

UncleSamtoshi (approved):

PR #1434, opened 2023-03-09
chore: add usePriceConversion tests

UncleSamtoshi:


yarn.lock
```diff
@@ -17757,6 +17765,13 @@ react-element-to-jsx-string@^14.3.4:
     is-plain-object "5.0.0"
     react-is "17.0.2"
 
+react-error-boundary@^3.1.0:
+  version "3.1.4"
+  resolved "https://registry.yarnpkg.com/react-error-boundary/-/react-error-boundary-3.1.4.tgz#255db92b23197108757a888b01e5b729919abde0"
```
nicolasburtey, line 17770: interesting that here the link is on `yarnpkg`. when I did yarn upgrade it migrate link to `npmjs.org` from `yarnpkg.com`. maybe we use different version (not an issue)

__tests__/hooks/use-price-conversion.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      __typename: "User",
+      defaultAccount: {
+        __typename: "ConsumerAccount",
+        id: "f2b1d0bf-816c-51db-aea4-4b773cfdf7a7",
+        realtimePrice: {
+          __typename: "RealtimePrice",
+          btcSatPrice: {
+            __typename: "PriceOfOneSat",
+            base: 10118784000000,
+            currencyUnit: "NGNCENT",
```
nicolasburtey, line 38: do we need to fetch currency unit? I don't think we are using it?

__tests__/hooks/use-price-conversion.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    mockUseRealtimePriceQuery.mockReturnValue({ data: undefined })
+
+    const { result } = renderHook(() => usePriceConversion())
+    expect(result.current).toEqual({
+      convertMoneyAmount: undefined,
+      usdPerBtc: {
+        amount: NaN,
+        currency: WalletCurrency.Usd,
+      },
+      usdPerSat: null,
```
nicolasburtey, line 90: feels like an inconsistent behavior that `usdPerBtc` return an object with `NaN` amount, but `usdPerSat` return `null`

__tests__/hooks/use-price-conversion.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    it("should make proper conversions", () => {
+      // test all conversions
+      for (const fromCurrency of Object.keys(amounts)) {
+        for (const toCurrency of Object.keys(amounts)) {
+          const fromAmount = amounts[fromCurrency as keyof typeof amounts]
+          const toAmount = amounts[toCurrency as keyof typeof amounts]
+
+          const convertedAmount = convertMoneyAmount(fromAmount, toAmount.currency)
+          // expect amounts to be within .1% of each other due to rounding
```
nicolasburtey, line 111: .1% would feels a big difference? does the test pass on 0.01%? wondering how big the round issues can be if that were the case.

nicolasburtey (commented):
some questions/comments

yarn.lock
UncleSamtoshi: That is strange.

__tests__/hooks/use-price-conversion.spec.ts
UncleSamtoshi: Switched to .01%

nicolasburtey (approved):

PR #1436, opened 2023-03-09
chore: removing TextCurrencyForAmount

nicolasburtey:


app/components/balance-header/balance-header.tsx
```diff
@@ -88,19 +93,39 @@ type Props = {
...
   // so there is no need to pass loading from parent?
   const { data } = useBalanceHeaderQuery({ skip: !isAuthed })
-  const usdWalletDisplayBalance = isAuthed
-    ? data?.me?.defaultAccount?.usdWallet?.displayBalance ?? NaN
-    : 0
 
-  const btcWalletDisplayBalance = isAuthed
-    ? data?.me?.defaultAccount?.btcWallet?.displayBalance ?? NaN
-    : 0
+  let balanceInDisplayCurrency = "$0"
```
UncleSamtoshi, line 103: Not in the scope of this PR, but there are several areas where we are making the assumption that the default display currency is "$". It would be better to define the default display currency in one place and the reference that config.

UncleSamtoshi (approved):

PR #1437, opened 2023-03-10
chore: remove conversion within cache.ts for more consistency

nicolasburtey:


app/types/amounts.ts
```diff
@@ -22,3 +22,36 @@ export type WalletAmount<T extends WalletCurrency> = MoneyAmount<T>
...
+
+export const toBtcMoneyAmount = (amount: number | undefined): BtcMoneyAmount => {
+  if (amount === undefined) {
+    return {
+      amount: NaN,
+      currency: WalletCurrency.Btc,
+    }
+  }
+  return {
+    amount,
```
UncleSamtoshi, line 34: We should round here so that amount is an integer.

UncleSamtoshi (changes requested):

app/types/amounts.ts
nicolasburtey: why?

nicolasburtey: in that case the type should be moving to bigint ? 

but I don't see why this function should take care about rounding. this is just creating a MoneyAmount type by adding wrapping on object with currency around the value provided

UncleSamtoshi: We haven't switched to the BigInt type yet, but ideally everything should be an integer because it is in minor units similar to the backend. All invoice creation mutations expect an integer.

nicolasburtey: then the rounding should be at the mutation stage right now?

only a fraction of the MoneyAmount type created for now will be done through this function. the code I refactoring from didn't do any rounding before.

UncleSamtoshi (approved):

PR #1447, opened 2023-03-12
chore: enable offline mode use case

nicolasburtey:


Extheoisah (approved):

Extheoisah:
removing the `notifyOnNetworkStatusChange` could also prevent the multiple re-renders we get when there's no network. We had this non-fatal error of recent in crashlytics

PR #1476, opened 2023-03-13
chore: refactoring persistent state

nicolasburtey:
- less bloated implementation of PersistentState:
-- only store id, token (and custom property if `id === Custom`)
-- for default network, use value from Array instead of story them
--- will make it easier to change bank name (ie: for forks)
--- less need for migration going forward
-- custom property like theme or Modals  like stablesatsModals who have been discarded will be stored eventually as apollo-client client-only data. not that critical if the value are been reseted eventually

- fix https://github.com/GaloyMoney/galoy-mobile/issues/684

-- TODO in a future PR: do not reset the apollo-client on every app update. but only on schema changes.

### DIFF
--- a/app/config/galoy-instances.ts
+++ b/app/config/galoy-instances.ts
@@ -1,13 +1,30 @@
 import { NativeModules } from "react-native"
 
+// this is used for local development
+// will typically return localhost
 const scriptHostname = (): string => {
   const { scriptURL } = NativeModules.SourceCode
   const scriptHostname = scriptURL?.split("://")[1].split(":")[0] ?? ""
   return scriptHostname
 }
 
-export const possibleGaloyInstanceNames = ["BBW", "Staging", "Local", "Custom"] as const
+export const standardInstances = ["BBW", "Staging", "Local"] as const
+export const possibleGaloyInstanceNames = [...standardInstances, "Custom"] as const
 export type GaloyInstanceName = (typeof possibleGaloyInstanceNames)[number]
+
+export type StandardInstance = {
+  name: "BBW" | "Staging" | "Local"
+}
+
+export type CustomInstance = {
+  name: "Custom"
+  graphqlUri: string
+  graphqlWsUri: string
+  posUrl: string
+  lnAddressHostname: string
+}
+
+export type GaloyInstanceInput = StandardInstance | CustomInstance
 
 export type GaloyInstance = {
   name: GaloyInstanceName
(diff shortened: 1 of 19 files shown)


app/store/persistent-state/state-migrations.ts
```diff
@@ -144,28 +180,25 @@ const stateMigrations: StateMigrations = {
...
-  galoyInstance: GALOY_INSTANCES[0],
+  schemaVersion: 5,
+  galoyInstance: { id: "Main" },
   galoyAuthToken: "",
-  isAnalyticsEnabled: true,
-  theme: defaultTheme,
 }
 
-export const deserializeAndMigratePersistentState = async (
+export const migrateAndGetPersistentState = async (
```
nicolasburtey, line 194: json is already deserialized

app/config/galoy-instances.ts
```diff
@@ -1,42 +1,84 @@
...
+
+  if (instance) {
+    return instance
+  }
+
+  console.error("instance not found") // should not happen
+  return GALOY_INSTANCES[0]
 }
 
 export const GALOY_INSTANCES: GaloyInstance[] = [
```
UncleSamtoshi, line 56: We could convert this to an object with the keys being the "Id" to avoid some of the typing issues.

app/config/galoy-instances.ts
```diff
@@ -1,42 +1,84 @@
...
+  id: GaloyInstanceName
+  name: string
   graphqlUri: string
   graphqlWsUri: string
   posUrl: string
   lnAddressHostname: string
+  blockExplorer: string
+}
+
+export const maybeAddDefault = (input: GaloyInstanceInput): GaloyInstance => {
```
UncleSamtoshi, line 41: This name doesnt really make sense to me. Maybe something like `resolveGaloyInstanceOrDefault` could be better.

app/config/galoy-instances.ts
nicolasburtey: I gave it a try, ie:

```
const GALOY_INSTANCES = {
  Main: {
    name: "BBW",
    graphqlUri: "https://api.mainnet.galoy.io/graphql",
    graphqlWsUri: "wss://api.mainnet.galoy.io/graphql",
    posUrl: "https://pay.bbw.sv",
    lnAddressHostname: "pay.bbw.sv",
    blockExplorer: "https://mempool.space/tx/",
  },
  Staging: {
    name: "Staging",
    graphqlUri: "https://api.staging.galoy.io/graphql",
    graphqlWsUri: "wss://api.staging.galoy.io/graphql",
    posUrl: "https://pay.staging.galoy.io",
    lnAddressHostname: "pay.staging.galoy.io",
    blockExplorer: "https://mempool.space/signet/tx/",
  },
  Local: {
    name: "Staging",
    graphqlUri: `http://${scriptHostname()}:4002/graphql`,
    graphqlWsUri: `ws://${scriptHostname()}:4002/graphql`,
    posUrl: `http://${scriptHostname()}:3000`,
    lnAddressHostname: `${scriptHostname()}:3000`,
    blockExplorer: "https://mempool.space/signet/tx/",
  },
} as const
```

but because we have the Custom flow, at the end of the day it's not changing much the need for complex type. I'm going to stick with the Array for now but see if I can improve the type

nicolasburtey: so the type issue is related to `find()` that always return `T | undefined`. there would be other ways to go with that does not rely on `find()` but that would make the code more complex (using filter and map). I think the current tradeoff is a good one.

UncleSamtoshi (approved):

PR #1520, opened 2023-03-13
chore: refactor common display logic

UncleSamtoshi:


app/hooks/use-display-currency.ts
```diff
@@ -173,6 +179,56 @@ export const useDisplayCurrency = () => {
...
+      walletAmount: WalletAmount<WalletCurrency>
+    }) => {
+      // if the display currency is the same as the wallet amount currency, we don't need to show the secondary amount (example: USD display currency with USD wallet amount)
+      if (walletAmount.currency === displayCurrency) {
+        return undefined
+      }
+
+      if (primaryAmount.currency === DisplayCurrency) {
+        return walletAmount
+      }
```
nicolasburtey, line 199: the mix between `displayCurrency` and `DisplayCurrency` make the code hard to read/understand. 

I understand than DisplayCurrency is hardcoded to `String(DisplayCurrency)` but I'm wondering if another name could help clarify the difference? (not obvious suggestion to mind right now)

nicolasburtey: maybe now that I get to read it multiple time, I get accustom to it.

app/hooks/use-display-currency.ts
```diff
@@ -173,6 +179,56 @@ export const useDisplayCurrency = () => {
...
+      const primaryAmountWithDefault = primaryAmount || displayAmount
+
+      const secondaryAmount = getSecondaryAmountIfCurrencyIsDifferent({
+        primaryAmount: primaryAmountWithDefault,
+        displayAmount,
+        walletAmount,
+      })
+
+      return `${formatMoneyAmount(primaryAmountWithDefault)}${
+        secondaryAmount ? ` (${formatMoneyAmount(secondaryAmount)})` : ""
```
nicolasburtey, line 226: this is hard to read/understand the logic

app/hooks/use-display-currency.ts
```diff
@@ -173,6 +179,56 @@ export const useDisplayCurrency = () => {
...
+      primaryAmount,
+      displayAmount,
+      walletAmount,
+    }: {
+      primaryAmount?: MoneyAmount<WalletOrDisplayCurrency>
+      displayAmount: DisplayAmount
+      walletAmount: WalletAmount<WalletCurrency>
+    }) => {
+      // if the primary amount is not provided, we use the display amount as the primary amount by default
+      const primaryAmountWithDefault = primaryAmount || displayAmount
```
nicolasburtey, line 217: in what scenario would the primary amount not be provided?

app/screens/receive-bitcoin-screen/receive-btc.tsx
```diff
@@ -278,10 +276,30 @@ const ReceiveBtc = () => {
...
   if (!paymentRequestDetails || !setAmount) {
     return <></>
   }
 
+  const togglePaymentRequestType = () => {
+    const newPaymentRequestType =
+      paymentRequestDetails.paymentRequestType === PaymentRequest.Lightning
+        ? PaymentRequest.OnChain
+        : PaymentRequest.Lightning
+    setPaymentRequestType(newPaymentRequestType, true)
```
nicolasburtey, line 300: I know this is just a code moving around, but using object as input to the function would make it more clear what `true` is used for here. 

the only way to know currently is to look at the function. with an object, the `true` would have more context also here.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -248,58 +249,40 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
...
-  const { moneyAmountToDisplayCurrencyString, formatMoneyAmount, displayCurrency } =
-    useDisplayCurrency()
+  const {
+    formatDisplayAndWalletAmount,
+    formatMoneyAmount,
+    getSecondaryAmountIfCurrencyIsDifferent,
+  } = useDisplayCurrency()
   const { LL } = useI18nContext()
-  const { convertMoneyAmount } = usePriceConversion()
+  const { convertMoneyAmount: _convertMoneyAmount } = usePriceConversion()
```
nicolasburtey, line 258: why is there a need to rename to `_convertMoneyAmount`?

nicolasburtey:
PR looks very good. few comments and questions.

app/hooks/use-display-currency.ts
UncleSamtoshi: In some cases you always want it to be formatted `displayAmount (walletAmount)`. If you know that you always want the displayAmount to be first, you can leave out the primaryAmount which is what is used to specify what goes first.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
UncleSamtoshi: To ensure the same price is being used everywhere, convertMoneyAmount should come from the paymentDetails object, so we use _convertMoneyAmount to be put into the paymentDetails object and then from the paymentDetails object we grab the function and name it convertMoneyAmount.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -329,65 +312,69 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
...
+    displayAmount: convertMoneyAmount(usdBalanceMoneyAmount, DisplayCurrency),
+    walletAmount: usdBalanceMoneyAmount,
+  })
+
+  let validAmount = true
+  let invalidAmountErrorMessage = ""
+
+  if (
+    moneyAmountIsCurrencyType(settlementAmount, WalletCurrency.Btc) &&
+    btcBalanceMoneyAmount
```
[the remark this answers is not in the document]
UncleSamtoshi: `btcBalanceMoneyAmount` is actually always defined at this point, so I will remove it from the conditional.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -587,22 +568,18 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
       <Text {...testProps("lnurl-min-max")}>
         {"Min: "}
         {formatMoneyAmount(
-          convertMoneyAmount(
-            { amount: lnurlParams.min, currency: WalletCurrency.Btc },
-            DisplayCurrency,
-          ),
+          convertMoneyAmount(toBtcMoneyAmount(lnurlParams.min), primaryAmount.currency),
```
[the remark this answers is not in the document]
UncleSamtoshi: Yeah I think this should be revisited. Wherever we are checking for nonzero amounts we should ensure that we are using the correct amount between the display amount and the wallet amount. In some cases one may be 0 where the other is not.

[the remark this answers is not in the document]
UncleSamtoshi: I created an issue.

app/screens/receive-bitcoin-screen/receive-btc.tsx
UncleSamtoshi: fixed it

nicolasburtey (approved):

Issue #1435, opened 2023-03-09
Support LNURL payRequest with predefined amount


UncleSamtoshi:
- Current behavior: The mobile wallet can parse an lnurl pay endpoint, however it does not populate the amount field when the min amount is equal to the max amount. Instead it allows the user to enter an amount.

- Expected behavior: The mobile wallet should parse the lnurl string and pre-populate the correct amount without allowing the user to change the amount.

PR #1523, opened 2023-03-14
chore: support LNURL payRequest with predefined amount

Extheoisah:
addresses issue https://github.com/GaloyMoney/galoy-mobile/issues/1435

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
@@ -555,9 +555,23 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
     )
   }
 
+  const isLnurlAmountEditable = lnurlParams && lnurlParams.min === lnurlParams.max
   let LnUrlMinMaxAmount: React.ReactNode = null
+  let primaryMoneyAmount: MoneyAmount<WalletOrDisplayCurrency> = primaryAmount
+  let secondaryMoneyAmount: MoneyAmount<WalletOrDisplayCurrency> = secondaryAmount
 
   if (lnurlParams && convertMoneyAmount) {
+    const { min, max } = lnurlParams
+    if (min === max) {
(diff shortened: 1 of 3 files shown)


app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -580,22 +580,36 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
     )
   }
 
+  const isLnurlAmountEditable = lnurlParams && lnurlParams.min === lnurlParams.max
```
UncleSamtoshi, line 583: I think these changes should be made in `createLnurlPaymentDetails`. Where `canSetAmount` should be false if `lnurlParams.min === lnurlParams.max` and the amount should be prepopulated.

UncleSamtoshi (changes requested):

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -558,15 +558,16 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
   let LnUrlMinMaxAmount: React.ReactNode = null
 
   if (lnurlParams && convertMoneyAmount) {
+    const { min, max } = lnurlParams
     LnUrlMinMaxAmount = (
```
UncleSamtoshi, line 562: This should only be shown if `canSetAmount` is true

UncleSamtoshi (changes requested):

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
Extheoisah: addressed here https://github.com/GaloyMoney/galoy-mobile/pull/1523/commits/aa5b944c59919282764b6c3e4f35c455027ada28

UncleSamtoshi (approved):

PR #1661, opened 2023-03-14
tests: first test for useDisplayCurrency

nicolasburtey:
- adding test for useDisplayCurrency
- using a different pattern than the tests used for usePriceConversion (we should eventually converge to way of mocking up the gql calls)
-- Note that I ran into some randomness issue with mocking. if the tests for the 0 digits is at the end, it's failing. not sure why but now re-order solves it. 

maybe more function can be tests on this hooks. this is a start.

### DIFF
--- a/__tests__/hooks/use-display-currency.spec.tsx
+++ b/__tests__/hooks/use-display-currency.spec.tsx
@@ -3,76 +3,141 @@ import { renderHook } from "@testing-library/react-hooks"
 import { useDisplayCurrency } from "@app/hooks/use-display-currency"
 import { MockedProvider } from "@apollo/client/testing"
 import { PropsWithChildren } from "react"
-import mocks from "@app/graphql/mocks"
 import * as React from "react"
 import { IsAuthedContextProvider } from "@app/graphql/is-authed-context"
 import { CurrencyListDocument, RealtimePriceDocument } from "@app/graphql/generated"
 
+const mocksNgn = [
+  {
+    request: {
+      query: RealtimePriceDocument,
+    },
+    result: {
+      data: {
+        me: {
+          __typename: "User",
+          id: "70df9822-efe0-419c-b864-c9efa99872ea",
+          defaultAccount: {
+            __typename: "Account",
+            id: "84b26b88-89b0-5c6f-9d3d-fbead08f79d8",
+            realtimePrice: {
+              btcSatPrice: {
+                base: 24015009766,
+                offset: 12,
+                currencyUnit: "USDCENT",
+                __typename: "PriceOfOneSat",
+              },
+              denominatorCurrency: "NGN",
+              id: "67b6e1d2-04c8-509a-abbd-b1cab08575d5",
+              timestamp: 1677184189,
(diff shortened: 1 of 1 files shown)


__tests__/hooks/use-display-currency.spec.tsx
```diff
@@ -0,0 +1,175 @@
...
+    const wrapper = ({ children }: PropsWithChildren) => (
+      <IsAuthedContextProvider value={true}>
+        <MockedProvider mocks={mocks}>{children}</MockedProvider>
+      </IsAuthedContextProvider>
+    )
+    const { result, waitForNextUpdate } = renderHook(() => useDisplayCurrency(), {
+      wrapper,
+    })
+
+    expect(result.current).toMatchObject({
```
UncleSamtoshi, line 161: Do we want this to be our behavior of briefly using USD until the query is resolved or would it be better for it to just be undefined?

__tests__/hooks/use-display-currency.spec.tsx
```diff
@@ -0,0 +1,175 @@
...
+        currency: "DisplayCurrency",
+      })
+
+      expect(res).toBe(100)
+    })
+
+    it("with 2 digits", async () => {
+      const wrapper = ({ children }: PropsWithChildren) => (
+        <IsAuthedContextProvider value={true}>
+          <MockedProvider mocks={mocks}>{children}</MockedProvider>
```
UncleSamtoshi, line 91: I think using the global mocks may not be the best idea for unit testing. It is difficult to understand what is actually being tested here because there is no indication what is contained in the mock. It also makes the mock more fragile to being updated in the future with tight coupling to what is inside it without any explicit process for creating it.

__tests__/hooks/use-display-currency.spec.tsx
nicolasburtey: yup I have been thinking through this. I think is the `{ loading }` value on the `useQuery` is true, we may want to return undefined.

it's subtle because:
- beyond the undefined, what should we show in that case UX wise?
- we have to separate the use case loading with the use case where the query has an error (value could still be undefined, but we should show something else on the screen)

UncleSamtoshi (approved):

__tests__/hooks/use-display-currency.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+    (mocks) =>
+    ({ children }: PropsWithChildren) =>
+      (
+        <IsAuthedContextProvider value={true}>
+          <MockedProvider mocks={mocks}>{children}</MockedProvider>
+        </IsAuthedContextProvider>
+      )
+
+describe("usePriceConversion", () => {
+  describe("testing moneyAmountToMajorUnitOrSats", () => {
```
Extheoisah, line 137: I'm not sure why we have nested `describe` functions when we can have one??

nicolasburtey: nested describe should not be an issue when it's bringing useful context, you can look into the backend, it's a common practice there

PR #1670, opened 2023-03-15
feat: add max button UI 

GildedPleb:
This is the first part of an ongoing, or draft PR, addressing adding a max spend button to the send screen. 

A few notes on this first run:

- This "fixes" storybook: 
  1. Storybook for the send details and send confirmation screen where presented on backgrounds that did not match the app, now they do.
  2. It removes all the `// eslint/@ts err/ignore` comments and replaces them with appropriate types
  3. Lastly it adds a storybook flag, default to false. Storybook was breaking on the details screen because `{ returnPartialData: true }` busts the cache for mocking when calling for price. `cache-first` is not needed because it is the Apollo default, and `returnPartialData: true` is needed to prevent flickering in the prod app. So, the storybook flag is used to toggle adding returnPartialData. 
- The `storybook` flag can also be used to release this UI to prod without piping the actual functionality, which has not yet been built. (using the flag to merge this isn't added, but its easy.)
- It seems that `<=` was either removed from `lessThanOrEqualTo` or something else switched there. I have added it back in because it seems to be semantic, and for sending max leaving it as '<' causes an error. I could be mistaken, so, am i missing something here?
- I took the liberty of using the `max send` button to force the swap to be in the currency of the account displayed on the top, not doing this presented some conversion and value miss-matches. (i have not internalized the vocabulary yet, so i don't know what any of this is supposed to be called), if the swap is tapped after max, the send-max state is removed. 
- I also took the liberty of moving the swap currency call back out of the render to be more idiomatic.


nicolasburtey:
thanks, @GildedPleb 

could you fix the `check-code`?

Maybe @UncleSamtoshi can have a look. you would have the best context on this one?

app/i18n/en/index.ts
```diff
@@ -596,6 +596,9 @@ const en: BaseTranslation = {
     notConfirmed:
       "Payment has been sent\nbut is not confirmed yet\n\nYou can check the status\nof the payment in Transactions",
     note: "Note or label",
+    sendMax: "send max",
```
UncleSamtoshi, line 599: Can lines 599 and 601 start off with a capitalized letter.

GildedPleb:
Should be good to go now with that last commit as checks are passing locally...

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -637,13 +690,20 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
           {chooseWalletModal}
         </View>
         <View style={Styles.fieldContainer}>
-          <Text style={Styles.fieldTitleText}>{LL.SendBitcoinScreen.amount()}</Text>
+          <View style={Styles.amountHeader}>
```
UncleSamtoshi, line 693: One thing to keep in mind is that the `Send Max` functionality isn't universally available right now. Currently it can only happen with onchain spends from your btc wallet (using the OnChainPaymentSendAllMutation), as well as intraledger payments with their normal mutation because they have no fee. It will not work with lightning payments right now as there is no mutation for it yet.

app/types/amounts.ts
```diff
@@ -71,7 +71,7 @@ export const lessThanOrEqualTo = <T extends WalletOrDisplayCurrency>({
   value: MoneyAmount<T>
   lessThanOrEqualTo: MoneyAmount<T>
 }) => {
-  return value.amount < lessThanOrEqualTo.amount
+  return value.amount <= lessThanOrEqualTo.amount
```
UncleSamtoshi, line 74: Good catch!

UncleSamtoshi (changes requested):
Looking good so far! Left a few comments.

app/i18n/en/index.ts
GildedPleb: Done

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
GildedPleb: Indeed. I haven't fully digested how `send-max` works, so this is a good to note! One thing that may make this "universal" is to compare the final send on the confirm UI with the total wallet balance. If they are the same, then `max-send` is assumed. And the final send would become the total send (minus fee). As such, even if the BE doesn't have it for all payment types it can still be "faked" for all payment types. I am not sure if this is desirable, and was actually writing up some thoughts in the chat as you commented. Will post those soon. But i def need to do more research on how all the piping works for max-send.

GildedPleb:
This most recent commit adds a few things:
- Additional Storybook stories for various sending types on the sending screens
- Gen's for onChainPaymentSendAll
- New Mocking for fee finding 
- Piping for maxSend flag to send payment mutations 
- Send Max warning on confirmation (only for onchain)

And importantly, 

An MVP for sweeping the wallet if the transaction is onchain, and onchain only.

bodymindarts:
Please follow https://www.conventionalcommits.org/en/v1.0.0/ for your commit messages.

UncleSamtoshi (commented):
PR is looking really good so far! I left one comment. The only other issue is how to handle the confirmation page, but it looks like that conversation is happening in the mattermost channel, so I will post my thought over there.

app/screens/send-bitcoin-screen/payment-details/onchain.ts
```diff
@@ -184,7 +198,21 @@ export const createAmountOnchainPaymentDetails = <T extends WalletCurrency>(
   }
 
   if (sendingWalletDescriptor.currency === WalletCurrency.Btc) {
-    const sendPayment: SendPayment = async (sendPaymentFns) => {
+    const sendPayment: SendPayment = async (sendPaymentFns, sendingMax) => {
+      if (sendingMax) {
```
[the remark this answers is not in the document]
GildedPleb: Great catch! 

TIL BBW effectively has onchain bitcoin invoices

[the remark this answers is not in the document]
GildedPleb: Done

nicolasburtey:
closing this PR as another Pr has been merged that implemented the sendMax for onchain

Issue #848, opened 2023-01-12
react-native-keychain


nicolasburtey:
it seems to no longer be used

2624789:
@nicolasburtey I've been going through the app authentication logic and also looking where is keychain already used.

I saw that [app/utils/storage/secureStorage.ts](https://github.com/GaloyMoney/galoy-mobile/blob/main/app/utils/storage/secureStorage.ts) uses [react-native-secure-key-store](https://github.com/pradeep1991singh/react-native-secure-key-store) but [DeviceAccountModal](https://github.com/GaloyMoney/galoy-mobile/blob/main/app/screens/get-started-screen/device-account-modal.tsx) is using [react-native-keychain](https://www.npmjs.com/package/react-native-keychain).

I would lean to use [react-native-keychain](https://www.npmjs.com/package/react-native-keychain) because [react-native-secure-key-store](https://github.com/pradeep1991singh/react-native-secure-key-store) last update was two years ago. Although, [react-native-keychain](https://www.npmjs.com/package/react-native-keychain) use is limited to the [DeviceAccountModal](https://github.com/GaloyMoney/galoy-mobile/blob/main/app/screens/get-started-screen/device-account-modal.tsx) component, while [react-native-secure-key-store](https://github.com/pradeep1991singh/react-native-secure-key-store) is used more broadly in the app through the [KeyStoreWrapper](https://github.com/GaloyMoney/galoy-mobile/blob/main/app/utils/storage/secureStorage.ts).

I would like to know your opinion about which library would be a better fit to store the token using keychain. Do you think it is a good idea to extend the existing **KeyStoreWrapper** and keep using **react-native-secure-key-store**? Another alternative would be to refactor **KeyStoreWrapper** to use **react-native-keychain** and unify the use of keychain in the app through this class.

nicolasburtey:
another key aspect to keep in mind is to the upgrade path for existing user. we don't want them to having to log back in when we do this upgrade

2624789:
I agree that for maintainability it is better to have just one library. I'll focus on the feature of this issue that is storing the token on the keychain/keystore without users having to log back. The easiest way I find now for doing this would be to use [react-native-secure-key-store](https://github.com/pradeep1991singh/react-native-secure-key-store).

Removing one of the two libraries and improve maintainability could be a good task for another issue.

designsats:
Resolved — in two parts, both landed.

[...]

PR #1688, opened 2023-03-16
feat: adding darkmode

nicolasburtey:
image

the pattern I have used is the following:
when a styles has a dark/light mode, duplicate the styles and change the property. when this styles is called, wrap it around a `darkMode ? styleDark : styleLight`.

there could be another approach which is to change a theme, that have just some palette change across the app, but I think the approach would lack the granuarily sometime neededs. there is not necessarily a 1-1 mapping from light to dark mode (ie: lightGrey become darkGrey, it's the case most of the time, but on some occasion we need to put dark for instance)

some additional note:

- I'm starting to remove `EStyleSheet` for `StyleSheet` (default, from react native). the main reason is that `EStyleSheet` doesn't seem to type the styles correctly, and `StyleSheet` does. Also it seems `EStyleSheet` is no longer updates, and I don't think we use a lot (if any?) of the extensibility `EStyleSheet` provides. 
- some screen doesn't use the `<Screen />` Component. I've moved to use the Screen because it will provide the correct background for darkMode/lightMode and is a step forward for a cleaner implementation
- some screen are "hacky" in how they appear in darkMode, mostly in the send/receive/home/conversion flows that should change soon. I've favored having some darkMode shortly versus trying to do it perfect for the screens that we now we'll be working back on soon
- inconsistency between style at top or bottom of files
- scroll view inconsistency. should the below button be part of the scrollview?


app/graphql/local-schema.gql
```diff
@@ -15,7 +15,13 @@ extend type Query {
   hiddenBalanceToolTip: Boolean!
   price: String # FIXME test only?
   beta: Boolean!
-  darkMode: Boolean!
+  darkMode: AppearanceMode!
+}
+
+enum AppearanceMode {
```
UncleSamtoshi, line 21: I've brought this up before, but even if we switch to not clearing the apollo cache on every update, I'm still not sure if this is where the data should live. It seems like it would be jarring if a user logs out and then all of a sudden the theme switches.

app/screens/settings-screen/settings-screen.tsx
```diff
@@ -140,6 +143,24 @@ export const SettingsScreen: React.FC = () => {
     setIsContactModalVisible(!isContactModalVisible)
   }
 
+  const switchDarkMode = () => {
```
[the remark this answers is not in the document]
nicolasburtey: this commit should fix it: https://github.com/GaloyMoney/galoy-mobile/pull/1688/commits/9a0d4e2ea79992f48aaf9b8cbca96c7a3aa8be44

app/graphql/local-schema.gql
nicolasburtey: > It seems like it would be jarring if a user logs out and then all of a sudden the theme switches.

I don't think it's really a big issue, for a couple of reasons:
- users don't log out very frequently. devs do it during the development process, but user typically never ever log out
- users typically don't change the dark mode settings from an app. I don't recall ever doing it, expect when I start implementing this functionality, and I saw that WOS have this functionality for instance, but I never played with it before. I think most of the time app rely (exclusively?) on the OS settings for this. As we implement now the OS setting by default, most users would anyway rely on it and probably have little reasons to override the settings

Now, the frequently at which the combination of those elements may occur may be very rare, how critical is the issue? ok an user may see the theme changes on logout, and may be surprise by it, but is it inconvenient or critical in any way? I don't think so.

so all in all, this event is very unlikely with a low effect on user, so I don't think this is worth the complexity of storing the override value outside of where we store all of our other content (graphql).

now maybe it may be worth why I've spent time refactoring the PersistentStateProvider relied on "@react-native-async-storage/async-storage". this means this is not encrypted, which is not ideal for storing password. 

Previously we were using react-native-keychain (created a ticket to be reminded to re-introduce it https://github.com/GaloyMoney/galoy-mobile/issues/848), and this is the right place to store password/token (I don't know why we got rid of it? Maybe you know?). it will be encrypted and can't be extracted from someone who would have physical access to the phone because it uses the secure element in the phone. 

keychain is limiting on how much data can be stored. but currently, as we are back to only have server+password, we could be using back the keychain now. With the previous element where we had a lot more information stored in the json, this was not possible.

this is another reason (and the main one imo) to differentiate where we store token value and the rest of the values in the app. we could have 3 storages, token via keychain, json that is not tied to graphql, and the graphql store, but really I see of complexity doing that and not much downside to not have the "json that is not tied to graphql".

nicolasburtey:
test fix on this PR: https://github.com/GaloyMoney/galoy-mobile/pull/1744

app/graphql/local-schema.gql
UncleSamtoshi: Initially I thought the 3 separate storages made conceptual sense, but in reality its probably an indication that the "json not tied to gql" should actually be stored as account settings in the backend. Just as we have language and currency in the backend, it makes sense to store theme there as well. However, I agree that the rarity of the situation may lead to it not being worth the implementation.

app/rne-theme/colors.ts
```diff
@@ -1,12 +1,14 @@
-const colors = {
+import { palette } from "@app/theme"
+
+const light = {
   primary: "#1627C4",
-  primary3: "#1627C4",
```
UncleSamtoshi, line 3: What was the thinking behind changing these colors? The naming came from our Figma design and existing components already depend on these names.

app/rne-theme/colors.ts
```diff
@@ -39,4 +41,45 @@ const colors = {
...
+  error8: "#FECACA",
+  error9: "#FEE2E2",
+  warning: "#F59E0B",
+  warning4: "#F59E0B",
+  warning9: "#FEF3C7",
+  white: "#000000", // <-- this is the only change
+  black: "#FFFFFF", // <-- this is the only change
+  grey5: "#61637A",
+  grey7: "#9292A0",
+  grey8: palette.lightGrey, // <-- this is the only change
```
UncleSamtoshi, line 71: Are all other colors okay to stay the same in dark mode? Both the new currency input and Theo's home screen components use this new color theme. We will be blocked on merging our PRs until dark mode is supported, and that will require having the dark theme completed.

app/rne-theme/colors.ts
UncleSamtoshi: I see in a separate commit you updated the existing components. There is still a cognitive overhead because all of the components in Figma use this naming.

app/graphql/theme-setup.tsx
```diff
@@ -0,0 +1,40 @@
+import { useThemeMode } from "@rneui/themed"
+import React from "react"
+import { Appearance } from "react-native"
+import { useDarkModeQuery } from "./generated"
+
+export const ThemeSync = () => {
+  const { mode, setMode } = useThemeMode()
+  const { data } = useDarkModeQuery()
+  const darkMode = data?.darkMode
```
UncleSamtoshi, line 9: Shoudn't this component just the `useDarkMode` hook and whenever the useDarkMode hook is different than the useThemeMode value, it should set the theme. The logic is duplicated somewhat from that hook and it adds complexity and introduces the possibility of the useDarkMode hook getting out of sync with the theme provider.

app/rne-theme/colors.ts
nicolasburtey: the idea is to have them in the right order in storybook

image

didn't know the naming was coming from Figma. I can revert it.

app/graphql/theme-setup.tsx
nicolasburtey: I'm thinking to remove this code (everything related to graphql and darkmode) when it goes to prod. and solely rely on the OS settings. 

but for now maybe we can keep it while we're in development, to make it easier to switch between theme (without having to go to the OS settings, which also changes all the other apps design... quite cumbersome)

app/rne-theme/colors.ts
nicolasburtey: I'm settings the darkMode colors as I use them right now. 

it helps me to understand how they translate from light to dark. it's not always a 1-1 mapping in my design to darkMode (not ideal but we'll have to think about it more later)

app/graphql/theme-setup.tsx
```diff
@@ -0,0 +1,40 @@
+import { useThemeMode } from "@rneui/themed"
+import React from "react"
+import { Appearance } from "react-native"
+import { useDarkModeQuery } from "./generated"
+
+export const ThemeSync = () => {
+  const { mode, setMode } = useThemeMode()
+  const { data } = useDarkModeQuery()
```
UncleSamtoshi, line 8:
```suggestion
  const isDarkMode = useDarkMode()
  
  useEffect(()=> {
  const intendedMode = isDarkMode ? "dark" : "light"
  if (intendedMode !== mode) {
      setMode(intendedMode)
  }
  }, [isDarkMode, mode])
```
Could this not replace the entire file and be functionally the same?

app/rne-theme/colors.ts
UncleSamtoshi: I think it would be ideal to revert it. If we want it sorted we can manually specify the order in the component.

app/rne-theme/colors.ts
nicolasburtey: maybe I'm mistaken what you're saying, but the current component are quite inconsistent already (ie: some background screen are sometimes `white`, sometime `lighterGrey`, and so the rest of the button are not consistent at all. there is quite a bit of refactoring work to do on the styles but I think this is part of the redesign to bring more consistency). I don't think we should be wary to "break" existing design, we can fix it as we go as we're refactoring other screen to introduce the new design.

UncleSamtoshi: I believe the current components don't use this theme at all, but the new redesigned components do. That's why I would keep this theme in a good state so that it works for the new components. For the old components, I wouldn't use the the theme object.

nicolasburtey:
- I have revert the commit that shifted the number for grey in the theme/colors
- I remove most of the logic around darkMode value, and solely relies on the OS for now. It doesn't work on the simulator thought, I'm not sure why. it says that this is a known behavior for Chrome. didn't read anything on react native, but I think we can merge it and I can test on testflight. worse thing it'll be a "no-op" (ie: stay in light mode).

app/rne-theme/colors.ts
nicolasburtey: > That's why I would keep this theme in a good state so that it works for the new components

I don't think the new theme is very understandable as of now because of how the type is been done.

here is an example of all the typed showing up:

image

yet, they're really mixed

image

Extheoisah (changes requested):
I can apprive if @UncleSamtoshi comments have been addressed

Extheoisah (approved):

PR #1701, opened 2023-03-17
chore: improve e2e tests login and welcome screen flow

Extheoisah:
This PR adds some improvements to the e2e test. Some of the code in the login flow were actually not working properly e.g. https://github.com/GaloyMoney/galoy-mobile/pull/1701/files#diff-8ffce1538d4d0c0cd9a295395d3e111073f9cf483a2b104d9ddf92bba152b07eL68
 
Also, somehow the displayCurrency of some of the test accounts was not USD by default which caused some the send flow tests to fail ([see here for browserstack video](https://app-automate.browserstack.com/sessions/072ac3a75edbcf547362b0a9597aa153b0aa528f/video?token=UlNFT1dybThSbVAya0UvNTMwaHVhTG9qbW93SVlGWDdzK3FKMzNsbEVTdkdaT1Z2bDkyekJFNXhQZm5NL1hhTzZuaHBrdzl3WENCTVE2ODVNbFVYdFE9PS0tSmIrNDJKOG4zbTd2QzQwenZ5Ymp3UT09--f5829c3f5a15b4d3909a5f36a11aeef0a637d917&source=rest_api&diff=30.073861999)). Maybe someone used the token locally, changed it to different currency??. Regardless, I have added a function to reset the display currency.

### DIFF
--- a/e2e/02-login-flow.e2e.spec.ts
+++ b/e2e/02-login-flow.e2e.spec.ts
@@ -62,13 +62,12 @@ describe("Login Flow", () => {
 
   it("click Save Changes", async () => {
     const changeTokenButton = await $(selector("Save Changes", "Button"))
-    await changeTokenButton.waitForEnabled({ timeout })
+    await changeTokenButton.waitForDisplayed({ timeout })
     await changeTokenButton.click()
     // wait for token to be saved
-    await browser.pause(3000)
+    await browser.pause(2000)
     const tokenPresentText = await $(selector("Token Present", "StaticText"))
     const tokenPresent = await tokenPresentText.getText()
-    console.log("tokenPresent", tokenPresent)
     await browser.waitUntil(async () => tokenPresent.includes("true"))
   })
 
@@ -92,8 +91,8 @@ describe("Login Flow", () => {
   })
 
   it("navigates back to move home screen", async () => {
-    // scroll down to show phone setting for small screens
-    await scrollDown()
+    // scroll up to show phone setting for small screens
+    await scrollUp()
     const phoneSetting = await $(selector(LL.common.phoneNumber(), "StaticText"))
     await phoneSetting.waitForDisplayed({ timeout })
     const backButtonOnSettingsScreen = await $(goBack())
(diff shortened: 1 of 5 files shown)


e2e/02-login-flow.e2e.spec.ts
```diff
@@ -86,9 +94,12 @@ describe("Login Flow", () => {
     expect(logoutButton.isDisplayed()).toBeTruthy()
     const backButtonOnAccountScreen = await $(goBack())
     await backButtonOnAccountScreen.click()
+    await browser.pause(2000)
```
UncleSamtoshi, line 97: Same here

Extheoisah:
> Any idea why the tests are currently failing on Android?

Yeah. The list of buttons in the settings screen does not currently fit into the screen all together. So when we scroll down to enter the debug screen and come back to the settings screen, the phone number button(which we also use to validate that the user is logged in) is no longer in focus. For bigger screens, this is not an issue. So I added the scroll command in this PR to scroll up to show the phone button for smaller screens.

e2e/02-login-flow.e2e.spec.ts
Extheoisah: I couldn't find a way to use the `browser.waitUntil()` fn here. We need this pause esp for android because there is small lag when we move to the settings screen from the debug screen.

UncleSamtoshi: Is there something on the settings screen you could wait until it was displayed? There is also the `waitForExist` method I believe.

nicolasburtey:
@UncleSamtoshi should we merge this one? it seems it's fixing the e2e on android (at least partially?)

Extheoisah:
> @UncleSamtoshi should we merge this one? it seems it's fixing the e2e on android (at least partially?)

This PR was originally intended to fix the android tests. We can merge it if its okay by you and @UncleSamtoshi 
Might beed a rebase though

Extheoisah:
Closing this PR in as the tests has been fixed here https://github.com/GaloyMoney/galoy-mobile/pull/1742

PR #1706, opened 2023-03-17
feat: hide decimals for inflated currencies

UncleSamtoshi:
- if a currency's dollar has less value than a usd cent, do not display decimals

### DIFF
--- a/__tests__/screens/send-confirmation.spec.tsx
+++ b/__tests__/screens/send-confirmation.spec.tsx
@@ -17,7 +17,5 @@ it("SendScreen Confirmation", async () => {
   await act(async () => {})
 
   const { children } = await findByLabelText("Successful Fee")
-  expect(children).toEqual(["₦0.00 ($0.00)"])
-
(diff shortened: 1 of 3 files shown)


Extheoisah (approved):
Looks good

nicolasburtey (commented):
just skimmed thought the PR. so that I understand, currently it's not a setting correct? it will just show the 0 for the currency when amount is below 0.005?

if I understand correctly, the issue with that is sometime 0 should not the answer, if you're paying `5 sats` in fees, it should be shown `$0`, but should be shown `< $0.01`

UncleSamtoshi:
@nicolasburtey 

> just skimmed thought the PR. so that I understand, currently it's not a setting correct? it will just show the 0 for the currency when amount is below 0.005?

Its unrelated to the amount. It checks if the value of a display currency major unit is less than 1 cent. If that is the case, then it will not show decimals on any of the amounts because the major unit is already sufficiently small.

Issue #1166, opened 2023-02-17
"a bug? submit a ticket" in the settings


nicolasburtey:

Ayush0054:
Hey, can I work on this issue? I will show you my approach and then make a pull request

nicolasburtey:
please do!

Ayush0054:
approach :go to setting -> click on "raise ticket" ->navigated to raise ticket screen -> then input your bug through text or images ->  "ticket is raised" feedback message

Ayush0054:
hey @nicolasburtey ,  Could you tell me more about how the ticket is going to be handled? For example, will it be handled using the above approach, an external API, or a custom chat? Any context or leads would be enough for me.

nicolasburtey:
Didn't think thought it much

I think just opening the GitHub page can be enough. Maybe we can add a message saying we are open source which is why it will be opening the webpage to GitHub

Ayush0054:
ok

Ayush0054:
Redirecting to this page ?  Any feedback 
![Screenshot_20230322_033954_Mattermost.jpg](https://user-images.githubusercontent.com/97244608/226753077-01e9d43f-b0c6-4675-92ed-9228d9fd3567.jpg)

Ayush0054:
Sure, I will probably start contributing after some time since I am currently doing an internship and working on personal projects.

PR #1720, opened 2023-03-22
Feat: "a bug? submit a ticket" in the settings #1166

Ayush0054:
Feat #1166 

image

https://user-images.githubusercontent.com/97244608/226863542-a7897e73-00e2-43ae-9d96-40f69f3fa264.mp4


nicolasburtey:
we're going to implement this with a normal text box that will send a message directly to use. the idea of using github requires an account which make it a bit harder to give feedback for most of our users which are not developers

PR #1742, opened 2023-03-24
feat: redesign wallet overview component

Extheoisah:
Simulator Screen Shot - iPhone 13 Pro - 2023-03-24 at 16 34 22


app/components/wallet-overview/wallet-overview-redesigned.tsx
```diff
@@ -0,0 +1,214 @@
...
+import { gql } from "@apollo/client"
+import {
+  useHideBalanceQuery,
+  useWalletOverviewScreenQuery,
+  WalletCurrency,
+} from "@app/graphql/generated"
+import { useIsAuthed } from "@app/graphql/is-authed-context"
+import { useDisplayCurrency } from "@app/hooks/use-display-currency"
+import { toBtcMoneyAmount, toUsdMoneyAmount } from "@app/types/amounts"
+// import { testProps } from "@app/utils/testProps"
```
UncleSamtoshi, line 14: Commented code should be deleted

app/components/wallet-overview/wallet-overview-redesigned.tsx
```diff
@@ -0,0 +1,214 @@
...
+      foregroundColor="#ecebeb"
+    >
+      <Rect x="0" y="0" rx="4" ry="4" width="100%" height="100%" />
+    </ContentLoader>
+  </View>
+)
+
+const HidableArea = ({ hidden, style, children }: HidableAreaProps) => {
+  return (
+    <TouchableHighlight style={style} underlayColor="#ffffff00">
```
UncleSamtoshi, line 43: We shouldn't have hard coded colors in components but should instead rely on the theme.

UncleSamtoshi, line 43: Does this need to be a`TouchableHighlight`? There is no action being passed in for `onPress`.

app/components/wallet-overview/wallet-overview-redesigned.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const btcWalletBalance = toBtcMoneyAmount(
+      data?.me?.defaultAccount?.btcWallet?.balance ?? NaN,
+    )
+
+    const usdWalletBalance = toUsdMoneyAmount(
+      data?.me?.defaultAccount?.usdWallet?.balance ?? NaN,
+    )
+
+    btcInDisplayCurrencyFormatted =
+      moneyAmountToDisplayCurrencyString(btcWalletBalance) ?? "..."
```
UncleSamtoshi, line 106: Does `moneyAmountToDisplayCurrencyString` ever return a falsy value? Otherwise I don't think we will get the "..." ever.

app/components/wallet-overview/wallet-overview-redesigned.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    usdInDisplayCurrencyFormatted =
+      moneyAmountToDisplayCurrencyString(usdWalletBalance) ?? "..."
+
+    btcInUnderlyingCurrency = formatMoneyAmount(btcWalletBalance)
+
+    if (displayCurrency !== WalletCurrency.Usd) {
```
UncleSamtoshi, line 113: There is a function from `useDisplayCurrency` called `getSecondaryAmountIfCurrencyIsDifferent` that can be used to replace lines 113-115.

app/components/wallet-overview/wallet-overview-redesigned.tsx
```diff
@@ -0,0 +1,214 @@
...
+      usdInUnderlyingCurrency = formatMoneyAmount(usdWalletBalance)
+    }
+  }
+
+  return (
```
UncleSamtoshi, line 118: Are there any storybook stories that show the loading state?

app/screens/home-screen/home-screen.tsx
```diff
@@ -341,16 +255,18 @@ export const HomeScreen: React.FC = () => {
...
 
       <View style={styles.walletOverview}>
-        <WalletOverview loading={loading} setModalVisible={setModalVisible} />
+        <WalletOverview
+          loading={loading}
+          setIsStablesatModalVisible={setIsStablesatModalVisible}
+        />
       </View>
-
       <FlatList
```
UncleSamtoshi, line 270: I would get rid of the flat list and just have a row of all the icon buttons. The refresh then needs to be added as some 'onScrollDown` prop on the screen.

UncleSamtoshi (changes requested):
Looks good! Left a few comments.

app/components/wallet-overview/wallet-overview-redesigned.tsx
Extheoisah: No. But I can add that

app/screens/home-screen/home-screen.tsx
```diff
@@ -227,21 +139,17 @@ export const HomeScreen: React.FC = () => {
...
+  const TRANSACTIONS_TO_SHOW = 2
 
   if (isAuthed && transactionsEdges?.length) {
     recentTransactionsData = {
       title: LL.TransactionScreen.title(),
-      target: "transactionHistory",
-      style: styles.transactionViewButton,
       details: (
-        <View style={styles.transactionsView}>
+        <View>
```
nicolasburtey, line 152: you can use a `<>` component instead of the view if you don't need any style.

app/components/wallet-overview/wallet-overview-redesigned.tsx
```diff
@@ -0,0 +1,211 @@
...
+  },
+  displayTextView: {
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "space-between",
+    marginVertical: 5,
+  },
+  separator: {
+    height: 1,
+    backgroundColor: mode === "dark" ? colors.lighterGreyOrBlack : colors.primary9,
```
nicolasburtey, line 192: this is quite hacky. ideally we should have just one color and its color automatically switch between modes. 

I know it's a pattern that result from our current state of redesign. we can address this in another PR later.

nicolasburtey (commented):
few comments/questions

Extheoisah:
> few comments/questions

I have addressed your comments. Pls re-review.

UncleSamtoshi (approved):

PR #1748, opened 2023-03-27
feat: new currency input

UncleSamtoshi:
- create a custom number pad
- use the the custom number pad in the send flow
- create a params object for `formatMoneyAmount` fn


app/components/amount-input-screen/amount-input-screen-ui.tsx
```diff
@@ -0,0 +1,188 @@
...
+  secondaryCurrencyFormattedAmount,
+  secondaryCurrencyCode,
+  errorMessage,
+  onKeyPress,
+  onToggleCurrency,
+  onClearAmount,
+  onSetAmountPress,
+  setAmountDisabled,
+  goBack,
+}: AmountInputScreenUIProps) => {
```
nicolasburtey, line 41:
```suggestion
export const AmountInputScreenUI: React.FC<AmountInputScreenUIProps> = ({
  primaryCurrencySymbol,
  primaryCurrencyFormattedAmount,
  primaryCurrencyCode,
  secondaryCurrencySymbol,
  secondaryCurrencyFormattedAmount,
  secondaryCurrencyCode,
  errorMessage,
  onKeyPress,
  onToggleCurrency,
  onClearAmount,
  onSetAmountPress,
  setAmountDisabled,
  goBack,
}) => {
```

nicolasburtey: this also ensure the return type to be compatible. I don't think it's critical but this is the pattern I've been using with the refactor I've done.

app/components/amount-input-screen/number-pad-reducer.ts
```diff
@@ -0,0 +1,156 @@
...
+  1: "1",
+  2: "2",
+  3: "3",
+  4: "4",
+  5: "5",
+  6: "6",
+  7: "7",
+  8: "8",
+  9: "9",
+  Decimal: ".",
```
nicolasburtey, line 52: in Europe and probably other locale, this would be a `,` (I think). just fyi, I think we should focus on those i18n details later

nicolasburtey:
the padding seems off (not enough on the horizontal side). I feel the button are too close to the edges.

image

here is what cashapp looks like. there a bit more space on the left / right of the screen

image

also, do we need a "Enter Amount"? feels like it's quite obvious and this instruction may not be needed?

nicolasburtey:
no error have a red cross? 

that sounds weird to me

image

app/components/amount-input-screen/amount-input-screen.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { WalletCurrency } from "../../graphql/generated"
+import mocks from "../../graphql/mocks"
+import {
+  DisplayCurrency,
+  MoneyAmount,
+  WalletOrDisplayCurrency,
+} from "../../types/amounts"
+
+export default {
+  title: "Amount Input Screen",
```
nicolasburtey, line 16: why do we need 2 stories book? feels the same between `Amount Input Screen UI` and `Amount Input Screen`

nicolasburtey (commented):
some nits, but PR looks good otherwise

app/components/amount-input-screen/amount-input-screen-ui.tsx
UncleSamtoshi: I can make those changes.

UncleSamtoshi:
> also, do we need a "Enter Amount"? feels like it's quite obvious and this instruction may not be needed?

I think the screen looks a little weird if you remove it and the header contains just an x

UncleSamtoshi:
> the padding seems off (not enough on the horizontal side). I feel the button are too close to the edges.

I increased the padding.

app/components/atomic/galoy-warning/galoy-warning.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <UseCase text="Warning with highlight">
+        <GaloyWarning errorMessage="Something went wrong" highlight />
+      </UseCase>
+      <UseCase text="Warning with no icon">
+        <GaloyWarning errorMessage="Something went wrong" noIcon />
+      </UseCase>
+      <UseCase text="Really long warning">
+        <GaloyWarning
+          highlight
+          errorMessage="Something that is really long and takes up a lot of space to see what it looks like when it is really long and takes up a lot of space"
```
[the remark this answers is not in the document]
UncleSamtoshi: At this stage the component isn't designed to handle something with the length of a stack trace. It could be a good follow up PR.

app/components/amount-input-screen/amount-input-screen.stories.tsx
UncleSamtoshi: They're very similar. I developed the component by first creating the `AmountInputScreenUI`, which had no logic in it, and I was very easily able to test all permutations of it by just passing in different props without having to mock anything. `AmountInputScreen` uses app state, and therefore requires more boiler plate to be able to exercise the same permutations. I suppose `AmountInputScreen` is what we ultimately want to test, so even though there is more boiler plate there it could make sense to remove the `AmountInputScreenUI` stories and build them out in `AmountInputScreen`.

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -364,123 +365,118 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
...
+              </View>
+            </View>
+            <View style={styles.walletSelectorInfoContainer}>
+              <View style={styles.walletSelectorTypeTextContainer}>
+                {sendingWalletDescriptor.currency === WalletCurrency.Btc ? (
+                  <>
+                    <Text style={styles.walletCurrencyText}>
+                      {LL.common.btcAccount()}
+                    </Text>
+                  </>
```
nicolasburtey, line 432:
```suggestion
                  <Text style={styles.walletCurrencyText}>
                    {LL.common.btcAccount()}
                  </Text>
```

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -364,123 +365,118 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
...
+                ) : (
+                  <>
+                    <Text style={styles.walletCurrencyText}>
+                      {LL.common.usdAccount()}
+                    </Text>
+                  </>
```
nicolasburtey, line 438:
```suggestion
                  <Text style={styles.walletCurrencyText}>
                      {LL.common.usdAccount()}
                    </Text>
```

nicolasburtey (approved):

nicolasburtey:
> > also, do we need a "Enter Amount"? feels like it's quite obvious and this instruction may not be needed?
> 
> I think the screen looks a little weird if you remove it and the header contains just an x

if you look at the CashApp screenshot, there is no "enter amount".  the main reason I'm not sure about it is because it'll felt like if you tap on "Enter Amount", you may expect some response like if it was a text field where you can add something.

I think that is ok for now, let's use this screen that is already much better and we can iterate later.

nicolasburtey:
some nits but approved nonetheless. can be addressed later.

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
UncleSamtoshi: I will address this in a follow up PR.

Issue #622, opened 2022-09-13
Copying onchain address leads to popup saying invoice copied


krtk6160:
**Describe the bug**
When I tapped the QR code of the onchain address, the resulting popup said `invoice has been copied in the clipboard`, which confused me.

**To Reproduce**

1. Go to receive screen
2. Click on 'Use a Bitcoin onchain address'
3. Tap on QR code
4. See wrong popup

**Expected behavior**
Popup should say bitcoin address copied to clipboard

**Smartphone (please complete the following information):**

- Device: Galaxy S21
- OS: Android 12 (OneUI 4.0)
- Version 2.0.34 Build 343

PR #1749, opened 2023-03-28
Fix: Receive screen copy msg fixed

Aditi-Singh16:
Fixes for #622

**Preview**

https://user-images.githubusercontent.com/74586135/228163058-40e8ba80-c7a6-4c8c-8653-0b9d638db1d6.mp4


Aditi-Singh16:
> the useMemo dependancy array needs to be updated

Ah yes, so I have used the already added variable then

nicolasburtey (approved):

PR #1755, opened 2023-03-29
ci: fix scripts

sandipndev:
Small fixes to CI script

### DIFF
--- a/ci/tasks/phased-rollout.sh
+++ b/ci/tasks/phased-rollout.sh
@@ -24,7 +24,7 @@ else
   export PREV_ROLLOUT_TIME=$(date +%s -d "$(cat ./timestamp)")
   export NOW_ROLLOUT_TIME=$(date +%s -d "$TIMESTAMP")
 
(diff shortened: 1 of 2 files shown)


ci/tasks/phased-rollout.sh
```diff
@@ -24,7 +24,7 @@ else
   export PREV_ROLLOUT_TIME=$(date +%s -d "$(cat ./timestamp)")
   export NOW_ROLLOUT_TIME=$(date +%s -d "$TIMESTAMP")
 
-  if [[ $(( $NOW_ROLLOUT_TIME - $PREV_ROLLOUT_TIME )) -lt 86400 ]]; then
+  if [[ $(( $NOW_ROLLOUT_TIME - $PREV_ROLLOUT_TIME )) -lt 80000 ]]; then
     echo "Can't rollout; 1 day hasn't yet passed from last rollout"
     exit 0
```
sandipndev, line 29: @krtk6160 This part checks if 1 day has passed since last phased rollout. Only then it agrees to move forward with the next phase of the rollout.
The 1 day exactly isn't met because of varying start times (2-3 min difference) of Concourse run after daily trigger. Therefore I changed `86400` to `80000` to almost match a day.

PR #1759, opened 2023-03-30
chore: fix amount in receive in darkMode

nicolasburtey:


UncleSamtoshi (dismissed):
Check code is failing, otherwise it looks good.

PR #1780, opened 2023-03-31
test: improve e2e config

UncleSamtoshi:
- logs network traffic in e2e tests
- ends the test run if a single test fails

### DIFF
--- a/e2e/config/browserstack.conf.js
+++ b/e2e/config/browserstack.conf.js
@@ -9,6 +9,10 @@ let capabilities = {
   "os_version": "9.0",
   "app": process.env.BROWSERSTACK_APP_ID,
   "browserstack.debug": true,
+  "browserstack.networkLogs": true,
+  "browserstack.networkLogsOptions": {
+    captureContent: true,
+  },
   "autoGrantPermissions": true,
 }
 if (process.env.E2E_DEVICE === "ios") {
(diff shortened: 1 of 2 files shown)


UncleSamtoshi:
Usually if a test fails it messes up many downstream tests. This PR makes is so that the test run will fail immediately instead of wasting time performing downstream tests. The alternative solution would be to do a better job resetting at the end of each test so that each test suite starts off from a known good state.

nicolasburtey:
> Usually if a test fails it messes up many downstream tests. This PR makes is so that the test run will fail immediately instead of wasting time performing downstream tests. It also makes debugging easier because the last test performed is the one that must be investigated.
> 
> The alternative solution would be to do a better job resetting at the end of each test so that each test suite starts off from a known good state. This could look like have an `afterAll` step that tries to always navigate to the home screen.
> 
> I do like the fail fast option for its quicker turn around time in testing results, but I am open to the `afterAll` solution as well.

another (small, but it adds up) side benefits is reduction of cost. failing tests can often last 30 min+, and each minutes is been billed.

nicolasburtey:
> * logs network traffic in e2e tests
> * ends the test run if a single test fails

approved and merging. we see if we like it better and re-assess after we have some insights in the behavior for day to day development.

nicolasburtey (approved):

PR #1890, opened 2023-04-07
feat: adding onchain stablesats send

nicolasburtey:


Extheoisah (commented):
Looks good to me.... I will test manually tomorrow

nicolasburtey:
> Looks good to me.... I will test manually tomorrow

did some tests and there are some issue with the case where the bitcoin amount is set in the QR code. will debug and put it back as ready when review when I fixed it.

nicolasburtey:
> > Looks good to me.... I will test manually tomorrow
> 
> did some tests and there are some issue with the case where the bitcoin amount is set in the QR code. will debug and put it back as ready when review when I fixed it.

ok fixed and tested, it's working properly now.

Extheoisah (approved):
Tested it and it works as expected!

PR #1920, opened 2023-04-13
fix: homescreen ui issues

Extheoisah:


app/components/atomic/galoy-primary-button/galoy-primary-button.tsx
```diff
@@ -26,8 +26,8 @@ export const GaloyPrimaryButton: FunctionComponent<PropsWithChildren<ButtonProps
 
 const useStyles = makeStyles((theme) => ({
   titleStyle: {
-    fontSize: 28,
-    lineHeight: 32,
+    fontSize: 27,
+    lineHeight: 28,
```
nicolasburtey, line 30: line height should not be just one more pixel than fontSize I think. multine text would touch top to bottom otherwise afaiu. 

does this needs to be changed?

app/components/wallet-overview/wallet-overview.tsx
```diff
@@ -165,7 +167,8 @@ const WalletOverview: React.FC<Props> = ({ loading, setIsStablesatModalVisible }
               type={usdInUnderlyingCurrency ? "p3" : "p1"}
               bold={!usdInUnderlyingCurrency}
             >
-              {`~ ${usdInDisplayCurrencyFormatted}`}
+              {usdInUnderlyingCurrency && `~ `}
+              {usdInDisplayCurrencyFormatted}
```
nicolasburtey, line 171: the logic is quite convoluted. why checking on `usdInUnderlyingCurrency` when `usdInDisplayCurrencyFormatted` is used?

the logic should also not be located in the JSX part.

app/screens/home-screen/home-screen.tsx
```diff
@@ -240,59 +269,25 @@ export const HomeScreen: React.FC = () => {
...
-        <View>
-          <WalletOverview
-            loading={loading}
-            setIsStablesatModalVisible={setIsStablesatModalVisible}
-          />
-        </View>
-
+        {error && (
+          <Text style={styles.error} selectable>
+            {error.message}
```
nicolasburtey, line 274: we use to parse the error differently. was there an issue before?

```
error ? (
            <Text style={styles.error} selectable>
              {error.graphQLErrors.map(({ message }) => message).join("\n")}
            </Text>
```

nicolasburtey (commented):
few questions and comments on the PR

app/components/hideable-area.tsx/hideable-area.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { ReactNode, useEffect, useState } from "react"
```
nicolasburtey, line 1: folder should not be name .tsx

nicolasburtey (changes requested):
need to rename folder

app/components/balance-header/balance-header.tsx
```diff
@@ -160,14 +162,18 @@ export const BalanceHeader: React.FC<Props> = ({ loading }) => {
...
-        <TouchableOpacity
-          onPress={() => setBalanceHidden(!balanceHidden)}
-          style={styles.hiddenBalanceTouchableOpacity}
-        >
-          <Icon style={styles.hiddenBalanceIcon} name="eye" />
-        </TouchableOpacity>
-      ) : (
+      <HideableArea
+        isHidden={balanceHidden}
+        hideBalance={hidden}
```
nicolasburtey, line 167: the variable used here are confusing

app/components/hideable-area.tsx/hideable-area.tsx
```diff
@@ -0,0 +1,37 @@
...
+    } else {
+      setShouldBeHidden(isHidden)
+    }
+  }, [isHidden, hideBalance])
+
+  if (shouldBeHidden) {
+    return <View style={styles}>{hiddenContent || <Text>****</Text>}</View>
+  }
+
+  return <View style={styles}>{children}</View>
```
nicolasburtey, line 34: there is imo no point of having a wrapper `<View style={styles}>` when this could be outside the component, ie from from the parent.

app/components/stablesats-modal/stablesats-modal.tsx
```diff
@@ -46,18 +47,18 @@ const styles = StyleSheet.create({
...
   homeButton: {
-    backgroundColor: palette.blue,
+    backgroundColor: colors.primary5,
     borderRadius: 12,
     padding: 16,
     marginBottom: 10,
   },
   learnMoreButton: {
-    backgroundColor: palette.white,
+    backgroundColor: "transparent",
```
nicolasburtey, line 56: no color hardcoding

app/components/wallet-overview/wallet-overview.tsx
```diff
@@ -98,27 +85,39 @@ const WalletOverview: React.FC<Props> = ({ loading, setIsStablesatModalVisible }
...
+    )}`
 
     btcInUnderlyingCurrency = formatMoneyAmount({ moneyAmount: btcWalletBalance })
 
     if (displayCurrency !== WalletCurrency.Usd) {
       usdInUnderlyingCurrency = formatMoneyAmount({ moneyAmount: usdWalletBalance })
     }
+
+    usdInDisplayCurrencyFormatted = `${
+      usdInUnderlyingCurrency ? "~" : ""
```
nicolasburtey, line 99: the logic of whether the ~ sign should be present or not should not be in the component imo. it should be in some function in the `useDisplayCurrency` hooks.

Extheoisah: IMO, I think the logic for adding the tilde sign (~) should be on the component level. We want the results of hooks to be pure and reusable in various components. Also, the computation for `usdInDisplayCurrencyFormatted = moneyAmountToDisplayCurrencyString(usdWalletBalance)` as well as the other two balances (`btcInDisplayCurrencyFormatted`, `usdInUnderlyingCurrency` and `btcInUnderlyingCurrency`) use functions from the `useDisplayCurrency` hook and they're done in the component level. Except we want to move these balance computations into the hook so that we get the values already formatted for just this screen.

nicolasburtey:
needs to be rebase

Extheoisah:
> needs to be rebase

rebased and addressed comments.

app/components/wallet-overview/wallet-overview.tsx
UncleSamtoshi: `formatMoneyAmount` takes in an argument called `isApproximate` that will add the `~`. I think it makes sense for the component to say if it is approximate, however all approximate numbers should be formatted the same way which is why it is part of the formatMoneyAmount fn.

Extheoisah: I wasn't aware of that functionality. Thanks.
I have refactored the code here https://github.com/GaloyMoney/galoy-mobile/pull/1920/commits/9d3b9fb730aad1c4b5919c75e30dd36e44f5e39a
cc @nicolasburtey

Extheoisah: fixed the duplicated code here https://github.com/GaloyMoney/galoy-mobile/pull/1920/commits/1694bd7bc72af327272893a20643b85bf72b9605

app/components/hideable-area/hideable-area.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { ReactNode } from "react"
+import { Text } from "@rneui/themed"
+
+interface HideableAreaProps {
+  children: ReactNode
+  isContentVisible?: boolean
```
nicolasburtey, line 6: why is this optional? there is probably no point of using the component without this value?

nicolasburtey (approved):

Issue #1866, opened 2023-04-03
home screen - wheel refresh when price updates


nicolasburtey:
if someone were to let the app open, the "refresh wheel" seems to re-appears everything the price refresh. we should only show the wheel on user initiated action

designsats:
Resolved by #4079 (merged 2026-08-06).

The refresh control was bound to a composite of background query loading flags (`loadingAuthed || loadingPrice || loadingUnauthed`), so every realtime-price refresh flipped the spinner on. It now reflects only user-initiated pulls via a dedicated `isPullRefreshing` state.

Covered by regression tests in `__tests__/screens/home.spec.tsx` ("HomeScreen pull-to-refresh").

grimen:
Closing as part of old-ticket cleanup (tracked in blinkbitcoin/blink-wip#1169) — no longer applicable/relevant to the current app.

PR #1991, opened 2023-04-17
Fix: Refresh only when user initiates

Aditi-Singh16:
#1866 still persists after the update

### DIFF
--- a/app/screens/home-screen/home-screen.tsx
+++ b/app/screens/home-screen/home-screen.tsx
@@ -233,7 +233,7 @@ export const HomeScreen: React.FC = () => {
         contentContainerStyle={styles.scrollView}
         refreshControl={
           <RefreshControl
-            refreshing={loading}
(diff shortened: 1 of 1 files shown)


Extheoisah:
Hi @Aditi-Singh16 
in my opinion, I think we should keep it as it because of the following reasons:
- users can see the latest data without needing to take action to refresh but can still force a refresh if desired.
- Allows for more efficient data fetching and a better user experience overall.

Also, a manual refresh has its cons such as:
- Users might not know when a refresh is needed and may miss important updates.
- Adds an extra action for users to perform, which could affect user experience.

Is there a reason you prefer a manual refresh?

Aditi-Singh16:
Oh okay, no specific reasons though, I just saw an open issue regarding this, hence created a PR. After looking at your reasons, I think it makes sense to have automatic refresh

PR #2045, opened 2023-04-18
chore: various ui improvements

UncleSamtoshi:
- handle larger fonts in conversion flow
- handle larger fonts in atomic components
- clean up send flow styling

### DIFF
--- a/.storybook/storybook.tsx
+++ b/.storybook/storybook.tsx
@@ -5,7 +5,6 @@ import { NavigationContainer } from "@react-navigation/native"
 import { ThemeProvider } from "@rneui/themed"
 import { createStackNavigator } from "@react-navigation/stack"
 import theme from "@app/rne-theme/theme"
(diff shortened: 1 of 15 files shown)


.storybook/storybook.tsx
```diff
@@ -47,7 +46,6 @@ export const StorybookUIRoot: React.FC = () => (
           />
         </Stack.Navigator>
       </NavigationContainer>
-      <ThemeSync />
```
nicolasburtey, line 50: this is useful to manage dark/light theme switch. have you use another solution?

nicolasburtey: ok this is what the new story is for

UncleSamtoshi: It may not be a better solution, but I was experimenting with it. Perhaps we could add the toggle as part of the base ui screen.

nicolasburtey (approved):

PR #2054, opened 2023-04-21
fix: 401 retry for database blip

ntheile:


nicolasburtey:
I'll recommend not going for those hacks and working throught the root cause

PR #2097, opened 2023-04-25
fix: ensure price and display currency are in sync

UncleSamtoshi:
Goal: Prevents scenario where the display currency for displaying amounts is different from the display currency in the price function which could cause a person to send incorrect amounts of money.

Implementation:
- adds currencyCode to amount types


app/hooks/use-display-currency.ts
```diff
@@ -249,6 +233,15 @@ export const useDisplayCurrency = () => {
       const { symbol, minorUnitToMajorUnitOffset, showFractionDigits, currencyCode } =
         currencyInfo[moneyAmount.currency]
 
+      if (
+        moneyAmount.currency === DisplayCurrency &&
+        currencyCode !== moneyAmount.currencyCode
+      ) {
+        console.log("currencyCode mismatch", currencyCode, moneyAmount.currencyCode)
+        // todo: we should display the correct currency but this requires `showFractDigits` to come from the backend
+        return NaN.toString()
```
nicolasburtey, line 242: I don't think the `NaN.toString()` is the best value to return here. we should say "there is an issue with the display currency, please refresh". will think about a short way to put it.

nicolasburtey:
PR looks great! tested is locally and didn't notice any regression, and went over the code and the changes make sense, although I didn't went to deep into it. 

I've just one comment. I took a stab at the PR so we can merge the PR shortly today and add it to the beta release: https://github.com/GaloyMoney/galoy-mobile/pull/2099 (PR #2099 is set to be merged into this PR)

nicolasburtey (approved):

PR #2103, opened 2023-04-25
fix: ui accessibility issues for users using large text

Extheoisah:
fixes UI accessibility issues such as: 
- overflowing wallet balances in home screen
- skewed buttons and text in home screen for large text
- blink modal is now scrollable and has a max-height


nicolasburtey (approved):

nicolasburtey (changes requested):
tests are not passing

nicolasburtey (approved):

PR #2109, opened 2023-04-26
fix: prevent infinite 401 loop

UncleSamtoshi:
- reverts 401 modal changes and instead gives special retry mechanism for 401 errors before logging the user out
- reorder retryLink so that it runs before the errorLink, in the past we set the error before we even retried, this should retry first and then if it still fails it will set the network error (it should help with the 502 errors we are seeing)

### DIFF
--- a/app/graphql/client.tsx
+++ b/app/graphql/client.tsx
@@ -7,7 +7,6 @@ import {
   ApolloProvider,
   HttpLink,
   NormalizedCacheObject,
-  ServerError,
   split,
 } from "@apollo/client"
 import AsyncStorage from "@react-native-async-storage/async-storage"
@@ -21,7 +20,7 @@ import { createPersistedQueryLink } from "@apollo/client/link/persisted-queries"
 import { useAppConfig } from "@app/hooks"
 import { AsyncStorageWrapper, CachePersistor } from "apollo3-cache-persist"
 import jsSha256 from "js-sha256"
-import React, { PropsWithChildren, useEffect, useState } from "react"
+import React, { PropsWithChildren, useCallback, useEffect, useRef, useState } from "react"
 import { createCache } from "./cache"
 
 import { useI18nContext } from "@app/i18n/i18n-react"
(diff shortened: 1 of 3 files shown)


app/graphql/client.tsx
```diff
@@ -69,7 +69,13 @@ const getAuthorizationHeader = (token: string): string => {
 
 const GaloyClient: React.FC<PropsWithChildren> = ({ children }) => {
   const { appConfig, saveToken } = useAppConfig()
-  const [networkError, setNetworkError] = useState<ServerError>()
+
+  const [_, setHasNetworkError] = useState(false)
```
nicolasburtey, line 73: I don't understand the purpose of this function

app/graphql/client.tsx
```diff
@@ -124,6 +130,16 @@ const GaloyClient: React.FC<PropsWithChildren> = ({ children }) => {
...
       })
 
+      const retry401ErrorLink = new RetryLink({
+        attempts: {
+          max: 3,
+          retryIf: (error) => {
+            return error && error.statusCode === 401
+          },
+        },
+        delay: (count) => count * 20000,
```
nicolasburtey, line 140: intent is to wait for 1 min? (0 * 20 + 1 * 20 + 2 * 20 ?)

nicolasburtey (commented):
some questions on the code. I haven't tried locally yet

app/graphql/client.tsx
UncleSamtoshi: Updating a ref variable doesnt trigger a rerender. We need a way to trigger a rerender while still having the ability to set the error only once. If we were to store the networkError in a useState hook, then the client creation uesEffect would have to depend on the networkError which would cause a new client to be recreated whenever the networkError changed. Instead we set the networkErrorRef.current and then trigger a rerender with the setHasNetworkError. I suppose we could invert the two and make `hasNetworkError` be the ref and `setNetworkError` to useState.

app/graphql/client.tsx
UncleSamtoshi: Intent is to retry at 20 seconds, 40 seconds, and 1 minute. I actually think the expression should be delay `delay: 20000`

nicolasburtey (approved):
haven't tried locally but logic sounds good. I think that's a good/better compromise.

PR #2119, opened 2023-04-28
feat: enable feature flags

UncleSamtoshi:
- create a feature flags context
- use firebase remote config to set the feature

PR #2120, opened 2023-04-28
Feat  device account 2

UncleSamtoshi:
- add feature flag
- clean up level naming
- use new device account mutation

### DIFF
--- a/app/graphql/feature-flags-context.tsx
+++ b/app/graphql/feature-flags-context.tsx
@@ -1,0 +1,66 @@
+import React, { useState, createContext, useContext, useEffect } from "react"
+import remoteConfigInstance from "@react-native-firebase/remote-config"
+import { useAppConfig } from "@app/hooks"
+
+type FeatureFlags = {
+  deviceAccountEnabled: boolean
+}
+
+type RemoteConfig = {
+  deviceAccountEnabled: boolean
+}
+
+const defaultRemoteConfig = {
+  deviceAccountEnabled: false,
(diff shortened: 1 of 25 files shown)


app/graphql/feature-flags-context.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { useState, createContext, useContext, useEffect } from "react"
```
nicolasburtey, line 1: it's not really a graphql related feature?

nicolasburtey: looking at the filename

app/graphql/level-component.tsx
```diff
@@ -25,17 +25,17 @@ gql`
 
 export const LevelContainer: React.FC<PropsWithChildren> = ({ children }) => {
   const isAuthed = useIsAuthed()
-  const isLevel0 = isAuthed
+  const isAtLeaseLevelZero = isAuthed
```
nicolasburtey, line 28:
```suggestion
  const isAtLeastLevelZero = isAuthed
```

nicolasburtey:
some comments on https://github.com/GaloyMoney/galoy-mobile/pull/2119

app/graphql/feature-flags-context.tsx
UncleSamtoshi: Any suggestions on where to put it? Maybe under the `config` folder or just the components.

PR #2124, opened 2023-04-28
feat: add custom modal for device account pop-up

Extheoisah:
I made the modal custom because we might be using the modal in more than one places not just for the device account. e.g. -> https://www.figma.com/file/rH7A7NquSsXEH7v8eGuYM7/Blink-(BBW)-Mobile-App-Design?node-id=5215-121907&t=Uzw5QYVs1uo691ys-0

Simulator Screenshot - iPhone 14 Pro - 2023-04-28 at 15 25 26


app/components/custom-modal/custom-modal.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <View style={styles.modalTitleContainer}>
+            <Text style={styles.modalTitleText}>{title}</Text>
+          </View>
+          <View style={styles.modalBodyContainer}>{body}</View>
+        </ScrollView>
+        <View style={styles.modalActionsContainer}>
+          <GaloyPrimaryButton
+            title={primaryButtonTitle}
+            onPress={primaryButtonOnPress}
+            titleStyle={styles.buttonTitle}
```
UncleSamtoshi, line 61: I don't think we should have to override the `titleStyle`. We want the GaloyPrimary and SecondaryButtons to look the same everywhere, so that if we decide to make a change to them we can change them all at once. We previously changed the fontSize of GaloyPrimaryButton, but we will be changing it back, so I dont think a change should be neccessary any way.

app/assets/images/warning.png
UncleSamtoshi, line 1: I think we should use an svg instead of a png. The png shows up as slightly blurry on my phone.

UncleSamtoshi:
image
I don't think that the `X` button should be part of the scroll view. Potentially even the icon shouldn't be part of the scroll view.

app/components/custom-modal/custom-modal.tsx
Extheoisah: Alright, the reason I did it here because the font size was too large for this components use case.

Extheoisah:
> image
> I don't think that the `X` button should be part of the scroll view. Potentially even the icon shouldn't be part of the scroll view.

fixed in this commit https://github.com/GaloyMoney/galoy-mobile/pull/2124/commits/b4f019be7927e87218c4dd08c47a8a917da6faf4

UncleSamtoshi (approved):

PR #2158, opened 2023-05-10
ci: update versions

sandipndev:


sandipndev:
This is already applied but let's not merge this here. I want to have all these jobs on `galoy-mobile-deployments`

PR #2165, opened 2023-05-11
fix: wallet overview ui issues

Extheoisah:


app/components/wallet-overview/wallet-overview.tsx
```diff
@@ -179,48 +179,60 @@ const WalletOverview: React.FC<Props> = ({
...
-  },
-  loaderContainer: {
-    flex: 1,
-    justifyContent: "flex-end",
-    alignItems: "flex-end",
-    height: 43,
-  },
-}))
+const useStyles = makeStyles(
+  ({ colors }, { isContentVisible }: { isContentVisible: boolean }) => ({
```
nicolasburtey, line 183: that sounds quite hacky. I think we should try to solve this with a pure CSS way, maybe with some min/max margin, or something along those lines?

Extheoisah: Yeah, using flexbox `alignItems: center` was sufficient. I have removed the unnecessary complexity.

nicolasburtey (approved):

PR #2170, opened 2023-05-12
feat: send bitcoin screen dark mode support

sandipndev:
Adds dark mode support to send bitcoin screen.

How Dark Mode looks:
image

How Dark Mode handles error:
image

### DIFF
--- a/app/components/floor-tooltip/floor-tooltip.tsx
+++ b/app/components/floor-tooltip/floor-tooltip.tsx
@@ -9,7 +9,7 @@ import { useAppConfig } from "@app/hooks"
 import { LocalizedString } from "typesafe-i18n"
 import { makeStyles } from "@rneui/themed"
 
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles((theme, type) => ({
   modalStyle: {
     margin: 0,
     flexDirection: "column",
@@ -41,7 +41,7 @@ const useStyles = makeStyles((theme) => ({
     marginBottom: 20,
   },
   icon: {
-    color: theme.colors.grey1,
+    color: type == "info" ? theme.colors.grey1 : theme.colors.error5,
   },
 }))
 
@@ -66,7 +66,7 @@ export const FloorTooltip: React.FC<FloorTooltipProps> = ({
   } = useAppConfig()
   const [isVisible, setIsVisible] = React.useState(false)
   const toggleModal = () => setIsVisible(!isVisible)
-  const styles = useStyles()
+  const styles = useStyles(type)
 
   let iconParams: { name: string; type: string }
   let defaultTitle: LocalizedString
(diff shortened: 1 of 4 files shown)


sandipndev:
Dark Mode:
image

sandipndev:
image

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -66,7 +66,7 @@ const usestyles = makeStyles((theme) => ({
     flexDirection: "row",
     borderStyle: "solid",
     overflow: "hidden",
-    backgroundColor: palette.white,
+    backgroundColor: theme.mode == "light" ? theme.colors.white : theme.colors.grey10,
```
nicolasburtey, line 69: so we don't want to really do this pattern. the idea is that if we have theme.colors.white, then this is the value for the light mode, and for darkMode the value should be translate accordingly. (black in this scenario).

There is some exception here because in the darkMode we have basically a dark background, and in the lightNode, we have sometime white, sometime light grey background. 

we should see the fact the background is sometime lightGrey instead of white for the lightMode as an intermediary state.

Those screens should eventually be refactored to be on par with the darkMode so we don't need to tweak those values per (light/dark) mode

the way to deal with that in the meantime is to create intermediary colors that don't follow the "colors should be inverted between light and dark mode" type of pattern. I have some example of this hack with `lapisLazuliOrLightGrey` or `whiteOrDarkGrey`

sandipndev: Understood, implemented the change!

sandipndev:
The send bitcoin flow has some more screens, I think refactoring them together is in the scope of this PR -- work continues.

nicolasburtey:
> 

feel free to go as you prefer, but my recommendation is to make small PR so we can merge them regularly. merging a PR every day can be a good way for the rest of the team to see the progress and give feedback.

sandipndev:
Then I think it'd be good to get this shipped, the next screen is a lot of code changes.

app/components/floor-tooltip/floor-tooltip.tsx
```diff
@@ -39,7 +40,10 @@ const styles = StyleSheet.create({
     fontSize: 20,
     marginBottom: 20,
   },
-})
+  icon: {
+    color: type === "info" ? theme.colors.grey1 : theme.colors.error5,
```
nicolasburtey, line 44: I don't think conditional in styles is best practice ( no strong opinion though, but this is my guts feeling)

another way to deal with this would be to have a something like:

```
iconInfo: {
    color: theme.colors.grey1,
  },
iconWarning: {
   colors: theme.colors.error5
}
```

nicolasburtey (approved):

PR #2174, opened 2023-05-14
chore: reduce the use of overloaded sendPayment function name

nicolasburtey:


app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -219,7 +219,7 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
     destination,
     paymentType,
     sendingWalletDescriptor,
-    sendPayment: sendPaymentFn,
```
nicolasburtey, line 222: I feel `sendPayment` is quite convoluted currently, this is probably the line that is most confusing to me.

we rename `sendPayment` to `sendPaymentFn`, then wrap into into a `useSendPayment()`, to get back a `sendPayment` again.

I feel renaming `sendPayment` in function that live outside the send bitcoin confirmation help clarify which function do what with more clarity. 

happy to use another name if you think there could be something better.

app/screens/send-bitcoin-screen/payment-details/onchain.ts
```diff
@@ -275,12 +275,12 @@ export const createAmountOnchainPaymentDetails = <T extends WalletCurrency>(
     canGetFee: false,
   }
 
-  let sendPayment: SendPayment
+  let sendPaymentSelector: SendPaymentSelector
   let getFee: GetFee<T>
 
   if (sendingWalletDescriptor.currency === WalletCurrency.Btc) {
-    sendPayment = async (sendPaymentFns) => {
```
nicolasburtey, line 282: i have also renamed `sendPaymentFns` to `paymentMutations` just to clarify the later are just the mutation we are injecting.

Issue #2157, opened 2023-05-09
Pending payment shows generic or network error


dolcalmi:
**Describe the bug**
when you pay a hold invoice and payments mutation returns PENDING the app shows a generic error (something went wrong or network error)

**To Reproduce**

1. Pay a hold invoice (for ex from OBW) and wait 
1. See error

**Expected behavior**
finish the flow, show a similar screen to success screen (with a pending icon/msg?) and the go to main screen

**Additional context**
We can see the real impact in payments dashboard.. between robosats and lnp2pbot we had 108 payments with this behavior in the last month

dolcalmi:
@nicolasburtey any suggestion on how to handle this flow?

nicolasburtey:
> @nicolasburtey any suggestion on how to handle this flow?

from a UX perspective? 

not sure but I'll dive into it. I'll reproduce the error first.

designsats:
Resolved in two stages — neither PR referenced this issue, which is why it stayed open.

**#2180** (merged 2023-06-30) fixed the reported bug. The confirmation screen's success branch required an exact `SUCCESS`, so a `PENDING` result fell through to the catch-all `setPaymentError(errorsMessage || "Something went wrong")`:

```diff
- if (!errorsMessage && status === "SUCCESS") {
+ if (status === "SUCCESS" || status === "PENDING") {

grimen:
Closing as part of old-ticket cleanup (tracked in blinkbitcoin/blink-wip#1169) — no longer applicable/relevant to the current app.

PR #2180, opened 2023-05-15
fix: prevent multiple sends

UncleSamtoshi:
- only allow one attempt at sending a payment
- show a success message if payment status is `valid`


UncleSamtoshi:
This pr does not fully solve issue #2157 , however instead of showing an error message and allowing the user to continue clicking send, it shows a success message.

It also prevent sendPayment from being clicked more than once.

nicolasburtey (approved):

PR #2189, opened 2023-05-18
feat: redesign phone auth flow

UncleSamtoshi:
- redesign phone auth flow for device account
- clean up error handling


package.json
```diff
@@ -83,6 +83,7 @@
     "js-lnurl": "0.5.1",
     "js-sha256": "^0.9.0",
     "jwt-decode": "^3.1.2",
+    "libphonenumber-js": "^1.10.30",
```
nicolasburtey, line 86: shouldn't we use the mobile dataset (instead of min)?

https://gitlab.com/catamphetamine/libphonenumber-js#min-vs-max-vs-mobile-vs-core

it has more validation for mobile phone number

app/components/atomic/galoy-info/galoy-info.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+} as Meta<typeof GaloyInfo>
+
+export const Default = () => {
+  return (
+    <View>
+      <UseCase text="Info">
+        <GaloyInfo infoMessage="Some info" />
+      </UseCase>
+      <UseCase text="Info with highlight">
+        <GaloyInfo infoMessage="Some info with highlight" highlight />
```
nicolasburtey, line 20: I really like this component

app/screens/phone-auth-screen/phone-input.tsx
```diff
@@ -1,396 +1,216 @@
...
-                containerStyle={styles.button}
-                title={LL.PhoneInputScreen.whatsapp() + "  "}
-                onPress={submitViaWhatsapp}
-                icon={<Icon name="logo-whatsapp" size={14} color={color.primary} />}
-              />
-              <ContactSupportButton containerStyle={styles.button} />
-            </View>
-          </KeyboardAvoidingView>
+            >
+              {phoneInputInfo?.formattedPhoneNumber || DEFAULT_PHONE_NUMBER}
```
nicolasburtey, line 186: it seem we can't paste a phone number any more

nicolasburtey:
great PR! seen the improvement on those screens is awesome.

I've push a few commits. @UncleSamtoshi you should review it.

I've tested locally and it works great. I've not tested the "1 button paste code from iOS" as it only works on a physical device but can test that when it's merged

my main comment is here https://github.com/GaloyMoney/galoy-mobile/pull/2189/files#r1198820696

app/screens/phone-auth-screen/phone-input.tsx
UncleSamtoshi: Just to clarify you want to be able to paste your phone number in the phone input? What is the scenario where someone has first copied their phone number and then pastes it?

UncleSamtoshi:
I don't think its possible to have the "1 click paste code from ios" as that requires a standard OS keyboard.

nicolasburtey (approved):

PR #2197, opened 2023-05-22
test: fix for failing android test due to version updates

Aditi-Singh16:
The version update PR by the bot, updated modules one by one, but ideally all of them should have the same version. Earlier we were at v7 and now we are transitioning to v8, there are many breaking changes to be noted - [v8 changes](https://github.com/webdriverio/webdriverio/releases/tag/v8.0.0)

So, when I updated it to the latest version 8.10.4, I still got the following error:
```
D:\galoy-mobile\node_modules\@wdio\cli\node_modules\@wdio\logger\build\node.js:9
const chalk_1 = __importDefault(require("chalk"));
                                ^

Error [ERR_REQUIRE_ESM]: require() of ES Module D:\galoy-mobile\node_modules\@wdio\cli\node_modules\chalk\source\index.js from D:\galoy-mobile\node_modules\@wdio\cli\node_modules\@wdio\logger\build\node.js not supported. Instead change the require of index.js in D:\galoy-mobile\node_modules\@wdio\cli\node_modules\@wdio\logger\build\node.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (D:\galoy-mobile\node_modules\@wdio\cli\node_modules\@wdio\logger\build\node.js:9:33) {
  code: 'ERR_REQUIRE_ESM'
}
```

chalk is one of the dependencies of  @wdio/cli package. Now @wdio/logger latest version is 8.6.6 which uses require to import chalk package, thus updating all the @wdio modules to 8.6.6 did the work!


nicolasburtey:
@Aditi-Singh16 we are using https://www.conventionalcommits.org/en/ fyi

for some reason the e2e tests are not been executed in the PR. I'll have a deeper look tomorrow for why that is.

Issue #2208, opened 2023-05-23
notification bar issue with android


nicolasburtey:
bar is grey. 

it should be either white or black (and maybe blue for the earn section?)

PR #2212, opened 2023-05-23
fix: changed notification bar color on mode change

Aditi-Singh16:
Fix issue #2208

**Preview**
image
image

I haven't changed the earn screen to blue or green
if we want to match it with the background, then the earn screen also scrolls, and when we scroll the background color changes, should I change the notification bar color on scroll?

### DIFF
--- a/app/components/screen/screen.tsx
+++ b/app/components/screen/screen.tsx
@@ -11,7 +11,6 @@ import { ScreenProps } from "./screen.props"
 import { isNonScrolling, offsets, presets } from "./screen.presets"
 import { isIos } from "../../utils/helper"
 import { makeStyles, useTheme } from "@rneui/themed"
-import { light } from "@app/rne-theme/colors"
 
 const useStyles = makeStyles(({ colors }) => ({
   background: {
--- a/app/components/screen/screen.tsx
+++ b/app/components/screen/screen.tsx
@@ -41,7 +41,7 @@ function ScreenWithoutScrolling(props: ScreenProps) {
     >
       <StatusBar
         barStyle={props.statusBar || statusBarContent}
-        backgroundColor={props.backgroundColor}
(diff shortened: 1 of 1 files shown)


nicolasburtey (changes requested):
we don't want to put hard coded values. we should always refers to colors.ts thought our rne theme wrapper

nicolasburtey:
you can rebase this should fix the test

app/components/screen/screen.tsx
```diff
@@ -16,6 +16,12 @@ const useStyles = makeStyles(({ colors }) => ({
   background: {
     color: colors.white,
   },
+  notificationbarlight: {
+    color: colors._white,
```
nicolasburtey, line 20: You should use .white instead of ._white.

It automatically provide color conversion across mode.

This will reduce this PR to only 2 or 3 lines of code

nicolasburtey (changes requested):
This can be simplified

app/components/screen/screen.tsx
```diff
@@ -41,7 +41,7 @@ function ScreenWithoutScrolling(props: ScreenProps) {
     >
       <StatusBar
         barStyle={props.statusBar || statusBarContent}
-        backgroundColor={props.backgroundColor}
+        backgroundColor={styles.background.color}
```
nicolasburtey, line 44: you can actually use `colors.white` instead, `colors` being destructured from 

```
  const {
    theme: { colors },
  } = useTheme()
```

then you can probably refactored this style by removing it entirely:

```
const useStyles = makeStyles(({ colors }) => ({
  background: {
    color: colors.white,
  },
}))
```

nicolasburtey (commented):
it's in a good shape now. adding some refactoring suggestions to make this file cleaner.

nicolasburtey (approved):

PR #2214, opened 2023-05-24
feat: add back gestures for iOS

sandipndev:
I've noticed when using the app on iPhone that the back gesture don't work for most screens, this is an effort to make sure it does.


nicolasburtey:
> I've noticed when using the app on iPhone that the back gesture don't work for most screens, this is an effort to make sure it does.

I'm not sure why it was disabled in the first place. any clue @UncleSamtoshi ?

nicolasburtey (approved):

PR #2248, opened 2023-05-26
feat: receive from nfc/boltcard over lnurlw

nicolasburtey:


nicolasburtey:
> PR looks good, but I still need to test it with a physical card

try to see if there are some places where you could order some? 

Based on my research, it seems to be an initiative that started in Europe (or should I say UK?) and where it's mostly available in Europe as of now. The list of available online store are on the FAQ section of https://www.boltcard.org/

I want to do another iteration on it because for iOS there is a popup that appear from the OS. This is not the case for Android. So I think the modal I created should be explicitly opened on our end only on Android.

nicolasburtey:
> > PR looks good, but I still need to test it with a physical card
> 
> try to see if there are some places where you could order some?
> 
> Based on my research, it seems to be an initiative that started in Europe (or should I say UK?) and where it's mostly available in Europe as of now. The list of available online store are on the FAQ section of https://www.boltcard.org/
> 
> I want to do another iteration on it because for iOS there is a popup that appear from the OS. This is not the case for Android. So I think the modal I created should be explicitly opened on our end only on Android.

@UncleSamtoshi related to this PR: https://github.com/GaloyMoney/galoy-mobile/issues/2263

It's not needed to be merged IMO because it would already work if the url is bech32 encoded, but if we want boltcard to work we need to work on the issue 2263 eventually.

app/components/modal-nfc/modal-nfc.tsx
```diff
@@ -0,0 +1,228 @@
...
+      } catch (error) {
+        // TODO: error that show as an Alert or onscreen message
+        // but only when it's not user initiated
+        // currently error returned is empty
+        console.error({ error }, `can't fetch the Ndef payload`)
+        dismiss()
+        return
+      }
+
+      NfcManager.cancelTechnologyRequest()
```
UncleSamtoshi, line 107: Is this necessary right here?

nicolasburtey, line 107:
```suggestion
```

PR #2268, opened 2023-05-31
chore: upgrading to galoy-client 0.2

### LINKED ISSUE #2263: lnurlw spec prefix/ unencoded callback urls not managed properly
boltcard return a link in the form of `lnurlw://cards.galoy.io/cards?p=ABCABCABCABCABCBACBACB&c=ABCABCABCABC`

the lnurlw:// is not been decoded properly right now. 
putting the same link into our destination is the send bitcoin screen shows a "please enter a valid destination"

this works thought:

```
  const url =
    "https://cards.galoy.io/cards?p=ABCABCABCABCABCBACBACB&c=ABCABCABCABC"

  const words = bech32.toWords(Buffer.from(url, "utf8"))
  const encoded = bech32.encode("lnurl", words, 1024)
  console.log(encoded)
```

another side bug:

```
  const url =
    "cards.galoy.io/cards?p=ABCABCABCABCABCBACBACB&c=ABCABCABCABC"

  const words = bech32.toWords(Buffer.from(url, "utf8"))
  const encoded = bech32.encode("lnurl", words, 1024)
  console.log(encoded)
```

this one leads to an infinitely loading at the send destination screen stage, because of missing `https://` prefix 

related spec: https://github.com/lnurl/luds/blob/luds/17.md

nicolasburtey:
fix #2263


app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -346,7 +348,9 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
 
           const result = await fetchLnurlInvoice({
             lnUrlOrAddress: paymentDetail.destination,
-            tokens: btcAmount.amount as Satoshis,
+            // FIXME: remove any from lnurl-pay lib?
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            tokens: btcAmount.amount as any,
```
nicolasburtey, line 353: Should we remove `Satoshis` from lnurl-pay library? 

This is what I have done for galoymoney-client. 

Those type are not very portable across library (or maybe I just don't know how to do it properly), so I think using number or bigint may be better. I've switch from Satoshis to number in `galoymoney-client`

UncleSamtoshi: If you switched to number in `galoymoney-client` why do you need to cast `btcAmount.amount` as any? I think it could be beneficial to keep the Satoshi type and just export it from galoy-money client, or we could rename the field. Some of the libraries use millisatoshis, so it's important we know which one is required.

dolcalmi: well, not sure about the config in react but types should be loaded automatically https://github.com/dolcalmi/lnurl-pay/blob/main/package.json#L8 :thinking:

nicolasburtey: the issue I faced is that Satoshi in lnurl-pay is not the same as Satoshi defined in another library

dolcalmi: you dont need to import the type, just use utils.checkedToSats or utils.toSats, also removed the url warning and included lnurlp in the new version 2.3.0 https://www.npmjs.com/package/lnurl-pay

PR #2275, opened 2023-06-01
feat: select component

sandipndev:
Adds `<Select>` component.


nicolasburtey:
tests are not passing

nicolasburtey:
probably `match-currencies.spec.ts` should have a tsx extension

app/components/select/item.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const [showActivityIndicator, setShowActivityIndicator] = React.useState(false)
+
+  const onPress = () => {
+    if (selected || loading) return
+
+    const isPromise = onChange(value)
+    if (isPromise) {
+      isPromise
+        .then(() => {
```
nicolasburtey, line 42: we don't use `then` in our codebase. 
you can use async/await instead. it will make the code more readable (may not be too obvious when there is one level of nesting, but it's become very apparent where there is several level of nesting.)

app/components/select/select.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { SelectItemProps } from "./select-item"
+import { Item } from "./item"
+
+export type SelectProps = {
+  value: string
+  onChange: (optionKey: string) => void | Promise<void>
+  children: React.ReactElement<SelectItemProps> | React.ReactElement<SelectItemProps>[]
+}
+
+export const Select: React.FC<SelectProps> = ({ value, onChange, children }) => {
```
nicolasburtey, line 13: select feels a bit too generic imo. maybe `MenuSelect` is a better description.

app/components/select/select.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [loading, setLoading] = React.useState(false)
+
+  const childrenArray: React.ReactElement<SelectItemProps>[] = React.Children.toArray(
+    children,
+  ) as React.ReactElement<SelectItemProps>[]
+
+  return (
+    <View>
```
nicolasburtey, line 21:
```suggestion
    <>
```

you probably don't need a view here.

app/screens/settings-screen/default-wallet.tsx
```diff
@@ -5,23 +5,15 @@ import {
...
-const useStyles = makeStyles(({ colors }) => ({
-  viewSelectedIcon: { width: 18 },
-
-  container: { backgroundColor: colors.white },
-
-  text: {
-    color: colors.black,
-  },
-
+const useStyles = makeStyles(() => ({
```
nicolasburtey, line 16: we are now putting the style at the bottom (and whenever we touch the style we should move them)

app/screens/settings-screen/default-wallet.tsx
```diff
@@ -115,27 +108,16 @@ export const DefaultWalletScreen: React.FC = () => {
...
-              <ActivityIndicator />
-            ) : (
-              defaultWalletId === id && (
-                <Icon name="ios-checkmark-circle" size={18} color={colors.green} />
-              )
-            )}
-          </View>
-          <ListItem.Title {...testProps(name)} style={styles.text}>
+      <Select
+        value={newDefaultWalletId || defaultWalletId || ""}
```
nicolasburtey, line 112: is `|| ""` really needed?

app/components/select/item.tsx
```diff
@@ -0,0 +1,64 @@
...
+    theme: { colors },
+  } = useTheme()
+  const styles = useStyles()
+
+  const [showActivityIndicator, setShowActivityIndicator] = React.useState(false)
+
+  const onPress = () => {
+    if (selected || loading) return
+
+    const isPromise = onChange(value)
```
nicolasburtey, line 39: this variable name is misleading. a `isPromise` would appear at first to be a boolean - because of the `is`. It seems that's not the case here.

app/components/select/item.tsx
```diff
@@ -0,0 +1,64 @@
+import React from "react"
+import { ActivityIndicator, View } from "react-native"
+
+import { ListItem, ListItemProps, makeStyles, useTheme } from "@rneui/themed"
+import Icon from "react-native-vector-icons/Ionicons"
+
+const useStyles = makeStyles(() => ({
+  iconContainer: { width: 18 },
+}))
+
```
nicolasburtey, line 10: put at the bottom

app/components/select/item.tsx
sandipndev: I've removed it fully

nicolasburtey (approved):

PR #2293, opened 2023-06-04
chore: test inapp browser

nicolasburtey:


nicolasburtey:
another thing to try: https://www.npmjs.com/package/react-native-inappbrowser-reborn

PR #2294, opened 2023-06-04
fix: fixed login-flow for e2e tests

Aditi-Singh16:
image

### DIFF
--- a/e2e/02-login-flow.e2e.spec.ts
+++ b/e2e/02-login-flow.e2e.spec.ts
@@ -107,6 +107,7 @@ describe("Login Flow", () => {
   })
 
   it("navigates back to move home screen", async () => {
+    await scrollDown()
     await scrollUp()
     const phoneSetting = await $(selector(LL.common.phoneNumber(), "StaticText"))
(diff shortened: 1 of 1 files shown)


Aditi-Singh16:
This is new, login flow works but user flow doesn't now,

Aditi-Singh16:
test_in_android fails because it can't send btc to the user "galoytest" stating the reason that the user doesn't exist, however, when I try sending it to galoytest in staging env, I don't get the error

e2e/02-login-flow.e2e.spec.ts
```diff
@@ -107,6 +107,7 @@ describe("Login Flow", () => {
   })
 
   it("navigates back to move home screen", async () => {
+    await scrollDown()
```
nicolasburtey, line 110: I don't know what the issue is, but doing a `scrollDown()` to next do a `scrollUp()` feels like an anti-pattern to me

nicolasburtey: if this works merely because this is synonym to a `wait sleep()`, we should use better primitive

PR #2316, opened 2023-06-09
fix: set memo for onchain payments

sandipndev:
This was a bug


nicolasburtey:
> This was a bug

it's failing eslint check. should be replaced with just `memo`

nicolasburtey (approved):

PR #2320, opened 2023-06-10
fix: translate dates

nicolasburtey:
also remove dependency on moment()


app/screens/receive-bitcoin-screen/receive-usd.tsx
```diff
@@ -443,13 +442,17 @@ const TimeInformation = ({
...
+  const seconds = date.getUTCSeconds()
+
   return (
     <View style={styles.countdownTimer}>
       <Text style={timeLeft < 10 ? styles.lowTimer : undefined}>
         {LL.ReceiveWrapperScreen.expiresIn()}:{" "}
       </Text>
       <View style={styles.timer}>
-        <Text>{moment.utc(timeLeft * 1000).format("m:ss")}</Text>
+        <Text>{`${minutes}:${seconds}`}</Text>
```
UncleSamtoshi, line 455: Will this result in weird formatting for seconds that are less than 10? For example 2:08 would be formatted as 2:8?

nicolasburtey: mmm probably. good catch. will try and fix if this is the case

PR #2325, opened 2023-06-12
chore: prevent button flickering on app start

nicolasburtey:
currently the secondary button label change after 1 sec

this PR would preserve the "blink" loading screen from RNBootSplash until the call to the the remote config server has resolved

### DIFF
--- a/app/app.tsx
+++ b/app/app.tsx
@@ -48,8 +48,8 @@ export const App = () => (
   <PersistentStateProvider>
     <TypesafeI18n locale={detectDefaultLocale()}>
       <ThemeProvider theme={theme}>
-        <FeatureFlagContextProvider>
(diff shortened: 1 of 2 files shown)


UncleSamtoshi:
Does this show the loading screen or does it show nothing?

nicolasburtey:
It should show the RNBootSplash which is the blink branded screen that get a fade out after the app has loaded.

nicolasburtey:
> It should show the RNBootSplash which is the blink branded screen that get a fade out after the app has loaded.

hidden animation is done here https://github.com/GaloyMoney/galoy-mobile/blob/main/app/navigation/navigation-container-wrapper.tsx#L125

UncleSamtoshi (approved):

PR #2326, opened 2023-06-12
fix: Fixed e2e tests

Aditi-Singh16:


e2e/02-login-flow.e2e.spec.ts
```diff
@@ -101,15 +98,22 @@ describe("Login Flow", () => {
...
 
   it("navigates back to move home screen", async () => {
-    await scrollUp()
-    const phoneSetting = await $(selector(LL.common.phoneNumber(), "StaticText"))
-    await phoneSetting.waitForDisplayed({ timeout })
+    const maxScrolls = 5
+    const testID = LL.common.phoneNumber()
+    let element = await $(`~${testID}`)
+    for (let i = 0; i < maxScrolls; i++) {
+      if (await element.isDisplayed()) {
```
nicolasburtey, line 108: if is typically an antipattern for tests: 
https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-if.md

e2e/03-intraledger-flow.e2e.spec.ts
```diff
@@ -120,7 +120,6 @@ describe("Username Payment Flow", () => {
   it("Click Next again", async () => {
     const nextButton = await $(selector(LL.common.next(), "Button"))
     await nextButton.waitForDisplayed({ timeout })
-    await nextButton.waitForEnabled()
```
UncleSamtoshi, line 123: I believe you can removed `waitForDisplayed`, because if you wait for something to be enabled it will also be displayed. However, just because something is displayed does not mean its enabled, so I think its important to keep the `waitForEnabled`.

e2e/02-login-flow.e2e.spec.ts
```diff
@@ -29,11 +28,12 @@ describe("Login Flow", () => {
...
-    await browser.pause(50)
-    await buildButton.click()
-    await browser.pause(50)
-    await buildButton.click()
+    let i = 0
+    while (i < 3) {
+      await buildButton.click()
+      await browser.pause(100)
+      i = i + 1
+    }
```
nicolasburtey, line 36: the code is even longer with the loop. let's keep the loop unrolled and just replace the `50` to `100` is this is what fix this test

nicolasburtey (changes requested):
one changes request but looks good otherwise

e2e/03-intraledger-flow.e2e.spec.ts
```diff
@@ -152,14 +149,12 @@ describe("Conversion Flow", () => {
   it("Click Next", async () => {
     const nextButton = await $(selector(LL.common.next(), "Button"))
     await nextButton.waitForDisplayed({ timeout })
-    await nextButton.waitForEnabled()
```
UncleSamtoshi, line 155: This should be switched to be `waitForEnabled`

UncleSamtoshi:
One comment, otherwise it looks good.

nicolasburtey (approved):

PR #2334, opened 2023-06-13
chore: device account v2

UncleSamtoshi:


__tests__/config/galoy-instances.spec.ts
```diff
@@ -18,6 +18,7 @@ it("get a full object with Custom", () => {
     name: "Custom",
     graphqlUri: "https://api.custom.com/graphql",
     graphqlWsUri: "ws://ws.custom.com/graphql",
+    restUrl: "https://api.custom.com",
```
nicolasburtey, line 21: We could call it authUrl. 

I think Rest endpoints will only be limited to auth related task

nicolasburtey (approved):
Didn't review it yet, just one comment

nicolasburtey (changes requested):
Didn't meant to approve... using GitHub mobile, not used to

app/graphql/client.tsx
```diff
@@ -95,28 +95,46 @@ const GaloyClient: React.FC<PropsWithChildren> = ({ children }) => {
...
       )
 
-      const wsLinkAuth = async () => {
-        if (token) {
-          return {
-            Authorization: getAuthorizationHeader(token),
-          }
-        } else if (appConfig.isAuthenticatedWithDeviceAccount) {
+      const appCheckLink = setContext(async (_, { headers }) => {
+        if (appConfig.isAuthenticatedWithDeviceAccount) {
```
nicolasburtey, line 99: why not send the appCheckToken for all levels? this is what I was suggesting yesterday, the only drawback been the bandwidth

app/screens/get-started-screen/device-account-modal.tsx
```diff
@@ -18,17 +16,16 @@ import {
   logCreateDeviceAccountFailure,
   logCreatedDeviceAccount,
 } from "@app/utils/analytics"
+import * as Keychain from "react-native-keychain"
```
nicolasburtey, line 19: note: I think is also what we should use to store the current session token so that it's not easily extractable.

app/screens/phone-auth-screen/phone-validation.tsx
```diff
@@ -221,35 +227,49 @@ export const PhoneValidationScreen: React.FC<PhoneValidationScreenProps> = ({
             variables: { input: { phone, code } },
           })
 
-          sessionToken = data?.userLoginUpgrade?.authToken
+          const success = data?.userLoginUpgrade?.success
+          const sessionToken = data?.userLoginUpgrade?.authToken
+          if (success) {
+            logUpgradeLoginSuccess()
+
+            await Keychain.resetGenericPassword()
```
nicolasburtey, line 235: can this call throw an error?

app/screens/get-started-screen/device-account-modal.tsx
```diff
@@ -42,38 +39,82 @@ export const DeviceAccountModal: React.FC<DeviceAccountModalProps> = ({
...
+  const [loading, setLoading] = React.useState(false)
+
   const { LL } = useI18nContext()
   const navigation =
     useNavigation<StackNavigationProp<RootStackParamList, "getStarted">>()
 
   const createDeviceAccountAndLogin = async () => {
     try {
+      setLoading(true)
+      const credentials = await Keychain.getInternetCredentials("device-account")
```
nicolasburtey, line 58: `device-account` should be in a const

app/screens/get-started-screen/device-account-modal.tsx
```diff
@@ -42,38 +39,82 @@ export const DeviceAccountModal: React.FC<DeviceAccountModalProps> = ({
...
       }
+
+      logCreatedDeviceAccount()
+      analytics().logLogin({ method: "device" })
+      setAuthenticatedWithDeviceAccount(sessionToken)
+      navigation.replace("Primary")
+      closeModal()
     } catch (error) {
+      setHasError(true)
+      setLoading(false)
```
nicolasburtey, line 114: should be in `finally`?

app/screens/get-started-screen/device-account-modal.tsx
```diff
@@ -42,38 +39,82 @@ export const DeviceAccountModal: React.FC<DeviceAccountModalProps> = ({
...
       logCreateDeviceAccountFailure()
-      console.log("Error creating device account: ", error)
+      console.log("Error with device account: ", error)
```
nicolasburtey, line 116: maybe we could/should have an Alert here instead of a console.log? not sure what kind of error we can expect. I guess the main question I have is on whether all android device works with the react-native-keychain package.

app/screens/phone-auth-screen/phone-validation.tsx
UncleSamtoshi: I actually was going to remove this call now because we don't need to know whether or not an account is already created since the 'create' mutation is renentrant.

app/screens/get-started-screen/device-account-modal.tsx
UncleSamtoshi: The `setLoading` ?

app/screens/get-started-screen/device-account-modal.tsx
UncleSamtoshi: Agreed

app/graphql/client.tsx
UncleSamtoshi: I had thought that it might incur a cost, but it looks like it is free with a quota of 10,000 calls per day, so I think we can enable it on all calls.

app/screens/get-started-screen/device-account-modal.tsx
UncleSamtoshi: In the case of an error it will show an error modal, so I'm not sure if it is valuable to alert the user with. In order to detect the error, we could log it with crashlytics.

nicolasburtey (approved):

PR #2361, opened 2023-06-16
feat: feedback flow after first payment

### LINKED ISSUE #2338: Feedback flow after making the first payment
Reference from [here](https://github.com/GaloyMoney/galoy-mobile/pull/704#issuecomment-1522990789)

**Detailed flow**

Galoy

For skipping the whole process we can have a close button on the modal/pop-up

Aditi-Singh16:
Fixes #2338 

https://github.com/GaloyMoney/galoy-mobile/assets/74586135/0605edc5-0299-4d2d-a63d-04fc306b34e5


app/screens/send-bitcoin-screen/improvement-request.tsx
```diff
@@ -14,14 +15,51 @@ export const SuggestAnImprovement: React.FC = () => {
...
+        const newToken = response.headers.get("Token")
+        setToken(newToken ?? "")
+        await AsyncStorage.setItem("mattermostToken", newToken ?? "")
+      }
+    } else {
+      setToken(savedToken ?? "")
+    }
+
+    await fetch("https://chat.galoy.io/api/v4/posts", {
+      method: "POST",
```
nicolasburtey, line 50: this would have to be from the backend. we can't embed securely a token or username/password in a mobile app

Aditi-Singh16: oh okay, then should I make a PR in the galoy repo?

nicolasburtey: I would not advise for it.. there is a steep learning curve on the backend so maybe console.log() the answer for now and someone with knowledge of the backend can then make the appropriate mutation.

Aditi-Singh16: What would be the best way to know whether it's the first payment or not?

nicolasburtey: you can fetch the list of transactions

nicolasburtey:
needs to be rebased

nicolasburtey:
- we need to store the fact we have shown this modal to the user locally (using a dedicated local mutation in apollo)
- why are not using Alert api for the first modal asking for feedback?

Aditi-Singh16:
> * we need to store the fact we have shown this modal to the user locally (using a dedicated local mutation in apollo)

ok, changed to alert

Is this the right documentation to follow? https://www.apollographql.com/docs/react/v2/data/local-state/

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -10,38 +10,145 @@ import { useI18nContext } from "@app/i18n/i18n-react"
...
+          text: LL.common.yes(),
+          onPress: () => rateUs,
+        },
+      ])
+    }
+  }, [isActive, LL])
+
+  const showSuggestionModal = () => {
+    return (
+      <Modal
```
sandipndev, line 118: Can you make this it's own component?
Having so much code in the screen for suggestion seems counterintuitive to me - it's better to have this as it's own component.

Perhaps make the component auto display itself if the user is receiving for the first time as well?

And the local auth query must also reside there.

The send screen as well as the home screen can just have `<Suggestion ... />` in that way.

sandipndev:
@Aditi-Singh16 you can see how the theme is stored here:  https://github.com/GaloyMoney/galoy-mobile/blob/main/app/graphql/local-schema.gql#L18

And we use a local-query to fetch it at the theme-sync.ts file.

app/graphql/client-only-query.ts
```diff
@@ -91,3 +97,20 @@ export const updateColorScheme = (client: ApolloClient<unknown>, colorScheme: st
...
+  try {
+    client.writeQuery<FeedbackModalShownQuery>({
+      query: FeedbackModalShownDocument,
+      data: {
+        __typename: "Query",
+        feedbackModalShown: shown,
+      },
+    })
+  } catch {
+    console.warn("feedback modal is not shown")
```
nicolasburtey, line 114: message is not correct here. the error would arise if the apollo is unable to update the value

app/i18n/en/index.ts
```diff
@@ -874,9 +876,11 @@ const en: BaseTranslation = {
     joinTheCommunity: "Join the community",
     whatsapp: "WhatsApp",
     email: "Email",
+    enjoyingApp: "Enjoying the app?",
     statusPage: "Status Page",
     telegram: "Telegram",
     mattermost: "Mattermost",
+    thankyouText: "Thankyou for the feedback, would you like to like to suggest an improvement?",
```
nicolasburtey, line 883: typo

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,74 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
+import { ratingOptions } from "@app/config"
+import crashlytics from "@react-native-firebase/crashlytics"
+import { useApolloClient } from "@apollo/client"
+import { useFeedbackModalShownQuery } from "@app/graphql/generated"
+import { feedbackModalShownScheme } from "@app/graphql/client-only-query"
+import { SuggestAnImprovement } from "./suggestion-modal"
 
 const SendBitcoinSuccessScreen = () => {
   const styles = useStyles()
+  const [isActive, setIsActive] = React.useState(false)
```
nicolasburtey, line 26: variable name does not make it clear what is active

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -55,8 +92,37 @@ const SendBitcoinSuccessScreen = () => {
           </Text>
         </SuccessTextAnimation>
       </View>
+      {showImprovement && (
+        <SuggestAnImprovement
+          navigation={navigation}
+          showImprovement={showImprovement}
+          setShowImprovement={setshowImprovement}
```
nicolasburtey, line 99: not consistent with camelCase

app/screens/send-bitcoin-screen/suggestion-modal.tsx
```diff
@@ -0,0 +1,102 @@
...
+import Modal from "react-native-modal"
+import { Text, makeStyles, useTheme } from "@rneui/themed"
+import { View, TextInput } from "react-native"
+import { StackNavigationProp } from "@react-navigation/stack"
+import { useI18nContext } from "@app/i18n/i18n-react"
+import { GaloySecondaryButton } from "@app/components/atomic/galoy-secondary-button"
+import { testProps } from "../../utils/testProps"
+import { RootStackParamList } from "@app/navigation/stack-param-lists"
+
+export const SuggestAnImprovement: React.FC<{
```
nicolasburtey, line 11: stories (storybook) needed for the component

app/screens/send-bitcoin-screen/suggestion-modal.tsx
```diff
@@ -0,0 +1,102 @@
...
+    // })
+  }
+
+  const dismissSuggestionModal = () => {
+    navigation.popToTop()
+    setShowImprovement(false)
+  }
+
+  return (
+    <Modal
```
UncleSamtoshi, line 50: You should switch to using the CustomModal component similar to how it is being used in the SetAddressModal. This will help to ensure a consistent experience with Modals in the application.

app/graphql/client-only-query.ts
```diff
@@ -91,3 +97,20 @@ export const updateColorScheme = (client: ApolloClient<unknown>, colorScheme: st
     console.warn("impossible to update beta")
   }
 }
+
+export const feedbackModalShownScheme = (
```
UncleSamtoshi, line 101: This should be renamed to `setFeedbackModalShown`

UncleSamtoshi:
I think going straight into leaving a review might upset some people because they never agreed to do that. Are we able to give the user three options?

- Yes, leave a review
- Yes, skip review
- No

Otherwise we could use a second modal to ask them to review us.

A quick convo with ChatGPT suggests that the standard UX is to have a secondary prompt. https://chat.openai.com/share/24eacd69-b0f3-42c0-905d-77e9559b477b

Thoughts @nicolasburtey @agbegin

app/i18n/en/index.ts
```diff
@@ -874,9 +876,11 @@ const en: BaseTranslation = {
     joinTheCommunity: "Join the community",
     whatsapp: "WhatsApp",
     email: "Email",
+    enjoyingApp: "Enjoying the app?",
     statusPage: "Status Page",
     telegram: "Telegram",
     mattermost: "Mattermost",
+    thankyouText: "Thankyou for the feedback, would you like to suggest an improvement?",
```
nicolasburtey, line 883:
```suggestion
    thankyouText: "Thank you for the feedback, would you like to suggest an improvement?",
```

Aditi-Singh16:
image
after switching to CustomModal

Aditi-Singh16:
> I think going straight into leaving a review might upset some people because they never agreed to do that. Are we able to give the user three options?
> 
> * Yes, leave a review
> * Yes, skip review
> * No
> 
> Otherwise we could use a second modal to ask them to review us.
> 
> A quick convo with ChatGPT suggests that the standard UX is to have a secondary prompt. https://chat.openai.com/share/24eacd69-b0f3-42c0-905d-77e9559b477b
> 
> Thoughts @nicolasburtey @agbegin

Should I add this flow?

nicolasburtey:
Secondary prompt is already given by the OS window?

Aditi-Singh16:
> Secondary prompt is already given by the OS window?

The user does have an option to dismiss both the modals.

app/i18n/en/index.ts
UncleSamtoshi: You marked this as resolved, but it is still present. Do you have changes that need to be pushed?

UncleSamtoshi:
> image after switching to CustomModal

I would have the text start in the first line and not have it be centered in the middle. You should also provide a secondary button `Skip` in the custom modal.

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,74 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
 
+  const [showImprovement, setShowImprovement] = React.useState(false)
   const navigation =
     useNavigation<StackNavigationProp<RootStackParamList, "sendBitcoinSuccess">>()
 
+  const client = useApolloClient()
+  const feedbackShownData = useFeedbackModalShownQuery()
+  const modalShown = feedbackShownData?.data?.feedbackModalShown
+
+  const dismiss = () => {
```
UncleSamtoshi, line 36: As part of these actions, you should log a custom firebase event called something like `logAppFeedback` with the parameter `isEnjoingApp: boolean`. You can see example in app/utils/analytics.ts

Aditi-Singh16:
For the storybook, The modal takes over the sidebar too, like it's not contained inside the screen, is this the correct behavior?
image

app/i18n/en/index.ts
Aditi-Singh16: Yes, not pushed yet

app/utils/feedback-helper.ts
```diff
@@ -0,0 +1,31 @@
...
+    feedbackSubmit(input: $input) {
+      errors {
+        message
+      }
+      success
+    }
+  }
+`
+
+export const sendFeedback = async (
```
nicolasburtey, line 16: you should use `useSendFeedbackMutation` instead. this feedback helper boilerplate is not needed

nicolasburtey (changes requested):
- tests are not passing
- boilerplate around mutation is not needed

app/components/custom-modal/custom-modal.tsx
```diff
@@ -130,8 +142,8 @@ const useStyles = makeStyles(({ colors }, props: UseStylesProps) => ({
     fontSize: 24,
     fontWeight: Platform.OS === "ios" ? "600" : "700",
     lineHeight: 32,
-    maxWidth: "80%",
-    textAlign: "center",
+    maxWidth: props.titleMaxWidth || "80%",
+    textAlign: props.titleTextAlignment || "center",
```
nicolasburtey, line 146: I don't think we need to override this? why do you think that is necessary?

having consistency across screen is important, so we would need a good reason to change the pattern

app/graphql/client-only-query.ts
```diff
@@ -91,3 +97,20 @@ export const updateColorScheme = (client: ApolloClient<unknown>, colorScheme: st
...
+  try {
+    client.writeQuery<FeedbackModalShownQuery>({
+      query: FeedbackModalShownDocument,
+      data: {
+        __typename: "Query",
+        feedbackModalShown: shown,
+      },
+    })
+  } catch {
+    console.warn("unable to update status")
```
nicolasburtey, line 114:
```suggestion
    console.warn("unable to update feedbackModalShown")
```

app/i18n/en/index.ts
```diff
@@ -874,9 +876,11 @@ const en: BaseTranslation = {
     joinTheCommunity: "Join the community",
     whatsapp: "WhatsApp",
     email: "Email",
+    enjoyingApp: "Enjoying the app?",
     statusPage: "Status Page",
     telegram: "Telegram",
     mattermost: "Mattermost",
+    thankYouText: "Thankyou for the feedback, would you like to suggest an improvement?",
```
nicolasburtey, line 883: still a typo?

nicolasburtey:
an updated video on the flow would be useful. also, please fix the typo on thank you (already commented on this 3 times!)

Aditi-Singh16:
> an updated video on the flow would be useful. also, please fix the typo on thank you (already commented on this 3 times!)

Updated the desc with the latest video

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,95 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
+import { useFeedbackModalShownQuery } from "@app/graphql/generated"
+import { setFeedbackModalShown } from "@app/graphql/client-only-query"
+import { SuggestionModal } from "./suggestion-modal"
+import { logAppFeedback } from "@app/utils/analytics"
 
 const SendBitcoinSuccessScreen = () => {
   const styles = useStyles()
+  const [isFeedbackModalActive, setIsFeedbackModalActive] = React.useState(false)
 
+  const [showImprovement, setShowImprovement] = React.useState(false)
```
UncleSamtoshi, line 29: This should be renamed to `showSuggestionModal`

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,95 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
+import crashlytics from "@react-native-firebase/crashlytics"
+import { useApolloClient } from "@apollo/client"
+import { useFeedbackModalShownQuery } from "@app/graphql/generated"
+import { setFeedbackModalShown } from "@app/graphql/client-only-query"
+import { SuggestionModal } from "./suggestion-modal"
+import { logAppFeedback } from "@app/utils/analytics"
 
 const SendBitcoinSuccessScreen = () => {
   const styles = useStyles()
+  const [isFeedbackModalActive, setIsFeedbackModalActive] = React.useState(false)
```
UncleSamtoshi, line 27: It seems like `isFeedbackModalActive` is unnecessary. In line 65 you can just call a function that calls `Alert.alert` directly instead of using a `useEffect`.

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,95 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
   const CALLBACK_DELAY = 3000
   useEffect(() => {
+    if (!modalShown) {
+      const feedbackTimeout = setTimeout(
+        () => setIsFeedbackModalActive(true),
+        FEEDBACK_DELAY,
+      )
+      return () => {
+        clearTimeout(feedbackTimeout)
+        setFeedbackModalShown(client, true)
```
UncleSamtoshi, line 70: `setFeedbackModalShown` should be set at the same time as the `Alert.alert` is called in line 65. Currently it is part of the hook clean up, so there is the potential that it does not get shown because the hook gets cleaned up before the modal is shown.

app/screens/send-bitcoin-screen/suggestion-modal.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <CustomModal
+      title={LL.support.thankYouText()}
+      minHeight={"50%"}
+      titleMaxWidth={"100%"}
+      titleTextAlignment={"left"}
+      toggleModal={dismissSuggestionModal}
+      isVisible={showSuggestionModal}
+      primaryButtonTitle={LL.common.submit()}
```
UncleSamtoshi, line 57: The primary button should be disabled if `improvement` is an empty string.

app/screens/send-bitcoin-screen/suggestion-modal.tsx
```diff
@@ -0,0 +1,94 @@
...
+  )
+}
+
+const useStyles = makeStyles(({ colors }) => ({
+  noteInput: {
+    color: colors.black,
+  },
+  field: {
+    padding: 10,
+    marginTop: 10,
```
UncleSamtoshi, line 89: `marginTop` and `marginBottom` may not be necessary here because it is already provided by the CustomModal.

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
UncleSamtoshi: This is still not resolved

UncleSamtoshi: I see. I think it would be good to add back in the alertShown then.

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,88 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
+      }, FEEDBACK_DELAY)
+      return () => {
+        clearTimeout(feedbackTimeout)
+      }
+    }
+    if (!showSuggestionModal) {
+      const navigateToHomeTimeout = setTimeout(navigation.popToTop, CALLBACK_DELAY)
+      return () => clearTimeout(navigateToHomeTimeout)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
UncleSamtoshi, line 72: You shouldnt disable exhaustive hooks, because it is a vector for introducing bugs.

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,88 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
 const SendBitcoinSuccessScreen = () => {
   const styles = useStyles()
-
+  const [showSuggestionModal, setShowSuggestionModal] = React.useState(false)
   const navigation =
     useNavigation<StackNavigationProp<RootStackParamList, "sendBitcoinSuccess">>()
 
+  const client = useApolloClient()
+  const feedbackShownData = useFeedbackModalShownQuery()
+  const modalShown = feedbackShownData?.data?.feedbackModalShown
```
UncleSamtoshi, line 33: I would rename this to `feedbackModalShown`

app/screens/send-bitcoin-screen/send-bitcoin-success-screen.tsx
```diff
@@ -11,37 +11,88 @@ import { RootStackParamList } from "@app/navigation/stack-param-lists"
...
+      }
+    }
+    if (!showSuggestionModal) {
+      const navigateToHomeTimeout = setTimeout(navigation.popToTop, CALLBACK_DELAY)
+      return () => clearTimeout(navigateToHomeTimeout)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [client, modalShown, LL, showSuggestionModal, navigation])
+
+  const showAlert = () => {
```
UncleSamtoshi, line 75: `showAlert` should be more descriptive. It should be something like `requestFeedback`

PR #2364, opened 2023-06-19
feat: handle onchain transaction hash null for pending unbroadcast state

sandipndev:
This is how the screen will look when an Onchain Transaction returns `null` transactionHash

### DIFF
--- a/app/screens/transaction-detail-screen/transaction-detail-screen.tsx
+++ b/app/screens/transaction-detail-screen/transaction-detail-screen.tsx
@@ -27,56 +27,7 @@ import { useAppConfig } from "@app/hooks"
 import { makeStyles, Text, useTheme } from "@rneui/themed"
 import { toWalletAmount } from "@app/types/amounts"
 import { isIos } from "@app/utils/helper"
-
-const useStyles = makeStyles(({ colors }) => ({
-  closeIconContainer: {
-    flexDirection: "row",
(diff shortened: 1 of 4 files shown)


app/screens/transaction-detail-screen/transaction-detail-screen.tsx
```diff
@@ -256,6 +207,12 @@ export const TransactionDetailScreen: React.FC<Props> = ({ route }) => {
       </View>
 
       <View style={styles.transactionDetailView}>
+        {settlementVia.__typename === "SettlementViaOnChain" &&
+          settlementVia.transactionHash === null && (
```
nicolasburtey, line 211: to help clarify what we're doing here, this check `settlementVia.__typename === "SettlementViaOnChain" &&
          settlementVia.transactionHash === null` should not be in the "view" but before we do the rendering with a proper variable name. like 
          
`const transactionHasBroadcasted = settlementVia.__typename !== "SettlementViaOnChain" &&
          settlementVia.transactionHash !== null`

sandipndev: I agree, addressed in https://github.com/GaloyMoney/galoy-mobile/pull/2364/commits/c1bdf32802f534cb47f3f10b20eaafaa852412cd

nicolasburtey (approved):

PR #2366, opened 2023-06-19
feat: add auth intro screen

UncleSamtoshi:


app/components/screen/screen.tsx
```diff
@@ -21,6 +21,7 @@ function ScreenWithoutScrolling(props: ScreenProps) {
 
   const preset = presets.fixed
   const style = props.style || {}
+  console.log(style)
```
nicolasburtey, line 24:
```suggestion
```

app/i18n/en/index.ts
```diff
@@ -735,6 +735,11 @@ const en: BaseTranslation = {
     learn: "I don't mean to badger you, but there's lot more to learn, dig in...",
     learnToEarn: "Learn to Earn",
   },
+  AuthIntroScreen: {
+    upgradeAccountRequirements: "Upgrading your account only requires a phone number",
+    signInCreateAccountRequirements: "Signing in or creating an account requires a phone number",
```
nicolasburtey, line 740: should we say something like "come back and use the trial account if you want to explore the app to get started"?

ios/Podfile.lock
```diff
@@ -986,4 +986,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: bb6529ab88d2e95007f12b67f3974e0e60c964a6
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.12.0
```
nicolasburtey, line 989: I did a `brew update` which updated the cocoapods so I not have a discrepancy with the one running on circleci, fyi

PR #2381, opened 2023-06-22
feat: send all onchain

sandipndev:
This PR adds the feature of sending all of an user's bitcoin/stablesats usd onchain. It uses the `onChainPaymentSendAll` mutation supported by the backend. It's a flow only available in "onchain" payment.

- [x] Use onChainPaymentSendAll
- [x] Hook Send All with Payment Details
- [x] Max Amount Button
- [x] ~Get Fee with user's balance~ (Can't do because backend always says "Payment of xxx exceeds balance yyy")
- [x] Add Language conversions for texts
- [x] Fix Tests

## How it looks

<table>
  <tr>
  <td>Selected USD Wallet</td>
  <td>Tapped "Max Amount" button</td>
  <td>Details Screen</td>
  </tr>
  <tr>
    <td>
      image 1
    </td>
    <td>
      image 2
    </td>
    <td>
      image 3
    </td>
  </tr>
</table>

---
Supersedes https://github.com/GaloyMoney/galoy-mobile/pull/1670

PS: It uses @UncleSamtoshi's payment detail logic layer for the send all handling so a review from him would be awesome.

### DIFF
--- a/__tests__/payment-details/helpers.ts
+++ b/__tests__/payment-details/helpers.ts
@@ -132,6 +132,7 @@ export const createSendPaymentMocks = (): SendPaymentMutationParams => {
     lnNoAmountUsdInvoicePaymentSend: jest.fn(),
     onChainPaymentSend: jest.fn(),
     onChainUsdPaymentSend: jest.fn(),
+    onChainPaymentSendAll: jest.fn(),
     onChainUsdPaymentSendAsBtcDenominated: jest.fn(),
     intraLedgerPaymentSend: jest.fn(),
     intraLedgerUsdPaymentSend: jest.fn(),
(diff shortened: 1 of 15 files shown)


sandipndev:
> What happens if you click send all and then switch wallets?

I think this video recording would clear all doubts:

https://github.com/GaloyMoney/galoy-mobile/assets/40622610/594d35e7-0b2e-4edf-91b4-f3a594971a1f

> Normally on the confirm payment details screen, the amount is the amount that the end user will receive. In the case of sendAll, we need to subtract the fees from the amount or display it differently so its clear its not the amount the end user will receive

For the fees we are currently setting it to null for the time being. When the backend supports querying fee without erroring, we can add it in and display how much fees the user might have to pay.

> You can't currently sendMax from lightning payments. We may want to introduce the following fields canSendMax and change the name of sendMax to isSendingMax

I have addressed this in my last two commits.

nicolasburtey (approved):

Issue #2384, opened 2023-06-22
Add Support to "Need help? Contact us." section in settings


pretyflaco:
Should be as follows so 1:1 support requests do not land in Telegram community chat. Suggested changes highlighted in bold below.

Status Page
Telegram (Community)
WhatsApp **(Support)**
Email **(Support)**

![image](https://github.com/GaloyMoney/galoy-mobile/assets/32958103/6dc05c48-21ab-4507-aab3-803ecdc2af5c)

nicolasburtey:
there are been some changes here, closing this ticket as probably not relevant any more

but @pretyflaco feel free to re-open if needed.

PR #2394, opened 2023-06-23
fix: Fix contact options

daviroo:
Fixes issue #2384 by removing telegram from the available support options when user taps "Need help".  Instead I've added a "Join the community" option on the main settings menu which also includes a link to mattermost.  I've utilised the existing `supportChannelsToHide` prop on the ContactModal to show/hide either 1:1 support options or community options.  I also changed the ContactModal which is available through the ContactSupportButton which is currently only used on the PhoneInput.  

Example screenshots below:
### 1:1 Support options
image

### Community options
image


nicolasburtey:
that's a great message improvement. thanks!

nicolasburtey (approved):

PR #2408, opened 2023-06-27
fix: properly hide balance across the home and send flow

### LINKED ISSUE #2460: Improve "hide balance" interactions based on user feedback
Users have reported issues with how hide balance feature works: 

1. When user has set "hide balance" in security, they would like for balances also to be hidden in the account selector in the Send flow 
2. When user taps balance on home screen to hide their balance, expectation is that recent transactions should also be hidden
3. When user hides their balance, closes the app, and re-opens the app, expectation is that the app remembers they toggled hide balance on last time they used the app. (even if default hide balance set to off)

nicolasburtey:
new behavior is as follow:
when the app load, it takes what is stored on the settings (in settings / security / hide balance)

when the user tap the balance on the header in the home, then, this switch the hide Balance mode. this switch only stays in memory and is not saved. 

the idea is the following: you may want to always have the balance hidden when using the wallet in the wild and have a sizeable balance. but if no-one is around you can tap the **** and see use the wallet without this feature. but then the next time you load the wallet, you don't want the hide balance to be deactivated.

fix #2460


app/hooks/use-hide-amount.ts
```diff
@@ -0,0 +1,21 @@
+import { useHideBalanceQuery } from "@app/graphql/generated"
+import { makeVar, useReactiveVar } from "@apollo/client"
```
UncleSamtoshi, line 2: I think this is the only place where we are using `makeVar` and `useReactiveVar` in the application. If we wanted to prevent adding a new patter, I think similar behavior could be accomplished with a context provider like `TemporaryHideAmountContext`.

UncleSamtoshi (approved):

PR #2450, opened 2023-07-03
feat: adding idempotency header for payment mutations

nicolasburtey:


app/screens/send-bitcoin-screen/use-send-payment.ts
```diff
@@ -113,41 +114,51 @@ gql`
...
 
+const getUuid = () => {
+  const randomBytes = Array.from({ length: 16 }, () => Math.floor(Math.random() * 256))
+  return uuidv4({ random: randomBytes })
+}
+
 export const useSendPayment = (
   sendPaymentMutation?: SendPaymentMutation | null,
 ): UseSendPaymentResult => {
+  const idempotencyKey = getUuid()
```
UncleSamtoshi, line 125: I believe this will generate a new idempotency key on rerenders essentially making it useless. We could use a useState to just generate one idempotency key, but there is is still the risk that the user just goes back one screen and then reenters this screen and is able to generate a new request. Potentially it could be good to generate an idempotency key deterministically based on the destination+amount+current-time. One unintended consequence of all these approaches is it essentially gets rid of the ability to retry a payment, so as part of this pr we should include a better defined flow for what happens when a payment fails.

UncleSamtoshi: I think the best way to handle this is to `useMemo` and then navigate to an error screen incase of an error.

UncleSamtoshi: I can work on the error screen.

nicolasburtey: how about generating an idempotency key when we start the sendFlow and sending it across screens as a param?

nicolasburtey: > One unintended consequence of all these approaches is it essentially gets rid of the ability to retry a payment, 

I don't think we want to retry payment really for onchain and intraledger, at least this is the intended behavior. we could remove the key for lightning (when there is an invoice) because there is an inherent non ability to pay an invoice twice enforced by lnd. that said I'm not sure we really need to allow a retry even on lightning. lnd already try to send the payment across different routes automatically.

UncleSamtoshi: This PR automatically sends someone to the error screen if a payment fails, so it removes the possibility of a retry https://github.com/GaloyMoney/galoy-mobile/pull/2452. All you would have to change for this PR is to `useMemo` with no dependencies to create the uuid.

sandipndev:
The code looks good to me, let me try to run this locally and check.

UncleSamtoshi (approved):

PR #2462, opened 2023-07-07
feat: set default account on second receive

UncleSamtoshi:
- prompt the user to set their default wallet on the second time they go to the receive screen

### DIFF
--- a/app/components/set-default-account-modal/set-default-account-modal.stories.tsx
+++ b/app/components/set-default-account-modal/set-default-account-modal.stories.tsx
@@ -1,0 +1,71 @@
+import React, { useState } from "react"
+import { View } from "react-native"
+
+import { MockedProvider } from "@apollo/client/testing"
+
+import { StoryScreen } from "../../../.storybook/views"
+import { createCache } from "../../graphql/cache"
+import { IsAuthedContextProvider } from "../../graphql/is-authed-context"
+import mocks from "../../graphql/mocks"
+import {
+  SetDefaultAccountModalUI,
+  SetDefaultAccountModalUIProps,
+} from "./set-default-account-modal"
+import { GaloyPrimaryButton } from "../atomic/galoy-primary-button"
+
+const SetDefaultAccountModalWithToggle: React.FC<SetDefaultAccountModalUIProps> = (
+  props,
+) => {
+  const [isVisible, setIsVisible] = useState(true)
+  const toggleModal = () => setIsVisible(!isVisible)
+
+  return (
+    <View>
+      <SetDefaultAccountModalUI
+        {...props}
+        isVisible={isVisible}
+        toggleModal={toggleModal}
+      />
+      <GaloyPrimaryButton title={"Open modal"} onPress={toggleModal} />
+    </View>
+  )
+}
+
+export default {
+  title: "Set Default Account Modal",
+  component: SetDefaultAccountModalUI,
+  decorators: [
+    (Story) => (
+      <MockedProvider mocks={mocks} cache={createCache()}>
+        <StoryScreen>{Story()}</StoryScreen>
+      </MockedProvider>
+    ),
+  ],
+}
+
+export const Default = () => {
+  return (
+    <IsAuthedContextProvider value={true}>
+      <SetDefaultAccountModalWithToggle
+        isVisible={true}
+        toggleModal={() => {}}
+        onPressBtcAccount={console.log}
+        onPressUsdAccount={console.log}
+      />
+    </IsAuthedContextProvider>
+  )
+}
+
+export const Loading = () => {
+  return (
+    <IsAuthedContextProvider value={true}>
+      <SetDefaultAccountModalWithToggle
+        isVisible={true}
+        toggleModal={() => {}}
+        onPressBtcAccount={console.log}
+        onPressUsdAccount={console.log}
+        loadingUsdAccount={true}
+      />
+    </IsAuthedContextProvider>
+  )
+}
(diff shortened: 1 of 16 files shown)


UncleSamtoshi:
image

app/components/set-default-account-modal/set-default-account-modal.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+  title: "Set Default Account Modal",
+  component: SetDefaultAccountModalUI,
+  decorators: [
+    (Story) => (
+      <MockedProvider mocks={mocks} cache={createCache()}>
+        <StoryScreen>{Story()}</StoryScreen>
+      </MockedProvider>
+    ),
+  ],
+} as Meta<typeof SetDefaultAccountModalUI>
```
nicolasburtey, line 45:
```suggestion
}
```

I think this Meta thing is no longer needed

nicolasburtey:
> * prompt the user to set their default wallet on the second time they go to the receive screen

the modal is been shown on the home screen, not receive screen?

app/screens/home-screen/home-screen.tsx
```diff
@@ -332,6 +355,11 @@ export const HomeScreen: React.FC = () => {
           </>
         ) : null}
         <AppUpdate />
+        <SetDefaultAccountModal
+          isVisible={setDefaultAccountModalVisible}
+          toggleModal={toggleSetDefaultAccountModal}
+          onSelectedAccount={onSelectedAccount}
```
nicolasburtey, line 361: what do we need to pass onSelectedAccount from `home` to the modal? 

the logic feel convoluted.

app/screens/home-screen/home-screen.tsx
```diff
@@ -137,6 +150,16 @@ export const HomeScreen: React.FC = () => {
 
   const onMenuClick = (target: Target) => {
     if (isAuthed) {
+      if (
+        target === "receiveBitcoin" &&
+        !hasPromptedSetDefaultAccount &&
+        dataAuthed?.me?.defaultAccount?.transactions?.edges?.length ===
```
nicolasburtey, line 156:
```suggestion
        dataAuthed?.me?.defaultAccount?.transactions?.edges?.length >=
```

nicolasburtey: should probably `more or equal` than `equal`. imagine you do the quiz, you will get multiple tx without going back to the home screen

app/screens/home-screen/home-screen.tsx
```diff
@@ -137,6 +150,16 @@ export const HomeScreen: React.FC = () => {
 
   const onMenuClick = (target: Target) => {
```
nicolasburtey, line 151: I'm not sure it's best to show this menu on the second time someone would want to receive. that could make sense but maybe having some modal after someone do successfully an action, not before, would be best? at the same time I understand that here it's also a great time to ask for this early.

app/screens/home-screen/home-screen.tsx
UncleSamtoshi: Its essentially a callback on completion of setting the account. If we put the logic to navigate to the receive screen after completion, it makes it so that we can not reuse the modal in other places.

app/screens/home-screen/home-screen.tsx
UncleSamtoshi: I have it set equal to `===` so that if a person logs out and logs back in, they shouldnt get asked again.

app/screens/home-screen/home-screen.tsx
UncleSamtoshi: I think it makes sense to place it here because you dont slow down the first interaction with the application but its natural to ask which wallet you want to receive to before you navigate to the receive screen.

UncleSamtoshi:
> > * prompt the user to set their default wallet on the second time they go to the receive screen
> 
> the modal is been shown on the home screen, not receive screen?

Yup they click receive, and then the modal pop ups and then they select a default wallet and then it navigates to the receive screen.

UncleSamtoshi:
> > > * prompt the user to set their default wallet on the second time they go to the receive screen
> > 
> > 
> > the modal is been shown on the home screen, not receive screen?
> 
> Yup they click receive, and then the modal pop ups and then they select a default wallet and then it navigates to the receive screen.

UncleSamtoshi:
It could be on the receive screen as well, but I didn't want to touch the receive screen while Sandipan was working on it.

app/screens/home-screen/home-screen.tsx
nicolasburtey: > Its essentially a callback on completion of setting the account. If we put the logic to navigate to the receive screen after completion, it makes it so that we can not reuse the modal in other places.

I understand the goal but unless a modal is used in multiple place I'm not sure this is useful to get this callback as it make the code more spread around for a maybe future use case.

I guess it's just a matter of perspective.

app/screens/home-screen/home-screen.tsx
nicolasburtey: fair assumption. not sure how we should deal with the earn section.

nicolasburtey (dismissed):

app/screens/home-screen/home-screen.tsx
UncleSamtoshi: Interesting, I didn't think of that scenario. I suppose it could be better to show it multiple times rather that have some people be skipped. Ultimately, I think we need a Apollo Federation with a separate api that keeps track of things like this.

app/screens/home-screen/home-screen.tsx
UncleSamtoshi: I see your point. I can remove it for now until we need to reuse the component.

app/screens/home-screen/home-screen.tsx
UncleSamtoshi: Although if we switch it to be `>==` it will also show up for every existing account. Which maybe could be good.

nicolasburtey (approved):

PR #2475, opened 2023-07-12
feat: add slider component

Aditi-Singh16:
https://github.com/GaloyMoney/galoy-mobile/assets/74586135/aac88ae0-db10-404a-b41c-182e2d8b63b3

I tested on Android 6.0 API level 23(low-end), as I saw the build.gradle minSdkVersion set to 23

https://github.com/GaloyMoney/galoy-mobile/assets/74586135/52970a7c-7341-4383-a8a6-9c6654d1cd87

### DIFF
--- a/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
+++ b/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
@@ -1,0 +1,19 @@
+import React from "react"
+import { Story, UseCase } from "../../../../.storybook/views"
+import { GaloySliderButton } from "./galoy-slider-button"
+
+export default {
+  title: "Galoy Slider Button",
+  component: GaloySliderButton,
+}
+
+export const Default = () => (
+  <Story>
+    <UseCase text="Default">
+      <GaloySliderButton initialText="Happy path" />
+    </UseCase>
+    <UseCase text="Disabled button">
+      <GaloySliderButton disabled initialText="Disabled" />
+    </UseCase>
+  </Story>
+)
--- a/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
+++ b/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
@@ -10,13 +10,13 @@ export default {
 export const Default = () => (
   <Story>
     <UseCase text="Default">
-      <GaloySliderButton initialText="default" />
+      <GaloySliderButton initialText="default" buttonSize={60} />
     </UseCase>
     <UseCase text="Disabled slider">
-      <GaloySliderButton disabled={true} initialText="Disabled" />
+      <GaloySliderButton disabled={true} initialText="Disabled" buttonSize={60} />
     </UseCase>
     <UseCase text="loading slider">
-      <GaloySliderButton loading={true} initialText="" />
+      <GaloySliderButton loading={true} initialText="" buttonSize={60} />
     </UseCase>
   </Story>
 )
--- a/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
+++ b/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
@@ -13,7 +13,10 @@ export const Default = () => (
       <GaloySliderButton initialText="default" />
     </UseCase>
     <UseCase text="Disabled slider">
-      <GaloySliderButton disabled initialText="Disabled" />
+      <GaloySliderButton disabled={true} initialText="Disabled" />
+    </UseCase>
+    <UseCase text="loading slider">
+      <GaloySliderButton loading={true} initialText="" />
     </UseCase>
   </Story>
 )
--- a/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
+++ b/app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
@@ -10,9 +10,9 @@ export default {
 export const Default = () => (
   <Story>
     <UseCase text="Default">
-      <GaloySliderButton initialText="Happy path" />
+      <GaloySliderButton initialText="default" />
     </UseCase>
-    <UseCase text="Disabled button">
+    <UseCase text="Disabled slider">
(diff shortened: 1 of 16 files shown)


app/components/atomic/galoy-slider-button/galoy-slider-button.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export const Default = () => (
+  <Story>
+    <UseCase text="Default">
+      <GaloySliderButton initialText="default" />
+    </UseCase>
+    <UseCase text="Disabled slider">
+      <GaloySliderButton disabled initialText="Disabled" />
+    </UseCase>
```
UncleSamtoshi, line 17: You should add a loading state here

UncleSamtoshi:
You need to run `yarn get-stories` in order for the stories to actually show up.

UncleSamtoshi:
I would check out this library or something similar https://www.npmjs.com/package/rn-swipe-button. The slider component from react-native-elements doesn't quite fit our usecase. We want the slider to automatically slide back to the beginning or the end if its close to either side. This is something you should get for free from this library or a similar one.

UncleSamtoshi:
Another library with potential is "react-native-swipe-verify"

nicolasburtey:
> Android.Emulator.-.Pixel_4a_API_30_5554.2023-07-12.13-33-07.mp4

- this button should only be in the last screen to confirm the payment
- there are some apparent rounding issue

Aditi-Singh16:
Hi @UncleSamtoshi , I tried both the libraries

1. rn-swipe-button
In this, for adding the arrow icon, it only takes image and not a component

2. react-native-swipe-verify
This works fine, but while sliding I get this warning
 ```
WARN  Animated: `useNativeDriver` was not specified. This is a required option and must be explicitly set to `true` or `false`
    at RNSwipeVerify (http://10.0.2.2:8081/index.bundle?platform=android&dev=true&minify=false&app=com.galoyapp&modulesOnly=false&runModule=true:397534:36)
    at RCTView
    at View (http://10.0.2.2:8081/index.bundle?platform=android&dev=true&minify=false&app=com.galoyapp&modulesOnly=false&runModule=true:48150:43)
```
So, when I looked at the code, I think [here ](https://github.com/darwin-morocho/RNSwipeVerify/blob/3256cc0769363e1b1fd49ddaec987e9891e30273/src/index.js#L109) ` useNativeDriver: false` should be added

So, should I fork this and add useNativeDriver: false or just use the first one but replace GaloyIcon with image?

**Comment by nicolasburtey** [comment]
> Hi @UncleSamtoshi , I tried both the libraries
> 
> 1. rn-swipe-button
>    In this, for adding the arrow icon, it only takes image and not a component
> 2. react-native-swipe-verify
>    This works fine, but while sliding I get this warning
> 
> ```
> WARN  Animated: `useNativeDriver` was not specified. This is a required option and must be explicitly set to `true` or `false`
>    at RNSwipeVerify (http://10.0.2.2:8081/index.bundle?platform=android&dev=true&minify=false&app=com.galoyapp&modulesOnly=false&runModule=true:397534:36)
>    at RCTView
>    at View (http://10.0.2.2:8081/index.bundle?platform=android&dev=true&minify=false&app=com.galoyapp&modulesOnly=false&runModule=true:48150:43)
> ```
> 
> So, when I looked at the code, I think [here ](https://github.com/darwin-morocho/RNSwipeVerify/blob/3256cc0769363e1b1fd49ddaec987e9891e30273/src/index.js#L109) ` useNativeDriver: false` should be added
> 
> So, should I fork this and add useNativeDriver: false or just use the first one but replace GaloyIcon with image?

we want to use the nativeDriver, otherwise the performance will not be good on low end phone

**Comment by nicolasburtey** [comment]
image

there is some missing opacity here. when the slider goes there should be a before/after. here the button is in the middle but the left is not covered. 

here is just the first example I found from google to illustrate the point: https://cdn.dribbble.com/users/81645/screenshots/5083565/04-slide-to-unlock_4x.png?compress=1&resize=400x300

**Comment by Aditi-Singh16** [comment]
how about this?
https://github.com/GaloyMoney/galoy-mobile/assets/74586135/06a9a687-4cbb-4b62-88f0-bf946431dbc0

**Comment by Aditi-Singh16** [comment]
How about this?

https://github.com/GaloyMoney/galoy-mobile/assets/74586135/e48505e9-d8bb-4e36-bbcf-ded1177be5f5

**Comment by nicolasburtey** [comment]
> How about this?
> 
>  Android.Emulator.-.Pixel_4a_API_30_5554.2023-08-01.13-29-03.mp4

we don't use shadow in other part of the app so we should not use this just here. 

also not a huge fan of the way the text disappear in the middle.

@agbegin @pretyflaco @sandipndev @UncleSamtoshi  what do you guys think?

**Comment by agbegin** [comment]
I would expect that swiping may turn the color from light/inactive to colored/active state as it moves from left to right.

**Comment by Aditi-Singh16** [comment]
okay, I will remove the shadow and instead of white, I will keep the same color but with low opacity?

**Comment by Aditi-Singh16** [comment]
@nicolasburtey @agbegin 
https://github.com/GaloyMoney/galoy-mobile/assets/74586135/2ed060bf-91a8-4b56-b0b1-bab48999c068

**Comment by nicolasburtey** [comment]
It's buggy?

I would remove the gradient if that makes it easier

Should prob be bigger in height

File: e2e/utils/controls.ts [hunk 2]
```diff
@@ -183,3 +183,25 @@ export const clickOnText = async (text: string) => {
...
+  const thumbWidth = size.width * thumbWidthPercentage
+  const startX = location.x + thumbWidth / 2
+  const startY = location.y + size.height / 2
+
+  const endX = location.x + size.width - thumbWidth
+  await driver.touchAction([
+    { action: "press", x: startX, y: startY },
+    { action: "wait", ms: 1000 },
+    { action: "moveTo", x: endX, y: startY },
+    "release",
```
ntheile, line 205: @Aditi-Singh16 
 we might just need to add another wait `{ action: "wait", ms: 5000 }` before `release` to make the e2e work better

MaxwellDG:
Readdressed in PR #2932

PR #2488, opened 2023-07-18
feat: add / remove phone number

nicolasburtey:


app/i18n/en/index.ts
```diff
@@ -842,7 +842,7 @@ const en: BaseTranslation = {
     learn: "I don't mean to badger you, but there's lot more to learn, dig in...",
     learnToEarn: "Learn to Earn",
   },
-  PhoneInputScreen: {
+  PhoneLoginInitiateScreen: {
```
UncleSamtoshi, line 845: In may make sense to unify `PhoneLoginInitiateScreen` and `PhoneRegistrationInitiateScreen` to `PhoneAuthInitiateScreen` just because they seem to be almost identical. Same with `PhoneLoginValidationScreen` and `PhoneRegistrationValidationScreen`.

app/i18n/raw-i18n/translations/qu.json
```diff
@@ -723,7 +723,7 @@
         "learn": "Ñuqaqa kaniyta mana yachanakuyta munanki, hinaspak kaniyta kawsayniyta kaniy, ch'usaq kaniyta...",
         "learnToEarn": "Yachanakuyta kaniyta kawsayniyta kaniy"
     },
-    "PhoneInputScreen": {
+    "PhoneLoginInitiateScreen": {
```
UncleSamtoshi, line 726: These manual fixes will get overwritten by transifex the next time we sync the languages. Perhaps there is a control in transifex to just rename a field.

app/screens/phone-auth-screen/phone-login-input.tsx
```diff
@@ -124,16 +125,14 @@ export const PhoneInputScreen: React.FC = () => {
     setCountryCode,
     supportedCountries,
     loadingSupportedCountries,
-  } = useRequestPhoneCode({
-    skipRequestPhoneCode: appConfig.galoyInstance.name === "Local",
```
UncleSamtoshi, line 128: Was this intentional to remove the local bypass? How will a user go through the phone flow locally?

UncleSamtoshi:
It looks like most of the Registration is copy and pasted from the Login flow. I think this will cause a longer term maintainability problem and it would be beneficial to either adapt the Login flow to handle registration or extract the common logic/display between the two.

app/screens/phone-auth-screen/phone-login-input.tsx
nicolasburtey: It's still here, just not on the screen but in the hook

nicolasburtey:
> It looks like most of the Registration is copy and pasted from the Login flow. I think this will cause a longer term maintainability problem and it would be beneficial to either adapt the Login flow to handle registration or extract the common logic/display between the two.

I have followed Justin heuristics of "if you have something only twice do not refactor but instead copy and you can later see a pattern when you have a third use"

So yeah there is some duplication but I don't know if this needs a refactoring now. 

The flow diverges in:
- the network calls
- use of captcha or not 
- error returned 

the rest is fairly identical indeed

No strong opinion here but if you strongly feel I should refactor I can do another pass

UncleSamtoshi:
Could be a topic for a code review, but I don't have a strong opinion at the moment either, so I will approve.

UncleSamtoshi (approved):

sandipndev:
image

It's strange to me that Remove Email is it's own button while remove phone number has a cross mark.

PR #2500, opened 2023-07-20
test: e2e test fix for iOS

Aditi-Singh16:


e2e/03-intraledger-flow.e2e.spec.ts
```diff
@@ -107,28 +107,28 @@ describe("Username Payment Flow", () => {
...
-    const contexts = await browser.getContexts()
-    const nativeContext = contexts.find((context) =>
-      context.toString().toLowerCase().includes("native"),
-    )
-    await browser.pause(3000)
-    if (nativeContext) {
-      await browser.switchContext(nativeContext.toString())
-    }
+  if (process.env.E2E_DEVICE === "android") {
+    it("Clicks on not enjoying app", async () => {
```
nicolasburtey, line 111:
```suggestion
    it("Dismiss enjoying app popup", async () => {
```

e2e/03-intraledger-flow.e2e.spec.ts
```diff
@@ -107,28 +107,28 @@ describe("Username Payment Flow", () => {
...
-    await browser.pause(3000)
-    const contexts = await browser.getContexts()
-    const nativeContext = contexts.find((context) =>
-      context.toString().toLowerCase().includes("native"),
-    )
-    await browser.pause(3000)
-    if (nativeContext) {
-      await browser.switchContext(nativeContext.toString())
-    }
+  if (process.env.E2E_DEVICE === "android") {
```
nicolasburtey, line 110: why is that needed?

nicolasburtey: fyi, tests should typically not be skipped like that. it's not a good practice. at a minimum, we should document why we are skipping it if that's necessary

nicolasburtey: also, why may it pass on other tests, like here? https://github.com/GaloyMoney/galoy-mobile/pull/2499 is it flacky or always not passing?

Aditi-Singh16: because for ios we have autodismiss alert, so not required for ios
first I thought this was causing the issue, but it was the useEffect

sandipndev:
@Aditi-Singh16 it can happen - I had to retry the e2e test for my branch as well. Let me analyse why that is failing.

e2e/03-intraledger-flow.e2e.spec.ts
nicolasburtey: > because for ios we have autodismiss alert, so not required for ios

no anymore (afaik)

Aditi-Singh16: Yes, I have added it in the feedback flow pr

PR #2503, opened 2023-07-21
chore: upgrade to rn 0.72

nicolasburtey:


sandipndev:
The React Native Reanimated issue is fixed and so is the vision camera setup. I could replicate https://github.com/software-mansion/react-native-reanimated/issues/4522#issuecomment-1725814674

Building on android works now.

Issue now: I get no jsbundle on the mobile screen.

PR #2508, opened 2023-07-23
feat: adding registration and deletion of 2fa in the mobile app

nicolasburtey:


app/i18n/raw-i18n/source/en.json
```diff
@@ -688,7 +688,28 @@
...
+        "totpDeleteAlertContent": "Are you sure you want to delete your two-factor authentication?"
+    },
+    "TotpRegistrationInitiateScreen": {
+        "title": "Two-factor authentication",
+        "content": "Scan this QR code with your authenticator app. Alternatively, you can manually copy/paste the secret into your authenticator app."
+    },
+    "TotpRegistrationValidateScreen": {
+        "title": "Two-factor authentication",
+        "enter6digitCode": "Enter the 6-digit code from your authenticator app to validate your two-factor authentication.",
+        "success": "Two-factor authentication has been enabled. You will now only be able to log back in with your phone or email AND this two factor authentication.\n\nOnly fully KYC account may be recovered from account that has lost access to their account and access to their two-factor authentication."
```
UncleSamtoshi, line 704: Rewording suggestion
```
Only full KYC accounts may be recovered in the case a user has lost access to their two-factor authentication.
```

app/screens/settings-screen/account-screen.tsx
```diff
@@ -450,6 +502,22 @@ export const AccountScreen = () => {
     },
   ]
 
+  if (beta) {
```
UncleSamtoshi, line 505: I think in other places to remove a setting we set `hidden: true` instead of dynamically adding it to the list.

app/screens/totp-screen/totp-login-validate.tsx
```diff
@@ -0,0 +1,96 @@
...
+        })
+
+        const success = response.status === 200
+        if (success) {
+          analytics().logLogin({
+            method: "email-2fa",
+          })
+
+          saveToken(authToken)
+          navigation.replace("Primary")
```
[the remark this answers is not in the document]
nicolasburtey: can do that

app/screens/settings-screen/account-screen.tsx
nicolasburtey: sure thing

app/i18n/raw-i18n/source/en.json
nicolasburtey: replaced the end of the sentence

UncleSamtoshi (approved):

Issue #2527, opened 2023-07-26
[bug]: Add domain validation to support only valid (old) static QR codes 


dolcalmi:
**Describe the bug**
previous to lnurl we used static QR codes with a galoy pay url (ex https://pay.blink.sv/dolcalmi) but now we have 3 valid domains so seems the mobile app is not validating what domains are valid to assume that we are trying to pay, like in the example, to dolcalmi@blink.sv.

So for example, you can use any domain/path and the mobile app will assume that path is a valid user, i.e, with https://www.domain.com/sats mobile app will try to pay to sats@blink.sv

**To Reproduce**

1. Read next QR code
![image](https://github.com/GaloyMoney/galoy-mobile/assets/2079600/e47f9f2b-3a8f-4c5c-a725-9153a89ed5cd)

2. Go to the payment process and validate lightning address in the confirmation screen

**Expected behavior**
Only have this behaviour with domains related to the app

**Smartphone (please complete the following information):**

- OS: iOS
- Version 2.2.104.555

sandipndev:
This is actually a bug in galoy-client. I found a completely opposite test here: https://github.com/GaloyMoney/galoy-client/blob/main/src/parsing/index.spec.ts#L552-L564

sandipndev:
<table><tr><td>Wrong</td><td>Wrong</td><td>Right</td></tr><tr><td><img height="200" src="https://github.com/GaloyMoney/galoy-mobile/assets/40622610/a4d6732e-e371-45d4-be8e-95456735ebda" /></td><td><img  height="200" src="https://github.com/GaloyMoney/galoy-mobile/assets/40622610/89653b8a-2879-4e16-9907-eb1022015a29" /></td><td><img height="200" src="https://github.com/GaloyMoney/galoy-mobile/assets/40622610/40e4d953-19ee-4861-b1f6-ee63dd96b3c4" /></td></tr></table>

PR #2539, opened 2023-08-02
chore: update galoy client

sandipndev:
fixes https://github.com/GaloyMoney/galoy-mobile/issues/2527


app/screens/send-bitcoin-screen/payment-destination/intraledger.ts
```diff
@@ -24,8 +24,17 @@ export const resolveIntraledgerDestination = async ({
...
 
+  if (!valid) {
+    return {
+      valid: false,
+      invalidReason: InvalidDestinationReason.WrongDomain,
+      invalidPaymentDestination: parsedIntraledgerDestination,
+    }
+  }
+
+  const handle = parsedIntraledgerDestination.handle
```
nicolasburtey, line 37: better to be consistent in how we parse an object, specially in the same file. ie: either use destructuring { valid } or use .handle.

nicolasburtey (commented):
one comment

nicolasburtey (approved):

PR #2540, opened 2023-08-02
refactor: receive screen design improvements

sandipndev:
fixes padding and stuff:

buttons should be smame height and border radius as other fields. (Tap to set amount is correct)
buttons should have 10px between them and 0px on the sides
inactive fields should be much more obvious.
BTC icon and lightning bolt icon should be the same as in rest of app

### DIFF
--- a/app/components/button-group/button-group.tsx
+++ b/app/components/button-group/button-group.tsx
@@ -97,9 +97,9 @@ const useStyles = makeStyles(
       justifyContent: "space-between",
       padding: 10,
       paddingVertical: 14,
-      marginRight: last ? 0 : 6,
+      marginRight: last ? 0 : 10,
       borderRadius: 8,
-      backgroundColor: selected ? colors.grey4 : colors.grey5,
+      backgroundColor: colors.grey5,
     },
     text: {
       fontSize: 16,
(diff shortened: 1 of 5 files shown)


app/components/button-group/button-group.tsx
```diff
@@ -71,34 +73,42 @@ export const ButtonGroup: React.FC<ButtonGroupProps> = ({
...
+const useStyles = makeStyles(
+  ({ colors }, { selected, last }: { selected: boolean; last: boolean }) => ({
+    button: {
+      flex: 1,
+      flexDirection: "row",
+      alignItems: "center",
+      justifyContent: "space-between",
+      padding: 10,
+      paddingVertical: 14,
+      marginRight: last ? 0 : 10,
```
nicolasburtey, line 100: I don't think this is the right way to do it. this problem may be solve in pure css only without having to pass external property.

I'm not an expert in CSS, but I think using some flex that would calculate interval automatically would be better. gpt4 is typically very good at it.

nicolasburtey (changes requested):
we should try to do a CSS only solution instead of passing argument around.

nicolasburtey (approved):

PR #2544, opened 2023-08-08
feat: codegen supergraph

ntheile:
This PR adds the codegen config for the supergraph so the mobile app can query `my welcome profile`.

The query will look like this:
```
query myWelcomeProfile {
    me {
      defaultAccount {
        ... on ConsumerAccount {
            welcomeProfile {
              leaderboardName
              allTimePoints
              allTimeRank
              innerCircleAllTimeCount
              innerCircleThisMonthCount
              outerCircleAllTimeCount
              outerCircleThisMonthCount
              thisMonthPoints
              thisMonthRank
            }
        }
      }
    }
  }
```

The hook should look something like this:

```
useMyWelcomeProfileQuery({
  fetchPolicy: "no-cache",
  skip: !isAuthed
})
```

TODO
- [x] Where to read in supergraph schema from? vendir from circles project? separate shared project?
   - [x] codegen from introspection url to include subgraphs
- [x] codegen.yml `skipDocumentsValidation: true` should be false but get `ERROR Fragment cannot be spread here as objects of type "Account" can never be of type "ConsumerAccount".`
- [x] `graphql-codegen` apollo federation docs https://the-guild.dev/graphql/codegen/docs/integrations/federation
- [x] v5 improvements roadmap for `graphql-codegen` apollo federation  https://github.com/dotansimha/graphql-code-generator/issues/8296#:~:text=Back%2Dend%20code%20generation%20issues


codegen.yml
```diff
@@ -56,3 +60,7 @@ generates:
         TotpCode: "string"
         Feedback: "string"
         Minutes: "string"
+        LeaderboardName: "string"
+        join__FieldSet: "string"
```
UncleSamtoshi, line 64: What is the purpose of adding lines 64 through 66?

package.json
```diff
@@ -141,13 +141,14 @@
     "@babel/preset-env": "^7.20.2",
     "@babel/runtime": "^7.20.13",
     "@graphql-codegen/add": "^4.0.0",
-    "@graphql-codegen/cli": "^3.3.0",
-    "@graphql-codegen/client-preset": "^3.0.1",
-    "@graphql-codegen/typescript": "^3.0.0",
-    "@graphql-codegen/typescript-operations": "^3.0.0",
+    "@graphql-codegen/cli": "^5.0.0",
```
UncleSamtoshi, line 144: Was upgrading all the codegen tools required to get it to work with the super graph?

ntheile: i was originally getting al ot of errors when running codegen, so the first task did was update all the packages. but the main package to make federation work is `@graphql-codegen/typescript-resolvers`

app/screens/phone-auth-screen/request-phone-code-registration.ts
```diff
@@ -73,15 +73,6 @@ gql`
...
   }
-
-  query supportedCountries {
-    globals {
-      supportedCountries {
-        id
-        supportedAuthChannels
-      }
-    }
-  }
```
ntheile, line 84: This was duplicated in the codebase

app/graphql/generated.gql
```diff
@@ -936,41 +936,27 @@ query network {
   }
 }
 
-query onChainTxFee($walletId: WalletId!, $address: OnChainAddress!, $amount: SatAmount!, $targetConfirmations: TargetConfirmations) {
-  onChainTxFee(
-    walletId: $walletId
-    address: $address
-    amount: $amount
-    targetConfirmations: $targetConfirmations
```
ntheile, line 944: targetConfirmations is a depreciated field that is not used in the mobile app

app/screens/send-bitcoin-screen/use-fee.ts
```diff
@@ -69,50 +69,33 @@ gql`
     $walletId: WalletId!
     $address: OnChainAddress!
     $amount: SatAmount!
-    $targetConfirmations: TargetConfirmations
   ) {
-    onChainTxFee(
-      walletId: $walletId
-      address: $address
-      amount: $amount
-      targetConfirmations: $targetConfirmations
```
ntheile, line 78: targetConfirmations is a depreciated field that is not used in the mobile app

codegen.yml
```diff
@@ -1,6 +1,6 @@
 overwrite: true
-schema: "https://raw.githubusercontent.com/GaloyMoney/galoy/main/src/graphql/public/schema.graphql"
-# schema: "../galoy/src/graphql/public/schema.graphql"
+schema: "https://api.staging.galoy.io/graphql"
```
nicolasburtey, line 2: we will always have to target a live environment from now on? couldn't we write the graphql somewhere?

ntheile: this is the introspection endpoint. i am not sure how to easily get the raw graphql otherwise because its in the private circles repo (i.e the https://raw.githubusercontent.com... url won't work because they have a `?token` secret appended to the end).

@krtk6160 do you have any ideas on how to write the `supergraph.graphql` somewhere public?

krtk6160: @bodymindarts Any thoughts on where we can keep the generated supergraph?
Should we sync it to the mobile repo like we sync the proto files?

nicolasburtey (approved):

PR #2565, opened 2023-08-18
chore: handle null welcomeProfile for new users

ntheile:


sandipndev:
I think welcomeProfile null would be same as lonely. We shouldn't communicate about the day thing because there are already talks to make the circles realtime. This PR doesn't have to be this big. I'm closing this for now and will be referencing this to in a commit on my bigger improvements PR.

PR #2588, opened 2023-08-21
feat: september challenge

sandipndev:


app/components/september-challenge/modal.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const acknowledgeModal = () => {
+    setIsVisible(false)
+  }
+
+  return (
+    <Modal
+      isVisible={isVisible}
+      backdropOpacity={0.8}
+      backdropColor={colors.white}
+      onBackdropPress={acknowledgeModal}
```
UncleSamtoshi, line 51: There is a flicker in this modal sometimes when you close it. I think it can be fixed by adding `backdropTransitionOutTiming={0}` to the modal props.

app/components/september-challenge/card.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <View style={styles.beside}>
+            <Text type="p1" bold>
+              {LL.Circles.septChallenge.title()}
+            </Text>
+            <Text color={colors.grey3}>{countDown}</Text>
+          </View>
+          <Text type="p2">
+            {LL.Circles.septChallenge.description({
+              innerCircle:
+                data?.me?.defaultAccount.welcomeProfile?.innerCircleThisMonthCount || 0,
```
UncleSamtoshi, line 81: This value should be 0 if it is not September yet.

app/i18n/en/index.ts
```diff
@@ -1143,6 +1143,7 @@ const en: BaseTranslation = {
     circlesGrowingKeepGoing:
       "Your circles grow as you welcome people to Blink – keep going!",
     points: "points",
+    innerCircleGrow: "Your inner circle grows when you send a Blink user their first sats!",
     calculatingYourCircles: "Calculating your circles...",
     innerCircleExplainer: "You're driving Bitcoin adoption with Blink, keep it up!",
     innerCircleExplainerCard: "I'm driving Bitcoin adoption with Blink!",
```
UncleSamtoshi, line 1149: Not related to this pr, but I think it would make sense to change  `peopleYouWelcomed: "people you welcomed"` to "peopleWelcomedByYou: "people welcomed by you"`. This is parallel to how we do it with `people welcomed by your circle".

app/components/september-challenge/card.tsx
```diff
@@ -0,0 +1,116 @@
...
+      } else {
+        const sLeft = (SEPT_1_EPOCH - dateNow) / 1000
+        setCountDown(secondsToDDMMSS(sLeft))
+      }
+    }, 1000)
+
+    return () => clearInterval(t)
+  }, [setCountDown])
+
+  if (Date.now() < new Date("10/1/2023").getTime()) return <></>
```
UncleSamtoshi, line 65: What is this supposed to do?

UncleSamtoshi (changes requested):

app/components/september-challenge/card.tsx
sandipndev: Good catch!

app/components/september-challenge/card.tsx
sandipndev: Fixed the logic.

app/i18n/en/index.ts
sandipndev: I agree with the discrepancy. These things should be identified much earlier at Figma imo.

UncleSamtoshi (approved):

Issue #2616, opened 2023-09-03
Improve address/LN address copying in receive flow


agbegin:
Combining a couple issues relating to copying invoices/addresses here.

### 1. Remove additional data from on-chain bitcoin address (receive flow) in app 
When user copies an on-chain address from Blink, the address is appended with additional data that is complicating the receive experience. 

Currently, no matter if user taps "copy", "share", the address itself, or the QR, the following data format is copied:
- bitcoin:bc1qtfjdzslm9psmg3ua97r7pvzlz4rzkx496eelr4?message=Pay%2520to%2520Blink%2520Wallet%2520User

Users have complained that in many cases they need to delete the extra data, and they're worried about sending an incorrect address when they do. 

Proposed steps: 
- Remove the extra data after the onchain address 
- Double-check whether it is appropriate to keep "bitcoin:" at the beginning of any or all of the copied strings. 
    - My understanding is having "bitcoin:" is related to BIP21 compatibility, but since we are not generating unified QRs, in what situation does it help any payer or payee to have "bticoin:"?  
    - Resources to look into it: 
        - https://bitcoin.design/guide/how-it-works/payment-request-formats/
        - https://bitcoinqr.dev/
    - Possible solution, if "bitcoin:" prefix is helpful
        - keep "bitcoin:" in the QR, but do not append it when the user taps on the address itself. In this instance, copy only the address. 
        
### 2. Enable user to copy Lightning Address
In the Paycode screen in Receive, when you tap "username@blink.sv" we should copy the Lightning Address, not the LNURL string.

PR #2617, opened 2023-09-04
refactor: receive screen

sandipndev:
Fixes Spanish Receive screen
Fixes https://github.com/GaloyMoney/galoy-mobile/issues/2616

### DIFF
--- a/app/i18n/raw-i18n/translations/es.json
+++ b/app/i18n/raw-i18n/translations/es.json
@@ -524,7 +524,7 @@
         "paycode": "Código de pago",
         "onchain": "Onchain",
         "bitcoin": "Bitcoin",
-        "stablesats": "Stablesats\n ",
+        "stablesats": "Stablesats",
         "regenerateInvoiceButtonTitle": "Regenerar la factura",
         "setUsernameButtonTitle": "Establecer nombre de usuario",
         "invoiceHasExpired": "La factura ha vencido",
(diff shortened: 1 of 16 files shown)


app/i18n/raw-i18n/translations/es.json
```diff
@@ -524,7 +524,7 @@
         "paycode": "Código de pago",
         "onchain": "Onchain",
         "bitcoin": "Bitcoin",
-        "stablesats": "Stablesats\n ",
+        "stablesats": "Stablesats",
```
nicolasburtey, line 527: this may need to be changed in transifex? (not sure if the sync goes both way, we'll see)

app/screens/settings-screen/account-id.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const AccountId: React.FC = () => {
+  const { data } = useAccountScreenQuery()
+  const { LL } = useI18nContext()
+
+  const styles = useStyles()
+
+  const accountId = data?.me?.id || ""
+  const last6digitsOfAccountId = accountId?.slice(-6).toUpperCase()
+
+  const xs = "********—****—****—****—****** "
```
nicolasburtey, line 19:
```suggestion
  const xs = "********—****—****—****—******"
```

sandipndev: The extra space gives the text some padding. Didn't add a style for that.

nicolasburtey (approved):

PR #2653, opened 2023-09-13
chore: more webview testing

nicolasburtey:


app/screens/price/price-history-screen.tsx
```diff
@@ -2,16 +2,60 @@ import { makeStyles } from "@rneui/themed"
...
       <PriceHistory />
+      {isAtLeastLevelTwo && (
+        <GaloyPrimaryButton
+          title={LL.PriceHistoryScreen.buyAndSell()}
+          onPress={() =>
+            navigate("webViewDebug", {
+              url: urlWebView,
+            })
+          }
+          style={styles.button}
```
UncleSamtoshi, line 56: If I remember correctly during the demo the button was flashing black when pressed and i think it is because this line should be `containerStyle={styles.button}`.

app/screens/webview/webview.tsx
```diff
@@ -23,6 +23,29 @@ export const WebViewScreen: React.FC<Props> = ({ route }) => {
...
+      webview.current.goBack()
+      return
+    }
+
+    navigation.goBack()
+  }, [canGoBack, navigation])
+
+  React.useEffect(() => {
+    navigation.setOptions({
+      headerLeft: () => <Button onPress={handleBackPress} title="< Back" />,
```
UncleSamtoshi, line 40: I think there should be a translated `go back` string here.

app/screens/price/price-history-screen.tsx
nicolasburtey: yup thanks

PR #2655, opened 2023-09-14
chore: using log out mutation when logging out

nicolasburtey:


dolcalmi (dismissed):

codegen.yml
```diff
@@ -1,6 +1,6 @@
 overwrite: true
-# schema: "https://api.staging.galoy.io/graphql"
-schema: "./supergraph.graphql"
+schema: "https://api.staging.galoy.io/graphql"
+# schema: "./supergraph.graphql"
```
UncleSamtoshi, line 3: Does this need to be changed back?

nicolasburtey: @sandipndev could be your issue?

sandipndev: It was indeed.

PR #2671, opened 2023-09-23
fix: Fixed error on scan screen permissions

### LINKED ISSUE #2670: when the user declined permission to look for photo in the library
signal-2023-09-23-113013

- the description should be more useful to the user as the action to perform
- the description should be translated

daviroo:
This fixes #2670.  This library provides error codes in the callback that you can inspect and respond to, but this library also doesn't seem to need permissions for the photo library (at least it didn't for any of my emulators).

https://github.com/react-native-image-picker/react-native-image-picker#ErrorCode


nicolasburtey (approved):

nicolasburtey:
thanks for the PR!

PR #2682, opened 2023-09-27
test(e2e): fix issues

sandipndev:
This PR aims at fixing and making the end to end test robust and more reliant.
- Removes the auth token flow, uses one of the test accounts on staging `+503650555{60..99}`
- In case BrowserStack test fails, it auto retries for a maximum of three times (that won't take more than 5 mins on BrowserStack) using the same built apk/ipa.

To achieve this, there are two source code changes as well now:
- Aside from copy the token, there's a share token button. This was needed to call the resets for that user to the backend. We can't get the clipboard on an iOS physical device so needed to use share.
- The phone number parsing happens on change of the phone number input. This is so that the default country data (fetched from ipapi.co) gets overridden when `+503...` number is entered.


sandipndev:
The scrolling issue is fixed now, it was something to do with gestures.

e2e/utils/graphql.ts
```diff
@@ -62,21 +62,31 @@ const createGaloyServerClient = (config: Config) => (authToken: string) => {
...
 }
 
-const getRandomToken = (arr: string[]): string => {
-  const randomIndex = Math.floor(Math.random() * arr.length)
-  console.log("Choosing token at random index: ", randomIndex)
-  return arr[randomIndex]
+const getRandomPhoneNumber = (): string => {
+  const randomDigits = Math.floor(Math.random() * 40 + 60) // Generates a number between 60 and 99
+    .toString()
+  return `+503650555${randomDigits}`
```
nicolasburtey, line 68: why not getting a number from a much larger pool such that there would statistically never be a potential existing account?

sandipndev: That's a good idea, we can fund that account initially using the test user and replenish test user's funds once in a while

e2e/utils/graphql.ts
```diff
@@ -62,21 +62,31 @@ const createGaloyServerClient = (config: Config) => (authToken: string) => {
...
-export const userToken = getRandomToken(authTokens)
+const TokenStore = {
+  token: "",
+}
+export const setUserToken = (token: string) => {
+  TokenStore.token = token
+}
+export const userToken = () => {
+  return TokenStore.token
+}
```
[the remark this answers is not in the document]
sandipndev: Ok I'll try doing that, wanted to use the repository pattern here.

nicolasburtey (approved):

PR #2699, opened 2023-10-04
chore: test onfido

nicolasburtey:


app/screens/full-onboarding-flow/full-onboarding-flow.tsx
```diff
@@ -0,0 +1,168 @@
...
+gql`
+  mutation onboardingFlowStart($input: OnboardingFlowStartInput!) {
+    onboardingFlowStart(input: $input) {
+      workflowRunId
+      tokenAndroid
+      tokenIos
+    }
+  }
+
+  query debugScreen {
```
UncleSamtoshi, line 25: Should this query be named `fullOnboardingScreen`?

app/screens/full-onboarding-flow/full-onboarding-flow.tsx
```diff
@@ -0,0 +1,168 @@
...
+      >
+        <Text style={styles.textHeader}>
+          {LL.FullOnboarding.accountVerifiedAlready()}
+        </Text>
+      </Screen>
+    )
+  }
+
+  return (
+    <Screen
```
UncleSamtoshi, line 123: It would be nice to lay out this screen similar to how we style other screens. Here is how its styled in the send flow. It uses padding of 20, and then the PrimaryButton is aligned with the bottom of the screen. https://github.com/GaloyMoney/galoy-mobile/blob/b60b1e086f3b3a4f47caf41f9a00dd010006bc28/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx#L643

app/screens/full-onboarding-flow/full-onboarding-flow.tsx
```diff
@@ -0,0 +1,168 @@
...
+      preset="scroll"
+      keyboardShouldPersistTaps="handled"
+      keyboardOffset="navigationHeader"
+    >
+      <View style={styles.screenContainer}>
+        <Text style={styles.textHeader}>{LL.FullOnboarding.requirements()}</Text>
```
UncleSamtoshi, line 129: Are you able to use a `type="h1"` instead of a unique text style.

app/screens/full-onboarding-flow/full-onboarding-flow.tsx
```diff
@@ -0,0 +1,168 @@
...
+const useStyles = makeStyles(() => ({
+  view: {
+    marginHorizontal: 20,
+    padding: 20,
+    borderRadius: 20,
+  },
+
+  textHeader: {
+    fontSize: 18,
+    marginBottom: 24,
```
UncleSamtoshi, line 159: Instead of using margin everywhere. The top level container can use the field `rowGap` which will space out all items in the container by that much.

PR #2714, opened 2023-10-18
chore: use nix flakes

sandipndev:
Using Nix Flakes in both the Dev Environment as well as the CI system


sandipndev:
Has been paused because or resource prioritisation onwards feature and UI work but having this would surely be a good to have. Will pick this up when I am a bit free of other tasks or is open for anyone else to pick up.

PR #2726, opened 2023-10-26
fix: QR code too small to be scan on my old LG

nicolasburtey:
value was changed from 260 to 195 in the receive screen refactoring

why was it changed? some issues with some particular devices?

https://github.com/GaloyMoney/galoy-mobile/blame/main/app/screens/receive-bitcoin-screen/qr-view.tsx#L139C19-L139C19


nicolasburtey:
example of an unscannable QR code

image

sandipndev:
Blink logo taking too much of the QR Code space 🤔

sandipndev (approved):

PR #2728, opened 2023-10-27
fix: use same invoice for NFC

sandipndev:
Uses same LN invoice as shown on receive screen along with handling <1 sat error ahead of time.

### DIFF
--- a/app/screens/receive-bitcoin-screen/use-receive-bitcoin.ts
+++ b/app/screens/receive-bitcoin-screen/use-receive-bitcoin.ts
@@ -387,10 +387,7 @@ export const useReceiveBitcoin = () => {
         const lnurlResponse = await result.json()
         if (lnurlResponse?.status?.toLowerCase() !== "ok") {
           console.error(lnurlResponse, "error with redeeming")
-          Alert.alert(LL.RedeemBitcoinScreen.redeemingError())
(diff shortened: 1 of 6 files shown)


app/screens/receive-bitcoin-screen/use-receive-bitcoin.ts
```diff
@@ -369,6 +369,37 @@ export const useReceiveBitcoin = () => {
...
+
+      const result = await fetch(url)
+      if (result.ok) {
+        const lnurlResponse = await result.json()
+        if (lnurlResponse?.status?.toLowerCase() !== "ok") {
+          console.error(lnurlResponse, "error with redeeming")
+          Alert.alert(LL.RedeemBitcoinScreen.redeemingError())
+          if (lnurlResponse?.reason) {
+            Alert.alert(lnurlResponse.reason)
+          }
```
nicolasburtey, line 393: if `lnurlResponse?.reason` is set, it should be in the same Alert as the previous one, probably on the second argument (the body of the message is the second argument, first is the title)

sandipndev: Ahh... copy pasta side effects

nicolasburtey (approved):

PR #2735, opened 2023-10-31
test(e2e): fixes second test calling first test

sandipndev:


e2e/helpers.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { TranslationFunctions } from "@app/i18n/i18n-types"
+import {
```
UncleSamtoshi, line 2: Is there any reason for this file to end with `.spec.ts`? Since its not a test I dont think it should have the `.spec`.

UncleSamtoshi (approved):

PR #2763, opened 2023-11-08
fix: duplicate pending transactions

UncleSamtoshi:
Current transaction pagination is broken and leads to duplicate pending transactions. This pr uses the dedicated pendingTransaction query and then filters out all paginated pending transactions.


app/screens/home-screen/home-screen.tsx
```diff
@@ -145,8 +150,21 @@ export const HomeScreen: React.FC = () => {
...
+
+  const transactions = useMemo(() => {
+    const transactions: TransactionFragment[] = []
+    if (pendingTransactions) {
+      transactions.push(...pendingTransactions)
+    }
+    const settledTransactions =
+      transactionsEdges
+        ?.map((edge) => edge.node)
+        .filter((tx) => tx.status !== TxStatus.Pending) ?? []
```
nicolasburtey, line 164: are we going to remove this filter after the use have updated to the new app?
because we can then remove the pending tx from the "main" tx query?

UncleSamtoshi: We need this deployed and then we can fix the backend to remove pending transactions from the transactions query and then we can remove this filter.

nicolasburtey (dismissed):

nicolasburtey (approved):

PR #2795, opened 2023-11-20
feat: use appcheck if present to replace captcha

nicolasburtey:


app/screens/phone-auth-screen/request-phone-code-login.ts
```diff
@@ -97,14 +98,31 @@ export const useRequestPhoneCodeLogin = (): UseRequestPhoneCodeReturn => {
...
   const { appConfig } = useAppConfig()
-  const skipRequestPhoneCode =
-    appConfig.galoyInstance.name === "Local" || appConfig.galoyInstance.name === "Staging"
 
   const [error, setError] = useState<ErrorType | undefined>()
   const [captchaRequestAuthCode] = useCaptchaRequestAuthCodeMutation()
 
   const { data, loading: loadingSupportedCountries } = useSupportedCountriesQuery()
+
+  const [appCheckToken, setAppCheckToken] = useState<string | undefined>()
```
UncleSamtoshi, line 115: There already exisits a `useAppCheckToken` that I think can replace this.

app/screens/phone-auth-screen/request-phone-code-login.ts
```diff
@@ -264,8 +229,52 @@ export const useRequestPhoneCodeLogin = (): UseRequestPhoneCodeReturn => {
...
+          setError(ErrorType.RequestCodeError)
+        }
+      }
+    }
+
+    const skipRequestPhoneCode =
+      appConfig.galoyInstance.name === "Local" ||
+      appConfig.galoyInstance.name === "Staging"
+
+    if (skipRequestPhoneCode) {
```
UncleSamtoshi, line 271: I think these functions are being called multiple times due to them being in a useEffect. Can you create a function that handles splitting along the paths and then call it after line 331 when the phone number is submitted. That would be similar to the previous implementation.

nicolasburtey: you would like `captchaPath` and `appCheckPath` to be outside the useEffect?

UncleSamtoshi: Yes. It seems like the trigger for the paths is when a user submits their phone number, so I think it is better if it is explicitly called as a function during submitPhoneNumber instead of called with a useEffect which runs when dependencies change.

nicolasburtey: So actually the fact it's been called by useEffect and not directly is currently on purpose. when `submitPhoneNumber ` (renamed to userSubmitPhoneNumber) is called, it needs to go through captcha for the normal flow, which then asynchronously update the dependency when captcha is called. 

It's quite a convoluted logic, and I actually have added some comment to try to clarify it, but this logic is not something this PR is changing, it was already like this before. 

Maybe a better design would be to go with a proper/tested "state flow management" with a `useReducer` like we have in the send bitcoin destination flow, but I would not that as part of this PR

We will want to replace the captcha library imo so maybe would be a good time to do that at this time?

UncleSamtoshi: That makes sense. I guess the only potential change would be to keep `registerCaptcha` as part of `userSubmitPhoneNumber` under an if statement to make the the captchaPath simpler. Up to you if you think that would improve it.

nicolasburtey: I would want to look at it more when switching the captcha library, most of the current would then be removed

Issue #2812, opened 2023-11-22
Login header


MaxwellDG:
**Issue:**
Trying to log in with phone sends the user to the login screen with the wrong header. 

Pressing this:
![phone-login-1](https://github.com/GaloyMoney/galoy-mobile/assets/51105802/2a36498b-c3dd-49d6-8364-eb7c83070f40)

Goes to this:
![phone-login-2](https://github.com/GaloyMoney/galoy-mobile/assets/51105802/9834fddb-8297-4299-8b71-65930bf17367)

**Proposal**
Display a different header if the user is creating an account vs logging in with a phone number. I noticed from some unused code that this looked like the initial intention, and the latter scenario was intended to display the string 'Phone number'.

sandipndev:
The header can just be renamed to "Authenticate" perhaps.

PR #2814, opened 2023-11-22
Refactor(phone-auth-screen): issue #2812 Updated headers on phone authentication screens to more accurately reflect the type of authentication (creating account versus logging in)

MaxwellDG:
There was some unused code (unused as in it was being overridden) that looked like it was trying to make the header during login to say 'Phone number' as the header. So I followed suit and used that.

For ease of reading the PR, take note of changing the optional property 'type' (in PhoneValidationStackParamList.phoneLoginValidate) to required. There was only 1 instance of this property not being used and that was in 'earns-section.tsx'. It used to pass in undefined and then default to PhoneLoginInitiateType.CreateAccount. Now it's stated there explicitly.

### DIFF
--- a/app/navigation/root-navigator.tsx
+++ b/app/navigation/root-navigator.tsx
@@ -43,6 +43,7 @@ import { AllContactsScreen } from "@app/screens/people-screen/contacts/all-conta
 import { PeopleTabIcon } from "@app/screens/people-screen/tab-icon"
 import {
   PhoneLoginInitiateScreen,
+  PhoneLoginInitiateType,
   PhoneLoginValidationScreen,
 } from "@app/screens/phone-auth-screen"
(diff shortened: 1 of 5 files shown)


MaxwellDG:
Header during both 'phoneFlow' screens for login:
Phone-number-1

Header during both 'phoneFlow' screens for creating account:
phone-login-2

nicolasburtey:
- the title of the PR is not very meaningful. release note are auto generated from it and "2812" will not really give insight about what this PR is about. 
- there is missing a dependency in useEffect as highlight by the linter

nicolasburtey:
thanks for this. the main reason we use the same screen today is because there is no differenciation in the API between an account creation and login back in, because we use a passwordless authentication system.

I guess that is fine to have different title, but eventually the confusion might still be around if someone is logging in with the wrong account because they will have a new account with a zero balance.

Issue #2853, opened 2023-11-29
Partial values for currencies that use commas instead of decimals


MaxwellDG:
**Issue**: There are dozens of currencies that use commas instead of decimals to differentiate between the 'dollar' and 'cent' (or whatever they call them) values. The library we're using for countries accounts for this but our numpad and input does not. This means anyone trying to input partial values is blocked. If you'd like to see for yourself, try Albanian money.

**Proposal**: Depending on the currency, swap out the decimal value for a comma (in all instances - meaning the partial values as well as the every-3 commas we usually input for ease-of-viewing) in the numpad and all its related functionality

MaxwellDG:
Closing because the data regarding using commas or decimals isn't there. For now it'll all be commas for every 3 digits, and decimals for fractional values.

Issue #2497, opened 2023-07-19
When setting custom amount allow pasting sat amount


pretyflaco:
**Is your feature request related to a problem? Please describe.**
Inputting a specific amount during invoice creation

**Describe the solution you'd like**
Allow pasting amount from clipboard

**Describe alternatives you've considered**
TINA

**Additional context**
E.g. Robosats contract step for bonding on buyer side requires specific sat amounts to be entered, and there are users asking for this feature for more convenience and usability - see https://twitter.com/eko2one/status/1681745875737903105

MaxwellDG:
I'll switch the **Text** components to **Input** components that have their state shared with the number pad. This should allow for pasting values in both the SAT and currenecy of choice fields

PR #2854, opened 2023-11-29
feat(amount-input-screen): allow for pasting values in amount input s…

MaxwellDG:
Addresses issue #2497 

I've added the ability to paste into the amount screen. 
This accounts for: 
- validating that it's only numerical and/or decimals
- not opening the keyboard
- maintaining that there can varying amounts of integers after the decimal (depending on the currency)
- maintains auto-formatting with commas for ease of viewing 
- combining the stateful input of pastes with numpad inputs

PS While working on this I noticed a large problem. There are dozens of currencies that use commas instead of decimals to differentiate between the 'dollar' and 'cent' (or whatever they call them) values. The library we're using for countries accounts for this but our numpad and input does not. This means anyone trying to input partial values is blocked. If you'd like to see for yourself, try Albanian money. I've opened an issue for this #2853

### DIFF
--- a/app/components/amount-input-screen/amount-input-screen-ui.tsx
+++ b/app/components/amount-input-screen/amount-input-screen-ui.tsx
@@ -59,7 +59,7 @@ export const AmountInputScreenUI: React.FC<AmountInputScreenUIProps> = ({
               showSoftInputOnFocus={false}
               onChangeText={(e) => {
                 // remove commas for ease of calculation later on
-                let val = e.replaceAll(",", "")
+                const val = e.replaceAll(",", "")
                 // todo adjust for currencies that use commas instead of decimals
 
                 // test for string input that can be either numerical or float
--- a/app/components/amount-input-screen/amount-input-screen-ui.tsx
+++ b/app/components/amount-input-screen/amount-input-screen-ui.tsx
@@ -6,7 +6,7 @@ import { GaloyIconButton } from "../atomic/galoy-icon-button"
 import { GaloyPrimaryButton } from "../atomic/galoy-primary-button"
 import { GaloyErrorBox } from "../atomic/galoy-error-box"
 import { CurrencyKeyboard } from "../currency-keyboard"
-import { Key, Keys } from "./number-pad-reducer"
+import { Key } from "./number-pad-reducer"
 
 export type AmountInputScreenUIProps = {
   primaryCurrencySymbol?: string
@@ -18,7 +18,7 @@ export type AmountInputScreenUIProps = {
   errorMessage?: string
   setAmountDisabled?: boolean
   onKeyPress: (key: Key) => void
-  onPaste: (key: Keys) => void
+  onPaste: (keys: number) => void
   onToggleCurrency?: () => void
   onClearAmount: () => void
   onSetAmountPress?: () => void
@@ -67,7 +67,7 @@ export const AmountInputScreenUI: React.FC<AmountInputScreenUIProps> = ({
                 // test for string input that can be either numerical or float
                 if (/^\d*\.?\d*$/.test(val)) {
                   const num = Number(val)
-                  onPaste(`${num}`)
+                  onPaste(num)
                 }
               }}
               containerStyle={styles.primaryNumberContainer}
--- a/app/components/amount-input-screen/amount-input-screen-ui.tsx
+++ b/app/components/amount-input-screen/amount-input-screen-ui.tsx
@@ -6,7 +6,7 @@ import { GaloyIconButton } from "../atomic/galoy-icon-button"
 import { GaloyPrimaryButton } from "../atomic/galoy-primary-button"
 import { GaloyErrorBox } from "../atomic/galoy-error-box"
 import { CurrencyKeyboard } from "../currency-keyboard"
-import { Key } from "./number-pad-reducer"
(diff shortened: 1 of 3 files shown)


nicolasburtey:
> There are dozens of currencies that use commas instead of decimals to differentiate between the 'dollar' and 'cent' (or whatever they call them) values. 

this has less to do with currency but more with the language / country settings actually.

ie: in France, we would show
100€
100$

while in the US we would have:
€100
$100

app/components/amount-input-screen/amount-input-screen-ui.tsx
```diff
@@ -53,13 +54,26 @@ export const AmountInputScreenUI: React.FC<AmountInputScreenUIProps> = ({
...
-            ) : (
-              <Text style={styles.faintPrimaryNumberText}>0</Text>
-            )}
+            <Input
+              value={primaryCurrencyFormattedAmount}
+              showSoftInputOnFocus={false}
+              onChangeText={(e) => {
+                // remove commas for ease of calculation later on
+                const val = e.replaceAll(",", "")
+                // todo adjust for currencies that use commas instead of decimals
```
nicolasburtey, line 63:
```suggestion
                // TODO adjust for currencies that use commas instead of decimals
```

nicolasburtey:
The Logic on the PR looks good but I think we need an refactoring iteration so that the code doesn't casting and has better function boundary

app/components/amount-input-screen/number-pad-reducer.ts
```diff
@@ -117,6 +117,26 @@ export const numberPadReducer = (
...
+
+        return {
+          ...state,
+          numberPadNumber: {
+            majorAmount: splitByDecimal[0],
+            hasDecimal: splitByDecimal.length > 1,
+            minorAmount: splitByDecimal[1] ?? "",
+          },
+        }
+      }
```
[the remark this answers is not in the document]
MaxwellDG: Yea i get that. Ill be back online in a couple days and Ill make a new reducer action. Makes sense following that pattern more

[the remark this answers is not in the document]
MaxwellDG: @nicolasburtey The type for 'Keys' I added felt a bit weird, but I believe its the best option. You'll notice starting from **amount-input-screen-ui.tsx** line 69 I go from string -> Number -> string -> Number -> string again. If you think that's weird let me know and I can explain the justification.

app/components/amount-input-screen/amount-input-screen-ui.tsx
```diff
@@ -53,13 +56,27 @@ export const AmountInputScreenUI: React.FC<AmountInputScreenUIProps> = ({
...
+                // remove commas for ease of calculation later on
+                const val = e.replaceAll(",", "")
+                // TODO adjust for currencies that use commas instead of decimals
+
+                // test for string input that can be either numerical or float
+                if (/^\d*\.?\d*$/.test(val)) {
+                  const num = Number(val)
+                  onPaste(`${num}`)
+                }
+              }}
```
nicolasburtey, line 72: is this code executed both for the case when we add a key or when we paste a string?

app/components/amount-input-screen/number-pad-reducer.ts
```diff
@@ -54,6 +61,8 @@ export const Key = {
 
 export type Key = (typeof Key)[keyof typeof Key]
 
+export type Keys = `${number | ""}${"." | ""}${number | ""}`
```
nicolasburtey, line 64: assuming Keys is what the user can paste, I don't think it make sense really a type. because ultimately the user could paste anything: a `string`, a `number`. I think we will want to validate / trim the input to make sure that if the user paste something incorrect, this is correctly handled.

nicolasburtey: said another way, I don't think we need a `Keys` type at all, and we can accept a `string` instead that should be tentatively converted to a number.

nicolasburtey:
The PR looks much better overall, but I think we need a proper validation on what the user is pasting, to properly manage the case when the input is not convertible into a number.

app/components/amount-input-screen/number-pad-reducer.ts
MaxwellDG: Yea i agree the Keys type seemed unnecessary. Removing that would streamline the code here too by removing unnecessary type conversions. Ill adjust the whole process to accept a string.

The safety and validation is already there too. Note the regex matching on line 68 in amount-input-screen-ui.tsx.

app/components/amount-input-screen/amount-input-screen-ui.tsx
MaxwellDG: I double-checked, and no - it's only when the user changes the input field directly themselves.

MaxwellDG:
> The PR looks much better overall, but I think we need a proper validation on what the user is pasting, to properly manage the case when the input is not convertible into a number.

You don't think the regex string matching is enough?

nicolasburtey:
> > The PR looks much better overall, but I think we need a proper validation on what the user is pasting, to properly manage the case when the input is not convertible into a number.
> 
> You don't think the regex string matching is enough? Currently if the value can't be formatted into a number then it is a no-op and doesn't allow it to be pasted.

ok I think that might be good enough

PR #2866, opened 2023-12-02
chore: moving TOTP out of beta

nicolasburtey:


app/screens/settings-screen/account-screen.tsx
```diff
@@ -186,6 +186,7 @@ export const AccountScreen = () => {
 
   const dataBeta = useBetaQuery()
   const beta = dataBeta.data?.beta ?? false
+  beta
```
nicolasburtey, line 189: doing that as to not delete the beta related code because it's useful to keep it here

sandipndev (dismissed):

PR #2889, opened 2023-12-08
chore: proper error message on 409 error

nicolasburtey:


nicolasburtey:
> I wonder if this case can ever be hit

yup there has been quite a few times, with routing failures and pending lightning paymant that are not confirmed where people retry the payment

PR #2890, opened 2023-12-08
chore: improving onchain queue display timer

nicolasburtey:
note: the screen doesn't auto refresh after the tx got send to the mempool, but I'll dive into this tomorrow with a follow up PR. I dont want delay the release for this detail.


sandipndev:
> the screen doesn't auto refresh after the tx got send to the mempool
I tried on main, it does update the screen when the transaction list updates via subscription
It does feel like it's not updating because there is a time delay between "now" and when the transaction actually did broadcast

PR #2893, opened 2023-12-08
feat: settings screen fresh look

sandipndev:
## Settings Screen Revamp with better UI/UX

This PR uses the latest designs and design elements from our Figma Board and implements them in the settings screen. The scope of this PR is only UI updates and it doesn't touch logic anywhere unless it is needed. 

Apart from functionality, some semantics of this PR:
- Settings into newer component wise layout
- Uses Skeleton when data inside the component loads
- Provides user feedback every time (Spinner or Alerts) so that the user never feels stuck

Screenshots: https://github.com/GaloyMoney/galoy-mobile/pull/2893#issuecomment-2002437545

If you like this PR, a +1 would be appreciated


sandipndev:
Comments I received via Mattermost:

- [ ] Explore Wallet Settings: Have all the settings visible
- [ ] Red cross for Email and Phone
- [ ] Move default account and notifications to Preferences
- [ ] Developer Dashboard in advanced
- [ ] POS Information Screen?
- [ ] Transaction Limits straight in Settings
- [ ] 2FA/TOTP in Security

sandipndev:
FYI this also would fail the end to end tests, but I will fix it in a different PR. Don't want to clutter the diff of this.

app/components/modal-tooltip/modal-tooltip.tsx
```diff
@@ -56,6 +56,8 @@ export const ModalTooltip: React.FC<ModalTooltipProps> = ({
   }
   const modalTitle = title || defaultTitle
 
+  console.log({ modalTitle, text })
```
nicolasburtey, line 59:
```suggestion
```

app/components/set-lightning-address-modal/set-lightning-address-modal.tsx
```diff
@@ -138,9 +138,10 @@ export const SetLightningAddressModalUI = ({
 }: SetLightningAddressModalUIProps) => {
   const {
     appConfig: {
-      galoyInstance: { lnAddressHostname, name: bankName },
+      galoyInstance: { lnAddressHostname },
     },
   } = useAppConfig()
+  const bankName = "Lightning"
```
nicolasburtey, line 144: variable name feels wrong here

app/i18n/en/index.ts
```diff
@@ -2331,6 +2336,15 @@ const en: BaseTranslation = {
     }
   },
   AccountScreen: {
+    fundsMoreThan5Dollars: "Your account has funds more than $5",
```
nicolasburtey, line 2339:
```suggestion
    fundsMoreThan5Dollars: "Your account has more than $5",
```

better wording?

app/i18n/en/index.ts
```diff
@@ -2289,6 +2289,11 @@ const en: BaseTranslation = {
     },
   },
   SettingsScreen: {
+    setByOs: "Set by OS",
+    pos: "Point of Sale",
+    posCopied: "Your point of sale link has been copied",
+    yourAddress: "Your {bankName: string} Address",
```
nicolasburtey, line 2295: probably no need for a input here if we always want `Lightning address`

app/i18n/en/index.ts
```diff
@@ -2331,6 +2336,15 @@ const en: BaseTranslation = {
     }
   },
   AccountScreen: {
+    fundsMoreThan5Dollars: "Your account has funds more than $5",
+    itsATrialAccount: "Trial accounts have reduced transaction limits and no recovery method. If you lose your phone or uninstall the app before you upgrade, your funds will be unrecoverable.",
```
nicolasburtey, line 2340:
```suggestion
    itsATrialAccount: "Trial accounts have reduced transaction limits and no recovery method. If you lose your phone or uninstall the app, your funds will be unrecoverable.",
```

app/screens/phone-auth-screen/phone-registration-validation.tsx
```diff
@@ -189,7 +188,6 @@ export const PhoneRegistrationValidateScreen: React.FC<
         logAddPhoneAttempt()
         const { data } = await phoneValidate({
           variables: { input: { phone, code } },
-          refetchQueries: [AccountScreenDocument],
```
nicolasburtey, line 192: why is that no longer needed?

app/screens/settings-screen/account/account-screen.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { UpgradeAccountLevelOne } from "./settings/upgrade"
+import { TotpSetting } from "../totp"
+
+export const AccountScreen: React.FC = () => {
+  const styles = useStyles()
+  const { LL } = useI18nContext()
+
+  const { isAtLeastLevelOne } = useLevel()
+
+  const items = {
```
nicolasburtey, line 26: `items` feel very/too generic for a variable name

nicolasburtey: if only `authenticationMethods` and `misc` is been passed down throught the props, we probably don't need to defined a high level items.

app/screens/settings-screen/account/id.tsx
```diff
@@ -28,42 +42,49 @@ export const AccountId: React.FC = () => {
...
-              .fill(null)
-              .map((_, i) => (
-                <View style={styles.circle} key={i} />
-              ))}
+    <View>
+      <Text type="p2" bold>
+        {LL.AccountScreen.accountId()}
+      </Text>
+      {loading ? (
+        <Skeleton style={styles.wrapper} />
```
nicolasburtey, line 50: I'm not sure a skeleton here would be the best UX. most of the data should be already available here in most case, and if not, shouldn't we already show the title of what is about the shown instead of a skeleton? 

skeleton feels more useful when the data has to be accurate from the beginning or it's not very determinisc what to expect. just my thought but we should ask opinion from more people on this.

app/screens/settings-screen/account/settings/email.tsx
```diff
@@ -0,0 +1,191 @@
...
+                name="refresh"
+                size="medium"
+                onPress={reVerifyEmailPrompt}
+              />
+            )}
+            {(bothEmailAndPhoneVerified || (email && !emailVerified)) && (
+              <GaloyIconButton name="close" size="medium" onPress={deleteEmailPrompt} />
+            )}
+          </View>
+        ) : undefined
```
nicolasburtey, line 179: maybe this logic should not be in the JSX part

app/screens/settings-screen/account/settings/phone.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+`
+
+export const PhoneSetting: React.FC = () => {
+  const { LL } = useI18nContext()
+  const { navigate } = useNavigation<StackNavigationProp<RootStackParamList>>()
+
+  const { loading, phone, emailVerified, phoneVerified } = useLoginMethods()
+
+  const [phoneDeleteMutation, { loading: phDelLoad }] = useUserPhoneDeleteMutation()
```
nicolasburtey, line 42:
```suggestion
  const [phoneDeleteMutation, { loading: phoneDeleteLoad }] = useUserPhoneDeleteMutation()
```

acronym not very easy to read imo

app/screens/settings-screen/account/settings/phone.tsx
```diff
@@ -0,0 +1,93 @@
...
+      subtitle={phone || undefined}
+      leftIcon="call-outline"
+      action={phoneVerified ? null : () => navigate("phoneRegistrationInitiate")}
+      spinner={phDelLoad}
+      rightIcon={
+        phoneVerified ? (
+          emailVerified ? (
+            <GaloyIconButton name="close" size="medium" onPress={deletePhonePrompt} />
+          ) : null
+        ) : undefined
```
nicolasburtey, line 89: `null` or `undefined`, my heart balance :)

app/screens/settings-screen/row.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [hovering, setHovering] = useState(false)
+  const styles = useStyles({ hovering, shorter })
+
+  if (loading) return <Skeleton style={styles.container} animation="pulse" />
+  if (spinner)
+    return (
+      <View style={[styles.container, styles.center]}>
+        <ActivityIndicator />
+      </View>
+    )
```
nicolasburtey, line 41: I'm confused of the difference between `loading` and `spinner`

app/screens/settings-screen/settings/account-ln-address.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { gql } from "@apollo/client"
+import { useAccountLnAddressQuery } from "@app/graphql/generated"
+import { toastShow } from "@app/utils/toast"
+
+gql`
+  query AccountLNAddress {
+    me {
+      username
+    }
+  }
```
nicolasburtey, line 19: have you look into how the network queries behave with this break down?

if `fetchPolicy` is not set, I believe it should be "cache-first" by default, but would not trigger query, but if it's not populated, it might send too many query to the server. 

not sure what would be the right solution. this is where relay from facebook shines

you should also ensure that mutation properties are been refetched from the server, because the default behavior from apollo client would be to show the cache value if available.

app/screens/settings-screen/settings/account-ln-address.tsx
```diff
@@ -0,0 +1,70 @@
...
+        subtitleShorter={(data?.me?.username || "").length > 22}
+        leftIcon="at-outline"
+        rightIcon={hasUsername ? "copy-outline" : undefined}
+        action={() => {
+          if (hasUsername) {
+            Clipboard.setString(lnAddress)
+            toastShow({
+              type: "success",
+              message: (translations) =>
+                translations.GaloyAddressScreen.copiedAddressToClipboard({ bankName }),
```
nicolasburtey, line 56: `bankName` may not be needed?

app/screens/settings-screen/settings/account-ln-address.tsx
```diff
@@ -0,0 +1,70 @@
...
+    me {
+      username
+    }
+  }
+`
+
+export const AccountLNAddress: React.FC = () => {
+  const { appConfig } = useAppConfig()
+  const hostName = appConfig.galoyInstance.lnAddressHostname
+  const bankName = "Lightning"
```
nicolasburtey, line 25:
```suggestion
```

nicolasburtey:
I made some comment looking over the code. have not checked the code yet. The restructuring of the screens with smaller component is great. I'd love to explore them with the storybook if that one worked :( 

we have to be mindful of the refactoring and the effect to network queries and how the cache is been managed. this was curated before to be updated when necessary (`fetchPolicy: cache-and-network`) but it doesn't seem this PR is considering this cache behavior right now. 

I think we should wait to merge the PR until the e2e tests also pass:
- there is no urgency to merge this PR
- I don't think there would be many merge conflict if we wait a little longer, no one is really going to do lot of work on those screens in the upcoming days

app/screens/phone-auth-screen/phone-registration-validation.tsx
sandipndev: Caching. Relevant fields can be queried back in the mutation and everywhere those fields get used will be updated automatically.

app/screens/settings-screen/account/id.tsx
sandipndev: The account ID header is displayed irrespective of loading state or not. The skeleton and the container have same height so the content also just appears when the data is loaded. I can show you a GIF if you want.

app/screens/settings-screen/account/settings/email.tsx
sandipndev: true, looks convoluted in my eyes as well

app/screens/settings-screen/account/settings/phone.tsx
sandipndev: lmao, yeah will make this DX better

app/screens/settings-screen/row.tsx
sandipndev: loading shows skeleton of the same width
spinner shows spinner (in case user has done some action - exporting transactions, deleting totp, etc)

sandipndev:
> we have to be mindful of the refactoring and the effect to network queries and how the cache is been managed. this was curated before to be updated when necessary (fetchPolicy: cache-and-network) but it doesn't seem this PR is considering this cache behavior right now.

Wherever a fetchPolicy was being used, I have made sure the same is transferred to the split components in this PR as well. To my testing, there has been no side effects regarding this.

sandipndev:
Storybook is now working: https://github.com/GaloyMoney/galoy-mobile/pull/2921
Once that gets merged I'll rebase and add the storybook components in this PR.

app/screens/settings-screen/settings/account-ln-address.tsx
sandipndev: this has now been added. thanks for the pointer

.circleci/config.yml
```diff
@@ -46,22 +47,11 @@ jobs:
...
-            curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
-
-            NODE_MAJOR=18
-            echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | sudo tee /etc/apt/sources.list.d/nodesource.list
-
-            sudo apt-get update
-            sudo apt-get install nodejs -y
+      - node/install:
+          install-yarn: true
+          node-version: '18.18.2'
```
nicolasburtey, line 52: We should use node 20?

Any reason why the 18 here?

__tests__/hooks/use-show-warning-secure-account.spec.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react"
-import { useShowWarningSecureAccount } from "@app/screens/settings-screen/show-warning-secure-account"
+import { useShowWarningSecureAccount } from "@app/screens/settings-screen/account/show-warning-secure-account-hook"
 import { renderHook } from "@testing-library/react-hooks"
 import {
   CurrencyListDocument,
```
nicolasburtey, line 5: PNG below should be a svg.

Only the screenshot can be a png but everything else (logo, i ona of buttons) should be svg.

Svg are lighter and scale in dimension, which is necessary for mobile

nicolasburtey (commented):
Some comments, still need to go over the major part of the code

app/graphql/generated.gql
```diff
@@ -778,6 +778,44 @@ query ContactsCard {
...
+      csvTransactions(walletIds: $walletIds)
+      __typename
+    }
+    __typename
+  }
+}
+
+query SettingsScreen {
+  me {
+    username
```
nicolasburtey, line 795:
```suggestion
  me {
    id
    username
```

I think we should always have the id here, otherwise it won't get cache properly

cc @UncleSamtoshi

app/i18n/en/index.ts
```diff
@@ -7,6 +7,7 @@ const en: BaseTranslation = {
     title: "Receive payment by using:",
     buttonTitle: "Set your address",
     yourAddress: "Your {bankName: string} address",
```
nicolasburtey, line 9: we should probably remove this string, specially if it's still used

app/i18n/en/index.ts
```diff
@@ -15,6 +16,7 @@ const en: BaseTranslation = {
     yourCashRegister: "Your Lightning Cash Register",
     yourPaycode: "Your Paycode",
     copiedAddressToClipboard: "Copied {bankName: string} address to clipboard",
```
nicolasburtey, line 18: we should probably remove this string, specially if it's still used

app/i18n/en/index.ts
```diff
@@ -2293,6 +2295,12 @@ const en: BaseTranslation = {
     pendingPayment: "The payment has been sent, but hasn't confirmed yet.\n\nIt's possible the payment will not confirm, in which case the funds will be returned to your account.",
   },
   SettingsScreen: {
+    setByOs: "Set by OS",
+    pos: "Point of Sale",
+    posCopied: "Your point of sale link has been copied",
+    yourAddress: "Your {bankName: string} Address",
```
nicolasburtey, line 2301: this should be removed probably

app/i18n/en/index.ts
```diff
@@ -2293,6 +2295,12 @@ const en: BaseTranslation = {
...
+    setYourAddress: "Set Your {bankName: string} Address",
```
nicolasburtey, line 2302: this should be removed probably

app/i18n/en/index.ts
```diff
@@ -2335,6 +2343,15 @@ const en: BaseTranslation = {
     }
   },
   AccountScreen: {
+    fundsMoreThan5Dollars: "Your account has more than $5",
```
nicolasburtey, line 2346: amount should be localized eventually in the local currency used (maybe it should be done as part of another PR)

app/screens/settings-screen/settings-context.tsx
```diff
@@ -0,0 +1,23 @@
+import { PropsWithChildren, createContext, useContext } from "react"
```
nicolasburtey, line 1: what is the value of this context? seems we could get along without it

if you're using it to manage the cache, then just reused `useSettingsScreenQuery` with a `fetchPolicy: "cache-only"` for all but the main component?

app/screens/settings-screen/totp.tsx
```diff
@@ -0,0 +1,94 @@
...
+      LL.AccountScreen.totpDeleteAlertTitle(),
+      LL.AccountScreen.totpDeleteAlertContent(),
+      [
+        { text: LL.common.cancel(), onPress: () => {} },
+        {
+          text: LL.common.ok(),
+          onPress: async () => {
+            setSpinner(true)
+            const res = await totpDeleteMutation()
+            await refetchTotpSettings()
```
nicolasburtey, line 58: can this throw?

e2e/03-intraledger-flow.e2e.spec.ts
```diff
@@ -138,14 +138,19 @@ describe("Username Payment Flow", () => {
...
+      await suggestionInput.waitForDisplayed({ timeout })
+      await suggestionInput.click()
+      await suggestionInput.setValue("e2e test suggestion")
+      await clickButton(LL.AuthenticationScreen.skip())
+
+      // FIXME: this is a bug. we should not have to double tap here.
+      await browser.pause(1000)
+      await clickButton(LL.AuthenticationScreen.skip())
+    } catch {
+      /* empty */
```
nicolasburtey, line 152: maybe add a comment on why the catch is necessary or useful instead of just `empty`

e2e/06-other-tests.e2e.spec.ts
```diff
@@ -45,7 +44,7 @@ describe("Change Language Flow", () => {
 
   it("navigates back to move home screen", async () => {
     await clickBackButton()
-    await waitTillSettingDisplayed(enLL.common.account())
+    await waitTillTextDisplayed(enLL.common.preferences() + "-group")
```
nicolasburtey, line 47: I think we should avoid manipulation string that represent object and ensure there is a proper to target them (without having to add `-group`).

e2e/helpers.ts
```diff
@@ -7,13 +7,19 @@ import {
...
-  waitTillSettingDisplayed,
+  waitTillTextDisplayed,
 } from "./utils"
 
 export const getAccessTokenFromClipboard = async (LL: TranslationFunctions) => {
   await clickIcon("menu")
-  await waitTillSettingDisplayed(LL.common.account())
+
+  if (process.env.E2E_DEVICE === "ios") {
+    await waitTillTextDisplayed(LL.common.preferences() + "-group")
```
nicolasburtey, line 19: same here

.circleci/config.yml
sandipndev: I was using Node 18 for development locally, does it work with Node 20?

__tests__/hooks/use-show-warning-secure-account.spec.tsx
sandipndev: Which PNG are you talking about?

sandipndev:
I'm planning to cleanup the translations file in a separate PR. would be best if there's a Python script to do this.

nicolasburtey:
> I'm planning to cleanup the translations file in a separate PR. would be best if there's a Python script to do this.

Why a python script? We don't use python in our stack

app/screens/settings-screen/totp.tsx
sandipndev: yes, it is handled now

e2e/03-intraledger-flow.e2e.spec.ts
sandipndev: done

nicolasburtey:
my first comments trying the PR locally

- first loading the settings is flacky. I don't think the loading template is defined correctly
- the username is not consistent. we have username in the Settings page, where we have @username in the Account page
- the fact we need to scroll in the Account page to see the main option is not too great. I'm not sure to know the goal of having this big logo here (as long as we don't have the ability to upload a picture, this is just taking the most important real estate space on this page). I'm also not sure why the username has to be back here. 
- the `point of sale` button is not super obvious to me how it's designed. if you talk the copy icon then it copy it, but if you tap anywhere else it's opening a new tab. this is not what I would have expected given the opening a new tab is just an appendix to the text "Point of Sales" and smaller in size than the Copy --> my expectation would be that copy would be the default action. but maybe the default / action should indeed be the tab opening, but then it should be the bigger icon here. 
- the fact that 2fa is a screen behind Security is akward in the flow, because all the other element in this Security screen do not open a new screen but are rather button that are present in app
- in the security screen, the hide balance button might not be visible if not scrolling down. the button should probably be appearing next to Hide Balance, not at the bottom of the text.

sandipndev:
For @designsats to take a look in the UI:

| | |
|---|---|
|  |  |
|  |  |

sandipndev:
Here are the updated screenshots:

| | |
|------------------------------|-----------------------------|
| Explore Wallet Settings Page | Explore Wallet Account Page |
| Simulator Screenshot - iPhone 15 Pro - 2024-03-17 at 17 31 12 | Simulator Screenshot - iPhone 15 Pro - 2024-03-17 at 17 31 24 |
| Logged In User Settings Page | Logged In User Account Page |
| Simulator Screenshot - iPhone 15 Pro - 2024-03-17 at 17 32 53 | Simulator Screenshot - iPhone 15 Pro - 2024-03-17 at 17 32 57 |
| Logged In User Security Page | |
| Simulator Screenshot - iPhone 15 Pro - 2024-03-17 at 17 45 07 | |

nicolasburtey:
spellcheck is going nuts lol

utils/find-unused-ll-keys.js
```diff
@@ -0,0 +1,97 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const fs = require("fs")
```
UncleSamtoshi, line 2: This is cool!

UncleSamtoshi:
PR looks good, and I really like the changes! It looks a lot prettier and feels more intuitive using the settings.

Two issues that i found that should be address, but could be done in a follow up PR:
- Deleting the account is slightly buggy and results in showing an alert for the session expiring. Likely the state is not being cleared correctly.
- When the lightning address is longer, the copy icon in the POS settings row is shifted right.

UncleSamtoshi (approved):

sandipndev:
Thank you @UncleSamtoshi for going through the PR thoroughly. I will address the two issues you flagged in a separate one now.

utils/find-unused-ll-keys.js
sandipndev: It has flagged a lot of unused keys and it'd be good to have them cleaned up eventually. Now that we have this on main, I think we'll be more vigilant.

Issue #2830, opened 2023-11-27
Center map automatically at user location


designsats:
Im using map almost every time Im paying irl to a Blink user. Its much better experience to find business on the map, click "Pay this business" and fill the amount than scanning QR. Ive noticed few other people operating like this. 

**Problem**
Map is always centered at El Zonte and I need to click the icon to pan to my location. 

**Solution**
Cache the last map position or center it at user (whichever is easier to code)

sandipndev:
We already ask for user's location when they open the maps tab. This would justify that request too.

MaxwellDG:
I agree that'd be useful. I'll get on it

PR #2895, opened 2023-12-08
Feature(maps-screen): center map on user's location when user first opens the maps tab

MaxwellDG:
This addresses issue #2830 

There are a few choices I made here that will need confirmation @nicolasburtey 

1. I decided to only centre the user upon first navigation to the Map screen. If they navigate away and return to the screen, it will continue with the current location instead of re-centring. I believe this is the correct decision because there is a button in the top-right for manual re-centring. 

<del>2. I added the Geolocation library config to the app.tsx file. Sometimes people don't like this and would prefer to have a wrapper component that does all the programmatic configs. Is there a different location you'd prefer this to be?</del>

<del>3. I don't have access to a macbook at the moment so I can't test on iOS. Either someone can test it for me, or it'll be a few days before I can.</del>

I think there's quite a few more improvements that are needed on the Maps screen. One of the first things I'd like to do after this is add a toggle somewhere in settings for allowing the user to change their location permissions. Otherwise they have to go into their phone settings and I know most users won't know how to do that easily.


nicolasburtey:
> ---------------- DO NOT MERGE YET, WIP -------------------

best is to use draft PR for that

nicolasburtey:
on the geolocation, I'm wondering if people will be comfortable sharing their location from the get go. maybe we should start by showing centering around the country (from the IP address) and then have a button for the user to allow sharing geolocation information?

@designsats @agbegin what do you guys think?

nicolasburtey:
for 2. if the component is used only in 1 place, I think having the config local is fine

agbegin:
> on the geolocation, I'm wondering if people will be comfortable sharing their location from the get go. maybe we should start by showing centering around the country (from the IP address) and then have a button for the user to allow sharing geolocation information?
> 
> @designsats @agbegin what do you guys think?

Agree, let the user trigger geo sharing when they're ready.

MaxwellDG:
@agbegin @nicolasburtey Okay so I'll center it around their IP by default. And if their account has geolocation turned on, then I'll center it around the fine_location. 

I imagine this toggle for geolocation should be on the user settings / preferences screens. This would require me to add a property to the user account object. Guess this means I'll be forced to venture into the server repo. I'll make a task there to be done first.

agbegin:
> The button to request could be in 1 of two places.

How about both? I imagine that the first time you want to use location, it would be while using the map. If you want to disable, you can always do so in settings. It could maybe stay on the map somewhere discreet as simple text that says "Location: Off" (by default, tap to turn on).. but I'll let you figure/recommend those details of UX 

FYI @sandipndev is actively working on settings so keep an eye on #2893

MaxwellDG:
> on the geolocation, I'm wondering if people will be comfortable sharing their location from the get go. maybe we should start by showing centering around the country (from the IP address) and then have a button for the user to allow sharing geolocation information?
> 
> @designsats @agbegin what do you guys think?

There are paid, unpaid, or "free-plan" options for getting a user's country based off of their IP.
1. Paid service that's a quick and easy API  (https://ipinfo.io/). 
2. Unpaid service that involves us downloading the IP-Country database. I imagine we'd throw this database onto our server and query it from the client (https://lite.ip2location.com/)
3. "Free-plan" on https://ipinfo.io has a 50,000/month limit

nicolasburtey:
We already use one service for getting ip to country on the onboarding flow.

We should use the same service. I don't think rate limit is an issue because each client will call the service, it's not coming from a single service. 

Maybe we could cache the result but even that I'm not sure we need it

MaxwellDG:
> We already use one service for getting ip to country on the onboarding flow.
> 
> We should use the same service. I don't think rate limit is an issue because each client will call the service, it's not coming from a single ip.
> 
> Maybe we could cache the result but even that I'm not sure we need it

It looks like the service you're using is broken due to rate-limiting issues. Here's a recording of Blink in production (downloaded from Play store). You probably never noticed cause it defaults to El Salvador.


https://github.com/GaloyMoney/galoy-mobile/assets/51105802/785ba24e-1e6e-4a0a-9109-6a6ccb09453d

error-1


I'll have to wait until someone else fixes this because I doubt you want to give me access to your ipapi account (hilarious name for a Spanglish speaker)

MaxwellDG:
@nicolasburtey you we're right about how the rate-limiting is done by device. So I added an asyncStorage cached version to fallback to.

You'll also notice the big .json file I added. Getting latitude/longitude from country code would've required using another API, so I figured this would be the most performant solution that didn't also didn't require utilizing another third-party service.

app/utils/location.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function getCountryFromIpAddress(): Promise<CountryCode> {
+  let countryCode = "SV" as CountryCode
+
+  try {
+    const response = await axios.get("https://ipapi.co/json/", {
+      timeout: 5000,
+    })
+    const _countryCode = response?.data?.country_code
+    if (_countryCode) {
+      await saveString("country_code", countryCode)
```
nicolasburtey, line 14: we currently rely on apollo to store value locally, ie:

https://github.com/GaloyMoney/galoy-mobile/blob/fe5bdbd4013e373a8b0a13dccb5d4985943eb106/app/graphql/cache.ts#L79C11-L79C22
https://github.com/GaloyMoney/galoy-mobile/blob/fe5bdbd4013e373a8b0a13dccb5d4985943eb106/app/graphql/client-only-query.ts#L104-L116

nicolasburtey: https://github.com/GaloyMoney/galoy-mobile/blob/fe5bdbd4013e373a8b0a13dccb5d4985943eb106/app/utils/theme-sync.tsx#L9

MaxwellDG: Adjusted to use graphql cache instead. Code snippet was real useful

app/screens/map-screen/map-screen.tsx
```diff
@@ -70,38 +99,84 @@ export const MapScreen: React.FC<Props> = ({ navigation }) => {
 
   const maps = data?.businessMapMarkers ?? []
 
+  // if getting location was denied and device's country code has been found (or defaulted)
+  // this is used to finalize the initial location shown on the Map
+  React.useEffect(() => {
+    if (countryCode && wasLocationDenied && !loading) {
+      // El Salvador gets special treatment here and zones in on El Zonte
```
nicolasburtey, line 106: most users in ES by now are not in El Zonte, so I think we don't need this exception, we could show the whole El Salvador map imo

app/graphql/cache.ts
```diff
@@ -79,6 +79,9 @@ export const createCache = () =>
           colorScheme: {
             read: (value) => value ?? "system",
           },
+          countryCode: {
+            read: (value) => value ?? null,
```
nicolasburtey, line 83:
```suggestion
            read: (value) => value ?? "SV",
```

you could alternatively use "SV" as the default value here, I think that would make CountryCode no longer optional in the type and simplify the code a little bit.

app/graphql/local-schema.gql
```diff
@@ -16,6 +16,7 @@ extend type Query {
   price: String # FIXME test only?
   beta: Boolean!
   colorScheme: String!
+  countryCode: String
```
nicolasburtey, line 19:
```suggestion
  countryCode: String!
```

if https://github.com/GaloyMoney/galoy-mobile/pull/2895/files#r1438319746 is implemented (no strong opinion but it may help to get cleaner code)

app/hooks/use-device-location.ts
```diff
@@ -0,0 +1,47 @@
...
+          })
+          const _countryCode = response?.data?.country_code
+          if (_countryCode) {
+            setCountryCode(_countryCode)
+            updateCountryCode(client, _countryCode)
+          } else {
+            console.warn("no data or country_code in response")
+          }
+          // can throw a 429 for device's rate-limiting. resort to cached value if available
+        } catch (e) {
```
nicolasburtey, line 30:
```suggestion
        } catch (err) {
```

we generally avoid 1 letter variable name

app/hooks/use-device-location.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const useDeviceLocation = () => {
+  const client = useApolloClient()
+  const { data } = useCountryCodeQuery()
+
+  const [loading, setLoading] = useState(true)
+  const [countryCode, setCountryCode] = useState<CountryCode>("SV")
+
+  useEffect(() => {
+    if (data) {
```
nicolasburtey, line 16: are you waiting for `data` for the throw case?

nicolasburtey: an alternative that would have less nesting would be to split the useEffect in 2:
- one useEffect that `setCountryCode` on data when its available (almost immediately)
- one useEffect that goes through `https://ipapi.co/json/` and `setCountryCode` again when the code is resolved. there is no longer a need to have the `catch` branch here because the value would already have been set.

functionality that should be the same but that may make code easier to read

package.json
```diff
@@ -64,6 +64,7 @@
     "@galoymoney/react-native-geetest-module": "^0.1.3",
     "@react-native-async-storage/async-storage": "^1.19.3",
     "@react-native-clipboard/clipboard": "^1.12.1",
+    "@react-native-community/geolocation": "^3.1.0",
```
nicolasburtey, line 67: does this have any effect on the permission required by the App stores?

app/screens/map-screen/map-screen.tsx
```diff
@@ -70,38 +99,84 @@ export const MapScreen: React.FC<Props> = ({ navigation }) => {
...
     }
-    asyncRequestLocationPermission()
-    // disable eslint because we don't want to re-run this function when the language changes
+
+    const negativeResponse = (error: GeolocationPermissionNegativeError) => {
+      console.debug("Permission location denied: ", error)
+      setLocationDenied(true)
+    }
+
+    Geolocation.requestAuthorization(permittedResponse, negativeResponse)
```
nicolasburtey, line 171: I've yet to try the PR locally, but what is this permission for? is it to show the current location of the user on the map? I think we should "wait for permission" in that case following the user pushing some button to give us the permission to ask for his precise geo localization. because asking for precise location for a bitcoin wallet is something bitcoiners can be wary of. 

maybe this could be done with another PR to not get this PR increase too much in scope

nicolasburtey:
great PR. provided some suggestions for improvements

app/screens/map-screen/map-screen.tsx
MaxwellDG: removed

app/graphql/cache.ts
MaxwellDG: changed

app/graphql/local-schema.gql
MaxwellDG: updated

app/hooks/use-device-location.ts
MaxwellDG: err

app/hooks/use-device-location.ts
MaxwellDG: I've put the two cases (useCountryCodeQuery and iPapi) together because they both need to finish in order for me to change **loading** to false. This is to indicate that the **initialLocation* has been established and the Map can go ahead and mount. I think having the two of them together shows that they're coupled.

As per the data throw case I agree that for good form, I should account for the error as well. I'll add another useEffect for that that'll setLoading to false and it'll allow the Map to render at the default 'SV' country code.

package.json
MaxwellDG: Android is already dealt with previously with this in the manifest
`<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />`

iOS i'm 99% sure. The docs say this is required and we already have it
`NSLocationWhenInUseUsageDescription`
This is not needed cause our deployment target is > 10
`NSLocationAlwaysUsageDescription`
And this one sucks and I really don't want to add it. The docs say it's mandatory, but I've tested it on iOS and it worked regardless
`NSLocationAlwaysAndWhenInUseUsageDescription`

app/screens/map-screen/map-screen.tsx
MaxwellDG: This behaviour was always in the app, it was just done before with the AndroidPermissions module instead (and iOS was done automatically). I agree though we can add that button later on so it's less pushy.

app/hooks/use-device-location.ts
MaxwellDG: For the useCountryCodeQuery I added an error case in a useEffect in case it throws. 

As per the dividing of code here, I think they should stay together because technically it's a race condition. useCountryCodeQuery would never realistically finish second, but nonetheless it's async and should be treated as possible. And I think it's best to show that all this is executing in the same order because I can only toggle **loading** to false when both are finished.

app/screens/map-screen/map-screen.tsx
nicolasburtey: > This behaviour was always in the app.

maybe I'm not sure to understand why the package `@react-native-community/geolocation` is necessary if this was the behavior before? (it's been years since I've coded this and I don't really remember all the details)

MaxwellDG: > > This behaviour was always in the app.
> 
> maybe I'm not sure to understand why the package `@react-native-community/geolocation` is necessary if this was the behavior before? (it's been years since I've coded this and I don't really remember all the details)

So the app in production right now gets permission to retrieve the user's location, but never actually does anything about it. Geolocation get's the user's coordinates after permission is given. And an added feature of the Geolocation library is that it asks for permissions automatically (which is why I was able to remove the AndroidPermissions section from before).

MaxwellDG: It's a small thing, but kind of the main point of this PR.

The app in prod currently gets the user's location permission and then does nothing about it. It does however give you access to that button in the top-right corner which will go to the user's location.

Geolocation allows for us to get the user's location before the MapView is created. Therefore it's required if we want to center on the user's location for the initial rendering.

nicolasburtey (approved):

PR #2898, opened 2023-12-08
fix: DarthCoin rant

nicolasburtey:
aka: saving unit of account used when typing an amount

### DIFF
--- a/app/components/amount-input-screen/amount-input-screen.tsx
+++ b/app/components/amount-input-screen/amount-input-screen.tsx
@@ -1,5 +1,5 @@
 import * as React from "react"
-import { WalletCurrency } from "@app/graphql/generated"
+import { WalletCurrency, useSwapDefaultUnitOfAccountQuery } from "@app/graphql/generated"
 import { CurrencyInfo, useDisplayCurrency } from "@app/hooks/use-display-currency"
 import { useI18nContext } from "@app/i18n/i18n-react"
 import { ConvertMoneyAmount } from "@app/screens/send-bitcoin-screen/payment-details"
@@ -19,6 +19,8 @@ import {
   NumberPadReducerActionType,
   NumberPadReducerState,
 } from "./number-pad-reducer"
+import { setSwapDefaultUnitOfAccount } from "@app/graphql/client-only-query"
+import { useApolloClient } from "@apollo/client"
 
 export type AmountInputScreenProps = {
   goBack: () => void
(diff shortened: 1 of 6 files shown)


Darth-Coin:
The force is strong with this one

ekzyis:
[This just did go down in bitcoin history :)](https://stacker.news/items/344399)

ralyodio:
i approve of this message.

app/components/amount-input-screen/amount-input-screen.tsx
```diff
@@ -198,11 +200,32 @@ export const AmountInputScreen: React.FC<AmountInputScreenProps> = ({
...
+
+  const onToggleCurrency = React.useCallback(() => {
+    if (secondaryNewAmount) {
+      setSwapDefaultUnitOfAccount(client, !swapDefaultUnitOfAccount)
       setNumberPadAmount(secondaryNewAmount)
-    })
+    }
+  }, [secondaryNewAmount, setNumberPadAmount, client, swapDefaultUnitOfAccount])
+
+  useEffect(() => {
```
UncleSamtoshi, line 217: Does this cause a flicker allowing it to render the initial amount and then on the next render switching to the default? I wonder if the change needs to be made before getting to this component, so that the first time it renders it is rendering the correct money amount.

nicolasburtey: I have not seen any flicker in my simulator. I think there were to do some calculation before rendering but I don't remember what it was anymore, I'll look for it

app/components/amount-input-screen/amount-input-screen.tsx
```diff
@@ -198,11 +200,32 @@ export const AmountInputScreen: React.FC<AmountInputScreenProps> = ({
     [currencyInfo],
   )
 
-  const onToggleCurrency =
-    secondaryNewAmount &&
-    (() => {
+  const client = useApolloClient()
+  const swapDefaultUnitOfAccount =
```
UncleSamtoshi, line 204: I think this implementation may have a bug. Its not a given that the primary amount is always the display currency or always the settlement currency. If swap was set to true and one use of the component passes in a display amount and another use of the component passes in a settlement amount, they would both end up swapping . A more accurate name would be `useDisplayAmountAsInitialAmount` and then you could explicitly handle the amount currencies before they reach this component.

app/components/amount-input-screen/amount-input-screen.tsx
nicolasburtey: nothing obvious coming to mind from my convo with GPT4. let's merge and see how it behaves and this is an issue?

app/components/amount-input-screen/amount-input-screen.tsx
nicolasburtey: in what scenario would we not have a display currency or settlement currency? ie: what would make it that the order would be different?

UncleSamtoshi: If on receive screen the initial primary amount was set to `0 Display` as the default and if on the send flow the primary amount was set to `0 Settlement` as the default. This isn't what we do now, but we dont have anything in place to prevent someone from misusing the component. 

Right now I believe we leave the amounts as undefined and then the initial amount is set in this line https://github.com/GaloyMoney/galoy-mobile/blob/c7c81a1f9d2df633e76752b964ab1a82555c3041/app/components/amount-input-screen/amount-input-screen.tsx#L147 with the `zeroDisplayAmount`. We could get rid of `needSwapCurrencyInit` and in this line use the value `useDisplayAmountAsDefaultAmount` to decide between `zeroDisplayAmount` and `zeroSettlementAmount` as the default.

Darth-Coin:
I just want to make some additional comments here, to clear up a bit.
I was talking about this screen, when user have to input the amount for the invoice.
image

I myself I did several times the mistake to input let's say 10000 in that field forgetting that is set automatically to EUR/fiat.
And twice I was even trying to pay that invoice that wasn't 10k sats but millions and wonder why is not going out.
Imagine my surprise when I found out going back Blink and see that the input was in fiat not in sats.

If I could make that mistake, imagine how a total noob could do it. And this can end up in big mistakes.
Yes, there's the switch button between fiat/sats but you have to press all the time before you input the amount.

If I set Blink in settings to display in BTC, I suppose to be displayed on all screens in BTC and not in fiat. Including the main top screen, that it display ONLY in fiat.

Make it optional: who wants to deal in fiat, fine, display in fiat all. But also I want to use it only in sats. I don't care too much about fiat.

I hope this comment make some clarifications and make your life easier with the changes.

nicolasburtey:
> If I could make that mistake, imagine how a total noob could do it. And this can end up in big mistakes.

Actually the reason we have fiat as default is because we consider the noob scenario as a key thing here. noob will always set their price in fiat term, so if we always have fiat by default, they're less likely to enter an amount under a wrong denomination

my assumption is this is a feature for a power user that want to have sats as the default unit of account when they type an amount

Darth-Coin:
Education of noobs start by changing their mindset, slowly, from fiat to BTC.
By making their life easier with displaying everything in fiat, you are not making anything else than a new Paypal over LN and they will never ever learn the lesson.
You are literally pushing the use of fiat doing that. Please don't say that is to help them. You are not helping them towards a Bitcoin standard.

Who wants to live in the fiat standard, fine. They can keep displaying fiat.
But many wants to move forward and break the fiat chains and embrace the Bitcoin standard.

PR #2903, opened 2023-12-09
BLIP 16 - display recently used contacts for quick send

marcoleder:
### Scenario 1: User types and list gets filtered accordingly
| Light Theme | Dark Theme |
|:-----------:|:----------:|
| Light AR | Dark AR |

---

### Scenario 2: User makes a typo (armo instead of arno) and gets feedback
| Light Theme | Dark Theme |
|:-----------:|:----------:|
| Light ARMO | Dark ARMO |

---

### Scenario 3: User selects from List and name then gets filled automatically in the search bar
| Light Theme | Dark Theme |
|:-----------:|:----------:|
| Light ARNO | Dark ARNO |


app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -63,6 +66,18 @@ gql`
...
+      id
+      contacts {
+        id
+        username
+        alias
+        transactionsCount
+      }
+    }
+  }
 `
```
nicolasburtey, line 81: there shouldn't be a separate query to the server when there is already one for this screen (this is a basic graphql optimization 101 of reducing the number of query back to the server to reduce latency and therefore provide a better UX)

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -111,6 +126,110 @@ const SendBitcoinDestinationScreen: React.FC<Props> = ({ route }) => {
...
+    if (contact.alias) {
+      contactPrettyNameMatchesSearchWord = contact.alias
+        .toLowerCase()
+        .includes(searchWord.toLowerCase())
+    } else {
+      contactPrettyNameMatchesSearchWord = false
+    }
+
+    return contactNameMatchesSearchWord || contactPrettyNameMatchesSearchWord
+  }
```
nicolasburtey, line 186: this function doesn't need to be in the component. it would actually be better outside because it would make it easier to unit test it

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -111,6 +126,110 @@ const SendBitcoinDestinationScreen: React.FC<Props> = ({ route }) => {
...
+          .split(" ")
+          .filter((text) => text.trim().length > 0)
+        const matchingContacts = allContacts.filter((contact) =>
+          searchWordArray.some((word) => wordMatchesContact(word, contact)),
+        )
+        setMatchingContacts(matchingContacts)
+      } else {
+        setMatchingContacts(allContacts)
+      }
+    },
```
nicolasburtey, line 229: this function should probably also be unit tested

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -111,6 +126,110 @@ const SendBitcoinDestinationScreen: React.FC<Props> = ({ route }) => {
...
+    data: contactsData,
+    error,
+  } = useContactsQuery({
+    skip: !isAuthed,
+    fetchPolicy: "cache-and-network",
+  })
+
+  if (error) {
+    toastShow({ message: error.message, LL })
+  }
```
nicolasburtey, line 140: have you tested this use case? I think it will likely loop indefinitely, ie: retrigger the toastShow message on every render. I would guess this might not be the intent.

nicolasburtey (commented):
great PR; awesome functionality. 

have provided some feedback to improve the code structure

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
marcoleder: Still getting started with graphql, my bad sorry!
[Fixed here](https://github.com/GaloyMoney/galoy-mobile/pull/2903/commits/98a4d174bec1131a50f9d99420be1f08eba58557#diff-9f28efbc29720d1ca95f1b3446fe21a4ee0ab32c4b0e0ff8e6bf989ebbd4c7c6)

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
marcoleder: Was taken directly from [screens/people-screen/contacts/all-contacts.tsx](https://github.com/GaloyMoney/galoy-mobile/blob/84eaa80cde72b5d8fc0ad65402066ff1ed6b72d0/app/screens/people-screen/contacts/all-contacts.tsx#L53-L55) but together with the useContactsQuery. 
Removed it, let me know what you think

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
marcoleder: Was taken directly from [screens/people-screen/contacts/all-contacts.tsx](https://github.com/GaloyMoney/galoy-mobile/blob/84eaa80cde72b5d8fc0ad65402066ff1ed6b72d0/app/screens/people-screen/contacts/all-contacts.tsx#L53-L55) but together with the useContactsQuery. 
Removed it, let me know what you think

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
marcoleder: Taken directly from [screens/people-screen/contacts/all-contacts.tsx](https://github.com/GaloyMoney/galoy-mobile/blob/84eaa80cde72b5d8fc0ad65402066ff1ed6b72d0/app/screens/people-screen/contacts/all-contacts.tsx#L90C3-L106C4)
If it shall be placed outside only here please confirm again :+1:

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
marcoleder: Also heavily inspired from [screens/people-screen/contacts/all-contacts.tsx](https://github.com/GaloyMoney/galoy-mobile/blob/84eaa80cde72b5d8fc0ad65402066ff1ed6b72d0/app/screens/people-screen/contacts/all-contacts.tsx#L72C3-L88C4)
Let me know if revision is _really_ needed

marcoleder:
Thanks for your feedback @nicolasburtey :=)

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
nicolasburtey: yes, we don't want logic in code to not be unit tested. things will break otherwise (now or on a later refactoring)

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
nicolasburtey: ok I guess this should have been manually tested then, and I may have this logic wrong. closing it.

nicolasburtey:
thanks for the further commits. some more feedbacks:
- we typically avoid the use of merge, we really use rebase to have a cleaner commit history
- one of your commit (the merge commit I think actually) has brought up a lot of png. not sure where those png comes from, but we are generally trying to not use any png for icons. compare to svg, they are heavier (and download apk size matters for the market we are addressing) and also doesn't scale on different screen, which is kind of important for android

PR #2905, opened 2023-12-09
chore: updating typescript to 5, removing ts-auto-mock

nicolasburtey:
ts-auto-mock is been deprecated

https://github.com/Typescript-TDD/ts-auto-mock/

and it was what prevented us to go to typescript 5

ts-auto-mock relies on ttypescript and ttypescript doesn't work with typescript 5

I tried ttsc unsuccesfully

I'm not sure if this is the best way to go forward, this PR adds quite a bit of not-very-useful code, mostly generated by GPT.

but otherwise not been able to update typescript is not ideal either


__tests__/payment-request/payment-request.spec.ts
```diff
@@ -19,62 +17,93 @@ const btcAmountInvoice =
...
+        errors: [],
+      },
+    },
+    errors: [],
+  }),
+)
 
 export const mutations: GeneratePaymentRequestMutations = {
+  // eslint-disable-next-line
+  // @ts-ignore
```
UncleSamtoshi, line 97: Why is ts-ignore rquired here? I think if we're going to add ts-ignore there should be a comment that explains why it is neccessary.

UncleSamtoshi:
I think this approach makes sense and will approve once the ts-ignore issue is addressed.

PR #2908, opened 2023-12-10
chore: improve log out experience in local dev

nicolasburtey:


UncleSamtoshi (approved):

UncleSamtoshi:
Strange that the tests are failing now.

nicolasburtey:
> Strange that the tests are failing now.

it's probably because I have update github action to node20. they were failing with another error message before. 

need to dive into why tests fail now. I think it has to do with dandling reference preventing jest to shutdown properly

PR #2910, opened 2023-12-11
chore: update lightning decoder used to a local blink one

nicolasburtey:


openoms (approved):
Improving the mobile view in https://github.com/GaloyMoney/dev.blink.sv/pull/35

PR #2914, opened 2023-12-12
chore: auto fill info on local dev

nicolasburtey:


UncleSamtoshi:
How common of a scenario is this for you? In general I think its preferable to not include special code for local testing, but I will approve if you think it is worth the improvement in dev flow.

nicolasburtey:
> How common of a scenario is this for you? In general I think its preferable to not include special code for local testing, but I will approve if you think it is worth the improvement in dev flow.

I'm doing this maybe 5 times a day or more when I'm working on the mobile wallet. I think it's worth the tradeoff.

nicolasburtey:
> > How common of a scenario is this for you? In general I think its preferable to not include special code for local testing, but I will approve if you think it is worth the improvement in dev flow.
> 
> I'm doing this maybe 5 times a day or more when I'm working on the mobile wallet. I think it's worth the tradeoff.

I'm frequently going between local dev and mainnet to test stuff, each of them provide different context. which is why I often have to log back in to the local dev environment

UncleSamtoshi (approved):

PR #2917, opened 2023-12-12
chore: improve message on successful 2fa activation

nicolasburtey:


app/i18n/en/index.ts
agbegin, line 1: - "Devices on this account which **are not logged in** with". (instead of "which have not login")
- Only level 2 accounts can be recovered by customer support in the event of lost 2FA access.

PR #2909, opened 2023-12-11
fix: btc destination input

MaxwellDG:
This was initially done in response to #2298 

There were several other minor UI improvements to be made, and 1 large refactoring done to address a lag issue on this input component. I'll reiterate the changes here:
  - There was a 1px jump from borderWidth on pasting
  - Buttons didn't have opacity indication when being pressed
  - Added a toast so users know what's happening when they press the paste button (I was confused when I was trying to press it and I didn't have anything in my clipboard. Initially thought it was broken)
  - Pressing the paste button twice was messing up the validation process and removing it on 2nd press (it went into manual 'entering' mode)
  - theres a multiple seconds lag on pressing the paste button for the first time which is very uncomfortable. Gonna work on making that more snappy. Likely it was only happening on Android
  
  Before merging, I'd like @krtk6160 to confirm if he sees the main issue already resolved here

PR #2919, opened 2023-12-13
Feat(send-bitcoin-destination-screen): UI improvements

MaxwellDG:
Closing the old PR #2909 in order to separate it into 2 separate PRs that address different things.

This PR makes a few UI improvements to the send bitcoin destination screen:

- fix 1px jump from borderWidth on pasting
- Buttons didn't have opacity indication when being pressed
- Added a toast so users know what's happening when they press the paste button (I was confused when I was trying to press it and I didn't have anything in my clipboard. Initially thought it was broken)
- Pressing the paste button twice was messing up the validation process and removing it on 2nd press (it went into manual 'entering' mode)
- multiple seconds of lag after pressing paste button on android

@nicolasburtey This PR has been updated with all the code changes you requested on the previous PR

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-reducer.ts
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-reducer.ts
@@ -1,6 +1,6 @@
 import { Destination, InvalidDestination } from "./payment-destination/index.types"
 
-const DestinationState = {
+export const DestinationState = {
   Entering: "entering",
   Pasting: "pasting",
   Validating: "validating",
@@ -25,7 +25,7 @@ export type SendBitcoinDestinationState = {
   destination?: Destination
 }
 
-const SendBitcoinActions = {
+export const SendBitcoinActions = {
   SetUnparsedDestination: "set-unparsed-destination",
   SetUnparsedPastedDestination: "set-unparsed-pasted-destination",
   SetValidating: "set-validating",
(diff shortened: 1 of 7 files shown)


app/screens/send-bitcoin-screen/send-bitcoin-reducer.ts
```diff
@@ -1,151 +1,142 @@
...
-export const DestinationState = {
-  Entering: "entering",
-  Validating: "validating",
-  Valid: "valid",
-  RequiresUsernameConfirmation: "requires-destination-confirmation",
-  Invalid: "invalid",
-} as const
-
-export type DestinationState = (typeof DestinationState)[keyof typeof DestinationState]
+export enum DestinationState {
```
nicolasburtey, line 3: we don't use `enum` in our codebase, it has some flaws for typescript IIRC / don't remember what exactly.

instead we use const object,with the `as const` to tell typescript that even the properties are immutable, ie: 

```
export const DestinationState = {
  Entering: "entering",
  Validating: "validating",
  Valid: "valid",
  RequiresUsernameConfirmation: "requires-destination-confirmation",
  Invalid: "invalid",
} as const
```

nicolasburtey:
except the enum everything else looks good

app/screens/send-bitcoin-screen/send-bitcoin-reducer.ts
MaxwellDG: Reverted the enums to how they were previously

nicolasburtey (approved):

Issue #2298, opened 2023-06-05
Double tapping input box doesn't select entire string


krtk6160:
**Describe the bug**
Double tapping inside destination input box does not select entire string.

**To Reproduce**

- video attached in screenshots section

**Expected behavior**
Entire string should be selected when double tapping the string inside the input box.

**Screenshots**

https://github.com/GaloyMoney/galoy-mobile/assets/17320781/ccefb7d3-2285-4bde-b28d-061c198af3ca

**Smartphone (please complete the following information):**

- Device: S21 5G
- OS: Android 12

MaxwellDG:
@krtk6160 Are you still experiencing this behaviour? I'm a bit confused here because in the video you posted it looks like 3 out of 4 times you tapped/double-tapped it did select the entire string. And I've tested it on two android devices and a double-tab is selecting the entire string.

Regardless, I can see this component needs some work. Things I've fixed so far:
- There was a 1px jump from borderWidth on pasting
- Buttons didn't have opacity indication when being pressed
- Added a toast so user's know what's happening when they press the paste button
- Double pressing paste button was confusing the validation process and removing it on 2nd press (it went into manual 'entering' mode)

nicolasburtey:
still seeing the same behavior by checking out from this PR

https://github.com/GaloyMoney/galoy-mobile/assets/11156370/20ad00c6-1255-46d7-9f1c-5d4df1569dca

MaxwellDG:
I've tried for a while now to get this working smoothly, but there's consistently some shortcomings.

Most solutions involved the prop **selection**, or forcing focus() but these all resulted in regressions elsewhere. The main issue is that the context menu can't be controlled by React Native in any reasonable manner. On top of that, the context menu behaves differently on different devices.

I can think of two possible solutions, but they are tedious so I'd like confirmation that it's worthwhile before I attempt them.
- A custom made context menu. We could disable the native context menu and create our own with all the 'copy', 'paste' etc... options
- Native modules. In Android you can control the context menu programmatically https://developer.android.com/reference/android/app/Activity.html#openContextMenu%28android.view.View%29
On iOS it's a bit easier apparently and you can control the contextMenu with ActionSheetIOS.showActionSheetWithOptions

nicolasburtey:
> I've tried for a while now to get this working smoothly, but there's consistently some shortcomings.
> 
> 
> 
[...]

PR #2920, opened 2023-12-13
feat: Can copy destination on send bitcoin screen

MaxwellDG:
Made as a substitute solution for issue #2298 

Allows for the the user to copy to clipboard on the send bitcoin screen. This is done here because it's really only a desired behaviour when the user has gotten to this screen from scanning a QR.

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
@@ -45,6 +45,7 @@ import { PaymentDestinationDisplay } from "@app/components/payment-destination-d
 import { useHideAmount } from "@app/graphql/hide-amount-context"
 import { ConfirmFeesModal } from "./confirm-fees-modal"
 import { GaloyIcon } from "@app/components/atomic/galoy-icon"
+import { toastShow } from "@app/utils/toast"
 
 gql`
   query sendBitcoinDetailsScreen {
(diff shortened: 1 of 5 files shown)


nicolasburtey:
> Made as a substitute solution for issue #2298
> 
> Allows for the the user to copy to clipboard on the send bitcoin screen. This is done here because it's really only a desired behaviour when the user has gotten to this screen from scanning a QR.

that looks good. I think we might want to show a Toast with a message "copied to the Clipboard" similarly in how we do it in the other place in the app - like when we paste from Clipboard in the previous screen - to give a better hint at the user at what this action is doing.

nicolasburtey:
I think it would also make sense to have the destination copy-able in the send confirmation screen

MaxwellDG:
> > Made as a substitute solution for issue #2298
> > Allows for the the user to copy to clipboard on the send bitcoin screen. This is done here because it's really only a desired behaviour when the user has gotten to this screen from scanning a QR.
> 
> that looks good. I think we might want to show a Toast with a message "copied to the Clipboard" similarly in how we do it in the other place in the app - like when we paste from Clipboard in the previous screen - to give a better hint at the user at what this action is doing.

Copying to clipboard automatically forces a toast on modern Android. So now there's two toasts whenever someone copies. It's not bad, but maybe it's something we could adjust in the toast hook another day. Specifically - if it's Platform.OS === Android and version > 13, then don't show toast for setting to clipboard.

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -432,11 +444,20 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
...
+            <View style={styles.disabledFieldBackground}>
+              <PaymentDestinationDisplay
+                destination={paymentDetail.destination}
+                paymentType={paymentDetail.paymentType}
+              />
+            </View>
+            <TouchableOpacity
+              style={styles.iconContainer}
+              onPress={copyToClipboard}
+              hitSlop={30}
```
nicolasburtey, line 457: I was not aware of this property. feels like a very useful one!

nicolasburtey (approved):

PR #2930, opened 2023-12-15
fix: pending transaction should be not be "success"

### LINKED ISSUE #2927: Timeout payment reported as successful
When ln send payment are timing out, we currently says they are successful, which is incorrect

nicolasburtey:
the issue is that the use case for queued transaction is conflated with the pending lightning transaction that may fail

fix #2927


app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx
```diff
@@ -92,26 +106,43 @@ const SendBitcoinSuccessScreen: React.FC<Props> = ({ route }) => {
...
+  }
+
+  const SuccessText = () => {
+    switch (status) {
+      case "SUCCESS":
+        return LL.SendBitcoinScreen.success()
+      case "QUEUED":
+        return LL.TransactionDetailScreen.txNotBroadcast({
+          countdown: formatTimeToMempool(
+            timeToMempool(arrivalAtMempoolEstimate as number),
```
nicolasburtey, line 127: @sandipndev I refactored `formatTimeToMempool` and needed to wrap the function with `timeToMempool`. this will fix the `27,123 minutes to confirm`

sandipndev (approved):

PR #2932, opened 2023-12-15
feat(galoy-slider-button): new sliding button implemented on send-bitcoin-confirmation-screen

MaxwellDG:
@nicolasburtey @UncleSamtoshi I revamped the slider component a bit. Let me know if you think it feels satisfying and if the UI is cool. 

PS <del>This is a draft cause I still have yet to get my XCode working. Gotta test on iOS to ensure it's all good. </del> And looks like I've gotta see what's going on with the e2e swiping test too.


nicolasburtey:
> @nicolasburtey @UncleSamtoshi I revamped the slider component a bit. Let me know if you think it feels satisfying and if the UI is cool.
> 
> Another thing to note is I don't believe the handleSendPayment function is correctly handling all error cases. There's a variety of things that could happen (I tried sending 10 sats and it couldn't calculate the fee and broke the handleSendPayment call) but regardless, there should be some more feedback there. The reason I'm bringing this up is because that can make the slider button hang. This has nothing to do with the slider component, and is instead related to the parent component not updating the **loading** state correctly.
> 
> PS This is a draft cause I still have yet to get my XCode working. Gotta test on iOS to ensure it's all good. And looks like I've gotta see what's going on with the e2e swiping test too.

something to check is whether this work for our end 2 end test also. this is a tricky thing to get right because it's not just a tap button (it's actually the goal of the slider!) but therefore the tests should also reproduce the human aspect of sliding the button to confirm

the tests are not running here probably because the PR is from your username and and not the main repo I guess? the e2e tests for Apple needs to run on a Mac machine and can't be run in github actions, so we use [circle](https://circleci.com/) for that.

MaxwellDG:
This is ready for review except for the e2e testing. Up to the team how soon you want the button integrated cause I have a feeling it'll be several days before e2e is fully functional.

nicolasburtey:
> This is ready for review except for the e2e testing. Up to the team how soon you want the button integrated cause I have a feeling it'll be several days before e2e is fully functional.

@sandipndev ?

bodymindarts:
@MaxwellDG can you please follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) - also in the title of the PR (as they get squashed on merge).

ie: `feat(...):. ...` and not uppercase.

PR #2942, opened 2023-12-18
chore: only ask for rate us on successful payment

nicolasburtey:


app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx
```diff
@@ -91,20 +91,24 @@ const SendBitcoinCompletedScreen: React.FC<Props> = ({ route }) => {
 
   const FEEDBACK_DELAY = 3000
   const CALLBACK_DELAY = 3000
+
   useEffect(() => {
-    if (!feedbackModalShown) {
+    if (!showSuggestionModal) {
```
UncleSamtoshi, line 96: I think this may be a bug. `showSuggestionModal` always starts as false until feedback is given. As it is written, the screen will navigate to home at the same time that feedback is requested.

nicolasburtey: this is indeed what it is doing: going to the home screen and asking for a feedback at the same time.

and this was also what was done before, so this PR doesn't change anything for this.

what behavior would you suggest? first ask for feedback and delay the return to home? or the opposite?

UncleSamtoshi: I think before it was requesting feedback and after that happened it would go to the home screen. Previously, line 100 would be an early return if feedbackModalShown was false and so it wouldn't call the setTimeout to navigate to the home screen in line 105. I think now this screen will be unmounted by the time a suggestion modal would be popped up.

PR #2956, opened 2023-12-26
feat: confirm slider on send bitcoin confirmation screen

MaxwellDG:
I made a typo on the previous PR #2932. useCallback was returning a function instead of calling it directly. 

This caused a pretty bad error last time, so I'd like to really ensure that I get the proper testing done beforehand. Gonna leave this here until then as a draft until I can get the e2e running locally.


nicolasburtey:
> I made a typo on the previous PR #2932. useCallback was returning a function instead of calling it directly.
> 
> This caused a pretty bad error last time, so I'd like to really ensure that I get the proper testing done beforehand. Gonna leave this here until then as a draft until I can get the e2e running locally.

the e2e are been rewritten so that they run on a local backend instead of on staging, but it will take a little while. I'm fine merging this now on my end.

MaxwellDG:
> > I made a typo on the previous PR #2932. useCallback was returning a function instead of calling it directly.
> > This caused a pretty bad error last time, so I'd like to really ensure that I get the proper testing done beforehand. Gonna leave this here until then as a draft until I can get the e2e running locally.
> 
> the e2e are been rewritten so that they run on a local backend instead of on staging, but it will take a little while. I'm fine merging this now on my end.

Okay. Please double-check it with a local test too for me to make absolutely sure it's functional. I felt bad getting this one into production last time with that error.

nicolasburtey:
> I felt bad getting this one into production last time with that error.

it was not in production. we do a nightly build precisely to catch those type of issues. only the team (and few folks who may be fetching the apk directly from github... but they would typically only fetch the release, not the pre-release) may have encounter the issue. so that's totally fine :)

nicolasburtey (approved):

nicolasburtey:
> > I felt bad getting this one into production last time with that error.
> 
> it was not in production. we do a nightly build precisely to catch those type of issues. only the team (and few folks who may be fetching the apk directly from github... but they would typically only fetch the release, not the pre-release) may have encounter the issue. so that's totally fine :)

I'm merging this now so we have this in the build for tomorrow.

I think one small improvement needed is to increase the size of the text, it's a bit too small. I'm also wondering if the font used is not different than the rest?

MaxwellDG:
> > > I felt bad getting this one into production last time with that error.
> > 
> > 
> > it was not in production. we do a nightly build precisely to catch those type of issues. only the team (and few folks who may be fetching the apk directly from github... but they would typically only fetch the release, not the pre-release) may have encounter the issue. so that's totally fine :)
> 
> I'm merging this now so we have this in the build for tomorrow.
> 
> I think one small improvement needed is to increase the size of the text, it's a bit too small. I'm also wondering if the font used is not different than the rest?

Yea it's a bit smaller, and a bit thinner, and slightly different shade of grey. Same font though, and the greys part of the theme.

I think it looks nice, but if you don't like it or think it's too inconsistent with the theme could you open another issue and I'll adjust it?

nicolasburtey:
maybe @designsats can provide some feedback after the next build is out

Issue #2955, opened 2023-12-26
ios Map markers sometimes jump to nearby marker


MaxwellDG:
**Issue**: On the map on iOS devices, sometimes when a user presses a Marker that is close to another Marker - it'll jump. It happens about 1.5 seconds after pressing the first marker. 

**Information**: I've investigated this for a while, and it appears to be due to iOS calling the **onPress** method twice, which is followed by an **onSelect** callback on the Marker component. React-native-maps has repeatedly acknowledged this issue, but hasn't fixed it for years. https://github.com/react-native-maps/react-native-maps/issues/1353 or https://github.com/react-native-maps/react-native-maps/issues/2492. There's tons of open issues about it.

**Solution**: There's a few options. None of them ideal.
1. I've created a solution that if this ever happens, it'll flash for half a second and then jump back to the correct marker. Required creating a ref for every single marker.
2. There's a proposed solution with 'pointerEvents="auto" on react-native-maps that makes the map respond really slow. About 1 full second before the map responds to any touch. Fixes the jumping though.
3. Completely remove the overhead popup for Markers on iOS and do everything with the custom interface I created in #2894  
4. Some people have gone in and directly patched the package using https://www.npmjs.com/package/patch-package. The key is to remove the 'onSelect' callback which is screwing everything up on iOS
5. Could try forcing iOS to use Google Maps. Not sure if it'll work but it's worth trying

@nicolasburtey @designsats Which of these solutions would you find satisfactory?

nicolasburtey:
I'm not sure I have a strong opinion on the element above, but I can share some context

The map was develop as a way to convince business, specially the small one (Tienda) to adopt bitcoin. it's actually quite a powerful tool, convincing one business to accept bitcoin because other surrounding business are also accepting bitcoin. this is the main way this is been used today, and continue to work well for new community

The frequency at which users are paying business directly from the map is not super common AFAIK. this is because:
- some businesses would also have a QR code, so that would be another way to pay them where you would be sure to pay the right business
- recurring customers are often using the contact tab and know who to pay directly
- ... and also because we haven't put as much work into the map, so maybe it's not used as much because it's a less pleasent UX

To come back to the earlier topic, I think a way to solve it also would be to do grouping, like what is been done on BTCMap and many other services where there is a high density of point. Instead of showing 10 overlapping cursors, we could just show 10 and the user will need to zoom to show the individual cursor, at which point it the current github issue would not longer be because the cursor would no longer overlap

PR #2968, opened 2023-12-29
feat: map speed increased

MaxwellDG:
Addresses #2955 

Loading time on Android used to be around 7 - 10 seconds. Now it's reduced to around 2.

This solution works for now, but it requires each map Marker to create a ref and use it in a useEffect. If the amount of markers reaches several thousands, we might have to revert back to the first solution proposed here which used a bottom-up slider.


MaxwellDG:
that's a lot of commits. let me redo that and clean it up.

Issue #2894, opened 2023-12-08
Speed of Maps screen


MaxwellDG:
**Issue:** Pressing the Map tab lags the app for several seconds (at least on Android. iOS often doesn't have these problems). This means the user has to wait a few seconds for the Map screen to load before they can navigate away. Very uncomfortable UI/UX feeling. 

**Cause:** There are a ton of Markers being added to the map. No matter where the user is looking, all Markers are being at least partially loaded.

**Solution**: At minimum, some sort of async thread should be loading the Markers into the Map 1 at a time (I think a generator function with yield could be cool for this). It could also load only the Markers that are seen in the screen as the user scrolls over them. In addition, some of the code could be memoized in that file in order to minimize re-renders.

MaxwellDG:
Here's what I've come up @designsats can you confirm this looks alright?

https://github.com/GaloyMoney/galoy-mobile/assets/51105802/e6deb8db-c4f7-416a-bcdb-87fe1ab95e30

- Map now loading all markers in 1.5 seconds.
- Aside from those 1.5 seconds during initial loading, the UI is no longer blocked. Namely, there's no other times that the bottom navigation bar won't respond to taps
- Added an animated bottom sheet that slides up when user taps on a marker
- A user can go to 'Send bitcoin' screen by tapping the 'Pay business' button, or tapping the white box with their name above the marker will also do that
- Ensured that the Google Maps toolbar is still in view with an additional change to animation timing

nicolasburtey:
> Here's what I've come up @designsats @sandipndev can you confirm this looks alright?
> 
>  map-speed.mp4 
> * Map now loading all markers in 1.5 seconds.
> * Aside from those 1.5 seconds during initial loading, the UI is no longer blocked. Namely, there's no other times that the bottom navigation bar won't respond to taps
> * Added an animated bottom sheet that slides up when user taps on a marker
> * A user can go to 'Send bitcoin' screen by tapping the 'Pay business' button, or tapping the white box with their name above the marker will also do that
> * Ensured that the Google Maps toolbar is still in view with an additional change to animation timing

I like the implementation given the constraints. 

I think I generally prefer the floating window as opposed to the dock one, but getting a frozen screen while it's been loading up is not ideal either. specially if this time is linear to the number of point of the map, which is increasing by the week.

MaxwellDG:
> > Here's what I've come up @designsats @sandipndev can you confirm this looks alright?
> > map-speed.mp4
[...]

PR #2969, opened 2023-12-29
feat: map markers load the callout section after they're pressed

MaxwellDG:
Addresses #2894 

Final test on iOS needed before removing draft status of PR


MaxwellDG:
duplicate

PR #2971, opened 2023-12-30
test(e2e): making the e2e testing suite more reliable with detox and local backend

sandipndev:
For a very long time now, e2e tests on mobile has been very unreliable and eventually we have gotten into the habit of ignoring the circleci test results.

This PR takes this problem head on. I have found out:
- Appium is not really ideal and often what happens is that it misses out points (example - double tap to remove keyboard)
- Appium is not "native"

This is a draft PR at the moment and I have introduced Detox as an alternative to Appium. The detox tests will run on iOS Simulator or Android Emulator and they can just directly be run on CI.

BlueWallet uses Detox: https://github.com/BlueWallet/BlueWallet/blob/master/.github/workflows/ci.yml


sandipndev:
With GitHub Actions, the time taken is a lot. Perhaps it's not worth it to use GitHub Actions for this because we already have a setup at CircleCI. Moving over to self hosting on CircleCI.

sandipndev:
Will split this PR into multiple for better review.

PR #2976, opened 2024-01-03
feat: increase speed of map markers initial load

MaxwellDG:
Improves #2894.

The solution here was to only render buttons within the marker callout section after the marker has been pressed. Then manually open the callout once that's happened.

Fortunately, I believe that this solution will aid in finding a good solution for #2955. With this PR as is, there is a side-effect behaviour regarding pressing an iOS marker that's very close to another one. Instead of jumping to a nearby marker (the previous erroneous behaviour), they now won't show a callout at all. Obviously neither of these behaviours are good enough, but it should be noted that this change has occurred - and will be improved upon in an upcoming PR.


app/components/map-interface/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  readonly mapInfo: {
+    readonly __typename: "MapInfo"
+    readonly title: string
+    readonly coordinates: {
+      readonly __typename: "Coordinates"
+      readonly longitude: number
+      readonly latitude: number
+    }
+  }
+}
```
nicolasburtey, line 30: you can probably import this directly from `generated.ts`?

app/components/map-interface/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export default function MapInterface({
```
nicolasburtey, line 32:
```suggestion
export default function MapComponent({
```

we generally use more the `Component` wording

app/components/map-interface/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                isFocused={focusedMarker?.username === item.username}
+                text={text}
+                styles={styles}
+              />,
+            )
+          }
+
+          return arr
+        },
+        [] as React.ReactElement[],
```
nicolasburtey, line 77: is casting necessary? can't we have a `const empty: React.ReactElement[] = []` instead?

app/components/map-interface/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+      ref={mapViewRef}
+      style={styles.map}
+      showsUserLocation={true}
+      initialRegion={userLocation}
+      customMapStyle={themeMode === "dark" ? MapStyles.dark : MapStyles.light}
+      onPress={handleMapPress}
+    >
+      {(data?.businessMapMarkers ?? []).reduce(
+        (arr: React.ReactElement[], item: MarkerData | null) => {
+          if (item?.username) {
```
nicolasburtey, line 60: I think right now by definition we have a username for the every point on the map - although it might not be reflected in the type?

app/components/map-interface/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <MapView
+      ref={mapViewRef}
+      style={styles.map}
+      showsUserLocation={true}
+      initialRegion={userLocation}
+      customMapStyle={themeMode === "dark" ? MapStyles.dark : MapStyles.light}
+      onPress={handleMapPress}
+    >
+      {(data?.businessMapMarkers ?? []).reduce(
```
nicolasburtey, line 58: `.map` feels more appropriate than `reduce` given the logic

nicolasburtey:
a few comments on the PR

app/components/map-interface/index.tsx
MaxwellDG: The way everything was written before made it very clear that there will some items that were null. I followed suit with that in mind, so there's several instances where I watch out for it. The reduce function instead of map which you mentioned below is an example.

If you're sure they're all going to be populated now then I can certainly remove these kind of checks

app/components/map-interface/index.tsx
MaxwellDG: renamed

app/components/map-interface/index.tsx
MaxwellDG: I dunno - I was always told that was good form. It's to show the arrays got a type from initialisation even when it's empty. I'll remove it.

app/components/map-interface/index.tsx
MaxwellDG: Replaced

MaxwellDG:
Gonna move this to draft until I can get the iOS marker jumping solution fixed as well

app/components/map-interface/index.tsx
nicolasburtey: making a type nullable to non nullable is not a breaking change (the other way around is). let me look into that before we merge the PR

MaxwellDG:
Closing because I'm putting all the map-related PRs into a single PR

PR #2977, opened 2024-01-03
chore: earns-map-screen svg text is now translateable

### LINKED ISSUE #2319: chore: remove hardcoded string in earn
https://github.com/GaloyMoney/galoy-mobile/blob/main/app/screens/earns-map-screen/bottom-start-01.en.svg

contains a string hardcoded in english. there is another version for spanish. but we should make it translatable, or alternatively not show any string in non english/spanish version.

MaxwellDG:
Fixes #2319 

I've completed the task as it was described, but I think it would be far better UI to address this a different way. The text is part of the svg so it has to be positioned with x y coordinates. This gets weird when translations have different widths and lengths of words. I'd much rather rewrite this with the text section of the svgs removed and instead use a standard Text component. That way it can utilize { textAlign: 'center' } no matter the size of the strings.


app/i18n/en/index.ts
```diff
@@ -2036,7 +2036,11 @@ const en: BaseTranslation = {
     registerTitle: "Need to upgrade your account",
     registerContent: "Register with your phone number to receive sats",
     notYet: "Not yet",
-    availableTomorrow: "This section will be available tomorrow!"
+    availableTomorrow: "This section will be available tomorrow!",
+    motivatingBadger1: 'Dig your way',
```
nicolasburtey, line 2040: is there a way to do that: `motivatingBadger: Dig your way\n\nthrough these\n\n...` ?
that would make it easier for the translators to understand the intent. not a big deal if that's not practical

app/screens/earns-map-screen/earns-map-screen.tsx
```diff
@@ -34,11 +32,10 @@ import RightTodo from "./right-section-to-do-01.svg"
...
 import { makeStyles, useTheme } from "@rneui/themed"
-
-const BottomOngoingEN = React.lazy(() => import("./bottom-ongoing-01.en.svg"))
-const BottomOngoingES = React.lazy(() => import("./bottom-ongoing-01.es.svg"))
-const BottomStartEN = React.lazy(() => import("./bottom-start-01.en.svg"))
-const BottomStartES = React.lazy(() => import("./bottom-start-01.es.svg"))
+import BottomOngoing from "@app/components/translateable-svgs/bottom-ongoing"
+import BottomStart from "@app/components/translateable-svgs/bottom-start"
+import BottomOngoingES from "./bottom-ongoing-01.es.svg"
+import BottomStartES from "./bottom-start-01.es.svg"
```
nicolasburtey, line 38: do we still need those 2?

app/i18n/en/index.ts
MaxwellDG: Kinda sorta. I was thinking we could translate it as 1 entity, and then i'd divide it into 3 or 4 lines with a maximum amount of characters per line. I'll do that.

app/screens/earns-map-screen/earns-map-screen.tsx
MaxwellDG: Nope. They're there while I was waiting on as response for the alternative solution. I'll remove them since it seems like you've decided

app/i18n/en/index.ts
MaxwellDG: Nvm. Trying to do calculate all that was ridiculous. I came up with another solution. See below

MaxwellDG:
I removed the text from the svg and cropped the svg to fit its new size. Then I made a standard text component and aligned it to the layout of the rest of the quiz.

nicolasburtey (approved):

PR #2988, opened 2024-01-05
fix: flickering nfc button

UncleSamtoshi:
- fixes flickering nfc button on receive screen

### DIFF
--- a/app/screens/receive-bitcoin-screen/receive-screen.tsx
+++ b/app/screens/receive-bitcoin-screen/receive-screen.tsx
@@ -37,36 +37,32 @@ const ReceiveScreen = () => {
 
   const [displayReceiveNfc, setDisplayReceiveNfc] = useState(false)
 
+  const nfcText = LL.ReceiveScreen.nfc()
   useEffect(() => {
     ;(async () => {
       if (
         request?.type === "Lightning" &&
         request?.state === "Created" &&
         (await nfcManager.isSupported())
-      )
+      ) {
         navigation.setOptions({
           headerRight: () => (
             <TouchableOpacity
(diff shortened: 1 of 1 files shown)


app/screens/receive-bitcoin-screen/receive-screen.tsx
```diff
@@ -37,36 +37,30 @@ const ReceiveScreen = () => {
...
     })()
-  }, [
-    LL.ReceiveScreen,
-    colors.black,
-    navigation,
-    request?.state,
-    request?.type,
-    styles.nfcIcon,
-  ])
+  }, [nfcText, colors.black, navigation, request?.state, request?.type])
```
nicolasburtey, line 63: warning @UncleSamtoshi

nicolasburtey:
what was the issue here? how are we fixing it?

UncleSamtoshi:
The issue was that both `LL.ReceiveScreen` and `styles.nfcIcon` were new objects on every render, so it was causing the useEffect to trigger over and over again. During those triggers I think there were some race conditions where it was causing the button to flicker. It was fixed by relying directly and on the translations string (which doesnt change) and removing `styles.nfcIcon` from the useEffect dependency array which should have no effect because the styles should not change.

app/screens/receive-bitcoin-screen/receive-screen.tsx
UncleSamtoshi: Disabled the warning.

nicolasburtey (approved):

nicolasburtey:
this might be where using `why-did-you-render` would solve those issues or get us warning

Issue #2975, opened 2024-01-02
wait for intent before asking the user his geo localisation


nicolasburtey:
with the recent PR, we are now asking for the geo localization of the user right when we tap on Map. I don't think that is the right behavior, because a user might want to see the map first, because saying "ok great, but who am I again on this map, and is there a merchant nearby that accepts bitcoin?"

so I think we should have a button or otherwise on top of the map to say "center to user location" and this is when we should ask the user for permission. maybe @designsats have some idea on this?

Note: at least this is the behavior on iOS. not too sure if Android is the same.

nicolasburtey:
related, we need to update the config file on iOS:

```
We identified one or more issues with a recent delivery for your app, "Blink (Bitcoin Beach Wallet)" 2.2.212 (657). Your delivery was successful, but you may wish to correct the following issues in your next delivery:

ITMS-90683: Missing purpose string in Info.plist - Your app’s code references one or more APIs that access sensitive user data, or the app has one or more entitlements that permit such access. The Info.plist file for the “Blink.app” bundle should contain a NSLocationAlwaysAndWhenInUseUsageDescription key with a user-facing purpose string explaining clearly and completely why your app needs the data. If you’re using external libraries or SDKs, they may reference APIs that require a purpose string. While your app might not use these APIs, a purpose string is still required. For details, visit: https://developer.apple.com/documentation/uikit/protecting_the_user_s_privacy/requesting_access_to_protected_resources.
```

designsats:
Thats how it currently works on Android. There is a button that centers the map. So something like that for IOS too but with remembering the last position.

MaxwellDG:
> NSLocationAlwaysAndWhenInUseUsageDescription

This is the weird one I was mentioning before. We don't actually use this, and it can be explicitly disabled. But the docs require it nonetheless. Even though the app works without it. And it's the more invasive location request too. I'll see if anyone's gotten around that without having to ask for it.

MaxwellDG:
> with the recent PR, we are now asking for the geo localization of the user right when we tap on Map. I don't think that is the right behavior, because a user might want to see the map first, because saying "ok great, but who am I again on this map, and is there a merchant nearby that accepts bitcoin?"
> 
> so I think we should have a button or otherwise on top of the map to say "center to user location" and this is when we should ask the user for permission. maybe @designsats have some idea on this?
> 
> Note: at least this is the behavior on iOS. not too sure if Android is the same.

[...]

PR #2999, opened 2024-01-10
feat: map upgrades

MaxwellDG:
Apologies for the huge PR, but I found myself rewriting the same pages for a variety of different github issues. Dealing with the repeated merge conflicts would've been equally as frustrating.

This addresses:  #2975 #2955 #2894 and several other features discussed in person that add-on to #2830 

**Initial loading**
- If location permissions given -> initial loading is at user's location
- else if it's user's first time using map -> initial loading at centre of user's country's ip
- else -> initial loading at last viewed coordinates from previous session

**Speed**
- Map loading speed decreased to ~2 seconds from ~8 seconds on Android

**Asking permissions**
- Added `NSLocationAlwaysAndWhenInUseUsageDescription` permission to Info.plist so Apple will shutup. Even though we're not asking for that permission as shown by line 47 in maps-screen.tsx
`Geolocation.setRNConfiguration({
  skipPermissionRequests: true,
  enableBackgroundLocationUpdates: false,
  authorizationLevel: "whenInUse",
  locationProvider: "auto",
})`

- Permissions are no longer asked immediately upon visiting the map
- Added a button in top-right corner that looks like 'center-on user location' button. When pressed, it will ask for user's permission to access location
- If user has previously pressed 'Never ask again', then we can't ask for permissions. We can however open a modal that brings them directly to their system's Blink settings screen where they can update the permission themselves

**iOS bugfix**
- tapping a marker that's close to another marker would make it jump to the wrong one. now if that happens, it'll stay on the correct marker. It'll flash though. Definitely a big improvement, yet not perfect.

**TODO**
- ~~Only part not tested on iOS is the new button for asking permissions~~
- ~~latLng in cache should be an object not 2 separate values. syntax is confusing here~~
- ~~On some Android devices, the 2nd time permissions are asked for (only the 2nd time ever), the modal asking them to change it in their settings will popup after they say no. Will make separate issue for this cause I'm stumped at the moment.~~


ios/Podfile
```diff
@@ -1,14 +1,26 @@
...
+node_require('react-native-permissions/scripts/setup.rb')
 
 use_modular_headers!
 
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
+
+setup_permissions([
+  'LocationAccuracy',
+  'LocationAlways',
```
nicolasburtey, line 20: we only need permission when device is in used, could this be the permission for running in the background?

app/components/map-component/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  handleCalloutPress,
+  onError,
+}: Props) {
+  const {
+    theme: { colors, mode: themeMode },
+  } = useTheme()
+  const styles = useStyles()
+  const { LL } = useI18nContext()
+  const client = useApolloClient()
+  const text = React.useMemo(() => LL.MapScreen.payBusiness(), [LL.MapScreen])
```
nicolasburtey, line 49: why is that necessary? it's not a pattern we have used so far to cache translated text

app/components/map-component/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+      }
+      // Android can ask twice for permission, and initial checks cannot differentiate between BLOCKED vs DENIED
+    } else {
+      // TODO figure out how to not fire this after the 2nd time user blocks
+      toggleModal()
+    }
+  }
+
+  const centerOnUser = () => {
+    getUserRegion(async (region) => {
```
nicolasburtey, line 74: is this an async action?

app/components/map-component/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  data,
+  userLocation,
+  permissionsStatus,
+  setPermissionsStatus,
+  handleMapPress,
+  handleMarkerPress,
+  focusedMarker,
+  focusedMarkerRef,
+  handleCalloutPress,
+  onError,
```
nicolasburtey, line 41: does `onError` needs to be pass as a prop? is it only ever specific to `alertOnLocationError`?

app/components/map-component/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+            e.nativeEvent.coordinate.longitude ===
+            focusedMarker?.mapInfo.coordinates.longitude
+          if (!matchingLat || !matchingLng) {
+            if (focusedMarkerRef.current) {
+              focusedMarkerRef.current.showCallout()
+            }
+          }
+        }}
+      >
+        {(data?.businessMapMarkers ?? []).reduce(
```
nicolasburtey, line 137: this PR will help using a .map instead: https://github.com/GaloyMoney/galoy/pull/3856

nicolasburtey:
on android simulator, this is the first thing I see when I click on the map

image

is it intended?

MaxwellDG:
> on android simulator, this is the first thing I see when I click on the map
> 
> image
> is it intended?

No, that should only be shown when pressing the button in the top right corner. I haven't tested on android simulator. I'll check that out now.

nicolasburtey:
nevermind, it seems my app was not correctly updated.

app/components/map-component/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+    request(LOCATION_PERMISSION)
+      .then((status) => {
+        if (status === RESULTS.GRANTED) {
+          centerOnUser()
+        } else if (status === RESULTS.BLOCKED) {
+          respondToBlocked(status)
+        }
+        setPermissionsStatus(status)
+      })
+      .catch(() => alertOnLocationError())
```
nicolasburtey, line 93: could we conver this to async/await syntax?
we don't really then().catch(), imo it's less readable than the new code syntax

app/graphql/client-only-query.ts
```diff
@@ -41,6 +44,13 @@ export default gql`
     countryCode @client
   }
 
+  query latLng {
+    latLng @client {
+      lat
+      lng
```
nicolasburtey, line 50: I think we should also store the zoom. because otherwise we come back in a very "unzoomed" settings by default, and it feels almost like the settings was not saved as a result of it

app/screens/map-screen/map-screen.tsx
```diff
@@ -95,168 +120,120 @@ export const MapScreen: React.FC<Props> = ({ navigation }) => {
...
-          if (data) {
-            const region: Region = {
-              latitude: data.coords.latitude,
-              longitude: data.coords.longitude,
-              latitudeDelta: 0.02,
-              longitudeDelta: 0.02,
-            }
-            callback(region)
+        setInitialLocation(region)
+        // User is using maps for the first time, so we center on the center of their IP's country
```
nicolasburtey, line 172: is this comment for the else branch below? if that is the case, I would move this line below the else, to be consistent with [comment line 163](https://github.com/GaloyMoney/galoy-mobile/pull/2999/files#diff-2840b3b05c86d0fba4f9261d1ba0aaeca6798d1ad37be559b44495d847710f04R163) where the comment is within the block, not before it

app/screens/map-screen/map-screen.tsx
```diff
@@ -95,168 +120,120 @@ export const MapScreen: React.FC<Props> = ({ navigation }) => {
     toastShow({ message: error.message, LL })
   }
 
-  const maps = data?.businessMapMarkers ?? []
+  // On screen load, check (NOT request) if location permissions are given
+  React.useEffect(() => {
+    check(LOCATION_PERMISSION).then((status) => {
```
nicolasburtey, line 125: async/await syntax preferred

app/screens/map-screen/map-screen.tsx
```diff
@@ -95,168 +120,120 @@ export const MapScreen: React.FC<Props> = ({ navigation }) => {
...
+      console.log("Initial check: ", status)
```
nicolasburtey, line 126:
```suggestion
```

nicolasburtey:
generally the PR looks great. biggest comment is on the zoom not been stored across session. otherwise it's mostly some nits. I have not review every line throughly but I think that is fine - this is not a critical code like the payment flow that require deep attention.

MaxwellDG:
@nicolasburtey Added zoom, addressed the small refactors.

I also found a solution for that Android 2nd request edge case. Found the location where I could add a flag.

nicolasburtey (approved):

PR #3004, opened 2024-01-16
chore: improve deep linking

sandipndev:
With this, if you do `blink://circles` you will get inside the circles screen directly.
This pattern can also be used to [useLinkTo](https://reactnavigation.org/docs/use-link-to) any screen with push notifications.


sandipndev:
@UncleSamtoshi would you prefer to have the useLinkTo pattern in this PR or a separate one?

UncleSamtoshi:
This looks good! I think we should expand it so that we can link to other places in the app such as a transaction and the earn, map, and receive screen.

nicolasburtey (approved):

PR #3036, opened 2024-01-31
chore: improving messaging on next available session

nicolasburtey:
it seems with the prior message it was not very clear that only section was available per day


app/i18n/en/index.ts
agbegin, line 1: I would say "Come back tomorrow to continue learning about Bitcoin!"

pretyflaco: "Yay, it looks like you're falling down the rabbit hole with high speed. Time for a digestive break. Come back tomorrow to continue learning about Bitcoin!"

Issue #611, opened 2022-09-05
Localize info.plist file


daviroo:
The info.plist file is used to store strings which give users reasons for why we need certain permissions e.g. to access photos in their camera roll.  Right now these reasons are given in English but we should localize to Spanish.  Here is some documentation on how to achieve it https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/AboutInformationPropertyListFiles.html.

PR #3041, opened 2024-02-01
feat: plist localizations

MaxwellDG:
Addresses #611 

This is a draft, but if its approved that we want to do this, Ill write a script to make the addition of new languages automated. At the moment I've only written 1 line in the il8n README, but Ill expand that to include the aforementioned script.

Also take note that I made 2 small changes to info.plist. We had a duplicate in there, and I removed a variable. Doesn't look like xCode localization .string files permit system variables like that.


nicolasburtey:
I think this PR is great and very well needed

ideally we would want to add those files somehow into transifex which is where people who are helping doing translation are going, and having everything centralized to transifex makes it much easier. 

https://app.transifex.com/galoy/bitcoin-beach-wallet/

but now the challenge for integrating transifex is the upload/download from transifix and format management. I'm not sure if transifex would be working well with the `InfoPlist.strings` format. maybe worth researching. 

I could see how this could be done manually because it's only a few strings, but if we add a string into this list, or add a new language, then it become a higher burden to maintain.

nicolasburtey (commented):
.

grimen:
Closing as part of a stale-PR cleanup pass (no activity in 6+ months). Feel free to reopen if this is still active work.

PR #3043, opened 2024-02-05
fix: webview back button, webview dark-mode on android

siddhart1o1:
+ Fixed: Change WebView theme to match the app's theme. Previously, if the system theme was set to light but the app's theme was dark, the WebView theme would remain light. With this PR, the WebView theme will align with the app's theme.

image 
 --- to --->

image

+ Fixed: Back button for webview.

image

### DIFF
--- a/app/screens/webview/webview.tsx
+++ b/app/screens/webview/webview.tsx
@@ -1,13 +1,14 @@
 import { StackNavigationProp } from "@react-navigation/stack"
 import * as React from "react"
-import { Alert, Button } from "react-native"
+import { Alert, TouchableOpacity } from "react-native"
 import { injectJs, onMessageHandler } from "react-native-webln"
 import { WebView, WebViewNavigation } from "react-native-webview"
 import { Screen } from "../../components/screen"
 import { RootStackParamList } from "../../navigation/stack-param-lists"
 import { RouteProp, useNavigation } from "@react-navigation/native"
 import { useI18nContext } from "@app/i18n/i18n-react"
-import { makeStyles } from "@rneui/themed"
+import { makeStyles, useTheme } from "@rneui/themed"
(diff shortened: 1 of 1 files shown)


app/screens/webview/webview.tsx
```diff
@@ -45,17 +50,36 @@ export const WebViewScreen: React.FC<Props> = ({ route }) => {
   React.useEffect(() => {
     navigation.setOptions({
       headerLeft: () => (
-        //                                  FIXME < is not the same as for other screens
-        <Button onPress={handleBackPress} title={`< ${LL.common.back()}`} />
+        <TouchableOpacity style={styles.iconContainer} onPress={handleBackPress}>
+          <GaloyIcon
+            name="caret-left"
+            size={20}
+            color={mode === "light" ? colors.black : colors._white}
```
nicolasburtey, line 57:
```suggestion
            color={colors.black}
```

the `_white` is a hardcoded color to while, while the `black` is a color that would be different depending on the theme. it would be black for the light mode, but white for the dark mode. 

I think just having `colors.black` would just work the same, with the benefits of been clearer on the intent.

app/screens/webview/webview.tsx
```diff
@@ -45,17 +50,32 @@ export const WebViewScreen: React.FC<Props> = ({ route }) => {
...
       ),
     })
-  }, [navigation, handleBackPress, LL])
+  }, [
+    navigation,
+    handleBackPress,
+    LL,
+    mode,
+    colors.black,
+    colors._white,
```
nicolasburtey, line 64:
```suggestion
```

I guess this item is not longer needed

nicolasburtey: maybe mode is not either

nicolasburtey (approved):

PR #3047, opened 2024-02-06
fix: linking logic

UncleSamtoshi:
- prevent unhandled links
- fix flickering of themes
- pass payment into sendDestinationScreen
- fix stuck in circles bug


app/screens/webview/webview.tsx
```diff
@@ -86,7 +86,6 @@ export const WebViewScreen: React.FC<Props> = ({ route }) => {
           sendPayment: async (paymentRequestStr) => {
             navigate("sendBitcoinDestination", {
               payment: paymentRequestStr,
-              autoValidate: true,
```
nicolasburtey, line 89: Why is that been removed?

nicolasburtey (commented):
One question

app/screens/webview/webview.tsx
nicolasburtey: Were using it for blink fiat to skip showing the invoice this is confusing for the user

UncleSamtoshi: It was breaking the linking into the mobile application because autoValidate was not present so it was not receiving any invoice at all. Where is autoValidate supposed to be false? I only see it as true.

app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -369,7 +369,7 @@ const SendBitcoinDestinationScreen: React.FC<Props> = ({ route }) => {
   )
 
   useEffect(() => {
-    if (route.params?.autoValidate && route.params.payment) {
+    if (route.params.payment) {
```
UncleSamtoshi, line 372: @nicolasburtey right here autoValidate is defaulting to false for the linking and so it is not receiving the payment param.

app/screens/webview/webview.tsx
nicolasburtey: default = undefined = false, so for every use case where it's not explicitly true

UncleSamtoshi: Alright let me adjust.

UncleSamtoshi: The way the logic is set up currently, however, means that if autoValidate isn't present, no invoice is being passed in at all.

UncleSamtoshi: > Were using it for blink fiat to skip showing the invoice this is confusing for the user

Actually the destination is now present on the confirmation screen. Do you still want to show it on the destination screen as well?

nicolasburtey (approved):

PR #3079, opened 2024-02-17
test(e2e): detox e2e testing setup against local machine

sandipndev:
- Introduce Detox (https://wix.github.io/Detox/)
- Testing against Local Instance
- Deprecating use of CircleCI for PR based testing because they are very flaky

The dev setup keeps on generating regtest blocks so as to simulate time progress with the Galoy Backend.


UncleSamtoshi:
Should E2E ios and E2E android be separate jobs so that they can run in parallel, and so that they both receive fresh environments?

dev/bin/setup.sh
```diff
@@ -0,0 +1,25 @@
...
+ source ${REPO_ROOT}/dev/vendor/galoy-quickstart/bin/helpers.sh
+ source ${REPO_ROOT}/dev/.env
+
+ login_user "alice" "$ALICE_PHONE" "000000"
+ echo "alice logged in"
+ ALICE_TOKEN=$(read_value "alice")
+ receive_onchain
+ echo "alice funded"
+
+ login_user "alice" "$BOB_PHONE" "000000"
```
UncleSamtoshi, line 16: should this be "bob"

dev/bin/setup.sh
```diff
@@ -0,0 +1,25 @@
...
+ echo "bob logged in"
+ BOB_TOKEN=$(read_value "alice")
```
UncleSamtoshi, line 18: Should this be reading "bob"

dev/bin/setup.sh
sandipndev: The next line, receive_onchain declared in quickstart adds money to token "alice"'s wallet. It doesn't take a parameter unfortunately. Once it gets fixed in upstream, it could be incorporated here.

dev/bin/setup.sh
sandipndev: We need bob's token in "alice" variable for receive_onchain, same as other comment.

sandipndev:
> Should E2E ios and E2E android be separate jobs so that they can run in parallel, and so that they both receive fresh environments?

We currently have one self hosted mac runner. Running them in parallel would require us to host another one = money. And also, IMO there is no quantifiable value in knowing that our tests passed in ios but failed in android. If something has failed, one would look at the video recordings from the run.

UncleSamtoshi:
> > Should E2E ios and E2E android be separate jobs so that they can run in parallel, and so that they both receive fresh environments?
> 
> We currently have one self hosted mac runner. Running them in parallel would require us to host another one = money. And also, IMO there is no quantifiable value in knowing that our tests passed in ios but failed in android. If something has failed, one would look at the video recordings from the run. Or do you have other opinion @UncleSamtoshi?

agreed that if either test fails you'll still want to look at the recording. I think the value comes from tests finishing faster + fresh state for each galoy stack. However we can do the state refresh serially if its necessary.

UncleSamtoshi (approved):

PR #3124, opened 2024-03-20
feat: add price and marketing settings

UncleSamtoshi:


app/i18n/en/index.ts
```diff
@@ -2336,6 +2336,14 @@ const en: BaseTranslation = {
       Payments: {
         title: "Payments",
         description: "Notifications related to sending and receiving payments.",
+      },
+      Marketing: {
+        title: "Promotions and updates",
```
nicolasburtey, line 2341: not a big fan of `Promotions`; for me this is a trigger word that if I see that, I'll turn it off. I prefer `new features and general updates` of something like that

UncleSamtoshi: Changed it

nicolasburtey (approved):

PR #3136, opened 2024-03-26
ci: try both e2e tests in one go

sandipndev:


nicolasburtey:
I think ideally they should not be in the same "checks". they should be 2 different "checks". but they should run in parrallel

sandipndev:
@nicolasburtey for that, we would need two different scaleway workers. I'll do some research if there is a way to do this in one single scaleway droplet so as to not increase expenses.

sandipndev:
I think I have got a good way to do it in the same worker. Will implement in a separate PR.

PR #3145, opened 2024-03-29
feat: Add link opening functionality in ScanningQRCodeScreen

### LINKED ISSUE #3128: Offer option to open link when scanning QR that encodes URL
QRs are not human-readable, so sometimes I scan a QR with Blink thinking it is a Lightning Invoice or Bitcoin Address, when in fact the QR encodes a URL. Currently, Blink decodes the URL and tells me that it is "Invalid QR Code" and in addition displays the URL it found, noting that it is not a valid Bitcoin address or Lightning Invoice. This is better than just returning an error, but not as helpful as it could be.

After scanning a QR that encodes a URL, show the URL and offer user the option to open the link.
Breez does it this way and it's good UX. https://x.com/_pretyflaco/status/1771089232641057254?s=20

Blink
image

Breez
image

Prakhar-Agarwal-byte:
Fixes #3128

Show an alert to open a URL when QR encodes a URL instead of showing an invalid address error.

### DIFF
--- a/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
+++ b/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
@@ -161,8 +161,8 @@ export const ScanningQRCodeScreen: React.FC = () => {
               ],
             )
           : Alert.alert(
-              "Open Link",
-              `${data.toString()}\n\nAre you sure you want to open this link?`,
+              LL.ScanningQRCodeScreen.openLinkTitle(),
+              `${data.toString()}\n\n${LL.ScanningQRCodeScreen.confirmOpenLink()}`,
               [
                 {
                   text: LL.common.No(),
--- a/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
+++ b/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
@@ -104,7 +104,19 @@ export const ScanningQRCodeScreen: React.FC = () => {
   }, [hasPermission, requestPermission])
 
   const loadInBrowser = (url: string) => {
-    Linking.openURL(url).catch((err) => console.error("Couldn't load page", err))
+    Linking.openURL(url).catch((err) => Alert.alert(err.toString()))
+  }
+
+  function isValidHttpUrl(input: string) {
+    let url
+
+    try {
+      url = new URL(input)
+    } catch (_) {
+      return false
+    }
+
+    return url.protocol === "http:" || url.protocol === "https:"
   }
 
   const processInvoice = React.useMemo(() => {
@@ -147,8 +159,9 @@ export const ScanningQRCodeScreen: React.FC = () => {
           })
           return
         }
-        destination.invalidReason === "InvoiceExpired"
-          ? Alert.alert(
+        switch (destination.invalidReason) {
+          case "InvoiceExpired":
+            Alert.alert(
               LL.ScanningQRCodeScreen.invalidTitle(),
               LL.ScanningQRCodeScreen.expiredContent({
                 found: data.toString(),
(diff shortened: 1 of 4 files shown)


app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -143,23 +147,36 @@ export const ScanningQRCodeScreen: React.FC = () => {
...
+              [
+                {
+                  text: LL.common.ok(),
+                  onPress: () => setPending(false),
+                },
+              ],
+            )
+          : Alert.alert(
+              "Open Link",
+              `${data.toString()}\n\nAre you sure you want to open this link?`,
```
nicolasburtey, line 165: our app is meant to be translatable, so we don't hard code string. you can have a look at `app/i18n/en/index.ts` and also `make codegen` to be able to update our translations with the new string

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -143,23 +147,36 @@ export const ScanningQRCodeScreen: React.FC = () => {
...
         }
-
-        Alert.alert(
-          LL.ScanningQRCodeScreen.invalidTitle(),
-          destination.invalidReason === "InvoiceExpired"
-            ? LL.ScanningQRCodeScreen.expiredContent({
-                found: data.toString(),
-              })
-            : LL.ScanningQRCodeScreen.invalidContent({
+        destination.invalidReason === "InvoiceExpired"
```
nicolasburtey, line 150: if the assumption is that, if an invoice is expired then it contains a loadable link, I think this is a wrong assumption.

a destination can be invalid for a couple more reason. I think there should be a specific invalidReason that is something like "isLink" (which would means it's not a proper payment destination), and IIF this is the reason, then we suggest to the user to open the link from the browser

Prakhar-Agarwal-byte: If `paymentType` of the `parsedDestination` from QR does not match any, then `invalidReason` defaults to `UnknownDestination`.
https://github.com/GaloyMoney/galoy-mobile/blob/331d9552448bd22038131077cacbc8baddc7d3f6/app/screens/send-bitcoin-screen/payment-destination/index.ts#L63-L68

I can put a check for `UnknownDestination`, but then also QR can contain text other than URL. I can check once again that `data.toString()` is indeed a URL by using this method https://stackoverflow.com/a/43467144

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
Prakhar-Agarwal-byte: Added "en" version of these strings

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
nicolasburtey: Sounds like a good option

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
```diff
@@ -103,6 +103,10 @@ export const ScanningQRCodeScreen: React.FC = () => {
     }
   }, [hasPermission, requestPermission])
 
+  const loadInBrowser = (url: string) => {
+    Linking.openURL(url).catch((err) => console.error("Couldn't load page", err))
```
nicolasburtey, line 107: Should probably be showing an error message to the user here instead of using the console

Prakhar-Agarwal-byte: Put in an alert

app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx
Prakhar-Agarwal-byte: Committed

nicolasburtey (approved):

nicolasburtey:
thanks for working on this PR!

PR #3194, opened 2024-04-18
chore: testing if this directive is still necessary

nicolasburtey:


grimen:
Not relevant anymore - already concluded.

PR #3198, opened 2024-04-18
feat: improve deep linking

UncleSamtoshi:
- add more deeplinks
- enable deeplinks to open "set account", "set ln address", and "upgrade account" modals.


app/components/set-default-account-modal/set-default-account-modal.tsx
```diff
@@ -51,10 +48,9 @@ export const SetDefaultAccountModal = ({
   const [usdLoading, setUsdLoading] = React.useState(false)
 
   const [accountUpdateDefaultWallet] = useAccountUpdateDefaultWalletIdMutation()
-  const navigation = useNavigation<StackNavigationProp<RootStackParamList, "Primary">>()
 
   const { data } = useSetDefaultAccountModalQuery({
-    fetchPolicy: "cache-only",
+    fetchPolicy: "cache-first",
```
nicolasburtey, line 53: skimming though the code, I think this is adding a network request for every time the app launches. not ideal

UncleSamtoshi: Yeah i thought there might be a chance that it doesn't exist in the cache, but I actually think that wont happen or will at least be rare. I will switch it back.

nicolasburtey (approved):

PR #3209, opened 2024-04-23
feat: notifications history

sandipndev:

### DIFF
--- a/app/screens/notification-history-screen/notification.tsx
+++ b/app/screens/notification-history-screen/notification.tsx
@@ -1,0 +1,78 @@
+import {
+  StatefulNotification,
+  StatefulNotificationsDocument,
+  useStatefulNotificationAcknowledgeMutation,
+} from "@app/graphql/generated"
+import { Text, makeStyles } from "@rneui/themed"
+import { View } from "react-native"
+import { timeAgo } from "./utils"
+import { gql } from "@apollo/client"
+import { TouchableWithoutFeedback } from "react-native-gesture-handler"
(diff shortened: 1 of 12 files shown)


app/screens/notification-history-screen/notification.tsx
```diff
@@ -0,0 +1,78 @@
...
+  })
+
+  const linkTo = useLinkTo()
+
+  return (
+    <TouchableWithoutFeedback
+      onPress={() => {
+        setIsAcknowledged(true)
+        !isAcknowledged && ack()
+        deepLink && linkTo(deepLink)
```
UncleSamtoshi, line 49: Do you think we should have a separate button for following the link. It could be confusing if you are just trying to mark a notification as read and all of a sudden you are linked to a different page.

UncleSamtoshi:
I think we might want to put a notifications icon link on the home page or make it one of the bottom tabs. I think it may be too hidden in the settings menu. Could be nice to have some indicator that there were unread notifications as well.

nicolasburtey:
I'm not sure mocking with transaction is the best way to get a feel for this component assuming we don't want transaction data in the notification history?

app/screens/notification-history-screen/notification.tsx
sandipndev: I was thinking about this too but most other apps "react" on tap of notifications and they don't have a different button for that. Confused about this UX.

sandipndev:
Merging this in order to see notifications in app so that we can reiterate on UI

PR #3214, opened 2024-04-24
Update use-app-config.ts

pulkit-999:
No need of if condition which is used in the code, it will be handled internally i think.


nicolasburtey:
> [pulkit-999](/pulkit-999)

hey pulkit-999, thanks for making this PR!

first thing is we use conventional commit in our repos: https://www.conventionalcommits.org/en/v1.0.0/

I can have a deeper look when the tests are green

grimen:
No activity in long time and missing tests - closing.

PR #3218, opened 2024-04-24
feat: adding explicit accept TC steps in the onboarding flow

nicolasburtey:
adding this steps in the onboarding flow:

image


app/screens/accept-t-and-c/accept-t-and-c.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const action = () => {
+    if (flow === "phone") {
+      navigation.navigate("phoneFlow", {
+        screen: "phoneLoginInitiate",
+        params: {
+          type: PhoneLoginInitiateType.CreateAccount,
+        },
+      })
+    } else {
```
UncleSamtoshi, line 42: What about the trial flow? Also how do you leave the screen if the action just opens the modal?

app/i18n/en/index.ts
```diff
@@ -28,6 +28,13 @@ const en: BaseTranslation = {
     howToUseYourCashRegisterExplainer:
       "Allow people to collect payments via the Cash Register link, without accessing your wallet.\n\nThey can create invoices and payments will be sent directly to your {bankName: string} Wallet.",
   },
+  AcceptTermsAndConditionsScreen: {
+    title: "Terms and Conditions",
+    accept: "Accept",
+    termsAndConditions: "View terms and Conditions",
```
UncleSamtoshi, line 34: Capitalization needs to be fixed on line 34.

app/i18n/en/index.ts
```diff
@@ -28,6 +28,13 @@ const en: BaseTranslation = {
...
+    prohibitedCountry: "View prohibited countries",
+    text: "By clicking 'Accept', you agree to our Terms And Conditions. You also agree that you are not a resident or citizen from one of the prohibited countries",
```
UncleSamtoshi, line 36: Needs a period at the end.

app/screens/accept-t-and-c/accept-t-and-c.tsx
nicolasburtey: does this address #1 ? https://github.com/GaloyMoney/galoy-mobile/pull/3218/commits/2fd8b3a08c0aedcf6a8f1458613b7d09fbc8e3a4

on the second point, there is not much logic changes, I just moved the logic from getStartedScreen to here, because the logic is within a modal. but the action to move to other screen were already in the modal.

UncleSamtoshi: Nevermind, I was confused by the naming. I thought the confirmation modal was to confirm the T and C not confirm device account. I would probably have the fallback be navigate to the home screen instead of alerting.

UncleSamtoshi: Maybe out of scope for this PR but we probably want some audit trail that terms and conditions were accepted.

UncleSamtoshi (approved):

PR #3240, opened 2024-05-01
feat: disable slide gesture around confirm slider button

### LINKED ISSUE #3081: Scroll to right for confirming payment
If we're not selecting the button accurately but just above, then the action, at least on iOS, is to come back to the previous screen.

This can be quite confusing

This action of been able to come back to the previous screen should be prevented at least in the area surrounding the sliding cursor

Prakhar-Agarwal-byte:
Fixes #3081 

Added a `PanGestureHandler` which captures slide gestures in areas surrounding the confirm slider button. This prevents the native navigation handler from catching this event, and back action is not triggered.

**Demo:**

https://github.com/GaloyMoney/blink-mobile/assets/17739006/91ae2854-9160-4dcf-b6a2-879e176cba15

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from "react"
 import { ActivityIndicator, TouchableOpacity, View } from "react-native"
+import { PanGestureHandler } from "react-native-gesture-handler"
 import ReactNativeHapticFeedback from "react-native-haptic-feedback"
 
 import { gql } from "@apollo/client"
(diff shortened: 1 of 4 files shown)


nicolasburtey:
great PR thanks!

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -560,4 +567,7 @@ const useStyles = makeStyles(({ colors }) => ({
     justifyContent: "space-between",
     alignItems: "center",
   },
+  sliderContainer: {
+    padding: 20,
```
nicolasburtey, line 571: should we make it even a bit bigger?

```suggestion
    padding: 32,
```

nicolasburtey: merging for now. we can change in a future PR later.

Prakhar-Agarwal-byte: Yes, this may help.

PR #3251, opened 2024-05-08
feat: bulletin

sandipndev:


app/components/notifications/bulletins.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+import { useNotifications } from "."
+import { NotificationCardUI } from "./notification-card-ui"
+import {
+  StatefulNotificationsDocument,
+  UnacknowledgedNotificationCountDocument,
+  useBulletinsQuery,
+  useStatefulNotificationAcknowledgeMutation,
+} from "@app/graphql/generated"
+import { useLinkTo } from "@react-navigation/native"
```
UncleSamtoshi, line 11: We need to switch to using `Linking.openUrl` instead of useLinkTo.

PR #3271, opened 2024-05-16
feat: reactive price screen

UncleSamtoshi:
- allow hovering over prices
- use price of 1 btc


UncleSamtoshi:
image

nicolasburtey:
We should probably stack button for timeframe at the bottom of the screen

Would be good to have maybe the vertical axis back

nicolasburtey (approved):

UncleSamtoshi:
I think now that you can hover over prices the vertical axis may not be neccessary. It also keeps the chart symmetrical by not having them. As for the buttons we could move them down, but the but the buy bitcoin button will be below it for some users and personally think its looks better with them right next to the chart. Willing to change it if you feel strongly though!

PR #3272, opened 2024-05-17
feat: store multiple tokens

Prakhar-Agarwal-byte:
Token management for #3273 

- Update PersistantState schema to have a list of saved tokens for multiple accounts
- Create migrations for the same

### DIFF
--- a/app/hooks/use-app-config.ts
+++ b/app/hooks/use-app-config.ts
@@ -10,7 +10,7 @@ export const useAppConfig = () => {
     () => ({
       token: persistentState.galoyAuthToken,
       galoyInstance: resolveGaloyInstanceOrDefault(persistentState.galoyInstance),
-      savedAccounts: persistentState.galoyAllAuthTokens,
+      allTokens: persistentState.galoyAllAuthTokens,
     }),
     [
       persistentState.galoyAuthToken,
--- a/app/hooks/use-app-config.ts
+++ b/app/hooks/use-app-config.ts
@@ -10,12 +10,12 @@ export const useAppConfig = () => {
     () => ({
       token: persistentState.galoyAuthToken,
       galoyInstance: resolveGaloyInstanceOrDefault(persistentState.galoyInstance),
-      savedAccounts: persistentState.galoySavedAccounts,
+      savedAccounts: persistentState.galoyAllAuthTokens,
     }),
     [
       persistentState.galoyAuthToken,
       persistentState.galoyInstance,
-      persistentState.galoySavedAccounts,
+      persistentState.galoyAllAuthTokens,
     ],
   )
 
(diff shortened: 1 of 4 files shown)


sandipndev:
@Prakhar-Agarwal-byte unit tests are failing, it expects the new schema. Can you fix that?

app/store/persistent-state/state-migrations.ts
```diff
@@ -30,9 +30,24 @@ type PersistentState_6 = {
   galoyAuthToken: string
 }
 
-const migrate6ToCurrent = (state: PersistentState_6): Promise<PersistentState> =>
+type PersistentState_7 = {
+  schemaVersion: 7
+  galoyInstance: GaloyInstanceInput
+  galoyAuthToken: string
+  galoySavedAccounts: string[]
```
sandipndev, line 37: Perhaps we can be a little more explicit?

```suggestion
  galoyAuthToken: string
  galoyOtherAuthTokens: string[]
```

app/hooks/use-app-config.ts
```diff
@@ -35,6 +40,7 @@ export const useAppConfig = () => {
           return {
             ...state,
             galoyAuthToken: token,
+            galoySavedAccounts: [...state.galoySavedAccounts, token],
```
sandipndev, line 43: Should also update the `saveTokenAndInstance` function.

sandipndev (commented):
Good initial draft, I'll need to review this PR locally once after the comments are addressed and the tests are passing before approval.

app/store/persistent-state/state-migrations.ts
Prakhar-Agarwal-byte: `galoySavedAccounts` will contain the current token as well so maybe we can rename it to `galoyAllAuthTokens` instead?

app/hooks/use-app-config.ts
Prakhar-Agarwal-byte: Sure

nicolasburtey:
I think this PR should also be merged first: https://github.com/GaloyMoney/blink-mobile/pull/3122

Before we think of multiple token, we need to make storing tokens safer

app/store/persistent-state/state-migrations.ts
```diff
@@ -30,9 +30,24 @@ type PersistentState_6 = {
...
+}
+
+const migrate7ToCurrent = (state: PersistentState_7): Promise<PersistentState> =>
   Promise.resolve(state)
 
+const migrate6ToCurrent = (state: PersistentState_6): Promise<PersistentState> => {
+  return migrate7ToCurrent({
+    ...state,
+    schemaVersion: 7,
+    galoyAllAuthTokens: [],
```
sandipndev, line 47: Shouldn't this check if the user has a token preset and add it in the All array?

Prakhar-Agarwal-byte: Yes it should be present in the all tokens list

sandipndev:
I will do a thorough testing on logged in phones / non logged in phones and verify the user is not logged out before merging.

sandipndev:
Verified and merged!

PR #3346, opened 2024-12-03
feat: add Slovak translation

webiumsk:


openoms (dismissed):
Thank you for the PR,
it is correct as compared to this PR adding an other language previously: https://github.com/GaloyMoney/blink-mobile/pull/2978/files

The json with the translated text was already added here: https://github.com/GaloyMoney/blink-mobile/pull/3344

dolcalmi (approved):

PR #3349, opened 2024-12-05
fix: always show full address for pos and static qr in settings

openoms:
Fixing the issue where the `LnAdressHostname` is displayed in place of the `posUrl`

The correct address for the Pos and StaticQR should  start with `pay.blink.sv`:

IMG_20241205_150326

### DIFF
--- a/app/screens/settings-screen/settings/account-pos.tsx
+++ b/app/screens/settings-screen/settings/account-pos.tsx
@@ -10,7 +10,6 @@ import { SettingsRow } from "../row"
 export const AccountPOS: React.FC = () => {
   const { appConfig } = useAppConfig()
   const posUrl = appConfig.galoyInstance.posUrl
-  const shortUrl = appConfig.galoyInstance.lnAddressHostname
 
   const { LL } = useI18nContext()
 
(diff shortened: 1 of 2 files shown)


app/screens/settings-screen/settings/account-pos.tsx
```diff
@@ -10,21 +10,19 @@ import { SettingsRow } from "../row"
...
 
   const pos = `${posUrl}/${data.me.username}`
-  const short = `${shortUrl}/${data.me.username}`
 
   return (
     <SettingsRow
       loading={loading}
       title={LL.SettingsScreen.pos()}
-      subtitle={short}
+      subtitle={posUrl}
```
dolcalmi, line 25: should not use the same value used in the action

app/screens/settings-screen/settings/account-static-qr.tsx
```diff
@@ -10,21 +10,19 @@ import { SettingsRow } from "../row"
...
 
   const qrUrl = `${posUrl}/${data.me.username}/print`
-  const short = `${shortUrl}/${data.me.username}/print`
 
   return (
     <SettingsRow
       loading={loading}
       title={LL.SettingsScreen.staticQr()}
-      subtitle={short}
+      subtitle={posUrl}
```
dolcalmi, line 25: same here

dolcalmi (changes requested):

app/screens/settings-screen/settings/account-pos.tsx
openoms: Now matching the correct url used when opening the link.

app/screens/settings-screen/settings/account-static-qr.tsx
openoms: Now matching the correct url used when opening the link.

dolcalmi (approved):

openoms:
The fix is working in the emulator:
image
Merging.

PR #3362, opened 2025-02-18
Feature:  handle lnurl success action

tiankii:
**Implemented: LNURL successAction: LUD-09, LUD-10**

image


app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx
```diff
@@ -114,6 +116,52 @@ const SendBitcoinCompletedScreen: React.FC<Props> = ({ route }) => {
...
+
+  const SuccessActionComponent = () => {
+    if (!successAction) return null
+
+    const originalText = successAction.message ?? ""
+    const truncatedText =
+      originalText.slice(0, 300) + (originalText.length > 300 ? "..." : "")
+
+    switch (successAction.tag) {
+      case "message":
```
dolcalmi, line 136: it must have 3 cases where are url and aes? please add them with proper testing. it can be an util or a component (please check __tests__ so you can get an idea on how to implement it)

app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx
```diff
@@ -114,6 +116,52 @@ const SendBitcoinCompletedScreen: React.FC<Props> = ({ route }) => {
...
+      LL,
+    })
+  }
+
+  const SuccessActionComponent = () => {
+    if (!successAction) return null
+
+    const originalText = successAction.message ?? ""
+    const truncatedText =
+      originalText.slice(0, 300) + (originalText.length > 300 ? "..." : "")
```
dolcalmi, line 133: according to [lud09](https://github.com/lnurl/luds/blob/luds/09.md) the limit is 144 chars

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx
```diff
@@ -154,6 +154,7 @@ const SendBitcoinConfirmationScreen: React.FC<Props> = ({ route }) => {
               params: {
                 arrivalAtMempoolEstimate: extraInfo?.arrivalAtMempoolEstimate,
                 status,
+                successAction: paymentDetail?.successAction || null,
```
dolcalmi, line 157: this does not makes sense.. if there are not payment details or successAction then it would be `undefinied || null`

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -381,6 +381,15 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
           }
 
           const result = await requestInvoice(requestInvoiceParams)
+
+          setPaymentDetail((prev) => {
```
dolcalmi, line 385: is this being used? why prev?

app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx
```diff
@@ -381,6 +381,15 @@ const SendBitcoinDetailsScreen: React.FC<Props> = ({ route }) => {
...
+            if (!prev || !prev.setSuccessAction) {
+              return prev
+            }
+
+            return prev.setSuccessAction(result.successAction)
+          })
```
dolcalmi, line 391: early return is good but try to avoid negated if statements when is not needed (i.e try to follow "positive condition principle")

```suggestion
          setPaymentDetail((prev) => {
            if (prev?.setSuccessAction) {
              return prev.setSuccessAction(result.successAction)
            }
 
            return prev
          })
```

dolcalmi (changes requested):
also please add a screenshot

__tests__/screens/send-bitcoin-completed-screen.spec.tsx
dolcalmi, line 1: you are missing aes test... you can use this data
```js
    successAction: {
      tag: "aes",
      description: "Here is your redeem code",
      ciphertext: "564u3BEMRefWUV1098gJ5w==",
      iv: "IhkC5ktKB9LG91FvlbN2kg==",
    },
    preimage: "25004cd52960a3bac983e3f95c432341a7052cef37b9253b0b0b1256d754559b",
```

app/screens/send-bitcoin-screen/payment-details/lightning.ts
```diff
@@ -250,10 +254,17 @@ export const createAmountLightningPaymentDetails = <T extends WalletCurrency>(
...
       },
     })
+    const { settlementVia } = data?.lnInvoicePaymentSend.transaction || {}
 
     return {
       status: data?.lnInvoicePaymentSend.status,
       errors: data?.lnInvoicePaymentSend.errors,
+      extraInfo: {
+        preimage:
+          settlementVia?.__typename === "SettlementViaLn"
```
dolcalmi, line 264: do we need the type validation or is this enough `preimage: settlementVia?.preImage` ?

app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.stories.tsx
```diff
@@ -54,6 +58,8 @@ const pendingRoute = {
   params: {
     status: "PENDING",
     arrivalAtMempoolEstimate: undefined,
+    successAction: undefined,
+    preimage: undefined,
```
dolcalmi, line 62: if they are undefined then there is no need to set them here

app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.stories.tsx
```diff
@@ -37,6 +39,8 @@ const queuedRoute = {
   params: {
     status: "PENDING",
     arrivalAtMempoolEstimate: 10000,
+    successAction: undefined,
+    preimage: undefined,
```
dolcalmi, line 43: same here

dolcalmi (changes requested):
dont forget to disable auto redirect to home when it has success action

app/screens/send-bitcoin-screen/payment-details/lightning.ts
EsauGomez9: Yes, it is necessary because that was the reason why the previous check of the pull request failed. If I don't evaluate the type, TypeScript throws this error:

```
Property 'preImage' does not exist on type '{ readonly __typename: "SettlementViaIntraLedger"; } | { readonly __typename: "SettlementViaLn"; readonly preImage?: string | null | undefined; } | { readonly __typename: "SettlementViaOnChain"; }'.
  Property 'preImage' does not exist on type '{ readonly __typename: "SettlementViaIntraLedger"; }'.
```

This happens because, according to the types generated from the GraphQL schema, only "SettlementViaLn" contains preImage, while the others ("SettlementViaIntraLedger" | "SettlementViaOnChain") do not.

__tests__/screens/send-bitcoin-completed-screen.spec.tsx
EsauGomez9: Great, thanks!

tiankii:
@dolcalmi  I have already resolved the requested changes. Some I couldn't comment on because they no longer apply, but they have been addressed. Could you please review them again?

yanascz:
How does it work with a multiline success message? I mean a message that contains line feeds, e.g.:

```
Tada!
• one
• two
• three
```

Can be tested e.g. with:
QR code

esaugomez31:
> How does it work with a multiline success message? I mean a message that contains line feeds, e.g.:
> 
> ```
> Tada!
> • one
> • two
> • three
> ```
> 
> Can be tested e.g. with: QR code

Hi, I did the test you recommended, and this is how it looks:
imagen

app/screens/send-bitcoin-screen/use-send-payment.ts
```diff
@@ -68,6 +68,13 @@ gql`
         message
       }
       status
+      transaction {
+        settlementVia {
+          ... on SettlementViaLn {
```
dolcalmi, line 73: please add
```
... on SettlementViaIntraLedger {
          preImage
        }
```
this will fix ln txs settled intraledger (just in case a provider uses blink as backend)

app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx
```diff
@@ -198,6 +213,70 @@ const useStyles = makeStyles(({ colors }) => ({
     padding: 20,
     borderRadius: 20,
   },
+  truncatedText: {
```
dolcalmi, line 216: seems the styles in this files are duplicated in other componentes... please remove them if they dont belong to this file

app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.stories.tsx
```diff
@@ -63,3 +65,62 @@ const route = {
 } as const
 
 export const Intraledger = () => <SendBitcoinConfirmationScreen route={route} />
+
+const successActionMessageMock = {
```
dolcalmi, line 69: please move all this test code to the test file.. like with the other components

dolcalmi (changes requested):

dolcalmi (approved):

EsauGomez9 (approved):

PR #3393, opened 2025-03-05
chore: update gemfiles

dolcalmi:


grimen:
Already addressed in recent React Native upgrade.

PR #3396, opened 2025-03-05
feat: onchain fee settings multiple queues

tiankii:
-multiple queues in the mobile application
-separate fees according to the time given to build the batch

UX Mockups
image


tiankii:
The real PR is this one: https://github.com/blinkbitcoin/blink-mobile/pull/3462

PR #3410, opened 2025-04-16
chore: more visibility on the build task

k9ert:
* Some improvements for scripts
* Upgrading to fastlane 2.227.1 to solve this [issue](https://ci.blink.sv/teams/galoy-mobile/pipelines/blink-mobile-deployments/jobs/upload-to-play-store/builds/4) (regression [issue](https://github.com/fastlane/fastlane/issues/21944) for fastlane).


android/fastlane/Fastfile
```diff
@@ -41,6 +41,13 @@ platform :android do
       track: "internal",
       skip_upload_changelogs: true,
       skip_upload_images: true,
+      release_status: "completed",
```
dolcalmi, line 44: this is the default value? are you sure this makes a difference?

android/fastlane/Fastfile
```diff
@@ -41,6 +41,13 @@ platform :android do
...
+      # Add health-related declarations
+      app_store_metadata: {
```
dolcalmi, line 46: this is for appstore and it is in the android config file... what error are you getting?

dolcalmi (changes requested):

android/fastlane/Fastfile
k9ert: Yes, exactly. reverted.

PR #3446, opened 2025-05-27
refactor: multi account redesign

tiankii:
-PR: https://github.com/blinkbitcoin/blink-mobile/pull/3446

@dolcalmi This is done!

### DIFF
--- a/__tests__/screens/settings-screen/settings-screen.spec.tsx
+++ b/__tests__/screens/settings-screen/settings-screen.spec.tsx
@@ -55,6 +55,7 @@ describe("Settings Screen", () => {
         }),
     )
 
-    expect(screen.getByText("test1@blink.sv")).toBeTruthy()
+    const elements = screen.getAllByText("test1@blink.sv")
+    expect(elements.length).toBeGreaterThan(0)
   })
 })


app/utils/multi-account/index.ts
dolcalmi, line 1: app/utils/multi-account/ it is a component util.. why was moved to app utils?

app/components/set-lightning-address-modal/set-lightning-address-modal.tsx
```diff
@@ -99,6 +100,12 @@ export const SetLightningAddressModal = ({
       return
     }
 
+    // Short delay to ensure GraphQL updates propagate before refreshing the current profile
+    await new Promise<void>((resolve) => {
+      setTimeout(() => resolve(), 100)
+    })
+
+    await updateCurrentProfile()
```
dolcalmi: this does not makes sense, why are you adding the timeout before the update?  is this really related?

app/screens/settings-screen/account/multi-account/profile.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const handleProfileSwitch = async (nextToken?: string) => {
+    setSwitchLoading(true)
+    await saveToken(nextToken ?? token)
+
+    // Small delay to ensure the new session token is processed and updated in the global state before navigating.
+    // This prevents the "Primary" screen from initially loading data using the old session token.
+    await new Promise<void>((resolve) => {
+      setTimeout(() => resolve(), 100)
+    })
```
dolcalmi: we cant use timeouts every time we have a sync issue, please find the root cause

app/screens/settings-screen/account/multi-account/profile.tsx
```diff
@@ -0,0 +1,216 @@
...
+          <Text type="h1" bold>
+            {hasUsername ? `${identifier}@${lnAddressHostname}` : identifier}
+          </Text>
+          <Text type="h1" bold>
+            {LL.ProfileScreen.fromThisDevice()}
+          </Text>
+        </View>
+        <View style={styles.actionButtons}>
+          <GaloyPrimaryButton
+            title="Confirm"
```
dolcalmi: use localization

app/screens/settings-screen/account/multi-account/profile.tsx
```diff
@@ -0,0 +1,216 @@
...
+            onPress={async () => {
+              await handleLogout()
+              setLogoutLoading(false)
+            }}
+          />
+          <GaloySecondaryButton title="Cancel" onPress={closeModal} />
```
dolcalmi: same here

app/screens/settings-screen/account/multi-account/profile.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const [switchLoading, setSwitchLoading] = useState(false)
+  const [logoutLoading, setLogoutLoading] = useState(false)
+  const [modalVisible, setModalVisible] = useState(false)
+
+  const { saveToken } = useAppConfig()
+  const { logout } = useLogout()
+
+  const handleProfileSwitch = async (nextToken?: string) => {
+    setSwitchLoading(true)
+    await saveToken(nextToken ?? token)
```
dolcalmi: are we sure this will not throw? also here we can be a bit more "defensive", please use ||

app/screens/settings-screen/account/multi-account/profile.tsx
```diff
@@ -0,0 +1,216 @@
...
+    setSwitchLoading(false)
+    toastShow({
+      type: "success",
+      message: LL.ProfileScreen.switchAccount(),
+      LL,
+    })
+    navigation.navigate("Primary")
+  }
+
+  const handleLogout = async () => {
```
dolcalmi, line 57: same here, are we sure we are handling correctly possible exceptions? (token related methods)

app/components/balance-header/balance-header.tsx
```diff
@@ -62,15 +59,9 @@ export const BalanceHeader: React.FC<Props> = ({ loading, formattedBalance }) =>
 
 const useStyles = makeStyles(({ colors }) => ({
   balanceHeaderContainer: {
-    flex: 1,
```
dolcalmi, line 65: did you test the new styles/behavior with big amounts?

app/components/balance-header/balance-header.tsx
```diff
@@ -79,11 +70,6 @@ const useStyles = makeStyles(({ colors }) => ({
   marginBottom: {
     marginBottom: 4,
   },
-  hiddenBalanceTouchableOpacity: {
```
dolcalmi, line 82: same for this... specially if it centers it correctly

app/screens/home-screen/home-screen.tsx
```diff
@@ -609,16 +655,41 @@ const useStyles = makeStyles(({ colors }) => ({
     maxWidth: "25%",
     flexGrow: 1,
   },
+  balanceContainer: {
+    marginBottom: 30,
+    marginTop: 7,
```
dolcalmi: did you test this with different screen sizes and in iOS? (applies for all fixed values in margins or paddings)

app/screens/settings-screen/account/multi-account/switch-account.tsx
```diff
@@ -0,0 +1,83 @@
...
+  )
+}
+
+export const useStyles = makeStyles(() => ({
+  outer: {
+    marginTop: 4,
+    paddingBottom: 20,
+    display: "flex",
+    flexDirection: "column",
+    // rowGap: 2,
```
dolcalmi: remove

app/screens/home-screen/home-screen.tsx
```diff
@@ -453,6 +460,26 @@ export const HomeScreen: React.FC = () => {
     </Modal>
   )
 
+  React.useEffect(() => {
```
dolcalmi: we already discuss this in the original PR... this is unnecessary and/or we need to find an alternative

app/utils/multi-account/index.ts
JMoises-XCode: The multi account logic is no longer on the settings screen, it's now on a different screen, and it was moved because the utilities are also accessed from the home screen.

app/components/set-lightning-address-modal/set-lightning-address-modal.tsx
JMoises-XCode: On the settings screen, the lnAddress appears in the header if it's present; otherwise, it shows the username or phone number. Now, when the profile is updated to add an lnAddress, the update is sent to the settings screen upon completion. This delay ensures that the changes propagate and that I can see my lnAddress in the settings header.

In short, multi-accounting is a local process and is much faster than GraphQL. Therefore, even if I perform the latest update, GraphQL doesn't update globally as quickly, which means that `updateCurrentProfile` doesn't yet reflect the latest changes.

app/screens/settings-screen/account/multi-account/profile.tsx
JMoises-XCode: In fact, that delay was unnecessary. It was deleted

app/screens/settings-screen/account/multi-account/profile.tsx
JMoises-XCode: Done

app/screens/settings-screen/account/multi-account/profile.tsx
JMoises-XCode: Done

app/screens/settings-screen/account/multi-account/profile.tsx
JMoises-XCode: It shouldn't fail; its purpose is simply to switch to the next profile if you log out of the current session.

app/screens/settings-screen/account/multi-account/profile.tsx
JMoises-XCode: Yes, if you have no more accounts, closing the session will delete everything and send you back to the login; otherwise, you will remain in the application with the next session.

app/components/balance-header/balance-header.tsx
JMoises-XCode: I tried it with large quantities, and it caused problems, so I refactored it.

app/components/balance-header/balance-header.tsx
JMoises-XCode: fixed

app/screens/home-screen/home-screen.tsx
JMoises-XCode: Just in android devices

app/screens/settings-screen/account/multi-account/switch-account.tsx
JMoises-XCode: Done

app/screens/home-screen/home-screen.tsx
JMoises-XCode: I removed it, and instead used the user Settings Screen Query that was already in use on that screen

dolcalmi (approved):

PR #3447, opened 2025-05-28
feat: save lnaddress contacts

tiankii:


dolcalmi (changes requested):
send ln address (external) change is missing

PR #3473, opened 2025-07-25
fix: onboarding flow issues

tiankii:


app/components/option-selector/option-icon.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react"
+import { View } from "react-native"
+import { makeStyles, useTheme, Icon } from "@rneui/themed"
+import { GaloyIcon, IconNamesType } from "../atomic/galoy-icon"
+
+export const OptioIcon = ({
```
dolcalmi: `OptionIcon`

dolcalmi (changes requested):

app/components/option-selector/option-icon.tsx
esaugomez31: Ok, done!

dolcalmi (approved):

PR #3579, opened 2026-01-15
fix: send destination double confirm

esaugomez31:
## Summary
  - Prevent the “never sent money to this address” confirmation modal from reappearing for the same destination after the user confirms once.
  - Add tests to cover the single-confirmation behavior for known and unknown destinations.

  ## Changes
  - Track normalized handles and confirmation state to avoid re-triggering the modal for the same destination.
  - Add test coverage for:
    - confirm shown once for same destination
    - confirm shown again for a different destination
    - no confirm for known contact
    - confirm checkbox gating


grimen (approved):
LGTM

Issue #8, opened 2020-12-02
add bitcoin: and lightning: as compatible link


nicolasburtey:

juscamarena:
It'd also be nice to add deep linking: galoy:lightning:xxx, or bitcoinbeach:lightning:xxx. iOS is a struggle with wallet all vying to trigger on the lightning: or bitcoin: uri

theamodhshetty:
Opened a fix here: https://github.com/blinkbitcoin/blink-mobile/pull/3670\n\nThis adds  and  to the compatible deep-link prefixes (while keeping existing  / ) and includes a test to ensure both formats stay supported.\n\nCould maintainers also confirm whether the  label is currently active and, if so, current payout/claim details?

joelklabo:
To resolve this issue, support for parsing and interpreting links with the `bitcoin:` and `lightning:` URI schemes can be added. This would improve compatibility for users leveraging these widely-adopted payment protocols. If maintainers approve the implementation path, I can submit a draft PR.

PR #3657, opened 2026-02-14
fix: add bitcoin: and lightning: as compatible link

joelklabo:
## Summary
- Added support for bitcoin: and lightning: URI schemes in QR code scanner
- These URIs are now recognized as valid links that can be opened in browser
- Fixes issue #8


joelklabo:
Good point! JavaScript's URL API automatically normalizes the protocol scheme to lowercase, so  will work correctly. The scheme gets lowercased internally during URL parsing, so both uppercase and lowercase variants are handled automatically. 

That said, if you'd prefer explicit handling for clarity, let me know and I can add a case-insensitive check. Otherwise, I believe this is good to go as-is.

joelklabo:
Closing as superseded by #3670, which covers the same bitcoin:/lightning: URI behavior with tests. The underlying issue #8 is also closed.

PR #3663, opened 2026-02-17
feat(ui): amount input modal redesign

esaugomez31:
# Amount Input Redesign

## What changed

The **set amount modal** (the screen where users type how much to send/receive) was redesigned. Instead of navigating to a full-screen modal, it now opens as a **bottom sheet** (`@gorhom/bottom-sheet`) that slides up from the bottom. The user's preferred currency is remembered between sessions, and the send flow now shows an inline balance-exceeded error.

### Bottom sheet modal
- Replaces `react-native-modal` full-screen overlay with `@gorhom/bottom-sheet` `BottomSheetModal`
- Slides up on tap, dismisses on swipe down or backdrop tap
- `BottomSheetModalProvider` added to the app root
- Amount is applied only **after** the dismiss animation completes (`pendingSetAmountRef` pattern), preventing visual glitches

### Redesigned amount input UI
- New layout with primary/secondary currency rows inside a rounded card
- `CurrencyPill` (outlined variant) shows the active currency code per row
- Circular toggle button centered between rows to swap currencies
- Hidden `TextInput` overlay for clipboard paste support
- Inline error row with warning icon (replaces the old `GaloyErrorBox`)

### Currency preference persistence
- When the user confirms an amount, the app remembers which currency they last used (display currency vs wallet currency)
- Next time the modal opens, it starts with the preferred currency
- Stored locally via Apollo Client cache (`preferredAmountCurrency` client-only field)
- Uses the reactive `usePreferredAmountCurrencyQuery` hook

### Inline balance-exceeded error (send flow)
- The send details screen now passes the wallet balance as `maxAmount` when no LNURL limit exists
- New `maxAmountIsBalance` flag triggers a distinct error message: _"Exceeds available balance: {amount}"_
- New i18n key `exceedsAvailableBalance` translated across all locales

### Accessibility
- Hidden `TextInput` excluded from screen readers (`accessibilityElementsHidden`, `importantForAccessibility="no"`)
- Toggle button has `accessibilityLabel="Switch currencies"` and `accessibilityRole="button"`

### Code cleanup
- Extracted `formatNumberPadNumber` into `number-pad-reducer.ts` — now currency-aware (appends "SAT" suffix for BTC) and reused by both amount input screens
- Removed unused `goBack` prop from the entire `AmountInputScreen` → `AmountInputScreenUI` chain
- `AmountInput` now always renders both the button and the modal as siblings (modal visibility controlled by `isOpen` state) instead of conditionally swapping between them

### i18n fixes
- Lowercase "lightning" in non-sentence-initial positions across affected translations (ar, ca, el, id, lg, ms, ro, sw, th, vi, xh)
- Added missing Armenian translations for `ReceiveScreen` keys

## Tests

- **8 new test files**: `amount-input-modal`, `amount-input`, `currency-pill` (outlined), `number-pad-reducer`, `send-amount-input-screen-ui`, `send-amount-input-screen`, `client-only-query`, `amount-input` integration
- **Updated**: `receive.spec.tsx` (bottom-sheet mock, dismiss lifecycle tests), `send-details.spec.tsx` (dismiss animation test), `send-destination.spec.tsx` (extra flushAsync)
- Global `@gorhom/bottom-sheet` mock at `__mocks__/@gorhom/bottom-sheet.tsx` simulates present/dismiss with configurable animation timing

## Dependencies

- Added `@gorhom/bottom-sheet@^5.2.8` (peer deps `react-native-reanimated`, `react-native-gesture-handler` — already installed)

### DIFF
--- a/__tests__/components/send-amount-input-screen.spec.tsx
+++ b/__tests__/components/send-amount-input-screen.spec.tsx
@@ -7,7 +7,9 @@ import { AmountInputScreen } from "@app/components/amount-input-screen/amount-in
 const mockFormatMoneyAmount = jest.fn()
 const mockGetSecondaryAmount = jest.fn()
 const mockSavePreferredAmountCurrency = jest.fn()
-const mockPreferredCurrencyData = { data: null as { preferredAmountCurrency: string } | null }
+const mockPreferredCurrencyData = {
+  data: null as { preferredAmountCurrency: string } | null,
+}
 
 jest.mock("@apollo/client", () => ({
   ...jest.requireActual("@apollo/client"),


grimen:
## Code Review: Amount Input Modal Redesign

### Overview

This PR converts the amount input from a full-screen `react-native-modal` to a `@gorhom/bottom-sheet` bottom sheet, adds currency preference persistence via Apollo cache, extracts `formatNumberPadNumber` into a shared module, and fixes a deep link re-processing bug. Includes comprehensive test coverage.

**+1682 / -267** across ~20 files. Well-structured, well-tested.

---

### Strengths

- **Good test coverage**: 6 new test suites covering the modal, input, UI, number pad formatter, screen logic, and client-only query
- **Clean extraction** of `formatNumberPadNumber` into a reusable module with currency-aware formatting
- **`processedPaymentRef` fix** in `send-bitcoin-destination-screen.tsx` is a solid improvement — prevents duplicate deep link processing on re-renders
- **Removal of duplicated `AmountInputButton` branches** in `amount-input.tsx` — the old code had two nearly identical return blocks for `canSetAmount` true/false, now unified into one
- **Currency preference persistence** via Apollo cache is a nice UX improvement with proper error handling (try/catch in `savePreferredAmountCurrency`)

---

### Issues & Suggestions

#### 1. Missing `dismiss()` when `isOpen` transitions to `false` (Medium)

In `amount-input-modal.tsx`, the `useEffect` only calls `present()`:

```tsx
useEffect(() => {
  if (isOpen) {
    bottomSheetRef.current?.present()
  }
}, [isOpen])
```

If the parent sets `isOpen=false` programmatically (without the sheet initiating the close via swipe/backdrop), the sheet won't dismiss. Consider:

```tsx
useEffect(() => {
  if (isOpen) {
    bottomSheetRef.current?.present()
  } else {
    bottomSheetRef.current?.dismiss()
  }
}, [isOpen])
```

Currently the flows all work because dismiss is always triggered by the sheet itself, but this is fragile — a future caller could break it.

#### 2. Inline style creates new object every render (Low)

`amount-input-modal.tsx`:
```tsx
<BottomSheetView style={{ paddingBottom: bottom }}>
```

This creates a new style object on every render. Consider memoizing:
```tsx
const contentStyle = useMemo(() => ({ paddingBottom: bottom }), [bottom])
```

#### 3. Hardcoded font family (Low)

`currency-pill.tsx`:
```tsx
outlinedText: {
  fontFamily: "Source Sans Pro",
```

Hardcoded font family string. If the app has a font theme or constants file, this should reference that instead for consistency.

#### 4. `resolvedInitialAmount` creates new object references (Low)

In `amount-input-screen.tsx`, the IIFE:
```tsx
const resolvedInitialAmount = (() => {
  // ...
  return { amount: 0, currency, currencyCode: ... } as MoneyAmount<...>
})()
```

This runs every render and creates new objects. Since it's only consumed by `useReducer`'s initial state (which ignores subsequent values), this is functionally correct but wasteful. Could be wrapped in `useMemo` for clarity of intent.

#### 5. `compact` prop silently dropped (Informational)

The current code on `main` passes `compact` to `AmountInputScreen` from the modal. The new code doesn't pass it. If `compact` affected layout behavior, verify this is intentional and no other consumers depend on it.

#### 6. `noSuffix` on `formatMoneyAmount` removed (Verify)

In `amount-input-screen.tsx`, the secondary amount formatting changed:
```diff
- formatMoneyAmount({ moneyAmount: secondaryNewAmount, noSuffix: true, noSymbol: true })
+ formatMoneyAmount({ moneyAmount: secondaryNewAmount, noSymbol: true })
```

The suffix (e.g., "SAT") is now shown for secondary amounts. Since the currency code is now displayed via a `CurrencyPill` next to the amount, having both the pill and the suffix could result in double-labeling (e.g., pill says "BTC" and amount shows "1,500 SAT"). Verify this renders correctly.

#### 7. Double `flushAsync` in test (Nit)

`send-destination.spec.tsx`:
```tsx
await flushAsync()
await flushAsync()
```

Consider adding a comment explaining why two flush cycles are needed to avoid future confusion.

---

### Potential Risks

- **New dependency**: `@gorhom/bottom-sheet@5.2.8` — peer dependencies (`react-native-reanimated`, `react-native-gesture-handler`) are already in the project, so this should be safe. The `@gorhom/portal` transitive dep adds `nanoid`.
- **`BottomSheetModalProvider` at app root**: Added inside `RootSiblingParent` but wrapping `NotificationsProvider`. This is the correct placement for bottom sheets to work everywhere, but increases the provider nesting depth by one.
- **Generated code changes** (`generated.ts`): The PR includes regenerated GraphQL types from staging schema. Verify this was intentional and doesn't include unrelated schema changes.

---

### Verdict

Good PR overall. The main actionable item is **#1** (adding `dismiss()` to the `useEffect` for robustness). The rest are minor or informational. The code is well-structured, the test coverage is solid, and the UX improvements (bottom sheet + currency preference) are well-motivated.

esaugomez31:
@grimen and @blink-claw-bot review
### #1 - Missing `dismiss()` when `isOpen` transitions to `false` (Medium)
**Applied.** Good defensive practice — if in the future the modal is closed programmatically (setting `isOpen=false` from the parent without user interaction on the sheet), the sheet will now dismiss properly instead of staying visually open.

### #2 - Inline style creates new object every render (Low)
**Applied.** Instead of memoizing with `useMemo`, I preferred to move the style into the existing `makeStyles` hook using its dynamic props parameter — this follows the project's established pattern and is a cleaner approach.

### #3 - Hardcoded font family (Low)
**Not applied.** The font is correctly installed in the project (`app/assets/fonts/SourceSansPro-*.ttf`) and linked for both platforms. There's no font constants file or theme font config in the project at the moment, and it's only referenced once in this file, so I don't consider it necessary to create one.

### #4 - `resolvedInitialAmount` creates new object references (Low)
**Not applied.** `useReducer` only reads the initial value once (on the first render) and ignores it on subsequent renders. Wrapping it in `useMemo` would work but would be an unnecessary memoization since the value is never re-consumed.

### #5 - `compact` prop silently dropped (Informational)
**Intentional.** The previous keyboard had two variants (default and compact) — some screens used the default and others required the compact style for tighter layouts. The new keyboard is fully responsive and doesn't need variants, so `compact` was removed along with all its existing usages as part of the keyboard refactor.

### #6 - `noSuffix` on `formatMoneyAmount` removed (Verify)
**Verified, no issue.** The `CurrencyPill` only displays general currency codes like BTC, USD, EUR, etc. The formatted value uses a different representation — for example, USD shows `$5.00` (symbol prefix) and Bitcoin shows `15,200 SAT` (suffix). Since the pill and the formatted value use different labels, there's no duplication.

### #7 - Double `flushAsync` in test (Nit)
**Intentional.** This test was flaky — it failed roughly 4 out of 10 runs because a single flush cycle wasn't enough time for React to fully process the state updates. Adding the second `flushAsync` gives it an extra tick to settle and made the test consistently pass

grimen (approved):
LGTM

grimen:
@esaugomez31 Can you just clarify the above concerns. Btw the `fontFamily` issue I don't think makes sense to ignore (per earlier comments) since no where else in the `blink-mobile` aside of the theme `fontFamily` is set in stylesheet - I don't see why we should need to do it for the currency pill?

esaugomez31:
@grimen @blink-claw-bot Review

### Summary

High Priority

1. resolvedInitialAmount depends on preferredCurrency which is read synchronously outside useMemo — Applied, replaced client.readQuery with the reactive usePreferredAmountCurrencyQuery generated hook, following the same convention used by useHideBalanceQuery, useBetaQuery, etc.

2. Balance as maxAmount in send flow may be misleading — Not applicable, this PR reused the existing business logic for max amount calculation, no business logic was changed. Fee-aware balance validation could be evaluated as an improvement in a separate ticket.

3. Hidden TextInput paste sanitization is fragile — Not applicable, tested and it is not possible to paste text even using long-press. The hidden TextInput with opacity: 0 does not trigger the system paste menu. The onChangeText handler is a defensive fallback that is never reached in the current UI. Additionally, this PR reuses the previous business logic that was already working correctly, changing that logic is out of scope.

Medium Priority

4. BottomSheetModalProvider wraps GaloyToast — Not applicable, it was placed at that level and tested to confirm it does not affect toasts. GaloyToast uses react-native-root-siblings (via RootSiblingParent) which renders at the native root level, not inside the bottom sheet portal.

5. Hardcoded dismiss animation timing in tests (40ms) — Not applicable, the mock is a static file we control that simulates the bottom sheet screen. It exists only for testing purposes to simulate a similar dismiss animation behavior. Exporting the constant would add unnecessary coupling between mock internals and test files.

6. formatNumberPadNumber shows "0 SAT" for empty BTC amounts but empty string for others — Not applicable, this is intentional. BTC amounts display "0 SAT" so the user always knows the unit. Fiat currencies already show the symbol separately (e.g. "$"). Verified and covered in unit tests.

7. Transfer amount input screen uses noSuffix: true but the signature changed — Correct observation. The "SAT" suffix was added intentionally for UI purposes so the user can see the unit in the new amount input view. The transfer screen passes noSuffix: true to preserve the previous behavior, while send/receive screens omit it to display the new suffix. All call sites have been verified.

Low Priority

8. Inline pressed opacity style — Applied, extracted inline { opacity: 0.7 } to a toggleButtonPressed style in the stylesheet.

9. Error row always renders a warning icon (transparent when no error) — Not applicable, rendering the icon with color="transparent" reserves the exact layout space and prevents visual shift when the error appears or disappears. A spacer View would need to manually match the icon+text height.

10. Several translation files also lowercase "Lightning" — This ticket only updated the translation keys related to the screens being worked on, across all languages. It is not in scope to change all translations, only those belonging to the affected screens.

grimen (approved):
LGTM - should be good enough for this iteration.

PR #3670, opened 2026-02-23
fix: support bitcoin and lightning links without //

theamodhshetty:
## Summary
- add `bitcoin:` and `lightning:` deep-link prefixes while keeping existing `bitcoin://` and `lightning://`
- extract compatible payment prefixes into a dedicated navigation constant
- add a unit test to ensure both URI formats remain supported

## Context
Refs #8.
Some wallets/QR links use single-colon schemes (`bitcoin:...`, `lightning:...`), but navigation prefix matching only included `scheme://` variants.

## Testing
- `yarn test __tests__/navigation/linking-prefixes.spec.ts`
- `yarn test __tests__/screens/send-destination.spec.tsx -t "deep link payment processing"`

### DIFF
--- a/__tests__/navigation/linking-prefixes.spec.ts
+++ b/__tests__/navigation/linking-prefixes.spec.ts
@@ -1,0 +1,9 @@
+import { COMPATIBLE_LINK_PREFIXES } from "@app/navigation/linking-prefixes"
+
+describe("compatible link prefixes", () => {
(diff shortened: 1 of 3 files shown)


theamodhshetty:
Happy to adjust scope if you’d like support for additional app-prefixed formats (for example `blink:lightning:...`) in this PR or a follow-up.

Issue #534, opened 2022-07-06
Implement select primary wallet feature


UncleSamtoshi:
Add a row in settings

PR #3679, opened 2026-02-25
feat(card): card flow details integration


esaugomez31:
# Card Details Screen — Backend Integration

## Summary

Replaces mock data in the Card Details screen with real GraphQL data via the shared `useCardData()` hook, adds biometric authentication gating, and extracts a shared card number formatting utility.

### Core: Card Details Screen
- Replace `MOCK_CARD` / local `CardStatus` with real data from `useCardData()` hook
- Add biometric authentication gate on screen entry via new `useBiometricGate` hook
- Navigate back on biometric failure or when card is not found
- Show `ActivityIndicator` during loading / authentication
- Map all 6 `CardStatus` values (Active, Locked, Canceled, NotActivated, Requested, Failed) with correct colors
- Format card type from `CardType` enum via i18n (Virtual/Physical)
- Format issued date from `createdAt` using locale-aware `toLocaleDateString`
- Hardcode network as "Visa"
- Placeholder fields (expiry, CVV, cardholder name) show "—" with disabled interaction (pending backend support)
- Card number field shows formatted display number and copies `lastFour` on press
- Auto-clear clipboard after 60 seconds
- Settings header button navigates to `cardSettingsScreen`

### New: Biometric Gate Hook
- Reusable hook that wraps `BiometricWrapper.isSensorAvailable` + `authenticate`
- Returns `authenticated` boolean
- Skips biometric if sensor is not available
- Refs synced on every render to avoid stale closures with empty `[]` deps

### Shared Utility: `formatCardDisplayNumber`
- Extracted card number masking logic into a shared utility
- Configurable via options: `totalDigits`, `groupSize`, `visibleDigits` (all with sensible defaults)
- Used by both `BlinkCard` component and `CardDetailsScreen`

### Enhanced: `ActionField`
- `value` is now optional (`string | null | undefined`), defaults to placeholder "—"
- `onAction` is now optional
- When value is null/placeholder: press is disabled, icon grayed out
- Backwards-compatible — existing callers with string values work unchanged

### Enhanced: `useClipboard`
- New optional `clearAfterMs` parameter
- Schedules clipboard clear after the specified delay
- Clears previous timer on subsequent copies (no double-clear)
- Cleanup on unmount

### Refactored: `BlinkCard`
- Replaced inline masking logic with shared `formatCardDisplayNumber` call
- Removed duplicated masking constants and imports

### Display Utilities
- `formatCardType(cardType, LL)` — maps `CardType` enum to i18n label
- `formatIssuedDate(isoDate, locale)` — locale-aware date formatting
- Added `ViewCardDetails` to `AuthenticationScreenPurpose`

### i18n
- 7 new keys in `CardDetails`: `statusCancelled`, `statusNotActivated`, `statusPending`, `statusFailed`, `cardTypeVirtual`, `cardTypePhysical`, `authDescription`
- All 22 CardDetails keys properly translated in 28 language files with diacritical marks

### Tests
- Full rewrite of card details screen tests (24 tests): biometric auth, loading, card-not-found, data rendering, card info, copy interactions, frozen state
- Unit tests for `formatCardDisplayNumber`, `ActionField` disabled state, `useClipboard` auto-clear, `formatCardType`, `formatIssuedDate`
- Fixed unrelated test mocks broken by the new shared utility

## Not in Scope
- Full PAN / CVV reveal (pending backend: #534)
- Expiry date and cardholder name fields (pending backend)

### DIFF
--- a/__tests__/components/action-field/action-field.spec.tsx
+++ b/__tests__/components/action-field/action-field.spec.tsx
@@ -139,9 +139,7 @@ describe("ActionField", () => {
     })
 
     it("renders without onAction prop", () => {
-      const { getByText } = render(
-        <ActionField icon="copy-paste" value="Some value" />,
-      )
+      const { getByText } = render(<ActionField icon="copy-paste" value="Some value" />)
 
       expect(getByText("Some value")).toBeTruthy()
     })
--- a/app/components/action-field/action-field.tsx
+++ b/app/components/action-field/action-field.tsx
@@ -26,7 +26,7 @@ export const ActionField: React.FC<ActionFieldProps> = ({
     theme: { colors },
   } = useTheme()
 
-  const hasValue = value != null && value !== PLACEHOLDER
+  const hasValue = value !== null && value !== undefined && value !== PLACEHOLDER
   const iconColor = hasValue ? colors.primary : colors.grey3
 
   return (


esaugomez31:
@grimen @blink-claw-bot please Review

## Code Review

- **# 1** `formatCardDisplayNumber` regex can crash on non-standard `totalDigits`/`groupSize` combos — `app/utils/helper.ts`
  - **Response:** Applied — changed regex to `.{1,${groupSize}}` with `?? [masked]` fallback. Added tests for non-divisible and undersized combos.
- **# 2** `cardStatusConfig` is rebuilt on every render — `card-details-screen.tsx`
  - **Response:** Applied — wrapped `cardStatusConfig` in `useMemo` with `[colors, LL]` dependencies and moved it before early returns to respect React hooks rules.
- **# 3** `AuthenticationScreenPurpose.ViewCardDetails` is added but never used — `app/utils/enum.ts`
  - **Response:** Applied — removed the unused enum value.

## Nits

- **# 4** Hardcoded "Visa" network — raw English string instead of parameterized
  - **Response:** Applied — added `networkVisa` i18n key at `CardFlow` level and replaced the hardcoded string with `LL.CardFlow.networkVisa()`.
- **# 5** `useClipboard` timer cleanup edge case with dynamic `clearAfterMs`
  - **Response:** Applied — added `clearAfterMs` to the cleanup effect dependency array and clear the timer when it becomes falsy.
- **# 6** Commented-out `onAction` handlers are redundant since `value={null}` already disables interaction
  - **Response:** Intentionally kept commented out — these fields (expiry, CVV, cardholder) are not yet implemented in the backend. The `onAction` is commented so copy doesn't work while data is unavailable, and `value={null}` disables the field visually. Once the backend exposes these values, we'll uncomment the handlers.

PR #3689, opened 2026-03-04
fix(ui): padding change from 12 to 5px

designsats:
Small fix of bottom padding of WalletOverview component


designsats:
@grimen one line change

PR #3693, opened 2026-03-04
feat(card): card flow shipping address integration

esaugomez31:
## What was done

Implements the frontend logic for the Card Shipping Address screen. Covers **Phase 1** (frontend validation & UX) and partial **Phase 2** (read-only backend integration). Save mutation deferred until backend is ready.

---

### Phase 1: Frontend Validation & UX — Completed

- [x] P.O. Box detection on address lines (`validatePOBox`)
- [x] Postal code format validation by country using `postcode-validator` library
- [x] Required field validation with min-char rules (firstName/lastName: 2, line1: 5, city: 2)
- [x] Dynamic country/region dropdown data (USA + Canada with all states/provinces)
- [x] Dirty state tracking via `useRef` + `JSON.stringify` comparison
- [x] Discard confirmation dialog on back navigation when form has unsaved changes (save button pending Phase 2)
- [x] InputField enhanced with `required`, `minLength`, `validate`, `error` props and inline validation on blur/type

### Phase 2: Backend Integration — Partial

**Completed (read):**
- [x] Fetch shipping address from `Card.shippingAddress` via `usePersonalDetailsQuery`
- [x] `useShippingAddressData` hook with `cache-first` fetch policy
- [x] Pre-populate form with existing address (map nullable GQL fields to non-null form values)
- [x] Handle empty state (no existing shipping address → empty form)
- [x] Loading state with activity indicator
- [x] Error handling with warning toast
- [x] Memoized `initialAddress` to prevent infinite re-render loop

**Pending (blocked on backend):**
- [ ] Save/update shipping address via mutation (`updateShippingAddress` or `cardUpdate` extension — mutation doesn't exist yet)
- [ ] Save button with loading state
- [ ] Success toast + navigate back + cache invalidation on save
- [ ] `CardError` extension code handling (`CARD_NOT_FOUND`, `OPERATION_NOT_PERMITTED`, `SERVICE_UNAVAILABLE`)

### Additional changes

- Aligned `ShippingAddress` type with GQL field names, split `fullName` into `firstName`/`lastName`
- Improved placeholder contrast in InputField

### Notes

> **City & Postal Code as dropdowns**: Originally designed as editable text fields in Figma, and implemented that way in this phase (postal code validated via `postcode-validator` library). The current Figma now shows them as dropdowns like State/Country. This change can be addressed in a follow-up — for now the text input approach works and validates correctly.

### DIFF
--- a/__tests__/screens/card-screen/card-personal-details-screen/card-personal-details-screen.spec.tsx
+++ b/__tests__/screens/card-screen/card-personal-details-screen/card-personal-details-screen.spec.tsx
@@ -24,18 +24,6 @@ jest.mock("@react-navigation/native", () => {
   }
 })
 
-type MockShippingAddress = {
-  firstName: string | null
-  lastName: string | null
-  line1: string
-  line2: string | null
-  city: string
-  region: string
-  postalCode: string
-  country: string | null
-  countryCode: string
-}
-
 type PersonalDetailsReturn = {
   firstName: string
   lastName: string
@@ -43,21 +31,9 @@ type PersonalDetailsReturn = {
   onboardingStatus: OnboardingStatus | null
   email: string
   phone: string
-  shippingAddress: MockShippingAddress | null
+  shippingAddress: null
   loading: boolean
   error: Error | undefined
-}
-
-const defaultMockShippingAddress: MockShippingAddress = {
-  firstName: "John",
-  lastName: "Doe",
-  line1: "456 Oak Avenue",
-  line2: "Suite 7",
-  city: "Miami",
-  region: "FL",
-  postalCode: "33101",
-  country: "United States",
-  countryCode: "US",
 }
 
 const defaultMockData: PersonalDetailsReturn = {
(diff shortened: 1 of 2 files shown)


esaugomez31:
@grimen @blink-claw-bot please Review

## Feedback

## 1. InputField validation timing

**Resolution:** Kept as-is — on mobile there's no tab between fields, so blur-only would let users fill the entire form with invalid data before seeing any error. Immediate feedback on type is better UX here.

---

## 2. Container gap reduced from 20 to 4

**Resolution:** Applied — replaced `helperTextSpacer` (View with `height: 13`) with `helperTextContainer` (View with `minHeight: 13`) for cleaner convention. The gap reduction from 20 to 4 is intentional: it reserves space to prevent layout shift when validation error messages appear/disappear.

---

## 3. Country displayed as code

**Resolution:** The country code is displayed per spec. The full country name is already shown in the selection list (COUNTRIES data), so the user sees "United States" when picking, and the stored code in the field. No change needed.

---

## 4. `validate` in useMemo deps

**Resolution:** Applied — stored `validate` prop in a `useRef` inside InputField and removed it from useMemo deps. The ref always holds the latest closure without triggering recomputation.

grimen (approved):

PR #3707, opened 2026-03-11
feat(card): card onboarding investment flow

esaugomez31:
### What's new

**6 flow screens** for the investment onboarding experience:

1. **Welcome Invest** - invitation to participate in the Blink financing round
2. **Company Valuation** - explains Blink's pre-money valuation and investment terms
3. **Select Invest** - choose the investment amount and equity percentage
4. **Term Sheet** - restates the chosen amount as units and equity
5. **Transfer Invest** - the step that moves the money, described below
6. **Insufficient Balance** - states what the investor holds, naming the wallet that balance sits in (the fullest of the two, which is the one a payment can draw on), and what is still missing; the "Deposit" button opens the receive screen

Flow: Welcome Invest > Company Valuation > Select Invest > Term Sheet > Transfer Invest > Insufficient Balance > Receive

**Not yet wired:**
- **Deposit Pending** - a waiting-for-confirmations status screen (clock icon + "Okay"). Registered in the navigator but not wired into the linear flow yet.

**Deep link:** only the flow's entry, `card/onboarding/investment`, is linkable. The later steps take figures the signing produces and are reached from it, so a link landing on one would open it with nothing to show.

**Developer tools:**
- Phase 2 (Investment) and Phase 3 (Card Dashboard) bulletin triggers in the developer screen

### The figures are the investor's own

The amount chosen on Select Invest travels through the flow, so the Term Sheet and the Transfer copy describe the investment the user picked. `investment-terms.ts` derives the units, the price per unit and the equity from that one figure, and the round's valuation stays the same whatever the choice.

The balance is read the same way. `useInvestmentFunding` converts both wallets to dollars and measures the fullest one against what was chosen, since a payment draws on a single wallet; the two added together only say whether consolidating would be enough. It reads through the active account, so it answers for a custodial and a self-custodial investor alike. The comparison is made in dollars, which is the currency the agreement is written in: the balance is shown to the investor in whatever currency they chose to read it in, and that choice has no bearing on whether the amount is covered.

The Insufficient Balance screen states that balance and the shortfall against the chosen amount, in place of the fixed figures it used to print. Once the balance covers the amount it closes itself, but only when it is back in front: a deposit landing while the receive screen is still up must not pop that screen from underneath.

### Paying for it

Transfer Invest decides where the investor goes. Short of the amount, the shortfall screen; holding it, the send flow.

The send flow is opened on an **invoice**, not on the receiving account: an invoice carries its amount, and an account alone would let the investor send any sum against an agreement that names one. `lnInvoiceCreateOnBehalfOfRecipient` mints it against the receiving wallet - the mutation is unauthenticated by design, which is how any payer asks a Blink account for an invoice, so it answers for both account types. The invoice carries a memo and an expiry, so an abandoned attempt leaves no payable claim behind. When the API refuses to mint one, its reason is reported, so a wrong wallet id can be told apart from an outage.

The amount billed is the satoshis the signed agreement names, carried on the route from the signing step. The rate is fixed when the investor signs, so converting the dollars again at the current price would charge a different amount of bitcoin than the document states; the conversion stands in only while the route carries no figure.

Where the money goes is one remote-config key, `cardInvestmentDepositBtcWalletId`, set per environment. It names the receiving account's bitcoin wallet by id rather than by handle, so a renamed account cannot redirect the money, and the transfer step does not offer to pay while it is unset: the button holds and the step says payments are not open yet, rather than sitting grey over the investor's own money with nothing to explain it.

### Supporting changes

- Flow screens reuse the shared `IconHero` component; the deposit-pending screen reuses `StatusScreenLayout`
- 1 new icon (btc-outline)
- `formatUsdAmount` cuts to cents, which is as far as dollars go: the shortfall is the amount less the balance, and that subtraction leaves a floating-point tail that would otherwise reach the screen as is
- Investment mock data for the credit limits; the balance and amount placeholders are gone, replaced by the figures above
- Translations for all 28 supported languages
- Navigation routes for all new screens, and the entry deep link
- One `CloseHeader` draws the top-right close control for this flow and for the account migration, which had its own copy of it; it can be drawn without the control so a gated screen keeps its layout
- Every dollar figure in the flow goes through one formatter, in the app's locale, and the share each option buys is derived from the valuation rather than written beside it

### Test coverage

11 test files covering every investment flow screen and both hooks at 100% coverage, plus the terms the screens are written from and the shared `IconHero` and `FeatureItem` layout fixes. The behaviour worth naming: that a covered investor reaches the send flow and a short one reaches the shortfall screen, that the shortfall screen closes only once it is back in front (pinned on a real native stack), that the invoice is billed for the agreement's figure rather than today's conversion, that a failed mint says so, reports the API's reason and stays put, that a retry clears the failure line, and that nothing acts on a balance that has not arrived.

### Depends on

- #3706 - `feat(card): card onboarding KYC flow`


kngako (changes requested):
1 Blocking finding.

---

## Findings

### Finding 1 — Covering the balance while the receive screen is up pops the receive screen and strands the investor on a shortfall screen that says nothing is missing [BLOCKING]

**When it comes up:** An investor short by $5,000 taps Deposit on the shortfall screen, which pushes `receiveBitcoin` on top of it, and pays the invoice from another wallet while looking at the QR. `receive-screen.tsx` is wrapped in `withMyLnUpdateSub`, which refetches `HomeAuthedDocument` the instant the invoice is PAID ([my-ln-updates-sub.tsx:34](app/screens/receive-bitcoin-screen/my-ln-updates-sub.tsx:34)); the custodial wallet context updates, the *unfocused* shortfall screen re-renders, `isCovered` turns true and [insufficient-balance-screen.tsx:55-57](app/screens/card-screen/onboarding/investment-flow/insufficient-balance-screen.tsx:55) calls `navigation.goBack()`. That action carries `source` but no `target` (`useNavigationCache.tsx:154`), so `StackRouter`'s `POP` resolves `currentIndex` to `state.index` — the focused route ([StackRouter.tsx:549-553](node_modules/@react-navigation/routers/src/StackRouter.tsx:549)) — and removes the **receive screen**, not the shortfall screen. The investor's "payment received" view vanishes, they land on "Insufficient balance … Deposit more than $0 to your account to reach the investment amount of $25,000" with a Deposit button, and the effect never fires again because `isCovered` has not changed. Reproduced with a real native stack (see Verification). The spec at [insufficient-balance-screen.spec.tsx:133](__tests__/screens/card-screen/onboarding/investment-flow/insufficient-balance-screen.spec.tsx:133) cannot see this: it mocks `useNavigation` and rerenders the screen focused.

**Proposal:** Gate the effect on focus — `const isFocused = useIsFocused()` (23 precedents in `app/`, e.g. `card-dashboard-screen.tsx:48`) and `if (isCovered && isFocused) navigation.goBack()` with `isFocused` in the deps — so the check runs when the investor closes the receive screen and the shortfall screen regains focus, which is the moment the commit message describes ("right after the investor comes back from making one"). `useFocusEffect` with the same condition is the other idiomatic shape. Pin it with a spec that mounts a real stack (the helper's `createNativeStackNavigator` is enough): push the shortfall screen, push a stub receive screen, flip the funding to covered, and assert the receive screen is still on top and that closing it lands on the transfer step.

### Finding 2 — The split-funds Convert still ends on Home, four screens away from the transfer step [NON-BLOCKING]

**When it comes up:** The shortfall screen's Convert CTA navigates to `conversionDetails` ([insufficient-balance-screen.tsx:63](app/screens/card-screen/onboarding/investment-flow/insufficient-balance-screen.tsx:63)); `conversionConfirmation` resets the stack to `[Primary, conversionSuccess]` ([conversion-confirmation-screen.tsx:97](app/screens/conversion-flow/conversion-confirmation-screen.tsx:97)) and the success screen `popToTop()`s ([conversion-success-screen.tsx:66](app/screens/conversion-flow/conversion-success-screen.tsx:66)). The investment screens are unmounted by the reset, so the new auto-close never runs, and the investor who has just consolidated the money re-enters from the notification and walks Welcome → Valuation → Select → Term Sheet again. The Deposit path was fixed by `bc33c813a`; this half of the same finding was not.

**Proposal:** Use the mechanism that already exists for exactly this: `drain-conversion.ts` arms a `DrainConversionReturn` before entering the convert screen and `conversionSuccess` resumes the arming flow (`Migration`, `ModeSelection`). Add an `Investment` return carrying `selectedAmountUsd`, arm it in `handleConvert`, and have the success screen `replace` to `cardOnboardingTransferInvestScreen` for it — without the 100 % USD→BTC prefill the drain applies, which is not what an investor consolidating a shortfall wants. One spec on the success screen's `continueAfterSuccess` pins the new branch.

### Finding 3 — The invoice step swallows the API's reason, so the one misconfiguration the branch anticipates is invisible in Crashlytics [NON-BLOCKING]

**When it comes up:** [use-investment-invoice.ts:80-83](app/screens/card-screen/onboarding/investment-flow/use-investment-invoice.ts:80) returns `null` whenever `invoice` is missing and never reads `errors[]`; only a transport rejection reaches `reportError`. The remote-config comment at [feature-flags-context.tsx:217](app/config/feature-flags-context.tsx:217) predicts the failure — "A dollar wallet is refused by the API and reads as a failed invoice" — so if ops pastes the USD wallet id, every investor sees the borrowed "Failed to generate invoice" line and nobody is paged. `a2c7a1492` fixed the *unconfigured* case with a visible message; the *misconfigured* case is still silent.

**Proposal:** When the payload carries `errors`, call `reportError("investment-invoice", new Error(errors.map((e) => e.message).join("; ")))` before returning `null`, and consider surfacing that message in place of the receive-screen line, since "recipient wallet not found" is actionable for support in a way "Failed to generate invoice" is not. Pin it: the spec currently stays green when `reportError` is deleted outright (mutation M4), and the `errors[]` case at [use-investment-invoice.spec.ts:72](__tests__/screens/card-screen/onboarding/investment-flow/use-investment-invoice.spec.ts:72) asserts only `null`.

### Finding 4 — "Proceed to Sign" signs nothing, the payment carries no investor identity, and the Deposit Pending screen is now reachable by nothing [NON-BLOCKING]

**When it comes up:** [term-sheet-screen.tsx:57](app/screens/card-screen/onboarding/investment-flow/term-sheet-screen.tsx:57) navigates straight from "Proceed to Sign" to a screen that opens with "You have signed the subscription agreement for … units" ([en.json:3730](app/i18n/raw-i18n/source/en.json:3730)); `settlementSats` — "carried from the signing step" — has no producer anywhere in the tree, so the "TEMPORARY" today's-price conversion is the only path that runs. The invoice memo is a constant and the input's `externalId` is unused, so once `cardInvestmentDepositBtcWalletId` is set, anyone who opens `blink://card/onboarding/investment` can pay an arbitrary tier into the deposit wallet with no record of who or for which agreement. `9ab4d5b30` removed the deposit-pending deep link, which was right, but nothing else navigates to `cardOnboardingDepositPendingScreen` ([root-navigator.tsx:307](app/navigation/root-navigator.tsx:307)), so the screen, its two i18n keys, its spec and the `StatusScreenLayout.iconPadding` prop added for it are dead until "the home once signed" exists.

**Proposal:** Until the signing step exists, gate entry on something the server decides (an eligibility query, or at minimum a remote-config flag separate from the wallet id) and stop saying "signed" — "You chose … units" is true today. Put the payer's account id and the chosen tier in the invoice's `externalId` (the input already accepts `TxExternalId`) so the receiving ledger can tie each payment to a subscriber without depending on the memo. Either drop `DepositPendingScreen` from this branch and land it with its caller, or leave a `TODO` on the registration naming the producer, so the next reader does not search for one.

### Finding 5 — 20 of the 49 commits do not type-check on their own, and the fix-ups are appended rather than folded [NON-BLOCKING — history hygiene]

**When it comes up:** From `4dfd5f274` (screens added) through `308cd000c` (deposit pending) the tree imports `@react-navigation/stack`, which the repo does not have (fixed at `e10f3ddb8`), and calls `headerCloseControl()`, which `main` had already renamed to `headerCloseControlOptions` before this branch was cut (fixed at `ef1090bfa`) — measured `tsc` counts of 5, 11, 6 and 7 errors along the way. In the same span a `TopUpScreen` is added, given a spec "at 100 %" (`392d5591c`) and deleted (`20ab36434`), and a debug home shortcut is added and dropped. The eight review fix-ups sit on top as further commits, so the branch now records the bug and its correction as separate history for the deep links, the locale, the shortfall copy and the close header.

**Proposal:** Squash the two type fix-ups into the commits they fix, fold the TopUp add/test/remove and the debug-shortcut add/drop into nothing, and fold each of the eight fix-ups into the commit whose behaviour it corrects — or squash-merge the branch. If it is merged as-is, note in the merge commit that `4dfd5f274..308cd000c` is not bisectable.

---

## Test Coverage Gaps

1. **The shortfall auto-close is only tested focused.** [insufficient-balance-screen.spec.tsx:133-156](__tests__/screens/card-screen/onboarding/investment-flow/insufficient-balance-screen.spec.tsx:133) mocks `useNavigation` with a `goBack` spy and rerenders the screen itself, so it asserts the call, not its effect on the stack, and cannot see Finding 1. A real-stack case (push shortfall, push a stub receive screen, flip funding, assert the top route) is what pins the behaviour; once `useIsFocused` gates the effect, a second case with the screen unfocused asserts `goBack` is *not* called.

2. **`select-invest-screen.spec.tsx › selects an option on press` still proves nothing.** [select-invest-screen.spec.tsx:105-120](__tests__/screens/card-screen/onboarding/investment-flow/select-invest-screen.spec.tsx:105) presses an option and asserts `expect(option).toBeTruthy()` on the node it already found. Removing the selected style entirely leaves the suite 7/7 green (mutation M3). Assert the pressed row carries `styles.limitOptionSelected` (or an `accessibilityState.selected`), and that pressing a second option moves it.

3. **`use-investment-invoice.spec.ts` does not pin the report or the API error.** Deleting `reportError` is not caught (mutation M4). Add `expect(reportError).toHaveBeenCalledWith("investment-invoice", expect.anything())` for the rejection and — after Finding 3 — for the `errors[]` payload with the message in it.

4. **The split-funds boundary is unpinned.** `resolveInvestmentFunding`'s `combinedUsd >= totalUsd` survives being tightened to `>` (mutation M1) because every split case uses 524 vs 500. Add a case with `combinedUsd === totalUsd`, which is the investor who converts exactly what they are short.

5. **The invoice-failure reset is unpinned.** Removing `setHasInvoiceFailed(false)` at [transfer-invest-screen.tsx:74](app/screens/card-screen/onboarding/investment-flow/transfer-invest-screen.tsx:74) is not caught (mutation M2). A fail → retry-succeeds sequence asserting the error line is gone before `navigate` fires would pin it.

6. **The locale-parity spec cannot see an untranslated paragraph.** `el` and `hy` still carry the English `CompanyValuation…body2` verbatim ([el.json:3703](app/i18n/raw-i18n/translations/el.json:3703), [hy.json:3703](app/i18n/raw-i18n/translations/hy.json:3703)) while their `body1` is translated; parity passes because the key exists. Fix the two strings, and consider a parity case that flags multi-word values identical to English outside a small loanword allowlist ("Okay", "Deposit", "for").

7. **The funding hook has no self-custodial or tie case.** `use-investment-funding.spec.ts` never runs with a single BTC wallet (the self-custodial shape), where `isSplitAcrossWallets` must be false and `largestWalletUsd` equals `combinedUsd`; and with two equal wallets the `>` in the reduce at [use-investment-funding.ts:50](app/screens/card-screen/onboarding/investment-flow/use-investment-funding.ts:50) names whichever comes first, which nothing states.

8. **The shared-component layout fixes have no pin.** `IconHero` gains `alignSelf: "stretch"` / `width: "100%"` and `FeatureItem` gains `flex: 1` to stop long titles clipping (`cf51ecebd`, `7b1c6a7b3`), affecting 15 pre-existing `IconHero` consumers; `icon-hero.spec.tsx` does not assert the title's style and there is no `feature-item` spec.

9. **`deposit-pending-screen.spec.tsx` tests a screen nothing can reach.** Its four cases pass against a route with no caller (Finding 4); they will keep passing whatever the eventual caller passes it, since it takes no params and reads no state.

10. **`send-confirmation.spec.tsx` is still flaky after `a698c8f0c`.** The commit correctly waits on the fee amount, and at the merge-base that test was the one going red (1 of 4 runs); on the earlier head a different test, "renders the warning when fee quote reports IncreasedToAvoidDust…", went red in 1 of 3 isolated runs. The `hide balance` block's siblings at [send-confirmation.spec.tsx:1324](__tests__/screens/send-confirmation.spec.tsx:1324) and `:1345` still assert straight after `flushEffects()`; the same `waitFor` shape belongs on all of them, in its own PR.

---

## Code Smells

1. **`useInvestmentFunding`'s header doc still describes the sum.** [use-investment-funding.ts:17-19](app/screens/card-screen/onboarding/investment-flow/use-investment-funding.ts:17) — "Both wallets count … the money may sit in either" — contradicts the `useMemo` doc below it, commit `ffa385c9e`, and the `balanceCurrency` the hook now returns. Rewrite to say the fullest wallet decides and the sum only says whether consolidating would do.

2. **`useInvestmentSats` doubles the price subscription to save one multiplication.** `021cafe2d` split the satoshi figure into its own hook so "the screens that merely measure the balance should not pay for a conversion they discard" — but the conversion is one `useMemo`'d multiply, while the split makes the transfer screen call `usePriceConversion()` twice ([use-investment-funding.ts:32](app/screens/card-screen/onboarding/investment-flow/use-investment-funding.ts:32), `:83`), each mounting two `useQuery` hooks on `cache-and-network`. The separation reads well; the stated reason is inverted. Either pass `convertMoneyAmount` in, or say the split is for readability.

3. **`FIGURE_LOCALE`'s doc names the wrong mechanism.** [investment-terms.ts:75-79](app/screens/card-screen/onboarding/investment-flow/investment-terms.ts:75) says "a European phone would print $25.000 here"; what actually happened was `app/i18n/mapping.ts` force-installing the formatjs polyfill with `af` loaded first, so *every* phone printed "$25 000". The pin is right; the comment will send the next reader looking for device-locale handling that does not exist.

4. **Credit-limit vocabulary for an investment amount.** `MOCK_CREDIT_LIMIT_VALUES` ([onboarding-mock-data.ts:41](app/screens/card-screen/onboarding/onboarding-mock-data.ts:41)), `CREDIT_LIMITS`, `selectedLimit`, `limitOption`/`limitText` and the `SelectInvest.desiredCreditLimit` key ("How much would you like to invest?") are all named for the card's credit tiers. Rename to `INVESTMENT_TIERS`/`selectedAmount`/`tierOption`/`desiredAmount` before the strings ship to Transifex under the wrong name.

5. **An unrelated test fix rides on the feature branch with no body.** `a698c8f0c` touches only `send-confirmation.spec.tsx`, has no commit body saying what was flaky or why it is here, and (Gap 10) fixes one of several identical races in that file. Split it out with the sibling tests fixed together.

6. **`formatUsdAmount`'s doc misdescribes the tail it trims.** [investment-terms.ts:85-87](app/screens/card-screen/onboarding/investment-flow/investment-terms.ts:85) says a balance converted from satoshis "carries a tail that would otherwise reach the screen as $3,333.756", but `convertMoneyAmount` rounds to whole cents (`Math.round` in `use-price-conversion.ts:139`), so the only tail is floating-point noise on `/ 100`. The `maximumFractionDigits: 2` is still right; the reason given is not.

7. **`DepositPendingScreen` renders its message with React Native's bare `Text` and a `fontWeight`.** [deposit-pending-screen.tsx:2,50-57](app/screens/card-screen/onboarding/investment-flow/deposit-pending-screen.tsx:50) copies `migration-transferring-funds-screen.tsx`'s pattern, so the line is in the system font rather than the themed `SourceSansPro` every other line in the flow uses. `Text type="p1"` from `@rn-vui/themed` with a `textAlign` is enough — if the screen stays (Finding 4).

8. **The developer-screen bulletins duplicate the i18n title in a string literal.** [developer-screen.tsx:314-339](app/screens/developer-screen/developer-screen.tsx:314) hardcodes "Welcome to become part of Blink" and a body that appears nowhere else. Dev-only, but reading it through `LL.CardFlow.Onboarding.WelcomeInvest.welcomeMessage.title()` keeps the two from drifting, and the "Phase 3: Card Dashboard" trigger is not part of this feature.

9. **`INVOICE_MEMO`'s doc says the payer's name is "missing on purpose" but offers no substitute.** [use-investment-invoice.ts:36-40](app/screens/card-screen/onboarding/investment-flow/use-investment-invoice.ts:36) — the input has an `externalId` field this branch leaves unused (Finding 4); if the memo is deliberately anonymous, the identifier belongs there, and the comment should point at it rather than at "the agreement holds it".

esaugomez31:
@kngako thanks for the review, everything is applied. Here is how each point ended up.

---

## Findings

### Finding 1: Covering the balance while the receive screen is up
Fixed. The auto-close is gated on `useIsFocused`, so it runs once the shortfall screen is back in front. A new spec mounts a real native stack, flips the funding to covered while the receive stub is on top, and asserts the receive screen stays and closing it lands on the transfer step; the existing spec covers the unfocused case.

### Finding 2: Split-funds Convert ends on Home
Resolved in #4279. The conversion flow carries where to return to, the shortfall screen arms an investment return, and the success screen replaces itself with the transfer step, without the drain prefill.

### Finding 3: Invoice step swallows the API's reason
Fixed. `errors[]` messages are reported under `investment-invoice` before returning null, same as a transport rejection. The spec pins both reports. The generated error type carries no `code`, so only the message goes out.

### Finding 4: Signing, identity and the deposit-pending caller
Resolved across the stack. #4234 holds the signing step and produces `settlementSats`, so the transfer copy is true there. #4279 gives the deposit-pending screen its caller from the home card and gates entry on an eligible, resolved account. The invoice's `externalId` now carries the paying account and the amount signed for, stamped per mint since the ledger keeps the id unique per receiving account and never drops an unpaid invoice; nothing is minted while the account is unknown.

### Finding 5: History hygiene
Agreed. The branch will be squash-merged, so only the final tree lands on main.

---

## Test Coverage Gaps

1. Real-stack case added, plus the unfocused one.
2. The pressed option carries `accessibilityState.selected` and `accessibilityRole="radio"`; the spec asserts it moves between options.
3. Both reports pinned, including the `errors[]` message.
4. `combinedUsd === totalUsd` case added.
5. Fail-then-retry case added; the failure line is gone before `navigate` fires.
6. Greek and Armenian translated.
7. Single BTC wallet and equal-wallets tie cases added.
8. `IconHero` and `FeatureItem` title styles pinned in their own specs.
9. Resolved with the caller in #4279.
10. Moved to #4284 with the sibling cases.

---

## Code Smells

1. Funding header rewritten: the fullest wallet decides, the sum only says whether consolidating would be enough.
2. `useInvestmentSats` doc now gives readability as the reason.
3. `FIGURE_LOCALE` doc names the polyfill and its first-loaded default.
4. Renamed in #4234.
5. Moved to #4284.
6. `formatUsdAmount` doc and spec attribute the tail to the shortfall subtraction.
7. Themed `Text` on the deposit-pending screen.
8. Developer-screen title read from i18n.
9. Memo doc points at the external id.

Ready for another look.

grimen (changes requested):
## Findings

**1. The transfer step decides the investment is covered on a wallet the send flow may not pay from.**
`useInvestmentFunding` measures the fullest wallet from `useActiveWallet()`, dollar or bitcoin ([use-investment-funding.ts:45-56](app/screens/card-screen/onboarding/investment-flow/use-investment-funding.ts:45)), and `TransferInvestScreen` then opens `sendBitcoinDestination` with the invoice alone ([transfer-invest-screen.tsx:82](app/screens/card-screen/onboarding/investment-flow/transfer-invest-screen.tsx:82)). The send flow does not know which wallet was judged. `useSendWallets` seeds from the account's `defaultWalletId`, or from the persisted default currency when self-custodial ([use-send-wallets.ts:81](app/screens/send-bitcoin-screen/hooks/use-send-wallets.ts:81), `:99`). When the dollar balance is gated (restricted region or anon self-custodial mode), `restrictToBitcoinWhenBlocked` removes the dollar wallet from the send flow entirely ([use-send-wallets.ts:40-51](app/screens/send-bitcoin-screen/hooks/use-send-wallets.ts:40)). The funding hook applies no such gate, so a dollar balance the send flow will not offer still counts as covering the investment. The step's copy says "You will pay with Bitcoin" ([en.json:3734](app/i18n/raw-i18n/source/en.json:3734)), while the rule accepts the dollar wallet as the one that pays. This is unchanged through #4234 and #4279.

*Example:* An investor in a restricted region holds $30,000 in the dollar wallet and $200 in bitcoin, and picks $25,000. Transfer Invest reads that as covered, mints the invoice and opens the send flow, which offers only the $200 bitcoin wallet and refuses the amount. The shortfall screen and its Convert path never appear, because the flow believes nothing is missing.

*Proposal:* one option is to measure only the wallets `useSendWallets` would offer, and to open the send flow on the wallet that was judged. `sendBitcoinDestination` takes no sending wallet today, so that needs a small param threaded into `createPaymentDetail`. If paying from the dollar wallet is deliberately out of scope, the bitcoin wallet alone should decide, and the dollar balance should go to the Convert branch.

**2. Builds cut from this PR before #4279 become payable, ungated, the moment the deposit-wallet key is set.**
Until #4279 lands, the only safeguard is an empty `cardInvestmentDepositBtcWalletId` ([feature-flags-context.tsx:219](app/config/feature-flags-context.tsx:219)). At this head the entry is deep-linkable by any signed-in user ([deep-link-screens.ts:62](app/navigation/deep-link-screens.ts:62)), and nothing checks eligibility. The transfer step says the agreement was signed, although nothing was signed, and the invoice carries no payer identity. #4279 fixes all of this, but it reads the same key. Remote Config values reach every installed build unless a condition restricts them. So if a release is cut from `main` between the squash-merges of this PR (or #4234) and #4279, setting the key for #4279's release also turns that older build on.

*Example:* A release ships with #3707 merged, a later one ships the full stack, and ops sets the wallet id. A user still on the earlier build opens `blink://card/onboarding/investment` and pays $25,000 into the round. No agreement exists, the invoice carries no account id, and support has only the constant memo to reconcile the payment against.

*Proposal:* one option is for the gated code in #4279 to read a new key, so that builds from this PR see an empty value forever. Another is an app-version condition on the key, documented beside its default. If the plan is to merge the three PRs with no release in between, a note in the PR body saying so would help.

## Test Coverage Gaps

- Nothing ties the wallet judged as covering to the wallet the payment is made from. `transfer-invest-screen.spec.tsx` asserts only `navigate("sendBitcoinDestination", { payment })`, and `use-investment-funding.spec.ts` has no case with the dollar wallet gated, which is where Finding 1 hides.

PR #3702, opened 2026-03-10
feat(card): card flow replace integration

esaugomez31:
## What was done

Implements the business logic and API integration for the Card Order Replacement flow (Phase 1 — Virtual Cards). Replaces mock data with real `cardReplace` mutation, adds dynamic step flow (virtual=2 steps, physical=3), locks card immediately on lost/stolen via `useLockCard`, shows different warning messages depending on the issue type, hides delivery info for virtual cards, and passes real card data to the success screen.

---

### GraphQL Mutation — Completed

- [x] `cardReplace(input: { cardId })` mutation added to `generated.gql`
- [x] `useCardReplaceMutation` auto-generated via codegen
- [x] Returns `{ id, lastFour, cardType, status }` from replaced card

### Translations (i18n) — Completed

- [x] Error messages: `lockFailed`, `replaceFailed`, `replaceFailedCardLocked`
- [x] Confirmation warnings restructured into `DamagedInfo` and `LostStolenInfo` sub-objects with separate bullet messages for each scenario
- [x] Translated to all 28 languages with correct diacritical marks
- [x] Types generated via `typesafe-i18n`

### Shared Types & Layout — Completed

- [x] `EMPTY_ADDRESS` constant added to `card-screen/types.ts` for null address fallback
- [x] `loading` prop added to `SteppedCardLayout` — disables button during mutations

### Hooks — Completed

- [x] **`useLockCard`** — Calls `cardUpdate(LOCKED)`, returns `{ lockCard, loading }`
  - Error handling: GQL errors, null response, network errors — all with toast
  - Pattern follows `useCardFreeze`
- [x] **`useReplaceCard`** — Calls `cardReplace(cardId)`, returns `{ replaceCard, loading }`
  - Returns `{ lastFour, cardType }` on success, `null` on failure
  - Error handling: GQL errors, null response, network errors — all with toast
- [x] **`useReplaceCardFlow`** — Refactored flow state management
  - Moved from `replace-card-screens/` to `replace-card-screens/hooks/`
  - Accepts `{ isVirtualCard, initialAddress }` params
  - Dynamic step order: physical = `[ReportIssue, Delivery, Confirm]`, virtual = `[ReportIssue, Confirm]`
  - Steps changed from numeric (1,2,3) to string enum (`"ReportIssue"`, `"Delivery"`, `"Confirm"`)
  - Exposes `currentStep`, `stepNumber`, `totalSteps`, `stepOrder`
  - `useRegisteredAddress` defaults to `false` when no address exists
[...]

Issue #427, opened 2022-04-12
White text on payment timeout screen


vindard:
## Description

Our usual "white text" issue, this time on payment timeout screen

![image](https://user-images.githubusercontent.com/17693119/162853511-37553ca1-d10c-4798-9e5d-038f71abe3de.png)

daviroo:
This screen will be rewritten during the USD integration work.  I've been trying to ensure that all fonts have an explicit colour while doing this work.  I'll keep the issue open for now but we can probably close it after USD work has been merged.

PR #3709, opened 2026-03-12
feat(card): card flow - order physical card integration

esaugomez31:
## Summary

Replaces all mock data (`MOCK_USER`) in the Order Physical Card flow with real GraphQL mutations and backend integration. Follows the same patterns established in the Replace Card flow (#422 / PR #3702).

### What this PR does

- **`cardCreate` mutation hook**: New `useCreateCard` hook following `useReplaceCard` pattern — handles GQL errors, null responses, and network errors with toast notifications
- **Extended card query**: Added `cardConsumerApplications` to the card query to derive `applicationId` from the first approved application
- **Real shipping address data**: `OrderCardScreen` now uses `useShippingAddressData()` to fetch the user's registered address; supports both registered and custom address entry
- **Form validation**: Added `onValidityChange` callback to `ShippingAddressForm` so the parent screen can disable/enable the continue button based on form validity
- **Shared types**: Moved `CardDesign`, `Delivery`, and their type aliases to shared `types.ts` (used by both Order and Replace flows)
- **Navigation fix**: Fixed `approved-screen.tsx` navigating to `"cardOrderScreen"` instead of `"orderCardScreen"`
- **Error translations**: Added `createFailed` error message in all 28 languages
- **Full test coverage**: 7 test suites updated/created (82 tests), all passing

### Ticket coverage (#449)

This PR covers **all v1 scope** defined in the ticket:

| Acceptance Criteria | Status |
|---|---|
| Access from Card Approved screen | Covered (+ nav bug fix) |
| Access from Card Settings | Already wired in #410 |
| 2-step flow: Shipping → Confirm (+ Success screen) | Covered |
| Back button returns to previous step | Covered |
| Progress indicator updates | Covered |
| Shipping address form with validation | Covered (PO Box, postal code, min lengths) |
| Registered address pre-fill + toggle | Covered |
| Confirm screen with order summary | Covered |
| `cardCreate` mutation call | Covered |
| Success screen with `lastFour` | Covered |
| Error handling (GQL errors, null response, network) | Covered |
| Loading states | Covered |

### Not in scope (per ticket — v2 / separate tickets)

- Eligibility check (hide option if user already has physical card) — Card Settings concern
- Shipping method selection — v1 is standard only
- KYC address pre-fill — blocked on blink-kyc
- Server-side address validation — requires blink-card proxy
- "Add to Wallet" on success — separate ticket #427
- Specific `CardError` extension codes (`CARD_NOT_FOUND`, `SERVICE_UNAVAILABLE`) — requires backend error codes
- Backend changes in blink-card (remove physical card guard, wire shipping, add `method` field)


grimen (commented):
## Code Review: Order Physical Card Integration

Solid PR that cleanly replaces mock data with real backend integration. The `useCreateCard` hook follows established patterns well, error handling is comprehensive, and test coverage is thorough (82 tests).

### Issues Found

#### 1. Missing translations (`bullet4`) — 6 languages still have English text
The following locales have untranslated `bullet4` strings (still English):
- `hr.json` (Croatian)
- `lg.json` (Luganda)
- `sk.json` (Slovak)
- `sr.json` (Serbian)
- `sw.json` (Swahili)
- `xh.json` (Xhosa)

These should be translated or flagged for follow-up before merge.

#### 2. `xh.json` indentation bug
The `errors` block in `xh.json` is at the wrong nesting level compared to the sibling `infoCard` block:
```json
        "bullet4": "Shipping address cannot be changed after order is placed"
      },
    "errors": {    // <-- wrong indent, should match sibling blocks
```

#### 3. Exhaustive switch checks removed — risk of silent failures
In `order-card-screen.tsx`, the `default` cases with `never` exhaustiveness checks were removed from both `getStepConfig()` and `renderStepContent()`. These functions now silently return `undefined` for unhandled step values. If a new step is added later, TypeScript won't flag that these switches need updating. Consider restoring:
```ts
default: {
  const _exhaustive: never = step
  throw new Error(`Unhandled step: ${_exhaustive}`)
}
```

#### 4. Validation logic duplicated in `shipping-address-form.tsx`
The `isValid` computation (useMemo) duplicates the validation rules already defined inline in the form fields (PO Box check, postal code format, min lengths). If validation rules change, they'd need updating in two places. Consider extracting a shared validation utility.

#### 5. `mapToShippingInput` hardcodes `phoneNumber: ""`
In `order-card-screen.tsx`, `mapToShippingInput` always sets `phoneNumber: ""`. If `ShippingAddress` gains a `phoneNumber` field later, this will silently override it. Minor, but worth a comment or using `address.phoneNumber ?? ""`.

#### 6. `loading` prop on `SteppedCardLayout` — verify behavior
`SteppedCardLayout` now receives `loading={createLoading}`. Worth verifying it properly disables the button during submission to prevent double-submit.

### Summary
Main actionable items:
1. **Fix 6 untranslated `bullet4` strings** (or create follow-up tickets)
2. **Fix `xh.json` indentation**
3. **Restore exhaustive switch checks** to catch future step additions at compile time

esaugomez31:
## Review Response — @grimen (review #3946238700)

All 13 items addressed across 5 commits:

### Issues
1. **No error state handling** — Added `useEffect` that shows toast + `goBack()` on `cardError`. `useShippingAddressData` already handles its own error internally.
2. **Late `initialAddress` sync** — Added `useEffect` with `hasInitializedRef` guard: syncs once when data arrives, won't override user edits.
3. **Exhaustive switch** — Changed both switches to `throw new Error(...)` keeping `_exhaustive: never` compile-time check.
4. **`phoneNumber: ""`** — Now sends `me.phone` from account data via `useShippingAddressData`. Still needs a UI input field (not in current design).

### Suggestions
5. **`CardDesign` enum** — No change needed; value is only used as UI display text, never sent to backend.
6. **`registeredAddress` memoization** — Wrapped in `useMemo([initialAddress])`.
7. **`applicationId` null UX** — Merged with #1: toast + `goBack()` if null after loading.
8. **`DEFAULT_COUNTRY`** — Hardcoded `"USA"` explicitly.
9. **Non-Error catch** — Generic `createFailed` toast for non-Error throws.

### Test Gaps
10. **Loading state** — Added 2 tests (cardData loading, addressData loading) + `testID="activity-indicator"`.
11. **`applicationId` null** — Added test: navigates back when null.
12. **`mapToShippingInput`** — Already covered by submit test asserting exact shape.
13. **Query error states** — Added test for card query failure + late `initialAddress` sync tests.

### Additional Fix
- **Double-submission risk** — Added `createLoading` to `isButtonDisabled` on Confirm step to prevent rapid taps firing multiple mutations.

grimen (changes requested):
## Blocking: Address form only supports USA/Canada — needs worldwide support

Rain ships physical cards to 100+ countries. The current implementation hardcodes only USA and Canada in `SUPPORTED_COUNTRIES` (`country-region-data.ts`), making the address form **unusable for international users** — they literally cannot select their country.

### What needs to change

**1. Country list must be comprehensive**
`country-region-data.ts` needs all countries Rain ships to. Consider using an established library (e.g., `i18n-iso-countries`, `country-region-data` npm package) rather than maintaining a manual list.

**2. Region/state should be optional or adapt per country**
Currently region is a required dropdown populated per country. Most countries don't have a standardized state/province list. The region field should:
- Show a dropdown for countries with well-known subdivisions (US states, CA provinces, etc.)
- Fall back to a free-text input for other countries
- Be optional for countries where it's not part of the address format (e.g., many European countries)

**3. Postal code validation must be country-aware (already partially there)**
`postcode-validator` supports 80+ countries, so the validation library is fine. But `getIsoAlpha2()` only maps USA→US and CAN→CA. Once the country list expands, the ISO mapping must cover all entries — or better, store ISO codes directly as country values.

**4. Address field requirements should vary by country**
- `line1 >= 5 chars` is too strict for many international addresses
- Some countries don't use postal codes at all (e.g., many Caribbean nations, some African countries)
- "Region" means different things (state, province, prefecture, canton, county) — labels should adapt
- PO Box regex is English-only — needs French ("Boîte postale", "Case postale"), Spanish ("Apartado"), German ("Postfach") patterns at minimum

**5. `isAddressValid` gate must match actual field validation**
Currently the Continue button gate (`isAddressValid`) doesn't run `postcodeValidator` — it only checks non-empty. A user can proceed with an invalid postal code. The gate should use the same validation logic as the field-level checks.

### What's NOT affected

The **registered address** (from "use registered address" toggle) comes from the backend via KYC and is display-only — no client validation needed there.

### Additional issue: `MOCK_USER.registeredAddress` still referenced

`shipping-step.tsx:48` and `use-order-card-flow.ts:44` still use `MOCK_USER.registeredAddress` — this mock data was supposed to be replaced with real backend data in this PR.

### Suggested approach

- Use a well-maintained country/region library instead of a hand-rolled list
- Define per-country address format configs: which fields are required, field labels, postal code format, whether region is a dropdown or free text
- Keep validation as **common-sense sanity checks** (non-empty required fields, valid postal format where applicable) — not frustrating over-validation. The backend should do the authoritative validation.
- Consider a lightweight config object per country tier:
  - **Tier 1** (US, CA, UK, EU): full validation with postal code format + region dropdown
  - **Tier 2** (rest of world): basic required fields only, free-text region, optional postal code

esaugomez31:
> ## Additional note: `MOCK_USER.registeredAddress` still used in both flows
> 
> `MOCK_USER.registeredAddress` needs to be replaced with real backend data in both the order and replace card flows:
> 
> **In this PR (#3709):**
> 
>     * `order-card-screens/steps/shipping-step.tsx:48`
> 
>     * `order-card-screens/use-order-card-flow.ts:44`
> 
> 
> **Already on `main` from merged PR #3702:**
> 
>     * `replace-card-screens/steps/delivery-step.tsx:52`
> 
>     * `replace-card-screens/use-replace-card-flow.ts:48`
> 
> 
> Since the shared components (`ShippingAddressForm`, `country-region-data.ts`, validation utils) are used by both flows, the international address fixes requested in this PR will automatically apply to the replace flow too. But the `MOCK_USER` references are per-flow and need to be cleaned up in both places.
> 
> Please address the `MOCK_USER` cleanup for **both** flows in this PR since the replace flow (#3702) is already merged with this tech debt.

Reviewed this, MOCK_USER was actually not used in either flow's runtime code. The only references were inside commented-out sections in card-personal-details-screen.tsx (placeholder code for when dateOfBirth and registeredAddress become available via the API). Since it wasn't being consumed anywhere, I removed MOCK_USER, UserInfo, and MOCK_SHIPPING_ADDRESS entirely to avoid confusion. The commented TODOs were updated to reference real variables instead of the mock.

app/components/card-screen/shipping-address-form.tsx
```diff
@@ -1,28 +1,69 @@
...
+export const isAddressValid = (
+  address: ShippingAddress,
+  { checkFullName = true }: { checkFullName?: boolean } = {},
+): boolean => {
+  const hasPOBox =
+    validatePOBox({ value: address.line1, errorMessage: "" }) !== undefined ||
+    (address.line2 !== "" &&
+      validatePOBox({ value: address.line2, errorMessage: "" }) !== undefined)
+
+  const hasInvalidPostal =
```
grimen: Avoid inverted booleans, instead `hasValidPostalCode`

app/components/card-screen/shipping-address-form.tsx
```diff
@@ -1,28 +1,69 @@
...
+    validatePostalCode({
+      value: address.postalCode,
+      countryCode: address.countryCode,
+      errorMessage: "",
+    }) !== undefined
+
+  const postalRequired = postcodeValidatorExistsForCountry(address.countryCode)
```
grimen: `isPostalCodeRequired`

app/components/card-screen/shipping-address-form.tsx
```diff
@@ -1,28 +1,69 @@
...
+  const postalCodeOk = postalRequired
```
grimen: `isPostalCodeValid`

app/components/card-screen/shipping-address-form.tsx
```diff
@@ -1,28 +1,69 @@
...
+} from "@app/utils/country-region-data"
 
 type ShippingAddressFormProps = {
   address: ShippingAddress
   onAddressChange: (address: ShippingAddress) => void
+  onValidityChange?: (isValid: boolean) => void
   showFullName?: boolean
 }
 
+export const isAddressValid = (
```
grimen: Move to generic utils library, this validation can be useful in more places potentially.

app/components/card-screen/shipping-address-form.tsx
esaugomez31: Done!

app/components/card-screen/shipping-address-form.tsx
esaugomez31: Done!

app/components/card-screen/shipping-address-form.tsx
esaugomez31: Done!

app/components/card-screen/shipping-address-form.tsx
esaugomez31: Done!

app/components/card-screen/shipping-address-form.tsx
grimen, line 1: Address validator should be able to flag which fields that fail, not just if it failed something or not and show global error. This is basic UX that needs to be right. In other words `validateAddress` method that returns affectd fields, see https://ajv.js.org/ for reference.

esaugomez31: Done! Now validateAddress returns which fields failed and why, instead of just a boolean. The form only imports that one function and passes errors directly to each field

esaugomez31:
@grimen Done! Implemented the country-aware validation using Google's libaddressinput metadata as you described. Replaced postcode-validator and country-region-data with a single static JSON bundled at build time. buildAddressSchema + validateAddress follow the AJV-inspired pattern — schema in, per-field errors out.

PR #3712, opened 2026-03-15
fix(send): dismiss payment completed screen when app is backgrounded

designsats:
## Problem

When a user pays in-store and then leaves the app, the `SendBitcoinCompletedScreen` persists indefinitely. Returning to the app later shows a stale confirmation screen with no way to know it's from a previous session.

Closes https://github.com/blinkbitcoin/blink-wip/issues/554

## Solution

An `AppState` listener in `SendBitcoinCompletedScreen` dismisses the receipt when the user returns from a stale background trip (30 seconds or longer). Short trips away — opening the success-action link, locking the phone, sharing, a quick app switch — keep the receipt intact; coming back later lands the user on Home instead of a stale receipt.

## Implementation

- `useOnStaleBackgroundReturn` (hook local to the screen): records `Date.now()` when the app goes to background and, on return to `active`, dismisses only if the trip lasted `STALE_BACKGROUND_MS` (30s). Working from the `background`/`active` events alone covers both platforms — iOS backgrounds through `inactive`, Android straight from `active` — with no previous-state tracking.
- The subscription is created once and reads the callback through a ref (same invariant as the global `AppStateWrapper`), so it never tears down mid-transition.
- Dismissal is guarded by `navigation.isFocused()`: a covered receipt (e.g. under the app-lock screen after a resume) cannot steal navigation from the screen on top.
- Dismissal uses `popToTop()` — the same pattern as `conversion-success-screen`. Under react-navigation v7, `navigate("Primary")` would push a duplicate Home instead of popping the receipt off the stack. The close button shares this path.

## Tests

22 tests in `__tests__/screens/send-bitcoin-completed-screen.spec.tsx`, 100% line coverage of the screen:

- Dismisses via `popToTop` on return from a stale (≥30s) background trip, including the iOS path through `inactive`
- Keeps the receipt on quick trips, foreground-only interruptions, and when the screen is covered by another screen
- No dismissal when the screen mounts while the app is already backgrounded
- Close button dismisses through the same path
- AppState subscription is removed on unmount (the mock implements real listener removal and rejects unknown event names)

## How to verify

1. Complete a payment and let the success screen settle into the receipt
2. Background the app and return within a few seconds — the receipt is still there
3. Background the app for 30+ seconds and return — you land on Home
4. The close button still returns Home

### DIFF
--- a/__tests__/navigation/app-state.spec.tsx
+++ b/__tests__/navigation/app-state.spec.tsx
@@ -58,6 +58,10 @@ const mockedKeyStore = jest.mocked(KeyStoreWrapper)
 /** Drives the real AppState listener the wrapper subscribes with. */
 let emitAppState: (state: AppStateStatus) => Promise<void>
 
+/** The same listener, left mid-flight so that what the wrapper does synchronously can be
+ *  told apart from what it does once the keystore has answered. */
+let emitAppStateWithoutSettling: (state: AppStateStatus) => void
+
 const START_TIME_MS = 1_790_000_000_000
 const GRACE_PERIOD_SECONDS = 60
 
@@ -100,8 +104,9 @@ describe("AppStateWrapper", () => {
       if (type === "change") listener = handler as (state: AppStateStatus) => void
       return { remove: jest.fn() } as ReturnType<typeof AppState.addEventListener>
     })
+    emitAppStateWithoutSettling = (state) => listener(state)
     emitAppState = async (state) => {
-      listener(state)
+      emitAppStateWithoutSettling(state)
       await flushEffects()
     }
   })
@@ -175,6 +180,29 @@ describe("AppStateWrapper", () => {
       await emitAppState("background")
       jest.setSystemTime(START_TIME_MS - 60 * 60 * 1000)
       await emitAppState("active")
+
+      expect(mockSetAppLocked).toHaveBeenCalledTimes(1)
+      expect(mockNavigate).toHaveBeenCalledWith("authenticationCheck", { isResume: true })
+    })
+
+    it("raises the lock only once the keystore has answered, never in the same tick", async () => {
+      /** Screens that dismiss themselves on a stale return, the send receipt among them,
+       *  run from their own synchronous AppState listener and rely on still being focused
+       *  when they do. Locking in this tick would cover them first, so the await before the
+       *  navigate is a contract those screens depend on rather than an implementation
+       *  detail free to change. */
+      renderWrapper()
+      await flushEffects()
+
+      await emitAppState("background")
+      jest.setSystemTime(START_TIME_MS + (GRACE_PERIOD_SECONDS + 1) * 1000)
+
+      emitAppStateWithoutSettling("active")
+
+      expect(mockSetAppLocked).not.toHaveBeenCalled()
+      expect(mockNavigate).not.toHaveBeenCalled()
+
+      await flushEffects()
 
       expect(mockSetAppLocked).toHaveBeenCalledTimes(1)
       expect(mockNavigate).toHaveBeenCalledWith("authenticationCheck", { isResume: true })
--- a/__tests__/screens/send-bitcoin-completed-screen.spec.tsx
+++ b/__tests__/screens/send-bitcoin-completed-screen.spec.tsx
@@ -51,7 +51,12 @@ jest.mock("react-native-view-shot", () => {
 })
 
 const mockNavigate = jest.fn()
-const mockNavigation = { navigate: mockNavigate, popToTop: jest.fn() }
+const mockIsFocused = jest.fn(() => true)
+const mockNavigation = {
+  navigate: mockNavigate,
+  popToTop: jest.fn(),
+  isFocused: mockIsFocused,
+}
 jest.mock("@react-navigation/native", () => {
   const actual = jest.requireActual("@react-navigation/native")
   return {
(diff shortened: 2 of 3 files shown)


grimen (changes requested):
Add test coverage to cover the change.

designsats:
@blink-claw-bot review please

esaugomez31:
Rebased onto `main` (resolved a conflict with the removed in-app-review code) and pushed a series of fixes on top of the original commit. Branch is current with `main`; 22 tests, typecheck and lint green.

**@blink-claw-bot's finding is fixed** — dismissal no longer depends on a direct `active → background` transition, so the standard iOS backgrounding path (`active → inactive → background`) now works, and the test that pinned the old behavior was replaced.

Also addressed while reviewing:

1. **Focus guard.** The mounted-but-covered receipt reacted to backgrounding from *any* screen stacked above it (including the app-lock prompt after a resume) and hijacked navigation. Dismissal now only runs when the receipt is focused (`navigation.isFocused()`).
2. **`popToTop()` instead of `navigate("Primary")`.** Under react-navigation v7, `navigate` pushes a duplicate `Primary` instead of popping back, so the receipt was never actually unmounted. `popToTop()` really removes it — same pattern as `conversion-success-screen`. The close button shares this path.
3. **Staleness threshold.** Dismissal now happens on *return* from background, and only if the trip lasted 30s or more. Quick exits caused by the screen itself — opening the LNURL success-action link, sharing on Android, locking the phone — no longer destroy the receipt, while the issue's actual scenario (coming back later) still lands on Home.
4. **Spec hardening.** The AppState mock now implements real listener removal and rejects unknown event names; added coverage for the receipt phase, the covered-screen case, the close button, and a behavioral unmount test. 100% line coverage of the screen.

Commits: b82a7df32, 02ef7e30c, 31125f31c, 112554ca2.

esaugomez31 (dismissed):
Works as expected!!

grimen (changes requested):
## Findings

**1. A skipped stale return is dropped rather than deferred (blocking).**
`handleStaleReturn` skips dismissal when the receipt is not focused, and nothing records that a stale return happened, so a covered receipt outlives the trip permanently — the same stale-receipt bug the PR fixes.

*Example:* from the receipt the user taps a push notification (which pushes another screen onto the root stack), backgrounds the app for an hour, returns, then backs out of that screen — the stale receipt from the previous session is still there.

*Proposal:* one option is to record the skipped dismissal in a ref and pop on the screen's next `focus` event, instead of dropping it.

## Test Coverage Gaps

- A quick trip followed by a stale trip is untested — nothing verifies that resetting `backgroundedAt` on the quick return doesn't suppress a later dismissal.
- A stale return during the success-icon phase is untested; the hook is live then too and would `popToTop` mid-animation.
- The ordering contract documented on `handleStaleReturn` (receipt pops before the resume lock is raised because the relock awaits the keystore) is enforced by no test on either side; a refactor in `app-state.tsx` would break it silently.

esaugomez31:
Rebased onto `main` (resolved a conflict in `__tests__/screens/send-bitcoin-completed-screen.spec.tsx` against the timezone tests that landed in #4065) and pushed the review fixes on top. Branch is current with `main`; 47 tests across the two touched specs, 100% line/branch/function coverage on the screen, full suite green (562 suites, 6414 tests), typecheck and lint clean.

**@grimen's blocking finding is fixed.** A stale return that arrives while the receipt is covered is no longer dropped. It is recorded in a ref and spent on the screen's next `focus` event, as proposed. The focus guard still holds, so a covered receipt never tears the app-lock screen off the stack; the pop simply waits until whatever covered it is gone. The push-notification case from the report now lands on Home instead of stranding the receipt for the rest of the session.

The three coverage gaps are closed:

1. **Quick trip followed by a stale one.** Pins that clearing `backgroundedAt` on the quick return does not suppress the later dismissal, and that the second trip is measured from its own departure rather than the first.
2. **Stale return during the success icon phase.** The clock is moved with `jest.setSystemTime` instead of `advanceTimersByTime`, so the icon timer stays unfired and the trip is genuinely taken from the icon phase rather than after it.
3. **The relock ordering contract**, now pinned on both sides. The receipt asserts it pops from inside the AppState event; `app-state.spec.tsx` asserts the relock raises nothing in that same tick and only navigates once the keystore has answered. A refactor on either side now breaks a test instead of failing silently.

One thing worth flagging from the rebase: besides the marked conflict, the auto-merge silently duplicated `mockNavigate` and the `@react-navigation/native` mock, and left two close-button tests, one of which still asserted `navigate("Primary")` after the branch had moved to `popToTop()`. Both are reconciled.

Commits: e524affb4, 96c69ba21.

esaugomez31:
@grimen 
## Before / After

Same scenario in both runs, with the branch as the only variable: complete a payment, let the success screen settle into the receipt, press home, wait 35 seconds, return to the app. The dismissal threshold is 30 seconds, so 35 clears it.

### Before (`main`)

Returning to the app still shows the receipt from the previous session. This is the stale confirmation screen described in the issue.

https://github.com/user-attachments/assets/81ef4a48-3e6d-4c43-91c8-f42798e27de9

### After (this branch)

The same trip now lands on Home, and the stale receipt is gone.

https://github.com/user-attachments/assets/f4c97e11-97bf-47f5-b770-7c8b2104adde

grimen (approved):

PR #3717, opened 2026-03-17
feat: home add card account

esaugomez31:
## What was done

Adds a **Card** account row to the Home screen's "My Accounts" section, alongside the existing Bitcoin and Dollar rows.

### New behavior
- Card row is shown when the user has a card with status `ACTIVE` or `LOCKED`
- Displays card balance: sats on top (large), display currency below (small) — same format as BTC/USD rows
- Card balance is included in the total balance shown in the header
- Tapping the Card row navigates to `cardDashboardScreen`
- Respects the hide/show amounts toggle (`****` placeholder when hidden)
- No notification badge (out of scope for Phase 1)

### Card pill
- Gray pill (`grey4` background, white text) labeled "Card" — distinct from BTC (blue) and USD (green)
- Verified color against Figma design
- All three pills equalize width via `useEqualPillWidth`

### Files modified
| File | Change |
|------|--------|
| `app/i18n/en/index.ts` + 28 JSON files | Added `common.card` translation key |
| `app/types/amounts.ts` | Added `CARD` constant and `CardCurrency` type |
| `app/components/atomic/currency-pill/currency-pill.tsx` | Added `CARD` case with gray styling |
| `app/components/atomic/currency-pill/use-equal-pill-width.ts` | Extended `PillKey` to include `CARD` |
| `app/components/balance-header/use-total-balance.ts` | Added optional `cardBalanceSats` param to include card in total |
| `app/components/wallet-overview/wallet-overview.tsx` | Added Card row (visibility, balance display, navigation, hide amounts) |
| `app/screens/home-screen/home-screen.tsx` | Fetches card data, passes it down to `WalletOverview` and `useTotalBalance` |

### Refactor included
- Moved `useCardData` and `useCardBalance` from `app/screens/card-screen/` to `app/hooks/` since they are now shared across the home screen and the card flow
- All existing consumers updated to import from `@app/hooks`

### DIFF
--- a/__tests__/screens/card-screen/order-card-screens/order-card-screen.spec.tsx
+++ b/__tests__/screens/card-screen/order-card-screens/order-card-screen.spec.tsx
@@ -73,7 +73,7 @@ const mockAddress = {
 }
 
 const mockUseCardData = jest.fn()
-jest.mock("@app/screens/card-screen/hooks", () => ({
+jest.mock("@app/hooks/use-card-data", () => ({
   useCardData: () => mockUseCardData(),
 }))
 
--- a/__tests__/screens/card-screen/pin-screens/hooks/use-card-pin-update.spec.ts
+++ b/__tests__/screens/card-screen/pin-screens/hooks/use-card-pin-update.spec.ts
@@ -5,13 +5,17 @@ import { useCardPinUpdate } from "@app/screens/card-screen/pin-screens/hooks/use
 
 const mockMutate = jest.fn()
 jest.mock("@app/graphql/generated", () => ({
+  ...jest.requireActual("@app/graphql/generated"),
   useCardPinUpdateMutation: () => [mockMutate, { loading: false }],
 }))
 
 let mockCard: { id: string } | undefined = { id: "card-123" }
 const mockFetchPublicKey = jest.fn()
-jest.mock("@app/screens/card-screen/hooks", () => ({
+jest.mock("@app/hooks/use-card-data", () => ({
   useCardData: () => ({ card: mockCard }),
+}))
+
+jest.mock("@app/screens/card-screen/hooks", () => ({
   useCardEncryption: () => ({ fetchPublicKey: mockFetchPublicKey, loading: false }),
 }))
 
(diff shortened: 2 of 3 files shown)


app/components/atomic/currency-pill/currency-pill.tsx
```diff
@@ -22,7 +23,7 @@ export const CurrencyPill = ({
   containerStyle,
   onLayout,
 }: {
-  currency?: WalletCurrency | "ALL"
+  currency?: WalletCurrency | "ALL" | CardCurrency
```
grimen: nitpick: `CardCurrency | WalletCurrency | "ALL"` more natural order of type.

app/components/atomic/currency-pill/use-equal-pill-width.ts
grimen, line 19: `CARD` constant is imported, but `BTC` and `USD` is not. Consistency needed here I think.

grimen (changes requested):

app/components/atomic/currency-pill/currency-pill.tsx
esaugomez31: Reordered to `CardCurrency | WalletCurrency | "ALL"`.

app/components/atomic/currency-pill/use-equal-pill-width.ts
esaugomez31: Declared `CARD` as a local constant with `CardCurrency` type, consistent with `BTC` and `USD`.

app/screens/home-screen/home-screen.tsx
```diff
@@ -530,6 +546,9 @@ export const HomeScreen: React.FC = () => {
           wallets={wallets}
           showBtcNotification={isOutgoing ? false : hasUnseenBtcTx}
           showUsdNotification={isOutgoing ? false : hasUnseenUsdTx}
+          showCardRow={showCardRow}
+          cardBalancePrimary={cardBalancePrimary}
+          cardBalanceSecondary={cardBalanceSecondary}
         />
```
grimen, line 554: Per discussion on call: Unify the interface where card is just another wallet.

esaugomez31: Done! Unified card as another account in the same array. Instead of passing card as separate props, home-screen now builds an `accounts: AccountBalance[]` array that includes BTC, USD, and optionally Card. WalletOverview receives this single array and handles formatting internally.

```suggestion
        <WalletOverview
          loading={loading}
          setIsStablesatModalVisible={setIsStablesatModalVisible}
          accounts={accounts}
          showBtcNotification={isOutgoing ? false : hasUnseenBtcTx}
          showUsdNotification={isOutgoing ? false : hasUnseenUsdTx}
        />
```

PR #3720, opened 2026-03-18
fix: send receive flows v36

esaugomez31:
## Summary

Fixes and improvements for send/receive flows found during v36 release testing ([#575](https://github.com/blinkbitcoin/blink-wip/issues/575))

**Bugs**
- **Phone number sanitization in send flow** — Strip all non-numeric characters (spaces, dashes, parentheses) from phone input on type and paste. Added `sanitizePhoneNumber` utility that preserves the `+` country prefix when present.
- **Paste buttons always active** — Both paste buttons (search input and phone input) now work regardless of which input is currently active. Previously each was disabled when its input wasn't focused.
- **CurrencyPill size mismatch with Dollar default** — Pill was smaller than expected when Dollar account was the default because `useEqualPillWidth` only measured via `onLayout` (which requires both pills to render). Now passes label text to the hook so it can pre-measure both pill widths upfront.
- **Expiration time modal colliding with iOS status bar** — Replaced React Native's `SafeAreaView` with `useSafeAreaInsets` from `react-native-safe-area-context` in the expiration time modal. Applied only on iOS to avoid affecting Android layout.
- **Expiration time shows "300 sec" instead of "5 min" on iOS** — Added `@formatjs/intl-numberformat` polyfill to fix `Intl.NumberFormat` with `style: "unit"` on Hermes iOS. The native iOS implementation (`NSMeasurementFormatter`) converts to the base unit (seconds), so "5 minutes" became "300 sec". The polyfill replaces the native implementation with a JavaScript-based one using CLDR data that formats correctly.

**Fixes**
- **hitSlop on currency toggle button** — Added `hitSlop={8}` to the `onToggleCurrency` pressable in the amount input screen for a larger tap target.

**Features**
- **Lightning address copy toast** — Changed success notification from "Paycode/LNURL has been copied in the clipboard" to "Lightning address has been copied in the clipboard". Updated all 28 translation files.

### DIFF
--- a/app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
+++ b/app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
@@ -884,11 +884,12 @@ const PhoneInputSection: React.FC<PhoneInputSectionProps> = ({
     setKeepCountryCode(false)
 
     try {
-      const clipboard = sanitizePhoneNumber(await Clipboard.getString())
+      const clipboardValue = await Clipboard.getString()
+      const clipboardPhoneNumber = sanitizePhoneNumber(clipboardValue)
 
-      let parsed = null
-      parsed = parseValidPhone(clipboard)
-      const parseNumber = parsed && parsed?.isValid() ? parsed.number : clipboard
+      const parsed = parseValidPhone(clipboardPhoneNumber)
+      const parseNumber =
+        parsed && parsed?.isValid() ? parsed.number : clipboardPhoneNumber
 
       updateMatchingContacts(parseNumber)
       dispatchDestinationStateAction({


app/screens/send-bitcoin-screen/send-bitcoin-destination-screen.tsx
```diff
@@ -883,7 +884,7 @@ const PhoneInputSection: React.FC<PhoneInputSectionProps> = ({
     setKeepCountryCode(false)
 
     try {
-      const clipboard = await Clipboard.getString()
+      const clipboard = sanitizePhoneNumber(await Clipboard.getString())
```
grimen: This is a bit confusing, for clarity something like this would be more clear:

```
const clipboardValue = await Clipboard.getString()
const clipboardPhoneNumber = sanitizePhoneNumber(clipboardValue)
```

esaugomez31: Applied both suggestions: renamed to `clipboardValue` / `clipboardPhoneNumber` for clarity, and changed `let parsed = null` + reassignment to a single `const parsed` declaration.

PR #3721, opened 2026-03-18
fix: notification settings label

esaugomez31:
## Summary

Fixes the notification settings label in the preferences screen that was hardcoded to always show "Some".
image

The label now reflects the actual notification state by querying `disabledCategories` from the notification settings:

- **"All"** — all notification categories are enabled (no disabled categories)
   image

- **"Some"** — at least one category is enabled but not all
   image
- **"None"** — all notification categories are disabled
   image

Added `statusAll`, `statusNone`, and `statusSome` i18n keys with translations for 28 languages.

### DIFF
--- a/app/i18n/raw-i18n/translations/ar.json
+++ b/app/i18n/raw-i18n/translations/ar.json
@@ -2358,19 +2358,19 @@
     "displayCurrency": "عملة العرض"
   },
   "NotificationSettingsScreen": {
-    "title": "Notification Settings",
-    "pushNotifications": "Push Notifications",
+    "title": "إعدادات الإشعارات",
+    "pushNotifications": "إشعارات الدفع",
     "statusAll": "الكل",
     "statusNone": "لا شيء",
     "statusSome": "بعض",
     "notificationCategories": {
       "Circles": {
-        "title": "Circles",
-        "description": "Notifications about your circles."
+        "title": "الدوائر",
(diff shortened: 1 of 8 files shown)


app/i18n/raw-i18n/translations/qu.json
```diff
@@ -2360,6 +2360,9 @@
   "NotificationSettingsScreen": {
     "title": "Notification Settings",
     "pushNotifications": "Push Notifications",
+    "statusAll": "All",
+    "statusNone": "None",
+    "statusSome": "Some",
```
grimen: English translations does not make sense here.

app/i18n/raw-i18n/translations/ar.json
grimen, line 2374: BTW not caused by this PR, but some translations have random English translations for some keys - please review while on it.

app/i18n/raw-i18n/translations/qu.json
esaugomez31: Translated to Quechua: `Lliw` (All), `Mana ima` (None), `Wakin` (Some).

app/i18n/raw-i18n/translations/ar.json
esaugomez31: Translated the missing strings in `NotificationSettingsScreen` across all affected languages: ar, cs, de, hy, sr, th, vi, and qu.

PR #3729, opened 2026-03-21
feat(card): integrate card holder and secrets details

esaugomez31:
## Summary

Integrates `cardHolder` and `cardSecretsEncrypted` GraphQL queries into the card flow, replacing hardcoded/empty values with real data from the API.

### Changes

- **Card holder name**: `BlinkCard` component now resolves the holder name internally via `useCardHolder` hook when receiving a `cardId` prop — eliminates duplicate hook calls across screens
- **Card secrets (PAN/CVC)**: Card details screen fetches and decrypts PAN and CVC using the encrypted secrets endpoint, displaying the full card number and CVV with copy-to-clipboard support
- **Holder name display**: Card visual shows the name in uppercase (matching physical card convention), while the info field shows it in original format
- **Route params cleanup**: Replaced `holderName` string param with `cardId` across all navigation routes — screens no longer pass display data through navigation, they resolve it from the API
- **AES-GCM decryption**: Added `decryptAesGcm` utility to complement the existing `encryptAesGcm`

### Screens updated

| Screen | Change |
| --- | --- |
| Card details | Shows full PAN, CVC, and cardholder name from API |
| Card dashboard | Shows holder name on card visual |
| Card status (approved/ordered) | Shows holder name on card visual |
| Add to mobile wallet | Shows holder name on card visual |
| Order card / Replace card | Passes `cardId` instead of `holderName` to status screen |

### Pending

- **Expiry date**: Not yet available in the backend API — field remains empty until the endpoint is implemented

### Tests

All new hooks, utilities, and screen modifications are covered by tests.

### DIFF
--- a/app/screens/card-screen/card-details-screen.tsx
+++ b/app/screens/card-screen/card-details-screen.tsx
@@ -65,17 +65,11 @@ export const CardDetailsScreen: React.FC = () => {
     }
   }, [authenticated, card?.id, secrets, fetchSecrets])
 
-  const copyAction = useCallback(
-    (value: string | undefined, label: string): (() => void) | undefined => {
-      if (!value) return undefined
-      return () =>
-        copyToClipboard({
-          content: value,
-          message: LL.common.hasBeenCopiedToClipboard({ type: label }),
-        })
-    },
-    [copyToClipboard, LL],
-  )
+  const handleCopy = (content: string, label: string) =>
+    copyToClipboard({
+      content,
+      message: LL.common.hasBeenCopiedToClipboard({ type: label }),
+    })
 
   const handleSettingsPress = useCallback(() => {
     navigation.navigate("cardSettingsScreen")
@@ -153,7 +147,7 @@ export const CardDetailsScreen: React.FC = () => {
             value={cardNumberFormatted}
             icon="copy-paste"
             testID="card-number-field"
-            onAction={copyAction(cardNumber, LL.CardFlow.CardDetails.cardNumber())}
+            onAction={() => handleCopy(cardNumber, LL.CardFlow.CardDetails.cardNumber())}
           />
 
           <View style={styles.rowFields}>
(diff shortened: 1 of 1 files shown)


__tests__/utils/crypto.spec.ts
grimen, line 1: Utility libraries unit tests should test actual values rather than mocked, looks like too much magic here?

app/screens/card-screen/card-details-screen.tsx
```diff
@@ -37,19 +38,29 @@ export const CardDetailsScreen: React.FC = () => {
...
-        content,
-        message: LL.common.hasBeenCopiedToClipboard({ type: label }),
-      })
+  useEffect(() => {
+    if (authenticated && card?.id && !secrets) {
+      fetchSecrets(card.id)
+    }
+  }, [authenticated, card?.id, secrets, fetchSecrets])
+
+  const copyAction = useCallback(
```
grimen: Seems very universal, so would make sense to have a dedicated hook `useCopyToClipboard`.

app/screens/card-screen/card-details-screen.tsx
grimen, line 1: We should discuss where this should go, but we need enable the biometric authentication for showing sensitive/encrypted details. Let's discuss on a call.

__tests__/utils/crypto.spec.ts
esaugomez31: Done. Rewrote tests using real `node:crypto` as backend, no more mocked cipher/decipher. Tests now do actual encrypt then decrypt roundtrips verifying real values (simple strings, unicode, card numbers, CVCs, wrong key rejection, tampered data detection).

app/screens/card-screen/card-details-screen.tsx
esaugomez31: `useClipboard` already exists as the universal clipboard hook. Instead of creating a duplicate hook, simplified `copyAction` to a plain `handleCopy` function that calls `copyToClipboard` directly. `ActionField` already handles the `undefined` case internally (disables action when value is null), so no wrapper needed.

app/screens/card-screen/card-details-screen.tsx
esaugomez31: Done. Biometric authentication was already implemented but was set as not required, so it allowed access without a biometric sensor. Now set `required: true` following the same pattern as `card-change-pin-screen`. When authentication fails or no sensor is available, a toast message is shown so the user knows why they cannot access the screen, then navigates back. Also handles the deeplink edge case: if no navigation history exists, resets to home instead of `goBack()`.

Issue #4, opened 2021-02-01
MISSING_INSTANCEID_SERVICE issue


nicolasburtey:
![WhatsApp Image 2021-02-01 at 12 57 51 PM](https://user-images.githubusercontent.com/11156370/106512550-60b13e80-6497-11eb-8f76-aa7ec94bec2c.jpeg)

(some quick  research leads to google cloud messaging: https://github.com/oney/react-native-gcm-android/issues/21)

Issue #11, opened 2020-10-17
error if amount is too low for sendOnChain


nicolasburtey:
lnService.getChainFeeEstimate will return : `transaction output is dust`
manage this properly from the app

PR #3731, opened 2026-03-24
feat(spark): spark onboarding backup screens UI

esaugomez31:
## Summary

Visual components and layout for the Spark self-custodial onboarding backup flow ([#572](https://github.com/blinkbitcoin/blink-wip/issues/572)).

### Screens implemented

| Screen | Description |
| --- | --- |
| **Backup Method** | Choose between Google Drive/Apple iCloud, Password Manager, or Manual backup — provider name adapts per platform |
| **Cloud Backup** | Optional encryption with password (12+ chars), PBKDF2 key derivation + AES-GCM, upload to Google Drive appDataFolder with overwrite confirmation |
| **Security Alerts** | Three sequential checkboxes with fade-in animation — all must be checked to proceed |
| **Backup Phrase (Step 1)** | Display words 1-6 in two groups of 3, with copy-to-clipboard, countdown timer, and Spark-compatible wallet link |
| **Backup Phrase (Step 2)** | Display words 7-12 in two groups of 3, with I have saved it button navigating to confirmation |
| **Confirm Phrase** | Verify 3 random words with BIP39 autocomplete suggestions above keyboard, auto-navigate on all correct, error state with red border |
| **Success** | Animated success screen that auto-navigates to home after animation completes |

### Cloud Backup (Google Drive)

- Uses @react-native-google-signin/google-signin with drive.appdata scope
- Google Sign-In with account picker, checks for existing backup, optional overwrite confirmation alert
- PATCH existing file or POST new file via Drive REST API v3 multipart upload
- Optional encryption: PBKDF2 (600k iterations, SHA-256) + AES-128-GCM
- Backup format: JSON with version, encrypted, createdAt fields for retrocompatibility
- Network-based filenames: blink-spark-backup-{network}.json
- Password validation: 12+ char minimum, confirm match, reserved error space in UI
- Passwords cleared on screen blur via useFocusEffect

### Password Manager Backup (Keychain)

- Uses react-native-keychain via useKeychainBackup hook
- Network-based service name: blink-spark-backup-{network}
- Success toast + navigate to success screen

### Backup Method Selection

- Subtitle adapts per platform using getCloudProviderName(LL) utility
- Primary CTA: Google Drive / Apple iCloud (dynamic per platform)
- iOS cloud backup returns early (not yet implemented)
- Secondary options: Password manager, Manual backup

### Security Checkboxes (Alerts)

- Three checkboxes revealed one at a time with fade-in animation
- Continue button disabled until all three are checked

### Backup Phrase Display (2 steps)

- Split into 2 screens per design review: words 1-6 (step 1) and 7-12 (step 2)
- PhraseStep enum (First / Second) for type-safe navigation
- Step 1: countdown timer then Continue. Step 2: I have saved it then confirm
- Copy button with toast confirmation and 60s auto-clear
- Logic extracted to useBackupPhrase hook

### Backup Phrase Confirmation

- BIP39 autocomplete via SuggestionBar positioned above keyboard
- Word number as Text outside input, green/red border states
- Auto-navigates after all 3 words validated (400ms delay)
- keyboardType visible-password for Android soft keyboard
- Logic extracted to useBackupConfirm hook

### Reusable components

- **SuccessScreenLayout** — Reusable success screen with animations
- **IconHero** — Icon in circle + title + optional subtitle
- **InfoBanner** — Left-bordered info container
- **SuggestionBar** — Horizontal chip bar with keyboard-aware positioning
- **PasswordInput** — Password field with eye toggle and error display
- **CheckboxRow** — Added centered prop for layout flexibility

### Hooks architecture

| Hook | Location | Responsibility |
| --- | --- | --- |
| useGoogleDriveBackup | app/hooks/ | Google Sign-In + Drive REST API |
| useKeychainBackup | app/hooks/ | Keychain save |
| useBackupMethods | spark-onboarding/hooks/ | Backup method orchestration |
| useBackupPhrase | spark-onboarding/hooks/ | Phrase display, copy, countdown |
| useBackupConfirm | spark-onboarding/hooks/ | Word validation, auto-navigate |
| useCloudBackupForm | spark-onboarding/hooks/ | Password form state and validation |
| useCloudBackup | spark-onboarding/hooks/ | Encrypt + upload + overwrite confirm |

### Translations

- 28 languages with diacritical marks
- CloudBackup namespace with 20 keys

### Test coverage

- 14 new test files covering all hooks, components, and screens

### Pending

- iOS iCloud backup (button visible but no-op on iOS)
- Real seed phrase generation via Spark SDK (currently mock data)
- Google Drive API access pending GCP configuration
- Wallet restoration flow


esaugomez31:
* **#3731** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3731?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3731?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):
Nice work — clean hook separation, strong test coverage, thorough i18n. The structure is mostly right, but I'd like to see a few things tightened before this merges. Two reasons:

1. **Format lock-in.** The PR description notes Drive access is still pending GCP config and the seed is still mocked — so no real user has a backup yet. That's the window to get the on-disk format right; after that, every fix is a migration.
2. **This PR sets templates.** It's the first user-chosen password in the codebase, the first cloud-backup adapter, and the first Google Drive integration. Whatever lands here will be copy-pasted. Worth investing in the shape now.

Most of the feedback is refactoring that extracts pure logic out of hooks so it's testable without React — which is also how the format round-trip test (the thing that catches the IV bug) becomes trivial to write.

## 🔴 Blocking

**1. AES-GCM IV should be 12 bytes, not 16** — `app/utils/crypto.ts:48`
GCM is specified for 96-bit IVs. Other lengths trigger GHASH preprocessing and hurt interop. Change `Crypto.randomBytes(16)` → `Crypto.randomBytes(12)`.

**2. Backup JSON must persist crypto parameters.**
Only `version: 1` is stored; `iterations`, `digest`, `keyLen`, and `cipher` are hard-coded constants. Any future change (including fixing #1 or moving to AES-256) breaks restore. Persist them now:
```json
{
  "version": 1,
  "kdf": { "name": "PBKDF2", "iterations": 600000, "digest": "SHA-256", "keyLen": 16 },
  "cipher": "AES-128-GCM",
  ...
}
```
With this in place, most Medium items below become safely migratable in follow-ups.

**3. Extract the backup format into `app/utils/spark-backup-format.ts`.**
Currently inline as an IIFE in `use-cloud-backup.ts`:
```ts
const payload = isEncrypted ? (() => { ... })() : JSON.stringify({...})
```
Pull this out into a pure module:
```ts
// spark-backup-format.ts
export type BackupPayload = /* discriminated union: encrypted | plain */
export const buildBackupPayload = (mnemonic, opts) => ...
export const parseBackupPayload = (raw: string) => ...   // for restore
```
This is the natural home for #2's schema, lets you write format round-trip tests without React, and gives the restore PR a ready-made parser. Single biggest structural win in the PR.

**4. `findExistingFileId` silently returns `undefined` on non-OK** — `use-google-drive-backup.ts:36`
A 401/5xx is treated as "no existing file" → the next step **creates a duplicate backup** with no user signal. Throw and surface the error.

**5. Add `decryptAesGcm` + a crypto round-trip test.**
Restore screens are out of scope and I'm not asking for them — but the *format* is being committed here and never exercised in reverse. A ~15-line `decryptAesGcm` in `crypto.ts` plus:
```ts
const { key, salt } = deriveKeyFromPassword(pw)
const { data, iv } = encryptAesGcm(mnemonic, key)
const { key: key2 } = deriveKeyFromPassword(pw, salt)
expect(decryptAesGcm(data, key2, iv)).toBe(mnemonic)
```
catches #1 immediately and validates #3's schema end-to-end. Combined with #3, this is trivial to add.

**6. Extract the Drive REST client into `app/utils/google-drive-client.ts`.**
`useGoogleDriveBackup` currently mixes three concerns: React `loading` state, Google Sign-In session lifecycle, and raw Drive REST (`findExistingFileId`, `buildMultipartBody`, `uploadOrUpdate`). Extract the REST layer:
```ts
// google-drive-client.ts
export const findAppDataFile = (name, token) => ...
export const uploadAppDataFile = ({ name, content, token, existingId }) => ...
export const downloadAppDataFile = (id, token) => ...   // for restore
```
The hook becomes a thin wrapper owning sign-in + `loading`. Benefits:
- Testable without `renderHook` + a React tree.
- `buildMultipartBody` (which has the fixed-boundary bug in Low) gets its own unit test.
- Fixes #4 (error handling) in a focused way.
- Restore PR has a clean place to add `downloadAppDataFile` without re-entangling the hook.

**7. Extract password validation into `app/utils/validators/password.ts`.** — `use-cloud-backup-form.ts:5`
```ts
const MIN_PASSWORD_LENGTH = 12
if (password.length < MIN_PASSWORD_LENGTH) return LL...passwordTooShort()
```
Length-only validation lets `"aaaaaaaaaaaa"` or `"123456789012"` through for the password protecting the seed. Given PBKDF2's work factor, a low-entropy password *is* the weak link, not the KDF.

I searched the repo — there's no existing password validator. This PR is the first user-chosen password in the codebase, so whatever lands here becomes the template. Worth doing once, properly:

- Reusable module with typed result: `{ valid: boolean; errors: PasswordIssue[]; strength: 'weak' | 'fair' | 'strong' }`.
- Prefer `zxcvbn-ts` (maintained, tree-shakeable, dictionary-aware) — or if bundle size is a concern, at minimum enforce min length + character-class diversity + a small common-password deny list.
- Keep i18n keys generic (`common.passwordTooWeak`, `common.passwordCommon`) rather than nesting under `SparkOnboarding.CloudBackup`, so the next password screen can reuse them.
- Drive the form error from the validator result, not an inline `useMemo` length check.

(Not AJV — JSON Schema validates *shape*, not password *quality*. Wrong tool.)

## 🟡 Non-blocking (strong suggestions)

**8. Consider AES-256 instead of AES-128** — `crypto.ts`
Cheap upgrade to `PBKDF2_KEY_LENGTH = 32` + `aes-256-gcm` for a long-lived seed-phrase backup. Safely deferrable **once #2 lands**, since the payload will carry the cipher name.

**9. Drive query not parent-scoped or escaped** — `use-google-drive-backup.ts:32`
```js
const query = encodeURIComponent(`name='${fileName}'`)
```
Add `and 'appDataFolder' in parents and trashed = false`, and escape `'` in `fileName`. Defense-in-depth. Natural to fix inside the extracted Drive client (#6).

**10. Keychain options are minimal** — `use-keychain-backup.ts:14`
```js
Keychain.setGenericPassword(service, value, { service })
```
For a seed phrase, explicitly set `accessible: ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY` (so it's not swept into iCloud Keychain) and consider `accessControl: BIOMETRY_CURRENT_SET_OR_DEVICE_PASSCODE`. Defaults vary by platform/OS. Once set, pin them in the keychain test.

**11. iOS cloud backup is a silent no-op** — `use-backup-methods.ts`
`handleCloudBackup` early-returns on iOS but the button is still rendered. Hide on iOS or disable + "coming soon". UX bug.

**12. Raw Drive error body shown in toast** — `use-cloud-backup.ts`
Google returns verbose JSON. Show a localized message; log the raw error.

**13. Wrap `MOCK_WORDS` behind `useWalletMnemonic()`.**
Imported directly in three hooks (`use-cloud-backup`, `use-backup-methods`, `use-backup-phrase`). When the real Spark SDK lands, you'll be grepping across hooks. One hook, one line to change later.

**14. Extract `buildConfirmChallenges(words, count)`.** — `use-backup-phrase.ts:handleContinue`
Pull the challenge construction out of the navigation callback. Pure function, trivially testable, and the single place to swap `Math.random` for `Crypto.randomBytes`.

**15. Unify the "Important" warning card with `InfoBanner`.**
The warning card in `cloud-backup-screen.tsx` (icon + title + body, primary color) is the same visual pattern as the new `InfoBanner`. Either extend `InfoBanner` with `variant: 'info' | 'warning'` + optional `title`/`icon`, or add `WarningCard` alongside it. Otherwise you'll have two near-identical implementations on the next PR that needs a warning UI.

**16. Extract rich-text helper for the bold-in-paragraph IIFE.** — `cloud-backup-screen.tsx`
The split-around-`boldText` IIFE is awkward and will be copy-pasted. Either `<BoldSpan before={...} bold={...} after={...} />` or a proper `<RichText>` that parses a `<bold>…</bold>` token. Second option scales better.

**17. Promise-wrapped `Alert.alert` → reusable `confirmDialog()` helper.** — `use-cloud-backup.ts`
```ts
const confirmOverwrite = ({...}) => new Promise((resolve) => Alert.alert(...))
```
This is generic — extract to `app/utils/confirm-dialog.ts`. Any future "are you sure?" flow will want it.

## 🟢 Low / nits

- **Fixed multipart boundary** (`"blink_backup_boundary"`) — randomize per request; if `content` ever contained that string, the upload corrupts. Fix inside the extracted Drive client (#6).
- **`useGoogleDriveBackup.loading` only tracks `upload`**, not `startSession` — no spinner during sign-in.
- **`useBackupConfirm` auto-completes after 400ms** — `hasCompleted` guard is good, but also disable inputs once it flips to avoid post-completion state churn.
- **`expiresAt = useRef(new Date(...)).current`** resets on remount (step1 → step2 → back). Confirm that matches intended timer semantics; otherwise lift into route params.
- **`configureGoogleSignIn()` runs on every `signIn()`** — harmless, could be one-time.
- **PATCH metadata includes `name`** — unnecessary on update, harmless.
- **Unused exports in `__mocks__/react-native-keychain.js`** (`getGenericPassword`, `resetGenericPassword`) — remove, or wire up alongside a keychain `read()` in a later PR.
- **`useKeychainBackup` tests only cover `save`** and don't pin the options object — once #10's options land, add an assertion.
- **`useBackupPhrase` branches internally on `isStep1`** for word slicing, button title, disabled state, and continue behavior. Readable today, but consider returning a pre-computed `buttonConfig` object or splitting into two small hooks as it grows.
- **Collocate Spark backup ID helpers.** `getSparkDriveBackupFilename` and `getSparkKeychainService` live in `app/config/appinfo.ts` among unrelated config — a small `spark-backup-ids.ts` would be clearer.

---

**Summary.** The seven blockers break into three groups:

- **Format correctness** (#1, #2, #5): IV length, schema metadata, round-trip test. Locks the format down before any user depends on it.
- **Structural extractions** (#3, #6, #7): format module, Drive client, password validator. Each one pulls pure logic out of a hook, makes the corresponding correctness fix trivial, and sets the template for future work.
- **Error propagation** (#4): silent duplicate on Drive query failure.

Collectively they're ~a day of focused work, not a rewrite. Everything in Non-blocking and Low can land in follow-ups — but if the Blocking items land together, the follow-ups become safe incremental PRs instead of format migrations.

grimen (changes requested):
Nice work — clean hook separation, strong test coverage, thorough i18n. The structure is mostly right, but I'd like to see a few things tightened before this merges. Two reasons:

1. **Format lock-in.** The PR description notes Drive access is still pending GCP config and the seed is still mocked — so no real user has a backup yet. That's the window to get the on-disk format right; after that, every fix is a migration.
2. **This PR sets templates.** It's the first user-chosen password in the codebase, the first cloud-backup adapter, and the first Google Drive integration. Whatever lands here will be copy-pasted. Worth investing in the shape now.

Most of the feedback is refactoring that extracts pure logic out of hooks so it's testable without React — which is also how the format round-trip test (the thing that catches the IV bug) becomes trivial to write.

## 🔴 Blocking

**1. AES-GCM IV should be 12 bytes, not 16** — `app/utils/crypto.ts:48`
GCM is specified for 96-bit IVs. Other lengths trigger GHASH preprocessing and hurt interop. Change `Crypto.randomBytes(16)` → `Crypto.randomBytes(12)`.

**2. Backup JSON must persist crypto parameters.**
Only `version: 1` is stored; `iterations`, `digest`, `keyLen`, and `cipher` are hard-coded constants. Any future change (including fixing #1 or moving to AES-256) breaks restore. Persist them now:
```json
{
  "version": 1,
  "kdf": { "name": "PBKDF2", "iterations": 600000, "digest": "SHA-256", "keyLen": 16 },
  "cipher": "AES-128-GCM",
  ...
}
```
With this in place, most Medium items below become safely migratable in follow-ups.

**3. Extract the backup format into `app/utils/spark-backup-format.ts`.**
Currently inline as an IIFE in `use-cloud-backup.ts`:
```ts
const payload = isEncrypted ? (() => { ... })() : JSON.stringify({...})
```
Pull this out into a pure module:
```ts
// spark-backup-format.ts
export type BackupPayload = /* discriminated union: encrypted | plain */
export const buildBackupPayload = (mnemonic, opts) => ...
export const parseBackupPayload = (raw: string) => ...   // for restore
```
This is the natural home for #2's schema, lets you write format round-trip tests without React, and gives the restore PR a ready-made parser. Single biggest structural win in the PR.

**4. `findExistingFileId` silently returns `undefined` on non-OK** — `use-google-drive-backup.ts:36`
A 401/5xx is treated as "no existing file" → the next step **creates a duplicate backup** with no user signal. Throw and surface the error.

**5. Add `decryptAesGcm` + a crypto round-trip test.**
Restore screens are out of scope and I'm not asking for them — but the *format* is being committed here and never exercised in reverse. A ~15-line `decryptAesGcm` in `crypto.ts` plus:
```ts
const { key, salt } = deriveKeyFromPassword(pw)
const { data, iv } = encryptAesGcm(mnemonic, key)
const { key: key2 } = deriveKeyFromPassword(pw, salt)
expect(decryptAesGcm(data, key2, iv)).toBe(mnemonic)
```
catches #1 immediately and validates #3's schema end-to-end. Combined with #3, this is trivial to add.

**6. Extract the Drive REST client into `app/utils/google-drive-client.ts`.**
`useGoogleDriveBackup` currently mixes three concerns: React `loading` state, Google Sign-In session lifecycle, and raw Drive REST (`findExistingFileId`, `buildMultipartBody`, `uploadOrUpdate`). Extract the REST layer:
```ts
// google-drive-client.ts
export const findAppDataFile = (name, token) => ...
export const uploadAppDataFile = ({ name, content, token, existingId }) => ...
export const downloadAppDataFile = (id, token) => ...   // for restore
```
The hook becomes a thin wrapper owning sign-in + `loading`. Benefits:
- Testable without `renderHook` + a React tree.
- `buildMultipartBody` (which has the fixed-boundary bug in Low) gets its own unit test.
- Fixes #4 (error handling) in a focused way.
- Restore PR has a clean place to add `downloadAppDataFile` without re-entangling the hook.

**7. Extract password validation into `app/utils/validators/password.ts`.** — `use-cloud-backup-form.ts:5`
```ts
const MIN_PASSWORD_LENGTH = 12
if (password.length < MIN_PASSWORD_LENGTH) return LL...passwordTooShort()
```
Length-only validation lets `"aaaaaaaaaaaa"` or `"123456789012"` through for the password protecting the seed. Given PBKDF2's work factor, a low-entropy password *is* the weak link, not the KDF.

I searched the repo — there's no existing password validator. This PR is the first user-chosen password in the codebase, so whatever lands here becomes the template. Worth doing once, properly:

- Reusable module with typed result: `{ valid: boolean; errors: PasswordIssue[]; strength: 'weak' | 'fair' | 'strong' }`.
- Prefer `zxcvbn-ts` (maintained, tree-shakeable, dictionary-aware) — or if bundle size is a concern, at minimum enforce min length + character-class diversity + a small common-password deny list.
- Keep i18n keys generic (`common.passwordTooWeak`, `common.passwordCommon`) rather than nesting under `SparkOnboarding.CloudBackup`, so the next password screen can reuse them.
- Drive the form error from the validator result, not an inline `useMemo` length check.

(Not AJV — JSON Schema validates *shape*, not password *quality*. Wrong tool.)

## 🟡 Non-blocking (strong suggestions)

**8. Consider AES-256 instead of AES-128** — `crypto.ts`
Cheap upgrade to `PBKDF2_KEY_LENGTH = 32` + `aes-256-gcm` for a long-lived seed-phrase backup. Safely deferrable **once #2 lands**, since the payload will carry the cipher name.

**9. Drive query not parent-scoped or escaped** — `use-google-drive-backup.ts:32`
```js
const query = encodeURIComponent(`name='${fileName}'`)
```
Add `and 'appDataFolder' in parents and trashed = false`, and escape `'` in `fileName`. Defense-in-depth. Natural to fix inside the extracted Drive client (#6).

**10. Keychain options are minimal** — `use-keychain-backup.ts:14`
```js
Keychain.setGenericPassword(service, value, { service })
```
For a seed phrase, explicitly set `accessible: ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY` (so it's not swept into iCloud Keychain) and consider `accessControl: BIOMETRY_CURRENT_SET_OR_DEVICE_PASSCODE`. Defaults vary by platform/OS. Once set, pin them in the keychain test.

**11. iOS cloud backup is a silent no-op** — `use-backup-methods.ts`
`handleCloudBackup` early-returns on iOS but the button is still rendered. Hide on iOS or disable + "coming soon". UX bug.

**12. Raw Drive error body shown in toast** — `use-cloud-backup.ts`
Google returns verbose JSON. Show a localized message; log the raw error.

**13. Wrap `MOCK_WORDS` behind `useWalletMnemonic()`.**
Imported directly in three hooks (`use-cloud-backup`, `use-backup-methods`, `use-backup-phrase`). When the real Spark SDK lands, you'll be grepping across hooks. One hook, one line to change later.

**14. Extract `buildConfirmChallenges(words, count)`.** — `use-backup-phrase.ts:handleContinue`
Pull the challenge construction out of the navigation callback. Pure function, trivially testable, and the single place to swap `Math.random` for `Crypto.randomBytes`.

**15. Unify the "Important" warning card with `InfoBanner`.**
The warning card in `cloud-backup-screen.tsx` (icon + title + body, primary color) is the same visual pattern as the new `InfoBanner`. Either extend `InfoBanner` with `variant: 'info' | 'warning'` + optional `title`/`icon`, or add `WarningCard` alongside it. Otherwise you'll have two near-identical implementations on the next PR that needs a warning UI.

**16. Extract rich-text helper for the bold-in-paragraph IIFE.** — `cloud-backup-screen.tsx`
The split-around-`boldText` IIFE is awkward and will be copy-pasted. Either `<BoldSpan before={...} bold={...} after={...} />` or a proper `<RichText>` that parses a `<bold>…</bold>` token. Second option scales better.

**17. Promise-wrapped `Alert.alert` → reusable `confirmDialog()` helper.** — `use-cloud-backup.ts`
```ts
const confirmOverwrite = ({...}) => new Promise((resolve) => Alert.alert(...))
```
This is generic — extract to `app/utils/confirm-dialog.ts`. Any future "are you sure?" flow will want it.
## 🧪 Test coverage gaps

Volume is high (14 new specs) but the most important branches aren't exercised. Several gaps map directly onto blocking items:

- **No real crypto runs anywhere.** `crypto.spec.ts` mocks `react-native-quick-crypto` at the module level, and `use-cloud-backup.spec.ts` then mocks `@app/utils/crypto` entirely on top. The "deterministic output with same salt" test is asserting the mock returns the same value twice. The IV length (blocking #1), schema (blocking #2), and the missing decrypt primitive (blocking #5) are all invisible to this suite. A test would pass even if the encryption were broken.
- **One round-trip test fixes most of this.** Extract the format per blocking #3, add `decryptAesGcm` per blocking #5, then in `spark-backup-format.spec.ts` use the *real* crypto module (don't mock) and assert `parse(build(mnemonic, password)) === mnemonic` plus `iv.length === 12`. If this test passes, the format is real. If it fails, the PR can't ship.
- **`findExistingFileId` non-OK path is untested** (blocking #4). The silent-duplicate bug is invisible to the Drive hook suite.
- **`buildMultipartBody` has no isolated test.** The fixed-boundary bug (Low) has no coverage. Both issues are much easier to test once the Drive client is extracted per blocking #6.
- **Drive query escaping / `appDataFolder` parent constraint** (Medium #9) — no assertion on the query string shape.
- **Upload PATCH vs POST** — should be table-tested over `{ existingId }` present/absent.
- **Keychain options beyond `{ service }` aren't pinned** (Medium #10). When `accessible` / `accessControl` land, they must come with a pinning test.
- **`useKeychainBackup.error` clearing on a subsequent successful save** is untested — the hook does `setError(undefined)` and nothing notices.
- **Password form accepts trivially weak passwords** (`"aaaaaaaaaaaa"`) — no test asserts this is rejected, which is why a length-only validator shipped (blocking #7).
- **`useFocusEffect` password clearing** is effectively untested — the effect is typically mocked in this codebase, so the "passwords cleared on blur" behavior in the PR description has no end-to-end coverage.
- **iOS `handleCloudBackup` early-return** (Medium #11) has no platform-specific test.
- **`useBackupConfirm` 400ms auto-navigate race** — `hasCompleted.current` protects against double-fire but not against state mutating after completion. Untested.
- **`isWordCorrect` case-insensitivity** — no explicit mixed-case test.

The pattern is the same across all of these: tests exercise the React wiring (state transitions, mock call counts) but not the underlying contracts (does real crypto round-trip, does the Drive query have the right shape, does a weak password get rejected). Most of these gaps close automatically once the blocking refactors (#3 format module, #6 Drive client, #7 password validator) extract the logic into pure modules that can be tested directly.


## 🟢 Low / nits

- **Fixed multipart boundary** (`"blink_backup_boundary"`) — randomize per request; if `content` ever contained that string, the upload corrupts. Fix inside the extracted Drive client (#6).
- **`useGoogleDriveBackup.loading` only tracks `upload`**, not `startSession` — no spinner during sign-in.
- **`useBackupConfirm` auto-completes after 400ms** — `hasCompleted` guard is good, but also disable inputs once it flips to avoid post-completion state churn.
- **`expiresAt = useRef(new Date(...)).current`** resets on remount (step1 → step2 → back). Confirm that matches intended timer semantics; otherwise lift into route params.
- **`configureGoogleSignIn()` runs on every `signIn()`** — harmless, could be one-time.
- **PATCH metadata includes `name`** — unnecessary on update, harmless.
- **Unused exports in `__mocks__/react-native-keychain.js`** (`getGenericPassword`, `resetGenericPassword`) — remove, or wire up alongside a keychain `read()` in a later PR.
- **`useKeychainBackup` tests only cover `save`** and don't pin the options object — once #10's options land, add an assertion.
- **`useBackupPhrase` branches internally on `isStep1`** for word slicing, button title, disabled state, and continue behavior. Readable today, but consider returning a pre-computed `buttonConfig` object or splitting into two small hooks as it grows.
- **Collocate Spark backup ID helpers.** `getSparkDriveBackupFilename` and `getSparkKeychainService` live in `app/config/appinfo.ts` among unrelated config — a small `spark-backup-ids.ts` would be clearer.

---

**Summary.** The seven blockers break into three groups:

- **Format correctness** (#1, #2, #5): IV length, schema metadata, round-trip test. Locks the format down before any user depends on it.
- **Structural extractions** (#3, #6, #7): format module, Drive client, password validator. Each one pulls pure logic out of a hook, makes the corresponding correctness fix trivial, and sets the template for future work.
- **Error propagation** (#4): silent duplicate on Drive query failure.

Collectively they're ~a day of focused work, not a rewrite. Everything in Non-blocking and Low can land in follow-ups — but if the Blocking items land together, the follow-ups become safe incremental PRs instead of format migrations.

esaugomez31:
@grimen please review

# PR Review Response

---

## Blocking Items

### 1. AES-GCM IV should be 12 bytes, not 16

Changed the default IV length in `encryptAesGcm` to 12 bytes (`DEFAULT_IV_LENGTH = 12`). The function now accepts an `options` parameter so callers can override the length when needed. The existing card encryption flow passes `{ ivLength: 16 }` explicitly to preserve its current behavior.

### 2. Backup JSON must persist crypto parameters

The new `spark-backup-format.ts` module includes `kdf` and `cipher` fields in every encrypted payload. The schema looks like this:

```json
{
  "version": 1,
  "kdf": { "name": "PBKDF2", "iterations": 600000, "digest": "SHA-256", "keyLen": 16 },
  "cipher": "AES-128-GCM",
  "encrypted": true,
  "data": "...",
  "iv": "...",
  "salt": "..."
}
```

The KDF constants are imported from `crypto.ts` so there is a single source of truth.

### 3. Extract the backup format into a pure module

Created `app/utils/spark-backup-format.ts` with three exports:

- `buildBackupPayload(mnemonic, opts)` builds the JSON string.
- `parseBackupPayload(raw)` handles unencrypted payloads.
- `parseEncryptedBackupPayload(raw, password)` handles decryption.

The inline IIFE in `use-cloud-backup.ts` has been replaced with a single `buildBackupPayload` call. A dedicated test file covers plain and encrypted round-trips, schema validation, IV length assertion, and wrong-password rejection.

### 4. findExistingFileId silently returns undefined on non-OK

Moved the Drive REST logic into `app/utils/google-drive-client.ts`. The `findAppDataFile` function now throws on any non-OK response, including the status code and response body in the error message. The hook (`use-google-drive-backup.ts`) catches the error and surfaces it through the existing result pattern. Direct contract tests cover the throw behavior.

### 5. Add decryptAesGcm and a crypto round-trip test

Added `decryptAesGcm` to `crypto.ts`. The backup format spec file includes a full round-trip test that uses real Node.js crypto, not mocked wrappers. It verifies that `parseEncryptedBackupPayload(buildBackupPayload(mnemonic, { password }), password)` returns the original mnemonic.

### 6. Extract the Drive REST client

Created `app/utils/google-drive-client.ts` with `findAppDataFile`, `uploadAppDataFile`, `downloadAppDataFile`, and `buildMultipartBody`. The boundary is now randomized per request. The hook is a thin wrapper that owns Google Sign-In and the loading state. Direct contract tests cover query shape, error throwing, POST vs PATCH routing, multipart body structure, and file download.

### 7. Extract password validation

Created `app/utils/validators/password.ts` with a typed result (`{ valid, errors, strength }`). It enforces minimum length, requires at least three character classes (uppercase, lowercase, digits, special), and rejects a set of common passwords. The form hook (`use-cloud-backup-form.ts`) now consumes the validator instead of an inline length check. The i18n keys live under `common.*` so they can be reused elsewhere. Direct tests cover weak, common, short, and strong passwords.

---

## Non-Blocking Items

### 8. Consider AES-256 instead of AES-128

Deferred. Now that the cipher name is persisted in the payload (thanks to item 2), upgrading to AES-256 is a safe, incremental change that will not require a format migration.

### 9. Drive query not parent-scoped or escaped

The query now includes `'appDataFolder' in parents and trashed = false`, and single quotes in the file name are escaped. Covered by a dedicated test.

### 10. Keychain options are minimal

Added `accessible: ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY` to the `setGenericPassword` call. The test pins this value in its assertion. Did not add `accessControl` for now since it changes the UX on some devices, but the option is easy to add later.

### 11. iOS cloud backup is a silent no-op

The cloud button is now disabled on iOS and a "coming soon" label is shown below it. The handler also shows a toast as a fallback. Both behaviors are covered by tests.

### 12. Raw Drive error body shown in toast

Upload failures now show a localized message (`uploadFailed`) and log the raw error to the console. Sign-in failures also show a localized message (`signInFailed`) instead of surfacing the raw SDK error.

### 13. Wrap MOCK_WORDS behind useWalletMnemonic

Created `app/hooks/use-wallet-mnemonic.ts`. All three backup hooks now import from this single hook instead of referencing `MOCK_WORDS` directly. The hook includes a `__DEV__` guard that throws in production builds to prevent accidentally shipping mock data.

### 14. Extract buildConfirmChallenges

Extracted to `app/screens/spark-onboarding/utils.ts` as a pure function. The backup phrase hook imports and uses it instead of building challenges inline.

### 15. Unify the warning card with InfoBanner

Extended `InfoBanner` with a `variant` prop (`"info"` or `"warning"`), plus optional `title` and `icon`. Dynamic styles are handled through `makeStyles` with runtime parameters, no inline style objects. The cloud backup screen uses the warning variant.

### 16. Extract rich-text helper for the bold-in-paragraph IIFE

Created `app/components/rich-text/rich-text.tsx`. It takes `text` and `bold` props, splits around the bold segment, and renders styled spans. Includes a guard for the case where the bold text is not found. The cloud backup screen uses this instead of the inline IIFE.

### 17. Promise-wrapped Alert.alert as reusable confirmDialog

Created `app/utils/confirm-dialog.ts` with a typed, generic helper. The cloud backup hook uses it instead of the inline `confirmOverwrite` wrapper.

---

## Test Coverage

| Area | Coverage |
|------|----------|
| Crypto round-trip (real crypto, not mocked) | `spark-backup-format.spec.ts`, 7 tests |
| Backup format schema and IV length | `spark-backup-format.spec.ts` |
| Drive client contract (throw, query, POST/PATCH, download) | `google-drive-client.spec.ts`, 7 tests |
| Password validator (weak, common, strong, short) | `password-validator.spec.ts`, 6 tests |
| Keychain accessible option pinned | `use-keychain-backup.spec.ts` |
| IV length default and override | `crypto.spec.ts` |
| Card encryption backward compatibility | `card-encryption.spec.ts` |

grimen (approved):
All blocking and non-blocking feedback addressed. Remaining low-priority nits are fine as follow-ups. LGTM.

grimen:
### Merge activity

* **May 20, 10:43 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3731).
* **May 20, 10:43 AM UTC**: @grimen merged this pull request with [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3731).

PR #3762, opened 2026-04-23
fix(spark): address settings + onboarding UI/UX issues from


esaugomez31:
# Spark: settings + onboarding UI/UX fixes

## What this PR does

Hardening pass on the self-custodial onboarding and settings surfaces. Tightens the manual backup and restore flows (auto-advance between word inputs, paste-and-skip on the first restore step, clearer error feedback, post-success routing that respects re-backup vs first backup), reorganizes the settings screen, fixes several bugs around the backup nudge banner and modal, and replaces the SDK status-check loop in the wallet provider with a snapshot-first lifecycle that uses `AbortController` so the wallet no longer hangs on initial render when Spark services are slow.

## Restore phrase flow

The 12-word restore screen now treats step 1 as a fast path: if the user pastes the full phrase and it is a valid BIP39 mnemonic, the screen navigates straight to step 2 with the parsed words pre-filled instead of forcing them to scroll back to a continue button. Word-by-word entry also auto-advances to the next field when what the user typed is unambiguously a single BIP39 match — typing `abandon` jumps to the next slot, but typing `run` waits because both `run` and `runway` exist. The mnemonic input was lifted to `forwardRef` with `useImperativeHandle` so the parent can drive focus programmatically; both restore and manual-confirm screens use `MnemonicWordInputHandle` for this.

The error UX on step 2 is no longer hidden behind a button press: when all twelve fields are filled and the joined phrase fails `validateMnemonic`, the screen surfaces "Invalid backup phrase. Please check if the word order is correct." inline with a warning icon, in a container with reserved height so the layout doesn't jump when the message appears or disappears. The same reserved-space pattern was applied to the manual backup confirm screen.

After a successful restore, the wallet is now marked as backed up via `setBackupCompleted(BackupMethod.Manual)` (previously the restore path called `resetBackupState()` which left the user looking like they still owed a backup).

## Manual backup confirm flow

Typing a correct word in the confirm screen now auto-focuses the next challenge input instead of waiting for a tap. The focus-advance hook (`useBackupConfirm`) returns a `focusRequest` index and a `clearFocusRequest` callback; the screen mounts ref handles for each input and applies focus from the effect. Same pattern as the restore screen.

The completion handler now distinguishes three cases:

1. **Migrating from custodial with funds** → routes to the existing `sparkMigrationTransferringFunds` screen.
2. **Re-backup from settings** (already backed up) → routes to the success screen with `reBackup: true`, which renders the generic `LL.common.success()` copy instead of the first-time "Welcome to non-custodial" headline. The migration check uses both the `useMigrationCheckpoint` checkpoint and the current `BackupStatus`, because the alerts screen sets the checkpoint even when the user enters from settings — checking only the checkpoint misroutes re-backups into the migration screen.
3. **Fresh manual backup** → routes to the success screen with `reBackup: false`.

The `sparkBackupSuccessScreen` route param shape was widened to `{ reBackup?: boolean } | undefined` and the screen reads it from `useRoute`.

## Cloud backup screen polish

Password validation errors on the cloud-backup form now only show after the field has been blurred, not while the user is still typing. The cloud hero icon and the backup/restore method icons in onboarding switched to the success color, and the "Important" `InfoBanner` icon switched to the warning color, so they match the rest of the design system.

## Backup nudge

The dismissal cooldown only applies to the home banner; the settings banner now ignores it (`shouldShowSettingsBanner` no longer waits for the `loaded` flag from `AsyncStorage.getItem(DISMISSAL_KEY)`). The home `BackupNudgeModal` is now gated behind `useIsFocused()`, so it doesn't pop up on top of unrelated screens (e.g. settings) when the user has the modal threshold balance and walks away from the home tab.

The threshold check moved from the active wallet's BTC sats to the combined display balance via `useTotalBalance`. Mixed-currency users with USD weight that crosses the threshold no longer slip past the modal because their BTC alone was below it.

## Account-type selection card

The selected account-type card on the picker screen now uses `colors.grey6` as the background. The previous color blended into the unselected variant on dark mode and made it hard to tell which option was selected.

## Self-custodial SDK lifecycle

Rewrote `useSdkLifecycle` to be snapshot-first: a refresh tries `getSelfCustodialWalletSnapshot` directly and only falls back to the operational status check on failure. Before this, every refresh ran a `getSparkStatus` call upfront, and on cold start that call could hang past any practical timeout — leaving the home screen in an indefinite loading state even though the SDK had already initialized correctly. The new flow only consults `getServiceStatus` when the snapshot rejects, which is when the operational state actually matters.

`getServiceStatus` itself was rewritten to use `AbortController` instead of `Promise.race`. The 5-second timeout aborts the underlying `getSparkStatus` call cleanly, no orphan promises. The bridge wrapper for `getSparkStatus` now accepts an optional `AbortSignal` and forwards it to the SDK's async-opts.

The `OFFLINE_EXEMPT_STATUSES` set keeps `Error` and `Unavailable` states from being downgraded to `Offline` on a refresh failure, so a user with no mnemonic doesn't see "wallet offline" copy.

## Settings reorganization

"View backup phrase" sits under **Security & privacy** alongside `TotpSetting` and `OnDeviceSecuritySetting` for self-custodial wallets. An earlier draft introduced a standalone **Recovery method** group, but custodial users would have seen an empty card under that header (the entry only renders for self-custodial), so the section was collapsed back into the existing security group.

The settings screen also skips its GraphQL queries when the user isn't authenticated, which fixes a flicker where the network requests fired against a missing token and produced a brief error state.

## Receive polish

The displayed Lightning invoice now strips a leading `lightning:` URI prefix before rendering. Some external sources prepend it; copying or showing it to the user added noise.

## i18n

New keys: `SettingsScreen.recoveryMethod`, the post-success `common.success` reuse for re-backup, and the updated `RestoreScreen.invalidMnemonic` copy ("Invalid backup phrase. Please check if the word order is correct."). All 28 locales updated via the project's Python-script i18n workflow, preserving the existing per-locale indentation and trailing-newline quirks. `Phrase.headerTitle` was retitled to "Backup phrase" (was "Your backup phrase") to match the new settings entry.

## Tests

3667/3667 green. New coverage includes the prefix-unique BIP39 advance and `clearFocusRequest` in `use-bip39-input`, the `forwardRef` focus exposure on `MnemonicWordInput`, the auto-navigate paste path in `use-restore-phrase`, the focus-advance hook in `use-backup-confirm`, the migration / re-backup routing branches in `backup-confirm-screen` (including the no-funds case), the `reBackup` param in `backup-success-screen`, the AbortController + signal forwarding in `is-online`, the snapshot-first / Offline / Error / Loading→Offline transitions in `wallet-provider`, and the combined-balance threshold path in `use-backup-nudge-state`. Existing specs were updated to drop the removed balance-stale heuristic and the old `resetBackupState` call site.

## Breaking notes

None for consumers. The self-custodial wallet provider's status snapshot no longer maintains an internal `isBalanceStale` ref/state, since the post-convert sync from the receive auto-convert pipeline (PR #3764) makes the heuristic unnecessary. All in-tree callers were updated. Custodial flows are untouched.


esaugomez31:
* **#3762** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

> [!IMPORTANT]
> **The structural debt from #3758 carries forward here, and this PR layers more on top.** Deferred findings include:
> - **SOLID violations** — `useSdkLifecycle` has at least six responsibilities in one hook (lifecycle, refresh, AppState polling, 10s interval polling, retry orchestration, snapshot-vs-status reducer, user-settings fetch); `useBackupNudgeState` mixes async-load state, threshold computation, and dismissal-cooldown into one return shape; `OFFLINE_EXEMPT_STATUSES` is reducer-internal but exposed at module scope; the bridge `getSparkStatus` wrapper still mixes pass-through with the new abort plumbing.
> - **Incomplete adapter pattern (carries from #3758)** — mode still leaks into call sites instead of being hidden behind a uniform Port; `setBackupCompleted(BackupMethod.Manual)` is called from screens directly rather than through a backup-completion port; the migration check in `backup-confirm-screen.tsx` reads from two different providers (`useMigrationCheckpoint` + `useBackupState`) and combines them inline rather than via a single `useBackupRouteDecision()`-style port.
> - **High cyclomatic complexity in the screen + lifecycle layer** — `backup-confirm-screen.tsx`'s `onComplete` is a 3-way routing decision based on `checkpoint × backupState × activeWallet.balance`, all combined inline; `use-sdk-lifecycle.ts`'s `refreshWallets` mixes snapshot-success / snapshot-fail / connectivity-3-state / status-preservation rules in one function; `restore-phrase-screen.tsx` combines step-1 paste, per-input focus forwarding, inline error rendering, and reserved-height layout in one component; `useBip39Input.updateWord` mixes normalisation, BIP39-set membership, prefix-uniqueness, and last-word-of-step bounds in one branch.
> - **Manual concurrency control smells** — `refreshingRef` + `pendingRefreshRef` dance, `hasCompleted.current` latch, ref-based focus-request pattern (`focusRequest` index + `clearFocusRequest`) instead of imperative `inputRef.focus()`. Each of these is a known foot-gun pattern (and Critical #5 / #6 are direct consequences).
> - **Code smells** — empty `catch {}` blocks (Important I1, I4, I5, I6), fire-and-forget AsyncStorage persists, boolean route params (`reBackup?: boolean` instead of a route-mode discriminator), test mocks that discard the very argument under test (`__tests__/self-custodial/bridge/status.spec.ts:14`), three-state status preserved via a Set named after the *exclusion* (`OFFLINE_EXEMPT_STATUSES`).
> - **Naming smells** — overlapping vocabulary `BackupStatus` (state) / `BackupMethod` (enum) / `useBackupState` (hook) / `setBackupCompleted` (action) without a coherent dictionary; `OFFLINE_EXEMPT_STATUSES` describes what's *excluded* not what it represents; `hasCompleted` is ambiguous (completed what?); `focusRequest` / `clearFocusRequest` is a request/clear pair where an imperative API would do; `reBackup` boolean leaks an internal mode through the navigation contract.
>
> **Why this is not in scope for this PR:** this PR sits in the same stack as #3758–#3769. Refactoring the lifecycle hook, route-decision plumbing, or vocabulary here would force a multi-day restack across the rest of the stack and create combinatorial conflict pain — particularly with the rename passes already deferred from #3758. Refactoring against `main` *after* the stack settles is materially cheaper.
>
> **Decision: this review covers correctness only.** The structural / SOLID / complexity / naming review will fold into the **same cleanup ticket opened against #3758**, scoped as post-stack work against `main`. **The cleanup ticket should reach at least the lifecycle-hook decomposition and adapter-pattern unification milestones before #3768 (production rollout) ships**, otherwise the structural debt ships and gets normalised.

---

Severity:
- **Critical** — user-visible hang, mis-routed funds-affecting flow, or silent regression of the *exact* class this PR is meant to fix. Must fix before merge.
- **Important** — correctness/observability that doesn't lose funds but degrades UX or hides incidents.

---

## Critical

### 1. Migration routing race vs. AsyncStorage load

`app/screens/spark-onboarding/manual-backup/backup-confirm-screen.tsx:43-50`

`useMigrationCheckpoint()` loads `checkpoint` async from AsyncStorage and exposes a `loading` flag — the screen ignores it. The auto-navigate timer in `useBackupConfirm` fires at 400 ms. If the user (or paste/auto-fill) drives `allCorrect` true before AsyncStorage resolves, `checkpoint === null`, `isMigrating === false`, and a real **migration-with-funds** user is routed to `sparkBackupSuccessScreen` instead of `sparkMigrationTransferringFunds` — they skip the funds-transfer step entirely.

**Fix:** gate the auto-navigate (or the route decision) on `loading === false`; refuse to route while the checkpoint hook is unsettled.

### 2. `getSelfCustodialWalletSnapshot` has no timeout/abort

`app/self-custodial/providers/use-sdk-lifecycle.ts:80-115`

This PR adds a 5 s `AbortController` timeout to `getServiceStatus` to fix the cold-start hang — but the new **snapshot-first hot path is unprotected**. If `getWalletInfo`/`listPayments` hang, `refreshingRef.current` stays `true`, every subsequent refresh is coalesced via `pendingRefreshRef`, and the user is stuck in `Loading` forever. Same bug class the PR is meant to fix, just shifted to a different call.

**Fix:** forward a per-call AbortController into `getSelfCustodialWalletSnapshot` (and the `getWalletInfo`/`listPayments` underneath), or wrap with a timeout race using the same `STATUS_TIMEOUT_MS`.

### 3. `getOnlineState()` has no timeout

`app/self-custodial/providers/is-online.ts:41-46` + caller at `use-sdk-lifecycle.ts:96-100`

The connectivity-fallback probe used in the snapshot-failure catch path calls `getSparkStatus()` with **no signal, no timeout**. Bare `catch {}` returns `Unknown`. The PR's new abort plumbing was added to `getServiceStatus` only; the recovery-path probe was left unbounded — same hang class as #2.

**Fix:** delegate `getOnlineState` to `getServiceStatus` (which is already abort-armed) or apply the same `AbortController + STATUS_TIMEOUT_MS` pattern.

### 4. `Loading + Unknown` connectivity is a dead-end state

`app/self-custodial/providers/use-sdk-lifecycle.ts:~108`

Cold-start sequence: prev = `Loading` → snapshot fails → `getOnlineState` returns `Unknown` (network probe also hung/failed) → status update returns `prev` → **stays in `Loading` forever**. `Offline`/`Ready` transitions are handled; this combination was missed. Exactly the indefinite-loading regression the PR is supposed to fix.

**Fix:** when `prev === Loading` and connectivity is `Unknown`, transition to `Error` (or a new `Stalled`) rather than preserving `Loading`.

### 5. Stale-state write race on retry

`app/self-custodial/providers/use-sdk-lifecycle.ts:80-115`

`refreshWallets` does not capture an epoch/cancellation token. On `retryCount` change the SDK is replaced, but an in-flight refresh against the **old** SDK resolves later and calls `setWallets`/`setStatus(Ready)` — overwriting fresh session state with stale data from a disconnected session.

**Fix:** capture an epoch (or generation counter) at call time, bail before writing if it changed.

### 6. `hasCompleted.current` latches forever

`app/screens/spark-onboarding/hooks/use-backup-confirm.ts:48-55`

After the auto-navigate timer flips the latch, `if (allCorrect && !hasCompleted.current)` is permanently false. If the user transiently de-corrects a word inside the 400 ms window (then re-corrects), the headline auto-advance UX is silently disabled for the rest of the session. The manual button still works, so degraded-not-broken — but it silently breaks the marquee improvement of this PR.

**Fix:** reset `hasCompleted.current = false` whenever `allCorrect` becomes false.

### 7. Empty "Recovery method" group renders for non-self-custodial users

`app/screens/settings-screen/settings-screen.tsx:166-169`

`ViewBackupPhraseSetting` returns `null` for non-self-custodial accounts, but `SettingsGroup` always renders the header label + empty card. Custodial users see an orphan section with a label and nothing inside it.

**Fix:** conditionally render the group only when `activeAccount?.type === SelfCustodial`, or have `SettingsGroup` skip rendering when `filteredItems.length === 0`.

---

## Important

| # | Item | Location |
|---|---|---|
| I1 | Bare `catch {}` discards original error and conflates abort/timeout/network/SDK-bug into a single return; no Crashlytics record | `app/self-custodial/providers/is-online.ts:18, 43` |
| I2 | `crashlytics().log` is a breadcrumb, not a recordable error — original cause from `getSparkStatus` is dropped, won't surface in Crashlytics error list | `app/self-custodial/providers/use-sdk-lifecycle.ts:96-100` |
| I3 | Snapshot-success unconditionally sets `Ready`, never consults service status — undocumented trade-off; degraded service surfaces only on op failure | `app/self-custodial/providers/use-sdk-lifecycle.ts:85-88` |
| I4 | `catch {}` in `generateRequest` swallows `createReceiveLightning` errors (no Crashlytics record); sibling effect at L117-127 does it correctly | `app/self-custodial/hooks/use-payment-request.ts:88-90` |
| I5 | `KeyStoreWrapper.deleteMnemonic().catch(() => {})` could leave a partial mnemonic on disk silently after failed restore — affects next attempt | `app/screens/spark-onboarding/restore/hooks/use-restore-wallet.ts:59` |
| I6 | Redundant `await restore(mnemonic).catch(() => {})` (restore handles its own errors); silently hides any future bug-class throw | `app/screens/spark-onboarding/restore/hooks/use-restore-phrase.ts:78` |
| I7 | PR description says "strips a leading `lightning:` prefix"; actual change just stops *adding* it. If the SDK ever returns invoices already prefixed, those display unchanged | `app/self-custodial/hooks/use-payment-request.ts` |
| I8 | AbortSignal is forwarded to the SDK but no test asserts the SDK actually honors it — the bridge spec mock at `__tests__/self-custodial/bridge/status.spec.ts:14` actively discards the argument | `app/self-custodial/bridge/status.ts:6-7` |

---

## Test coverage

Estimated behavioural coverage of the changed surfaces: **~70%** — strong on hooks (BIP39, backup-confirm, restore, cloud form, lifecycle state transitions) and bridge plumbing; sparse on the screen layer and on the new behaviours that justify this PR (combined-balance threshold, focus-aware modal, auth-gated settings, inline mnemonic error).

| File | LOC | Notes |
|---|---|---|
| `app/screens/spark-onboarding/restore/restore-phrase-screen.tsx` | ~200 | Inline `invalidMnemonic` rendering, reserved-height container, focus-forwarding `useEffect`, `wrong={showError}` propagation — no screen-level spec exists. Hook spec covers `useRestorePhrase` only. |
| `app/screens/home-screen/home-screen.tsx` | ~640+ | New `useIsFocused()` gate on `BackupNudgeModal` is the bug-fix, but no home-screen spec exists or was added. If `useIsFocused` is wired to the wrong tree, the PR's regression returns. |
| `app/screens/settings-screen/settings-screen.tsx` | (modified) | New `useUnacknowledgedNotificationCountQuery({ skip: !isAuthed })` and the new "Recovery method" `SettingsGroup` rendering have no screen-level test. `skip-queries-when-unauthed.spec.tsx` covers leaf components only. |
| `__tests__/i18n/*` (locale parity) | n/a | 28 locales updated with `recoveryMethod` / `invalidMnemonic` / `Phrase.headerTitle`; missing keys silently fall back to English. No contract test asserting per-locale parity. |

Each should be made to **fail** on current code first, then fixed alongside the production fix.

|---|---|---|
| `__tests__/hooks/use-backup-nudge-state.spec.ts:24-31` | `useTotalBalance` mock only sums BTC: `walletCurrency === "BTC" ? sum + w.balance : sum`. Every existing assertion would also pass against the pre-PR BTC-only implementation. The "USD weight crosses the threshold" regression — the headline behaviour change in this file — is not actually exercised. | (combined-balance regression claim) |
| `__tests__/screens/spark-onboarding/backup-confirm-screen.spec.tsx` | `useMigrationCheckpoint` is mocked synchronously (`mockCheckpoint.mockReturnValue(...)`) — the async-load race that mis-routes migration-with-funds users is invisible. | #1 |
| `__tests__/screens/spark-onboarding/hooks/use-backup-confirm.spec.ts` | `hasCompleted.current` latch isn't tested with the de-correct → re-correct sequence; auto-advance disabling silently is uncovered. | #6 |
| `__tests__/self-custodial/providers/wallet-provider.spec.tsx` | Snapshot-call hang (no timeout) is not tested; `Loading + Unknown` deadlock not tested; retry-with-old-SDK stale-write race not tested. The suite covers the easy enum transitions but not the failure modes that re-introduce the cold-start hang. | #2, #4, #5 |
| `__tests__/self-custodial/providers/is-online.spec.ts` | Walks the `getServiceStatus` abort path well, but does not test that **`getOnlineState` itself can hang** (no timeout). | #3 |
| `__tests__/self-custodial/bridge/status.spec.ts:14` | Mock signature `getSparkStatus: () => mockBreezGetSparkStatus()` discards the `{ signal }` argument — so the bridge's signal-forwarding contract isn't actually verified end-to-end. | I8 |

|---|---|
| #1 | `backup-confirm-screen` test: `useMigrationCheckpoint` returns `{ checkpoint: null, loading: true }` with `BackupStatus.None` and funded wallet → drive `allCorrect` to true → assert no navigation fires until `loading` flips to `false`. Then flip `checkpoint` to `"backupAlerts"` → assert `sparkMigrationTransferringFunds` is the destination. |
| #2 | `use-sdk-lifecycle` test: `getSelfCustodialWalletSnapshot` returns a never-resolving promise → assert status transitions to `Offline`/`Error` within `STATUS_TIMEOUT_MS + ε`, not stuck in `Loading`. |
| #3 | `is-online` test: `getSparkStatus` (the bare path used by `getOnlineState`) hangs → assert `getOnlineState` resolves to `Unknown` within timeout, not indefinitely. |
| #4 | `use-sdk-lifecycle` test: prev = `Loading`, snapshot rejects, `getOnlineState` returns `Unknown` → assert status becomes `Error` (or `Stalled`), not preserved as `Loading`. |
| #5 | `use-sdk-lifecycle` test: trigger `retryCount` increment while a refresh against the old SDK is in flight; resolve the old refresh after the new SDK is connected → assert `setWallets`/`setStatus` are not called with the stale snapshot. |
| #6 | `use-backup-confirm` test: drive `allCorrect` true → wait < 400 ms → mutate one word to be wrong → mutate it back to correct → assert `onComplete` fires (latch reset). |
| #7 | `settings-screen` test: render with `activeAccount.type !== SelfCustodial` → assert `getByText(LL.SettingsScreen.recoveryMethod())` is **not** present (no orphan empty group). |
| (combined-balance) | `use-backup-nudge-state` test: rewrite mock to combine BTC + USD-as-sats; case with BTC alone under threshold + USD weight pushing combined balance over `backupNudgeModalThreshold` → assert `shouldShowModal === true`. |
| (focus-aware modal) | `home-screen` spec: mock `useIsFocused` × `shouldShowModal` matrix; assert `<BackupNudgeModal isVisible>` follows `shouldShowModal && isFocused`. |
| (auth-gated settings + recovery method group) | `settings-screen` spec under `IsAuthedContextProvider value={false}` → assert `useUnacknowledgedNotificationCountQuery` was called with `skip: true`; under `value={true}` with self-custodial account → assert "Recovery method" group renders containing "View backup phrase". |
| (locale parity) | Parameterised test iterating all 28 locales: load locale, assert `LL.SettingsScreen.recoveryMethod()` and `LL.RestoreScreen.invalidMnemonic()` resolve to non-empty strings. |

---

1. **Land the Critical fixes in this PR** — #1–#7 are all small and scoped; none of them require touching the wider stack. #2/#3/#4 form a single hardening pass on the new lifecycle, do them as one commit.
2. **Write the Tier 3 regression tests alongside the production fixes** — each Critical should have a failing test against current code before the fix lands, then green after.
3. **Fix the Tier 2 false-confidence tests** — un-mock `useTotalBalance` to reflect real BTC+USD combination; switch the `useMigrationCheckpoint` mock to expose the `loading` flag; un-discard `{ signal }` in the bridge spec mock. These are quality fixes that prevent the next regression from sneaking past CI.
4. **Add the Tier 1 missing screen specs** — `restore-phrase-screen`, the home-screen modal-gating spec, and a screen-level `settings-screen` spec for the auth-gate + recovery-method group.
5. **Important items I1–I8** — most are small (3–10 lines each); fold the observability ones (I1, I2, I4, I5) into the same commit as the lifecycle hardening, so the new error paths are also instrumented.

esaugomez31:
@grimen

## Critical

### 1. Migration routing race vs. AsyncStorage load

`useBackupConfirm` now accepts an optional `disabled` flag and short-circuits its auto-navigate effect while it is true. The screen consumes `loading` from `useMigrationCheckpoint` and forwards it as `disabled`, so the 400 ms timer cannot fire (and the manual button stays disabled) until AsyncStorage settles. A migration-with-funds user who completes the words via paste/autofill no longer skips `sparkMigrationTransferringFunds`; the route decision waits for `checkpoint` to be authoritative before running.

### 2. `getSelfCustodialWalletSnapshot` has no timeout/abort

The snapshot await in `useSdkLifecycle.refreshWallets` is now wrapped in a `withTimeout` race that rejects with a labelled error after `STATUS_TIMEOUT_MS` (the same 5 s budget already used by `getServiceStatus`). When the SDK hangs on `getWalletInfo`/`listPayments`, the race fires, the rejection lands in the existing `catch` (which records it to crashlytics and lets the connectivity-fallback decide between Offline/Error), and the existing `finally` releases `refreshingRef` so subsequent refreshes are no longer coalesced behind a stuck call. The new helper lives in `app/utils/with-timeout.ts` so the same pattern (already inlined in the logout flow) has a single, tested implementation.

### 3. `getOnlineState()` has no timeout

`getOnlineState` now arms an `AbortController` with the same `STATUS_TIMEOUT_MS` budget that `getServiceStatus` already uses, forwards the signal to `getSparkStatus`, and clears the timer in `finally`. The shape of the function is unchanged: success still maps to `Online`/`Offline`, the existing catch still routes through `reportSparkStatusFailure` and returns `Unknown`. The recovery-path probe can no longer hang; on timeout it aborts, falls into the existing catch, and the caller in `useSdkLifecycle` keeps its three-state contract (`Online`/`Offline`/`Unknown`) intact so the "preserve previous status" branch still applies when connectivity is genuinely indeterminate.

### 4. `Loading + Unknown` connectivity is a dead-end state

The `Unknown` branch in the snapshot-failure `setStatus` updater now transitions out of `Loading` instead of preserving it: when the previous status is `Loading` (cold start) and connectivity is indeterminate, it returns `Error`, which the UI already surfaces with a retry button. All other previous statuses keep the existing "preserve previous status" behaviour, so a `Ready` session that briefly loses connectivity still survives a flaky probe. A new `Stalled` enum value would have churned the entire status reducer, mappers and translations for one edge case; reusing the existing `Error` state keeps the change to a single line and pulls the user out of the indefinite-loading dead-end the PR set out to fix.

### 5. Stale-state write race on retry

`refreshWallets` now treats the SDK reference captured at call time as the session epoch. A small `isStale = () => sdkRef.current !== sdk` predicate gates the two writes that follow an `await`: the success-path setters (after the snapshot resolves) and the catch-path `setStatus` updater (after the connectivity probe). When `retryCount` increments and the cleanup of the previous effect run reassigns `sdkRef.current`, an in-flight refresh against the disconnected SDK now bails before touching React state, so a late-resolving stale snapshot can no longer overwrite the fresh session's wallets/status. No new ref or generation counter is introduced; the SDK identity already changes exactly when the session changes, which is the same invariant the bug is about.

### 6. `hasCompleted.current` latches forever

The auto-advance effect now resets `hasCompleted.current = false` whenever `allCorrect` flips back to false, so a transient de-correct (paste-then-edit, typo correction inside the 400 ms window) no longer disables the marquee auto-advance for the rest of the session. The "no double-fire" guarantee is preserved by an explicit `if (hasCompleted.current) return undefined` guard before the timer starts; the timer body itself is unchanged.

### 7. Empty "Recovery method" group renders for non-self-custodial users

`ViewBackupPhraseSetting` was moved into the `securityAndPrivacy` array next to `TotpSetting` and `OnDeviceSecuritySetting`, and the standalone "Recovery method" group was removed. Custodial users no longer see an orphan section with an empty card; self-custodial users see "View backup phrase" sit naturally under "Security & privacy" alongside the rest of the security-related rows. The standalone group was the cheapest place the bug could surface; collapsing it into the existing security group keeps the surface narrower and doesn't depend on `SettingsGroup` learning to skip empty children.

## Important

### I1. Bare `catch {}` in is-online.ts

Already addressed in a follow-up commit on the upstream branch that landed after this review was submitted. Both `getServiceStatus` and `getOnlineState` now route their catch through a shared `reportSparkStatusFailure` helper that records to crashlytics (deduped per session under `spark-status-fetch-failed`) before returning the sentinel. No further action needed in this PR; the fix is inherited from the parent branch.

### I2. `crashlytics().log` is a breadcrumb, not a recordable error

Audited the four call sites individually rather than swapping every `.log` for `.recordError`. Three are kept as breadcrumbs by design: `logging.ts` dispatches Info/Warn through `.log` and only Error through `.recordError`; `convert.ts` already pairs a `.log` (context) with a follow-up `.recordError` (the actual error); and `use-sdk-lifecycle.ts:117` is a side-note breadcrumb inside a catch path whose error was already recorded a few lines above. The genuine offender was `payment-adapter.ts` line 89 in `createGetFee`, where the catch only emitted `.log` and the failure was invisible in crashlytics search. That call site now wraps the rejection in an `Error` (preserving the existing `[SparkSDK] Fee quote failed` prefix) and routes it through `crashlytics().recordError`. Covered by a new assertion in `payment-adapter.spec.ts` that mocks `prepareSendPayment` to reject and verifies `recordError` was called once with the same message.

### I3. Snapshot-success unconditionally sets `Ready`

Documented trade-off rather than adding a second probe. The snapshot's resolution already proves the SDK reached the network and produced valid data, so a follow-up `getServiceStatus()` call would only add latency to every poll/AppState refresh while duplicating signal that already arrives from the catch path on op failure. The "degraded but still functional" case is intentionally not surfaced from the success path: today nothing in the UI consumes a `Degraded` signal, and once it does the right home for that subscription is the SDK event listener (which fires service-status changes asynchronously), not a synchronous probe inside `refreshWallets`. No code change shipped under I3 because the existing behaviour is correct given those constraints.

### I4. Bare `catch {}` in `use-payment-request.ts`

The `} catch {` in `generateRequest` now binds the error and forwards it to `crashlytics().recordError` (wrapping non-Error rejections in a labelled `Error`) before setting `PaymentRequestState.Error`. The user-visible state machine is unchanged; the silent reject path is no longer invisible to crashlytics. Covered by a new assertion in `use-payment-request.spec.ts` that mocks the adapter to reject with a labelled message and asserts `recordError` was called with the same message.

### I5. `KeyStoreWrapper.deleteMnemonic` silent catch leaves partial mnemonic

`deleteMnemonic` now records both failure modes to crashlytics: the inner `.catch(() => {})` on the network-key removal binds the error and reports it (the function still returns `true` because the primary mnemonic was successfully removed; the network leftover is a soft failure but no longer invisible), and the outer catch records the primary-removal failure before returning `false`. The boolean contract callers depend on stays the same; the partial-state scenario is now traceable in crashlytics. Covered by three new assertions in `secure-storage-mnemonic.spec.ts`: primary-removal failure is recorded and returns false, network-key failure is recorded but still returns true, and the success path records nothing.

### I6. Redundant `restore().catch(() => {})`

Investigated and held back from this PR. The pattern lives in two screens (`restore-method-screen` and `use-restore-phrase`) and the underlying `useRestoreWallet.restore()` re-throws after toast/state/crashlytics handling so callers can choose how to react. Removing the throw is the structurally correct fix, but `feat--self-custodial-launch-polish` retains the same `throw err` and the same `.catch(() => {})` callers in its multi-account refactor of `use-restore-wallet.ts`, so changing this here would conflict during restack and bounce back to the same shape. Once polish merges and the multi-account pattern is settled, the `throw err` can be dropped in a follow-up and the silenced catches will become structurally unnecessary across both branches.

### I7. PR description claim vs. actual change



### I8. AbortSignal forwarding test

The bridge spec mock now forwards every argument to `mockBreezGetSparkStatus` instead of dropping them, and two new assertions cover the contract: when a caller passes a signal, the SDK is invoked with `{ signal }`; when none is passed, the SDK is invoked without arguments. The signal-forwarding path that protects `getServiceStatus`/`getOnlineState` from indefinite hangs is now verified end-to-end at the bridge boundary.

## Test coverage

### Tier 1 (production code with no test file)

Four new spec files now cover surfaces that previously had none:

- `__tests__/i18n/locale-parity.test.ts` walks all 27 locales and asserts the three backup/restore keys touched by this PR (`SettingsScreen.recoveryMethod`, `RestoreScreen.invalidMnemonic`, `BackupScreen.ManualBackup.Phrase.headerTitle`) exist and are non-empty, so a missing translation in any locale fails CI instead of shipping as an English fallback.
- `__tests__/screens/home.spec.tsx` was extended (rather than splitting into a new file) with a `BackupNudgeModal focus gating` block that renders the real `HomeScreen` and asserts the matrix `useIsFocused × shouldShowModal`. The first pass added a separate spec that rendered an inline gate component, which only re-tested a JS boolean AND; the production gate now lives behind a real `HomeScreen` render and a mocked `BackupNudgeModal` that captures its `isVisible` prop, so the assertion exercises the actual binding at the JSX call site.
- `__tests__/screens/spark-onboarding/restore-phrase-screen.spec.tsx` covers the inline `invalidMnemonic` rendering, the `wrong=true/false` propagation to `MnemonicWordInput`, the restoring spinner, the error screen with retry CTA, and the custom `validationError` text fallback.
- `__tests__/screens/settings-screen/settings-screen.spec.tsx` was extended with two assertions that `useUnacknowledgedNotificationCountQuery` is invoked with `skip: true` when unauthenticated and `skip: false` when authenticated, locking in the auth-gated query behaviour the settings refactor relied on.

### Tier 2 (tests that pass while a Critical bug is alive)

For each Critical, a class of existing assertions would have kept passing while the bug was live because they never modelled the specific race or edge condition the bug exploited. Rather than retrofit them into brittle race simulators, every Critical commit lands a dedicated regression next to the existing happy-path test, so the original suite keeps its narrow contract and the new test pins down the bug directly.

- **Critical #1 (migration routing race):** `useBackupConfirm` already had `calls onComplete when all words are correct`, which presumed the consumer was always ready to navigate. It would have stayed green even when the screen was wired with `disabled` permanently true. The new `does not auto-complete while disabled, then fires once unlocked` assertion in `use-backup-confirm.spec.ts` plus the screen-level `does not auto-navigate while the migration checkpoint is still loading` in `backup-confirm-screen.spec.tsx` close the gap by toggling the loading flag inside the test.
- **Critical #2 (snapshot timeout):** `wallet-provider.spec.tsx` already had refresh-failure assertions where the snapshot rejected synchronously, so the `.catch` ran and the suite was happy even though a hanging snapshot would never reach the catch in production. The new `recovers from a snapshot that hangs past the timeout instead of staying in Loading` test mocks a never-resolving promise and asserts the `withTimeout` race transitions out of Loading.
- **Critical #4 (`Loading + Unknown` dead-end):** `preserves Ready status when connectivity is 'unknown'` covered the Ready→preserve case and would have passed unchanged with the dead-end alive on cold start. The new `transitions out of Loading to Error when both snapshot and connectivity` assertion forces the previous-status to `Loading`, runs both probes through the unhappy path, and verifies the reducer escapes to `Error`.
- **Critical #5 (stale-write race):** Pre-existing tests asserted that a successful refresh wrote `wallets`/`status`. They had no notion of an SDK swap mid-refresh, so a stale write from a disconnected SDK would have landed without flagging the test. The new `Critical #5 — stale-write safety` block flips `sdkRef.current` mid-await and asserts the late resolver is dropped before any setter runs.
- **Critical #6 (auto-advance latch):** `calls onComplete when all words are correct` only covered the linear fill path and never re-triggered the effect. The new `re-arms the auto-advance timer after a de-correct then re-correct sequence` test types a wrong word into a previously-correct slot and back, then asserts the timer fires exactly once on the second correction.
- **Critical #3 (online state timeout):** The legacy `returns 'unknown' when getSparkStatus throws` test only covered the synchronous reject path. The new `aborts and returns 'unknown' when the SDK call hangs past the timeout` assertion uses fake timers plus a never-resolving SDK to verify the AbortController fires and the catch path runs.
- **Critical #7 (empty Recovery method group):** Existing settings-screen tests exercised the SC=true path where the group rendered as expected. They never asserted what custodial users saw, so the empty-card regression was invisible. The new `does not render a standalone Recovery method group` assertion renders the custodial path and `queryByTestId` proves the group is gone.

### Tier 3 (Critical bugs with no test coverage anywhere)

Every Critical landed with a dedicated regression assertion in the same commit, so the suite catches a future revert before it ships:

- **Critical #1** — `__tests__/screens/spark-onboarding/hooks/use-backup-confirm.spec.ts` (`does not auto-complete while disabled, then fires once unlocked`) plus the screen-level race in `__tests__/screens/spark-onboarding/backup-confirm-screen.spec.tsx` (`does not auto-navigate while the migration checkpoint is still loading`).
- **Critical #2** — `__tests__/self-custodial/providers/wallet-provider.spec.tsx` (`recovers from a snapshot that hangs past the timeout instead of staying in Loading`) plus the helper unit test in `__tests__/utils/with-timeout.spec.ts`.
- **Critical #3** — `__tests__/self-custodial/providers/is-online.spec.ts` (`forwards an AbortSignal to getSparkStatus` and `aborts and returns 'unknown' when the SDK call hangs past the timeout`).
- **Critical #4** — `__tests__/self-custodial/providers/wallet-provider.spec.tsx` (`transitions out of Loading to Error when both snapshot and connectivity` are unhappy on cold start) sitting next to the pre-existing Ready-preserve case.
- **Critical #5** — `__tests__/self-custodial/providers/wallet-provider.spec.tsx` (`SelfCustodialWalletProvider — stale-write safety (Critical #5)`) covers the SDK-swap race for both the success and catch paths.
- **Critical #6** — `__tests__/screens/spark-onboarding/hooks/use-backup-confirm.spec.ts` (`re-arms the auto-advance timer after a de-correct then re-correct sequence`).
- **Critical #7** — `__tests__/screens/settings-screen/settings-screen.spec.tsx` (`does not render a standalone Recovery method group`).

The bridge-level `forwards every argument to the underlying SDK` assertions added under I8 also act as a defence-in-depth backstop for Criticals #2/#3, since the AbortSignal escape hatch only works if the bridge actually forwards it.

grimen (dismissed):

grimen:
### Merge activity

* **May 20, 10:43 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762).
* **May 20, 10:46 AM UTC**: [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762) couldn't merge this pull request because a downstack PR #3742 failed to merge.
* **May 20, 10:52 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762).
* **May 20, 10:55 AM UTC**: [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762) couldn't merge this pull request because a downstack PR #3746 failed to merge.
* **May 20, 12:23 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762).
* **May 20, 12:23 PM UTC**: @grimen merged this pull request with [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3762).

grimen (approved):

Issue #645, opened 2022-10-07
Formatting issues in BBW address confirmation screen


daviroo:
https://twitter.com/cryptstock9/status/1578357339954241537?s=46

nicolasburtey:
duplicated

PR #3768, opened 2026-04-29
feat(spark): production rollout and SDK hardening


esaugomez31:
## Summary

Production-readiness layer for the self-custodial (Spark) experience: settings screens, multi-account orchestration, SDK lifecycle hardening, feature-flag rollback, contacts adapter, send-flow integration for SC, error translation, and auto-convert.

Primary ticket: blinkbitcoin/blink-wip#645 — *feature(spark): Rollout, Settings, and Hardening*. The work also touches adjacent tickets (#646 stable balance / convert, #647 payments) where send-flow and convert-flow integration was needed for SC parity.

## Ticket #645 — Tasks

- [x] **6.1 Self-custodial settings screens** — Account Information, Bitcoin Deposit, and Transaction Limits screens; recovery method group; profile row with delete-wallet flow; custodial-only settings rows hidden in SC mode.
- [x] **6.2 Account switching and multi-account coexistence** — `useAccountRegistry` hook orchestrates multi-account selection across the app.
- [x] **6.3 SDK lifecycle hardening** — polling, backoff, retries; `network-status-banner` for offline / degraded network.
- [x] **6.4 Feature-flag rollback and data preservation** — `temporarily-unavailable` SC state preserves wallet data while allowing graceful rollback.
- [x] **6.6 Contacts adapter** — provider-agnostic contact bridge wired into contacts UI and send-bitcoin.
- [x] **6.7 Backend-only feature gating** — `BackendFeatureGate` component hides custodial-only tabs (Circles, Earn, Card explanations) for SC users.
- [x] **6.8 Observability + support patterns** — SDK error classifier + translator hook + i18n strings for self-custodial errors.

## Additional work (beyond #645 strict scope)

- **Send flow (Spark integration)** — USD-wallet Lightning sends use SDK `conversionOptions`; LNURL pay support via `prepareLnurlPay` bridge wrappers and dedicated payment-details; LNURL `successAction` plumbed through confirmation; Spark labels surfaced in destination headers and transaction detail.
- **Send-flow fixes** — skip Intraledger optimization for Blink LNURL; skip lnurl-pay invoice fetch when irrelevant; classified SC errors translated at confirmation time.
- **Convert / Stable Balance** — exact-input BTC↔USDB convert with DisplayAmount; auto-convert BTC→USDB after Lightning receive; sync wallet after convert to surface token balance; remove obsolete balance-stale heuristic.
- **Receive polish** — hide receive-toggle icon when wallet is locked.
- **i18n** — translations across 28 locales for new SC rollout copy, settings, contacts, delete flow, errors, and convert.

## Test plan

- [ ] Switch between custodial and SC accounts; verify settings show the correct rows for each mode.
- [ ] Open Account Information, Bitcoin Deposit, and Transaction Limits screens in SC mode.
- [ ] Disable network mid-session; confirm `network-status-banner` and graceful SDK degradation.
- [ ] Toggle remote `nonCustodialEnabled` off; confirm SC wallet enters `temporarily-unavailable` and data persists.
- [ ] Send via Lightning from BTC wallet, USD wallet, and to a Blink LN address (intraledger and external).
- [ ] Send via LNURL with success action returned.
- [ ] Trigger SDK errors (invalid invoice, insufficient funds, network); confirm translated copy appears.
- [ ] Auto-convert: receive Lightning → BTC swept to USDB; verify token balance.
- [ ] Contacts: add and pay an SC contact; confirm provider-agnostic behavior.
- [ ] Verify custodial regressions: Circles, Earn, Card flows, send/receive flows still work.


esaugomez31:
* **#3768** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

> [!IMPORTANT]
> **#3768 carries the same structural debt called out on #3758, plus net-new debt added by this PR's screen-layer integration.** As before, this review covers correctness only — the structural / SOLID / complexity / naming review remains a separate post-stack cleanup ticket against `main`. Findings below are net-new to #3768 (or surface in code first introduced here):
>
> - **SOLID violations.**
>   - **`useContacts` returns `ContactAdapter & { loading: boolean }`** (`app/hooks/use-contacts.ts:8`) — the loading state is bolted on with an intersection rather than declared in the contract; every consumer has to remember to spread it. ISP/SRP smell: the adapter type doesn't fully describe its observable state.
>   - **`useCustodialContactAdapter` declares `add` / `delete` / `update.paymentIdentifier` then throws `unsupported(...)` at runtime** (`app/custodial/contact-adapter.ts:16-17, 78-80, 82-84`). LSP-ish: the adapter formally implements the interface but partially refuses it. Either capabilities-gate the methods (which the type already exposes via `capabilities: { canAdd, canDelete, ... }` — so just narrow the method shape too) or split into a smaller `ReadOnlyContactAdapter`.
>   - **Three overlapping status enums** for what is the same conceptual axis: `WalletStatus = Loading | Ready | Offline | Degraded | Error | Unavailable` (`app/types/wallet.types.ts`), `CreationStatus = Idle | Creating | Error` (`use-create-wallet.ts:16-22`), plus auto-convert's own status set. `Loading` vs `Idle` and two different `Error`s mean the same string lives in two enums with different semantics.
>   - **Per-account secureStorage is implemented as 5 free functions on the global key store** (`app/utils/storage/secureStorage.ts:~194`) instead of a typed `MnemonicStore` adapter with key isolation enforced by construction. Carries the same "module of stringly-keyed verbs" pattern that #3758 flagged for adapters.
>
> - **Incomplete adapter pattern.** `useContacts()` IS clean — returns one `ContactAdapter` and picks SC/custodial under the hood. **But the contacts UI bypasses it.** `all-contacts.tsx:57-106` and `contacts-card.tsx` declare `const isSelfCustodial = activeAccount?.type === AccountType.SelfCustodial`, then run **two parallel data paths** (`useContactsQuery({ skip: !isAuthed || isSelfCustodial })` for custodial and a separate `useState<Contact[]>` + `useFocusEffect` calling `bridgeListContacts(sdk)` for SC), then a `useMemo` picks one. The adapter exists; the call sites don't trust it. Same shape as the screen-layer mode-leak previously called out — newly added in this PR.
>
> - **High cyclomatic complexity in the screen layer.**
>   - **70 net-added `isSelfCustodial` branches** across **14 production files** in this PR alone (`git diff <base>..HEAD | grep "^+.*isSelfCustodial"`).
>   - `home-screen.tsx` (756 LOC, +37 in this PR): **19 hook calls** in the body prelude before the first `return`, plus mode branches on `isSelfCustodial` for username title, wallet selection, banner visibility, and switch-account UX.
>   - `all-contacts.tsx` (315 LOC) and `contacts-card.tsx` (233 LOC): mode-leak described above; the parallel state + parallel query + mode-pick triple is repeated almost verbatim in both files.
>   - `account-screen.tsx` and `settings-screen.tsx`: row visibility forks per mode rather than via `BackendFeatureGate` (which exists for exactly this purpose but is not consistently applied).
>   - `network-error-component.tsx` introduces a *third* spelling — `isSelfCustodialActive` — for the same boolean used as `isSelfCustodial` everywhere else.
>
> - **Code smells.**
>   - Empty-`.catch(() => …)` swallow with no telemetry: `is-online.ts`, `wallet-provider.tsx` (LN address resolve, line ~115), `secureStorage.ts` per-account methods (5 sites), `use-save-lnaddress-contact.ts` SC branch, `use-sdk-lifecycle.ts:114, 122`. The pattern hides incidents — already flagged on #3758, **persisted unchanged here** and extended to new surfaces.
>   - `useFocusEffect` + `useEffect` doing similar work concurrently in the new contacts screens — same race shape called out for `unclaimed-deposit-banner` on #3758.
>   - `type X = (typeof X)[keyof typeof X]` const-as-enum pattern (`CreationStatus`) used inconsistently — some statuses are real `enum`-replacement objects, others are string-literal unions, others are the `WalletStatus` enum-like.
>   - Container-noun modules: `auto-convert/storage`, `storage/account-index`, `bridge/wallet`, `bridge/lifecycle` — repeats #3758's `Info` / `Result` / `Fn` / `helpers` / `utils` cluster.
>
> - **Naming smells (~30 in this PR alone).**
>   - **24 `selfCustodial`-prefixed exports inside `app/self-custodial/`** — the namespace already says it. `listSelfCustodialAccounts`, `addSelfCustodialAccountId`, `findSelfCustodialAccountByMnemonic`, `selfCustodialCreateWallet`, `selfCustodialRestoreWallet`, `mapSelfCustodialTransaction(s)`, `useSelfCustodialContactAdapter`, `useSelfCustodialWallet`, `useSelfCustodialAccountInfo`, `useSelfCustodialConversionLimits`, `getSelfCustodialWalletSnapshot`, `SelfCustodialAccountEntry`, `SelfCustodialErrorCode`, `SelfCustodialPaymentRequestState`, `createSelfCustodialLightningPaymentDetails`, `createSelfCustodialOnchainPaymentDetails`, `createSelfCustodialLnurlPaymentDetails`, `createSelfCustodialSparkPaymentDetails`, etc. Inside the namespace, `list()` / `add()` / `findByMnemonic()` would suffice.
>   - **Bridge-import aliasing as a symptom of leak**: `import { listContacts as bridgeListContacts } from "@app/self-custodial/bridge"` (and `bridgeListPayments`, `bridgeAddContact`, `bridgeDeleteContact`, `bridgeUpdateContact`) — call sites rename imports to `bridge*` because they collide with same-verb custodial/Apollo paths in the same module. The cure is the adapter pattern; #3758 said the same.
>   - Mode-prefixed hooks/state at call sites: `useSelfCustodialRollback`, `useSelfCustodialUnavailable`, the `selfCustodialContacts` state inside the contacts UI, `selfCustodialFragments` in the transaction-history screen.
>   - Inconsistent boolean spelling: `isSelfCustodial` (most files) vs `isSelfCustodialActive` (`network-error-component.tsx`) — same signal, two names.
>   - Suffix containers: `RollbackResult` (`use-self-custodial-rollback.ts`), `ConversionLimitsState`, `BackupState`, `CreationStatus` — `Result` / `State` / `Status` containers for shapes that could carry domain names.
>
> **Why this is not in scope for this PR (same logic as #3758):** #3768 sits at the top of the same stack. Restacking a refactor pass through #3761–#3769 would be multi-day pain — particularly the rename passes against the 24 redundant prefixes and the 70 `isSelfCustodial` branches. Refactoring against `main` *after* the stack settles is materially cheaper.
>
> **Decision: this review covers correctness only.** Structural cleanup remains the post-stack ticket. **However — #3768 is the rollout PR.** Per the prior #3758 note, production rollout was supposed to be gated on the cleanup ticket reaching at least the adapter-pattern unification milestone. The contacts-UI bypass added by this PR makes that gating *more* important, not less: shipping the rollout now normalises the parallel-data-path pattern in the screen layer and makes the eventual cleanup harder. Recommend either (a) the team explicitly waiving the #3758 gate before merge, *or* (b) the cleanup ticket landing the contact-adapter unification before #3768 ships.

---

Real strengths first: the bridge-layer rollback on SDK init failure during restore (`bridge/lifecycle.spec.ts` covers it explicitly), `auto-convert/{storage,executor}.spec.ts` are exemplary — fail-open behavior, tolerance bands, full status-transition matrix, dedup + prune. `is-online.spec.ts` walks every `ServiceStatus`. `wallet-provider.spec.tsx` covers Ready/Offline/Error/Degraded transitions, AppState polling, coalescing, lifecycle disconnect. `use-delete-self-custodial.spec.ts` covers all 4 post-delete branches plus partial-failure resilience. `use-restore-wallet.spec.ts` covers existing-mnemonic dedup AND in-flight-guard reentrancy. Persistent state migrations 7→8→9 are pure additive widenings. `BackendFeatureGate` and `network-status-banner` cleanly cover all branches.

Severity:
- **Critical** — fund-loss, mis-attribution, data-integrity loss, or silent failure that misleads the user. Must fix before merge.
- **Important** — correctness/reliability that doesn't lose funds but degrades UX, hides incidents, or sets up future bugs.
- **Test** — coverage gap or a test that passes while a confirmed bug is alive.

---

## Critical

### 1. Cold-start race silently flips authed SC users to custodial — **and persists it**

`app/config/feature-flags-context.tsx:285-286` + `app/hooks/use-self-custodial-rollback.ts:25-33`

`FeatureFlagContextProvider` defaults `nonCustodialEnabled = false` and only blocks rendering until remote config returns when `currentLevel === "NonAuth"`:

```ts
// feature-flags-context.tsx:285-286
if (!remoteConfigReady && currentLevel === "NonAuth") {
  return null
}
```

For authed users, `useLevelQuery` is `cache-only`, so `currentLevel` resolves to the cached value on the first frame and the gate is not taken. Children render with the default `false` flag. `useSelfCustodialRollback` then runs:

```ts
// use-self-custodial-rollback.ts:25-33
useEffect(() => {
  if (nonCustodialEnabled) return
  if (activeAccount?.type !== AccountType.SelfCustodial) return
  const fallback = accounts.find((a) => a.type === AccountType.Custodial)
  if (!fallback) return
  setActiveAccountId(fallback.id)
}, [nonCustodialEnabled, activeAccount, accounts, setActiveAccountId])
```

`setActiveAccountId` writes to persistent state via `updateState`. By the time remote config arrives and flips `nonCustodialEnabled = true`, the active-account switch is already written to AsyncStorage. The effect is one-directional — there is no "snap back" branch.

A user with both a custodial token and a self-custodial wallet who chose SC as their active account gets silently bumped to custodial **on every cold start**. Re-selecting SC doesn't survive the next launch. Users with only SC briefly see `TemporarilyUnavailableScreen` until remote config arrives. Directly violates the stated rollback guarantee — the *user-intent* data being lost is itself rollback data.

**Fix:** gate the rollback on `remoteConfigReady` (expose readiness from the provider), or extend the render gate to authed users (`if (!remoteConfigReady) return null` regardless of auth). Treat the default flag value as "unknown," not "off."

### 2. SDK lifecycle cleanup never awaits `disconnectSdk` and never calls `removeSdkEventListener`

`app/self-custodial/providers/use-sdk-lifecycle.ts:213-218`

```ts
return () => {
  mounted = false
  if (backoffTimerRef.current) { clearTimeout(...); ... }
  if (sdkRef.current) {
    disconnectSdk(sdkRef.current)   // not awaited; no removeSdkEventListener
    sdkRef.current = null
    setSdk(null)
    setConnectedAccountId(null)
  }
}
```

The bridge exports `removeSdkEventListener` — it is never called outside tests. On rapid account switches (custodial↔SC, multi-SC) a new `initSdk` can fire while the previous one is still resolving; the listener (closed over the old `mounted` flag and account ID) lives on. `connectAndListen`'s `if (!mounted) return` guard runs *before* writing to `sdkRef.current`, so a late resolver still latches the new SDK after cleanup ran — and the cleanup that already executed observed the old or null ref. Stale `PaymentSucceeded` / wallet-update events fire `refreshWallets` against the wrong active-account context — money-adjacent.

**Fix:** capture the listener id from `addSdkEventListener` in a ref and call `removeSdkEventListener` in cleanup. Track the in-flight init via an abort token so a late-resolving init disconnects itself when stale.

### 3. Mnemonic normalisation mismatch creates duplicate accounts at the same Spark identity

`app/self-custodial/bridge/lifecycle.ts:87` vs `app/self-custodial/storage/account-index.ts:90, 94`

```ts
// bridge/lifecycle.ts:87
const trimmed = mnemonic.trim().replace(/\s+/g, " ")
```
```ts
// storage/account-index.ts:90,94
const trimmed = mnemonic.trim()
...
if (stored?.trim() === trimmed) return entry.id
```

Restore stores `trim() + collapse-whitespace`. Lookup only `trim()`s. Tabs / double spaces / trailing newline-after-trim / NFC quirks miss the dedup → fresh `accountId`, fresh storage dir, **same on-chain identity**. Two index entries pointing at one wallet, separate active toggles, undefined SDK behavior when both storage dirs activate. Manual word-grid is safe today; cloud restore, paste paths, and any future entry surface are not.

**Fix:** define a single `normalizeMnemonic(s) = s.trim().replace(/\s+/g, " ")` helper used in both write and lookup, including on `stored` in the comparison. Adding a unit test pinning the prefix `${MNEMONIC}:${accountId}` while you're there avoids a rename silently orphaning everyone's wallet on upgrade.

### 4. Per-account secureStorage methods (mnemonic store) have **zero tests**

`app/utils/storage/secureStorage.ts:~194` — `getMnemonicForAccount` / `setMnemonicForAccount` / `deleteMnemonicForAccount` / `{get,set}MnemonicNetworkForAccount`

This is the physical key store for SC wallets. Every method silently swallows errors (`.catch(() => null/false)`). A regression here can:

- silently lose a freshly-generated mnemonic on Keychain failure (= lost funds, no recovery surface),
- leak one account's mnemonic into another via key collision,
- fail to clear on delete (privacy / security regression).

There is no `__tests__/utils/storage/secureStorage.spec.ts`. The methods are only indirectly mocked at call sites.

**Fix:** add a spec asserting (a) key isolation across `accountId`s, (b) success/failure return values, (c) `deleteMnemonicForAccount` removes BOTH the mnemonic and the network-label keys, (d) the prefix `MNEMONIC:${accountId}` is pinned. Critical #3 fix and Critical #4 share a test surface — land them together.

### 5. `useInFlightGuard` is untested → double-create-wallet risk

`app/hooks/use-in-flight-guard.ts` (no spec exists)

Wraps `selfCustodialCreateWallet` and `selfCustodialRestoreWallet`. A failure mode here means a second invocation generates **two wallets with two UUIDs** but only one is persisted as active — the orphan holds funds the user can't access. `use-restore-wallet.spec.ts` covers reentrancy at the integration level for restore; `use-create-wallet.spec.ts` does not, and there is no isolated test for the guard's invariants.

**Fix:** add a spec covering: second concurrent `run()` returns `undefined` and never invokes its operation; `isRunning()` flips during operation and resets after; guard releases on both resolution and rejection; separate hook instances have independent guards.

### 6. The flag driving the wallet-lockout navigator branch (`shouldShowUnavailable`) has zero behavioural coverage

`app/hooks/use-self-custodial-rollback.ts:35-40` ↔ `__tests__/hooks/use-self-custodial-rollback.spec.ts`

This boolean drives a top-level branch in `app/navigation/root-navigator.tsx` — when `true`, the user sees `TemporarilyUnavailableScreen` instead of the entire app. The existing spec asserts the `setActiveAccountId` side effect but **never asserts on the returned `shouldShowUnavailable` value**, and never mocks `useHasCustodialAccount`. The companion `use-self-custodial-unavailable.spec.ts` mocks both `useAccountRegistry` AND `useSelfCustodialRollback`, reducing the hook under test to "did it forward arguments?" — a tautology.

**This is the same untested seam where Critical #1 lives.** No test catches that bug today.

**Fix:** see Tier 3 below — `flag=false, active=SC, hasCustodial=false → true`; `flag=false, active=SC, hasCustodial=true → false`; `flag=true → false`; active=custodial → false; transition `false → true` (no flap).

---

## Important

| # | Item | Location |
|---|---|---|
| I1 | Creating a second SC wallet wipes the **previous** wallet's "backup completed" state. `resetBackupState()` runs before `updateState({ activeAccountId: newId })`; `BackupStateProvider.persist` keys on `activeSelfCustodialAccountId` from React context, which still resolves to the OLD account at that instant — `backupState:<old-id>` overwritten with default state. UX-trust signal flipped to "phrase not safe" on the prior wallet. | `app/screens/spark-onboarding/hooks/use-create-wallet.ts:41-45` + `app/self-custodial/providers/backup-state-provider.tsx` (`persist` reads from context) |
| I2 | `useDeleteSelfCustodial` calls `disconnectSdk(sdk)` before `setActiveAccountId(...)`. `useSdkLifecycle` still has `sdkRef.current` set; the 10s poll and backoff retry hammer a disconnected SDK until the active-account change propagates. `refreshWallets` errors flip status to `Offline` mid-deletion; UI gates that depend on `isSelfCustodial = status !== Unavailable` flap. | `app/screens/settings-screen/account/multi-account/hooks/use-delete-self-custodial.ts:50, 75, 81` + `app/self-custodial/providers/use-sdk-lifecycle.ts:115-123, 232-237` |
| I3 | `useSelfCustodialAccountInfo` and `useSelfCustodialConversionLimits` initialise `loading: true` and only flip to `false` inside an effect that early-returns when `sdk` is null. Opening "Account Information" or "Transaction Limits" while Offline/Unavailable/Error renders an infinite `<ActivityIndicator />` — directly contradicts the rollout doc's offline UX guarantees, in exactly the screens users open *to* troubleshoot offline state. | `app/self-custodial/hooks/use-self-custodial-account-info.ts:25, 38, 46` + `…/use-self-custodial-conversion-limits.ts:26` |
| I4 | Schema 8→9 migration: `selfCustodialDefaultWalletCurrency` is propagated by spread but the round-trip is never asserted in `state-migrations.spec.ts`. A user who already adopted SC and set USD as default can have it silently dropped on upgrade. | `app/store/persistent-state/state-migrations.ts` ↔ `__tests__/store/persistent-state/state-migrations.spec.ts` |
| I5 | `useActiveWallet` now returns `isReady=true` for both `Ready` and `Degraded`. Many call sites gate UI / payment availability on `isReady`. The behavior change is unasserted in `use-active-wallet.spec.ts` — regressing Degraded→`isReady=false` blocks payments for degraded users; regressing the inverse hides degradation. | `app/hooks/use-active-wallet.ts` ↔ `__tests__/hooks/use-active-wallet.spec.ts` |
| I6 | `useSaveLnAddressContact` self-custodial branch is entirely uncovered. The spec hardcodes `isSelfCustodial = false`, so the SC code path — including `bridgeAddContact` happy/error paths and the silent-failure swallow that returns `{ saved: false, handle }` while logging — is invisible. The screen-level `send-confirmation.spec.tsx` mocks the hook out, so doesn't compensate. | `app/screens/send-bitcoin-screen/use-save-lnaddress-contact.ts` ↔ `__tests__/hooks/use-save-lnaddress-contact.spec.ts` |
| I7 | SDK reconnect backoff (`RECONNECT_BACKOFF_MS = [1000, 3000, 9000]`) is untested. `wallet-provider.spec.tsx` covers Ready→Offline transitions but never advances fake timers. Off-by-one on `failureCountRef`, missing `clearTimeout` on success, or scheduling after unmount can produce a thundering herd or no recovery. Overlaps Critical #2 — same lifecycle module. | `app/self-custodial/providers/use-sdk-lifecycle.ts:115-123` |
| I8 | `account-index.ts` (canonical SC account list + one-shot legacy migration `selfCustodialAccountIds[]` → `selfCustodialAccountIndex[]` + `findSelfCustodialAccountByMnemonic`) has **no spec**. Bugs here mean lost accounts post-migration or duplicate entries. Same file as Critical #3. | `app/self-custodial/storage/account-index.ts` |
| I9 | `validate-network` records to crashlytics AND emits an SDK log event AND returns `false` on mismatch. The spec asserts only the boolean — a refactor that drops `crashlytics().recordError(…)` would silently kill the team's primary mainnet/regtest mix-up signal. | `__tests__/self-custodial/providers/validate-network.spec.ts` |
| I10 | `BackupStateProvider` account-switch transition untested. Backup status is keyed per-account (`backupState:${accountId}`); on active-account change, the provider must reload the new account's state and not leak the previous account's status during the transition. Failure means the nudge banner doesn't appear when it should. `removeBackupStateFor` (used by `useDeleteSelfCustodial`) is also exported but not directly tested. | `app/self-custodial/providers/backup-state-provider.tsx` ↔ `__tests__/self-custodial/providers/backup-state-provider.spec.tsx` |
| I11 | Auto-convert `isRetryableNow` 30s cooldown / 2-min orphan-timeout is the gate that decides when a previously-attempted convert is retried on app launch. The listener spec covers max-attempts and Failed retry, but not the time-based gate. Money-handling code with no time-gate test = under-retry (orphans) or over-retry (rate-limits). | `app/self-custodial/hooks/use-auto-convert-listener.ts` |
| I12 | `feature-flags-context` rollout-exposure logging — new effect logs `self_custodial_rollout_exposed` once per session via `rolloutLoggedRef`. If the deduping ref is wrong, every remote-config refresh re-fires the analytics event, polluting funnel data + cost metrics. No spec exists. | `app/config/feature-flags-context.tsx` |
| I13 | `use-save-session-profile` now writes `activeAccountId = DefaultAccountId.Custodial` on every successful `saveProfile`. If broken, users who sign into Blink while a SC account was active stay on the wrong account post-sign-in (SC wallet visible but the new auth token is custodial). No spec exists. | `app/hooks/use-save-session-profile.ts` |
| I14 | `wallet-provider.tsx` `resolveAndPersist` LN-address path: existing spec mocks `getLightningAddress` to return `null`, so the persistence + `reloadSelfCustodialAccounts` subpath (lines 91-115) is never exercised. The error path is wrapped in a swallowing `.catch` that still chains the reload — verify whether that's intended. | `app/self-custodial/providers/wallet-provider.tsx:91-115` ↔ `__tests__/self-custodial/providers/wallet-provider.spec.tsx` |
| I15 | `extractLightningAddressUsername` (new) and 4 new SC analytics helpers (`logSelfCustodialBackupCompleted`, `…RestoreCompleted`, `…StableBalanceActivated`, `…RolloutExposed`) have no tests. The snake_case payload-key rename (`nonCustodialEnabled → non_custodial_enabled`) is exactly the kind of thing that drifts silently. | `app/utils/pay-links.ts`, `app/utils/analytics.ts` |

---

## Test coverage

Estimated behavioural coverage of #3768's net-new code: **~55-60%** — strong on bridge / providers / auto-convert; sparse on root-level hooks (rollback, in-flight-guard, save-session-profile), secure storage, account-index, and the contacts UI surface.

| File | LOC | Notes |
|---|---|---|
| `app/utils/storage/secureStorage.ts` (new per-account methods) | ~80 added | Critical #4 — physical mnemonic keystore, every method silently swallows |
| `app/hooks/use-in-flight-guard.ts` | ~30 | Critical #5 — gates wallet-create / restore against double-invocation |
| `app/self-custodial/storage/account-index.ts` | ~120 | I8 — canonical SC account list + legacy migration + mnemonic dedup |
| `app/hooks/use-save-session-profile.ts` | ~40 modified | I13 — switches `activeAccountId` on sign-in |
| `app/config/feature-flags-context.tsx` | (modified) | I12 — new rollout-exposure logging effect (`rolloutLoggedRef`) |
| `app/utils/pay-links.ts` (`extractLightningAddressUsername`) | ~10 added | I15 |
| `app/utils/analytics.ts` (4 new SC helpers) | ~40 added | I15 |
| `app/utils/error-logging.ts` | ~15 | Trivial wrapper but the only crashlytics surface for SC paths |
| `app/screens/people-screen/contacts/contacts-card.tsx` | (modified) | New SC branching + SDK error-swallow |
| `app/screens/people-screen/contacts/all-contacts.tsx` | (modified) | New SC branch via `useFocusEffect` |
| `app/self-custodial/bridge/wallet.ts` | ~50 | Trivial passthroughs, but no direct test |

These create false confidence. Each should be made to **fail** on current code first, then fixed alongside the production fix.

|---|---|---|
| `__tests__/hooks/use-self-custodial-rollback.spec.ts` | Spec only asserts the `setActiveAccountId` side effect; never asserts on `shouldShowUnavailable` and never mocks `useHasCustodialAccount`. The cold-start race (where remote config defaults to `false`, the rollback writes, and the snap-back never happens) is untested at the seam where it lives. | #1, #6 |
| `__tests__/hooks/use-self-custodial-unavailable.spec.ts` | Mocks both `useAccountRegistry` AND `useSelfCustodialRollback` — the hook under test is reduced to forwarding mocked arguments. Provides no behavioural coverage at all. Adjacent to the seam where #1 lives. | #1, #6 |
| `__tests__/store/persistent-state/state-migrations.spec.ts` | `activeAccountId` round-trip asserted; `selfCustodialDefaultWalletCurrency` round-trip never asserted. Schema-8 user with USD default loses it on migrate to 9. | I4 |
| `__tests__/persistent-storage.spec.ts` | "Migrates persistent state from schema 0" passes `schemaVersion: 0` which isn't in `stateMigrations` — exercising the default-fallback path, not migration. The test name lies; either delete or replace with real schema-3 / 4 / 7 / 8 round-trips. | (test quality) |
| `__tests__/hooks/use-save-lnaddress-contact.spec.ts` | Hardcodes `isSelfCustodial = false`. The entire SC branch — including `bridgeAddContact` happy + error paths and the silent-failure swallow — is invisible. | I6 |
| `__tests__/self-custodial/providers/wallet-provider.spec.tsx` | Covers Ready/Offline/Error/Degraded transitions but never advances fake timers — the 1s/3s/9s reconnect backoff schedule, success-resets-counter, unmount-clears-timer, max-attempts-cap are all unasserted. | #2 (lifecycle leak), I7 |
| `__tests__/self-custodial/providers/validate-network.spec.ts` | Asserts the boolean return only. `crashlytics().recordError(...)` side effect on mismatch — the production canary for catastrophic mainnet/regtest mix-ups — would survive deletion. | I9 |
| `__tests__/hooks/use-active-wallet.spec.ts` | Existing tests cover Ready / Unavailable; the new `Degraded → isReady=true` mapping (a behavior change in this PR) has no direct assertion. | I5 |
| `__tests__/people-screen/contacts/legacy-contact-shims.spec.ts` | Happy-path only; no `paymentIdentifier=""` or SDK contacts where `name` is missing — both real cases coming from the SDK that map silently to empty `alias` / `handle`. | (test quality) |
| `__tests__/self-custodial/bridge.spec.ts` | Pre-dates `bridge/lifecycle.spec.ts` split; both test create/restore. The newer file is more thorough. Either consolidate or document the split intent — keeping both creates drift risk. | (test quality) |

For each, no existing test exercises the behavior. Add a regression test alongside the fix.

| Critical | Recommended regression test |
|---|---|
| #1 | Test the `useSelfCustodialRollback` *transition*: mount with `nonCustodialEnabled = false` (default value before remote config), `activeAccount = SC`, custodial fallback present → assert `setActiveAccountId` is **not** called yet (gated on `remoteConfigReady`). Then advance to `remoteConfigReady = true`, `nonCustodialEnabled = true` → assert no rollback ever fired. Without the gating fix the test must demonstrate the problematic behavior. |
| #2 | `useSdkLifecycle` test: mount with account A → switch to account B before A's `initSdk` resolves → A's `removeSdkEventListener` is called AND A's listener never fires `refreshWallets` after the switch. Verify `disconnectSdk(A)` completes before `initSdk(B)` writes to `sdkRef.current`. |
| #3 | `findSelfCustodialAccountByMnemonic` test: store via `selfCustodialRestoreWallet` with `"  word1\tword2  word3  "` → look up with the same input AND with `"word1 word2 word3"` AND with `" word1  word2 word3"` → all three return the same `accountId`. Pin `normalizeMnemonic` as a single source of truth. |
| #4 | `secureStorage` per-account spec: `setMnemonicForAccount("a", X)` then `getMnemonicForAccount("b") === null` (key isolation); rejection of underlying store → method returns `false` not throws; `deleteMnemonicForAccount` removes both mnemonic + network keys; refactor-pin on the key prefix. |
| #5 | `useInFlightGuard` direct spec: second `run()` while first is in flight returns `undefined` and the operation is invoked exactly once; guard releases on both resolve and reject; instances are independent. AND a `useCreateWallet` integration test mirroring the existing restore reentrancy test. |
| #6 | Direct `useSelfCustodialRollback` spec asserting all 5 cases of `shouldShowUnavailable` (see #6 "Fix" above), composing the real hook (only mocking `useFeatureFlags` + `useHasCustodialAccount` primitives). Delete the tautological `use-self-custodial-unavailable.spec.ts`. |

### Tier 4 — Important items without coverage anywhere

| Important | Recommended test |
|---|---|
| I1 | `useCreateWallet` test: starting with active SC account A whose `backupState:A` is "completed" → run `create()` for new account B → `backupState:A` is **unchanged** (only `backupState:B` is reset). |
| I2 | `useDeleteSelfCustodial` test: assert order — `setActiveAccountId` is called BEFORE `disconnectSdk` / `KeyStoreWrapper.deleteMnemonicForAccount`, OR assert no `refreshWallets` fires after `disconnectSdk` resolves. |
| I3 | `useSelfCustodialAccountInfo` and `useSelfCustodialConversionLimits` — `sdk = null` → `loading: false` and an explicit unavailable/empty state, NOT initial `loading: true` indefinitely. |
| I7 | Fake-timer test for `RECONNECT_BACKOFF_MS`: first failure → 1000ms; second → 3000ms; third → 9000ms; fourth → no schedule; success resets counter; unmount clears timer. |
| I9 | `validate-network` mismatch → `crashlytics().recordError(expect.objectContaining({ message: stringContaining("Network mismatch") }))`. |
| I10 | `BackupStateProvider`: rerender with active account B (different `useAccountRegistry` mock) → loads `backupState:B` and A's status does not leak; while loading, status is default (None), not A's stale value. Plus a direct `removeBackupStateFor` test. |
| I11 | `isRetryableNow`: cooldown (5s ago → skip), eligible (35s ago), orphan fast-path (3min ago), first-attempt (`undefined` → eligible). |

---

1. **Block merge on Critical #1.** It is silent, persistent, and directly contradicts the rollout's data-preservation guarantee. Land the gating fix together with the regression test from Tier 3 #1 — the seam is the same spec file flagged in Tier 2.
2. **Critical #2, #3** ship with this rollout; both sit in money-adjacent code (lifecycle event delivery; account-identity dedup). Land their fixes with Tier 3 #2 and #3 in the same commits.
3. **Critical #4, #5** are pure test gaps — write the specs now. They cost ~half a day combined and lock down two of the most security-critical seams in the PR (mnemonic store, in-flight guard).
4. **Critical #6** delete `use-self-custodial-unavailable.spec.ts` and replace with the direct `useSelfCustodialRollback` spec from Tier 3 #6 — same scope, real coverage. Same PR as #1.
5. **Important I1–I3** are confined to small, identifiable seams in this PR — fold each into the Critical-fix commits where the lifecycle is already touched.
6. **Important I4, I5, I6** are test-only and small — same PR. **I7, I8, I10, I11** are also test-only but larger; OK as a follow-up if the team prefers, but I7 (backoff) and I10 (backup-state account switch) sit close to money paths and are worth doing now.
7. **I9, I12, I13, I14, I15** are belt-and-suspenders — follow-up PR is fine.
8. **Structural / SOLID / naming feedback** — remains the post-stack cleanup ticket against `main` per the prior #3758 review note. Production rollout #3768 should still be gated on the cleanup ticket reaching the adapter-pattern unification milestone, otherwise the structural debt called out previously normalises.

esaugomez31:
@grimen

## Critical

### 1. Cold-start race silently flips authed SC users to custodial — and persists it

`FeatureFlagContext` now exposes `remoteConfigReady: boolean` as part of its public shape — `false` until `fetchAndActivate` resolves (success or catch in the existing `finally` clause that already drove `setRemoteConfigReady(true)`). `useSelfCustodialRollback`'s effect adds `if (!remoteConfigReady) return` as its first guard, treating the default flag value as "unknown" rather than "off". The `shouldShowUnavailable` lockout signal also gates on `remoteConfigReady` so the boot window never flaps the user into `TemporarilyUnavailableScreen`. The render gate at `feature-flags-context.tsx:285` was left intact (still NonAuth-only) — went with the surgical option (expose readiness from the provider, rather than extend the render gate to authed users) to avoid a white-screen on every authed cold start while still treating the boot window as undecided. Covered by three new tests in `__tests__/hooks/use-self-custodial-rollback.spec.ts` "cold-start race (Critical #1)": ("does not roll back while remoteConfigReady is false, even with the default flag value") asserts the boot window is inert; ("does not roll back when remote config arrives with the flag enabled") drives the canonical race — `false` default, then arrival with `nonCustodialEnabled: true` — and asserts no rollback ever fires; ("rolls back exactly once when remote config arrives with the flag disabled") asserts the post-ready legitimate rollback still works. Without the gating fix the second test demonstrates the pre-fix bug.

### 2. SDK lifecycle cleanup never awaits `disconnectSdk` and never calls `removeSdkEventListener`

`useSdkLifecycle` now tracks the listener id from `addSdkEventListener` in a dedicated `listenerIdRef` and signals in-flight `initSdk` via `abortRef`. The cleanup path goes through a single `teardownSdk(sdk, listenerId)` helper that awaits `removeSdkEventListener` first and then `disconnectSdk` — the same helper is reused at the two abort-bail sites inside `connectAndListen` (post-init and post-listener-registration), so the remove-then-disconnect ordering lives in one place. A late-resolving `initSdk` for account A that lands after the cleanup ran observes `abortRef.current === true` and disconnects the orphan SDK without writing to `sdkRef.current` or registering a listener — stale `PaymentSucceeded` / wallet-update events can no longer fire `refreshWallets` against the new account context. Covered by two new tests in `__tests__/self-custodial/providers/wallet-provider.spec.tsx`: ("aborts a late-resolving initSdk if unmount happens before it lands") drives `mockInitSdk` with a deferred Promise, unmounts mid-flight, resolves after, and asserts `disconnectSdk` was called and `addSdkEventListener` was never called against the stale SDK; ("removes the listener and disconnects when abort fires after listener registration but before refs land") drives the variant where `mockAddSdkEventListener` is the deferred — asserts both `removeSdkEventListener` and `disconnectSdk` were called with the correct listener id. The pre-existing "disconnects SDK on unmount" was extended to also assert `removeSdkEventListener` is called before `disconnectSdk`.

Folded a SOLID extraction alongside: `RECONNECT_BACKOFF_MS` scheduling and counter management are now in `app/self-custodial/hooks/use-backoff-retry.ts` exposing `{ schedule, reset }`. The lifecycle hook calls `scheduleBackoffRetry(() => { if (sdkRef.current) refreshWallets() })` in the catch and `resetBackoff()` on success and unmount — `failureCountRef` and `backoffTimerRef` are privatized inside the new hook, and the unmount cancellation runs through the hook's own `useEffect(() => reset, [reset])`. Closes Important I7 in the same surface: `__tests__/self-custodial/hooks/use-backoff-retry.spec.ts` covers the fake-timer schedule sequence (1000 → 3000 → 9000 → no-op), reset rewinding, schedule replacing pending timer, and unmount auto-cancel.

### 3. Mnemonic normalisation mismatch creates duplicate accounts at the same Spark identity

Extracted a single `normalizeMnemonic(s) = s.trim().replace(/\s+/g, " ")` helper at `app/utils/mnemonic.ts` (project-wide flat-helper convention — sibling of `bip39-wordlist.ts`, `phone.ts`, `format-public-key.ts`; not coupled to `self-custodial/`). Both consumers now go through it: `selfCustodialRestoreWallet` (write side — used for `validateMnemonic`, `setMnemonicForAccount`, AND `initSdk` so the SDK seeds the same normalized form that the index will compare against) and `findSelfCustodialAccountByMnemonic` (lookup — both `mnemonic` input AND `stored` value are normalized before the equality check). Tabs / double spaces / trailing newline / mixed runs all collapse to the same canonical form on both sides, so the dedup gate finally holds. Covered by three specs: `__tests__/utils/mnemonic.spec.ts` pins the helper's contract directly (trim, collapse runs, mixed whitespace, single-space preserve, whitespace-only → empty, idempotent — six cases); `__tests__/self-custodial/storage/account-index.spec.ts` "findSelfCustodialAccountByMnemonic (Critical #3)" walks the matrix at the lookup site (exact match, leading/trailing whitespace, collapsed runs of tabs+multi-space, stored noisy with leading newline, no match, no stored mnemonic); `__tests__/self-custodial/bridge/lifecycle.spec.ts` "selfCustodialRestoreWallet (I14)" gains two new whitespace-variant tests (tabs+newlines, mixed runs) alongside the original space-only assertion. The pre-fix bug (lookup used `trim()` only while restore used `trim() + collapse-whitespace`) would surface immediately on any of the four whitespace-mismatch tests across the two specs.

Closes Important I8 in the same surface — `app/self-custodial/storage/account-index.ts` had zero direct coverage. The new `__tests__/self-custodial/storage/account-index.spec.ts` walks all five public verbs (`listSelfCustodialAccounts`, `listSelfCustodialAccountIds`, `addSelfCustodialAccountId`, `removeSelfCustodialAccountId`, `setSelfCustodialLightningAddress`) plus the canonical-index/legacy-id-list dual format: parsed entries from the canonical index, malformed-entry filtering (bad id type, missing id, bad lightningAddress shape), JSON parse-error tolerance, one-shot legacy migration (legacy `["a","b"]` → canonical `[{id:"a",lightningAddress:null},…]` AND persisted via `setItem`), non-string entries from the legacy list dropped, append-with-null-LN-address, dup no-op, filter no-op when id is absent, set-LN-address writes / no-ops on unchanged-value / no-ops on unknown id / clears via `null`. Twenty-four tests covering both Critical #3 and I8.

### 4. Per-account secureStorage methods (mnemonic store) have zero tests

Extended `__tests__/utils/secure-storage-mnemonic.spec.ts` with a new "per-account mnemonic methods (Critical #4)" block — 15 tests covering the four assertions you listed (a–d) for all five verbs (`getMnemonicForAccount`, `setMnemonicForAccount`, `deleteMnemonicForAccount`, `getMnemonicNetworkForAccount`, `setMnemonicNetworkForAccount`). **(a) Key isolation across `accountId`s**: `getMnemonicForAccount` / `setMnemonicForAccount` each have an "isolates accounts by hitting/writing a different key per id" test driving alice and bob through `mockGet`/`mockSet` and asserting `toHaveBeenNthCalledWith(1, "mnemonic:alice", …)` / `toHaveBeenNthCalledWith(2, "mnemonic:bob", …)`; `deleteMnemonicForAccount` adds "never touches the global 'mnemonic' / 'mnemonic_network' keys" asserting the legacy global keys, sibling account keys, and the network-key for the wrong id are all NOT touched. **(b) Success/failure return values**: every verb has both a success path (return `true` or the stored value) and a failure path (Keychain rejects → silent fallback to `false` / `null`) — the silent-failure surface is now structurally pinned, so a regression that throws instead of returning the documented sentinel breaks the test immediately. **(c) `deleteMnemonicForAccount` removes BOTH the mnemonic and the network-label keys**: "removes both 'mnemonic:{id}' and 'mnemonic_network:{id}', returns true" asserts both `mockRemove` calls; "returns true even when the network-key removal fails (tolerated)" pins the existing `.catch(() => {})` swallow on the network key (primary success + network failure → still `true`); "returns false when the primary mnemonic removal fails" asserts the inverse — primary failure short-circuits to `false` before even attempting the network key. **(d) Prefix pinning**: every per-account test uses the literal string `"mnemonic:alice"` / `"mnemonic_network:alice"` rather than reconstructing the key from the constant — a rename of `KeyStoreWrapper.MNEMONIC` from `"mnemonic"` to anything else now breaks `expect(mockGet).toHaveBeenCalledWith("mnemonic:alice")` immediately, so an upgrade can't silently orphan everyone's wallet by re-namespacing. Total spec count is now 31 tests across the global + per-account surfaces; Critical #3's helper guarantees normalization parity, Critical #4's tests guarantee the underlying key store actually stores and retrieves what the helper hands it.

### 5. `useInFlightGuard` is untested → double-create-wallet risk

Added direct + integration coverage; the hook implementation already satisfied the four invariants (the `useRef(false)` + `try/finally` pattern), so this commit pins them against future regression rather than fixing a behavioural bug. Direct spec at `__tests__/hooks/use-in-flight-guard.spec.ts` (5 tests, one per invariant + a happy-path baseline): "resolves to the operation's result on first call" (baseline); **invariant 1** "returns undefined and never invokes the operation when a previous run is still in flight" — drives `firstOp` as a deferred Promise, calls `run(secondOp)` while it's pending, asserts the second call resolves to `undefined` AND `secondOp` was never invoked AND `firstOp` was called exactly once; **invariant 2** "flips isRunning() during the operation and resets it after resolution" — asserts `isRunning() === false` before, `=== true` after `run` is invoked but before its operation resolves, `=== false` after the operation resolves; **invariant 3** "releases the guard on rejection so a follow-up run() can proceed" — drives the operation to throw, asserts `run` rejects AND `isRunning()` resets to `false` AND a subsequent `run()` actually invokes its operation and returns its value (proves the `finally` block runs); **invariant 4** "isolates the guard per hook instance" — two `renderHook(() => useInFlightGuard())` instances, asserts that while instance A has a deferred operation in flight, instance B's `isRunning()` reads `false` and a `run()` on B invokes its operation and resolves normally without interference. Integration test at `__tests__/screens/spark-onboarding/hooks/use-create-wallet.spec.ts` "ignores reentrant create while one is already in flight (Critical #5)" mirrors `use-restore-wallet.spec.ts:148` exactly: drives `mockCreateWallet` as a deferred Promise, calls `create()` synchronously (status flips to `Creating`), then calls `create()` a second time and asserts `mockCreateWallet` was invoked exactly once AND none of the success-path side effects fired (`mockUpdateState`, `mockDispatch`, `mockReinitSdk`, `mockResetBackupState` all not called) — closes the asymmetry the review flagged ("`use-restore-wallet.spec.ts` covers reentrancy at the integration level for restore; `use-create-wallet.spec.ts` does not"). The double-UUID orphan-wallet failure mode now has both unit-level invariant coverage AND the call-site integration coverage.

### 6. The flag driving the wallet-lockout navigator branch (`shouldShowUnavailable`) has zero behavioural coverage

Deleted the tautological `__tests__/hooks/use-self-custodial-unavailable.spec.ts` (the hook is a 3-line wrapper; the spec mocked both `useAccountRegistry` AND `useSelfCustodialRollback`, reducing the test to argument forwarding). Replaced with the direct 6-case `shouldShowUnavailable` matrix on `useSelfCustodialRollback` itself, mocking only the abstractions the hook actually depends on (`useFeatureFlags`, `useHasCustodialAccount`). The matrix covers: flag=disabled + active=SC + no custodial fallback → `true` (locks user out); flag=disabled + active=SC + custodial fallback present → `false` (rollback handles it); flag=enabled regardless of fallback → `false`; active=custodial → `false`; `remoteConfigReady=false` → `false` (boot window inert); transition `false → true` on remote-config arrival without flap. The seam where Critical #1 lived now has real coverage instead of a tautological adjacent.

---

## Important

### I1. `useCreateWallet` `resetBackupState()` ordering wipes previous wallet's backup state

Removed the `resetBackupState()` call (and the surrounding `useBackupState` import + destructuring + useCallback dep) from `useCreateWallet`. The call wasn't fixable by reordering — `BackupStateProvider.persist` is a `useCallback` that closes over `activeSelfCustodialAccountId`, so even calling `updateState({ activeAccountId: newId })` first, the ref still resolves to the OLD id at the moment `resetBackupState()` runs (state updates aren't synchronous and the closure was captured at the prior render). The provider's existing `useEffect` at `backup-state-provider.tsx:86-103` already loads the new account's persisted state on `activeSelfCustodialAccountId` change — for a freshly-created `Crypto.randomUUID()` accountId there's no prior record, so `readBackupState` returns null and the provider falls back to `defaultState`, which is exactly what `resetBackupState()` was meant to achieve. The redundant-and-harmful call is gone; `backupState:<old-id>` stays intact across the create flow. `__tests__/screens/spark-onboarding/hooks/use-create-wallet.spec.ts` was updated alongside: dropped the `mockResetBackupState` declaration, the `jest.mock("@app/self-custodial/providers/backup-state-provider", …)` block, and three `expect(mockResetBackupState).…` assertions; renamed "reinits SDK and resets backup state on success" → "reinits SDK on success".

### I2. `useDeleteSelfCustodial` calls `disconnectSdk(sdk)` before `setActiveAccountId`

Reordered `deleteWallet` so the active-account switch happens before `disconnectSdk(sdk)`. `useSdkLifecycle`'s accountId-keyed effect now observes the new active id immediately and tears down the outgoing SDK through its own cleanup path, so the 10s poll and backoff retries no longer have a window to hammer the stale `sdkRef`. The explicit `disconnectSdk(sdk)` is still called (kept to release the storage dir before `RNFS.unlink`), but it now runs after the lifecycle has already started teardown — it is effectively idempotent. The three branches (remaining SC, custodial fallback, no fallback → clear `activeAccountId`) now run their `setActiveAccountId` / `updateState` call up-front; the navigation phase at the end of the flow no longer repeats the active-account write. `__tests__/screens/settings-screen/multi-account/use-delete-self-custodial.spec.ts` adds three Important #2 ordering tests, one per branch, each asserting via `mock.invocationCallOrder` that the active-account write fires before `disconnectSdk` (and the remaining-SC test additionally asserts `disconnectSdk` fires before `deleteMnemonicForAccount`, pinning the full chain `switch → disconnect → delete`). The seven existing tests (initial state, no-custodial → getStarted, custodial fallback → Primary, remaining SC, sdk-null skip, disconnect-rejects-and-continues, deleteMnemonic-fails error capture) continue to pass without modification — semantic behavior is preserved end-to-end, only the ordering is fixed.

### I3. `useSelfCustodialAccountInfo` / `useSelfCustodialConversionLimits` infinite spinner offline

Both hooks now surface a sentinel `Error("Self-custodial wallet unavailable")` instead of staying stuck in `loading: true`. When the effect observes `sdk === null` (Offline / Unavailable / Error states) it sets `loading: false` + the unavailable-error in a single `setInfo` / `setState` call and returns; consumers' existing `if (loading) → spinner` / `if (error) → loadError text` branches now route to the error message instead of the indicator. The reverse transition (`sdk` going from `null` back to a value) is also pinned: the effect calls `setInfo((prev) => ({ ...prev, loading: true, error: null }))` before kicking off the fetch, so the screen flips back to the spinner during a real load instead of leaving the prior unavailable-error visible mid-fetch. This unblocks the exact screens users open *to* troubleshoot offline state — Account Information and Transaction Limits — and the same fix applies to every consumer of `useSelfCustodialAccountInfo` (also `account-fields.tsx`, `id.tsx`, `use-backup-methods.ts`, `use-cloud-backup.ts`); consumers that read only `identityPubkey` are unaffected because the empty-string fallback is unchanged. Tier 2 tests that documented the old loading-forever contract were replaced with the fixed contract: each spec now has "flips to loading=false with an unavailable-error when sdk is null (Important #3: no infinite spinner offline)" and "flips back to loading=true and clears the unavailable-error when sdk transitions from null to a value", on top of the existing happy-path / error-path tests.

### I4. Schema 8→9 migration: `selfCustodialDefaultWalletCurrency` round-trip unasserted

Added four targeted tests in `__tests__/store/persistent-state/state-migrations.spec.ts` covering the field's round-trip across the migration boundary: schema 8 → 9 with `selfCustodialDefaultWalletCurrency: "USD"`, schema 8 → 9 with `selfCustodialDefaultWalletCurrency: "BTC"`, schema 8 → 9 with the field absent (asserts `undefined` survives without being coerced), and schema 9 → 9 round-trip. All four also assert `schemaVersion === 9` post-migration so a future spread regression that drops or zeroes the field would surface immediately. Spec count went from 8 → 12; the existing schema-3-through-8 chain coverage is unchanged.

### I5. `useActiveWallet` `isReady=true` for both `Ready` and `Degraded` unasserted

Added 6 new tests to `__tests__/hooks/use-active-wallet.spec.ts`: two positive tests pinning `isReady === true` when status is `Degraded` (one per accountType — SelfCustodial and Custodial — to cover both branches of `resolveBaseState`), plus a parametrised negative loop over the four non-Ready / non-Degraded statuses (`Loading`, `Error`, `Offline`, `Unavailable`) that asserts `isReady === false` and that `result.current.status` faithfully echoes the source state. The Degraded SC test also asserts `isSelfCustodial === true` so the existing call-site gate `isSelfCustodial = status !== Unavailable` is preserved alongside the new readiness contract. A regression that drops Degraded from the Ready list would block payments for degraded users; a regression that adds Loading/Error/Offline to Ready would hide degradation. Both surface immediately. Spec count went from 4 → 10.

### I6. `useSaveLnAddressContact` self-custodial branch entirely uncovered

Restructured `__tests__/hooks/use-save-lnaddress-contact.spec.ts` to make `isSelfCustodial` and `sdk` parameterisable via top-level `let` mocks (previously hardcoded to `false` / `null`), then added a dedicated "self-custodial branch (Important #6)" describe block with 4 tests: (1) "saves the contact through the bridge when sdk is available" pins the happy path — `bridgeAddContact` invoked with `{ name: handle, paymentIdentifier: handle }` and `{ saved: true, handle }` returned; (2) "returns saved=false with the handle and logs to crashlytics when bridgeAddContact rejects" pins the silent-swallow that the call site relies on (`crashlytics().log` is called with the handle in the message, the return value carries the handle so the caller can still surface it to the user); (3) "returns saved=false without a handle when sdk is null" pins the early-return for the offline path — bridge is not called AND the Apollo mutation is not called (no fallback); (4) "does not call the Apollo mutation when self-custodial is active even on the happy path" guards against a regression that would double-write the contact via both paths. Spec count went from 4 → 8.

### I7. SDK reconnect backoff `RECONNECT_BACKOFF_MS = [1000, 3000, 9000]` untested

Closed in the Critical #2 commit by extracting the scheduler into `useBackoffRetry` and adding 7 fake-timer tests in `__tests__/self-custodial/hooks/use-backoff-retry.spec.ts` covering: schedule fires after the first delay (1000ms boundary); walks the delay sequence on successive calls (1000 → 3000 → 9000); becomes a no-op after exhaustion; schedule replaces a pending timer with the next delay; reset rewinds the counter and cancels any pending timer; cancels any pending timer on unmount; empty delays array no-ops. Off-by-one on counter, missing `clearTimeout` on success, or scheduling-after-unmount all surface immediately.

### I8. `account-index.ts` (canonical SC account list + legacy migration) has no spec

Closed in the Critical #3 commit. `__tests__/self-custodial/storage/account-index.spec.ts` covers every public verb plus the canonical/legacy dual format — see Critical #3 above for the full test enumeration.

### I9. `validate-network` mismatch crashlytics canary unasserted

The mock surface in `__tests__/self-custodial/providers/validate-network.spec.ts` was already in place but the existing assertions only covered the boolean return; the crashlytics + SDK-log side effects were silent. Promoted the inline `recordError: jest.fn()` and `logSdkEvent: jest.fn()` to top-level `mockRecordError` and `mockLogSdkEvent` references and added two dedicated tests: "records the mismatch to crashlytics with the wallet/config networks in the message (Important #9)" pins the `recordError` payload — asserts `toBeInstanceOf(Error)`, message contains `"Network mismatch"`, `"wallet=mainnet"`, AND `"config=regtest"`; "emits an SDK log event at Error level with the mismatch message (Important #9)" pins `logSdkEvent` was called once with the `"error"` level and the same wallet/config strings in the message. The two pre-existing pass-through cases (no stored network, matching network) gained negative assertions ensuring `recordError` and `logSdkEvent` are NOT called when the wallet is in sync. A refactor that drops either the `crashlytics().recordError(...)` call or the `logSdkEvent(SdkLogLevel.Error, ...)` call would now fail the spec immediately. Spec count went from 3 → 5.

### I10. `BackupStateProvider` account-switch transition untested

Refactored `__tests__/self-custodial/providers/backup-state-provider.spec.tsx` so the `useAccountRegistry` mock is parameterisable via a top-level `mockActiveAccount` (previously hardcoded to one account), then added two new describe blocks covering the gaps. "account-switch transition (Important #10)" has 4 tests: (1) reloads the new account's persisted state when `activeAccount.id` changes — drives `mockGetItem` to return distinct payloads per key (`backupState:test-sc-uuid` → manual, `backupState:other-sc-uuid` → cloud), reads the first, switches the mock, rerenders, asserts `getItem` was called with the new key and the visible state matches the new payload; (2) falls back to default state when the new account has no persisted record — first account has data, second account doesn't, after switch state goes to `BackupStatus.None`; (3) clears the backup state when the active account becomes non-self-custodial — switching to a custodial account drops state to default (no leak across the SC→Custodial transition); (4) scopes `setBackupCompleted` writes to the new active account after a switch — calls `setBackupCompleted("cloud")` post-switch, asserts the write hits `backupState:other-sc-uuid` and never touches `backupState:test-sc-uuid`. "removeBackupStateFor (Important #10)" has 2 tests: removes under the namespaced `backupState:{accountId}` key, and never touches the legacy `backupState` global key or other accounts' keys. Spec count went from 6 → 12.

### I11. Auto-convert `isRetryableNow` 30s cooldown / 2-min orphan-timeout untested

Added a dedicated `useAutoConvertListener — isRetryableNow gate (Important #11)` describe block in `__tests__/self-custodial/hooks/use-auto-convert-listener.spec.ts` with `jest.useFakeTimers()` + `jest.setSystemTime(FIXED_NOW)` so `Date.now()` reads deterministically inside the listener. Five tests pin the gate from both entry points: live-trigger blocked when `lastAttemptAtMs = FIXED_NOW - 5_000` (within `RETRY_COOLDOWN_MS = 30_000`) — `executeAutoConvert` and `recordAutoConvertAttempt` are not called; live-trigger allowed when `lastAttemptAtMs = FIXED_NOW - 35_000` (past cooldown) — convert runs once; live-trigger orphan recovery when `lastAttemptAtMs = FIXED_NOW - 3 * 60 * 1000` (past `ORPHAN_TIMEOUT_MS = 120_000`) — convert runs; mount-replay blocked under cooldown; mount-replay allowed past orphan timeout. Off-by-one on either constant (e.g. cooldown bumped to 60s, orphan widened to 5min) flips at least two of the boundary tests; under-retry (orphans never recover) and over-retry (rate-limit thrash) are now both covered. Total spec count went from 26 → 31.

### I12. `feature-flags-context` `rolloutLoggedRef` once-per-session dedup untested

Added `__tests__/config/feature-flags-rollout-exposure.spec.ts` — a focused spec mirroring the dedup gate via a small `useRolloutExposureLogger` hook with the same `useRef(false) + useEffect` shape and per-field deps as the source. Following the same pattern as `feature-flag-cascade.spec.ts`, this avoids mounting the real provider's heavy dependency graph (remote-config plugin, `useLevel`, `useAppConfig`) while still pinning the React semantics that matter. Seven tests cover: log doesn't fire while `remoteConfigReady` is false; logs exactly once when it transitions false → true; payload reflects the flag snapshot at readiness; no re-log when `nonCustodialEnabled`, `stableBalanceEnabled`, or `hasCustodialAccount` flips after readiness (one test each); no re-log when `remoteConfigReady` round-trips true → false → true. A regression that drops the ref guard fires the second test more than once; a regression that widens the dedup scope (e.g. by depending on the whole flags object) fires one of the three flag-flip tests.

### I13. `use-save-session-profile` writes `activeAccountId = Custodial` on sign-in untested

Created `__tests__/hooks/use-save-session-profile.spec.ts` (no spec existed) with 8 tests covering `saveProfile`. The Important #13 contract is pinned by three tests: "writes activeAccountId = DefaultAccountId.Custodial after saving the token" inspects the `updateState` callback and asserts the next state has `activeAccountId: "custodial-default"`; "flips activeAccountId to Custodial even when the previously-active was a self-custodial account" feeds the updater a state where `activeAccountId === "self-custodial-uuid"` and asserts it gets overwritten — the SC-active user can no longer end up on the wrong account post-sign-in; "calls updateState BEFORE the profile fetch so an SC-active user lands on Custodial even if /me fails" rejects the lazy `getUsernames` query and asserts `updateState` was still called with the Custodial id, so a network failure on `/me` doesn't strand a freshly authed user on the wrong account. Additional coverage pins related contracts: `saveProfile("")` is a no-op (no `saveToken`, no `updateState`); `saveToken` fires before `updateState` (auth context propagates first); existing-token short-circuit doesn't re-fetch or re-save; brand-new profile is saved alongside the deselected previous ones; existing profile is re-selected when re-signing-in with a fresh token. A regression that drops the Custodial write or moves it after `tryFetchUserProps` fails the I13 tests immediately.

### I14. `wallet-provider.tsx` `resolveAndPersist` LN-address persistence subpath unexercised

Promoted the inline `getLightningAddress: jest.fn().mockResolvedValue(null)` mock to a top-level `mockGetLightningAddress` so individual tests can override it, then added a new "resolveAndPersist LN-address path (Important #14)" describe block with 4 tests pinning the previously dark subpath. (1) "persists the resolved LN address and reloads the registry on success" — drives `mockGetLightningAddress` to return `{ lightningAddress: "alice@blink.sv" }`, asserts `setSelfCustodialLightningAddress("test-sc-uuid", "alice@blink.sv")` was called AND `listSelfCustodialAccounts` ran twice (initial render + post-persist reload). (2) "skips both persist and reload when getLightningAddress resolves to null" — the prior default — explicitly asserts neither `setSelfCustodialLightningAddress` nor a second `listSelfCustodialAccounts` fire, with a 20ms tail to catch a delayed call. (3) "swallows getLightningAddress rejection without persisting or reloading" — drives the bridge call to throw, asserts the outer `try/catch` swallows it (no persist, no reload, no crashlytics required). (4) "still reloads the registry when setSelfCustodialLightningAddress rejects (current swallow-and-chain behaviour)" — explicitly pins the verification question of whether the `.catch(() => {})` on the persist call is intentional. It is — the reload chains regardless so the registry view stays in sync even when storage rejects, and the ledger of behaviour is now pinned for any future refactor that revisits the swallow-and-chain pattern. Spec count went from 44 → 48.

### I15. `extractLightningAddressUsername` + 4 SC analytics helpers untested

Created two new specs. `__tests__/utils/pay-links.spec.ts` covers all four exports of `app/utils/pay-links.ts` (the file had no spec): `getPosUrl`, `getPrintableQrCodeUrl`, `getLightningAddress` (already-has-@ pass-through and append-hostname branches), and the new `extractLightningAddressUsername` with seven edge cases — full LN address (`alice@blink.sv` → `alice`), null, undefined, empty string, leading-`@` (no username portion → null), no separator (returned verbatim), and multiple `@` symbols (preserves only the first segment). `__tests__/utils/analytics-self-custodial.spec.ts` covers the four new SC analytics helpers and explicitly pins the snake_case payload-key contract: `logSelfCustodialBackupCompleted({ backupMethod: "manual" | "google_drive" | "icloud" })` → emits `"self_custodial_backup_completed"` with `{ backup_method: ... }`; `logSelfCustodialRestoreCompleted()` → emits `"self_custodial_restore_completed"` with no payload; `logSelfCustodialStableBalanceActivated({ label })` → emits `"self_custodial_stable_balance_activated"` with `{ label }`; `logSelfCustodialRolloutExposed({ nonCustodialEnabled, stableBalanceEnabled, hasCustodialAccount })` → emits `"self_custodial_rollout_exposed"` with `{ non_custodial_enabled, stable_balance_enabled, has_custodial_account }` AND a separate test asserts `false` booleans are preserved (not coerced or dropped). The camelCase → snake_case rename — drift-prone if it ever changes silently — is now structurally pinned. Spec files: 2 new, 17 tests total.

grimen (dismissed):

grimen:
### Merge activity

* **May 20, 10:43 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768).
* **May 20, 10:46 AM UTC**: [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768) couldn't merge this pull request because a downstack PR #3742 failed to merge.
* **May 20, 10:52 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768).
* **May 20, 10:55 AM UTC**: [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768) couldn't merge this pull request because a downstack PR #3746 failed to merge.
* **May 20, 12:38 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768).
* **May 20, 12:39 PM UTC**: @grimen merged this pull request with [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3768).

grimen (approved):

PR #3795, opened 2026-05-29
fix(settings): persist theme preference across account switches


### LINKED ISSUE #3793: Theme settings reset to 'Set by OS' when switching accounts
## Bug Report

**Reported by:** @ihaveadifferentname (internal testing)

**Description:**
When switching between accounts, the theme setting resets to "Set by OS" instead of preserving the user's chosen theme (e.g. Dark).

**Steps to Reproduce:**
1. Have multiple accounts in the app
2. Go to Settings → switch account
3. Create a non-custodial account
4. Switch back to the previous account
5. Observe that the theme (was set to Dark) is now reset to "Set by OS"

**Expected behavior:**
Theme preference should persist per-account (or globally) and not reset when switching accounts.

**Actual behavior:**
Theme resets to "Set by OS" after account switching.

**Severity:** Low — cosmetic/UX annoyance, no data loss

esaugomez31:
## Summary
- Fixes the bug where the theme setting reset to "Set by OS" when switching between accounts (ticket #3793). For example, if the theme was set to Dark, creating a new self-custodial account and switching back would silently revert it.
- Root cause: the theme was stored as a local field inside the Apollo cache. When the active account changes, the Apollo client is rebuilt, the cache is reset, and (for self-custodial sessions) the cache restore step is skipped on purpose to avoid leaking the custodial cache. The theme value got wiped in that process and fell back to its default.
- The theme now lives in `persistentState`, the same storage layer that already handles language, display currency and the default wallet. It is keyed by the active account id, so each account keeps its own theme without ever mixing with the others.

## What changed
- New schema version `12` adds a `themeByAccountId` map to `persistentState`.
- New helpers `getThemePreference` / `withThemePreference` mirror the pattern already used for language and display currency.
- New `useEffectiveTheme` hook reads and writes through `persistentState`.
- The theme provider, the theme screen and the settings row now consume the new hook instead of the Apollo query.
- The old Apollo `colorScheme @client` query, the `updateColorScheme` writer, the type policy entry and the schema field were removed since they are no longer needed.

## Why this approach
The codebase already uses `persistentState` for account-scoped preferences (language, display currency, default wallet). Moving the theme there:
- Keeps it safe from the Apollo cache lifecycle on account switch.
- Isolates it per account, so custodial and self-custodial accounts never share or overwrite each other.
- Removes dead Apollo code that was only there to support this one preference.

### DIFF
--- a/.storybook/views/story-screen.tsx
+++ b/.storybook/views/story-screen.tsx
@@ -5,7 +5,7 @@ const PersistentStateWrapper: React.FC<React.PropsWithChildren> = ({ children })
   <PersistentStateContext.Provider
     value={{
       persistentState: {
-        schemaVersion: 11,
+        schemaVersion: 12,
         galoyInstance: {
           id: "Main",
         },
--- a/__tests__/hooks/use-effective-display-currency.spec.ts
+++ b/__tests__/hooks/use-effective-display-currency.spec.ts
@@ -14,7 +14,7 @@ const mockUseAccountUpdateDisplayCurrencyMutation = jest.fn(
 )
 const mockUpdateState = jest.fn()
 let mockPersistentState = {
-  schemaVersion: 11,
+  schemaVersion: 12,
   galoyInstance: { id: "Main" },
   galoyAuthToken: "",
   activeAccountId: undefined as string | undefined,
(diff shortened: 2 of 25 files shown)


esaugomez31:
* **#3795** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3795?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3795?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (approved):

PR #3799, opened 2026-05-29
feat(stablesats): restrict Stablesats/Stablecoin by country (default: HK)

esaugomez31:
## Summary

For users in Hong Kong (and any other country added later via Firebase Remote Config), the app now hides the Dollar account on custodial accounts only. Self-custodial accounts are unaffected and can use the Dollar account anywhere.

Driven by regulatory uncertainty as discussed in [blink-wip#745](https://github.com/blinkbitcoin/blink-wip/issues/745).

## What changes for custodial users in a restricted country

- The Dollar row on Home shows a disabled state with a "not available" label, and opens an info modal
- The Transfer button is disabled and opens the same modal
- The total balance excludes the Dollar wallet
- On Receive, the currency toggle is blocked and the initial wallet is forced to Bitcoin, even if the user had Dollar saved as their default
- The "Default account" row in Settings is hidden
- Direct navigation to the Convert or Default Wallet screens (via deep link or the back stack) redirects to Home

The modal explains the restriction and offers two actions: a primary `Create new` CTA that opens the account creation flow (the Dollar account is available on self-custodial accounts), and a secondary `Close` button to dismiss.

## Converting an existing Dollar balance to Bitcoin

If a restricted custodial user still holds a Dollar balance, dismissing the restriction modal opens a `Convert your Dollar balance to Bitcoin` modal. It shows the current Dollar balance and the approximate Bitcoin amount they will receive, and a single `Approve` action moves the entire Dollar balance to Bitcoin.

The conversion runs behind a shared `useIntraLedgerConversion` hook that the existing Convert confirmation screen now consumes as well, so both entry points share one intraledger mutation, analytics, haptics and error path. The self-custodial conversion hook was aligned to the same surface for symmetry; the conversion mechanics for both custodial and self-custodial are unchanged.

## What stays the same for self-custodial users

Self-custodial accounts can use the Dollar account (USDB) without any restriction, regardless of country.

## How the list of countries is controlled

A Firebase Remote Config entry called `stablesatsBlockedCountries` (default `["HK"]`) controls which countries are restricted. Ops can add or remove countries without releasing a new app.

The serialization of the list is handled by the shared `app/utils/remote-config.ts` util (`serializeRemoteConfigDefault` on write, `getRemoteConfigList` on read), so the Remote Config layer does not leak JSON parsing into business logic.

PR #3804, opened 2026-06-01
feat(custodial): block US signup and gate first-signup by country

### LINKED ISSUE #3780: feat(custodial): Block custodial account creation for US users under all conditions
## Summary

US users must not be able to create custodial accounts under any conditions — including the second-account (multi-account) flow.

## Context

From the 2026-05-21 sync meeting: **No custodial for US users under any conditions (including second-account flow).**

For other regions/legislations that are self-custodial-only by default, users should still be able to create a custodial account as their 2nd+ account (first account must always be self-custodial). The custodial vs self-custodial selection when creating an account should appear for 2nd+ accounts in all countries **except** the US, which carries significant regulatory risk.

## Acceptance Criteria

- [x] US users (determined by phone number / jurisdiction) cannot create a custodial account — not as first account, not as second account, not under any flow
- [x] Non-US self-custodial-default regions still allow custodial as 2nd+ account
- [x] The account-type selection screen is suppressed entirely for US users on all account creation flows
- [x] Leverage the existing country/region constant lists introduced in the self-custodial implementation

/cc @grimen

esaugomez31:
## Summary

Ticket: https://github.com/blinkbitcoin/blink-mobile/issues/3780

Blocks custodial account creation for US users under all conditions, and gates the first custodial signup in regions where self-custodial is the regulated default. The account-type selection screen is now suppressed whenever only one option (self-custodial or custodial) is genuinely available for the user.

## Behavior by region

**United States**: Custodial creation is blocked in every flow, both for the first account and any subsequent account. The selection screen does not appear and the user is sent directly to the self-custodial creation flow.

**Self-custodial-default regions** (OFAC sanctioned countries, EU-27 under MiCA, and the non-EU set from Google Play policy 16329703): The first account must be self-custodial. The selection screen is suppressed and the user is sent directly to the self-custodial flow. Once the user already has an account, custodial becomes available as a second or subsequent account, and the selection screen is shown so the user can choose.

**Unrestricted regions**: The selection screen is shown for every account creation, allowing the user to choose between self-custodial and custodial.

The selection screen appears only when the user has a real choice between two options. When the rules leave a single valid option, the screen is suppressed and the user is sent directly to the appropriate creation flow.

## Configuration

Two new Remote Config flags drive the lists. Defaults ship with the values previously hardcoded in the app, so behavior matches the existing region map out of the box:

- `custodialSignupBlockedCountries` — countries where custodial signup is blocked under any condition. Defaults to `["US"]`.
- `custodialFirstSignupBlockedCountries` — countries where the first custodial signup is blocked. Custodial remains allowed as a 2nd+ account. Defaults to the OFAC-sanctioned countries plus the Google Play 16329703 non-EU and EU-27 (MiCA) sets, with the US excluded since it sits in the always-blocked list.

Either list can be updated from Firebase without shipping a new app version.

## Restore flow

Account restore is unaffected. Users restoring an existing custodial account can do so regardless of country. The restriction applies only to new account creation.

## Notes

**Hardcoded country list moved to Remote Config.** The previous implementation kept the list of restricted countries hardcoded in `app/config/custodial-countries.ts` with a `TODO: replace with backend-driven list` annotation. That follow-up is resolved here: both lists are now served from Firebase Remote Config, with the original values shipped as in-app defaults so behavior is preserved out of the box and can be updated remotely going forward.

**Restriction logic refined to match the ticket.** The previous logic treated the entire 41-country list as a flat block: custodial creation was denied in those countries regardless of whether the user was creating their first account or an additional one. The ticket requires a finer-grained model:

- The US must be blocked under any condition (first account, second account, any flow).
- Other restricted countries (OFAC, MiCA, Google Play 16329703 non-EU) must only force self-custodial for the first account. Once the user already holds an account, custodial becomes a valid choice for additional accounts in those regions.

The new logic implements this distinction by splitting the single list into two Remote Config lists (`custodialSignupBlockedCountries` and `custodialFirstSignupBlockedCountries`) and combining them with the user's existing-account count when deciding whether custodial is available for the current signup attempt.


esaugomez31:
* **#3799** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3799?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3804** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3804?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3804?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

app/config/feature-flags-context.tsx
```diff
@@ -228,6 +253,16 @@ export const FeatureFlagContextProvider: React.FC<React.PropsWithChildren> = ({
...
 
+        const custodialSignupBlockedCountries = JSON.parse(
+          remoteConfigInstance().getValue(CustodialSignupBlockedCountriesKey).asString(),
+        )
+
+        const custodialFirstSignupBlockedCountries = JSON.parse(
+          remoteConfigInstance()
+            .getValue(CustodialFirstSignupBlockedCountriesKey)
+            .asString(),
+        )
```
grimen: Whenever native APIs like `JSON.parse` is used in middle of app/business logic it automatically should signal: Use existing or new utility function with a fitting name for behavior (depending on abstraction layer may be a universal helper/utility, app helper/utility, or both).

app/config/feature-flags-context.tsx
```diff
@@ -109,6 +130,10 @@ const defaultFeatureFlags: FeatureFlags = {
 remoteConfigInstance().setDefaults({
   ...defaultRemoteConfig,
   replaceCardDeliveryConfig: JSON.stringify(defaultReplaceCardDeliveryConfig),
+  custodialSignupBlockedCountries: JSON.stringify(defaultCustodialBlocks.blockAnySignup),
+  custodialFirstSignupBlockedCountries: JSON.stringify(
+    defaultCustodialBlocks.blockFirstSignup,
+  ),
```
grimen: Same here, `JSON.parse` is code smell in this abstraction layer.

grimen (changes requested):

app/config/feature-flags-context.tsx
esaugomez31: Extracted to a new util `app/utils/remote-config.ts` (`getRemoteConfigList<T>`, `getRemoteConfigObject<T>`).

app/config/feature-flags-context.tsx
esaugomez31: Addressed via the new util `serializeRemoteConfigDefault`.

grimen:
### Code review

Solid PR — extracting `useCustodialEligibility` and the `remote-config.ts` helpers is the right direction and addresses the earlier `JSON.parse` abstraction feedback. One change worth making before merge, plus two cheap hardening items.

**High prio — add an `Array.isArray` guard (can crash the app)**

`getRemoteConfigList` parses Remote Config JSON and casts to `T[]` with no shape check. A valid-but-non-array value (e.g. an operator typo like `{"US":true}`) parses fine, skips the fallback, and then `custodialSignupBlockedCountries.includes(country)` throws `TypeError: .includes is not a function` mid-render.

https://github.com/blinkbitcoin/blink-mobile/blob/9ee4b21ad32cac1dff46dd5cb57664a4bf3771ea/app/utils/remote-config.ts#L21-L26

Suggest validating before returning:
```ts
export const getRemoteConfigList = <T>(key: string, fallback: T[]): T[] => {
  const parsed = tryParseJson<unknown>(remoteConfigInstance().getValue(key).asString(), fallback)
  return Array.isArray(parsed) ? (parsed as T[]) : fallback
}
```
Worth two matching test cases (valid-JSON-non-array, valid-JSON-primitive) in `__tests__/utils/remote-config.spec.ts` — currently only empty-string and invalid-JSON hit the fallback path.

**Medium — make the unknown-country policy explicit**

The deleted `custodial-eligibility.ts` was deliberately fail-closed (`if (!countryCode) return false`). The new hook is fail-open: when `country` is undefined, both blocks are `false` so `signupAllowed = true`.

https://github.com/blinkbitcoin/blink-mobile/blob/9ee4b21ad32cac1dff46dd5cb57664a4bf3771ea/app/hooks/use-custodial-eligibility.ts#L20-L27

This is safe today only because country is undefined solely while `loading` is true, during which the Create button is disabled. But the reversal is implicit and unguarded for any future consumer that doesn't gate on `loading`. Either restore fail-closed, or keep the behavior with a comment stating callers must gate on `loading`.

**Medium — lock in the invariant that keeps the above safe**

Nothing currently asserts the Create button stays disabled while detecting country — and that disabled state is the only thing preventing the fail-open from going live.

https://github.com/blinkbitcoin/blink-mobile/blob/9ee4b21ad32cac1dff46dd5cb57664a4bf3771ea/app/screens/get-started-screen/get-started-screen.tsx#L117-L120

A `loading: true` -> press -> `expect(mockNavigate).not.toHaveBeenCalled()` test would guard `disabled={!canCreateAccount || detectingCountry}` against a future refactor silently dropping it.

Non-blocking: removing the hardcoded US floor is intentional (the PR's purpose; defaults still `["US"]`); the `useAccountTypeOptions` concern-mixing is pre-existing debt already tracked.

🤖 Generated with [Claude Code](https://claude.ai/code)

grimen (changes requested):
Requesting changes: please add the `Array.isArray` guard in `getRemoteConfigList` (crash risk on valid-but-non-array Remote Config values) plus the two test cases. The fail-open unknown-country reversal in `useCustodialEligibility` and the disabled-while-loading test are also worth addressing. Details in https://github.com/blinkbitcoin/blink-mobile/pull/3804#issuecomment-4616827003

esaugomez31:
@grimen feedback applied.

**1. HIGH PRIO: `Array.isArray` guard in `getRemoteConfigList`**

Done. Added an `Array.isArray` check after `JSON.parse` in `app/utils/remote-config.ts`. Valid JSON that is not an array (object, primitive, `null`) now falls back instead of leaking a non-array into callers. Covered with two new cases in `__tests__/utils/remote-config.spec.ts`: object payload (`{"US":true}`) and primitive payload (`"US"`).

**2. MEDIUM: explicit unknown-country policy in `useCustodialEligibility`**

Done. Restored fail-closed semantics in `app/hooks/use-custodial-eligibility.ts`: `signupAllowed` is `false` when `country` is `undefined`. Added a JSDoc above the expression so the contract is explicit (consumers that forget to gate on `loading` still default to the safe outcome). The existing test in the `country undefined` block now asserts the new behavior.

**3. MEDIUM: lock in the disabled-while-loading invariant**

Done. The existing test in `__tests__/screens/get-started-screen/get-started-screen.spec.tsx` now also dispatches a `fireEvent.press` while loading and asserts `mockNavigate` was never called. Locks in the contract that the button cannot be activated mid-detection even if a future refactor relocates the disabled wiring.

grimen (changes requested):

## Critical

### C1. "Fail-closed when country unknown" is unreachable — geo failure silently resolves to `"SV"` and opens the gate

`app/hooks/use-custodial-eligibility.ts:20-30` + `app/hooks/use-device-location.ts:9,66-68`

The hook's safety premise is `country !== undefined` ⇒ trustworthy. But `useDeviceLocation` **never settles on `undefined`** — every failure path resolves to a concrete country, `DEFAULT_COUNTRY_CODE = "SV"`:

```ts
// use-device-location.ts — ipapi timeout / 429 / no data:
} catch (err) {
  setCountryCode((data.countryCode as CountryCode) ?? DEFAULT_COUNTRY_CODE) // "SV"
}
```

Phone-parse throw, `useCountryCodeQuery` error, ipapi 5s timeout, and the explicitly-anticipated 429 rate-limit all land on cached-or-`"SV"`. So the `undefined` guard only protects the millisecond loading window — not the dominant real-world failure mode.

**Silent scenario:** US user opens the app, `ipapi.co` times out → country = `"SV"` → in no block list → `signupAllowed = true` → **custodial signup offered to a US user**. No log, no Sentry event, nothing.

`useDeviceLocation` is pre-existing (not in this diff), but this PR is what turns its silent fallback into a compliance hole. Adjacent: the `/** Fail-closed when the country is unknown… */` doc comment (`use-custodial-eligibility.ts:27-28`) documents a guarantee that doesn't hold.

**Fix:** express fail-closed against *detection failure*, not against `undefined`. Expose a `detectionFailed`/`isFallback` flag from `useDeviceLocation` and treat fallback-derived countries as blocked for signup-gating purposes (restore stays unaffected), + `logError` so the "compliance decision made on fallback country" rate is observable. Update the doc comment to the real contract.

### C2. The entire compliance chain has zero production observability

`app/utils/remote-config.ts:13-20` + `app/config/feature-flags-context.tsx:302-306`

Grep-verified: not a single `logError`/Sentry call anywhere in the chain — only bare `catch {}`, `console.warn`, `console.error`.

```ts
// remote-config.ts
} catch {
  return fallback
}
```

```ts
// feature-flags-context.tsx
} catch (err) {
  console.error("Error fetching remote config:", err)
} finally {
  setRemoteConfigReady(true)   // "ready" even on total failure
}
```

The fallback *values* are safe (defaults include the block lists — a Firebase outage does **not** empty them, verified). The silence is the defect:

- An ops typo in the Firebase console (`"US,CU,IR"` instead of `["US","CU","IR"]`, trailing comma, smart-quote) silently discards the remote update and reverts to the **stale baked-in defaults of the last app release**. The entire reason these lists moved to Remote Config — updating sanctions lists without shipping — silently fails, unauditably.
- A fleet-wide fetch failure (everyone running on baked-in defaults, no remote compliance update can ever land) is invisible in telemetry, and `remoteConfigReady = true` overstates confidence to downstream consumers.

**Fix:** `logError` + dedicated `errorIds` entry in `tryParseJson`'s catch, `getRemoteConfigList`'s non-array branch (include the config key + truncated raw value), and the context's fetch catch. Consider a `remoteConfigFetchFailed` flag so consumers can distinguish "remote values" from "defaults because fetch threw."

### C3. Case-normalization asymmetry — a lowercase entry in the Firebase console silently unblocks a country

`app/hooks/use-custodial-eligibility.ts:20-24` + `app/utils/remote-config.ts:22-28`

The hook uppercases the *country* (`countryCode?.toUpperCase()`) but trusts the *list* is uppercase. The deleted code compared against a list it controlled in-repo; now the list comes from Remote Config, where an ops person typing `"us"` silently fails the `.includes("US")` match — **the US block evaporates with no error and no signal**. Combined with C2, the failure is doubly invisible.

Related unsoundness in the same helper: `getRemoteConfigList<T>` checks `Array.isArray` but blind-casts elements (`parsed as T[]`) — `[1, 2]` passes as `string[]` and flows into `.includes(country)` as a permanent non-match.

**Fix:** normalize and validate in one place — `getRemoteConfigList` (or the comparison site): keep only `typeof x === "string"` elements, `.toUpperCase()` them, and `logError` when entries were dropped or coerced.

---

## Important

- **I1 — `isFirstSignup` races `useAccountRegistry`'s async hydration.**
  `app/hooks/use-custodial-eligibility.ts:26` + `app/hooks/use-account-registry.ts`
  `isFirstSignup = accounts.length === 0`, but the registry populates `accounts` asynchronously with **no loading flag**, and eligibility's `loading` covers country detection only. Failure direction is over-blocking (safe), but the result can flip after the CTA is already enabled, and nothing marks it provisional. **Fix:** expose `loading` from the registry and OR it into eligibility's `loading`.

---

## Design

**SOLID violations**

- **D1 — ISP: `useCustodialEligibility` exposes 5 fields, its only consumer uses 2.**
  `app/hooks/use-custodial-eligibility.ts:6-12`
  `signupBlocked` / `firstSignupBlocked` / `isFirstSignup` have **zero consumers** (grep-verified). Exporting the intermediate terms invites future callers to recombine the policy wrongly — the correct combination lives only inside the hook. Shrink the public type to `{ signupAllowed, loading }`.

- **D2 — SRP: pure policy regressed into hook soup.**
  `app/hooks/use-custodial-eligibility.ts:14-39`
  The deleted `isCustodialAllowedForCountry` was a pure function; its replacement interweaves the decision with three hook data sources, so the new spec needs 185 lines of mock scaffolding for a 4-line boolean formula. Extract a pure `decideCustodialEligibility({country, accountCount, lists})` core; keep the hook as a thin adapter. Also makes the compliance matrix exhaustively testable.

- **D3 — OCP: shotgun surgery in `feature-flags-context.tsx`.**
  Adding one remote-config key touches **six** sites (key const, `RemoteConfig` type, defaults object, `setDefaults`, fetch block, `setRemoteConfig`) — this PR adds two keys → ~12 edit sites. Follow-up ticket: a key registry (single `{ key: default }` map driving type, defaults, and fetch).

- **D4 — The `remote-config.ts` adapter is the right shape but inconsistently rigorous.**
  `app/utils/remote-config.ts:22-31`
  `getRemoteConfigList` validates `Array.isArray`; its sibling `getRemoteConfigObject` doesn't check the parsed value is a plain object — a remote value of `"42"` or `"[1,2]"` flows through and gets spread (`{...defaults, ...42}` silently no-ops; `...[1,2]` injects numeric keys). Neither validates element/field types (see C3). Apply the same validation discipline in both helpers.

- None introduced. The new hook is flat guard expressions; the screen gains one conditional branch.

**Code smells**

- **D5 — Duplicated `AccountOption → flow` mapping with a silent-else to the restricted flow.**
  `app/screens/get-started-screen/get-started-screen.tsx:61-66` (ternary) + `app/screens/account-type-selection/account-type-selection.tsx:42-58` (if/else)
  A hypothetical third `AccountOption` falls through the ternary to `"trial"` — the *restricted* flow. One exhaustive `Record<AccountOption, flow>` next to the enum fixes both call sites.

- **D6 — Lost provenance.**
  `app/config/feature-flags-context.tsx:84-98`
  The deleted `custodial-countries.ts` carried ISO-3166 notes and the OFAC/Google Play policy source links; `defaultCustodialBlocks` is a `// prettier-ignore` literal mid-file and the support.google.com policy URL is **gone**. Restore the provenance comments (own module, or above the literal).

- See also the comment rot on the "fail-closed" doc comment, kept under C1 because it documents the broken guarantee.

**Naming smells**

- **D7 — Three vocabularies for the same two concepts.**
  RC keys say `custodialSignupBlockedCountries`/`custodialFirstSignupBlockedCountries`, the defaults object says `blockAnySignup`/`blockFirstSignup`, the hook says `signupBlocked`/`firstSignupBlocked`. The key names are the wire contract — align internals to those.

- **D8 — `signupAllowed` vs `signupBlocked` is a false-antonym pair.**
  They read as negations but aren't: `signupAllowed === false` while `signupBlocked === false` whenever country is undefined or first-signup gating fires. A reader writing `!signupBlocked` instead of `signupAllowed` introduces a compliance bug that compiles. Fixed automatically by D1.

- **D9 — `loading` under-promises.**
  It covers country detection only, not account-registry hydration (see I1) — widen the semantics or rename so the gap is visible.

---

## Test coverage

- **`app/config/feature-flags-context.tsx` (RC wiring)** — the "US always blocked" guarantee lives in the default `["US"]` literal and the `catch` fallback **in this file**, and no test imports the real context — `feature-flag-cascade.spec.ts` re-implements a standalone `deriveFlags` helper. A typo in a key string or default array ships silently: `remote-config.spec.ts` tests against `"any-key"` and never asserts the real key constants or default arrays. Must-add: (a) empty/unfetched RC → context exposes `custodialSignupBlockedCountries === ["US"]`; (b) `getValue` throwing → defaults still served, `useCustodialEligibility`'s `.includes()` doesn't crash.
- **Default-lists structure** — the deleted `custodial-countries.spec.ts` validated list shape (uppercase ISO codes, no duplicates). That structural check vanished — the 38-country list is now an untested literal. One small shape test recovers it.

- **`__tests__/hooks/use-custodial-eligibility.spec.ts:128`** (→ C1) — the undefined-country fail-closed test mocks `useDeviceLocation` returning `countryCode: undefined`, a state the real hook **never settles on** (it settles on `"SV"`). The test pins the guard against a phantom input while the actual failure mode (fallback country) sails through.
- **`__tests__/utils/remote-config.spec.ts`** (→ C2) — asserts fallback *values* on malformed input but not that the failure is *observable*; a regression swapping `logError` back to bare `catch {}` (or never adding it) ships invisibly.
- **`__tests__/hooks/use-custodial-eligibility.spec.ts`** (→ C3) — every case passes uppercase lists; a lowercase list entry (`["us"]` vs country `"US"`) is never exercised — the case-mismatch non-match is encoded by omission.
- **`__tests__/hooks/use-account-type-options.spec.ts` + `__tests__/screens/get-started-screen/get-started-screen.spec.tsx`** (→ seam) — each layer mocks the adjacent hook wholesale, so no test exercises the real chain `device-location → eligibility → options → navigation`. A contract rename between layers (e.g. `defaultSelected`) passes every suite.

- **C1** — mock ipapi/query failure so `useDeviceLocation` settles on the fallback → assert a US/restricted-region user is still **blocked** (fails on current code).
- **C2** — feed malformed / non-array RC values → assert `logError` called with the config key (fails on current code).
- **C3** — lists `["us"]`, country `"US"` → assert `signupAllowed === false` (fails on current code); plus `[1,2]` as list → assert dropped + logged.
- **I1** — registry still hydrating (`accounts: []`, loading) in a restricted region with an existing account → assert eligibility reports loading / not authoritative.
- **Precedence** — a country present in **both** lists with an existing account → assert still blocked (locks `signupBlocked` priority over the first-signup carve-out).
- **Seam** — one integration-level `GetStartedScreen` render with only `useDeviceLocation`/`useAccountRegistry`/RC mocked (real hooks in between) — closes the Tier 2 seam gap and I1 together.

esaugomez31:
@grimen 
## Critical

**C1 — "Fail-closed when country unknown" is unreachable; geo failure resolves to `"SV"` and opens the gate**
Done! `useDeviceLocation` now exposes a `detectionFailed` flag, set true on every fallback path (phone-parse fail/empty, country-code query error, ipapi timeout / no-data / 429), each with a `logError`. `decideCustodialEligibility` fails closed on `detectionFailed`, so a fallback-derived country is blocked for signup gating. The doc comment now states the real contract.

**C2 — the entire compliance chain has zero production observability**
Done! Added a `logError` seam (`app/utils/log-error.ts`) and wired it into `tryParseJson`'s catch, the non-array / non-object branches, the dropped-entries path, and the context's `fetchAndActivate` catch. The config key is folded into the Error message so a malformed-JSON ops typo is attributable in Crashlytics. Bare `catch {}` / `console.error` removed.

**C3 — case-normalization asymmetry; a lowercase Firebase entry silently unblocks a country**
Done! Normalization + validation centralized in `getRemoteConfigStringList`: entries are uppercased and non-string elements dropped + logged, so a `"us"` in the console still matches and `[1,2]` can no longer flow into `.includes` as a permanent non-match.

## Important

**I1 — `isFirstSignup` races `useAccountRegistry`'s async hydration**
Done! `useAccountRegistry` now exposes `loading` (covering both async reads), and `useCustodialEligibility` ORs it into its own `loading`, so the result is marked provisional until the registry settles.

## Design

**D1 — ISP: hook exposed 5 fields, consumer used 2**
Done! Public surface shrunk to exactly `{ signupAllowed, loading }`.

**D2 — SRP: pure policy regressed into hook soup**
Done! Extracted a pure `decideCustodialEligibility({ country, detectionFailed, accountCount, lists })` to `app/utils/custodial-eligibility.ts`; the hook is now a thin adapter, and the matrix is exhaustively unit-tested.

**D3 — OCP: shotgun surgery in `feature-flags-context.tsx`**
Leaving as the follow-up ticket you suggested (a key registry driving type, defaults, and fetch). Not in this PR.

**D4 — adapter rigor: `getRemoteConfigObject` didn't validate shape**
Done! It now validates the parsed value is a plain object (rejects primitives/arrays) with a `logError`, symmetric with `getRemoteConfigList`'s `Array.isArray` check.

**D5 — duplicated `AccountOption → flow` mapping with silent-else to the restricted flow**
Done! One exhaustive `Record<AccountOption, AccountFlow>` consumed by both screens; a hypothetical third option fails to compile instead of falling through to `trial`.

**D6 — lost provenance**
Done! Restored the ISO-3166 note + OFAC and Google Play policy (16329703) source links above the default lists.

**D7 — three vocabularies for the same two concepts**
Done! Internals aligned to the Remote Config key names (`custodialSignupBlockedCountries` / `custodialFirstSignupBlockedCountries`).

**D8 — `signupAllowed` vs `signupBlocked` false-antonym pair**
Done! `signupBlocked` is gone from the public surface (falls out of D1); only `signupAllowed` remains, so the `!signupBlocked` footgun can't be written.

**D9 — `loading` under-promises**
Done! `loading` now covers registry hydration in addition to country detection.

## Test coverage

**Tier 1 — production code with no test file**
Done! Added a default-lists shape test against the real `defaultRemoteConfig` (uppercase ISO-3166, no duplicates, no cross-list overlap, US floor present), recovering the deleted `custodial-countries.spec`.

**Tier 2 — tests passing while a bug is alive**
Done! The phantom undefined-country test is replaced by a real detection-failure test; the remote-config spec now asserts the failure is observable (`logError` called with the config key in the message), not just the fallback values; the lowercase-list case is exercised.

**Tier 3 — regression tests per finding**
Done! Added: C1 (fallback country blocked), C2 (malformed RC → `logError` with key), C3 (`["us"]` → blocked + `[1,2]` → dropped & logged), I1 (registry hydrating → `loading`), and precedence (country in both lists with an existing account → still blocked). The full real-chain `device-location → eligibility → options → navigation` Seam test I left as the recommended (non-blocking) follow-up.

grimen (approved):

PR #3781, opened 2026-05-26
refactor: upgrade React Native to 0.85.3 (React 19)


esaugomez31:
## Summary

Upgrades React Native from 0.76.9 to 0.85.3, React from 18.3.1 to 19.2.3, and the reanimated stack from 3.16.7 to 4.3.1 (with the new `react-native-worklets` 0.8.0 dependency split). Required to keep the iOS app on a supported toolchain (Apple App Store iOS 26 SDK requirement).

## Stack delta

- React Native: 0.76.9 → 0.85.3
- React / react-dom: 18 → 19.2.3 (+ react-test-renderer 19.2.3)
- @testing-library/react-native: 12.x → 13.3.3
- react-native-reanimated: 3.16.7 → 4.3.1
- react-native-worklets: NEW 0.8.0 (split out of reanimated v4)
- react-native-gesture-handler: 2.25 → 2.31.2
- react-native-screens: ≥4.0 <4.5 → ≥4.25 <4.26
- react-native-camera-kit: 16.x → 18.x (16.x C++ codegen breaks on RN 0.85)
- react-native-svg: 15.8.0 → 15.15.5 (ImageResponseObserver API change in RN 0.84+)
- react-native-view-shot: 4.x → 5.x (RCTScrollView removed in RN 0.84+)
- @react-navigation/{native,stack,bottom-tabs}: 6.x → 7.x (required by screens 4.25 + Fabric)
- @react-navigation/elements: 1.x → 2.x
- @react-native-community/slider: 4.5.7 → 5.2.0 (required for storybook bundle on RN 0.85)
- @react-native/{babel,jest,codegen,metro-config}-preset: 0.76.x → 0.85.3
- @types/react: ^19.2.15
- @breeztech/breez-sdk-spark-react-native: 0.13.2-dev3 → 0.15.0 (16 KB page-size compliance + real-time `TokenTransactionEvent`)
- Gradle: 8.10.2 → 8.13, compileSdkVersion 35 → 36

## Notable changes by area

**iOS build**
- Build React-Core from source (`RCT_USE_PREBUILT_RNCORE=0`) until RNFirebase supports the prebuilt binary (RN 0.84+)
- Downgrade non-modular include errors to warnings for all pod targets (`use_frameworks!` + static linkage)
- Register `RCTAppDependencyProvider` in AppDelegate so Fabric discovers third-party native components at runtime
[...]

PR #3805, opened 2026-06-01
fix(price-history): restore Bitcoin price text under reanimated v4


esaugomez31:
## Summary

Restores the Bitcoin price text on the chart screen, which stopped rendering after the React Native 0.85 / reanimated v4 upgrade (PR #3781).

## Root cause

The price-history component bound a `SharedValue<string>` directly to a `text` prop on a component returned by `Reanimated.createAnimatedComponent(TextInput)`. That pattern relied on reanimated v3 behavior that automatically wired whitelisted native props to shared values. The reanimated v4 release removed that shortcut: shared values must now be threaded through `useAnimatedProps` explicitly. With v4 in place, the underlying `TextInput` no longer received a `value`, so the price text rendered as an empty string.

The file already contained a correctly written wrapper, `AnimatedText`, using `useAnimatedProps`, but it was defined and exported without ever being consumed by the component itself.

## Fix

Replaced the raw animated component usage with the existing `AnimatedText` wrapper in the two places that render shared-value-driven text: the current price and the active timestamp shown on chart press. The wrapper passes both a static `value` and an animated `animatedProps`, so reanimated v4 correctly drives the native text updates on every shared-value change.

Additional cleanup applied in the same change:

- Removed the two `// @ts-ignore-next-line` comments that were silencing the type error caused by the prior incorrect usage. The typed wrapper covers the contract correctly.
- Moved `color` from a separate prop into the `style` array, matching the standard `TextInput` styling API.
- Demoted `AnimatedText` to an internal helper. The previous `export` had no external consumers.

## Coverage

Added a unit test for `PriceHistory` that renders the component with mocked Apollo, reanimated, and victory-native, and asserts that the latest formatted price reaches the underlying `TextInput.value`. The test guards the rendering chain `useDerivedValue` → `AnimatedText` → `TextInput.value`, so a regression on the same code path will fail the build.

### DIFF
--- a/__tests__/components/price-history.spec.tsx
+++ b/__tests__/components/price-history.spec.tsx
@@ -1,0 +1,211 @@
+import React from "react"
+import { render } from "@testing-library/react-native"
+
+import { PriceHistory } from "@app/components/price-history/price-history"
+
+const mockUseBtcPriceListQuery = jest.fn()
+const mockFormatMoneyAmount = jest.fn()
+
+jest.mock("@app/graphql/generated", () => {
+  const actual = jest.requireActual("@app/graphql/generated")
+  return {
+    ...actual,
+    useBtcPriceListQuery: (args: unknown) => mockUseBtcPriceListQuery(args),
+  }
+})
+
+jest.mock("@app/hooks/use-display-currency", () => ({
+  useDisplayCurrency: () => ({
+    formatMoneyAmount: (args: unknown) => mockFormatMoneyAmount(args),
+  }),
+}))
+
+jest.mock("@app/i18n/i18n-react", () => ({
+  useI18nContext: () => ({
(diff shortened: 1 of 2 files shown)


esaugomez31:
* **#3805** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3805?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3805?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

esaugomez31:
Before fix:

Image

After fix:

Image

grimen (approved):

PR #3816, opened 2026-06-03
fix(test): suppress InteractionManager deprecation warning in CI logs

### LINKED ISSUE #3813: fix(test): Deprecation warning floods Concourse/CI logs with gibberish: InteractionManager (via @react-navigation/stack)
## Description

**Every test in the CI suite** emits a React Native deprecation warning for `InteractionManager`. The warning is triggered by dependencies that still use the deprecated `InteractionManager` API. While tests currently **PASS**, this warning floods the **Concourse/CI logs with repeated gibberish**, drowning out anything actually useful. It also signals an upcoming breaking change — `InteractionManager` will be removed in a future React Native release.

## Warning

```
console.warn

  InteractionManager has been deprecated and will be removed in a future release.
  Please refactor long tasks into smaller ones, and use 'requestIdleCallback' instead.

  at warn (node_modules/react-native/Libraries/Utilities/warnOnce.js:27:11)
  at Object.warnOnce [as InteractionManager] (node_modules/react-native/index.js:236:5)
  at InteractionManager (node_modules/@react-navigation/stack/src/views/Stack/Card.tsx:187:9)
  at latestCallback (node_modules/use-latest-callback/lib/src/index.js:21:28)
  at onStartInteraction (node_modules/@react-navigation/stack/src/views/Stack/Card.tsx:246:9)
  at latestCallback (node_modules/use-latest-callback/lib/src/index.js:21:28)
  at Timeout.animate [as _onTimeout] (node_modules/@react-navigation/stack/src/views/Stack/Card.tsx:404:9)
```

> Note: the stack trace above (`@react-navigation/stack`'s `Card.tsx`) is **just one** of several origins. The deprecated `InteractionManager` API is referenced from **multiple places** — across `@react-navigation` packages and other RN dependencies — so the warning is not attributable to a single file. The root issue is the deprecated API itself being called from several sources.

## Scope

This is **not isolated to a few suites or a single source file** — the warning appears across the **entire test suite in CI**, triggered from multiple dependencies. Any test that renders a screen wrapped in a navigator (or otherwise exercises code paths that touch `InteractionManager`) surfaces it, which in practice is nearly every suite. As a result the **Concourse/CI logs are flooded with this repeated warning + stack trace gibberish**, making the output extremely noisy and hard to read.

## Impact

- **Concourse/CI logs are flooded with repeated warning + stack-trace gibberish**, burying real warnings and failures and making CI output painful to scan.
- Signals a future-compatibility risk: `InteractionManager` is scheduled for removal from React Native.

## Suggested fix

Since the warning comes from multiple sources, it needs to be addressed once for the whole test suite, e.g.:

- Upgrade the offending dependencies (`@react-navigation/*` and any others still on the deprecated API) to versions that no longer rely on `InteractionManager` (or migrate to `requestIdleCallback`), once available, **or**
- If no fixed upstream versions exist yet, suppress this specific warning globally in the Jest setup (e.g. filter it in `jest.setup`) until the dependencies are fixed — so it's handled in one place rather than per-suite and the Concourse logs stay clean.

## Notes

This is a non-blocking warning — tests pass. Filed to track the deprecation and clean up the noisy Concourse/CI logs before it becomes a breaking change.

grimen:
## Summary

Fixes #3813.

Every suite that renders a navigator (or otherwise touches `InteractionManager`) re-triggers React Native's `InteractionManager has been deprecated` warning once per test file (`warnOnce` dedupes per module registry, and Jest resets the registry per file), flooding Concourse/CI logs with repeated warning + stack-trace noise.

### Changes

- **`jest.setup.js`** (new): wraps `console.warn` and drops **only** messages matching `InteractionManager has been deprecated` — and **only when `CI` is set**, so the deprecation stays visible in local dev runs. All other warnings pass through untouched.
- **`jest.config.js`**: registers the setup file via `setupFilesAfterEnv` (re-applied per test file, so it survives Jest's per-file module resets).
- **`ci/pipeline.yml`**: adds `params: { CI: true }` to the Concourse `test-unit` task. GitHub Actions sets `CI=true` automatically, but Concourse deliberately does **not** expose build-metadata env vars to task containers (concourse/concourse#790), so it must be set explicitly — same convention the `build` tasks already use.

### Why not fix the root cause instead

- `@react-navigation` v7 (current, latest stable) still uses `InteractionManager`; it is only removed in v8, which is **alpha-only** today and a breaking migration.
- The app's own `app/screens/transaction-history/transaction-history-screen.tsx` also calls `InteractionManager.runAfterInteractions` directly, so a react-navigation upgrade alone would not silence the warning.

That migration is real future-proofing work but out of scope for this CI-log cleanup; it can be tracked separately.

## Test plan

- [x] Local run (no `CI`): warning still visible — `yarn test __tests__/screens/transaction-history-screen.spec.tsx` → 1 occurrence, tests pass
- [x] `CI=true` run: warning suppressed — same spec → 0 occurrences, tests pass
- [x] `CI=false` behaves like local (warning visible)
- [x] Filter is surgical: a non-matching `console.warn` still prints under `CI=true`; only the InteractionManager message is dropped
- [x] Full suite with `CI=true`: **0** `InteractionManager has been deprecated` occurrences across all suites; 142 suites pass — the single failure (`__tests__/components/blink-card/blink-card.spec.tsx`, locale/env-dependent `"12/ 28"` expectation) fails identically on untouched `main`
- [x] `ci/pipeline.yml` still parses as valid YAML (ytt `#@` lines are comments)

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -86,6 +86,8 @@ jobs:
       - name: repo
       run:
         path: pipeline-tasks/ci/tasks/test-unit.sh
+      params:
+        CI: true
 
 - name: check-code
   plan:
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,7 +4,7 @@ module.exports = {
     "./node_modules/@react-native/jest-preset/jest/setup.js",
     "./node_modules/react-native-gesture-handler/jestSetup.js",
   ],
-  setupFilesAfterEnv: ["@testing-library/jest-native/extend-expect"],
+  setupFilesAfterEnv: ["@testing-library/jest-native/extend-expect", "./jest.setup.js"],
   transform: {
     "\\.(ts|tsx)$": [
       "ts-jest",
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,0 +1,16 @@
+// In CI, suppress the noisy, non-actionable "InteractionManager has been deprecated"
+// warning emitted during tests by @react-navigation/* internals and by
+// app/screens/transaction-history/transaction-history-screen.tsx, so Concourse/CI
+// logs stay readable. Locally the warning is left intact so developers still see
+// the deprecation. See https://github.com/blinkbitcoin/blink-mobile/issues/3813
+const isCI = Boolean(process.env.CI) && process.env.CI !== "false"
+
+if (isCI) {
+  const SUPPRESSED_WARNINGS = [/InteractionManager has been deprecated/]
+  const originalWarn = console.warn.bind(console)
+  console.warn = (...args) => {
+    const message = typeof args[0] === "string" ? args[0] : ""
+    if (SUPPRESSED_WARNINGS.some((re) => re.test(message))) return
+    originalWarn(...args)
+  }
+}


esaugomez31 (approved):

openoms:
@grimen applied the ci change from here so Concourse adds the CI=true env:
```
./ci/repipe
Updating pipeline @ ciblink
could not find a valid token.
logging in to team 'dev'

navigate to the following URL in your browser:

  https://ci.blink.sv/login?fly_port=57809

target saved
jobs:
  job test-unit has changed:
  name: test-unit
  plan:
  - in_parallel:
      steps:
      - get: repo
        trigger: true
      - get: bundled-deps
        trigger: true
      - get: pipeline-tasks
  - config:
      image_resource:
        name: ""
        source:
          password: ((docker-creds.password))
          repository: us.gcr.io/galoy-org/galoy-mobile-pipeline
          username: ((docker-creds.username))
        type: registry-image
      inputs:
      - name: bundled-deps
      - name: pipeline-tasks
      - name: repo
+     params:
+       CI: "true"
      platform: linux
      run:
        path: pipeline-tasks/ci/tasks/test-unit.sh
    task: test-unit
  
pipeline name: blink-mobile

apply configuration? [yN]: y
configuration updated
unpaused 'blink-mobile'
```

PR #3769, opened 2026-04-29
feat(spark): UI/UX launch polish across settings and flows

esaugomez31:
## Summary

UI/UX launch polish for the self-custodial (Spark) experience: stable-balance gating, settings reorganization, account redesign, delete flow, default-account preference, receive flow QR by default, single-screen backup phrase, dynamic success messaging, stablecoins-aware Dollar modal, removal of the legacy "You Are in Control" trust-model modal, country-aware account-type gating, a comprehensive multi-account stack with auto-convert UX polish, and SDK amount-adjust block on the convert-details screen.

Primary ticket: blinkbitcoin/blink-wip#692 — *feat(spark/ui): Review and fixes*. Also addresses blinkbitcoin/blink-wip#664 (Stablesats/Dollar popup modal — non-custodial variant).

## Ticket blinkbitcoin/blink-wip#692 — Tasks

- [x] **1. Hide Stable Balance Setting (Feature Flag)** — `stableBalanceEnabled` remote flag (default `false`) gates `isStableBalanceActive`. Stable-balance settings hidden when off.
- [x] **2. Change Stable Balance Default Flag to False** — new wallets no longer auto-activate stable balance on creation.
- [x] **3. Move "Backup Phrase" to "Security & Privacy"** — relocated from Advanced.
- [x] **4. Hide "Advanced" Section for Non-Custodial Users** — Advanced group hidden when active wallet is SC; empty `SettingsGroup` no longer renders.
- [x] **5. Remove Modal with Alert in Transfer Flow (First Time Only)** — first-time convert modal dropped; unused i18n removed.
- [x] **6. Remove orange text from Review Transfer screen** — the silent SDK amount-adjust path is gone; the convert-details screen now blocks the Review-transfer button when the SDK would auto-adjust the requested amount and surfaces an inline error explaining the user must use the full balance. The confirm screen no longer shows orange adjustment text.
- [x] **7. Hide "Bitcoin Deposit Address" from "Ways to Get Paid"** — row hidden for all users; unused i18n removed.
- [x] **8. Add "Delete Account and Data" Button in Your Account Section** — Account screen bifurcated for SC. SC layout matches the Figma: avatar + LN address + "Non-custodial" subtitle, then `Wallet identifier` / `Lightning address` / `Backup status` boxed read-only fields (shared with the AccountInformation sub-screen via `SelfCustodialAccountFields`), then the danger zone with the delete flow (balance check + warning screen + SDK storage wipe + `activeAccountId` clear). Custodial layout untouched.
[...]

PR #3818, opened 2026-06-03
fix: persist default-account modal flag across sessions


### LINKED ISSUE #3811: fix(receive): Default account modal re-appears on every "Receive" tap — choice not persisted (regression since 2.4.46)
## Description

On release **2.4.50**, the "Select default account" modal (Bitcoin vs Dollar/Stablesats) pops up **every time** the user taps "Receive", even after the user has already made a selection. The user's choice is not being persisted across sessions.

This is a **regression** — the modal should only appear once (after the user has >= 1 transaction) and remember the choice permanently.

## Steps to Reproduce

1. Open Blink on v2.4.50 (custodial account, Mainnet, with at least 1 transaction)
2. Tap **Receive**
3. The "Select default account" modal appears — choose Bitcoin or Dollar
4. Go back to home screen
5. Tap **Receive** again
6. The modal appears **again**, despite having already chosen

## Expected Behavior

The modal should appear **once**. After the user selects a default account, the choice is persisted and the modal does not reappear.

## Actual Behavior

The modal reappears on every "Receive" tap. The persisted flag `hasPromptedSetDefaultAccount` is not being restored from the Apollo cache on app startup.

## Affected Versions

- **Regression introduced in:** 2.4.46
- **Still present in:** 2.4.50 (latest)

## Root Cause Analysis

The default account choice is persisted via a client-only Apollo cache field `hasPromptedSetDefaultAccount`, which is written to disk by `apollo3-cache-persist`. The trigger condition is in [`app/screens/home-screen/home-screen.tsx:411`](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/screens/home-screen/home-screen.tsx#L411):

```ts
if (
  !isSelfCustodial && target === "receiveBitcoin" && !hasPromptedSetDefaultAccount && // <-- reads false when cache not restored
  numberOfTxs >= 1 && galoyInstanceId === "Main"
) {
  toggleSetDefaultAccountModal()
  return
}
```

### Regressing commit

**Commit [`c276cf1ee8`](https://github.com/blinkbitcoin/blink-mobile/commit/c276cf1ee8) — PR #3769** ("feat(spark): UI/UX launch polish", 2026-05-20) changed cache restore in [`app/graphql/client.tsx:296-300`](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/graphql/client.tsx#L296) from unconditional to token-gated:

```diff
- await persistor.restore()
+ // Skip restore in self-custodial mode so the persisted custodial cache cannot leak.
+ if (token) {
+   await persistor.restore()
+ }
```

The `token` now comes from [`useEffectiveAuthToken()`](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/graphql/hooks/use-effective-auth-token.ts) which resolves **asynchronously**. On cold start it transiently returns `""` while the async account-type check (`listSelfCustodialAccounts()`) is in flight. If the Apollo client effect runs during this window, `token` is falsy and `persistor.restore()` is **skipped** — so `hasPromptedSetDefaultAccount` is never loaded back into the in-memory cache and defaults to `false`.

### Key files | File | Relevance |
|------|-----------|
| `app/graphql/client.tsx:296-312` | Token-gated `persistor.restore()` — the regression |
| `app/graphql/hooks/use-effective-auth-token.ts` | Async token resolution (transiently returns `""`) |
| `app/screens/home-screen/home-screen.tsx:402-420` | Modal trigger condition |
| `app/components/set-default-account-modal/set-default-account-modal.tsx` | Modal component + persist write |
| `app/graphql/client-only-query.ts:196-208` | `setHasPromptedSetDefaultAccount()` writer |
| `app/graphql/cache.ts:107-109` | Cache read policy (defaults to `false`) |

## Suggested Fix Direction

Ensure `persistor.restore()` runs for custodial sessions regardless of token-resolution timing. Options:

1. **Distinguish "still resolving" from "self-custodial (no token)"** in `useEffectiveAuthToken()` — e.g., return a tri-state (`loading | token | null`) — and only skip restore when explicitly in self-custodial mode, not during the loading phase.
2. **Restore unconditionally**, and instead purge/namespace the cache by account type when entering self-custodial mode.
3. **Await token resolution** before building the client, ensuring `token` is accurate by the time the restore decision is made.

---

/cc @grimen @esaugomez31 @ihaveadifferentname

esaugomez31:
## Summary

Fixes #3811 — the "Select default account" modal re-appearing on every **Receive** tap.

The flag tracking whether the modal has already been shown lived in the Apollo client-only cache (`hasPromptedSetDefaultAccount`). Its restore was [token-gated in PR #3769](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/graphql/client.tsx#L296-L300) and `useEffectiveAuthToken()` resolves asynchronously, so on cold start the restore is often skipped and the flag defaults to `false` — the modal re-appears.

This PR moves the flag out of Apollo cache into `persistentState`, keyed by `accountId`. Same persistence pattern used by PR #3795 (theme preference).

## Why persistentState

- Independent of Apollo client lifecycle and token-gating
- Loaded synchronously on app boot via `PersistentStateProvider`
- Per-account isolation comes for free (`defaultAccountModalShownByAccountId[id]`), which also fixes the secondary issue where the choice was lost on account switches

## Changes

**New**
- `app/store/persistent-state/default-account-modal-shown.ts` — `getDefaultAccountModalShown` / `markDefaultAccountModalShown`
- `app/hooks/use-default-account-modal-shown.ts` — `useDefaultAccountModalShown`
- `__tests__/store/persistent-state/default-account-modal-shown.spec.ts` — 9 specs covering per-account isolation, fallback, immutability

**Schema migration**
- `PersistentState_13` adds `defaultAccountModalShownByAccountId?: Record<string, boolean>`
- `migrate12ToCurrent` → bumps to 13, `migrate13ToCurrent` is identity

**Removed (Apollo client-only field)**
- `hasPromptedSetDefaultAccount` from `local-schema.gql`
- Type policy from `cache.ts`
- `setHasPromptedSetDefaultAccount` writer + query from `client-only-query.ts`
- Regenerated `generated.ts`

**Consumers migrated**
- `home-screen.tsx` — reads `defaultAccountModalShown` from the new hook
- `set-default-account-modal.tsx` — calls `markDefaultAccountModalShown()` instead of writing the Apollo cache

### DIFF
--- a/.storybook/views/story-screen.tsx
+++ b/.storybook/views/story-screen.tsx
@@ -5,7 +5,7 @@ const PersistentStateWrapper: React.FC<React.PropsWithChildren> = ({ children })
   <PersistentStateContext.Provider
     value={{
       persistentState: {
-        schemaVersion: 12,
+        schemaVersion: 13,
         galoyInstance: {
           id: "Main",
         },
--- a/__tests__/hooks/use-default-account-modal-shown.spec.ts
+++ b/__tests__/hooks/use-default-account-modal-shown.spec.ts
@@ -1,0 +1,81 @@
+import { act, renderHook } from "@testing-library/react-native"
+
+import { useDefaultAccountModalShown } from "@app/hooks/use-default-account-modal-shown"
+import { PersistentState } from "@app/store/persistent-state/state-migrations"
+
+const mockUpdateState = jest.fn()
+let mockPersistentState: PersistentState
+
+jest.mock("@app/store/persistent-state", () => ({
+  usePersistentStateContext: () => ({
+    persistentState: mockPersistentState,
+    updateState: mockUpdateState,
+  }),
+}))
+
+const baseState: PersistentState = {
+  schemaVersion: 13,
+  galoyInstance: { id: "Main" },
+  galoyAuthToken: "",
+}
+
+describe("useDefaultAccountModalShown", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockPersistentState = { ...baseState }
+  })
+
+  describe("defaultAccountModalShown (read)", () => {
+    it("reflects false when the flag has never been set", () => {
+      const { result } = renderHook(() => useDefaultAccountModalShown())
+
+      expect(result.current.defaultAccountModalShown).toBe(false)
+    })
+
+    it("reflects true when the active account's flag is set", () => {
+      mockPersistentState = {
+        ...baseState,
+        activeAccountId: "self-custodial-1",
+        defaultAccountModalShownByAccountId: { "self-custodial-1": true },
+      }
+
+      const { result } = renderHook(() => useDefaultAccountModalShown())
+
+      expect(result.current.defaultAccountModalShown).toBe(true)
+    })
+  })
+
+  describe("markDefaultAccountModalShown (write)", () => {
+    it("calls updateState with a functional updater, not a direct value", () => {
+      const { result } = renderHook(() => useDefaultAccountModalShown())
+
+      act(() => result.current.markDefaultAccountModalShown())
+
+      expect(mockUpdateState).toHaveBeenCalledTimes(1)
+      expect(typeof mockUpdateState.mock.calls[0][0]).toBe("function")
+    })
+
+    it("the captured updater sets the flag for the active account", () => {
+      const { result } = renderHook(() => useDefaultAccountModalShown())
+
+      act(() => result.current.markDefaultAccountModalShown())
+
+      const updater = mockUpdateState.mock.calls[0][0]
+      const next = updater({ ...baseState, activeAccountId: "self-custodial-1" })
+
+      expect(next.defaultAccountModalShownByAccountId).toEqual({
+        "self-custodial-1": true,
+      })
+    })
+
+    it("the captured updater returns a falsy value when prev is undefined (no write)", () => {
+      const { result } = renderHook(() => useDefaultAccountModalShown())
+
+      act(() => result.current.markDefaultAccountModalShown())
+
+      const updater = mockUpdateState.mock.calls[0][0]
+
+      expect(updater(undefined)).toBeFalsy()
+    })
+  })
+})
--- a/__tests__/persistent-storage.spec.ts
+++ b/__tests__/persistent-storage.spec.ts
@@ -33,7 +33,7 @@ it("migration from 5 to current returns ok with the migrated state", async () =>
   expect(result).toEqual({
     status: MigrationStatus.Ok,
     state: {
-      schemaVersion: 12,
+      schemaVersion: 13,
       galoyInstance: { id: "Main" },
       galoyAuthToken: "myToken",
     },
--- a/__tests__/store/persistent-state/default-account-modal-shown.spec.ts
+++ b/__tests__/store/persistent-state/default-account-modal-shown.spec.ts
@@ -1,6 +1,6 @@
 import {
   getDefaultAccountModalShown,
-  markDefaultAccountModalShown,
+  withDefaultAccountModalShown,
 } from "@app/store/persistent-state/default-account-modal-shown"
 import { PersistentState } from "@app/store/persistent-state/state-migrations"
 import { DefaultAccountId } from "@app/types/wallet"
@@ -76,14 +76,14 @@ describe("getDefaultAccountModalShown", () => {
   })
 })
 
-describe("markDefaultAccountModalShown", () => {
+describe("withDefaultAccountModalShown", () => {
   it("creates the per-account map when absent and sets the flag for the active id", () => {
     const state: PersistentState = {
       ...baseState,
       activeAccountId: "self-custodial-1",
     }
 
-    const next = markDefaultAccountModalShown(state)
+    const next = withDefaultAccountModalShown(state)
 
     expect(next.defaultAccountModalShownByAccountId).toEqual({
       "self-custodial-1": true,
@@ -100,7 +100,7 @@ describe("markDefaultAccountModalShown", () => {
       },
     }
 
-    const next = markDefaultAccountModalShown(state)
+    const next = withDefaultAccountModalShown(state)
 
     expect(next.defaultAccountModalShownByAccountId).toEqual({
       "self-custodial-1": true,
@@ -109,7 +109,7 @@ describe("markDefaultAccountModalShown", () => {
   })
 
   it("uses the custodial slot when activeAccountId is undefined", () => {
-    const next = markDefaultAccountModalShown(baseState)
+    const next = withDefaultAccountModalShown(baseState)
 
     expect(next.defaultAccountModalShownByAccountId).toEqual({
       [DefaultAccountId.Custodial]: true,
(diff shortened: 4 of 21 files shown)


esaugomez31:
* **#3818** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3818?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

## Critical

None.

---

## Important

- **I1 — Add the upgrade tradeoff to the PR description.**
  `app/store/persistent-state/state-migrations.ts:106-107`
  Users who already dismissed the modal under the Apollo mechanism will see it **one more time** after updating — the old flag lived in Apollo cache and is structurally unavailable to `migrate12ToCurrent`. One-time, then it sticks. QA should expect it.

- **I2 — Backdrop/close-icon dismiss does not mark the flag — pre-existing, unchanged.**
  `app/components/set-default-account-modal/set-default-account-modal.tsx:150-159`
  Only the BTC/USD selection handlers mark it; dismissing via backdrop re-prompts later. Identical on `main`, so not a regression — recorded so the semantics are a decision, not an accident.

---

## Design

**SOLID violations**

- None.

- **D1 — `resolveAccountKey` copy-pasted; the per-account pattern's 5th use has no shared key primitive.**
  `app/store/persistent-state/theme-preference.ts:7-8` + `app/store/persistent-state/default-account-modal-shown.ts:5-6`
  Character-for-character identical (`activeAccountId ?? DefaultAccountId.Custodial`), and it encodes the write-key-must-equal-read-key invariant this PR depends on — one definition means it can't silently diverge between features. Extract to a shared `app/store/persistent-state/account-key.ts` before the 6th copy. Keep `self-custodial-language.ts:5-9`'s resolver separate — different semantics (excludes custodial).

- None.

**Code smells**

- **D2 — Duplicated `markDefaultAccountModalShown(); toggleModal()` sequence in both handlers.**
  `set-default-account-modal.tsx:80,103`
  Extract `closeAsAcknowledged()` — dropping the mark from one branch silently reintroduces the bug for that choice path.

**Naming smells**

- **D3 — `markDefaultAccountModalShown` names two different things.**
  `app/store/persistent-state/default-account-modal-shown.ts:13` + `app/hooks/use-default-account-modal-shown.ts:23`
  A pure transformer `(state) => state` in the store and a void imperative callback returned by the hook — the hook imports one and exports the other under the identical name, so auto-import will pick the wrong one. Also breaks the sibling `with*` convention for pure transformers (`withThemePreference`, `withSelfCustodialLanguage`). Rename the store fn to `withDefaultAccountModalShown` — fixes both.
- `UseDefaultAccountModalShownReturn` vs siblings' `EffectiveThemeReturn` / `EffectiveLanguageReturn` (no `Use` prefix). Cosmetic.
- Stale spec title at `__tests__/store/persistent-state/state-migrations.spec.ts:113` — says *"preserves schema 11 …"* over a `schemaVersion: 12` input. Rename.

---

## Test coverage

- **`app/hooks/use-default-account-modal-shown.ts`** (25 LOC, new) — contains the only non-trivial wiring in the feature: the functional updater `updateState((prev) => prev && mark(prev))`. Exact copy-able precedent exists: `__tests__/hooks/use-effective-language.spec.ts:158-166` (mock context → capture updater from `mock.calls[0][0]` → apply to fake `prev` → assert).

### Tier 2 — tests that pass while a regression could ship invisibly

- **`__tests__/store/persistent-state/state-migrations.spec.ts`** — all 18 changed lines are mechanical `12→13` find-replace; `defaultAccountModalShownByAccountId` and `themeByAccountId` appear **nowhere** in the spec. No behavioral spec for the new v12→v13 step or the v13 identity — a future migration step that forgets to spread a field ships invisibly.
- **`__tests__/store/persistent-state/default-account-modal-shown.spec.ts`** — optional gaps: explicit `activeAccountId === DefaultAccountId.Custodial` case; empty-string accountId (`"" ?? x` keeps `""`). Low priority.

- **Hook updater semantics** (→ Tier 1) — `__tests__/hooks/use-default-account-modal-shown.spec.ts`: assert `markDefaultAccountModalShown` calls `updateState` with a **functional updater** (regressing to a direct-value update would clobber concurrent writes); apply the captured updater to a fake `prev` and assert the resulting map; assert the updater returns falsy (no write) when `prev` is undefined.
- **Migration field preservation** (→ Tier 2) — `"migrates a v12 state to v13 preserving themeByAccountId and self-custodial maps"`; `"v13 identity migration preserves defaultAccountModalShownByAccountId untouched"`.
- **Both modal handlers mark the flag** (→ D2) — mock the hook in a `set-default-account-modal` component spec and assert `markDefaultAccountModalShown` fires on **both** the keep-current and set-new paths — or skip the spec and rely on the `closeAsAcknowledged()` extraction (one of the two suffices).

esaugomez31:
@grimen 
## Important

**I1 — add the upgrade tradeoff to the PR description**
Done! Added to the description: users who dismissed the modal under the old Apollo mechanism will see it one more time after updating (the old flag lived in Apollo cache, unreachable from `migrate12ToCurrent`), then it sticks. QA should expect the one-time re-prompt.

**I2 — backdrop/close-icon dismiss does not mark the flag (pre-existing)**
Kept as-is by decision. Identical to `main`, so out of scope for this fix — only the BTC/USD selection handlers acknowledge the modal; backdrop/close intentionally re-prompts later. Recording it as a deliberate semantic rather than folding an unrelated behavior change into this PR.

## Design

**D1 — `resolveAccountKey` copy-pasted; no shared key primitive**
Done! Extracted to `app/store/persistent-state/account-key.ts`; `theme-preference.ts` and `default-account-modal-shown.ts` now both import the single definition. Left `self-custodial-language.ts`'s resolver separate (different semantics — excludes custodial).

**D2 — duplicated `mark(); toggle()` sequence in both handlers**
Done! Extracted `closeAsAcknowledged()`; both the keep-current and set-new paths call it, so dropping the mark from one branch can't silently reintroduce the bug.

**D3 — `markDefaultAccountModalShown` names two different things**
Done! Renamed the store transformer to `withDefaultAccountModalShown`, matching the `with*` sibling convention (`withThemePreference`, `withSelfCustodialLanguage`); the hook keeps the imperative `markDefaultAccountModalShown` callback, so auto-import can no longer pick the wrong one.
- Return type renamed `UseDefaultAccountModalShownReturn` → `DefaultAccountModalShownReturn` for sibling parity.
- Fixed the stale spec title that said "preserves schema 11" over a `schemaVersion: 12` input.

## Test coverage

**Tier 1 — `use-default-account-modal-shown.ts`**
Done! Added `__tests__/hooks/use-default-account-modal-shown.spec.ts` following the `use-effective-language` precedent: asserts `updateState` is called with a **functional updater**, applies the captured updater to a fake `prev` and asserts the resulting map, and asserts the updater returns falsy (no write) when `prev` is undefined.

**Tier 2 — migration field preservation**
Done! Added a v12→v13 test preserving `themeByAccountId` and the self-custodial maps, plus a v13-identity test asserting `defaultAccountModalShownByAccountId` passes through untouched.

**Tier 3 — per-finding regression**
Done! The hook updater semantics and migration field-preservation are covered above; the "both handlers mark the flag" case is covered by construction via the `closeAsAcknowledged()` extraction (D2), which you noted suffices over a component spec.

grimen (approved):

PR #559, opened 2022-07-12
feat: enable send payment when fee probe fails

daviroo:
Fixes #554 .  https://github.com/GaloyMoney/galoy-client/pull/53 needs to be merged first before this can be merged.

PR #3821, opened 2026-06-03
fix(send): USD wallet payment drains entire balance to BTC


### LINKED ISSUE #3809: USD wallet payment drains entire balance to BTC instead of retaining remaining USD balance
## Bug Report

**Reported by:** kemal (via Slack)
**App version:** iOS 2.4.49 (self-custodial)

## Steps to Reproduce

1. On **Device A** (self-custodial), transfer 1000 sats to the Dollar (USD) account → receives $0.69 at time of exchange
2. On **Device B**, create a Lightning invoice for 200 sats
3. On **Device A**, pay the 200 sat invoice **from the Dollar account**

## Expected Behavior

- The Dollar account on Device A should pay 200 sats worth (~$0.14) from the USD balance
- The remaining balance of approximately $0.55 should be **retained in the Dollar account**

## Actual Behavior

- The Dollar account pays the 200 sat invoice successfully
- However, the remaining ~$0.55 USD balance is **not retained** in the Dollar account
- Instead, the remaining amount is converted and transferred to the **Bitcoin account as 795 sats**
- The Dollar account ends up with $0.00 balance

## Root Cause

This is a **fundamental limitation of spark/USDB** — there is an 800 sat minimum for transfers between Bitcoin and Dollar accounts. When the remaining USD balance (795 sats worth) falls below this threshold after a payment, the system silently sweeps it back to the Bitcoin account.

The **Transfer flow** already handles this gracefully, but the **Send flow** does not.

## Impact

Users lose their USD-denominated position unexpectedly. The entire USD balance is liquidated when only a partial payment should have been deducted.

## Suggested Fix

The Send flow should handle the 800 sat minimum threshold the same way Transfer does — either:
1. Retain the sub-threshold USD balance as-is (do not force-convert it), or
2. Warn the user upfront that the remaining balance will be converted to BTC

/cc @grimen @esaugomez31

esaugomez31:
## Summary

Fixes #3809 — paying a Lightning invoice from the USD (USDB) account silently sweeps the entire remaining USD balance to BTC whenever the leftover would fall below the Spark/USDB protocol minimum (~800 sats equivalent).

The bug existed because the `Send` flow ignored the SDK signal `conversionEstimate.amountAdjustment === IncreasedToAvoidDust` that the `prepareSendPayment` call returns when the source balance would be left sub-threshold. The Convert flow already consumes this signal via `useSelfCustodialConversionGuard`; the Send flow did not.

This PR adds a non-blocking warning to the confirmation screen — user keeps control of the operation but is no longer surprised by the sweep.

## Why a warning, not a block

The ticket explicitly suggests two paths: (1) retain the sub-threshold balance, or (2) warn the user upfront. Option 1 is not feasible from the app side — the threshold lives in the SDK pool, not in app code. Option 2 is the realistic path: surface the existing SDK signal transparently and let the user decide.

## Repro (validated on regtest)

Starting state on a fresh self-custodial wallet:

| Step | Action | USD wallet | BTC wallet |
|---|---|---|---|
| 0 | Initial state | $0.00 | 1600 sats |
| 1 | Transfer 1000 sats → USD | $0.66 (1000 sats) | 600 sats |
| 2 | Pay 300-sat Lightning invoice from USD | **$0.00** ← bug: drains | **+695 sats** ← swept |

Expected: USD wallet keeps ~698 sats remainder. Actual (pre-fix): USD wallet drained entirely, the ~698 sats appear in BTC wallet.

## Architecture

The fix uses the **same code paths** as the existing Convert flow — zero duplication of detection/limits logic. Both flows now share:

- `resolveDustAdjustment` in `app/types/payment.ts` — the predicate that decides "is this a real dust sweep or is the user already draining intentionally"
- `mapAmountAdjustment` in `app/self-custodial/bridge/convert.ts` (now exposed via barrel) — maps the SDK `AmountAdjustmentReason` enum to the app's `ConvertAmountAdjustment`
- `useNonCustodialConversionLimits(ConvertDirection.BtcToUsd)` — fetches the threshold from `SDK.fetchConversionLimits()`, no hardcoded value

The new `useSendDustWarning` hook in `app/self-custodial/hooks/` is the Send analogue of `useSelfCustodialConversionGuard`:

```
SDK.prepareSendPayment
  → conversionEstimate.amountAdjustment
  → mapAmountAdjustment        ← shared bridge mapper
  → ConvertAmountAdjustment    ← shared app enum
  → useFee (Send) / useConversionQuote (Convert)
  → useSendDustWarning (Send) / useSelfCustodialConversionGuard (Convert)
  → resolveDustAdjustment      ← shared predicate
```

## Custodial untouched

Custodial users are unaffected:
- `GetFee.amountAdjustment` is an optional field; custodial fee probes (GraphQL mutations) never populate it
- `useNonCustodialConversionLimits` returns `null` for custodial (no SDK)
- Double-gate in the hook (`if (!fee.amountAdjustment) return HIDDEN; if (!limits?.minFromAmount) return HIDDEN`) ensures the warning is impossible to render outside the self-custodial USD-Lightning path

## Commits

Split for review clarity:

1. **`refactor(self-custodial): extract dust-adjustment helpers from Convert flow`** — pure refactor, byte-equivalent Convert behavior (verified by 15/15 guard tests + convert bridge spec)
2. **`fix(send): warn before USD wallet remainder is auto-converted to BTC`** — the actual fix plus tests, i18n, and screen integration

## Screens

Confirmation screen with the warning rendered when remaining USD would fall below the SDK-reported minimum:

> ⚠️ Remaining $0.45 (693 SAT) will be converted to Bitcoin. USD minimum: 800 SAT.

Visual treatment mirrors the existing `lightningRecommended` warning (icon + warning color + `p3` text), placed inside a `fieldContainer` for proper horizontal padding. The "Slide to Confirm" button is additionally disabled while the fee quote is loading, to prevent the user from sliding before the warning has a chance to surface.

## Translations

New i18n key `usdRemainderSweep` added with proper diacritics in all 28 languages. Each JSON file's original indent preserved (some use 2-space, others 4-space) — diff per language is exactly 3 lines.
Pasted image (2)

### DIFF
--- a/__tests__/screens/send-confirmation.spec.tsx
+++ b/__tests__/screens/send-confirmation.spec.tsx
@@ -4,6 +4,7 @@ import { Satoshis } from "lnurl-pay"
 import { act, fireEvent, render, screen } from "@testing-library/react-native"
 
 import { DisplayCurrency, toUsdMoneyAmount } from "@app/types/amounts"
+import { ConvertAmountAdjustment } from "@app/types/payment"
 import { WalletCurrency } from "@app/graphql/generated"
 import * as PaymentDetails from "@app/screens/send-bitcoin-screen/payment-details/intraledger"
 import { ConvertMoneyAmount } from "@app/screens/send-bitcoin-screen/payment-details/index.types"
@@ -14,6 +15,8 @@ import {
   Intraledger,
   LightningLnURL,
 } from "@app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.stories"
+
+import { flushEffects } from "../helpers/flush-effects"
 import { ContextForScreen } from "./helper"
 
 jest.mock("@app/store/persistent-state", () => ({
@@ -179,6 +182,14 @@ const mockUseSendBalances = jest.fn()
 jest.mock("@app/screens/send-bitcoin-screen/hooks/use-send-wallets", () => ({
   ...jest.requireActual("@app/screens/send-bitcoin-screen/hooks/use-send-wallets"),
   useSendBalances: () => mockUseSendBalances(),
+}))
+
+jest.mock("@app/self-custodial/hooks/use-non-custodial-conversion-limits", () => ({
+  useNonCustodialConversionLimits: () => ({
+    limits: { minFromAmount: 800, minToAmount: null },
+    loading: false,
+    error: null,
+  }),
 }))
 
 const useActiveWalletMock = jest.fn(() => ({
@@ -661,6 +672,96 @@ describe("SendBitcoinConfirmationScreen — fee-currency conversion", () => {
     )
 
     expect(screen.getByText(/exceeds your balance/i)).toBeTruthy()
+  })
+})
+
+describe("SendBitcoinConfirmationScreen — USD remainder sweep warning", () => {
+  const usdRemainderSweepMatcher = /will be converted to Bitcoin\. USD minimum:/i
+
+  beforeEach(() => {
+    mockUseSendBalances.mockReturnValue({
+      btcWallet: {
+        id: "btc-wallet-id",
+        balance: 0,
+        walletCurrency: WalletCurrency.Btc,
+      },
+      usdWallet: {
+        id: "usd-wallet-id",
+        balance: 1000,
+        walletCurrency: WalletCurrency.Usd,
+      },
+    })
+  })
+
+  it("renders the warning when fee quote reports IncreasedToAvoidDust and user is not draining balance", async () => {
+    mockUseFee.mockReturnValue({
+      status: "set",
+      amount: { amount: 0, currency: WalletCurrency.Usd, currencyCode: "USD" },
+      amountAdjustment: ConvertAmountAdjustment.IncreasedToAvoidDust,
+    })
+
+    render(
+      <ContextForScreen>
+        <Intraledger route={buildUsdSettlementRoute(200)} />
+      </ContextForScreen>,
+    )
+
+    await flushEffects()
+
+    expect(screen.getByText(usdRemainderSweepMatcher)).toBeTruthy()
+  })
+
+  it("does NOT render the warning when there is no amountAdjustment in the fee quote", async () => {
+    mockUseFee.mockReturnValue({
+      status: "set",
+      amount: { amount: 0, currency: WalletCurrency.Usd, currencyCode: "USD" },
+    })
+
+    render(
+      <ContextForScreen>
+        <Intraledger route={buildUsdSettlementRoute(200)} />
+      </ContextForScreen>,
+    )
+
+    await flushEffects()
+
+    expect(screen.queryByText(usdRemainderSweepMatcher)).toBeNull()
+  })
+
+  it("does NOT render the warning when the user is already draining the full USD balance", async () => {
+    mockUseFee.mockReturnValue({
+      status: "set",
+      amount: { amount: 0, currency: WalletCurrency.Usd, currencyCode: "USD" },
+      amountAdjustment: ConvertAmountAdjustment.IncreasedToAvoidDust,
+    })
+
+    render(
+      <ContextForScreen>
+        <Intraledger route={buildUsdSettlementRoute(1000)} />
+      </ContextForScreen>,
+    )
+
+    await flushEffects()
+
+    expect(screen.queryByText(usdRemainderSweepMatcher)).toBeNull()
+  })
+
+  it("does NOT render the warning for FlooredToMin (benign SDK floor)", async () => {
+    mockUseFee.mockReturnValue({
+      status: "set",
+      amount: { amount: 0, currency: WalletCurrency.Usd, currencyCode: "USD" },
+      amountAdjustment: ConvertAmountAdjustment.FlooredToMin,
+    })
+
+    render(
+      <ContextForScreen>
+        <Intraledger route={buildUsdSettlementRoute(200)} />
+      </ContextForScreen>,
+    )
+
+    await flushEffects()
+
+    expect(screen.queryByText(usdRemainderSweepMatcher)).toBeNull()
   })
 })
 
--- a/__tests__/screens/send-confirmation.spec.tsx
+++ b/__tests__/screens/send-confirmation.spec.tsx
@@ -3,7 +3,7 @@ import { TouchableOpacity, Text } from "react-native"
 import { Satoshis } from "lnurl-pay"
 import { act, fireEvent, render, screen } from "@testing-library/react-native"
 
-import { DisplayCurrency, toUsdMoneyAmount } from "@app/types/amounts"
+import { DisplayCurrency, toBtcMoneyAmount, toUsdMoneyAmount } from "@app/types/amounts"
 import { ConvertAmountAdjustment } from "@app/types/payment"
 import { WalletCurrency } from "@app/graphql/generated"
 import * as PaymentDetails from "@app/screens/send-bitcoin-screen/payment-details/intraledger"
@@ -611,6 +611,22 @@ const buildUsdSettlementRoute = (
   } as const
 }
 
+const buildBtcSettlementRoute = (unitOfAccountSats: number) => {
+  const btcDescriptor = { currency: WalletCurrency.Btc, id: "btc-wallet-id" } as const
+  const params: PaymentDetails.CreateIntraledgerPaymentDetailsParams<WalletCurrency> = {
+    handle: "test",
+    recipientWalletId: "testid",
+    convertMoneyAmount: usdBtcConvert,
+    sendingWalletDescriptor: btcDescriptor,
+    unitOfAccountAmount: toBtcMoneyAmount(unitOfAccountSats),
+  }
+  return {
+    key: "sendBitcoinConfirmationScreen",
+    name: "sendBitcoinConfirmation",
+    params: { paymentDetail: PaymentDetails.createIntraledgerPaymentDetails(params) },
+  } as const
+}
+
 describe("SendBitcoinConfirmationScreen — fee-currency conversion", () => {
   beforeEach(() => {
     // Balance: $10.00 = 1000 cents.
@@ -763,6 +779,36 @@ describe("SendBitcoinConfirmationScreen — USD remainder sweep warning", () => 
 
     expect(screen.queryByText(usdRemainderSweepMatcher)).toBeNull()
   })
+
+  it("does NOT render the warning for a BTC source wallet even when the fee quote reports IncreasedToAvoidDust (false-positive guard)", async () => {
+    mockUseSendBalances.mockReturnValue({
+      btcWallet: {
+        id: "btc-wallet-id",
+        balance: 1_000_000,
+        walletCurrency: WalletCurrency.Btc,
+      },
+      usdWallet: {
+        id: "usd-wallet-id",
+        balance: 1000,
+        walletCurrency: WalletCurrency.Usd,
+      },
+    })
+    mockUseFee.mockReturnValue({
+      status: "set",
+      amount: { amount: 0, currency: WalletCurrency.Btc, currencyCode: "BTC" },
+      amountAdjustment: ConvertAmountAdjustment.IncreasedToAvoidDust,
+    })
+
+    render(
+      <ContextForScreen>
+        <Intraledger route={buildBtcSettlementRoute(200)} />
+      </ContextForScreen>,
+    )
+
+    await flushEffects()
+
+    expect(screen.queryByText(usdRemainderSweepMatcher)).toBeNull()
+  })
 })
 
 describe("SendBitcoinConfirmationScreen — skipBalanceCheck matrix", () => {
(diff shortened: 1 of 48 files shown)


esaugomez31:
* **#3821** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3821?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3821?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

## Critical

### C1. Fee error swallows the dust signal AND the user can still slide-to-confirm

`app/screens/send-bitcoin-screen/use-fee.ts:21-35` + `app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx:541` + `app/self-custodial/payment-details/send-helpers.ts:53-62`

The slider is gated only on loading:

```tsx
disabled={!validAmount || hasAttemptedSend || fee.status === "loading"}
```

It is **not** disabled on `fee.status === "error"`. When the Spark `prepareSend` probe fails, `createGetFee` returns `{ amount: undefined }` — the new `amountAdjustment` threading only happens on the success path, and the `FeeType` error arm added in this PR makes it structurally impossible to carry (`amountAdjustment?: undefined`). `useSendDustWarning` receives `undefined` → `HIDDEN`.

Net effect: a flaky SDK call → no warning, slider enabled → user sweeps their USD balance into BTC unwarned — exactly the surprise this PR exists to prevent.

**Fix:** disable the slider on `fee.status === "error"` as well (or at minimum fail closed when the source wallet is self-custodial USD and the dust signal could not be evaluated). Fixing D2 (move `amountAdjustment` onto the `"set"` arm only) removes the type stub that makes the loss structural — one commit fixes the bug and the type.

### C2. Limits fetch race / failure silently hides the warning

`app/self-custodial/hooks/use-send-dust-warning.ts:48-60`

The hook consumes only `{ limits }` and ignores the `loading`/`error` states returned by `useNonCustodialConversionLimits`:

```ts
const { limits } = useNonCustodialConversionLimits(
  isUsdSource ? ConvertDirection.BtcToUsd : undefined,
)
...
if (!limits?.minFromAmount) return HIDDEN
```

- **Race:** fast fee + slow limits fetch → the slider enables the instant the fee resolves, before the warning is evaluable. User can confirm before the warning appears.
- **Failure:** if `fetchConversionLimits` rejects, the limits hook captures the error into its own `error` state, the dust hook never reads it, and the warning is permanently suppressed — "we couldn't fetch limits" is treated identically to "no dust problem exists."

**Fix:** consume `loading`/`error`. While limits are loading, disable the slider (treat like fee loading). On limits error when the dust adjustment **is** present, fail closed (block or show a degraded warning) and log — don't render nothing.

---

## Important

- **I1 — Minimum rendered in the wrong unit vs. its "USD minimum" label, and likely the wrong direction's limit.**
  `use-send-dust-warning.ts:48-49, 75`
  The hook fetches `ConvertDirection.BtcToUsd` limits, whose `minFromAmount` comes back raw in **sats** (`bridge/limits.ts:46`), then formats it via `toBtcMoneyAmount` — while the copy says *"USD minimum: {minimum}"*. The actual sweep is **UsdToBtc**, whose `minFromAmount` would be USD cents. The Convert flow formats this in the from-wallet currency (`conversion-details-screen.tsx:510-513`); Send is inconsistent with it. Fix the direction or the label.

- **I2 — `feeSats={fee.amount?.amount ?? 0}` masks a missing fee as a plausible zero.**
  `send-bitcoin-confirmation-screen.tsx:127`
  Overstates the "remaining" figure in the warning math. Derive `feeSats` only from `fee.status === "set"`.

- **I3 — Load-bearing comment dropped when extracting `resolveDustAdjustment`.**
  `app/types/payment.ts:204`
  The rationale for the full-drain `>=` suppression (`/** Only the dust gate blocks; FlooredToMin is a benign SDK floor… */`) didn't survive the move. Carry it over — the boundary is now shared by two call sites with no documentation.

---

## Design

**SOLID violations**

- **D1 — SRP: `useSendDustWarning` mixes data-fetching, the business gate, and i18n string formatting in one hook.**
  `use-send-dust-warning.ts:36-88`
  Its sibling `useSelfCustodialConversionGuard` stops at the decision and leaves formatting to the screen. Return the decision + raw `MoneyAmount`s; format in the component. This is also why the hook is only testable via full screen render.

- **D2 — ISP: `amountAdjustment` bolted onto all three `FeeType` union arms.**
  `use-fee.ts:21-35` + `payment-details/index.types.ts:68`
  The `"loading"`/`"error"` arms exist solely to carry `amountAdjustment?: undefined`; only the self-custodial USD-Lightning adapter ever populates it. Put it on the `"set"` arm only. This is also the structural cause of C1.

- **D3 — Convert-flow concept leaks into the generic fee contract.**
  `use-fee.ts:15` + `payment-details/index.types.ts:30`
  `ConvertAmountAdjustment` is now imported into the shared `GetFee` contract that custodial fee probes also implement — the dust concern should stay behind the self-custodial port instead of every fee implementer inheriting a field that is meaningless for them. Same fix as D2.

- None introduced. Measured: the new hook is flat early-return guards (6 branch points / 90 lines); the screen gains one conditional render + one boolean clause.

**Code smells**

- **D4 — New `dustWarning` style near-duplicates `feeWarning` with gratuitous diffs.**
  `send-bitcoin-confirmation-screen.tsx:636-654`
  `flex-start` vs `center`, `columnGap` vs `{" "}`, `flex: 1` vs `0.95` — two warning banners on one screen will drift. Factor a shared `WarningBanner` (icon + `p3` warning text) for both.
- See also I2 (`?? 0` default masking) and I3 (dropped comment), kept under Important because they affect displayed values/shared logic.

**Naming smells**

- None introduced by this PR. (`ConvertAmountAdjustment` reading as convert vocabulary inside Send code is the symptom of D3, not a rename target.)

---

## Test coverage

- **`app/self-custodial/hooks/use-send-dust-warning.ts`** — no direct spec. The remainder math (`balanceSats - sentSats - feeSats`), the three `formatMoneyAmount` conversions, and the limits-missing branch are unexercised as units — only reachable through the screen spec, which fully mocks `useFee` and never asserts rendered values. This is what let I1 ship.

- **`__tests__/screens/send-confirmation.spec.tsx`** (→ C1) — mocks `useFee` entirely, so it *injects* `amountAdjustment` rather than verifying the real chain (`createGetFee` → `useFee` → screen) produces it. The production wiring is untested at every level — cover it in jest by asserting `createGetFee`'s returned `amountAdjustment` directly (see Tier 3 "Wiring").
- **`__tests__/screens/send-confirmation.spec.tsx:679`** (→ I1) — warning assertion is a regex on the static sentence fragment (`/will be converted to Bitcoin\. USD minimum:/i`) — never asserts the rendered value or unit, so the sats-labeled-as-USD minimum passes untested.
- **`__tests__/screens/send-confirmation.spec.tsx:188-192`** (→ C2) — `useNonCustodialConversionLimits` mocked unconditionally to `{ minFromAmount: 800 }` — the loading/error/missing-limits branches are never executed.
- **(absent)** (→ C1) — no test sets `fee.status: "loading"` — the PR's own new slider-disable clause is unverified. Nothing covers `"error"` either.
- **(absent)** (→ false positive) — every dust-warning test uses a **USD** source wallet. A regression dropping or inverting the currency gate — firing the warning on ordinary BTC/custodial sends — would pass the entire suite. Highest-value missing counter-case.
- **`__tests__/self-custodial/payment-details/send-helpers.spec.ts`** (→ C1) — only patches the SDK mock to avoid an import error; no test feeds `conversionEstimate.amountAdjustment` and asserts `result.amountAdjustment`.
- **(absent)** (→ map arm) — `mapAmountAdjustment(FlooredToMin)` unasserted anywhere — trivial to lock down now that the function is exported.

- **C1** — screen case with `mockUseFee.mockReturnValue({ status: "error" })` + USD source → assert slider `disabled === true` (current code: enabled). Matrix row for `status: "loading"` → `disabled === true`.
- **C2** — mock limits hook returning `{ limits: null, loading: true }` and `{ limits: null, error }` with dust signal present → assert fail-closed (slider disabled / warning state), not silent HIDDEN.
- **I1** — `renderHook` on `useSendDustWarning` asserting the rendered `minimum` string's value **and unit** (USD for a USD-source sweep); or screen assertion on the full interpolated copy.
- **I2** — warning rendered with `fee.amount: undefined` → assert no numeric remainder displayed.
- **Wiring** — extend an existing `createGetFee` test: SDK resolves with `conversionEstimate.amountAdjustment` → assert `result.amountAdjustment === ConvertAmountAdjustment.IncreasedToAvoidDust`.
- **False positive** — screen case: BTC source wallet + `amountAdjustment: IncreasedToAvoidDust` → assert warning absent.
- **Map arm** — one-line unit test of `mapAmountAdjustment` over all three inputs.

esaugomez31:
@grimen 

## Critical

**C1 — fee error swallows the dust signal AND the user can still slide-to-confirm**
Done! Fixed via D2 as you suggested (one change fixes the bug and the type): `amountAdjustment` now lives only on the `"set"` arm, so the stub that structurally dropped the signal is gone. The slider fails closed when the fee couldn't be evaluated: `feeUnavailable = fee.status === "loading" || (fee.status === "error" && !fee.amount)`. This blocks the exact self-custodial case you named (`createGetFee` returns `{ amount: undefined }` on probe failure) while preserving the deliberate `#559` max-fee fallback (error **with** an amount stays sendable). Regression test added for both directions.

**C2 — limits fetch race / failure silently hides the warning**
Done! `useSendDustWarning` now consumes `{ limits, loading, error }`. While limits load it returns `pending` (screen disables the slider, same as fee loading); on error or missing minimum while a dust sweep is in play it returns `blocked` (slider disabled) and `reportError`s — no longer treats "couldn't fetch" as "no problem."

## Important

**I1 — minimum in the wrong unit/direction vs the "USD minimum" label**
Done! The hook now requests `ConvertDirection.UsdToBtc` (the actual sweep), whose `minFromAmount` is USD cents, and the value is carried as a `UsdMoneyAmount` and formatted in the from-wallet currency in the screen — consistent with the Convert flow and the label.

**I2 — `feeSats ?? 0` masks a missing fee as a plausible zero**
Done! `feeSats` is derived only from the settled fee (`settledFee = fee.status === "set" ? fee : undefined`, passed as `settledFee?.amount.amount`), so it's `undefined` rather than a phantom zero when the fee isn't set.

**I3 — load-bearing comment dropped from `resolveDustAdjustment`**
Done! Restored the rationale documenting both suppression conditions (FlooredToMin is a benign SDK floor; a full-balance send leaves no remainder to sweep), now that the boundary is shared by Convert and Send.

## Design

**D1 — SRP: hook mixed fetching, the gate, and i18n formatting**
Done! `useSendDustWarning` is now decision-only — it returns a `hidden | pending | blocked | visible` union carrying raw `MoneyAmount`s; all `formatMoneyAmount` calls moved to the screen, mirroring `useSelfCustodialConversionGuard`. It's now unit-testable without a screen render.

**D2 — ISP: `amountAdjustment` on all three `FeeType` arms**
Done! It lives on the `"set"` arm only; the `loading`/`error` arms no longer carry the `?: undefined` stub.

**D3 — convert-flow concept leaked into the generic fee contract**
Done! `amountAdjustment` is off the shared `GetFee` contract entirely (custodial probes no longer inherit it). It now lives behind the self-custodial port as `SelfCustodialFeeResult<T>` in `send-helpers.ts`; `use-fee.ts` reads it via a localized `import type` + cast. `index.types.ts` no longer imports `ConvertAmountAdjustment` at all.

**D4 — `dustWarning` style near-duplicated `feeWarning`**
Done! Extracted a shared `WarningBanner` (icon + `p3` warning text, optional `numberOfLines`), now backing both the lightning-recommended banner and the dust banner; the duplicated `feeWarningText`/`dustWarning`/`dustWarningText` styles were removed.

## Test coverage

**Tier 1 — `use-send-dust-warning.ts` had no direct spec**
Done! Added a direct `renderHook` spec covering the decision states (hidden/pending/blocked/visible), the I1 USD-unit assertion, the `UsdToBtc` direction, the `reportError`-on-failure path, and the BTC-source / FlooredToMin / full-drain negative cases.

**Tier 2 — tests passing while a bug is alive**
Done! `createGetFee` now has a direct wiring test asserting `conversionEstimate.amountAdjustment → result.amountAdjustment`; `mapAmountAdjustment` has a standalone unit over all three inputs; the false-positive counter-case (BTC source + `IncreasedToAvoidDust` → warning absent) is asserted at both the hook and screen level; the slider-disable clause is exercised for `fee.status` `loading` and `error`.

**Tier 3 — regression tests per finding**
Done! C1 (error-without-amount → slider disabled; error-with-amount → still enabled, #559), C2 (loading → pending, error → blocked + logged), I1 (minimum value **and** USD unit), false-positive (BTC source), and the `createGetFee` wiring are all covered.

grimen (approved):

PR #3823, opened 2026-06-04
fix(conversion): preserve chip-press amount when focused-input identity changes mid-cycle

esaugomez31:
## Result

Pressing a percentage chip (25 / 50 / 75 / 100%) on the conversion screen now applies the amount and clears the loading state correctly, even on a fresh entry without any prior typing in the input. The bug affected both custodial and self-custodial accounts.

Before the fix the chip showed an indefinite spinner and the amount never reached the inputs on the fresh-entry path. After the fix the full cycle completes within the normal debounce window (~1s) and the converted value renders in the other input as expected.

### DIFF
--- a/__tests__/components/amount-input-screen.spec.tsx
+++ b/__tests__/components/amount-input-screen.spec.tsx
@@ -12,32 +12,40 @@ const mockFormatMoneyAmount = jest.fn()
 const mockOnAmountChange = jest.fn()
 const mockOnSetFormattedAmount = jest.fn()
 
-jest.mock("@app/hooks/use-display-currency", () => ({
-  useDisplayCurrency: () => ({
-    currencyInfo: {
-      BTC: {
-        symbol: "",
-        minorUnitToMajorUnitOffset: 0,
-        showFractionDigits: false,
-        currencyCode: "SAT",
-      },
-      USD: {
-        symbol: "$",
-        minorUnitToMajorUnitOffset: 2,
-        showFractionDigits: true,
-        currencyCode: "USD",
-      },
-      DisplayCurrency: {
-        symbol: "$",
-        minorUnitToMajorUnitOffset: 2,
-        showFractionDigits: true,
-        currencyCode: "USD",
-      },
+jest.mock("@app/hooks/use-display-currency", () => {
+  const currencyInfo = {
+    BTC: {
+      symbol: "",
+      minorUnitToMajorUnitOffset: 0,
+      showFractionDigits: false,
+      currencyCode: "SAT",
     },
-    formatMoneyAmount: mockFormatMoneyAmount,
-    zeroDisplayAmount: { amount: 0, currency: "DisplayCurrency", currencyCode: "USD" },
-  }),
-}))
+    USD: {
+      symbol: "$",
+      minorUnitToMajorUnitOffset: 2,
+      showFractionDigits: true,
+      currencyCode: "USD",
+    },
+    DisplayCurrency: {
+      symbol: "$",
+      minorUnitToMajorUnitOffset: 2,
+      showFractionDigits: true,
+      currencyCode: "USD",
+    },
+  }
+  const zeroDisplayAmount = {
+    amount: 0,
+    currency: "DisplayCurrency",
+    currencyCode: "USD",
+  }
+  return {
+    useDisplayCurrency: () => ({
+      currencyInfo,
+      formatMoneyAmount: mockFormatMoneyAmount,
+      zeroDisplayAmount,
+    }),
+  }
+})
 
 jest.mock("@app/hooks/use-debounce", () => ({
   useDebouncedEffect: (callback: () => void) => {
@@ -130,5 +138,52 @@ describe("AmountInputScreen", () => {
 
     const errorElement = getByTestId("error-message")
     expect(errorElement.props.children).toBe("")
+  })
+
+  it("propagates a fresh initialAmount even when focusedInput identity changes in the same commit", () => {
+    const initialFocusedInput = {
+      id: ConvertInputType.FROM,
+      currency: "BTC" as const,
+      formattedAmount: "",
+      isFocused: true,
+      amount: toBtcMoneyAmount(0),
+    }
+
+    const { rerender } = render(
+      <AmountInputScreen
+        inputValues={defaultInputValues}
+        onAmountChange={mockOnAmountChange}
+        convertMoneyAmount={mockConvertMoneyAmount}
+        onSetFormattedAmount={mockOnSetFormattedAmount}
+        focusedInput={initialFocusedInput}
+      />,
+    )
+
+    mockOnAmountChange.mockClear()
+
+    const chipAmount = toBtcMoneyAmount(25000)
+    const refocusedInput = {
+      ...initialFocusedInput,
+      amount: {
+        ...initialFocusedInput.amount,
+        currency: "USD" as const,
+        currencyCode: "USD",
+      },
+    }
+
+    rerender(
+      <AmountInputScreen
+        inputValues={defaultInputValues}
+        onAmountChange={mockOnAmountChange}
+        convertMoneyAmount={mockConvertMoneyAmount}
+        onSetFormattedAmount={mockOnSetFormattedAmount}
+        focusedInput={refocusedInput}
+        initialAmount={chipAmount}
+      />,
+    )
+
+    expect(mockOnAmountChange).toHaveBeenCalledWith(
+      expect.objectContaining({ amount: 25000 }),
+    )
   })
 })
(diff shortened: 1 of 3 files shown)


esaugomez31:
* **#3823** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

## Critical

None found.

---

## Important

- **I1 — Hardcoded magic value used as both input and expectation.**
  `__tests__/components/amount-input-screen.spec.tsx:164, 186`
  Bare `25000` is the chip input and the assertion value with nothing tying them together — they can drift independently. Name it: `const CHIP_AMOUNT_SATS = 25000` and reference it in both places.

---

## Design

**SOLID violations**

- **D1 — SRP: state-machine-in-disguise.**
  `app/components/transfer-amount-input/amount-input-screen.tsx:149-167`
  The component coordinates **nine mutable refs** (`lastValuesRef`, `lastSnapshotRef`, `skipNextRecalcRef`, `focusedIdRef`, `prevFocusSigRef`, `freezeFormatRef`, `typingRef`, `forceDebounceRef`, + this PR's `initialAmountAppliedRef`) across four effects plus a debounced callback — an implicit state machine as a bag of order-dependent booleans. Each fix adds another flag; complexity grows super-linearly. **Fix (follow-up ticket, not this PR):** collapse into a `useReducer` state machine (`idle / focusing / typing / applyingChipAmount / debouncing`) so "chip applied, don't recalc on focus" is a first-class transition.

- None introduced.

- **D2 — Focus-sig effect now branches on five independent flags.**
  `amount-input-screen.tsx:243-290`
  `prevFocusSigRef`, `skipNextRecalcRef`, `initialAmountJustApplied`, `forceDebounceRef`, amount/currency mismatch — this PR adds two of those paths. Acceptable for a hotfix; folds into D1's `useReducer` follow-up.

**Code smells**

- **D3 — Invisible effect-ordering invariant.**
  `amount-input-screen.tsx:235-251`
  The fix only works because the `initialAmount` effect (`:235`) is declared before the focus-sig effect (`:243`); React runs effects in declaration order. Nothing notes or protects this — reordering the blocks silently reintroduces the bug. **Fix (this PR):** one-line comment at the consume site (`:251`).
- **D4 — Uncommented consume-once token + early-return.**
  `amount-input-screen.tsx:251-274`
  The read-then-clear (`:251-252`) and the `if (initialAmountJustApplied) return` that *is* the fix (`:274`) ship with zero rationale; a reader can't tell why bailing before `setNumberPadAmount` is safe. **Fix (this PR):** 1–2 lines above `:274` — number pad already holds the chip amount; the debounced recalc re-pushes formatting.
- **D5 — Self-referential sticky flag.**
  `amount-input-screen.tsx:268-272`
  `forceDebounceRef.current = initialAmountJustApplied || forceDebounceRef.current || ...` — set in three places, cleared in two, across ~250 lines. **Fix (this PR):** comment naming it "sticky until next debounce"; lifecycle cleanup folds into D1.

**Naming smells**

- **D6 — `initialAmount` is not initial.**
  `amount-input-screen.tsx:40`
  Per this PR's own title it's a *chip-press amount applied mid-cycle*; "initial" implies one-time/at-mount — exactly the wrong mental model. **Fix (this PR, new code only):** `initialAmountAppliedRef` → `chipAmountAppliedRef` (`:167, :239, :251`), `initialAmountJustApplied` → `chipAmountJustApplied`. The public prop rename goes with D1's follow-up.
- **D7 — Test name uses React jargon.**
  `amount-input-screen.spec.tsx:143`
  "…identity changes in the same commit" reads as a *git* commit. **Fix:** "in the same render".

---

## Test coverage

- None — both changed files have co-located coverage. The new test is a verified genuine regression test (fails on pre-fix code: chip `25000` clobbered to `0`).

### Tier 2 — tests that pass while a regression could be alive

Each should be made to **fail** on the current code first where applicable.

- **`__tests__/components/amount-input-screen.spec.tsx:143-188`** (→ the fix's mechanism) — asserts the outcome (`25000` arrives) but not the mechanism: it would still pass if the early-return at `:274` were removed but the value eventually arrived, because the `useDebouncedEffect` mock (`:50-56`) runs synchronously. Add `expect(mockOnAmountChange).not.toHaveBeenCalledWith(expect.objectContaining({ amount: 0 }))` to pin the clobber.
- **(same mock)** (→ spinner-clear) — the synchronous debounce mock means the spinner-clear half of the bug (`notifyTyping(false)` / `onAfterRecalc`) is never exercised at unit level; covered only via the integration spec.

- **USD as the FROM wallet** (crit 4/10) — every integration percent scenario in `conversion-details-screen.spec.tsx` uses `btcBalance: 100000, usdBalance: 50000`, so the source is always BTC. Add `btcBalance: 0, usdBalance: 50000` → press 25% → `assertConversionValues`.
- **Self-custodial chip happy-path** (crit 3/10) — SC tests only assert gating (limit failure / quote error / boot window); the PR claims both account types. Add: SC account, limits OK → press chip → advance timers → Next enables, nav params carry the amount.
- **Unit-level spinner-clear** (crit 3/10) — sibling test with fake timers (not the synchronous debounce mock) asserting `onTypingChange(false, …)` / `onAfterRecalc` fires after the same-commit identity race.
- **Consecutive chip presses** (crit 2/10) — 25% → 100% without intervening settle. Partially exercised by scenarios 43–45; low risk.

esaugomez31:
@grimen

## Important

**I1 — hardcoded magic value used as both input and expectation**
Done! Extracted `const CHIP_AMOUNT_SATS = 25000` and reference it at both the chip input and the assertion, so they can't drift independently.

## Design

**D1 / D2 — state-machine-in-disguise + focus-sig branching on five flags**
Agreed, and kept out of this hotfix as you suggested. The order-dependent refs really do want to become an explicit `useReducer` (`idle / focusing / typing / applyingChipAmount / debouncing`), so "chip applied, don't recalc on focus" becomes a first-class transition instead of another boolean. Tracking it as the follow-up.

**D3 — invisible effect-ordering invariant**
Done! One-liner at the consume site noting the `initialAmount` effect must stay declared before the focus-sig effect (React runs effects in declaration order), so reordering the blocks can't silently reintroduce the bug.

**D4 — uncommented consume-once token + early-return**
Done! Comment above the early-return: the number pad already holds the chip amount, so we bail before the focus reset and let the forced debounce re-push formatting instead of clobbering it to 0.

**D5 — self-referential sticky flag**
Done! `forceDebounceRef` is now annotated "sticky until the next debounce" (set on chip-apply and focus changes, cleared only when the debounce fires).

**D6 — `initialAmount` is not initial**
Done! Renamed the new code only: `initialAmountAppliedRef` > `chipAmountAppliedRef`, `initialAmountJustApplied` > `chipAmountJustApplied`. Left the public `initialAmount` prop as-is for the D1 follow-up, per your note.

**D7 — test name uses React jargon**
Done! "...in the same commit" > "...in the same render".

## Test coverage

**Tier 2 — pin the clobber mechanism**
Done, with one adjustment worth flagging. The literal `not.toHaveBeenCalledWith({ amount: 0 })` false-fails on the fixed code: the `useDebouncedEffect` mock runs synchronously, so it fires once on the intermediate `useReducer` render (before the chip dispatch lands) and a transient `0` is always emitted, fix or no fix. The mechanism-faithful pin is asserting the *final* emitted amount is the chip value: `mockOnAmountChange.mock.calls.at(-1)?.[0]` equals `CHIP_AMOUNT_SATS`. That still fails on pre-fix code (where the focus-sig clobber to 0 wins last) and passes on the fix.

**Tier 3 — USD as the FROM wallet**
Done, with a caveat on the setup. `btcBalance: 0` alone doesn't make USD the source: the screen defaults the from-wallet to BTC, so 25% of a 0-balance BTC wallet just yields an empty field (orthogonal from-selection behavior, not this fix). I make USD the source via a `toggle` first (both balances present), then press 25%, which gives `$125.00` (25% of 50000) and validates via `assertConversionValues`.

**Tier 3 — self-custodial chip happy-path**
Done! New test: self-custodial account + limits OK, press the chip, advance timers, Next enables and the nav params carry the amount. The existing self-custodial tests only asserted gating (limit failure / quote error / boot window), so this is the positive path. It also exercises the spinner-clear half end-to-end (real debounce + fake timers).

**Tier 3 — consecutive chip presses**
Done! Added 25% then 100% as a scenario.

**Tier 3 — unit-level spinner-clear**
Left out. The spinner-clear path is already covered end-to-end by the self-custodial happy-path test above (real debounce + fake timers). A unit-level version would need a separate spec file without this one's module-level synchronous `useDebouncedEffect` mock, for no added coverage over the integration test

grimen (approved):

PR #3824, opened 2026-06-04
fix(lnurl-withdraw): support self-custodial vouchers end-to-end


### LINKED ISSUE #3808: fix(self-custodial): LNURL withdraw voucher redemption button unresponsive with self-custodial account
## Bug Report

**Reported by:** @pretyflaco (kemal)
**Version:** 2.4.49
**Platform:** iOS (likely also Android)

## Description

When redeeming an LNURL withdraw voucher (generated from terminal.blinkbtc.com) using a **self-custodial** (external signing device) account, the redeem screen loads correctly but the "Redeem Bitcoin" button does not respond to taps. The same flow works correctly with a **custodial** account.

## Expected behavior

The "Redeem Bitcoin" button should work for both custodial and self-custodial accounts, completing the LNURL withdraw flow.

## Actual behavior

The "Redeem Bitcoin" button is unresponsive when a self-custodial account is selected. No error message is shown — the button simply doesn't react.

## Steps to reproduce

1. Generate an LNURL withdraw voucher on terminal.blinkbtc.com
2. Open Blink app (v2.4.49), select a **self-custodial** account
3. Scan the LNURL voucher QR code
4. Arrive at the redeem screen (this part works correctly)
5. Tap "Redeem Bitcoin"
6. Observe: button does not respond

**Workaround:** Switch to a custodial account to redeem the voucher.

## Screenshots

_(Attached in the linked Slack thread — redeem screen with unresponsive button)_

## Analysis

Likely the LNURL withdraw callback requires generating a Lightning invoice, and the self-custodial wallet path may not have the invoice generation handler wired up for this flow, or the button's onPress handler has a conditional that silently fails for non-custodial wallet types.

/cc @grimen @esaugomez31

esaugomez31:
## Summary

Fixes #3808. Self-custodial accounts could not redeem LNURL withdraw vouchers because two separate bugs blocked the end-to-end flow:

1. **Redemption screen**: the detail screen silently bailed when `btcWalletId` was undefined (SC has no Blink-side wallet), so the result screen was never reached.
2. **QR scanner**: `sdk.parse` from the Breez Spark SDK hung indefinitely on any non-Spark bech32 input (e.g. LNURL strings), blocking the entire destination-resolution pipeline before the LNURL parser could run.

### Redemption flow (`app/screens/redeem-lnurl-withdrawal-screen/`)

- Extracted custodial and self-custodial redemption into a single hook `useLnurlWithdrawRedemption` that branches by `accountType` and exposes a uniform `{ paid, pending, errorMessage, lnServiceErrorReason }` shape to the screen.
- **Custodial path preserved 1:1**: same `lnInvoiceCreate` mutation, same k1+pr callback POST, same `useLnUpdateHashPaid` subscription wait, same `HomeAuthedDocument` refetch on paid (relocated from a render-phase `useMemo` into a proper `useEffect`, same observable effect).
- **Self-custodial path**: new `lnurlWithdraw` adapter on the SC bridge wraps `sdk.lnurlWithdraw` with `completionTimeoutSecs: 120`. On resolve it branches on the SDK's authoritative `payment.status`: `Completed` maps to success; otherwise (the completion timeout elapsed with no settled payment) it maps to a distinct `pending` state, so the screen never reports premature success. Thrown errors are reported to Crashlytics (`reportError`), classified via the existing `classifySdkError`, localised via `useTranslateSdkError`, and a genuine `LnurlError` service reason is threaded to `lnServiceErrorReason`. While the SC SDK is still connecting, a bounded wait surfaces a "wallet not connected" message instead of an indefinite spinner.
- Detail screen drops the silent `btcWalletId && navigation.replace(...)` guard, gates the custodial wallet query with `skip: isSelfCustodial`, and forwards `receivingWalletDescriptor` as optional so SC nav goes through.
- Result screen is now a thin renderer (286 lines > 165 lines) consuming the hook. Custodial and self-custodial share the same `payment-success` icon, error rendering, and `ActivityIndicator`; self-custodial additionally renders a "payment may still be completing" message when the SDK can't confirm settlement within the timeout.

### QR scanner unblock (`app/self-custodial/`)

- Added `hasSparkAddressShape` in `config.ts` (co-located with `SparkToken` and other Spark protocol constants).
- `parseSparkAddressDetailed` short-circuits to `NotSparkAddress` when the input cannot structurally be a Spark address (HRPs `sp1` / `sprt1`, case-insensitive, after trim). Inputs outside the SDK's contract never reach `sdk.parse`.

### Supporting

- New `lnurlWithdraw` slot on `usePayments` (custodial gets `undefined`, SC gets the bridge adapter when the SDK is connected).
- New `satsToMsats` / `msatsToSats` helpers in `app/utils/amounts.ts` (used by the bridge and the detail screen).

## Notes

- The SC `completionTimeoutSecs` default of 120s is intentional, not arbitrary. Custodial waits on the GraphQL `myLnUpdates` subscription with no client-side timeout. SC has no equivalent subscription path; the SDK timeout is the wait mechanism. A future improvement could surface a configurable timeout if Rain (or other LNURL services) routinely exceed 2 minutes.
- `hasSparkAddressShape` lives in `config.ts` alongside `SparkToken` rather than in a dedicated file: both are Spark protocol constants tied to the same domain


esaugomez31:
* **#3824** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3824?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3824?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

## Critical

### C1. SC path: permanent spinner when the SDK isn't connected at screen open

`app/screens/redeem-lnurl-withdrawal-screen/hooks/use-lnurl-withdraw-redemption.ts:62` + `app/hooks/use-payments.ts`

```ts
useEffect(() => {
  if (!enabled || !lnurlWithdraw) return
```

`usePayments` hands SC accounts `lnurlWithdraw: undefined` until the SDK is connected. Opening the redeem screen while the SC wallet is still initializing hits this guard: `paid=false`, `errorMessage=""` → the result screen renders the `ActivityIndicator` branch forever. No error, no timeout, no retry signal. The effect re-runs if the SDK connects later; if it never does (boot failure, storage contention), the user is stranded.

**Fix:** when `enabled && !lnurlWithdraw`, surface a distinct "wallet not connected" error — immediately or after a bounded wait.

### C2. SC 120s timeout reported as a flat "network error" — the payment may still complete

`app/self-custodial/bridge/lnurl-withdraw.ts:42-46` + `app/self-custodial/sdk-error.ts:42`

```ts
} catch (err) {
  const code = classifySdkError(err)
  return { status: PaymentResultStatus.Failed, errors: [{ message: code }] }
}
```

When `completionTimeoutSecs: 120` expires the SDK rejects and `classifySdkError` maps `"timeout"` → `NetworkError`. By then the LNURL callback has already fired — the service may still settle after the timeout. The user is told it *failed*, may retry the voucher or contact support, then funds arrive. `PaymentResultStatus.Pending` exists in `app/types/payment.ts:13` but is produced nowhere in `app/`.

**Fix:** detect the completion-timeout case in the adapter (before `classifySdkError` collapses it) and return `PaymentResultStatus.Pending`; render "payment may still be completing — check your balance before retrying" rather than a failure.

---

## Important

- **I1 — `run()` is a floating promise with no `.catch`.**
  `use-lnurl-withdraw-redemption.ts:89`
  Any synchronous throw (param marshaling, `satsToMsats`, a future adapter edit) is a swallowed rejection → permanent spinner. Add a defensive `.catch` setting a generic error. (this PR)

- **I2 — SC `lnServiceErrorReason` hardcoded to `""`.**
  `use-lnurl-withdraw-redemption.ts:108`
  Custodial forwards the LNURL service's human-readable `reason` to the UI (`:186`); SC discards it. Thread it through from `LnurlError` if the SDK surfaces it. (this PR)

- **I3 — SC adapter classifies and swallows the raw error.**
  `bridge/lnurl-withdraw.ts:45`
  An unknown SDK tag shows "generic error" with zero diagnostic trail in Sentry/Crashlytics. Log the raw error (tag + inner) before returning the classified code. (this PR)

- **I4 — All four custodial error paths log via bare `console.error`.**
  `use-lnurl-withdraw-redemption.ts:157, 163, 184, 189`
  Nothing reaches Crashlytics. (this PR, or fold into I5's ticket)

- **I5 — Pre-existing defects carried 1:1 from the old result screen.**
  Verified byte-for-byte against merge-base `4fa04eb53`; now isolated in the hook, ~20 lines total. (follow-up ticket, not this PR — fixing them here would break the 1:1-preservation property that makes the custodial side easy to verify)
  - Unhandled rejections in `submitToCallback` — `new URL`, `fetch`, `result.json()` have no try/catch and the call is a floating promise; LNURL service unreachable or non-JSON 200 → permanent spinner, no log (`use-lnurl-withdraw-redemption.ts:171-194`)
  - `console.error(result.text(), ...)` logs a Promise — the failed response body is never captured (`:189`)
  - Raw `` `${err}` `` rendered to the user on invoice-create throw (`:164`)
  - Double `HomeAuthedDocument` refetch on custodial success — the `withMyLnUpdateSub` HOC refetches on every `PAID` (`my-ln-updates-sub.tsx:34`) and the hook refetches again when `paid` flips (`:202`)

---

## Design

**SOLID violations**

- **D1 — ISP: mode-dependent return shape on the public hook.**
  `use-lnurl-withdraw-redemption.ts:108, 186`
  `useLnurlWithdrawRedemption` exposes `lnServiceErrorReason` that is meaningful for custodial but hardcoded `""` for SC — same pattern flagged on `useEffective{DisplayCurrency,Language}` in the #3769 review. Fixing I2 (thread the SDK's reason) dissolves it. (this PR)

- **D2 — boolean `enabled` mode dispatch.**
  `use-lnurl-withdraw-redemption.ts:36, 112, 212-225`
  Both sub-hooks are always invoked, one disabled by flag. Effects are properly gated so no state leakage; acceptable for internal sub-hooks, but it's the same mode-dependent architecture the cleanup ticket tracks. (follow-up ticket, not this PR)

- None introduced. `createLnurlWithdraw` follows the established bridge pattern (`bridge/index.ts:19`, `LnurlWithdrawAdapter` in `types/payment.ts:189-191`, wired through `usePayments` like `SendPaymentAdapter` et al.). Only 2 mode-branches added at call sites (`redeem-bitcoin-detail-screen.tsx:58` `skip: isSelfCustodial`, `:99` wallet guard), both justified.

- None introduced. Measured: `useMutationRedemption` ~8 branch points / 87 lines (justified by the invoice-create → callback-submit → subscription-wait choreography); `useAdapterRedemption` 3; the dispatch hook 1 ternary; detail-screen `navigate()` 2.

**Code smells**

- **D3 — error-reporting inconsistency in the added code.**
  `console.error` ×4 in the custodial sub-hook, complete silence in the SC adapter, zero Crashlytics anywhere — see I3/I4. Same four-way-inconsistency theme recorded on the #3769 cleanup ticket. (this PR for the new SC adapter; rest with I5's ticket)

**Naming smells**

- **D4 — `useAdapterRedemption` / `useMutationRedemption` name the transport, not the domain.**
  `useSelfCustodialRedemption` / `useCustodialRedemption` would say what they are instead of how they talk. Rename-only. (this PR)

---

## Test coverage

- None — every new module ships a spec.

- **`__tests__/screens/redeem-lnurl-withdrawal-screen/hooks/use-lnurl-withdraw-redemption.spec.tsx`** (→ C1) — the `lnurlWithdraw`-undefined case asserts the silent early return, i.e. it encodes the eternal spinner as expected behavior. Should assert an error state surfaces.
- **`__tests__/self-custodial/bridge/lnurl-withdraw.spec.ts`** (→ C2) — timeout rejection asserted as `Failed` + network-error code, encoding C2 by assertion. Should assert a distinct pending outcome.
- **`use-lnurl-withdraw-redemption.spec.tsx`** (→ stale-error reset) — `errorMessage`/`lnServiceErrorReason` are only ever set, never reset. Fail with `status:"ERROR"`, rerender with new `amountSats` + OK response + matching hash → assert `errorMessage===""`, `paid===true`. **Fails today** — ghost error renders over a succeeding flow.
- **`use-lnurl-withdraw-redemption.spec.tsx`** (→ stale resolution) — SC re-render mid-flight: changed `amountSats` re-runs `lnurlWithdraw`; no test that the stale first resolution can't flip state after params changed. Deferred first promise, rerender, resolve first as Failed → assert no error set and adapter called twice with the new amount.
- **`use-lnurl-withdraw-redemption.spec.tsx`** (→ contract) — custodial steady-state pending (invoice created, callback OK, hash not yet seen) never pinned as `paid=false, errorMessage=""`.
- **`__tests__/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-detail-screen.spec.tsx`** (→ rounding wire-through) — only clean msats values (1000→1). A voucher with `minWithdrawable: 1500` msats should assert `minWithdrawableSatoshis.amount === 2` — the wire-through of the `msatsToSats` rounding that `amounts.spec.ts` covers in isolation.
- **`__tests__/screens/redeem-lnurl-withdrawal-screen/redeem-bitcoin-result-screen.spec.tsx`** (→ HOC contract) — mocks the redemption hook entirely; fine as a presentational split, but nothing pins the `withMyLnUpdateSub` wrapping the custodial paid signal depends on (`redeem-bitcoin-result-screen.tsx:132`). An accidental unwrap ships invisibly.

- **C1** — mock `usePayments` with `lnurlWithdraw: undefined`, `enabled: true` → assert an error message surfaces (post-fix), not a permanent `paid=false, errorMessage=""`.
- **C2** — mock `sdk.lnurlWithdraw` rejecting with a timeout-tagged error → assert `PaymentResultStatus.Pending` (post-fix) and the may-still-complete message, not `sc_network_error`.
- **I1** — make the mocked adapter throw synchronously → assert a generic error message surfaces instead of a swallowed rejection.
- **I2** — mock the SDK rejecting with an `LnurlError` carrying a service `reason` → assert it reaches `lnServiceErrorReason`.

esaugomez31:
Hi @grimen

## Critical

**C1 — SC permanent spinner when the SDK isn't connected at screen open**
Done! `useSelfCustodialRedemption` now starts a bounded wait when `enabled && !lnurlWithdraw`: after 10s with no SDK it surfaces a distinct `walletNotConnected` error, and if the SDK connects first the run-effect clears it and proceeds. No more silent eternal spinner.

**C2 — completion timeout reported as a flat network error**
Done — but the mechanism turned out different from catch-side detection, and the difference matters. The SDK's `completionTimeoutSecs` does **not** reject on expiry: it **resolves** with `LnurlWithdrawResponse { payment: Payment | undefined }` (empty/pending payment). The adapter was discarding that response and returning `Success`, so the timeout actually surfaced as *premature success*, not as a network error. Fixed SDK-first: branch on the resolved `response.payment?.status` — `Completed` → `Success`, otherwise (`undefined`/`Pending`) → `PaymentResultStatus.Pending`, which the screen renders as "payment may still be completing — check your balance before retrying." Thrown errors stay in the `catch` and are classified as before: a thrown transport timeout can predate dispatch, so it must remain `Failed`/retryable, never `Pending`.

## Important

**I1 — `run()` is a floating promise with no `.catch`**
Done! `run().catch(...)` (guarded by `cancelled`) sets a generic error so a synchronous throw or rejection can't strand the spinner.

**I2 — SC `lnServiceErrorReason` hardcoded to `""`**
Done! `PaymentError` carries an optional `reason`, the adapter threads it from the SDK error, and the SC hook sets `lnServiceErrorReason` from it. Per your note I gated `getSdkErrorReason` to the `LnurlError` tag specifically, so only a genuine service reason reaches the UI — never a raw internal SDK string for a non-LNURL failure.

**I3 — SC adapter classifies and swallows the raw error**
Done! The adapter calls `reportError("lnurlWithdraw", err)` (the existing Crashlytics helper used by the sibling bridges) before classifying.

**I4 — four custodial error paths log via bare `console.error`**
Folded into I5's follow-up ticket, per your D3 steer ("rest with I5's ticket") — swapping those lines would break the 1:1-with-merge-base property of the custodial side.

**I5 — pre-existing defects carried 1:1 from the old result screen**
Agreed, left for the follow-up ticket. The custodial `submitToCallback` rejections, `console.error(result.text())`, the raw `${err}` render, and the double `HomeAuthedDocument` refetch are untouched so the 1:1-preservation holds.

## Design

**D1 — ISP: mode-dependent return shape**
Dissolved by I2 — `lnServiceErrorReason` is now meaningful on both paths.

**D2 — boolean `enabled` mode dispatch**
Left for the follow-up ticket, as flagged.

**D3 — error-reporting inconsistency in the added code**
Done for the new SC adapter (I3, `reportError`); the custodial `console.error` cleanup rides with I5's ticket.

**D4 — `useAdapterRedemption` / `useMutationRedemption` name the transport, not the domain**
Done! Renamed to `useSelfCustodialRedemption` / `useCustodialRedemption` (and `SelfCustodialParams` / `CustodialParams`).

## Test coverage

**Tier 2 (made to fail first):**
- C1 — the `lnurlWithdraw`-undefined test now advances the bounded-wait timer and asserts the error surfaces, instead of asserting the eternal spinner.
- C2 — the adapter spec asserts `Pending` on a resolve with no settled payment (and on a still-pending payment), `Success` only on a completed payment, and that a thrown timeout is classified `Failed`, never `Pending`.
- Stale-error reset — error state set, then a new attempt + success → asserts `errorMessage` clears and `paid` flips.
- Stale resolution — a deferred first promise resolving after `amountSats` changes mid-flight no longer flips state; the adapter is re-invoked with the new amount.
- Rounding wire-through — a `1500` msats voucher asserts `minWithdrawableSatoshis.amount === 2` end-to-end.
- HOC contract — pins the `withMyLnUpdateSub` wrapping (wrapper testID), so an accidental unwrap fails.

**Tier 3 regressions:** custodial steady-state pending (`paid=false, errorMessage=""`); the SC `pending` state; I2 reason threading; plus direct unit tests for `getSdkErrorReason` (`LnurlError` → reason, everything else → `undefined`)

grimen (approved):

PR #3825, opened 2026-06-05
fix(android): credentials-manager crash on password save


esaugomez31:
## Summary

Fixes [blink-wip#769](https://github.com/blinkbitcoin/blink-wip/issues/769). On Android, the app crashed right after the password-manager Save dialog during self-custodial onboarding (`Choose your backup method` > `Password manager` > Save).

## Root cause

The crash is in the third-party `react-native-credentials-manager` (v0.8.1), not in our code. `signUpWithPassword` resolves its promise with a Kotlin `mapOf(...)` — a `java.util.LinkedHashMap`. React Native's promise-resolve converter `Arguments.fromJavaArgs` matches the value against an exact-class whitelist (`null`, `Boolean`, `Int`, `Double`, `Float`, `String`, `WritableNativeMap`, `WritableNativeArray`) and throws on anything else:

```
java.lang.RuntimeException: Cannot convert argument of type class java.util.LinkedHashMap
  at com.facebook.react.bridge.Arguments.fromJavaArgs(Arguments.kt:159)
  at com.facebook.react.bridge.PromiseImpl.resolve(PromiseImpl.kt:29)
  at com.credentialsmanager.CredentialsManagerModule$signUpWithPassword$1.invokeSuspend(CredentialsManagerModule.kt:76)
```

The credential save succeeds — only resolving the result crashes. The sibling methods (`signIn`, `createPasskey`) already build their response with `Arguments.createMap()`, so only `signUpWithPassword` was affected. Independent of device, Android OS, and Kotlin version.

## Fix

`patches/react-native-credentials-manager+0.8.1.patch` rebuilds the response as a `WritableMap` in both the `newarch` and `oldarch` modules:

```kotlin
val result = Arguments.createMap().apply {
  putString("type", "password")
  putString("username", username)
  putBoolean("success", true)
}
promise.resolve(result)
```

## Hardening (review follow-ups)

- **Pinned the dependency to exact `0.8.1`** (was `^0.8.1`) so resolution can't drift off the patched version.
- **`patch-package --error-on-fail`** in both `postinstall` and the `patch` script — a patch that stops applying now fails the install (and CI) instead of printing a silent warning and exiting 0.
- **Upstream issue filed and linked from the patch:** https://github.com/benjamineruvieru/react-native-credentials-manager/issues/23 — the exit criterion for deleting this patch is a release that includes the fix.
- **Regression test** `__tests__/patches/credentials-manager-patch.spec.ts` pins the patch version to the installed package version and asserts the `WritableMap` fix is present (not a raw `promise.resolve(mapOf`).
- **Jest mock aligned** to the native response shape (now includes `username`).

## Why a patch and not a version bump

- `0.8.1` is the latest published version; no upstream fix or relevant open PR.
- Per project policy, patch-package is a last resort reserved for when no upstream fix/bump exists. This qualifies.

## Test plan

- [x] Reproduced the original crash on an Android device (`adb logcat` stack trace above).
- [x] Rebuilt with the patch; the save flow no longer crashes and onboarding continues.
- [x] `postinstall` auto-applies the patch on `yarn install`, now with `--error-on-fail`.
- [x] 26/26 tests (patch regression spec + credential-backup), `tsc` + `eslint` clean.

## Notes

- No JS-side result-shape contract test: the bug manifested as a *rejection* in native, the consumer discards the resolved value, and such a test would break on harmless mock drift while catching nothing real.


esaugomez31:
* **#3825** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3825?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

## Critical

None.

---

## Important

- **I1 — The patch can silently stop applying on a dependency bump — the crash would return to production with a green pipeline.**
  `package.json:30,122` + `.github/workflows/test.yml`
  Two compounding gaps: the dependency is caret-ranged (`"^0.8.1"`) while the patch is pinned to `+0.8.1`, and `postinstall` runs `yarn patch-package` **without `--error-on-fail`** — patch-package's default on a failed hunk is a warning and exit 0. A future resolution to 0.8.2+ produces a non-fatal warning buried in install logs and CI stays green while the native crash ships again.
  **Fix (both one-liners):** `yarn patch-package --error-on-fail` in postinstall, and pin the dep exactly (`"0.8.1"`). This is the actual regression net for this PR — see Test coverage Tier 2/3.

- **I2 — File the upstream PR now and link it from the patch.**
  `patches/react-native-credentials-manager+0.8.1.patch`
  The PR description says "when we open the PR [upstream]" — do it as part of this change and reference the upstream PR/issue URL inside the patched comment block. That's the exit criterion for deleting the patch: when a release containing the fix ships, the link tells the future maintainer exactly what to check before bumping.

- **I3 — Stale jest mock diverges from the native contract.**
  `__mocks__/react-native-credentials-manager.js:1-3`
  Resolves `{ type: "password", success: true }` — missing `username`, inconsistent with both the native shape and the inline mock in `use-credential-backup.spec.ts`. Harmless today (the consumer discards the result), but if a future consumer reads `result.username` the mock masks the dependency. One-line alignment.

---

## Design

**SOLID violations**

- None.

- None.

- None.

**Code smells**

- **D1 — Caret range on a patched dependency.** `package.json:122` — `"^0.8.1"` under a version-pinned patch is the structural smell behind I1; a patched dep should be exact-pinned so resolution can't drift the patched version. Same fix as I1.

**Naming smells**

- None.

---

## Test coverage

- None.

### Tier 2 — tests that pass while a regression could ship invisibly

- **`.github/workflows/test.yml`** (→ I1) — CI runs `yarn install` (postinstall → `patch-package`, exit 0 on hunk failure) then `yarn test`. The entire pipeline passes while the patch silently fails to apply. `--error-on-fail` converts this from invisible to a hard install failure.

- **I1 (optional belt-and-suspenders, after the two one-liners)** — `__tests__/patches/credentials-manager-patch.spec.ts`: assert the patch filename version matches `require("react-native-credentials-manager/package.json").version`; assert the installed newarch `CredentialsManagerModule.kt` contains `Arguments.createMap()` and not `promise.resolve(mapOf` in the `signUpWithPassword` block — the only test that pins the *fix* rather than the version. Source-coupling is acceptable here; the artifact under test is a source patch.
- **Anti-recommendation** — do **not** add a JS result-shape contract test (`{type, username, success}`): the consumer discards the resolved value, the native bug manifested as a *rejection* (not a wrong shape), and such a spec would break on harmless mock drift while catching nothing real.

esaugomez31:
@grimen

## Important

**I1 — the patch can silently stop applying on a dependency bump**
Done! Both one-liners applied: the dependency is now pinned exactly (`"react-native-credentials-manager": "0.8.1"`), and `patch-package --error-on-fail` runs in both `postinstall` and the `patch` script — so a patch that stops applying fails the install (and CI) instead of a silent warning + exit 0.

**I2 — file the upstream PR/issue and link it from the patch**
Done! Filed upstream: https://github.com/benjamineruvieru/react-native-credentials-manager/issues/23. The patch's KDoc block (both `newarch` and `oldarch`) now links it with "remove this patch once a release includes the fix", so the exit criterion is recorded at the source.

**I3 — stale jest mock diverges from the native contract**
Done! `__mocks__/react-native-credentials-manager.js` `signUpWithPassword` now echoes `username` (`{ type: "password", username, success: true }`), matching the native response shape.

## Design

**D1 — caret range on a patched dependency**
Done! Pinned to exact `0.8.1` (same fix as I1), so resolution can't drift the patched version.

## Test coverage

**Tier 2 — pipeline passes while the patch silently fails to apply**
Done! `--error-on-fail` converts a non-applying patch from an invisible warning into a hard install failure, so CI can no longer go green with the crash shipped.

**Tier 3 — patch regression test**
Done! Added `__tests__/patches/credentials-manager-patch.spec.ts`: it asserts the patch filename version matches the installed package version, and that the installed `newarch` `CredentialsManagerModule.kt` contains `Arguments.createMap()` and not a raw `promise.resolve(mapOf` — pinning the fix, not just the version.

**On the JS result-shape contract test**
Left it out, agreed — it wouldn't add real coverage here.

grimen (approved):

PR #3829, opened 2026-06-09
fix(transactions): Dollar conversion missing from Bitcoin and All transaction lists


esaugomez31:
## Summary

Fixes `blinkbitcoin/blink-wip#766`: in the Transaction History screen, self-custodial **Dollar conversions were missing from the Bitcoin and All lists**. The screen derived and reordered the self-custodial list per filter, so conversion entries could be absent from the Bitcoin and All views.

The fix moves ordering ownership to the self-custodial data layer: every filter now renders from a single, globally time-ordered source, so conversions show up in every list, newest-first.

This PR also includes two **extra** self-custodial fixes found along the way (see below).

### Transaction ordering owned by the data layer (the ticket)

- `app/self-custodial/providers/wallet-snapshot.ts`: the snapshot exposes `allTransactions` (deduped, newest-first); added `dedupeTransactionsById` and `mergeOrderedTransactions`.
- `app/self-custodial/hooks/use-sdk-lifecycle.ts` and `app/self-custodial/providers/wallet.tsx`: thread `allTransactions` through the lifecycle and the provider value.
- `app/screens/transaction-history/transaction-history-screen.tsx`: consume the ordered `allTransactions`; the All / Bitcoin / Dollar filters slice the same source instead of reordering it.
- Tests cover the ordering across the All, Bitcoin and Dollar filters.

### Extra: mark-as-read by timestamp (highlight recency)

While verifying the lists, self-custodial transactions were found to never be marked as read — the unseen highlight never cleared (most visibly for Dollar), while custodial worked fine. The highlight decided recency with a lexicographic `txId > lastSeenId` comparison, which holds only for custodial ids (time-monotonic); self-custodial ids are Breez payment hashes with no time order, so the comparison was effectively random.

- For self-custodial, recency is compared by `createdAt` (the numeric timestamp) instead of the id.
- The two recency models are split per domain, each a pure module: `app/self-custodial/mappers/transaction-highlight.ts` (`shouldHighlightByTimestamp`, `resolveHighlightBaseline`) and `app/custodial/mappers/transaction-highlight.ts` (`shouldHighlightById`). The screen is a thin dispatcher that picks one by account type.
- The custodial recency logic was moved verbatim (behaviour-identical). Both mappers are covered by unit tests at 100% statement/branch/function/line coverage.

### Extra: self-custodial list shows the skeleton while navigating in

On self-custodial, opening the Transaction History (swipe up on Home) jumped straight to the list with no skeleton and felt laggy: the list rendered synchronously during the navigation animation and janked it. Custodial doesn't hit this because its data is async, so it naturally shows the skeleton during the transition (`!transactions`). Self-custodial had no equivalent hold.

- New generic, reusable hook `app/hooks/use-has-transitioned.ts` (`useHasTransitioned`) tracks the screen's `transitionEnd`. No domain logic.
- `app/screens/transaction-history/transaction-history-screen.tsx` holds the self-custodial skeleton via `selfCustodialSettling = isSelfCustodial && !hasTransitioned`, mirroring custodial's `!transactions`, so the list (the heavy render) lands after the animation. Result: self-custodial now shows the skeleton and animates in like custodial, with no flicker.
- The unseen highlight is unaffected: row highlighting uses the baseline frozen at mount, so a newly received transaction still highlights once the list appears.
- Hook covered by unit tests at 100% coverage; screen tests cover skeleton-while-settling and list-after-transition.

## Notes

- Custodial behaviour is unchanged throughout: the custodial list still flows through the Apollo `useTransactionListForDefaultAccountQuery` path, the custodial highlight model is a verbatim extraction (the only structural change collapsed an unreachable nested-ternary branch — `settlementCurrency` is a closed `{Btc, Usd}` enum — into its equivalent binary form), and every self-custodial gate clause is guarded by `isSelfCustodial`.
- Domain is conveyed by folder (`app/custodial/` vs `app/self-custodial/`); no module or function name carries a domain prefix, and no function mixes the two account types.


esaugomez31:
* **#3829** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3829?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3829?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (approved):

PR #3835, opened 2026-06-09
fix(mobile): non-custodial bug report & UI improvements


esaugomez31:
## Summary

Implements the non-custodial fixes and UI improvements from blinkbitcoin/blink-wip#781.

## Changes

**1. "Anonymous user" label** — `common.anonymousUser`: EN "Anonymous user" → "Anon user", ES "Usuario anónimo" → "Usuario Anon". The other locales keep their natural native term (a per-language review found "Anon" is an English clipping, not natural elsewhere; non-Latin scripts would mix alphabets).

**2. Cloud-backup password — relaxed validation** — the encryption password now enforces only a 12-character minimum. Removed the character-class complexity rule, the common-password list, and the unused strength scoring; the one-line validator was inlined into its sole consumer.

**4. "Amount below 800 SAT" warning — conditional + fiat** — the self-custodial Dollar-receive auto-convert warning now appears only after an amount is entered and it falls below the pool minimum (it no longer shows by default, and it clears when the amount is removed). The copy now includes the fiat equivalent (`{minSats} SAT / {minFiat}`).

**5. "Backup phrase" header → recovery framing** — ES "Frase de respaldo" → "Frase de recuperación". The English-loanword "backup" headers (da/de/it/ja/nl/pt/ro) were replaced with each language's natural recovery-phrase term; already-native locales kept theirs.

## Pending

**3. Notifications not working** — pending due to backend limitations.

## Notes

- **Custodial untouched:** every change is self-custodial-only or gated so the custodial render output stays byte-identical.
- i18n edits preserve each file's indentation and use real diacritics; types regenerated via `yarn update-translations`.
- Per-item tests added/updated; full typecheck, lint, and the relevant test suites are green.

Part of blinkbitcoin/blink-wip#781


esaugomez31:
* **#3835** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3835?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3835?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

app/screens/self-custodial/onboarding/hooks/use-cloud-backup-form.ts
```diff
@@ -48,22 +49,17 @@ export const useCloudBackupForm = () => {
     if (confirmPassword.length === 0) setConfirmPasswordTouched(false)
   }, [confirmPassword])
 
-  const passwordValidation = useMemo(() => {
-    if (!isEncrypted || password.length === 0) return null
-    return validatePassword(password)
-  }, [isEncrypted, password])
+  const isPasswordValid = useMemo(
+    () => password.length >= MIN_PASSWORD_LENGTH,
+    [password],
```
grimen: Use password validator interface that was implemented a while back, i.e. password validation using custom JS logic is almost always the wrong path.

app/utils/validators/password.ts
grimen, line 1: NO, this is an incredibly bad idea - please revert this removal and do this property. Password validation should be done with this utility with at most a custom regex.

grimen (changes requested):
Password validation solution is a step backwards - we don't want to go that route.

app/screens/self-custodial/onboarding/hooks/use-cloud-backup-form.ts
esaugomez31: Done!
dropped the inline check. The hook uses validatePassword again, no custom JS in the component.

app/utils/validators/password.ts
esaugomez31: Done here too!

grimen (approved):

PR #3847, opened 2026-06-12
fix(stablesats): region restriction is enforced inconsistently and can be bypassed


esaugomez31:
## Summary

Hardens the Stablesats/Dollar region restriction so it is enforced consistently and can no longer be bypassed, and recovers Dollar funds that get stranded in a restricted account. Also includes an unrelated email-registration navigation fix found while testing.

Part of blinkbitcoin/blink-wip#795

## Changes

**1. Region restriction is now sticky (closes the bypasses)** — the restriction was recomputed from live location signals on every render, so it vanished whenever the signal did. It is now persisted per account in `persistentState`: once an account is seen in a blocked country it stays restricted. This closes the two reported bypasses, removing the phone number and switching a non-custodial account to a custodial one in HK.

**2. Restriction also enforced by IP (phone OR IP)** — detection checks the phone-number country first; when the phone country does not restrict, it falls back to the device IP country. Either signal being in a blocked country enforces the restriction, so a user with an allowed phone number who is physically in a restricted region is now caught. The IP lookup runs only when the phone country does not already restrict.

**3. Forced-conversion auto-prompt for stranded Dollar funds** — when a restricted account still holds a Dollar balance, the convert-to-Bitcoin modal opens automatically on app start and when switching into that account, so the balance can be moved to Bitcoin. Previously no prompt appeared and the funds looked lost. The modal is non-dismissable (no close icon, backdrop tap disabled): the only way out is approving the conversion, so the funds cannot be left stranded.

**4. Email-add returns to Settings cleanly (unrelated nav fix)** — adding an email from Settings left the "Success" screen in the navigation stack (React Navigation v7's `navigate` no longer pops back to an existing screen), so pressing back from Settings re-opened Success. It now uses `popTo("settings")`. The onboarding email flow uses a separate branch and is unchanged.

## Pending

**Currency display flips USD/HKD when switching accounts** — display currency is per-account (custodial from the backend, self-custodial defaulting to USD), and the cache-and-network price refetch adds a transient flicker on switch. Not addressed here; pending a decision on the intended behavior.

## Notes

- The restriction is custodial-only; self-custodial accounts are exempt (gated at the query level).
- Detection precedence: persisted flag, then phone country, then IP country, with `SV` as the safe default.

### DIFF
--- a/.storybook/views/story-screen.tsx
+++ b/.storybook/views/story-screen.tsx
@@ -5,7 +5,7 @@ const PersistentStateWrapper: React.FC<React.PropsWithChildren> = ({ children })
   <PersistentStateContext.Provider
     value={{
       persistentState: {
-        schemaVersion: 13,
+        schemaVersion: 14,
         galoyInstance: {
           id: "Main",
         },
--- a/__tests__/components/stablesats-restriction-modal/stablesats-restriction-modal.spec.tsx
+++ b/__tests__/components/stablesats-restriction-modal/stablesats-restriction-modal.spec.tsx
@@ -77,41 +77,6 @@ describe("StablesatsRestrictionModal", () => {
     expect(mockNavigate).not.toHaveBeenCalled()
   })
 
-  it("fires onDismiss when Close is pressed", () => {
-    const onDismiss = jest.fn()
-    const { getByText } = render(
-      wrap(
-        <StablesatsRestrictionModal
-          isVisible={true}
-          toggleModal={jest.fn()}
-          onDismiss={onDismiss}
-        />,
-      ),
-    )
-
-    fireEvent.press(getByText("Close"))
-
-    expect(onDismiss).toHaveBeenCalledTimes(1)
-  })
-
-  it("does not fire onDismiss when Create new is pressed (navigates instead)", () => {
-    const onDismiss = jest.fn()
-    const { getByText } = render(
-      wrap(
-        <StablesatsRestrictionModal
-          isVisible={true}
-          toggleModal={jest.fn()}
-          onDismiss={onDismiss}
-        />,
-      ),
-    )
-
-    fireEvent.press(getByText("Create new"))
-
-    expect(onDismiss).not.toHaveBeenCalled()
-    expect(mockNavigate).toHaveBeenCalledWith("getStarted")
-  })
-
   it("renders nothing when isVisible is false", () => {
     const { queryByText } = render(
       wrap(<StablesatsRestrictionModal isVisible={false} toggleModal={jest.fn()} />),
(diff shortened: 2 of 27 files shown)


esaugomez31:
* **#3847** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

### Test coverage

**1. `useStablesatsForcedConversion` invariant is untested.** `home.spec` covers auto-open-with-balance and no-open-without, but nothing asserts the modal stays closed after conversion zeroes the balance, nor the close-only-on-success behavior the whole forced flow rests on.

**2. No test pins the sticky/no-reset behavior.** A "stays restricted after the country leaves the blocklist" test would lock in the decided stickiness rather than leaving it implicit.

### Code smells

**3. `useStablesatsForcedConversion` mirrors derivable state.** Given the modal can only close on success, visibility is effectively `isConvertModalVisible = hasConvertibleBalance`. The `useState` + effect works only because conversion zeroes the balance — a fragile, undocumented coupling. Either derive it directly or comment why the mirror is needed.

**4. `stablesatsRestrictedByAccountId` naming overpromises per-account granularity** — for custodial accounts every login resolves to the single `DefaultAccountId.Custodial` slot, so the map holds one shared custodial entry, not one per account.

**5. `fetchCountryFromIp` orchestration is duplicated** between `useDeviceLocation`'s own effect and the standalone `useIpCountryCode`, each with separate error logging.

esaugomez31:
@grimen

Test coverage
1. useStablesatsForcedConversion invariant is untested. Done. Added a hook test that the modal stays closed once a successful conversion zeroes the balance, and a modal test asserting close-only-on-success (the conversion's onSuccess is the only thing wired to close it, alongside the existing non-dismissable + no-close-icon assertions).

2. No test pins the sticky/no-reset behavior. Done. Added a test that a flagged account stays restricted after its country is removed from the remote blocklist (flagged + HK + empty blocklist → still restricted).

Code smells
3. useStablesatsForcedConversion mirrors derivable state. Kept the mirror (not derived) and documented why: deriving isConvertModalVisible = hasConvertibleBalance would keep the modal open until the balance query refetches to zero, whereas the state lets onSuccess close it immediately. One-line comment added.

4. stablesatsRestrictedByAccountId naming overpromises per-account granularity. Done. Collapsed it to a single stablesatsRestrictedCustodial: boolean — since the restriction only ever applies to custodial (every login resolves to the one DefaultAccountId.Custodial slot), the per-account map was always a single entry. Schema 14 is new in this PR so no extra migration was needed.

5. fetchCountryFromIp orchestration is duplicated. Done. Extracted a shared resolveIpCountryCode (fetch + error logging) now used by both useDeviceLocation and useIpCountryCode, so the IP-resolution and logging live in one place.

grimen (approved):

PR #3849, opened 2026-06-12
fix(self-custodial): Not Authorized on send to a Blink Lightning Address, plus account-resolution fixes

### LINKED ISSUE #3845: fix(send): self-custodial send to Blink Lightning Address returns Not Authorized
## Summary

On Blink mobile 2.4.54, sending from a self-custodial account to a custodial Blink Lightning Address fails with a `Not Authorized` error. Reported on both iOS and Android.

## Affected versions/platforms

- Blink mobile 2.4.54
- iOS
- Android
- Source account: self-custodial / non-custodial account
- Destination: custodial Blink Lightning Address, example `pretyflaco@blink.sv`

## Reproduction steps

1. Switch to a self-custodial account.
2. Go to Send.
3. Enter `pretyflaco` as the destination.
4. Observe `Not Authorized` error.

## Expected behavior

The app should resolve the custodial Blink Lightning Address and continue the payment flow, or show a clear actionable error if this route is intentionally unsupported.

## Actual behavior

The send flow returns `Not Authorized`.

## Notes

Reported by Kemal in Slack `#1-tech-bot` on 2026-06-11.
Slack thread: https://blinkbitcoin.slack.com/archives/C0AEYAALSCQ/p1781189559679239

## Update 2026-06-15

Kemal corrected the repro: step 3 should use `pretyflaco` as the destination, not `pretyflaco@blink.sv`. He also added a screen recording in https://github.com/blinkbitcoin/blink-mobile/issues/3845#issuecomment-4706844128.

esaugomez31:
## Summary

Fixes the reported "Not Authorized" when a self-custodial account sends to a Blink username, plus an account-switcher bug found alongside it.

Closes #3845

## Changes

**1. Self-custodial send to a username returned "Not Authorized"** — a bare Blink username is classified by the parser as a custodial intraledger handle (before any domain/LNURL logic and independent of the configured LN-address domains). The send then ran the custodial `intraLedgerPaymentSend` GraphQL mutation, which a self-custodial account has no JWT for, so it failed with "Not Authorized" (and showed Fee 0 SAT). For a self-custodial account, a username is now re-resolved as its Lightning Address (`username@<lnAddressHostname>`) and paid through the existing LNURL path. The pure parser stays account-agnostic and the routing lives in the self-custodial layer; the custodial flow is untouched.

**2. Custodial account missing from the account switcher** — the switcher loaded stored custodial profiles only when a custodial token was present, so a self-custodial-active user (no token) saw only their self-custodial account. Stored profiles are read from the keystore and do not need a token (it only marks which one is selected); the switcher now lists them regardless, and only persists a new profile when a token exists.

## Notes

- Custodial behavior is unchanged: the username re-routing runs only for a self-custodial account (when the Spark SDK is connected); the custodial path returns the parser output unchanged.
- The username re-resolution reuses the existing LNURL send path and the `getLightningAddress` helper, so no new destination or payment-detail shapes were introduced.

### DIFF
--- a/__tests__/payment-destination/resolve-destination.integration.spec.ts
+++ b/__tests__/payment-destination/resolve-destination.integration.spec.ts
@@ -1,0 +1,86 @@
+import { getParams } from "js-lnurl"
+import { requestPayServiceParams } from "lnurl-pay"
+
+import { Network } from "@app/graphql/generated"
+import { resolveDestination } from "@app/screens/send-bitcoin-screen/payment-destination/resolve-destination"
+import { PaymentType } from "@blinkbitcoin/blink-client"
+
+// Real parser + LNURL resolution; only the network boundaries are mocked.
+jest.mock("js-lnurl", () => ({
+  ...jest.requireActual("js-lnurl"),
+  getParams: jest.fn(),
+}))
+jest.mock("lnurl-pay", () => ({
+  ...jest.requireActual("lnurl-pay"),
+  requestPayServiceParams: jest.fn(),
+}))
+
+const mockParseSparkAddress = jest.fn()
+jest.mock("@app/self-custodial/bridge", () => ({
+  parseSparkAddress: (...args: unknown[]) => mockParseSparkAddress(...args),
+}))
+
+const mockWrapDestination = jest.fn()
+jest.mock("@app/self-custodial/payment-details/wrap-destination", () => ({
+  wrapDestination: (...args: unknown[]) => mockWrapDestination(...args),
+}))
+
+// Payment-detail builders are lazy (createPaymentDetail closures); stub them so the
+// real parser graph loads without pulling the heavy detail factories.
+jest.mock("@app/screens/send-bitcoin-screen/payment-details", () => ({
+  createIntraledgerPaymentDetails: jest.fn(),
+  createLnurlPaymentDetails: jest.fn(),
+}))
+
+const mockGetParams = getParams as jest.MockedFunction<typeof getParams>
+const mockRequestPayServiceParams = requestPayServiceParams as jest.MockedFunction<
+  typeof requestPayServiceParams
+>
+
+describe("resolveDestination (integration: real parser)", () => {
+  const fakeSdk = { id: "sdk" } as never
+  const lnAddressHostname = "blink.sv"
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockParseSparkAddress.mockResolvedValue(null)
+    mockWrapDestination.mockImplementation((result) => result)
+    mockGetParams.mockResolvedValue({} as never)
+    mockRequestPayServiceParams.mockResolvedValue({
+      callback: "https://blink.sv/callback",
+      fixed: false,
+      min: 1,
+      max: 1000000,
+      domain: lnAddressHostname,
+      metadata: [["text/plain", "pay esaudeveloper"]],
+      metadataHash: "hash",
+      identifier: `esaudeveloper@${lnAddressHostname}`,
+      description: "",
+      image: "",
+      commentAllowed: 0,
+      rawData: {},
+    } as never)
+  })
+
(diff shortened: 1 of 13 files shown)


esaugomez31:
* **#3849** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3849?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3849?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):
Non-blocking notes on test coverage and structure before merge.

### Test coverage

**1. The fix's core assumption isn't exercised end-to-end.** Every spec mocks `parseDestination`/`resolveUsername`, so the one thing this fix depends on — that `parsePaymentDestination("handle@<lnAddressHostname>")` resolves to `Lnurl` and not back to `Intraledger` — has no coverage. It's correct today only because self-custodial passes `lnAddressDomains: []`; if that list ever included the host, the address would route back through the intraledger path and `wrapDestination` would hit the custodial mutation again ("Not Authorized" returns), with all mocked tests still green. Could we add one integration test using the real parser (LNURL HTTP mocked) that asserts a self-custodial send to a bare username ends up as a `Lnurl` destination? That pins the assumption the whole fix rests on.

**2. No regression test for the preserved custodial self-save.** This PR moved the save behind `&& currentToken`. The new test asserts `saveProfile` is *not* called when there's no token, but nothing asserts it *is* still called when a token is present and stored profiles are empty — the branch the PR deliberately kept.

**3. The no-token test doesn't assert the meaningful state.** `fetchProfiles` recomputes `selected = profile.token === ""` → `false`, but the new test only checks the row renders, not that no profile is selected. It would pass even if selection were broken, which is the actual semantic this change enables. Worth an explicit `selected: false` assertion.

### Code smells

**4. `resolveUsername` lives in `wrap-destination.ts` but doesn't wrap anything.** It's a routing/resolution concern sitting in the wrapping module — a reader looking for the username→LN-address rerouting won't expect to find it here. Consider moving it next to `resolveDestination`.

**5. `intraledgerHandle` duplicates `wrapDestination`'s shape guards.** Both functions in this file now independently destructure a `ParseDestinationResult` (`valid` / `Send` / `paymentType`) — two places that have to stay in sync with the discriminated-union shape, and will drift silently if it changes.

esaugomez31:
@grimen

Test coverage
1. The fix's core assumption isn't exercised end-to-end. Done. Added an integration test (resolve-destination.integration.spec.ts) that runs the real parser with the LNURL HTTP boundary mocked: a self-custodial send to a bare username, with lnAddressDomains: [], resolves to a Lnurl destination. If the host were ever in the domains list, it would route back to Intraledger and fail the test — so the assumption is pinned.

2. No regression test for the preserved custodial self-save. Done. Added a test asserting saveProfile is still called with the current token when a token is present and no profiles are stored yet — the branch the PR deliberately kept.

3. The no-token test doesn't assert the meaningful state. Done — via a direct fetchProfiles unit test asserting selected: false when there's no current token (and selected: true only for the matching token). I put it on fetchProfiles rather than the component because the row's "current" indicator is selected && activeAccount.type === Custodial, so the active-account dependency makes a UI-level selection assertion unreliable; testing fetchProfiles pins the exact semantic instead.

Code smells
4. resolveUsername lives in wrap-destination.ts but doesn't wrap anything. Done. Moved it (and its intraledgerHandle helper) to resolve-username.ts, next to resolveDestination, with the spec moved alongside.

5. intraledgerHandle duplicates wrapDestination's shape guards. Done. Extracted a shared isSendDestination(result): result is PaymentDestination type predicate in index.types; both wrapDestination and intraledgerHandle now narrow through it, so the discriminated-union shape lives in one place.

grimen (approved):

PR #3853, opened 2026-06-15
fix(ios): remove iOS 26 Liquid Glass overlay from header buttons

### LINKED ISSUE #4072: fix(ios): iOS 26 Liquid Glass capsule misaligned on custom header buttons
## Description

On iOS 26, UIKit wraps custom navigation-bar button items in a Liquid Glass "shared background". Our custom header buttons (rendered through native-stack's `headerLeft`/`headerRight`) get this capsule automatically, and it renders misaligned with the glyphs — the icon sits offset from where the system places the glass, so buttons look broken.

## Affected controls

- Close buttons (`headerCloseControl`, 8 screens)
- Settings-screen notification bell
- Card details / card dashboard settings buttons
- Receive-screen NFC button
- Send-screen scan button
- Restore-phrase paste button
- Webview back button

Native back buttons are unaffected (standard system treatment).

## Expected

Custom header buttons render without the misplaced glass capsule, i.e. opted out via `UIBarButtonItem.hidesSharedBackground` (reachable through native-stack's `unstable_headerLeftItems`/`unstable_headerRightItems`, iOS-only; Android keeps the plain `headerLeft`/`headerRight` fallback).

designsats:
> [!NOTE]
> Rebased 2026-08-07: the original Liquid Glass opt-out landed on `main` through #3890 (this branch was stacked underneath it), so the two original commits became empty and were dropped. What remains here are the follow-ups from the review.

**Original scope (now on `main`):** iOS 26 wraps custom navigation-bar button items in a Liquid Glass "shared background", rendering a capsule misaligned with the glyphs. Every custom header button opts out via `UIBarButtonItem.hidesSharedBackground`, reached through native-stack's `unstable_header{Left,Right}Items` and centralized in the `header-no-glass` helper. Android falls back to `headerLeft`/`headerRight`, so each site emits both.

**This PR (review follow-ups):**
- The three card-onboarding screens that suppressed the back button with an empty custom `headerLeft` now use `headerBackVisible: false` — no custom item, so no capsule can be drawn around it. The NFC button's hidden state clears the slot with `noHeaderRight` instead of installing an empty no-glass item, so no empty custom header items remain anywhere.
- The dead `headerCloseControl` factory export is gone; `headerCloseControlOptions` is the only path, so a future screen can't silently reintroduce the capsule.
- `header-no-glass` is re-exported through an `index.ts` like its sibling components; the deep imports are gone.
- A new spec pins the no-glass contract: both the plain Android render and the `unstable_` item with `hidesSharedBackground: true` must be emitted, and `noHeaderRight` must clear both keys.


designsats:
@blink-claw-bot Review

grimen (commented):
## Findings

**1. Empty `headerLeft: () => <></>` on the card-onboarding screens may still draw a glass capsule.**
The equivalent empty branch in `nfc-header-button.tsx` was converted to `headerRightNoGlass(() => <></>)`, but `cardOnboardingPersonalInfoScreen`, `cardOnboardingProcessingScreen`, and `cardOnboardingPreapprovedScreen` keep a plain empty `headerLeft`, so the two empty-item cases are treated differently. For the implementer to confirm on device which behavior is right.
*Example:* if iOS 26 wraps even an empty custom item in a shared background, these three screens show a stray empty capsule top-left; if it doesn't, the NFC empty-branch conversion was unneeded.
*Proposal:* one option is `headerBackVisible: false` for these screens, which suppresses the back button natively without a custom item at all.

**2. `headerCloseControl` is now dead code and a footgun.**
After this PR no caller of the old `headerCloseControl` factory remains, but it is still exported from the component and its `index.ts`. A future screen that picks the old export gets the glass capsule back silently.
*Example:* a new modal screen written next month uses `headerRight: headerCloseControl()` (the pattern all over git history) and ships with the misaligned capsule again.
*Proposal:* removing the old export (or aliasing it to the options-based variant) would leave the opted-out path as the only path.

## Test Coverage Gaps

- Nothing asserts the no-glass contract. `headerRightNoGlass`/`headerLeftNoGlass` are pure functions, so a small spec could check that both `headerRight` and `unstable_headerRightItems` are emitted and that the custom item carries `hidesSharedBackground: true`. Today the helper could be reverted to plain `headerRight` and no test would fail — the existing `nfc-header-button.spec.tsx` only asserts `headerRight: expect.any(Function)`.

## Code Smells

- The Android back-button reposition (`BACK_BUTTON_INSET_CORRECTION = -10` in header-back-control, plus the `marginLeft: 10` removal in webview) is a native-stack-migration layout fix riding in a `fix(ios)` PR, and the PR body doesn't mention it. Both changes are well-commented in code, but Android testers won't know from the PR that back-button positions moved.
- `header-no-glass` has no `index.ts`, so every consumer deep-imports `@app/components/header-no-glass/header-no-glass`, unlike sibling components (`header-close-control` re-exports via `index.ts`).

grimen (changes requested):
## Findings

**1. Empty `headerLeft: () => <></>` on the card-onboarding screens may still draw a glass capsule (blocking).**
The equivalent empty branch in `nfc-header-button.tsx` was converted to `headerRightNoGlass(() => <></>)`, but `cardOnboardingPersonalInfoScreen`, `cardOnboardingProcessingScreen`, and `cardOnboardingPreapprovedScreen` keep a plain empty `headerLeft`, so the two empty-item cases are treated differently. For the implementer to confirm on device which behavior is right.
*Example:* if iOS 26 wraps even an empty custom item in a shared background, these three screens show a stray empty capsule top-left; if it doesn't, the NFC empty-branch conversion was unneeded.
*Proposal:* one option is `headerBackVisible: false` for these screens, which suppresses the back button natively without a custom item at all.

**2. `headerCloseControl` is now dead code and a footgun.**
After this PR no caller of the old `headerCloseControl` factory remains, but it is still exported from the component and its `index.ts`. A future screen that picks the old export gets the glass capsule back silently.
*Example:* a new modal screen written next month uses `headerRight: headerCloseControl()` (the pattern all over git history) and ships with the misaligned capsule again.
*Proposal:* removing the old export (or aliasing it to the options-based variant) would leave the opted-out path as the only path.

## Test Coverage Gaps

- (blocking) Nothing asserts the no-glass contract. `headerRightNoGlass`/`headerLeftNoGlass` are pure functions, so a small spec could check that both `headerRight` and `unstable_headerRightItems` are emitted and that the custom item carries `hidesSharedBackground: true`. Today the helper could be reverted to plain `headerRight` and no test would fail — the existing `nfc-header-button.spec.tsx` only asserts `headerRight: expect.any(Function)`.

## Code Smells

- The Android back-button reposition (`BACK_BUTTON_INSET_CORRECTION = -10` in header-back-control, plus the `marginLeft: 10` removal in webview) is a native-stack-migration layout fix riding in a `fix(ios)` PR, and the PR body doesn't mention it. Both changes are well-commented in code, but Android testers won't know from the PR that back-button positions moved.
- `header-no-glass` has no `index.ts`, so every consumer deep-imports `@app/components/header-no-glass/header-no-glass`, unlike sibling components (`header-close-control` re-exports via `index.ts`).

grimen:
Rebased onto `main` and pushed the review follow-ups. Note on the rebase: the original two commits landed on `main` via #3890 (this branch was stacked underneath it), so they dropped as empty — the PR diff is now just the review fixes.

**Finding 1 (blocking) — empty `headerLeft` on the card-onboarding screens:** switched `cardOnboardingPersonalInfoScreen`, `cardOnboardingProcessingScreen`, and `cardOnboardingPreapprovedScreen` to `headerBackVisible: false` (6aa25f8ae), as proposed. This suppresses the back button natively, so there is no custom item for iOS 26 to wrap. The repo already relies on this pattern (`contact-support-screen.spec.tsx` even documents that a null `headerLeft` alone doesn't reliably suppress the arrow on native-stack). The NFC empty branch went the same direction: its hidden state now uses `noHeaderRight` to clear the slot instead of installing an empty custom item — no empty custom header items remain anywhere.

**Finding 2 — dead `headerCloseControl` export:** removed the factory and its `index.ts` re-export (573ea81d3); `headerCloseControlOptions` is now the only path, with a comment stating why.

**Test coverage gap (blocking):** added `__tests__/components/header-no-glass/header-no-glass.spec.tsx` (b77d37dee) asserting both halves of the contract — the plain Android render and the `unstable_` custom item with `hidesSharedBackground: true` — plus that `noHeaderRight` clears both keys explicitly (keys present and `undefined`, since `setOptions` spreading only clears keys that exist). Reverting the helper to plain `headerRight` now fails the suite. The NFC spec also asserts the `unstable_` item and the `noHeaderRight` hidden state, and now renders the header button and fires the press. `header-no-glass.tsx` and `index.ts` sit at 100% coverage.

**Smells:** `header-no-glass` now has an `index.ts` and all ten deep imports are rewritten (573ea81d3). The Android back-button reposition note is moot after the rebase — that code lives on `main` — but the PR body was rewritten to describe the actual remaining diff.

Local verification: targeted suites + everything touching the changed screens (93 suites, 1289 tests) pass, `tsc -p .` and eslint clean.

grimen:
Simulator proof of the fix — same Maestro flow on an iOS 26.5 simulator both times (Get Started → Log in / Restore → Non-custodial → Manual backup), landing on the restore-phrase screen with its custom `Paste` header button.

**before** is the pre-fix rendering path (the `hidesSharedBackground` opt-out stripped from `header-no-glass`, i.e. a plain custom `headerRight`): iOS 26 wraps the button in the Liquid Glass capsule, misaligned with the glyph. **after** is this branch as-is: the custom item opts out, while the native Back button keeps its standard system glass treatment.

| before | after |
|---|---|
|  |  |

<sub>Demos live on branch `assets/pr-3853-demo` — deleting it breaks them.</sub>

grimen (approved):

PR #3860, opened 2026-06-15
fix(self-custodial): home intermittently shows a zero balance for a funded account

### LINKED ISSUE #3852: Android 2.4.54 intermittently shows zero balance for non-custodial account
## Summary

Android 2.4.54 intermittently shows a zero balance on the home screen for a non-custodial account that still has funds.

Reported by Kemal in Slack `#1-tech-bot` on 2026-06-15 with two screenshots from the same non-custodial account about 20 minutes apart, with no transactions in between.

Tagging @esaugomez31 and @grimen to investigate.

## Observed behavior

Sometimes the home screen shows:

- Main balance: `$0.00`
- Bitcoin balance: `0 SAT` / `~ $0.00`
- Dollar balance: `$0.00`

A later/other view of the same account shows the expected balance:

- Main balance: `$0.12`
- Bitcoin balance: `187 SAT` / `~ $0.12`
- Dollar balance: `$0.00`

Kemal also tested invoice generation while the UI showed `$0.00`, and invoice generation succeeded. That suggests the account/wallet state may still be available, while the displayed balance is intermittently stale, missing, or incorrectly defaulted to zero.

## Environment

- App: Blink Mobile
- Platform: Android
- Version: 2.4.54
- Account type: Non-custodial

## Expected behavior

The home screen should consistently show the available Bitcoin balance for the account, or clearly show a loading/error state if balance data cannot be loaded. It should not silently display `$0.00` / `0 SAT` for a funded account.

## Notes / initial investigation questions

- Check whether the non-custodial balance query/cache can transiently resolve as empty/zero on app resume or account switch.
- Check whether a loading/error/null state is being coerced to zero in the home balance UI.
- Check whether invoice generation uses a different source of wallet state than the home balance display.
- Screenshots are attached in the original Slack thread:
  - working state at 17:05: `$0.12`, `187 SAT`
  - incorrect zero state at 16:44: `$0.00`, `0 SAT`

esaugomez31:
## Summary

On slow connections, a funded self-custodial account could show a $0.00 balance on the home screen. The home only treated the account as loading while the SDK status said so; when the wallet snapshot did not respond within the 5s timeout, the status flipped to error/offline with the wallets not loaded yet but the price already cached, so the home rendered a real $0.00 instead of a loading state.

Closes #3852

## Changes

**1. Show a loading state instead of a false zero** — the home now keys the self-custodial loading state off whether the wallets have actually loaded (empty wallets means the balance has not loaded yet, not that it is zero) instead of the SDK status. A slow or failed snapshot now shows the balance skeleton, and once the wallets are loaded it keeps the last known balance instead of dropping to $0.00.

**2. More time to load on slow networks** — the wallet-snapshot fetch now has its own 10s timeout instead of sharing the 5s status budget, so congested connections have more time to load the balance before falling back.

## Notes

- Custodial behavior is unchanged: the loading-state change is gated to self-custodial accounts; the custodial path keeps its existing query-based loading logic.
- The snapshot timeout is a separate constant from the connectivity-probe budget, so raising it to 10s does not slow down offline detection (the probes stay at 5s).

### DIFF
--- a/__tests__/screens/home.spec.tsx
+++ b/__tests__/screens/home.spec.tsx
@@ -825,4 +825,63 @@ describe("HomeScreen", () => {
       expect(lastIsVisible()).toBe(false)
     })
   })
+
+  describe("self-custodial balance loading (#3852)", () => {
+    afterEach(() => {
+      mockActiveWalletOverride = null
+    })
+
+    it("shows the loading state instead of $0.00 when the self-custodial balance failed to load", async () => {
+      mockActiveWalletOverride = {
+        wallets: [],
+        status: "error",
+        accountType: "self-custodial",
+        isReady: false,
+        isSelfCustodial: true,
+        needsBackendAuth: false,
+      }
+
+      const { queryByTestId } = render(
+        <ContextForScreen>
+          <HomeScreen />
+        </ContextForScreen>,
+      )
+      await flushEffects()
+
+      expect(queryByTestId("balance-value")).toBeNull()
+    })
+
+    it("keeps showing the balance when a later refresh goes offline and the wallets are retained", async () => {
+      mockActiveWalletOverride = {
+        wallets: [
+          {
+            id: "btc-1",
+            walletCurrency: "BTC",
+            balance: { amount: 5000, currency: "BTC", currencyCode: "BTC" },
+            transactions: [],
+          },
+          {
+            id: "usd-1",
+            walletCurrency: "USD",
+            balance: { amount: 0, currency: "USD", currencyCode: "USD" },
+            transactions: [],
+          },
+        ],
+        status: "offline",
+        accountType: "self-custodial",
+        isReady: false,
+        isSelfCustodial: true,
+        needsBackendAuth: false,
+      }
+
+      const { getByTestId } = render(
+        <ContextForScreen>
+          <HomeScreen />
+        </ContextForScreen>,
+      )
+      await flushEffects()
+
+      expect(getByTestId("balance-value")).toBeTruthy()
+    })
+  })
 })
--- a/__tests__/screens/home.spec.tsx
+++ b/__tests__/screens/home.spec.tsx
@@ -883,5 +883,45 @@ describe("HomeScreen", () => {
 
       expect(getByTestId("balance-value")).toBeTruthy()
     })
+
+    it("shows the loading state during an account switch, before the new wallets load", async () => {
+      mockActiveWalletOverride = {
+        wallets: [],
+        status: "loading",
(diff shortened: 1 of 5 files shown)


esaugomez31:
* **#3860** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3860?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3860?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):
Clean, correctly-scoped fix. Notes before merge.

### Test coverage

**1. The actual point-#2 behavior is untested.** The change that matters — a slow snapshot resolving *between 5s and 10s* now succeeds where it previously aborted — has no positive test. The wallet.spec test only exercises the abort at `WALLET_SNAPSHOT_TIMEOUT_MS + 1`. A "resolves at ~6s → succeeds" case would pin the timeout bump.

**2. No home-level account-switch test.** The new `home.spec` cases cover empty→skeleton and populated+offline→balance, but not the switch transition (wallets cleared → skeleton). It's implicitly covered by the provider spec, but the home-screen invariant the fix introduced isn't asserted at the home level.

### Code smells

**3. Loading semantics depend on a cross-module invariant.** `queryLoading` keys off `activeWallet.wallets.length === 0`, which only means "not loaded" because `buildWallets` (`wallet-snapshot.ts`) unconditionally returns 2 wallets on success — an invariant living in a different module. The inline comment helps, but if `buildWallets` ever returns `[]` for some future account shape, the home silently shows a permanent skeleton. A `status`-aware guard or an explicit "loaded" flag would be more robust.

esaugomez31:
@grimen

Test coverage
1. The actual point # 2 behavior is untested. Done. Added a "slow snapshot resolves within the timeout → succeeds" test: the snapshot resolves at 6s (past the old 5s budget, within the new 10s), asserting Ready + no timeout. Reverting the bump to 5s fails it.

2. No home-level account-switch test. Done. Added a switch test — status loading + empty wallets asserts the skeleton at the home level.

Code smells
3. Loading semantics depend on a cross-module invariant. Done. queryLoading is now !isReady && wallets.length === 0, so it no longer relies on buildWallets always returning 2 wallets — a ready account with no wallets now shows $0.00 instead of a permanent skeleton (new test), and a retained balance still survives a later offline refresh.

grimen (approved):

PR #3861, opened 2026-06-16
feat(hide-balance): add HiddenBalancePlaceholder across balance views

### LINKED ISSUE #4081: feat(hide-balance): apply hide-balance state consistently across balance views
## Problem

The hide-balance setting is not applied consistently across the app. When a user enables hide balance, amounts are masked on the home screen balance header, but several other views still reveal them:

- Send-bitcoin screens (wallet overview shows real balances during the send flow)
- Transaction history list items
- Transaction detail screen

This defeats the purpose of the setting: someone glancing at the screen while the user sends a payment or browses history can still see the balances.

## Proposed solution

Introduce a reusable `HiddenBalancePlaceholder` component and apply the hide-balance state consistently across all balance views (home header, send flow, transaction list, transaction details), with tests covering the placeholder and the affected components.

## Related

Implemented by PR #3861.

> [!NOTE]
> This issue was created retroactively from @designsats' PR #3861 to give that work a tracking issue — the PR predates the issue.

designsats:
## Summary

Adds a reusable `HiddenBalancePlaceholder` and applies the hide-balance state consistently across the app's balance views.

- Add `HiddenBalancePlaceholder` component and apply it across balance views
- Respect hide-balance state in the send-bitcoin screens
- Bold balance amounts and center-align the balance header
- Tests covering the placeholder, transaction item, and hide-amount component

## Test plan

- [ ] Toggle hide-balance and verify amounts are masked on the home/balance header
- [ ] Verify send-bitcoin screens respect the hide-balance state
- [ ] Verify transaction list items mask amounts when hidden
- [ ] `yarn test` passes

### Home

home-hidden

### Transaction history

tx-history-hidden

### Transaction details

tx-detail-hidden

### Send flow - wallet overview

send-wallet-overview-hidden

Closes #4081


designsats:
@blink-claw-bot review

designsats:
@grimen should be rdy

grimen (changes requested):
## Findings

**1. Peeking now permanently disables the "Always hide balance" setting (blocking).**

The old `switchMemoryHideAmount` toggled local `useState` only, so revealing the balance was session-scoped and the persisted setting re-hid it on the next app start. The new implementation calls `saveHideBalance`, which writes to the Apollo cache persisted via `CachePersistor` — the same thing the Security-screen switch does — contradicting the setting's own description ("Hides your balance on the home screen when the app starts… You can still see your balance by tapping on the main amount").

*Example:* A user with "Always hide balance" enabled taps the header to check their balance; on every subsequent app launch the balance is visible until they re-enable the setting in Settings.

*Proposal:* One option is to keep the persisted setting untouched and layer a session-scoped "revealed" flag on top of it, so a peek never outlives the session.

**2. Wallet-picker modal no longer respects hide-balance (blocking).**

The `ChooseWalletModal` in `send-bitcoin-details-screen.tsx` previously rendered `hideAmount ? HIDDEN_AMOUNT_PLACEHOLDER : …` for both wallets' balance lines, so amounts were masked with `****` when hiding was on. Those branches were removed, so the modal now always shows the real amounts; the inline wallet selector on the same screen was converted to the placeholder, so this looks unintended.

*Example:* With hide-balance enabled, opening the wallet picker on the send-details screen shows both BTC and USD balances in plain text.

*Proposal:* Render `HiddenBalancePlaceholder size="small"` in the modal rows when `hideAmount` is true, mirroring the inline selector.

**3. Amount column in transaction rows intercepts navigation taps.**

The row previously navigated to the transaction detail everywhere, but the new `Pressable` with `hitSlop={10}` and `minHeight: 40` over the amount area in `transaction-item.tsx` toggles hide instead. Combined with finding 1, an accidental tap on the right side of a row rewrites the persisted setting for the whole app — for the implementer to confirm this tap target is intended inside navigable rows; if it is, a comment would help.

*Example:* A user tapping the right half of a transaction row to open its details instead hides every amount in the app.

*Proposal:* One option is to keep the toggle on the header only and let the full row keep its navigation affordance.

## Test Coverage Gaps

- No test pins that a peek leaves the persisted `hideBalance` setting untouched across an app restart — the exact behavior finding 1 changes.
- The send-bitcoin screens have no coverage for the hide-balance state, which is where the finding 2 regression slipped through.

grimen:
Pushed three fix commits addressing the review findings:

**1. Peek is session-scoped again** (`0e1fc879`) — `HideAmountContainer` no longer writes `saveHideBalance`/`saveHiddenBalanceToolTip` on toggle; it holds session state seeded from the persisted setting, with a re-sync effect so a mid-session change from the Security screen still propagates (which the old seed-once state missed). The rewritten `hide-amount-component` test pins that a peek never calls the persistence helpers and that a persisted change overrides an earlier session peek.

**2. Choose-wallet modal masks balances again** (`92a47a58`) — the modal rows render `HiddenBalancePlaceholder` while hidden, mirroring the inline wallet selector. New spec (`send-details-hide-balance.spec.tsx`) opens the picker and asserts no digits render in either wallet row while hidden (it fails on the previous commit, passes with the fix). The spec mocks `useSendWallets` because `MockedProvider`'s single-use mocks made query-driven wallet data flaky across provider remounts.

**3. Transaction rows navigate again** (`f6eb45bc`) — removed the `Pressable` toggle around the amount column, so the full row keeps its navigation affordance; hiding still renders the placeholder. The toggle affordance remains on the balance header, the wallet-overview eye icon, and the transaction-detail amount (kept there since it has no competing navigation and is now session-scoped via fix 1).

Verification: `tsc --noEmit` clean, lint clean on touched files, full `yarn test` green (449 suites / 4998 tests). Coverage on the touched files: `hide-amount-component.tsx` and `hidden-balance-placeholder.tsx` at 100%; `transaction-item.tsx` unchanged from the PR's own level (uncovered lines are pre-existing description-helper branches).

grimen:
Simulator demo (staging account, hide-balance toggled on via the eye icon, then transaction history → transaction detail):

| before (base) | after (this PR + review fixes) |
|---|---|
|  |  |

Before: the home header masks with `****`, but opening a transaction's detail reveals the full amount while hide-balance is on. After: the placeholder masks consistently — history rows, and the detail amount, which now supports tap-to-peek (session-only, per the review fix).

<sub>Demos live on branch `assets/pr-3861-demo` — deleting it breaks them.</sub>

grimen (approved):

designsats:
It doesnt work as my PR. 

Two differences: 
- setting doesnt persist 
- Amounts in balances arent shown in Send flow at all, user  has to exit send flow to display their balances. Bad UX. 
- flickering on android, size of Balances changes once loaded, causing jumping.

PR #3863, opened 2026-06-17
feat(migration): custodial to self-custodial migration UI

esaugomez31:
## Summary

UI for migrating a custodial Blink account to a self-custodial (Spark) account, and the EU wind-down states that drive it: the settings entry, the guided flow (explainer, optional API-keys and keep-receiving steps, history download, balances summary and transfer), the home bulletin, the receive-disabled prompt, and the non-dismissable gate once the account is closed. The migration/wind-down data is read entirely through a single temporary mock; the real backend queries are wired in a branch stacked above this one.

Part of blinkbitcoin/blink-wip#800

## In this PR

- **Settings entry.** Re-enabled the "Move to non-custodial" row in the Account section (custodial accounts only). It resumes an in-progress migration through the checkpoint route or starts a fresh one, and hides itself while the self-custodial kill-switch is off.
- **Wind-down states.** The whole feature is driven by a server-reported phase, never computed from the device clock: `PreCutoff` (works normally, migration is voluntary), `ReceiveDisabled` (receiving is turned off, the "Migrate now" prompt appears), and `GatedClosed` (the account is closed except for the guided migration). The phase and its dates come from a single mock (`utils/backend-mock.ts`, `Query.windDown` shape in `app/types/wind-down.ts`) so the integration branch swaps only that layer.
- **Entry gate.** A single choke point for both entry points (settings and the forced blocker) orders the pre-flow checks per the wind-down design: an API-keys warning first, then a custodial dollar-balance precondition (emptied manually pre-deadline; post-gate the flow converts it), then the intro in the mode the account's situation demands (voluntary, forced pre-deadline, or the armed gate).
- **Explainer.** Reworked into five **required acknowledgement checkboxes**, revealed one at a time as each is checked. The CTA stays disabled until all five are checked, then routes into the flow.
- **Optional per-account steps.** A **keep-receiving** screen shows the account's lightning address before it stops working (shown only when the account has one), and a **download-history** step exports the transaction CSV (shown only when there is history, and only on a fresh run). Both follow the per-screen navigation rules (back arrows on the mid-flow steps, no exit once committed).
- **Balances summary and transfer.** A balances-overview screen shows the current and resulting balances, the network fee (or that Blink covers it), and, post-gate only, the exchange rate. It never renders zeros while loading (spinner and disabled Approve). Approving lands on the transfer screen, which provisions the self-custodial account, swaps the active session to it, marks it backed up, and clears the migration state; any failure routes to contact support instead of hanging.
- **Backup provisions and lands on the new account.** The flow provisions a self-custodial account (without activating it) so the shared backup screens show its recovery phrase. Cloud, password-manager, and manual backup all work mid-migration, because the account identity is derived offline from the mnemonic. The provisioned-but-inactive wallet is tracked in a per-owner registry so it is reused on resume (never re-provisioned) and hidden from the account switcher until the migration activates it.
- **Home bulletin.** One-time dismissable row on the home screen for self-custodial accounts pointing at the blog post, persisted per account.
- **Receive-disabled prompt and receive gate.** In the `ReceiveDisabled` phase the home greys out Receive (its tap reopens the prompt) and shows a dismissable, server-dated "Migrate now" modal that yields to every other home modal; the dismissal is scoped per account for the session. A route-level guard bounces any path into the receive screen back home while receiving is blocked.
- **Forced "Time to upgrade" gate.** Custodial accounts in a restricted country, or any account once the gate arms, are blocked at every app-open by a non-dismissable full-screen that replaces the tabs until they migrate. The restricted countries are a Firebase remote config (`custodialMigrationRequiredCountries`, defaulting to `US`), and a persisted flag stops the gate being bypassed by changing location afterwards.
- **Contact support.** A support screen gathers the account diagnostics (account id, pubkey, username, email, phone, app version, platform, country) from a single keychain read and opens a pre-filled support email.
- **Kill-switch.** A single `useSelfCustodialDisabled` rule (`nonCustodialEnabled` remote config) pauses every entry into the migration (settings row, blocker, gate, and the migrate-now prompt) while leaving the server-driven receive greying in place.
- **De-minimis fee subsidy.** The balances preview mirrors the backend exactly: a zero balance previews zero, a balance at or below the de-minimis threshold has its fee covered by Blink and transfers whole, and anything above uses the backend's drain solve (`max(0.5%, 10 sats)` reserve) so the mocked figures match what the real query will return.
- **Shared component.** Extracted `RevealedCheckboxList` (the reveal-one-at-a-time checkbox list) and reused it in the existing backup security-checks screen; `CheckboxRow` forwards a `testID`. Feature-level layout is shared through `MigrationStepLayout`, `MigrationCloseHeader`, `BalancePairCard`, and `GateBalances`.
- **i18n.** The full `AccountMigration` copy group (explainer checks, the three intro modes, the balances summary and its "not available" states, the migrate-now and dollar-balance modals, keep-receiving, download-history, API-service, the contact-support diagnostics, and the resume-failed toast) plus `NonCustodialInfoBulletin`, translated across all 28 languages with the fee and dates as interpolated values.

## Pending (backend integration)

- **Backend queries and mutations.** The migration/wind-down data (`Query.windDown`, `Account.migration.preview`) is served by a temporary mock, and the funds transfer is a fixed-delay placeholder that swaps the session without moving funds. The real query and the `migrationStart` / `migrationCommit` / `migrationLnAddressTransfer` mutations, their error codes, and the "migrated" account status are wired in the branch stacked above this one. A ship-gate flag (`IS_ACCOUNT_AFFECTED`) keeps the wind-down UI off until then.

## Notes

- Custodial-only: the settings entry, the receive-disabled prompt, and the forced gate apply to custodial accounts; all are inert for self-custodial accounts.
- The required-checkbox, per-screen navigation, and modal-priority behaviors match the wind-down acceptance scenarios.


esaugomez31:
* **#3934** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3917** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3863** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

designsats:
@esaugomez31 Example of the error message which shows in case user has dollar balance bellow the limit. 

forced-upgrade-custody

designsats:
Many updates in oroginal ticket incoming

grimen (changes requested):
## Findings

### High

**1. Two custodial profiles on the same phone can end up sharing one migration wallet.**
The pending-wallet registry is keyed by `activeAccount.id` (`use-pending-migration-accounts.ts`). But every custodial profile on a device has the same id: the constant `custodial-default` (`app/types/wallet.ts`). So the app cannot tell two custodial profiles apart.
*Example:* Alice (profile 1) starts a migration. The app creates wallet W1 for her, she sees its recovery phrase, then she stops halfway. Bob (profile 2) later starts his own migration on the same phone. The app finds W1 in the registry and gives it to Bob. Bob's money is now on a wallet whose recovery phrase Alice has. Alice can take Bob's funds. The same problem lets Bob resume Alice's half-finished flow (skipping the terms and checkboxes he never accepted) and mixes up their "don't show again" dismissals.
*Proposal:* key the registry, the checkpoint ownership check, and the dismissals by something that is truly unique per profile — the Galoy account id. If the stored owner doesn't match the current profile, treat it as "not mine" and start fresh (don't delete it — the other profile may still need it). Add a test with two custodial profiles on one device.

**2. The kill-switch does not cover all ways into the migration.**
The remote config `nonCustodialEnabled` is meant to pause every entry into the migration. It works for the settings row, the blocker, and the migrate-now prompt. It is missing in two places: the home bulletin (`useMigrationReminderBulletin`) and the `blink://account-migration` deeplink (`migration-entry-screen.tsx`).
*Example:* Spark has an outage, so ops turns the flag off. The home bulletin still shows. A user taps its "Migrate" button and is walked through wallet creation and a fund transfer — against a system that is currently down.
*Proposal:* check the flag once, in the one place every entry passes through: `MigrationGate`. If the flag is off, the gate shows a "temporarily unavailable" screen. That protects every entry point automatically, including ones added in the future. The per-entry checks then only need to hide buttons (add that to the bulletin too, so users don't see a button that leads to a wall). One test on the gate covers it.

### Medium

**3. Deeplinks can open screens on top of the "account closed" blocker.**
The GatedClosed blocker only replaces the main tab area. Deeplinks open screens on the root stack, which sits above it.
*Example:* a blocked user taps a `lightning:` payment link in a chat app. The send screen opens on top of the blocker and works normally. The same is true for the convert, settings, and scan-qr links. Screens that were already open when the block started also keep working.
*Proposal:* enforce the block in the navigation container: while the gate is armed, ignore all deeplinks except the migration ones, and reset the stack when the gate arms so nothing is left open above it.

**4. The gate's safety checks pass when their queries fail. Also, the PR description is out of date.**
The gate only waits for `loading` (`migration-gate.tsx`). If the API-keys query errors, the app assumes "no API keys" and skips the warning. If the balance query errors, the app assumes "$0" and lets the user through the dollar-balance check.
*Example:* a user with $50 and two API keys opens the gate on bad Wi-Fi. Both queries fail. The user enters the flow with no warning and a full dollar balance.
Separately: the PR description says a persisted flag stops users from dodging the gate by changing location. That code was removed in commit a345d1b7e. Today, relaunching the app offline shows the normal, unblocked app. A comment in `root-navigator.tsx` still refers to the deleted code.
*Proposal:* only proceed when the data is actually there — `useCustodialWalletBalances` already exposes `isReady` for this; add the same for the API-keys query, and show a retry button on error. Decide on purpose what should happen on an offline relaunch (bring back a persisted flag, or accept the gap until the backend enforces it). Update the description and remove the stale comment.

**5. If the app dies at the wrong moment, a funded wallet can disappear from the account switcher — permanently.**
At the end of the migration, `use-complete-migration.ts` clears the pending-wallet record without awaiting the write. Pending records never expire.
*Example:* the transfer succeeds and the app switches to the new account, then the phone dies before the cleanup write lands. The stale record now lives forever. The new account is only shown in the switcher while it is active — so the next time the user switches profiles, their funded wallet vanishes from the list for good. The only way back is restoring from the recovery phrase.
*Proposal:* await the cleanup before navigating, and — more importantly — add a self-healing check: on load, drop any pending record whose wallet has already been activated.

**6. Coming back to an abandoned migration can create a second wallet and orphan the first.**
`use-migration-account.ts` reuses the already-created wallet only if it appears in the account registry. But the hook's `loading` flag doesn't include the registry's own loading state.
*Example:* a user abandoned a migration 3 days ago (the resume checkpoint has expired) after writing down wallet W1's phrase. They re-enter on a slow device and tap the button before the registry has loaded. The app doesn't find W1, creates W2, and overwrites the record. W1 now shows up in the switcher as a mystery empty account — and it's the one whose phrase the user has on paper. Killing the app between wallet creation and saving the record leaks the same kind of orphan.
*Proposal:* include the registry's loading state (and treat a failed read as "not ready", not "empty"). Also save the pending record *before* creating the wallet — the id is known up front — so a crash leaves a harmless dangling record instead of an orphaned wallet.

**7. The preview data contract won't survive contact with the real backend.**
`AccountMigrationPreview` is synchronous and only knows about the BTC balance — because the mock is synchronous and only uses BTC.
*Example 1:* the real backend query is async. While it loads, the fields default to zero — and the Approve button enables on a screen full of zeros. That breaks the "never show zeros while loading" rule this PR carefully enforces everywhere else.
*Example 2:* after the gate arms, the flow itself converts the user's dollars. A user with 0 sats and $50 would see a final screen saying they will receive nothing.
*Proposal:* since the contract lives in this PR, extend it here: add loading/error states (folded into `isReady`) and fields for the dollar balance and its converted amount, served by the mock. Add a test for the post-gate, nonzero-dollars case.

**8. Four timing bugs in the flow's state machine:**
- **a.** `migration-required-screen.tsx` decides whether to show the "keep receiving" step before the lightning-address query has finished. On a cold start, a fast tap skips the warning for a user who does have an address. *Proposal:* wait for the query, or always visit the screen and let its existing self-skip decide.
- **b.** After the migration completes, the balances screen (still mounted underneath) re-saves the checkpoint that was just cleared. Harmless today, a trap for any future code that checks "is there a resumable migration". *Proposal:* only save while the screen is focused (`useIsFocused()`).
- **c.** If the balance query fails on the final screen, the user is stuck: endless spinner, back button blocked, no retry — only "Contact support", for what may be a Wi-Fi blip. *Proposal:* show an error message with a Retry button; keep Approve disabled.
- **d.** With the kill-switch on while the account is gated, the home screen shows Receive as enabled — but tapping it bounces straight back with no explanation. *Proposal:* grey Receive using `useWindDownReceiveBlocked` (which covers both blocked phases), and show a short "receiving is disabled" message where the tap currently does nothing.

**9. One test suite can never fail.**
`use-wind-down-status.spec.ts` starts with `if (result.current === null) return` — and the mock always returns `null` right now (`IS_ACCOUNT_AFFECTED = false`). Every assertion in the file is skipped. The rules it is meant to protect (the ordering of the wind-down dates) could break without any test noticing.
*Proposal:* force the "affected" branch with `jest.mock`, test both branches explicitly, and delete the early return.

### Low

**10.** On the failure screen, "Contact us" does nothing if the phone has no mail app — `Linking.openURL("mailto:…")` fails silently (`use-contact-support.ts`). That screen has no other exit. *Proposal:* check `canOpenURL` first; if it fails, show the support address with a copy button.

**11.** One bad timezone string from the backend would crash the home screen for every custodial user: `formatDayAndMonth` passes it straight to `toLocaleDateString`, which throws. *Proposal:* try/catch, fall back to the device timezone, report the error.

**12.** The final screen reads the balance with Apollo's cache-first default, so a payment received moments earlier may not appear in the figure the user approves. *Proposal:* for the implementer to decide — if the approved figure should always be fresh, use `cache-and-network`; if cache-first is acceptable because the backend recomputes at commit time, a short comment would settle it.

**13.** The LNURL-withdraw screen (an incoming-funds path) is not behind the receive block. A user scanning a voucher while receiving is disabled gets a raw server error instead of the migrate prompt. *Proposal:* wrap `redeemBitcoinDetail` in `WindDownReceiveGate`, like `receiveBitcoin`.

**14.** On iOS, resuming from a cloud-backup checkpoint sends the user back to the start (all 5 checkboxes and the terms again). Android resumes forward. There's no comment explaining why, and the test pins the behavior without a reason. *Proposal:* if this is intended (e.g. an iCloud re-prompt concern), add a comment explaining why; otherwise make iOS resume forward like Android.

**15.** If saving the "backup completed" flag fails, the flow continues anyway (`use-complete-backup.ts`). The user who just did a backup is later nagged that they have none. *Proposal:* stop and offer retry, the same way checkpoint saves already do.

**16.** If wallet creation fails halfway, the recovery phrase written to the keychain is not cleaned up (`bridge/lifecycle.ts`), while the restore path does clean up after itself. *Proposal:* for the implementer to confirm — if keeping the entry is intentional (e.g. to never delete key material once written), add a comment; otherwise mirror the restore path's cleanup.

### Nits

**17.** `keep-receiving-screen.tsx` has an unused `buttonsContainer` style.
**18.** `icon-hero.tsx` now renders an empty text element when `subtitle=""`; the old code skipped it.
**19.** The new `info-row` tests pass even if the new props do nothing (styles are mocked away).
**20.** The explainer test mocks `useMigrationAccount` with an outdated shape, so "button disabled while creating the wallet" is never tested.
**21.** The backup-method screen loads the recovery phrase from the keychain as soon as it opens, before the user picks a method. Load it when a method is chosen.
**22.** `MigrateNowModal` is mounted/unmounted instead of toggled like its sibling modals, so it skips its exit animation.
**23.** The fee-math sweep test checks an equation the mock defines, so it can't fail. The real rules — the reserve is satisfied, and the payout is the largest possible — are untested, as is the exact-half rounding case.
**24.** A comment above the `useMigrationBlocker` import in `root-navigator.tsx` describes code that was deleted (see 4).

## Test Coverage Gaps

- Two custodial profiles on one device: wallet reuse, checkpoint ownership, and dismissals are only tested with a single profile (1).
- No test that the bulletin, the deeplink, or the gate respect the kill-switch (2); `use-self-custodial-disabled` has no test of its own.
- No test for deeplinks bypassing the armed gate (3), or for the home screen when the account is gated and the kill-switch is on (8d); the receive-gate test never uses the `GATED_CLOSED` status directly.
- No test for the gate letting users through when its queries error (4).
- No test for the crash windows at the end of the flow: cleanup not persisted before navigation, or a kill during wallet creation or the account swap (5, 6).
- The `use-migration-account` test always mocks a fully loaded registry, so the loading race and the failed-read path are untested (6).
- No test anywhere for the post-gate case with a dollar balance (7).
- No test for the "keep receiving" decision while its query is still loading (8a), or for the checkpoint being re-saved after completion (8b).
- No direct tests for `use-migration-balances-preview` or `use-custodial-wallet-balances`; the screen test stubs out currency display, so real formatting is untested.
- `use-wind-down-status.spec.ts` is dead code behind its `null` guard (9).
- The fee-math sweep can't fail (23).
- The `info-row` prop tests can't fail (19), and the explainer test's outdated mock skips the provisioning-disabled case (20).
- `migration-step-layout` and `migration-close-header` are only tested indirectly through screens.

## Code Smells

- **Rules enforced at every call site instead of one central place.** This is the one pattern that repeats: the kill-switch is checked at each entry point and two were forgotten (2); the receive block is applied route by route and one route was forgotten (13); the home screen's modal-priority list names each sibling modal and already omits two. In all three cases the fix is the same: enforce the rule once, at a choke point, so new code is covered automatically.
- **The data contract is shaped like the mock, not like the real backend.** `AccountMigrationPreview` is synchronous because the mock is (7). That's exactly why the real async query can't slot in behind the seam without changing the contract.
- **An id that doesn't mean what it looks like.** `activeAccount.id` reads like a unique owner, but for custodial profiles it's the same constant for everyone (root cause of 1). Any future per-owner feature will hit the same trap. A real owner id on the account record would fix the class of bug, not just this case.
- **Fire-and-forget writes at the moments that matter most.** Cleanup after the account swap isn't awaited (5), a failed backup-status save is ignored (15), and some promise chains have no error handling. This PR itself established the right pattern — checkpoint saves return false and stop the flow — it just isn't applied everywhere.
- **The home screen keeps growing.** The bulletin, the prompt, the receive greying, and the modal ordering all landed in the already-huge `home-screen.tsx`. Not new to this PR, but each addition makes it worse; the modal ordering in particular would be safer as one small dedicated component.
- **Loose prop contracts on the shared components.** `InfoRow` silently ignores `valueColor` when `isValueMuted` is set, and `RevealedCheckboxList` misbehaves if its `labels` array ever changes length. Harmless with today's callers, undocumented for tomorrow's.

esaugomez31:
@grimen  Decision on each point below.

## Findings

### High

**1. Two custodial profiles on the same phone can end up sharing one migration wallet.**
**Will fix.** The app supports multiple accounts of each type, so two custodial profiles on one device is a real scenario and `custodial-default` genuinely can't tell them apart. We'll key the pending-wallet registry, the checkpoint-ownership check, and the dismissals by the real Galoy account id, treat a stored owner that doesn't match the current profile as "not mine" (without deleting it), and add a two-custodial-profile test.

**2. The kill-switch does not cover all ways into the migration.**
**Will fix, centrally, and it needs to go one step further than the proposal.** Resume bypasses the gate today: both Settings and the deeplink decide resume-vs-fresh themselves and jump straight to a checkpoint screen (`navigateToCheckpoint`/`replaceToCheckpoint`), so a single check in the gate would still leak the resume-via-deeplink path. We'll make the gate the real choke point: Settings and the deeplink stop deciding and always route to the gate; the gate then, in order, shows a "temporarily unavailable" screen when the flag is off, else resumes to the checkpoint, else starts the intro. That covers every entry, present and future. We're deliberately NOT hiding the bulletin's button: with the flag off the gate shows a clear "temporarily unavailable" screen, so the button leads to an informative message, not a wall — no per-entry button logic needed. One gate test covers the disabled case.

### Medium

**3. Deeplinks can open screens on top of the "account closed" blocker.**
**Will fix.** Confirmed: the blocker replaces only the `Tab.Navigator`, and the root stack sits above it, so root-stack deeplinks (send, convert, settings, scan) open over it. Worth flagging that it's more than a gated-account UX gap: the same bypass now also perforates the migration lock this stack adds — during `IN_PROGRESS` the account is still Active server-side, so a deeplink send would let the user spend the very balance the transfer is about to claim. We'll enforce it in the navigation container: while the gate or lock is armed, ignore every non-migration deeplink and reset the stack when it arms so nothing stays open above it, with tests that a blocked deeplink is dropped and legitimate ones still work.

**4. The gate's safety checks pass when their queries fail. Also, the PR description is out of date.**
**Will fix, both parts.** (a) The gate will wait for the data to actually be there, not just for `loading` to end: we'll make `useActiveApiKeys` expose error/ready like `useCustodialWalletBalances` already does, keep the checks conservative on error (don't let through), and show a retry. Worth noting the dollar-balance leak is already caught server-side — the backend re-validates an empty USD wallet on `migrationStart` and `migrationCommit` — so that half is UX degradation rather than a real bypass; the API-key warning is client-only, so that half matters. (b) We'll update the PR description and remove the stale comment (same as 24). The offline-relaunch behavior is settled by the stacked backend work: the server's `migration.status` is the authority, and an offline relaunch shows the app until the next successful read — a deliberate, documented trade-off.

**5. If the app dies at the wrong moment, a funded wallet can disappear from the account switcher — permanently.**
**Will fix.** Confirmed: `use-complete-migration.ts` fires the checkpoint and pending-record clears without awaiting. We'll await the cleanup before navigating and, as the real fix, add the self-healing check you suggest: on load, drop any pending record whose wallet is already activated in the registry. This lands together with 1, since both reshape the pending-record store.

**6. Coming back to an abandoned migration can create a second wallet and orphan the first.**
**Will fix.** Confirmed on both counts: the hook's `loading` covers the checkpoint and pending record but not the account registry, so a tap before the registry loads misses the existing wallet and provisions a second one; and the wallet is created before the pending record is saved, so a crash in between leaks an orphan. We'll fold the registry's loading (and a failed read) into the gate, and save the pending record before creating the wallet so a crash leaves a harmless dangling record instead of an orphaned wallet. Done together with 5 and 7 as one end-of-flow hardening pass.

**7. Already done — in the stacked backend-integration PR (#3917) that sits on top of this one.** `use-migration-preview.ts` there drops the sync mock for the real `Query.migration.preview`, and loading/error are folded into a total `isReady`, so Approve never enables on a screen of zeros — exactly your Example 1. On Example 2, the real backend does not convert dollars: it rejects a non-empty USD wallet on `migrationStart` and `migrationCommit`. So instead of a "you receive nothing" conversion screen, we block entry whenever there's a dollar balance in every phase (the `!isGated` exemption and the "flow converts dollars" comment are both gone). That's why we don't add converted-amount fields to the contract — there's no conversion; emptying dollars is a gate precondition.

**8. Four timing bugs in the flow's state machine.**
**a — Will fix.** The "keep receiving" decision will wait for the lightning-address query (or always visit the screen and let its self-skip decide).
**b — Will fix.** The checkpoint save will be gated on `useIsFocused()` so a screen mounted underneath can't re-save a cleared checkpoint.
**c — Already done, in the stacked integration PR (#3917).** The final screen already splits a failed balance read into a Retry (network) and a support handover (server rejection), with Approve disabled — exactly your proposal.
**d — Will fix.** Receive will be greyed with `useWindDownReceiveBlocked` (covers both blocked phases) and show a short "receiving is disabled" message where the tap currently does nothing.

**9. One test suite can never fail.**
**Will fix.** Confirmed: the `if (result.current === null) return` guard plus the always-null mock skips every assertion. We'll force the affected branch with `jest.mock`, test both branches, and drop the early return.

### Low

**10.** Contact us does nothing without a mail app.
**Will fix.** We'll check `canOpenURL` first and fall back to showing the support address with a copy button.

**11.** A bad timezone string would crash the home screen.
**Will fix.** We'll wrap `formatDayAndMonth` in try/catch, fall back to the device timezone, and report the error.

**12.** The final screen reads the balance cache-first.
**Will fix.** We'll switch that query to `cache-and-network`, so the dollar check is fresh client-side too, not only re-validated by the backend at commit.

**13.** LNURL-withdraw is not behind the receive block.
**Will fix.** We'll wrap `redeemBitcoinDetail` in `WindDownReceiveGate`, like `receiveBitcoin` — and it's part of the same "enforce the receive block at one choke point" cleanup as the code smell.

**14.** iOS resumes a cloud-backup checkpoint to the start.
**Will fix.** No intended reason for the divergence — we'll make iOS resume forward like Android and drop the special-case.

**15.** A failed backup-completed save lets the flow continue.
**Will fix.** We'll stop and offer retry, the same way checkpoint saves already do.

**16.** A half-failed wallet creation leaves the recovery phrase in the keychain.
**Will fix.** We'll mirror the restore path's cleanup: on a half-failed creation, delete the keychain entry.

### Nits

**17.** Unused `buttonsContainer` style. **Will fix** (remove it).
**18.** `icon-hero` renders an empty text element for `subtitle=""`. **Will fix** (skip it when empty, as before).
**19.** `info-row` tests pass even if props do nothing. **Will fix** (assert the rendered style, not the mock).
**20.** Explainer test mocks `useMigrationAccount` with an outdated shape. **Will fix** (update the mock so the provisioning-disabled case is actually exercised).
**21.** Backup-method screen loads the phrase before a method is chosen. **Will fix** (load it when a method is chosen).
**22.** `MigrateNowModal` mounts/unmounts instead of toggling. **Will fix** (toggle like its siblings so the exit animation runs).
**23.** Fee-math sweep test can't fail. **Will fix** (assert the real rules: reserve satisfied, payout maximal, and the exact-half rounding case).
**24.** Stale comment above the `useMigrationBlocker` import. **Will fix** (remove it; same as 4b).

## Test Coverage Gaps

**Will close, with the fixes.** Each fix lands with the test it names — two custodial profiles (1), the kill-switch on every entry via the gate (2), deeplinks over the armed gate and the gated-home Receive state (3, 8d), the gate letting users through on query error (4), the end-of-flow crash windows (5, 6), the loading-race and failed-read in `use-migration-account` (6), the post-gate dollar case (already covered in the stacked PR, 7), the keep-receiving-while-loading and post-completion re-save (8a, 8b), and the fee-math and info-row tests that can't fail (23, 19). Standalone additions: `use-self-custodial-disabled`, a direct `GATED_CLOSED` receive-gate test, direct `use-migration-balances-preview` / `use-custodial-wallet-balances` tests, and direct `migration-step-layout` / `migration-close-header` tests. `use-wind-down-status.spec.ts` gets both branches (9).

## Code Smells

Agreed on all of them, and the fixes above target the patterns, not just the instances:
- **Rules at every call site.** The kill-switch moves to one choke point (the gate, 2) and the receive block to one place (13); we'll also pull the home's modal-ordering into a small dedicated component rather than a per-sibling list.
- **Contract shaped like the mock.** The real async contract lives in the stacked integration PR (7); the seam holds because loading/error fold into a total `isReady`.
- **An id that doesn't mean what it looks like.** We'll key the migration by the real Galoy account id (1). The broader "put a real owner id on the account record" refactor is the deeper fix for the whole class; we'd scope that as its own change rather than fold it in here — flagging it so it isn't lost.
- **Fire-and-forget writes.** We'll apply the return-false-and-stop pattern the PR already established to the account-swap cleanup (5), the backup-status save (15), and the provisioning order (6).
- **Home screen keeps growing.** Agreed; the modal-ordering extraction above is the first cut.
- **Loose prop contracts on shared components.** We'll tighten/document `InfoRow` (`valueColor` vs `isValueMuted`) and `RevealedCheckboxList` (the `labels`-length assumption).

esaugomez31:
@grimen this is how each one turned out, already up on this branch and restacked.

## High
1. Fixed. Pending registry, checkpoint ownership, and dismissals keyed by the real Galoy account id, not `custodial-default`. Non-matching owner starts fresh, isn't deleted. + two-profile test.
2. Fixed. Single choke point: the gate shows a "temporarily unavailable" screen when the flag is off, covering every entry. Bulletin button intentionally kept.

## Medium
3. Fixed. While the gate is armed, non-migration deeplinks are ignored and the stack is reset to Primary.
4. Fixed. Gate blocks until each query settles with data, shows a retry on error. Stale `root-navigator` comment removed (= #24).
5. Fixed. Cleanup awaited before navigating; self-healing drop of already-activated pending records on load.
6. Fixed. Registry loading folded into the gate; pending record saved before the wallet is created, and the save aborts provisioning on failure.
7. #3917. Real `Query.migration.preview`, loading/error folded into `isReady`.
8a. Fixed. CTA waits for the lightning-address query before routing.
8b. Fixed. Checkpoint saved only while focused (and once the preview has figures).
8c. #3917. Retry vs support-handover instead of an endless spinner.
8d. Fixed. Home Receive greyed via `useWindDownReceiveBlocked`, with a "receiving is disabled" message.
9. #3917. Real `Query.windDown`; spec tests both branches and can fail.

## Low
10. Fixed. Falls back to copying the support address. Uses open-then-catch (not `canOpenURL`, which false-negatives on Android 11+ for `mailto`).
11. Fixed. `formatDayAndMonth` try/catch with device-timezone fallback + report.
12. #3917. Preview read `no-cache`, so the approved figure is always fresh.
13. Fixed. `redeemBitcoinDetail` wrapped in `WindDownReceiveGate`.
14. Fixed. iOS special-case dropped; resumes forward like Android.
15. Fixed. Failed backup-completed save stops and offers a retry toast.
16. Fixed. `bridge/lifecycle.ts` mirrors the restore cleanup: deletes the phrase on a half-failed create.

## Nits
17. Removed the unused `buttonsContainer` style.
18. `icon-hero` skips the subtitle text for an empty string.
19. `info-row` tests now assert the rendered style.
20. Explainer test uses the current `useMigrationAccount` shape and covers the provisioning-disabled state.
21. Backup-method reads the phrase on the credential tap, not on mount.
22. `MigrateNowModal` kept mounted and toggled by `isVisible`.
23. #3917. Fee math moved to the backend; the mock and its test were removed.
24. Fixed with #4.

grimen (changes requested):
## Findings

### High

**1. Arming the gate removes the lock screen from the app.**
When the account becomes gated, the app resets navigation so only the blocker remains (`navigation-container-wrapper.tsx`). But that reset also skips `authenticationCheck` — the screen that asks for the PIN or fingerprint on launch.
*Example:* a gated user has a PIN set. They open the app. The PIN prompt never appears — they land directly on the migration gate, and everything behind it (balances, the transfer of their funds) works on a phone that was never unlocked. A second effect: the app still *believes* it is locked, so the one deeplink that is supposed to work while gated — the migration one — is queued waiting for an unlock that never comes. Nothing shows today because the mock is off; this becomes real the day the backend arms the gate.
*Proposal:* reset to `authenticationCheck` instead of Primary and let the normal unlock flow continue into the gate. Add a test for "gate armed + PIN enabled".

**2. Resuming a migration still bypasses the kill-switch.**
This is the previous round's finding 2, marked fixed — but the fix covers only a *fresh* start. `migration-entry-screen.tsx` first checks for a saved checkpoint, and if there is one it jumps straight to that screen — before the gate, the only place that checks the flag, ever runs. The comment in the bulletin hook says its Migrate button "routes through the gate, so the single choke point covers the disabled case"; the code doesn't do what the comment says.
*Example:* Spark has an outage, so ops turns `nonCustodialEnabled` off. A user who is halfway through a migration taps Migrate on the home bulletin. Instead of the "temporarily unavailable" screen, they continue creating a wallet and transferring funds — against a system that is currently down.
*Proposal:* what the decision comment already promised: every entry point goes to the gate, and the gate — in order — checks the flag, resumes the checkpoint, or starts fresh. One test: flag off + saved checkpoint.

**3. The safety checks added last round can never fire, because the storage layer never reports failure.**
`saveJson` (`app/utils/storage/storage.ts`) catches every error and returns `false` — and no caller reads that return value. Two of last round's fixes depend on a failed write being *loud*: checkpoint saves are documented to "resolve false so callers can stop the flow", and the pending-wallet record "MUST throw on failure" so a wallet is never created without one. Neither can actually happen. The same shape exists in the keychain: `setMnemonicNetworkForAccount` returns `false` instead of throwing, so the new cleanup in `bridge/lifecycle.ts` can't catch it, and a wallet can end up with no network label — which the network-mismatch check then reads as "no mismatch".
*Example:* the phone's storage write fails (disk full, app killed in the background) while a wallet is being provisioned. The guard that should abort cannot fire, so the wallet is created with no record of it — exactly the orphan it was built to prevent: a wallet whose recovery phrase may already be written down, that the app has now forgotten. The tests don't notice, because they replace the storage layer with one that *throws* — a behavior the real one does not have.
*Proposal:* make `saveJson` and the keychain setters report failure for real (throw, or a result that callers must check). Check it at the three places above, and change the tests to mock the layer as it actually behaves.

### Medium

**4. Pressing back during the success animation lands on a dead screen.**
The transfer screen moves to the success screen with `navigate`, so it stays alive underneath. The success screen has no back protection and only moves on after its animation plus a ~2-second hold.
*Example:* the user presses back during the animation. They return to the transfer screen — whose work is already done, so it shows a spinner that will never end, and it swallows the back button. The only way out is force-quitting the app.
*Proposal:* `replace` (or reset) at this point of no return, so there is nothing to go back to.

**5. The old account is closed before checking that the new one exists.**
`use-complete-migration.ts` first discards the custodial session, then activates the account id stored in the checkpoint — without checking that this account is still on the device. The wallet-creation path performs exactly this check; the completion path doesn't.
*Example:* a user resumes at the balances screen, but the wallet behind the stored id is gone (keychain loss within the resume window). Approving closes their custodial session, then switches to an account that doesn't exist. The user now has neither account — on a screen whose back button is blocked.
*Proposal:* check the registry first; if the account is missing, keep the custodial session and route to contact support.

**6. When the owner lookup has no answer, three things break.**
Last round's fix 1 keys all migration state by the real account id — which now comes from a server query. While that query is loading, or after it fails, the id is `null`, and there is no retry. Three separate failures fall out of that one state:
*Example:* a user is offline while receiving is disabled. (a) The "Migrate now" modal appears, but its close button silently does nothing (`use-account-session-dismissal.ts`) — the modal cannot be dismissed for the rest of the session. (b) Any checkpoint save made in this state quietly erases the stored owner and wallet id, so the resume forgets its progress. (c) Worst: the backup screens receive an *empty* recovery phrase, and the confirm screen — with zero words to confirm — concludes that everything was confirmed and shows success. The user believes they backed up; nothing was recorded.
*Proposal:* give the lookup an explicit error state with a retry, and block the dependent actions while it is unresolved: the modal's close falls back to a plain dismissal, checkpoint saves wait, and the backup screens refuse an empty phrase. Refusing zero confirm-challenges is worth doing regardless.

**7. The migration deeplink can send a self-custodial user into the custodial gate.**
The entry screen waits for the checkpoint to load, but not for the account list — and the two loads race.
*Example:* a self-custodial-only user taps `blink://account-migration` on a slow device. The account list hasn't loaded, so "is this a self-custodial account?" reads as no — and they are placed in the "Time to upgrade" gate meant for custodial accounts. By the time the list loads, the screen that could correct it is gone.
*Proposal:* wait for both loads before dispatching (this folds into fix 2 if the gate becomes the dispatcher).

**8. Finishing the in-app dollar transfer silently abandons the migration.**
The gate's "Transfer" button hands the user to the conversion flow — whose success screen pops the whole stack after 3 seconds, including the gate the user came from.
*Example:* a user taps Transfer to empty their dollars, completes it, and lands on Home. The migration they were in the middle of is simply gone; nothing invites them back. (Separately: the cancel-and-return path re-checks the balance only from cache — nothing refreshes it when the gate regains focus.)
*Proposal:* a product decision: either return to the gate after the conversion, or accept the round-trip and refetch balances when the gate regains focus so the return visit is at least accurate.

**9. The two most important saves in the flow don't wait for their result.**
Last round established the pattern: a checkpoint save reports failure and the flow stops. The two commit points don't use it. On the balances screen, Approve becomes tappable before the point-of-no-return checkpoint has been written (and its result is ignored); on the T&C screen the save is fired and the user advances regardless.
*Example:* a fast Approve before the write lands, then a crash during the transfer: the app resumes at the backup step instead of the commit screen. On the T&C side, a failed write means the user is later asked to consent to terms they already accepted.
*Proposal:* await the save and gate the buttons on it, the way the other steps already do.

**10. The gate's Retry button fails invisibly.**
Retry fires both refetches with no error handling, and nothing on screen indicates that a retry is running.
*Example:* the Wi-Fi is still bad. The retry fails — as an unhandled rejection nobody sees — while the error screen sits unchanged. The user taps again and again, stacking up requests.
*Proposal:* catch the failure, show progress (`notifyOnNetworkStatusChange: true`), and disable the button while a retry is in flight.

**11. Kill-switch on while the gate is armed: payment links die silently.**
The deeplink blocking checks only "is the gate armed"; the full-screen blocker additionally hides when the kill-switch is on. In the combined state the app looks normal — home and send work — but every payment deeplink is silently thrown away.
*Example:* ops pauses migrations while accounts are gated. A user taps a lightning invoice in a chat app. Nothing happens, and nothing explains why.
*Proposal:* key the deeplink blocking and the blocker on the same condition — whichever way product wants that state to behave.

### Low

**12.** On iOS there is no way back from Contact support: no header, gestures disabled, and the "return to the balances screen" guard only exists for Android's hardware back button. Add a close or back control.

**13.** The native signer object created to derive the wallet's public key holds key material derived from the seed, and it is never freed (`bridge/wallet.ts`). Each derivation leaves key material sitting in native memory until garbage collection gets to it. Call `uniffiDestroy()` after reading the pubkey.

**14.** The support screen loads the full recovery phrase into memory just to derive a public key (`use-migration-support-details.ts`). The codebase has a lazy variant (`useLoadWalletMnemonic`) for exactly this case — screens that don't display the phrase.

**15.** Cloud-backup filenames now come from the new offline key derivation instead of asking the connected wallet SDK. No test compares the two. If they ever differ — network, key set, upper/lower-case hex — the app stops finding existing backups and quietly creates a second file.

**16.** If the phrase or identity hasn't loaded before a cloud backup, the failure message says "sign-in failed" — pointing the user at their Google/iCloud account when the problem is local.

**17.** Nothing protects the `IS_ACCOUNT_AFFECTED` switch from being committed as `true` by accident — no test asserts the mock is off. One line of test is cheap insurance against shipping the account-closed lockout to every custodial user.

### Nits

**18.** `use-custodial-owner-id.ts` is the only new query that selects `me` without its `id`; Apollo will log cache-merge warnings.
**19.** `RevealedCheckboxList` uses the label text as the React key; two identical labels (possible after translation) would share one checkbox state. Index keys are safe here — the list never reorders.
**20.** `useMigrationDiagnostics` builds a fresh array on every render, so the memoization built on top of it never actually holds.
**21.** A migration link with a trailing slash (`…/account-migration/`) is treated as a non-migration link and dropped while the gate is armed.
**22.** The Android CSV export relies on the share library appending `.csv` for the MIME type `text/comma-separated-values` — a mapping Android doesn't reliably have. The old code always put `.csv` in the filename. Worth one device check.
**23.** The keep-receiving screen shows a blank page while its data loads, where the gate shows a spinner for the same situation. And offline, the required-screen reads a *failed* address query as "no address" and skips the keep-receiving warning for a user who has one.
**24.** The balances-overview and contact-support footers copy the shared layout's button styling verbatim instead of reusing it — the shared component needed a slot for a scrollable middle, not a fork.
**25.** A failed wallet creation is reported to crashlytics twice (once in the bridge, once in the hook).
**26.** The galoybot→blinkbitcoinbot CI rename is unrelated to this PR — and release PRs labeled with the old name disappear from the changelog lookup.

## Test Coverage Gaps

- The resilience tests (checkpoint storage, pending records, the create-wallet cleanup) mock the storage and keychain layers as if they threw on failure. The real layers don't — so the tests prove handling for failures that can't occur, and hide finding 3.
- The armed-gate navigation reset has no tests at all: not the lock-screen interaction (1), not the reset, not the deeplink behavior under the kill-switch (11).
- No test resumes a migration with the kill-switch off (2) — the exact case last round's fix claimed to cover.
- No test for any of the null-owner failures (6): the un-closable modal, the erased owner, or the empty-phrase backup.
- `use-migration-balances-preview` and `use-custodial-wallet-balances` still have no direct tests — carried over from last round, unaddressed.
- No test that unchecking an explainer box disables the CTA again.
- The offline key derivation is never checked against what the SDK reports (15) — only against mocks.

## Code Smells

- **Safety code proven by tests of a world that doesn't exist.** Last round's "tests that can never fail," one layer deeper: the guard is real, but the failure it guards against is unreachable, and the test hides that by mocking a failure mode the dependency doesn't have (3).
- **Storage that answers `false` to everything.** `saveJson` and the keychain setters turn every failure into an ignorable return value. Fixing call sites one at a time (last round's 5 and 15, this round's 9) will keep leaking until the layer itself reports failure.
- **The choke point still isn't one.** Entry points still decide resume-vs-fresh on their own — which is precisely how the kill-switch leaked again (2). Until the gate is the only dispatcher, every new entry point can re-create this bug.
- **One `null`, three meanings.** For the owner id, "still loading," "failed," and "doesn't exist" are the same value — and three separate bugs grew out of that single ambiguity (6). An explicit loading/error/ready state fixes the class.
- **Forked instead of extended.** Two screens copy the shared layout's footer styling rather than the layout growing a slot (24).

---

Verified fixed from the last round: owner keying by real account id (1), the deeplink drop and stack reset mechanism itself (3), gate queries waiting for data with retry (4), awaited cleanup and self-healing pending records (5), save-before-create ordering (6), receive greying (8d), the wind-down spec's dead null-guard (9), the mailto fallback (10), the timezone try/catch (11), the voucher receive gate (13), iOS resuming forward (14), and nits 17–22. The new findings above sit almost entirely inside the fix commits themselves — the usual price of a hardening round nobody re-reviewed.

esaugomez31:
@grimen

Thanks for the second pass — this was exactly the "nobody re-reviewed the hardening" catch it needed. Resolutions below, finding by finding. Net: **23/26 fixed with tests, 6/7 coverage gaps closed**, smells addressed at the root where feasible. The four exceptions (2 deferrals, 1 native-only test, 1 not-in-diff) are called out honestly, plus two product calls for you.

### High

**1. Arming the gate removes the lock screen.** **Fixed.** The reset is now lock-aware via a `blockerEntryRoute(isAppLocked)` helper: while locked it resets through `authenticationCheck`, so PIN/biometric is never skipped and `isAppLocked` no longer sticks at true (which also un-freezes the queued migration deeplink). Once unlocked it goes straight to the blocker (no re-prompt). Tests: `blockerEntryRoute` unit + render tests (locked→authenticationCheck, unlocked→Primary).

**2. Resuming a migration still bypasses the kill-switch.** **Fixed** — exactly what the decision comment promised. The entry screen is now the single dispatcher; Settings and the bulletin route through it instead of deciding resume-vs-fresh, and it enforces the flag (and waits for `remoteConfigReady`) before any resume. The bulletin comment is now accurate. Test: flag-off + saved checkpoint → gate, not a resume.

**3. Safety checks can never fire.** **Fixed at the layer.** `saveJson` now propagates the write error; its two must-succeed callers reject for real, and the one best-effort caller (persistent-state) opts into quiet with an explicit catch. The keychain path checks the boolean and throws (`if (!labelled) throw`) in both create and restore, so the orphan cleanup runs. Tests rewritten to the real behavior (rejecting `saveJson` / `false`-returning setter). *One honest nuance:* the keychain **setter** still returns `false` at the layer — I fixed the two migration call sites, not the setter itself (wider consumer trace needed). Follow-up if you want it.

### Medium

**4. Back during the success animation lands on a dead screen.** **Fixed.** The transfer screen `reset`s to success, so the finished screen is off the stack.

**5. Old account closed before checking the new one exists.** **Fixed.** `use-complete-migration` verifies the account is still on the device before discarding the custodial session; if it's gone it keeps the session and returns false (→ support).

**6. Null owner breaks three things.** **Fixed** all three: (a) dismissal falls back to a session-global key so the modal is closable; (b) a null-owner checkpoint save refuses instead of erasing the stored owner/account id; (c) zero challenges no longer auto-"confirm." *Smell note:* I added the guards; I did **not** yet refactor `use-custodial-owner-id` to an explicit loading/error/ready + retry state — that's the class fix and I'd land it as its own change.

**7. Deeplink can send a self-custodial user into the custodial gate.** **Fixed.** The dispatcher now waits for the account registry to hydrate (not just the checkpoint) before deciding.

**8. Finishing the in-app dollar transfer abandons the migration.** **Fixed (product call).** Implemented option B: accept the round-trip and refetch balances when the gate regains focus, so the return visit is accurate. Return-to-gate would mean threading migration context through the shared conversion success screen — tell me if you'd rather have that.

**9. The two commit-point saves don't await their result.** **Fixed.** Balances Approve gates on the checkpoint write landing; T&C awaits the save and only advances on success.

**10. Retry fails invisibly.** **Fixed.** Retry catches the rejection and disables/spins the button while in flight.

**11. Kill-switch + gate armed: payment links die silently.** **Fixed (product call).** The deeplink drop is now keyed on the blocker's own visibility (`useMigrationBlocker().isVisible`), so when the kill-switch hides the blocker, payment deeplinks flow normally. *Test nuance:* covered via the shared-visibility render test; a direct linking-subscribe test is a good follow-up.

### Low

**12. iOS: no way back from Contact support.** **Fixed.** Added a visible Back control returning to the commit point (the hardware-back guard was Android-only).

**13. Native signer never freed.** **Fixed.** `deriveWalletIdentityPubkey` frees the signer in a `finally` (verified it's called even when reading the pubkey throws).

**14. Support screen loads the full phrase eagerly.** **Fixed.** Switched to the lazy `useLoadWalletMnemonic`, deriving the pubkey from a transiently-loaded phrase.

**15. Backup filename offline vs SDK, no test.** **N/A in jest.** The equality needs the native Breez signer, which isn't available in the test environment — flagged as a device/integration check. Open to another approach if you have one.

**16. Local error surfaced as "sign-in failed".** **Fixed.** Now shows a local backup-failed message instead of pointing at the cloud account.

**17. Nothing guards `IS_ACCOUNT_AFFECTED`.** **Fixed.** A test asserts `windDownMock` is null (the switch is off) — it fails loudly if the toggle is ever committed as `true`.

### Nits

- **18.** Fixed — `me { id }` added + codegen.
- **19.** Fixed — index keys (with a duplicate-label collision test).
- **20.** Fixed — memoized.
- **21.** Fixed — trailing slash trimmed on both ends.
- **22.** Fixed — `text/csv` (AOSP maps it; `text/comma-separated-values` it doesn't).
- **23.** Fixed — spinner while loading, and an errored address query no longer reads as "no address" (both keep-receiving and the required screen).
- **24. Deferred.** This is a layout-changing refactor (a scrollable-middle slot on `MigrationStepLayout`); it needs a visual pass and belongs in its own commit. Will follow up.
- **25.** Fixed — single crashlytics report on wallet-create.
- **26. N/A.** The galoybot→blinkbitcoinbot rename isn't in this branch's diff — it merged to `main` via #3919 before this branch (`git diff main…HEAD -- ci/pipeline.yml` is empty). Your diff base likely predated that merge.

## Test Coverage Gaps

- **Resilience tests** — fixed at the root (loud layer): the create-wallet and persistent-state specs now mock the real `false`/reject behavior; the checkpoint/pending specs exercise a genuinely-rejecting wrapper.
- **Armed-gate reset** — covered (lock-screen, reset). Deeplink-under-kill-switch is covered via the shared visibility key; direct linking test is a follow-up.
- **Resume with kill-switch off** — covered.
- **Null-owner failures** — all three covered.
- **`use-migration-balances-preview` / `use-custodial-wallet-balances`** — added direct specs (15 tests).
- **Unchecking an explainer box disables the CTA** — added (component reverse-report + screen round-trip).
- **Offline key derivation vs SDK** — native-only; device/integration.

## Code Smells

- **Fake-world safety tests** — fixed at the root (loud layer).
- **Storage answers `false` to everything** — `saveJson` fixed at the layer; the keychain setter still returns `false` (call sites fixed) — follow-up to make the setter throw.
- **The choke point isn't one** — fixed (single dispatcher; new entry points route to it).
- **One `null`, three meanings** — guards added; the `use-custodial-owner-id` tri-state + retry is the remaining class fix.
- **Forked footer** — deferred with #24.

esaugomez31:
## Second review pass (review #2)

An independent `/review` over the full branch diff. Net: **9 findings fixed, each with a regression test; on the changed files I measured, coverage held at 100% (stmts/branches/funcs/lines).** Behavioral fixes first, cleanups after — all on this branch.

### Behavioral

**1. Greyed Receive tap was a silent no-op when the self-custodial flag is off.** **Fixed.** In `ReceiveDisabled` with `nonCustodialEnabled` off, tapping the greyed Receive called `reopen()` on the migrate-now prompt, but the modal's visibility is gated by the same flag, so nothing surfaced. The prompt now exposes `canReopen` (`isReceiveDisabled && !isSelfCustodialDisabled`), and the home falls back to `receivingDisabledToast` when the nudge can't reopen — so the tap is never a dead no-op. Test: flag off → tap → toast, not a silent reopen.

**2. Self-custodial dollar prediction read as unrestricted when the IP lookup failed.** **Fixed.** The prediction resolves region by IP (the future account has no phone). When the IP doesn't resolve it now falls back to the session country (`ipCountryCode ?? deviceCountryCode`), so a restricted user isn't previewed a dollar balance the new account can't hold. The IP still wins whenever it resolves. Tests: falls back to the session country when the IP is unresolved; prefers the IP over the session country when both resolve.

**3. History Download could fire against unloaded wallets.** **Fixed.** The Download CTA now waits on the wallet-balances query (`isDownloadDisabled = isBusy || !areWalletsReady`), while Skip stays gated only on `isBusy` — so a still-loading or failed balances query can never trap the user on this optional step. Tests: loading and error both keep Download disabled and Skip available.

**4. An empty transactions CSV surfaced as an error.** **Fixed.** `exportCsv` now distinguishes a missing field (`undefined` → throw, a real failure) from an empty export (`""` → resolve `false`, nothing to share), so a no-transaction export no longer shows the error toast. Test: an empty CSV resolves `false` without opening the share sheet.

**5. An expired checkpoint lent its account id to a fresh save.** **Fixed.** `saveCheckpointToStorage` read the prior record with `validateStoredCheckpoint`, which does not check expiry, so an expired record's `accountId` could be carried into a new save. It now treats an expired prior record as absent (matching `loadCheckpoint`), keeping the 48h expiry authoritative for the id. Test: an expired record's `accountId` is dropped on the next save.

**6. Keep-receiving hung on a spinner when the lightning-address query failed.** **Fixed.** The screen folded the query error into a single `isCheckReady`, so an errored `useAddressScreenQuery` (cache-first, no auto-retry) rendered a spinner with no CTA — reachable via the gate routing here on `addressError` — leaving the OS back button as the only escape mid-flow. Error is now its own state: **Try Again** refetches, and **Continue** advances the migration without the (purely informational) address, so a persistent failure never traps the flow. The skip decision stays gated on "no error", so a user who actually has an address still sees the warning instead of being skipped past. Tests: an error renders the retry state (not a spinner) and does not skip; Try Again refetches; Continue advances without blocking; a still-failing refetch reports and keeps the retry tappable.

### Nits

- **7.** Fixed — extracted `isConfirmDisabled` on the backup-phrase-confirm CTA and flattened the nested exchange-rate ternary on balances-overview to a single `&&` (outer stays one ternary). No behavior change.
- **8.** Fixed — `isMountedRef` on the checkpoint state actually tracks focus (set on `reloadCheckpoint`, cleared on blur), renamed to `isFocusedRef`.
- **9.** Fixed — lifted `Platform.OS !== "ios" && isCredentialBackupAvailable` on the backup-method screen into a named `canShowCredentialBackup`.

## Test Coverage

Each behavioral fix (1–6) landed with a regression test for the new branch; the two nits with a code change (8 = the `isFocusedRef` rename, and the extractions in 7/9) kept their specs green. On the files I measured — `keep-receiving-screen`, `use-dollar-balance-restricted`, `download-history-screen`, `use-export-transactions-csv`, `migration-checkpoint-storage`, `use-migration-checkpoint-state`, `backup-method-screen` — coverage is 100% across statements, branches, functions and lines. Typecheck and lint clean.

grimen (changes requested):
Verified every resolution from the last round against `69da3c9`: all 26 findings genuinely fixed or honestly deferred, with regression tests that fail if the fix regresses. One new finding blocks.

## Findings

**1. A failed checkpoint save permanently bricks the balances screen.**
The balances-overview screen saves the point-of-no-return checkpoint in an effect that runs once, and Approve stays disabled until that save lands (`balances-overview-screen.tsx:59-75`). Two things combine badly: nothing ever runs the save again — after the effect settles, none of its dependencies change for the lifetime of the screen, so the comment's "a later focus re-save heals it" can't happen — and this is the one screen where every exit is deliberately sealed: hardware back is swallowed, gestures are off, there is no header.
*Example:* a user resumes their migration and lands on the balances screen. The owner query errors, or the AsyncStorage write fails — the exact failures this round made loud. The save resolves `false`, Approve stays disabled, and nothing on screen says why. Back does nothing. The user is stuck at the commit point of a funds-moving flow, and the only way out is force-killing the app.
*Proposal:* give the failure a face, the way the gate's Retry already does: show error text when the save fails, with a Retry that re-runs it (or re-run it on refocus for real, by keying the effect on focus). Test: a rejected save renders the retry state, and a successful retry enables Approve.

esaugomez31:
> Verified every resolution from the last round against `69da3c9`: all 26 findings genuinely fixed or honestly deferred, with regression tests that fail if the fix regresses. One new finding blocks.
> ## Findings
> 
> **1. A failed checkpoint save permanently bricks the balances screen.** The balances-overview screen saves the point-of-no-return checkpoint in an effect that runs once, and Approve stays disabled until that save lands (`balances-overview-screen.tsx:59-75`). Two things combine badly: nothing ever runs the save again — after the effect settles, none of its dependencies change for the lifetime of the screen, so the comment's "a later focus re-save heals it" can't happen — and this is the one screen where every exit is deliberately sealed: hardware back is swallowed, gestures are off, there is no header. _Example:_ a user resumes their migration and lands on the balances screen. The owner query errors, or the AsyncStorage write fails — the exact failures this round made loud. The save resolves `false`, Approve stays disabled, and nothing on screen says why. Back does nothing. The user is stuck at the commit point of a funds-moving flow, and the only way out is force-killing the app. _Proposal:_ give the failure a face, the way the gate's Retry already does: show error text when the save fails, with a Retry that re-runs it (or re-run it on refocus for real, by keying the effect on focus). Test: a rejected save renders the retry state, and a successful retry enables Approve.

Done!

grimen (approved):

PR #3864, opened 2026-06-19
feat: add self-custodial Dollar Balance restrictions for EU and HK


esaugomez31:
## Summary

Self-custodial Dollar Balance restrictions, from blinkbitcoin/blink-wip#819. The restriction is driven by per-account-type, remote-configurable country lists, so the policy is tunable without a release.

Part of blinkbitcoin/blink-wip#819. Also enables the El Salvador restriction from blinkbitcoin/blink-wip#864.

## Phase 1 rollout matrix

Full Phase 1 of the Non-Custodial Rollout Plan. `C` = Custodial Mode, `NC` = Non-Custodial Mode.

**Restriction levels**

| Level | Meaning |
|---|---|
| **1** | No restriction |
| **2** | Transfer/swap disabled only (mode and Dollar Balance stay available) |
| **3** | Dollar Balance disabled (Bitcoin Balance stays available) |
| **5** | Fully restricted — new account creation is blocked (Unsupported region screen) |

(Level 4 — migration notification — is Phase 2+ and is not part of this PR.)

| Jurisdiction / region | C | NC | New custodial accounts |
|---|---:|---:|---|
| El Salvador | 1 | 3 | Allow |
| Hong Kong | 3 | 2 | Allow |
| Vietnam | 2 | 1 | Block |
| United States | 1 | 1 | **Block** |
| EU / EEA | 2 | 2 | Block |
| United Kingdom | 2 | 2 | Block |
| Singapore | 2 | 2 | Block |
| South Korea | 2 | 2 | Block |
| Canada | 2 | 2 | Block |
| Japan | 2 | 2 | Block |
| Switzerland | 2 | 2 | Block |
| Indonesia | 2 | 2 | Block |
| Israel | 2 | 2 | Block |
| Philippines | 2 | 2 | Block |
| Thailand | 2 | 2 | Block |
| Bahrain | 2 | 2 | Block |
| South Africa | 2 | 2 | Block |
| UAE | 2 | 2 | Block |
| Cuba | 5 | 5 | Already blocked |
| Iran | 5 | 5 | Already blocked |
| North Korea | 5 | 5 | Already blocked |
| Syria | 5 | 5 | Already blocked |
| Russia | 5 | 5 | Already blocked |
| Belarus | 5 | 5 | Already blocked |
| Rest of world | 1 | 1 | Allow |

## Changes

- **Generic region × account-type restriction hook.** Renamed `useStablesatsRestricted` / `useStablesatsRestrictionGuard` → `useDollarBalanceRestricted` / `useDollarBalanceRestrictionGuard`. An internal policy resolves per account type — custodial reads the Stablesats list/flag, self-custodial the stable-token list/flag. It gates on `accountType` (not `isSelfCustodial`) so the restriction stays stable through the self-custodial cold-start window while the SDK connects. `isBlockedCountry` was deduped into `use-device-location`, and the custodial forced-conversion moved to its own hook.
- **Full restriction (balance hidden).** `stableTokenBlockedCountries` remote-config list (default `["HK"]`) plus a persisted `stableTokenRestricted` anti-bypass flag (keeps the #795 protections: phone removal, account switch, restart, login/logout). A fully restricted account hides the Dollar balance, disables the transfer button, and shows a restriction modal — the same treatment HK custodial already had.
- **Bitcoin-only receive & send when restricted.** Receive locks to Bitcoin (`useReceiveAssetMode`: default Bitcoin, asset toggle disabled, Bitcoin-only rails) and send offers only the Bitcoin wallet (`useSendWallets`). Custodial now gets this 1:1 explicitly (it previously relied on the forced-conversion zeroing the USD balance); self-custodial has no forced-conversion, so it is enforced directly.
- **Allowed without transfer.** `stableTokenTransferBlockedCountries` (default the 27 EU member states) hides the home transfer button and guards the conversion page. The balance and receive flow stay available — lighter than the full restriction.
- **Transfer/swap disabled for custodial too.** `stablesatsTransferBlockedCountries` extends the same transfer block to custodial accounts through a shared `useTransferBlocked` policy (renamed from the self-custodial-only `useStableTokenTransferBlocked`), keyed by `accountType`.
- **New account creation blocked (Unsupported region screen).** `custodialCreationBlockedCountries` / `selfCustodialCreationBlockedCountries` redirect the create flow to a new Unsupported region screen — at the "Create new account" tap when every offered option is blocked, or after account-type selection when the chosen option is blocked, both before terms and conditions.
- **Dropped `custodialSignupBlockedCountries`.** The old always-block (which disabled the button) is superseded by the Unsupported region screen; US now goes through `custodialCreationBlockedCountries`. `custodialFirstSignupBlockedCountries` is kept.
- **Unified restriction modal.** A single `DollarBalanceRestrictionModal` ("Dollar Balance is not available in your region", single **Close** button) serves both custodial and self-custodial restricted accounts.
- **Forced-conversion modal copy.** The forced Dollar-to-Bitcoin conversion modal (`UsdConvertToBtcModal`, shown when a restricted account still holds a Dollar balance) reads "Dollar Balance is not available in your region" / "Transfer from Dollar Balance to Bitcoin Balance" with a **Transfer** button.
- **i18n.** Restriction, conversion, and Unsupported region copy across `en` + 28 locales (real diacritics, each file's indentation preserved); types regenerated via `yarn update-translations`.


esaugomez31:
* **#3864** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

### Code smells

**1. `wallet-overview.tsx` uses `LL.StablesatsRestriction.walletLabel()` for the in-wallet label on all restricted accounts.** Self-custodial HK users get the Stablesats-keyed copy rather than anything under the new `StableTokenRestriction` block. It works today because the copy is the same, but it leaves the UI unable to differentiate per account type without a code change if copy ever diverges. Either add a `StableTokenRestriction.walletLabel` key now (even with the same text) or leave a comment marking the shared key as intentional so a future translator doesn't wonder why the SC label is under `StablesatsRestriction`.

**2. `persist` in the `useDollarBalanceRestrictionSync` effect deps creates an undocumented account-type-switch side effect.** `persist` is a stable module-level function reference that changes only when `accountType` changes. If a user is in a blocked country and switches from custodial to self-custodial mid-session, the effect re-fires with the SC-specific `persist` function and writes `stableTokenRestricted` in addition to the already-written `stablesatsRestrictedCustodial`. This is the correct behavior, but it is entirely undocumented. A one-line comment explaining why `persist` is in the deps (and what double-write it causes on account-type switch) would prevent future reviewers from treating it as an accidental inclusion and removing it.

### Test coverage

**3. No home-screen test covers the HK self-custodial path (dollar-restricted, transfers not blocked).** The existing SC test sets `mockDollarBalanceRestrictedOverride = true` with `mockStableTokenTransferBlockedOverride = false` (default), which means the transfer button IS rendered but disabled. The test only asserts that the `stable-token-restriction-modal` component is mounted in the tree and the forced-conversion modal is absent — it does not assert that the button is present-and-disabled, nor that pressing it opens the SC modal (not the custodial `stablesats-restriction-modal`). A test that fires the button's `onDisabledPress` and confirms the SC modal becomes visible would close this gap.

**4. `useDollarBalanceRestrictionSync` has no test for the account-type-switch scenario.** Given finding 2 above — if a custodial user in HK switches to a self-custodial account, the effect fires again and should write `stableTokenRestricted`. The test suite covers custodial and SC in isolation but never tests transitioning between them in a blocked country. This is exactly the path where a future refactor dropping `persist` from the deps would silently break the anti-bypass protection for the SC flag.

esaugomez31:
@grimen

1. wallet-overview.tsx uses the Stablesats-keyed wallet label for SC. Done. Added a comment marking the shared StablesatsRestriction.walletLabel as intentional across custodial and self-custodial (copy is identical), so a future translator knows why the SC label lives there.

2. persist in the useDollarBalanceRestrictionSync deps. Done. Added a comment explaining its identity flips with accountType, so a custodial-to-self-custodial switch in a blocked country re-fires the effect and writes the SC flag too (anti-bypass double-write).

3. No home-screen test for the HK self-custodial path. Done. Added a test that asserts the transfer button is rendered (transfers not blocked), and pressing it (disabled) opens the self-custodial StableTokenRestrictionModal, not the custodial StablesatsRestrictionModal.

4. No account-type-switch test for useDollarBalanceRestrictionSync. Done. Added a test where a blocked-country user switches custodial to self-custodial mid-session and the effect re-fires to write the stableTokenRestricted flag.

esaugomez31:
@grimen Ready for Review

grimen (approved):

PR #3865, opened 2026-06-20
feat(self-custodial): support switching the Breez network with the active environment


esaugomez31:
## Summary

The self-custodial (Breez Spark) network used to be pinned by a build-time `BREEZ_NETWORK` env var. This binds it to the app's active environment instead, so self-custodial follows the same environment as custodial (Main maps to Mainnet, every other instance maps to Regtest). It also adds two supporting UX changes for working across networks.

Part of blinkbitcoin/blink-wip#823

## Changes

- **Network from the active environment.** `useSparkNetwork()` derives the Breez network from `appConfig.galoyInstance` through pure `networkForInstance` / `networkLabelFor` helpers, and the network is threaded as an explicit parameter through config, the bridge/SDK lifecycle, address parsing, scan context, and the delete/probe paths (no globals). `storageDir` and the persisted per-account network label are byte-identical to the previous env-based values, so wallets created on the old build keep working, and the network-vs-wallet validation is preserved. Token and API key stay env vars.
- **Delete funded regtest accounts.** On regtest, account deletion no longer blocks on a non-zero balance (regtest funds are throwaway); on mainnet the has-funds warning still applies. The profile row also skips the SDK balance probe entirely on regtest.
- **Network-mismatch toast.** When the active self-custodial account was created on a different network than the current environment (a multi-account dev/QA scenario), a toast names the network to switch back to, instead of leaving the wallet on a silent failure. It fires once per activation of the mismatched account (deduped while that account stays active) and re-warns if the account is deselected and reselected. Detection reuses the same `mismatchedNetworkLabel` rule as the SDK connect gate.

## Notes

- **Spark address HRP update.** The send-destination shape gate (`SPARK_ADDRESS_SHAPE_PATTERN`) was corrected from the legacy `sp1`/`sprt1` HRPs to the current `spark1` (mainnet) / `sparkrt1` (regtest) HRPs that the Breez SDK now emits, confirmed against real mainnet and regtest addresses. This is a parsing-only change: the pattern is a cheap pre-filter in front of `sdk.parse` (the source of truth for validity and network), so it only affects how a pasted or scanned destination is recognized, not any stored data. No migration is needed: the legacy form is no longer produced by the current SDK, and self-custodial addresses are persisted as display strings only, never re-validated through this gate. A wrong-network Spark address still surfaces the dedicated "wrong network" error rather than a generic one.
- Custodial behavior is unchanged (self-custodial-only paths).
- The obsolete `BREEZ_NETWORK` CI/build variable is removed in the stacked PR #3867.
- Toast i18n across `en` + 28 locales (real diacritics, each file's indentation preserved); types regenerated via `yarn update-translations`.
- 100% coverage on the new files; typecheck, lint, and the full test suite are green.


grimen (changes requested):

### Code smells

**1. `SPARK_ADDRESS_SHAPE_PATTERN` is a silent breaking change with no mention in the PR description.** The pattern changed from `/^(?:sp1|sprt1)/i` to `/^(?:spark1|sparkrt1)/i`. This is a protocol-level address format change, not a network-wiring refactor: any `sp1`/`sprt1` address a user has stored, copied, or is about to scan will pass through the shape-gate silently and be handed to the SDK. The test explicitly documents the rejection of the old HRPs, which is good, but the PR body makes no mention of the change. This needs a Notes entry explaining whether the old format has been fully retired SDK-side, and whether any migration path is needed for persisted addresses.

**2. `useSelfCustodialNetworkMismatchToast` claims one-time but resets on custodial round-trip.** The doc comment says "one-time toast", but `notifiedKeyRef.current` is cleared to `null` whenever `accountId` becomes `null` (i.e. when the user switches to a custodial account). If the user goes self-custodial → custodial → self-custodial, the toast fires again for the same mismatch. If re-firing is acceptable, remove "one-time" from the comment. If it is not, the ref needs to survive the null-accountId transition (e.g. only reset when the identity of the mismatch key changes, not when the account deselects).

### Test coverage

**3. `use-delete-account.spec.ts` hardcodes Regtest, leaving the mainnet delete path untested.** The mock is fixed to `useSparkNetwork: () => mockSparkNetwork.Regtest`, so every test runs the regtest branch (balance check bypassed). There is no case that exercises `isRegtestNetwork(network) === false` inside the hook, meaning the mainnet code path—where a non-zero balance triggers `DeleteAccountOutcome.HasFunds`—is not covered at the hook level. Given that this is the path that protects mainnet users from accidentally deleting a funded wallet, it should be tested here.

**4. No test for the custodial→self-custodial re-selection edge case in `use-network-mismatch-toast`.** All specs use a fixed `SC_ACCOUNT` for the full lifecycle. The reset-on-null-accountId behavior described in finding 2 is completely untested: there is no case that verifies whether the toast fires once or twice after a custodial round-trip.

esaugomez31:
@grimen

1. SPARK_ADDRESS_SHAPE_PATTERN change. Done. Added a Notes entry to the PR description. It is a parse-only pre-filter in front of sdk.parse; the legacy sp1/sprt1 form is no longer emitted by the SDK and persisted addresses are display-only, so no migration is needed.

2. "one-time" toast comment. Done. Reworded the JSDoc to match the behavior: it fires once per activation and re-warns on reselect. The re-firing is intentional, so I fixed the wording, not the behavior.

3. Mainnet delete path. Done. Added a mainnet case to use-delete-account.spec.ts. The hook itself has no balance check; the funded-delete guard lives in the screens and its mainnet path is already covered in delete-account.spec.tsx and profile-row.spec.tsx.

4. Custodial round-trip test. Done. Added a test in use-network-mismatch-toast.spec.ts that asserts the toast re-warns after a custodial round-trip and reselect.

grimen (approved):

grimen:
### Merge activity

* **Jun 22, 5:23 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3865).
* **Jun 22, 5:24 PM UTC**: @grimen merged this pull request with [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3865).

PR #3867, opened 2026-06-20
ci: remove the obsolete BREEZ_NETWORK build variable


esaugomez31:
## Summary

With #3865 deriving the Breez network from the app's active environment, the `BREEZ_NETWORK` build variable is no longer read by the app. This removes it from the CI/build configuration and the docs.

Part of blinkbitcoin/blink-wip#823

## Changes

- `ci/values.yml`: removed `breez_network: mainnet`.
- `ci/pipeline.yml`: removed `BREEZ_NETWORK` from both build params blocks.
- `ci/tasks/build.sh`: removed `BREEZ_NETWORK` from the `react-native-config` env file heredoc. The release `BuildConfig` survival assertion derives its key list from that same file, so it stops checking the variable automatically.
- `android/app/proguard-rules.pro`: comment only; the generic `-keep class com.galoyapp.BuildConfig { *; }` rule is unchanged.
- `docs/android-env-config-r8.md`: removed the sample references.

## Notes

- Stacked on #3865, which makes the variable obsolete.
- The repo now has zero `BREEZ_NETWORK` references (the `dev/vendor/galoy-quickstart` `NETWORK=regtest` is the galoy backend's own variable, unrelated).
- CI/build-only change; no app code is affected.

PR #3868, opened 2026-06-20
feat(self-custodial): register a Blink-domain lightning address via the Breez SDK

esaugomez31:
## Summary

Self-custodial (Breez Spark) wallets had no way to claim a Lightning address, so they were unreachable by `username@domain`. This lets a self-custodial user register a Blink-domain Lightning address (`username@blink.sv` on mainnet, `username@staging.blink.sv` on regtest) from Settings, using the same username rules and the same set-address modal as custodial. Registration goes through the Breez SDK against Blink's LNURL server; no Blink GraphQL mutation is involved.

## Changes

- **Register via the Breez SDK.** `config.lnurlDomain` is set to the Blink LNURL host (derived from the active network) before connect, so the SDK signs the chosen username with the Spark wallet key and registers it against Blink's LNURL server. A new `useRegisterLightningAddress` hook validates the username (reusing the custodial `validateUsername` rules), checks availability, registers, and refreshes the active account. There is no default address: the user opts in from Settings.
- **Adapter pattern over a shared UI.** The existing set-address modal UI is now pure (it takes `hostname` and `bankName` as props instead of reading config directly), with two thin containers over the same UI: custodial keeps the GraphQL `userUpdateUsername` path, self-custodial uses the SDK register path. The Settings row likewise extracts one shared presentational `LightningAddressRow` used by both account types; the self-custodial row is now set-or-show instead of hidden when empty.
- **Account list display.** The self-custodial account row shows the username without the `@domain` suffix, matching the custodial presentation one to one.
- **Network aware.** Staging works today; mainnet picks up the production host automatically once Blink deploys its LNURL server, with no app change.

## Notes

- Custodial behavior is unchanged (the custodial container renders the same modal with the same hostname and bank name it read internally before, and keeps the GraphQL flow).
- No new i18n; this reuses the existing set-address and Settings keys.
- 100% coverage on the new files; typecheck, lint, and the full test suite are green.
- Stacked on #3867 (which is stacked on #3865).

### DIFF
--- a/__mocks__/@breeztech/breez-sdk-spark-react-native.js
+++ b/__mocks__/@breeztech/breez-sdk-spark-react-native.js
@@ -83,6 +83,7 @@ module.exports = {
     IncreasedToAvoidDust: "IncreasedToAvoidDust",
   },
   PrepareSendPaymentRequest: { create: (p) => p },
+  RegisterLightningAddressRequest: { create: (p) => p },
   SendPaymentRequest: { create: (p) => p },
   SyncWalletRequest: { create: (p) => p },
   ReceivePaymentRequest: { create: (p) => p },
--- a/__tests__/screens/settings-screen/self-custodial/profile-row.spec.tsx
+++ b/__tests__/screens/settings-screen/self-custodial/profile-row.spec.tsx
@@ -227,12 +227,15 @@ describe("ProfileRow", () => {
     })
   })
 
-  it("renders the lightning address as the row title when one is set", () => {
-    const { getByText } = render(
-      <ProfileRow entry={{ id: TEST_ENTRY_ID, lightningAddress: "alice@example.com" }} />,
+  it("renders the lightning username without the domain as the row title when one is set", () => {
+    const { getByText, queryByText } = render(
+      <ProfileRow
+        entry={{ id: TEST_ENTRY_ID, lightningAddress: "alice@lnurl.staging.blink.sv" }}
+      />,
     )
 
-    expect(getByText("alice@example.com")).toBeTruthy()
+    expect(getByText("alice")).toBeTruthy()
+    expect(queryByText("alice@lnurl.staging.blink.sv")).toBeNull()
   })
 
   it("falls back to anonymous user label when no lightning address is set", () => {
@@ -560,7 +563,7 @@ describe("ProfileRow", () => {
       <ProfileRow entry={{ id: TEST_ENTRY_ID, lightningAddress: "stale@example.com" }} />,
     )
 
-    expect(getByText("magentamouse1845@breez.tips")).toBeTruthy()
+    expect(getByText("magentamouse1845")).toBeTruthy()
   })
 
   it("ignores the live lightning address for inactive rows and uses the persisted entry value", () => {
@@ -575,6 +578,6 @@ describe("ProfileRow", () => {
       />,
     )
 
-    expect(getByText("stored@example.com")).toBeTruthy()
+    expect(getByText("stored")).toBeTruthy()
   })
 })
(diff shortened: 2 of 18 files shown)


grimen (changes requested):

### Test coverage

**1. `liveLightningAddress ?? persistedLightningAddress` fallback is untested.** `account-ln-address.spec` covers the live-address-present and no-address cases, but nothing asserts that the persisted registry entry is shown when `liveLightningAddress` is `null`. The fallback could be silently removed without a test failure.

**2. Concurrent `register()` calls are unguarded and untested.** The hook sets `loading = true` but never gates on it, so calling `register()` twice before the first resolves fires two parallel SDK calls. Either add a guard (`if (loading) return`) or a test that documents the current behavior.

### Code smells

**3. `description: undefined` passed explicitly.** `sdk.registerLightningAddress({ username, description: undefined })` — an absent key and an `undefined` key are identical at runtime. Drop the field or supply a real value.

**4. `liveLightningAddress ?? persistedLightningAddress` in `SelfCustodialLightningAddressRow` is undocumented.** The fallback is load-bearing resilience (shows stored address while the SDK connects), but looks like a redundant expression. A one-line comment would prevent it from being "simplified" away.

designsats:
@grimen @esaugomez31 
- How will we prevent someone creating infinite usernames and squat them? In custodial mode its prevented by phone number requirement. 
- I thought we give user random username and ability to create custom aliases as per https://github.com/blinkbitcoin/blink-wip/issues/22 let me know if its out of scope here

esaugomez31:
@grimen

1. liveLightningAddress ?? persistedLightningAddress fallback untested. Already covered. account-ln-address.spec.ts has "shows the persisted address (not the set prompt) while the live address is still resolving" (live null plus persisted set asserts the persisted address is shown); removing the fallback fails it.

2. Concurrent register() unguarded. Done. Wrapped register() in useInFlightGuard (same pattern as use-create-wallet / use-restore-wallet), so a second call while one is in flight no-ops. Added a concurrency test.

3. description: undefined passed explicitly. Done. Now uses the SDK factory RegisterLightningAddressRequest.create({ username }), which drops the explicit field (description is the only optional field there).

4. Fallback undocumented. Done. Extracted const address with a JSDoc explaining it falls back to the persisted address while the SDK reconnects.

grimen (dismissed):

grimen:
### Merge activity

* **Jun 22, 5:23 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3868).
* **Jun 22, 5:25 PM UTC**: [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3868) couldn't merge this pull request because a downstack PR #3867 failed to merge.
* **Jun 22, 5:34 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3868).
* **Jun 22, 5:36 PM UTC**: [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3868) rebased this pull request as part of a merge.
* **Jun 22, 5:36 PM UTC**: [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3868) couldn't merge this PR because it failed for an unknown reason (At least 1 approving review is required by reviewers with write access).
* **Jun 22, 5:47 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3868).
* **Jun 22, 5:48 PM UTC**: @grimen merged this pull request with [Graphite](https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3868).

grimen (dismissed):

grimen (approved):

PR #3873, opened 2026-06-24
fix(self-custodial): resolve Blink username sends via their Lightning address


esaugomez31:
## Summary

Sending to a Blink Lightning address was rejected in two cases, both fixed here (part of blinkbitcoin/blink-wip#879):

- A **self-custodial** account paying a Blink username: a full address like `bulus@blink.sv` already worked, but a bare `bulus` (or any username that does not map to a custodial account) was rejected.
- A **custodial** account paying a **self-custodial** recipient's Blink Lightning address.

Both now resolve over LNURL so the payment goes through.

## Root cause

A Blink username/address is classified as an `Intraledger` destination and resolved through a custodial mutation. A self-custodial sender has no token for that mutation, and a custodial sender to a self-custodial recipient hits a username that has no custodial account. In both cases the intraledger lookup returns `UsernameDoesNotExist` and the send was never retried over LNURL.

## Fix

- **Self-custodial sender** (`resolve-username.ts`): recovers the handle from an invalid intraledger result scoped strictly to `UsernameDoesNotExist` and re-resolves `handle@<lnAddressHostname>` over LNURL. Any other invalid reason (for example `WrongDomain`) passes through unchanged, so a non-Blink URL is never rewritten into a Blink address.
- **Custodial sender** (`resolve-destination.ts`): when the intraledger lookup fails with `UsernameDoesNotExist`, re-parses the destination with the new `preferLnurlForInternalHandles` flag (added in `@blinkbitcoin/blink-client` 0.6.0, which this PR bumps to) so the Blink address resolves over LNURL. Custodial-to-custodial sends keep using intraledger; the fallback only triggers when the username is not a custodial account.

### DIFF
--- a/__tests__/payment-destination/resolve-destination.integration.spec.ts
+++ b/__tests__/payment-destination/resolve-destination.integration.spec.ts
@@ -89,6 +89,28 @@ describe("resolveDestination (integration: real parser)", () => {
     expect(mockWrapDestination).toHaveBeenCalled()
   })
 
+  it("routes a custodial send to a non-custodial Blink username through LNURL via the flag fallback", async () => {
+    // No custodial wallet for the handle: the recipient is self-custodial, not custodial.
+    const accountDefaultWalletQuery = jest.fn().mockResolvedValue({
+      data: { accountDefaultWallet: null },
+    })
+
+    const result = await resolveDestination(
+      {
+        rawInput: "esaudeveloper@blink.sv",
+        myWalletIds: ["my-wallet-id"],
+        bitcoinNetwork: Network.Mainnet,
+        lnurlDomains: [lnAddressHostname],
+        accountDefaultWalletQuery: accountDefaultWalletQuery as never,
+      },
+      { sdk: null, network: mockSparkNetwork.Regtest },
+      lnAddressHostname,
+    )
+
+    expect(result.valid).toBe(true)
+    expect(result.valid && result.validDestination.paymentType).toBe(PaymentType.Lnurl)
+  })
+
   it("resolves a matching-network Spark address to a valid Spark destination", async () => {
     mockParseSparkAddress.mockResolvedValue({
       address: "sparkrt1qabcdefghijklmn",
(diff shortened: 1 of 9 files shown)


esaugomez31:
* **#3873** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3873?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3873?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):

### Test coverage

**1. No test for double-failure in the custodial LNURL fallback.** `resolve-destination.ts` retries with `preferLnurlForInternalHandles: true` when the first `parseDestination` returns `UsernameDoesNotExist`, but there is no test for the case where the second call also fails. The caller presumably handles an invalid result, but the behavior is unverified.

**2. Integration test relies on a hardcoded external address.** The new integration test in `resolve-destination.integration.spec.ts` uses `esaudeveloper@blink.sv` as the recipient. If that account changes account type or is deleted, the test fails with no code change as the cause. The handle should be injected from a fixture or the test should mock at a lower level.

### Code smells

**3. `invalidPaymentDestination` is accessed without a null guard.** In `resolve-username.ts`, `parsedPaymentDestination` returns `result.invalidPaymentDestination` when the reason is `UsernameDoesNotExist`, but does not guard against the field being absent. If `blink-client` ever emits that reason without populating `invalidPaymentDestination`, `intraledgerHandle` silently returns `undefined` and the LNURL re-resolution is skipped with no error surfaced. The blink-client 0.6.0 type contract should be verified, or an explicit guard added.

**4. Two network round-trips for custodial-to-self-custodial sends.** The custodial fallback in `resolve-destination.ts` always calls `parseDestination` twice for any self-custodial recipient — once to discover `UsernameDoesNotExist`, then again with the flag. This is a latency regression on this path. Worth documenting with a comment at the call site so the intent is clear and a future reader does not collapse it into one call.

grimen (approved):

PR #3875, opened 2026-06-25
fix: ignore failed location fallback for dollar restrictions

openoms:
## Summary

Fixes a false-positive self-custodial Dollar account restriction when primary country detection falls back to `SV`.

`useDeviceLocation` can return `SV` with `detectionFailed=true` when phone/IP country detection fails. If Remote Config blocks `SV` for `selfCustodialDollarBalanceBlockedCountries`, the app could treat that fallback as a real blocked country and persist `stableTokenRestricted`, making the Dollar account remain unavailable even after switching regions.

This change ignores failed primary country fallbacks for Dollar balance restriction decisions, while still allowing the secondary IP lookup to persist the restriction when the real IP country is blocked.

## Changes

- Treat `detectionFailed=true` primary country values as non-blocking.
- Keep secondary IP lookup active for phone-source detection failures.
- Allow `stableTokenRestricted` persistence from a blocked IP result.
- Keep existing persisted restriction flags honored.
- Add regression coverage for:
  - failed `SV` fallback not blocking by itself
  - existing persisted stable-token restriction still blocking
  - failed phone fallback with blocked IP country still persisting the restriction

## Testing

- `direnv exec . yarn prettier --check app/hooks/use-dollar-balance-restricted.ts __tests__/hooks/use-dollar-balance-restricted.spec.ts`
- `direnv exec . yarn tsc:check`
- `direnv exec . ./node_modules/.bin/eslint app/hooks/use-dollar-balance-restricted.ts __tests__/hooks/use-dollar-balance-restricted.spec.ts --ext .js,.ts,.tsx`
- `direnv exec . yarn test __tests__/hooks/use-dollar-balance-restricted.spec.ts --runInBand`
- `direnv exec . yarn check:translations`
- `direnv exec . yarn check:codegen`
- `direnv exec . yarn graphql-check`
- `direnv exec . bash -lc 'git ls-files -z "*.js" "*.ts" "*.tsx" | xargs -0 ./node_modules/.bin/eslint'`

Results:
- Focused Jest passed: 22 tests.
- Prettier, TypeScript, focused ESLint, translations, codegen, and GraphQL checks passed.
- Tracked-file ESLint completed with 0 errors and existing warnings only.

### DIFF
--- a/__tests__/hooks/use-dollar-balance-restricted.spec.ts
+++ b/__tests__/hooks/use-dollar-balance-restricted.spec.ts
@@ -213,6 +213,22 @@ describe("useDollarBalanceRestrictionSync", () => {
       expect(mockUpdateState).toHaveBeenCalledTimes(1)
     })
 
+    it("persists the stable-token flag when phone fallback fails but IP is blocked", () => {
+      mockUseDeviceLocation.mockReturnValue({
+        countryCode: "SV",
+        source: "phone",
+        detectionFailed: true,
+      })
+      mockUseIpCountryCode.mockReturnValue("FR")
+
+      renderHook(() => useDollarBalanceRestrictionSync())
+
+      expect(mockUseIpCountryCode).toHaveBeenCalledWith(true)
+      expect(mockUpdateState).toHaveBeenCalledTimes(1)
+      const updater = mockUpdateState.mock.calls[0][0]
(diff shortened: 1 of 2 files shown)


grimen:
We do not want to skip country detection since compliance is key - instead more solid IP geolocation lookup fallbacks addressed in https://github.com/blinkbitcoin/blink-mobile/pull/3877.

PR #3880, opened 2026-06-25
feat(restrictions): add remote switch to disable cached account restrictions


esaugomez31:
## What

Adds a single Firebase Remote Config switch, `dollarRestrictionCacheEnabled` (default `true`), that turns off the caching/persistence of all account restrictions: Dollar Balance and Transfer block, for both custodial (Stablesats) and self-custodial (Stable Token).

## Why

When the IP geolocation lookup is rate-limited, device detection falls back to `SV`. If `SV` happens to be in a blocked list, the restriction gets computed, **persisted**, and then sticks even after detection recovers, leaving the user wrongly restricted with no way out from the client.

This is a remote kill-switch: flipping it to `false` makes the app ignore the persisted restriction so a falsely-stuck user is released, without shipping a new build.

## How

Both restriction reads gate the persisted flag behind the switch, and both sync hooks stop writing when it is off:

- read: `(dollarRestrictionCacheEnabled && isPersisted) || isBlockedCountry(country, blockedCountries)`
- sync: only persists (and only consults IP) when the switch is on.

Live country detection is **not** touched: a user genuinely in a blocked country stays restricted regardless of the switch. The flag only governs the cached/persisted part.

## Scope

Covers every cached restriction, all read through the gated hooks (no direct reads of the persisted flags):

| Restriction | Custodial | Self-custodial |
|---|---|---|
| Dollar Balance | `stablesatsRestrictedCustodial` | `stableTokenRestricted` |
| Transfer block | `stablesatsTransferBlocked` | `stableTokenTransferBlocked` |

The Transfer button visibility and the Dollar Balance gating both flow through these hooks, so both respond to the switch. Creation/region blocks are live-only (no persistence), so they are intentionally out of scope.

## Notes

- Default `true` preserves current behavior exactly: with the flag on the logic reduces to the previous code, so existing users see no change until the switch is flipped remotely.
- The `dollarRestrictionCacheEnabled` key must be created in Firebase Remote Config (boolean). Until it exists, it falls back to the `true` default.

## Tests

Added cache-disabled coverage for both hooks (read and sync) and updated the existing specs. Full suite green.

### DIFF
--- a/__tests__/hooks/use-dollar-balance-restricted.spec.ts
+++ b/__tests__/hooks/use-dollar-balance-restricted.spec.ts
@@ -49,6 +49,7 @@ const baseState: PersistentState = {
 const remoteConfig = {
   custodialDollarBalanceBlockedCountries: ["HK"],
   selfCustodialDollarBalanceBlockedCountries: ["FR"],
+  dollarRestrictionCacheEnabled: true,
 }
 
 const setup = (accountType: AccountType): void => {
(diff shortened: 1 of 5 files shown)


esaugomez31:
* **#3880** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (approved):
LGTM - per DMs earlier

PR #3890, opened 2026-06-27
feat(backup): add warning Card component and backup safety messaging

### LINKED ISSUE #3887: Update restore backup phrase warning copy
## Context
@designsats implemented the suggested copy changes and asked for a ticket to track them.

## Requested copy
- Icon: `warning.svg`
- If phrase is shown: `Do not show this to anyone including support`
- Restore flow title: `Do you recognize this backup phrase?`
- Restore flow body: `Never enter words someone else sent you.`

## Frontend changes needed
Warning card component has to be updated:
- [ ] color variable `warning/light`: `#E18E02`
- [ ] color variable `warning/dark`: `#FFC563`
- [ ] add new color variable `grey7/dark`: `#0F0F0F`
- [ ] add new color variable `grey7/light`: `#F9F9F9`

Reasoning:
- `grey5` zones should always be active.
- Warning color is too close to the primary accent color in dark mode. The accent color should stay easily identifiable as link/action.

Reference comment: https://github.com/blinkbitcoin/blink-mobile/issues/3887#issuecomment-4817709874

## Acceptance criteria
- Restore flow uses `warning.svg` for the warning state.
- The phrase warning uses the exact copy above.
- The restore prompt title/body use the exact copy above.
- Warning card color variables are updated as listed above.
- Android and iOS are checked for parity, if both surfaces are affected.

designsats:
## Summary

Ticket: https://github.com/blinkbitcoin/blink-mobile/issues/3887

Adds a reusable `WarningCard` component and new backup-safety messaging across the self-custodial onboarding/backup/restore flows. The goal is to make the seed-phrase screens clearer about social-engineering risk ("never show this to anyone, including support" and "never enter words someone else sent you").

> ### Stacked on #3853
>
> This branch is rebased **onto `ios-liquid-glass-disabled` (#3853)**, not onto plain `main`.
> GitHub will not let a PR base point at a branch on a fork, so the base here stays `main` and
> the diff currently carries #3853's two commits as well. **Review the last 6 commits**
> (`feat(backup): add the card component…` onward). Once #3853 merges, rebasing this branch
> onto `main` drops them and the diff shrinks to this PR's own work.
>
> **Why it is stacked:** this PR adds two new `navigation.setOptions({ headerRight })` sites
> (backup phrase, view backup phrase) and rewrites the restore Paste button. #3853 exists
> precisely to route every custom header button through `headerRightNoGlass`, so landing this
> PR on plain `main` would have reverted the restore button's opt-out and shipped two more
> header buttons still wrapped in the iOS 26 Liquid Glass capsule. All three now use
> `headerRightNoGlass`.

## Acceptance requirements

- The backup, view-backup and restore screens each carry a warning card that is legible in both light and dark themes.
- The warning card wraps its content: long titles wrap rather than overflowing the card padding, and the title is vertically aligned with the warning icon.
- The Spark-compatible wallet URL is tappable on **both** the onboarding backup screen and the view-backup screen, with identical behaviour.
- The anti-phishing warning on the restore screen is visible while the user is typing words, not pushed below the keyboard.
- Copy and paste header actions meet a usable touch-target size and announce their visible label to screen readers.
- The view-backup Copy action is unreachable until biometric authentication succeeds, and the phrase never leaves the screen before the gate resolves.
- No behavioural change to any screen outside the self-custodial backup/restore flows.

### `WarningCard` (`app/components/warning-card`)

- Always bordered: 1px warning border plus a warning icon, so the card never depends on its fill alone to be visible.
- Optional `title` (bold heading) and body children.
- Surface follows static vs. active: `grey7` when static, `grey5` with a press/scale animation when `onPress` is provided.
- Title typography matches Figma `Paragraph 2/Bold` — Source Sans Pro, weight 700, 16px, 22px line height.

### `SparkCompatibleInfo` (`app/components/spark-compatible-info`)

Shared component rendering the "works in any Spark-compatible wallet" notice with the wallet-list URL as a real `accessibilityRole="link"`. Used by both the backup and view-backup screens, so the copy and the link behaviour cannot drift apart again.

### Theme

- New `grey7` token (light `#F9F9F9`, dark `#0F0F0F`) for the static card surface, added to `colors.ts` and `themed.d.ts`.
- `warning` token retuned: light `#F59E0B` → `#E18E02`, dark `#F59E0B` → `#FFC563`.

> **Note for reviewers — this `warning` change is app-wide.** The token is read by roughly 20 call sites outside this feature (warning banners, send-bitcoin confirmation, KYC banners, `light.warning` in `galoy-toast.tsx`), so it is a global visual change riding inside a backup-messaging PR. Dark mode measures 12.3:1 on the card surface. Light mode improves from 2.04:1 to 2.47:1 but still sits below the 4.5:1 WCAG threshold that applies to the 16px bold card title; reaching it would need roughly `#8A5600`. Keeping the current values is a deliberate design decision.

### Screens

- **Backup phrase screen** — warning card: *"Do not show this to anyone, including support"*, plus the shared Spark-compatible info banner.
- **View backup phrase screen** — the same "do not share" warning card and info banner.
- **Restore phrase screen** — on step 1, warning card: *"Do you recognize this backup phrase? / Never enter words someone else sent you."*, positioned **above** the word inputs.

### Accessibility

- `hitSlop` of 12pt on the copy/paste header buttons, whose `clear` variant otherwise has a hit area no larger than its text bounds.
- Explicit `accessibilityLabel` on those buttons, since `testProps` overwrites the accessible name with the test id.

### i18n

Three English source keys, propagated across all locales:

- `BackupScreen.ManualBackup.Phrase.doNotShareWarning`
- `RestoreScreen.recognizePhraseTitle`
- `RestoreScreen.recognizePhraseBody`

`i18n-types.ts` regenerated via `yarn update-translations`.

## Review feedback addressed

| # | Feedback | Resolution |
| --- | --- | --- |
| 1 | *Blocking:* the Spark-compatible link stopped being tappable on the onboarding backup screen — it rendered as plain text while the sibling screen kept a real link | Extracted `SparkCompatibleInfo`, used by both screens. Removed the now-dead `handleOpenLink` / `handleOpenSparkLink` hook members. Added a test that fails against the plain-text rendering |
| 2 | *Blocking:* the borderless `default` card was invisible against the screen background (1.05:1 light, 1.10:1 dark) | Removed the `default` variant. The card is always bordered, so the shape no longer depends on the fill. `grey7` is retained as the static surface, matching the Figma spec |
| 3 | The `warning` token change is app-wide and light mode still falls short of 4.5:1 | Documented above. Values kept deliberately |
| 4 | The anti-phishing warning sat last in the scroll content, off-screen behind the keyboard | Moved above the word inputs |
| 5 | `Card` overlapped `InfoCard`, and `app/components/card/` collided with the `card-screen/` product area | Renamed to `WarningCard` in `app/components/warning-card/`. It is now a distinct, always-bordered primitive rather than a general card |
| 6 | Most of the card API was unused | Dropped `type` / `CardType` and the `default` branch |
| 7 | Header touch targets were smaller than 44pt/48dp | `hitSlop` on all three buttons |
| 8 | `testProps` overwrote the accessible name | `accessibilityLabel` applied after the spread |
| 9 | Copy nit: comma before "including" | Fixed in the English source; the translations already punctuated it that way |
| — | Card content was not vertically centred and the card did not wrap its content | Root cause was `fontFamily: "SourceSansPro-Bold"` combined with the theme's `fontWeight`, which makes Android synthesise the weight so glyphs draw wider than they measure. Switched to the `"Source Sans Pro"` + `700` pairing already used by `InfoCard`, and disabled `includeFontPadding` |
| 10 | *kngako:* `app/components/warning-card/index.ts` and `app/components/spark-compatible-info/index.ts` "look like empty files" | They were one-line `export *` barrels, which GitHub renders as a near-empty diff. Both now name their export (`export { WarningCard } from "./warning-card"`), matching every other barrel in `app/components` |
| 11 | Regression against #3853 (found while rebasing): the two new header buttons and the rewritten restore Paste button bypassed the Liquid Glass opt-out | All three header buttons now go through `headerRightNoGlass`. See the stacking note at the top |
| 12 | *Blocking:* the header Copy button escaped the biometric gate on the view-backup screen | The `useLayoutEffect` installing `headerRight` ran above the `if (!authenticated)` early return, so the button was mounted with the full mnemonic while the prompt was still pending. It is now installed only once `authenticated` is true and cleared otherwise, via a new `noHeaderRight` that resets both `headerRight` (Android) and `unstable_headerRightItems` (iOS). Three specs added; two fail against the ungated effect |

## Testing specs

Automated (`yarn test`, 555 suites / 6326 tests passing):

- `__tests__/components/warning-card/warning-card.spec.tsx` — title/body rendering, the warning icon in title-only, body-only and combined cases, and press behaviour.
- `__tests__/components/spark-compatible-info/spark-compatible-info.spec.tsx` — renders the link, opens the remote-config URL on press, exposes `accessibilityRole="link"`.
- `__tests__/screens/.../backup-phrase-screen.spec.tsx` — asserts the Spark-compatible link is present **and tappable**. This is the regression guard for finding 1: it fails against the previous plain-text rendering.
- `__tests__/screens/.../view-backup-phrase-screen.spec.tsx`, `restore-phrase-screen.spec.tsx` — warning-card placement on each screen.
- `__tests__/screens/.../view-backup-phrase-screen.spec.tsx` — the biometrics-**enabled** path: no header Copy button while the prompt is pending, none when auth fails, one that copies once auth succeeds. The first two are the revert-failing guards for the gate escape.

Manual:

1. Onboarding → backup phrase: confirm the warning card is legible in light **and** dark, that the title wraps instead of running past the padding, and that the icon lines up with the first line of the title.
2. Tap the Spark-compatible URL on the backup screen and on the view-backup screen — both should open the wallet list.
3. Restore → step 1: with the keyboard open, confirm the "never enter words someone else sent you" card is still on screen.
4. Check the copy/paste header buttons are comfortably tappable, and that a screen reader announces "Copy"/"Paste" rather than the test id.
5. Enable biometric unlock, open Settings → view backup phrase, and confirm no Copy action is present in the header until the prompt is answered; cancel it and confirm the screen leaves without ever showing one.
6. Spot-check a screen that uses the `warning` token outside this flow (e.g. a warning banner or the send-bitcoin confirmation) for the retuned colour.

## Translations for human reviewers

Please review the machine-generated translations below for your language.

| Locale | `doNotShareWarning` | `recognizePhraseTitle` | `recognizePhraseBody` |
| --- | --- | --- | --- |
| en | Do not show this to anyone, including support | Do you recognize this backup phrase? | Never enter words someone else sent you. |
| af | Moenie dit vir enigiemand wys nie, ook nie ondersteuning nie | Herken jy hierdie rugsteunfrase? | Moet nooit woorde invoer wat iemand anders aan jou gestuur het nie. |
| ar | لا تُظهر هذا لأي شخص بما في ذلك الدعم | هل تتعرّف على عبارة النسخ الاحتياطي هذه؟ | لا تُدخل أبدًا كلمات أرسلها لك شخص آخر. |
| ca | No ho mostris a ningú, ni tan sols a l'assistència | Reconeixes aquesta frase de còpia de seguretat? | No introdueixis mai paraules que t'hagi enviat algú altre. |
| cs | Neukazujte to nikomu, ani podpoře | Poznáváte tuto zálohovací frázi? | Nikdy nezadávejte slova, která vám poslal někdo jiný. |
| da | Vis ikke dette til nogen, heller ikke support | Genkender du denne backup-sætning? | Indtast aldrig ord, som en anden har sendt dig. |
| de | Zeige dies niemandem, auch nicht dem Support | Erkennst du diese Backup-Phrase? | Gib niemals Wörter ein, die dir jemand anderes geschickt hat. |
| el | Μην το δείξετε σε κανέναν, ούτε στην υποστήριξη | Αναγνωρίζετε αυτή τη φράση αντιγράφου ασφαλείας; | Μην εισάγετε ποτέ λέξεις που σας έστειλε κάποιος άλλος. |
| es | No se lo muestre a nadie, ni siquiera al soporte | ¿Reconoce esta frase de respaldo? | Nunca introduzca palabras que otra persona le haya enviado. |
| fr | Ne montrez ceci à personne, y compris au support | Reconnaissez-vous cette phrase de sauvegarde ? | N'entrez jamais des mots qu'une autre personne vous a envoyés. |
| hr | Nemojte ovo pokazivati nikome, uključujući podršku | Prepoznajete li ovu frazu za sigurnosnu kopiju? | Nikada ne unosite riječi koje vam je poslao netko drugi. |
| hu | Ne mutasd meg senkinek, az ügyfélszolgálatnak sem | Felismered ezt a biztonsági mentési kifejezést? | Soha ne írj be olyan szavakat, amelyeket valaki más küldött neked. |
| hy | Մի՛ ցույց տվեք սա որևէ մեկին, այդ թվում՝ աջակցման ծառայությանը | Ճանաչո՞ւմ եք այս պահուստային արտահայտությունը: | Երբեք մի՛ մուտքագրեք բառեր, որ ուրիշն է ուղարկել ձեզ: |
| id | Jangan tunjukkan ini kepada siapa pun, termasuk tim dukungan | Apakah Anda mengenali frasa cadangan ini? | Jangan pernah memasukkan kata-kata yang dikirim orang lain kepada Anda. |
| it | Non mostrarlo a nessuno, nemmeno all'assistenza | Riconosci questa frase di backup? | Non inserire mai parole che ti ha inviato qualcun altro. |
| ja | サポートを含め、誰にも見せないでください | このバックアップフレーズに見覚えがありますか？ | 他人から送られた単語は絶対に入力しないでください。 |
| lg | Tokiraga muntu yenna, n'abawagizi ba support | Omanyi ebigambo bino eby'okukuuma? | Toteekangamu bigambo omuntu omulala by'akuweereza. |
| ms | Jangan tunjukkan ini kepada sesiapa termasuk sokongan | Adakah anda mengenali frasa sandaran ini? | Jangan sekali-kali masukkan perkataan yang dihantar oleh orang lain. |
| nl | Laat dit aan niemand zien, ook niet aan support | Herken je deze back-upzin? | Voer nooit woorden in die iemand anders je heeft gestuurd. |
| pt | Não mostre isto a ninguém, incluindo o suporte | Você reconhece esta frase de backup? | Nunca insira palavras que outra pessoa lhe enviou. |
| qu | Aman pimanpas rikuchiychu, yanapaqkunamanpas | ¿Kay waqaychana rimaykunata reqsinkichu? | Ama hayk'aqpas hukkuna apachisqan simikunata churaychu. |
| ro | Nu arăta acest lucru nimănui, inclusiv asistenței | Recunoști această frază de backup? | Nu introduce niciodată cuvinte trimise de altcineva. |
| sk | Neukazujte to nikomu, ani podpore | Spoznávate túto zálohovaciu frázu? | Nikdy nezadávajte slová, ktoré vám poslal niekto iný. |
| sr | Не показујте ово никоме, укључујући подршку | Да ли препознајете ову фразу за резервну копију? | Никада не уносите речи које вам је послао неко други. |
| sw | Usimwonyeshe mtu yeyote ikiwa ni pamoja na timu ya usaidizi | Je, unaitambua kaulisiri hii ya nakala rudufu? | Kamwe usiweke maneno uliyotumiwa na mtu mwingine. |
| th | อย่าแสดงสิ่งนี้ให้ใครเห็น รวมถึงฝ่ายสนับสนุน | คุณจำวลีสำรองข้อมูลนี้ได้ไหม | อย่าป้อนคำที่คนอื่นส่งมาให้คุณเด็ดขาด |
| tr | Bunu destek dahil hiç kimseye gösterme | Bu yedekleme ifadesini tanıyor musun? | Başka birinin sana gönderdiği kelimeleri asla girme. |
| vi | Không cho bất kỳ ai xem điều này, kể cả bộ phận hỗ trợ | Bạn có nhận ra cụm từ sao lưu này không? | Không bao giờ nhập các từ mà người khác gửi cho bạn. |
| xh | Musa ukuyibonisa nakubani na, kuquka nenkxaso | Ngaba uyayazi le binzana yokhuselo? | Ungaze ufake amagama akuthunyelwe ngomnye umntu. |


designsats:
Fixing tests. @grimen 

This is how its supposed to look like: 
image

designsats:
@blink-claw-bot review

designsats:
@grimen Need review

kngako:
Yo, @designsats. Seems the Card content isn't vertically center aligned on the backup screen (or at least on the same level/height as the warning icon).  It would look correct if the message took two lines but it's only 1 line of text. Best make the card wrap around content.

image

kngako:
Claude also has a more designed oriented review for you. But I think the regression of the "spark-compatible" link is definitely a blocker. 

We should probably also have a test for that.
 


<h2>Blocking</h2>
<p><strong>1. The spark-compatible link stopped being tappable on the onboarding backup screen</strong></p>
<p><a href="https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx">backup-phrase-screen.tsx</a> dropped <code>handleOpenLink</code> from the destructure and now renders the whole interpolated string as plain text inside <code>InfoBanner</code>. The URL text still displays but does nothing when tapped. The sibling screen <a href="https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:63">view-backup-phrase-screen.tsx:63</a> still splits the string and renders a real <code>accessibilityRole="link"</code> span — identical copy, two different behaviors.</p>
<p><code>50458bf67</code>'s own message frames this as a test fix ("the standalone link assertion no longer matched"), so the assertion was rewritten to match the new behavior rather than catching it. Nothing now covers link tappability on that screen.</p>
<p>Dead code left behind: <code>handleOpenLink</code> at <a href="https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/hooks/use-backup-phrase.ts:53">use-backup-phrase.ts:53</a> has zero consumers (grep-confirmed), along with its <code>openExternalUrl</code> import and <code>sparkCompatibleWalletsUrl</code> read. Both screens carry an identical split-and-render block — extracting one shared component would fix the regression and the duplication together.</p>
<p><strong>2. <code>grey7</code> is invisible against the screen background in both themes</strong></p>
<p><code>Screen</code> paints <code>colors.white</code>, which is <code>#FFFFFF</code> in light and <code>#000000</code> in dark (<a href="https://claude.ai/epitaxy/app/components/screen/screen.tsx:23">screen.tsx:23</a>, <a href="https://claude.ai/epitaxy/app/rne-theme/colors.ts:72">colors.ts:72</a>). The new static-card surface:</p>

theme | card | screen bg | contrast
-- | -- | -- | --
light | #F9F9F9 | #FFFFFF | 1.05:1
dark | #0F0F0F | #000000 | 1.10:1


<p>For <code>type="warning"</code> the 1px border carries the shape. But <code>type</code> defaults to <code>"default"</code>, which has no border — so the first non-warning <code>&lt;Card&gt;</code> renders an invisible box. Either drop <code>grey7</code> and use <code>grey5</code> (what <code>InfoCard</code> already does), or pick a value with real separation.</p>
<h2>Should fix</h2>
<p><strong>3. The <code>warning</code> token change is app-wide, and light mode still fails contrast</strong></p>
<p><code>#F59E0B</code> → <code>#E18E02</code> (light) / <code>#FFC563</code> (dark) touches ~20 existing call sites outside this feature — warning banners, send-bitcoin confirmation, KYC banners, <code>light.warning</code> in <a href="https://claude.ai/epitaxy/app/components/galoy-toast/galoy-toast.tsx:21">galoy-toast.tsx:21</a>. None re-checked.</p>
<p>The light change is an improvement (2.04:1 → 2.47:1 on <code>#F9F9F9</code>) but still short: the Card title is <code>p2</code> bold ≈16px, below WCAG's large-text threshold of 18.66px bold, so 4.5:1 applies — you'd need roughly <code>#8A5600</code>. Dark is fine at 12.3:1. Either way this belongs in the PR description; it's a global visual change riding inside a backup-messaging PR.</p>
<p><strong>4. The anti-phishing warning sits where it's least visible</strong></p>
<p><a href="https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx:185">restore-phrase-screen.tsx:185</a> puts "Never enter words someone else sent you" as the <em>last</em> child of the scrollable content — after the subtitle, six word inputs, and error row. The keyboard is always up on this screen, so it's off-screen exactly while the user types words someone may have sent them. Both backup screens put their card at the top. Move it above <code>styles.inputList</code>.</p>
<p><strong>5. <code>Card</code> duplicates <code>InfoCard</code></strong></p>
<p><a href="https://claude.ai/epitaxy/app/components/card-screen/info-card.tsx">info-card.tsx</a> already renders warning icon + warning-colored bold title + body with the same <code>borderRadius: 8</code>, <code>padding: 14</code>, <code>grey5</code> background, and <code>GaloyIcon name="warning" size={16}</code>. It has a spec and existing consumers. The new <code>Card</code> adds a border and a press animation. With <a href="https://claude.ai/epitaxy/app/components/warning-banner/warning-banner.tsx">warning-banner.tsx</a> that's three overlapping ways to render a warning. A <code>bordered</code> prop on <code>InfoCard</code> would likely have done it — at minimum the PR should say why not.</p>
<p>Naming also collides: <code>app/components/card/</code> sits beside <code>app/components/card-screen/</code>, which is the debit-card product area, not a card primitive.</p>
<h2>Minor</h2>
<p><strong>6. Most of <code>Card</code>'s API is unused.</strong> All three consumers pass <code>type="warning"</code> and nothing else. Never used in production: <code>onPress</code> and its whole branch (<code>scaleAnim</code>, <code>breatheIn</code>, <code>breatheOut</code>, <code>Pressable</code>, <code>Animated.View</code>, <code>styles.active</code> — ~35 lines), plus <code>numberOfLines</code>, <code>style</code>, and <code>type="default"</code>. The press path exists only in the spec. <code>a3e145a04</code> fixed a real bug there — <code>onPress</code> fired from <code>breatheOut</code>, so it triggered on drag-off cancels too — but it's fixing code nothing calls.</p>
<p><strong>7. Tiny header touch targets.</strong> <code>GaloyTertiaryButton</code> with <code>clear</code> sets <code>paddingHorizontal: 0, paddingVertical: 0</code> at <code>fontSize: 14</code>, so the hit area is the text bounds — under 44pt/48dp. Pre-existing on the restore paste button, now propagated to two more screens. One <code>hitSlop={{top: 12, bottom: 12, left: 12, right: 12}}</code> on each fixes all three. (Side effect worth noting: restore's paste label went 16px → 14px.)</p>
<p><strong>8. <code>testProps</code> overwrites the accessible name.</strong> <a href="https://claude.ai/epitaxy/app/utils/testProps.ts:8">testProps.ts:8</a> sets <code>accessibilityLabel: testID</code>, so the new header buttons announce "backup-phrase-copy" instead of "Copy". Repo-wide pre-existing pattern, but these are brand-new controls — <code>accessibilityLabel={copyLabel}</code> after the spread is nearly free.</p>
<p><strong>9. Copy nit:</strong> "Do not show this to anyone including support" reads better with a comma before "including". The 28 translations already punctuate it that way (fr: "…à personne, y compris au support").</p>
<h2>What's solid</h2>
<p>All 28 locales updated with genuine translations — spot-checked fr/de/es/ja/ar/sw, none are English fallbacks — and <code>i18n-types.ts</code> is correctly machine-regenerated. Test coverage is real: the new component spec, both card placements, and the step-2 negative case on restore.</p></body></html>I'll start by looking at what's actually in these 8 commits.

Reviewed the 4 substantive commits (`0fd888b19`, `50458bf67`, `7bd4cca80`, `a3e145a04`) against the branch point `1f97ed827`. Verification: `tsc -p .` clean, `eslint --no-ignore` clean on all changed paths, and the 50 relevant jest suites (628 tests) pass.

## Blocking

**1. The spark-compatible link stopped being tappable on the onboarding backup screen**

[[backup-phrase-screen.tsx](https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx)](app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx) dropped `handleOpenLink` from the destructure and now renders the whole interpolated string as plain text inside `InfoBanner`. The URL text still displays but does nothing when tapped. The sibling screen [[view-backup-phrase-screen.tsx:63](https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:63)](app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:63) still splits the string and renders a real `accessibilityRole="link"` span — identical copy, two different behaviors.

`50458bf67`'s own message frames this as a test fix ("the standalone link assertion no longer matched"), so the assertion was rewritten to match the new behavior rather than catching it. Nothing now covers link tappability on that screen.

Dead code left behind: `handleOpenLink` at [[use-backup-phrase.ts:53](https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/hooks/use-backup-phrase.ts:53)](app/screens/self-custodial/onboarding/hooks/use-backup-phrase.ts:53) has zero consumers (grep-confirmed), along with its `openExternalUrl` import and `sparkCompatibleWalletsUrl` read. Both screens carry an identical split-and-render block — extracting one shared component would fix the regression and the duplication together.

**2. `grey7` is invisible against the screen background in both themes**

`Screen` paints `colors.white`, which is `#FFFFFF` in light and `#000000` in dark ([[screen.tsx:23](https://claude.ai/epitaxy/app/components/screen/screen.tsx:23)](app/components/screen/screen.tsx:23), [[colors.ts:72](https://claude.ai/epitaxy/app/rne-theme/colors.ts:72)](app/rne-theme/colors.ts:72)). The new static-card surface:

| theme | card | screen bg | contrast |
|---|---|---|---|
| light | `#F9F9F9` | `#FFFFFF` | 1.05:1 |
| dark | `#0F0F0F` | `#000000` | 1.10:1 |

For `type="warning"` the 1px border carries the shape. But `type` defaults to `"default"`, which has no border — so the first non-warning `<Card>` renders an invisible box. Either drop `grey7` and use `grey5` (what `InfoCard` already does), or pick a value with real separation.

## Should fix

**3. The `warning` token change is app-wide, and light mode still fails contrast**

`#F59E0B` → `#E18E02` (light) / `#FFC563` (dark) touches ~20 existing call sites outside this feature — warning banners, send-bitcoin confirmation, KYC banners, `light.warning` in [[galoy-toast.tsx:21](https://claude.ai/epitaxy/app/components/galoy-toast/galoy-toast.tsx:21)](app/components/galoy-toast/galoy-toast.tsx:21). None re-checked.

The light change is an improvement (2.04:1 → 2.47:1 on `#F9F9F9`) but still short: the Card title is `p2` bold ≈16px, below WCAG's large-text threshold of 18.66px bold, so 4.5:1 applies — you'd need roughly `#8A5600`. Dark is fine at 12.3:1. Either way this belongs in the PR description; it's a global visual change riding inside a backup-messaging PR.

**4. The anti-phishing warning sits where it's least visible**

[[restore-phrase-screen.tsx:185](https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx:185)](app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx:185) puts "Never enter words someone else sent you" as the *last* child of the scrollable content — after the subtitle, six word inputs, and error row. The keyboard is always up on this screen, so it's off-screen exactly while the user types words someone may have sent them. Both backup screens put their card at the top. Move it above `styles.inputList`.

**5. `Card` duplicates `InfoCard`**

[[info-card.tsx](https://claude.ai/epitaxy/app/components/card-screen/info-card.tsx)](app/components/card-screen/info-card.tsx) already renders warning icon + warning-colored bold title + body with the same `borderRadius: 8`, `padding: 14`, `grey5` background, and `GaloyIcon name="warning" size={16}`. It has a spec and existing consumers. The new `Card` adds a border and a press animation. With [[warning-banner.tsx](https://claude.ai/epitaxy/app/components/warning-banner/warning-banner.tsx)](app/components/warning-banner/warning-banner.tsx) that's three overlapping ways to render a warning. A `bordered` prop on `InfoCard` would likely have done it — at minimum the PR should say why not.

Naming also collides: `app/components/card/` sits beside `app/components/card-screen/`, which is the debit-card product area, not a card primitive.

## Minor

**6. Most of `Card`'s API is unused.** All three consumers pass `type="warning"` and nothing else. Never used in production: `onPress` and its whole branch (`scaleAnim`, `breatheIn`, `breatheOut`, `Pressable`, `Animated.View`, `styles.active` — ~35 lines), plus `numberOfLines`, `style`, and `type="default"`. The press path exists only in the spec. `a3e145a04` fixed a real bug there — `onPress` fired from `breatheOut`, so it triggered on drag-off cancels too — but it's fixing code nothing calls.

**7. Tiny header touch targets.** `GaloyTertiaryButton` with `clear` sets `paddingHorizontal: 0, paddingVertical: 0` at `fontSize: 14`, so the hit area is the text bounds — under 44pt/48dp. Pre-existing on the restore paste button, now propagated to two more screens. One `hitSlop={{top: 12, bottom: 12, left: 12, right: 12}}` on each fixes all three. (Side effect worth noting: restore's paste label went 16px → 14px.)

**8. `testProps` overwrites the accessible name.** [[testProps.ts:8](https://claude.ai/epitaxy/app/utils/testProps.ts:8)](app/utils/testProps.ts:8) sets `accessibilityLabel: testID`, so the new header buttons announce "backup-phrase-copy" instead of "Copy". Repo-wide pre-existing pattern, but these are brand-new controls — `accessibilityLabel={copyLabel}` after the spread is nearly free.

**9. Copy nit:** "Do not show this to anyone including support" reads better with a comma before "including". The 28 translations already punctuate it that way (fr: "…à personne, y compris au support").

## What's solid

All 28 locales updated with genuine translations — spot-checked fr/de/es/ja/ar/sw, none are English fallbacks — and `i18n-types.ts` is correctly machine-regenerated. Test coverage is real: the new component spec, both card placements, and the step-2 negative case on restore.

designsats:
@kngako  "2. grey7 is invisible against the screen background in both themes" - this is intentional. Im transitioning from using grey5 for all card and info components and using it only for active (tapable) elements. Its not an issue to resolve but a feature to apply app-wide.

designsats:
@kngako thanks — all of it is addressed. Force-pushed: the branch is rebased onto current `main` and the history is rewritten, so the diff is easiest to read fresh rather than incrementally.

**Both blockers**

1. **Spark-compatible link.** Confirmed — `backup-phrase-screen.tsx` had dropped `handleOpenLink` and rendered the whole interpolated string as plain text, while `view-backup-phrase-screen.tsx` still split it into a real link. Both screens now render a shared `SparkCompatibleInfo` component, so the copy and the behaviour cannot drift apart again. The dead `handleOpenLink` / `handleOpenSparkLink` hook members are gone. You asked for a test: `backup-phrase-screen.spec.tsx` now presses the link and asserts the URL opens, which fails against the previous plain-text rendering.

2. **Invisible card surface.** Resolved by removing the `default` variant rather than changing the fill. `WarningCard` is now always bordered, so its shape never depends on the surface. `grey7` stays as the static surface — it matches the Figma spec (`#0F0F0F` on the dark card), and the reason the borderless variant was the real problem is that it had no border to carry the shape. That variant was never used in production; all three call sites were `type="warning"`.

**Layout report.** Your screenshot was the visible symptom of a font bug: `card.tsx` was the only place in the app naming a weighted face (`fontFamily: "SourceSansPro-Bold"`) while also inheriting `fontWeight: "600"` from the theme's `bold` prop. Android synthesises the weight on top of the already-bold face, so glyphs draw wider than they measure — long titles ran past the padding instead of wrapping, and `includeFontPadding` left the card taller than its content with the title sitting above the icon. It now uses the `"Source Sans Pro"` + `700` + 22px pairing that `InfoCard` already uses, which is also exactly Figma's `Paragraph 2/Bold`.

**Also done:** anti-phishing warning moved above the word inputs on restore (it was behind the keyboard), `hitSlop` and real `accessibilityLabel`s on the copy/paste header buttons, comma added to the English warning copy, and `Card` renamed to `WarningCard` in `app/components/warning-card/` so it no longer collides with the `card-screen/` product area.

**One thing left deliberately as-is:** the `warning` token retune. It is app-wide (~20 call sites) and light mode is still 2.47:1, short of 4.5:1 for the 16px bold title. That is a conscious design decision, and it is now documented in the PR description as you asked rather than left implicit.

Green locally against current `main`: `tsc:check`, `eslint:check`, `check:translations`, `check:codegen`, and `yarn test` at 555 suites / 6326 tests.

app/components/warning-card/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./warning-card"
```
kngako: Is this basicallly an empty file?

app/components/spark-compatible-info/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./spark-compatible-info"
```
kngako: Another empty file?

kngako:
UI seems consistent.

image

image

image

kngako:
The LLM Review follows:

---


## 1. The header Copy button es

capes the biometric gate — confirmed regression

[[view-backup-phrase-screen.tsx:47](https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:47)](app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:47)

The `useLayoutEffect` that installs `headerRight` runs unconditionally, before the `if (!authenticated)` early return at [[:63](https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:63)](app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:63). Previously the copy affordance was an `IconTextButton` inside the authenticated JSX branch, so it could not be reached before auth. Moving it to the header took it outside the gate.

I confirmed this with a throwaway spec (since deleted): with `getIsBiometricsEnabled → true` and `authenticate` left pending, the body renders no words, but `setOptions` has already installed `headerRight`, and pressing it calls `copyToClipboard` with the full mnemonic.

The window is real but bounded — `useBiometricGate` awaits a keychain read *and* a sensor check ([[use-biometric-gate.ts:33-40](https://claude.ai/epitaxy/app/screens/card-screen/hooks/use-biometric-gate.ts:33)](app/screens/card-screen/hooks/use-biometric-gate.ts:33)) before the native prompt covers the app, and `onFailure → goBack()` leaves another moment. `useWalletMnemonic` loads the phrase eagerly regardless of the gate, so `words` is populated throughout.

```tsx
useLayoutEffect(() => {
  if (!authenticated) {
    navigation.setOptions({ headerRight: undefined })
    return
  }
  navigation.setOptions({ headerRight: () => (/* ... */) })
}, [navigation, authenticated, copyLabel, handleCopy, styles])
```

No spec covers the biometrics-enabled path — `beforeEach` in [[view-backup-phrase-screen.spec.tsx:69](https://claude.ai/epitaxy/__tests__/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.spec.tsx:69)](__tests__/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.spec.tsx:69) mocks it to `false`, so the gate always passes instantly. Worth adding the enabled case alongside the fix.

## 2. `colors.warning` is changed app-wide, not just for these cards

[[colors.ts:52,106](https://claude.ai/epitaxy/app/rne-theme/colors.ts:52)](app/rne-theme/colors.ts:52) — light `#F59E0B → #E18E02`, dark `#F59E0B → #FFC563`. That token has ~25 call sites outside this flow: send-bitcoin confirmation borders and text, card KYC banner states, migration screens, `temporarily-unavailable`, `payment-offline-notice`, `set-lightning-address`. The dark shift in particular (saturated amber → pale yellow) is a large visual change none of this branch's tests or screens exercise. Probably a deliberate design-token update, but it deserves explicit sign-off rather than riding in on a warning-card branch.

## 3. Light-mode warning title fails WCAG AA

`#E18E02` on the `grey7` card surface `#F9F9F9` is **~2.5:1**. The title renders at 16px/700 ([[warning-card.tsx:41-43](https://claude.ai/epitaxy/app/components/warning-card/warning-card.tsx:41)](app/components/warning-card/warning-card.tsx:41)), which doesn't qualify as WCAG "large text" (that needs ≥18.66px bold), so the bar is 4.5:1. Dark mode is fine at ~12.3:1.

The branch already improved this (`#F59E0B` was ~2.0:1), but this is the app's most safety-critical copy and it's the *entire* warning string, not an accent. Consider keeping the icon in `colors.warning` and giving the title the default text color, which is how `InfoCard` reads.

## 4. `WarningCard`'s pressable variant is unreachable

`onPress`, `numberOfLines`, and `style` have no production call site — all three screens pass only `title` (+ children on restore). That makes the `Animated`/`Easing`/`Pressable` breathe animation, the `grey5` active surface, and the `scaleAnim` ref about 35 lines of dead machinery, exercised only by the spec. Commit `8bc87673b` already dropped the unused "default" variant on exactly this reasoning; the same argument applies here.

## 5. Pre-existing, but you're editing the line: restore header goes stale on step 2

[[restore-phrase-screen.tsx:75](https://claude.ai/epitaxy/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx:75)](app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx:75) early-returns when `!isStep1` without clearing `headerRight`. Since `handleContinue` calls `navigate("selfCustodialRestorePhrase", { step: Second })` on the already-focused route, params merge in place and the instance stays mounted — so the Paste button survives into step 2. Not introduced here, and nothing in the restore spec touches the paste button at all, but it's the same pattern as finding 1 and a one-line fix while you're in there.

## 6. Smaller things

- `HEADER_BUTTON_HIT_SLOP` is duplicated verbatim, comment and all, in three files. It's a property of `GaloyTertiaryButton clear` having zero padding — better placed in that component.
- [[warning-card.spec.tsx:8-36](https://claude.ai/epitaxy/__tests__/components/warning-card/warning-card.spec.tsx:8)](__tests__/components/warning-card/warning-card.spec.tsx:8) mocks `makeStyles` to return empty objects, so the static-vs-active surface split (`grey7` / `grey5`) the commit message emphasizes is untested. And `expect(toJSON()).toBeTruthy()` at [[:91](https://claude.ai/epitaxy/__tests__/components/warning-card/warning-card.spec.tsx:91)](__tests__/components/warning-card/warning-card.spec.tsx:91) asserts nothing beyond "render didn't throw".
- `grey7` sits outside the existing scale ordering in both themes (lighter than `grey5` in light, darker in dark). The scale is already loose — the file's own comments say "grey8-ish" — and the `// static card surface` note mitigates it, but a semantic name would age better.
- The restore Paste label goes 16px/700 → 14px/bold via the tertiary button. Assuming intentional consistency; just flagging it's a visual delta, not a pure refactor.

## What's solid

`SparkCompatibleInfo` genuinely deduplicates the split-and-render block and fixes the onboarding screen's link being inert plain text — the new spec fails against the old rendering. The dead `handleOpenLink` / `handleOpenSparkLink` hook members are properly removed rather than left behind, `app/components/card` is fully deleted after the rename with no dangling references, and `IconTextButton` / `InfoBanner` are still used elsewhere so nothing was orphaned. Ordering `accessibilityLabel` after the `testProps` spread to reclaim the accessible name from the test id is the right call, and the translations are complete and real.

designsats:
@kngako thanks — force-pushed. Two structural things plus your blocker.

### The branch is now stacked on #3853

Your review didn't name it, but this PR was regressing that one. Moving the copy/paste actions into the headers added two new `navigation.setOptions({ headerRight })` sites and rewrote the restore Paste button, all with plain `headerRight`. #3853 exists precisely to route every custom header button through `headerRightNoGlass`, so landing this on plain `main` would have reverted the restore button's opt-out and shipped two more buttons still wrapped in the iOS 26 Liquid Glass capsule.

So this branch is rebased onto `ios-liquid-glass-disabled`, not onto `main`. GitHub won't let a PR base point at a fork branch, so the base stays `main` and the diff carries #3853's two commits — **review the last 7 commits**. All three header buttons now use `headerRightNoGlass`.

#3853 itself was ~50 commits stale, so I rebased it onto current `main` too. That surfaced a real break: `cardOnboardingAcknowledgementScreen` was added to `root-navigator.tsx` after #3853 branched and still called the `headerCloseControl` export that PR removes. Fixed there.

### 1. Header Copy button escaping the biometric gate — confirmed, fixed

You were right, and your suggested shape is what landed. The `useLayoutEffect` now returns early with `noHeaderRight` unless `authenticated`. Clearing needs both keys, since `headerRightNoGlass` writes `headerRight` for Android and `unstable_headerRightItems` for iOS, so I put the reset next to the setter rather than open-coding `headerRight: undefined` at the call site.

Three specs added on the biometrics-**enabled** path you flagged as uncovered: no header while the prompt is pending, none when auth fails, one that copies once auth succeeds. I verified the first two fail against the ungated effect before restoring the fix.

### 6. The `index.ts` files aren't empty

They were one-line `export *` barrels, which GitHub renders as a near-empty diff. Both now name their export, matching every other barrel in `app/components`.

### Still open, deliberately

- **2 — `colors.warning` is app-wide.** Agreed it deserves explicit sign-off rather than riding in here. Happy to split it into its own PR if you'd prefer that; say the word.
- **3 — light-mode title contrast.** ~2.5:1, below the 4.5:1 bar for 16px bold. Noted in the body as a design decision; your suggestion of icon-in-`warning` + default text colour is the cleanest fix if we want AA.
- **4 — the pressable `WarningCard` variant.** Fair, it's unused. Left in for now.
- **5 — restore header stale on step 2.** Pre-existing and untouched by this PR; I'd rather not widen the diff further, but it's a one-liner if you want it here.
- **6 — `HEADER_BUTTON_HIT_SLOP` duplicated three times**, and the `makeStyles` mock hiding the `grey7`/`grey5` split.

designsats:
@kngako Clanker response. Im free of any slopsposibility.

grimen (changes requested):
## Findings

**1. Restore Paste button persists into step 2.**
Step 1 → step 2 goes through `navigation.navigate("selfCustodialRestorePhrase", …)`, which updates the focused route's params instead of pushing a new screen, so when `isStep1` flips the header effect re-runs but its `if (!isStep1) return` guard leaves the previously installed Paste button mounted. The guard predates this PR, but the PR rewrites this effect and now ships `noHeaderRight` for exactly this situation.
*Example:* the user fills step 1 and taps Continue; step 2 still shows Paste in the header even though the guard reads as step-1-only intent.
*Proposal:* if Paste is not meant for step 2, one option is `navigation.setOptions(noHeaderRight)` in the `!isStep1` branch; if it is meant to stay, dropping the guard with a short comment would make that explicit.

## Test Coverage Gaps

- No spec asserts the header buttons' accessible name is "Copy"/"Paste" rather than the testID, so reverting the post-spread `accessibilityLabel` (review item 8) passes all tests; `hitSlop` is likewise unasserted.
- The rewritten restore header Paste button has no screen-level test: `restore-phrase-screen.spec.tsx` mocks `useRestorePhrase` and never renders or presses the header button, so dropping its `onPress` wiring would fail nothing — Paste is only covered at the hook level.

grimen:
Pushed my review fixes and rebased the branch onto current `main` to clear the merge conflict (it collided with #4088, which hardened these same screens — both sides' changes are kept: the param guards and their specs from `main`, this PR's header work on top). The branch still carries #3853's two commits per the stacking note.

The two new commits:

- 1b487d53f `fix(restore): clear the header Paste action on step 2` — the `!isStep1` branch now sets `noHeaderRight` instead of returning early, so the Paste button installed during step 1 is cleared when the same mounted screen advances to step 2 (previously it stayed in the header for the whole second half of seed entry). The step-1→2 transition spec and the step-2-only spec both fail against the previous early return.
- 18cc1d7c3 `test(backup): assert the header Copy buttons' accessible name and hit slop` — closes the coverage gaps: reverting the post-spread `accessibilityLabel` or dropping `hitSlop` now fails a spec on each of the three header buttons, and the restore header Paste button gained a screen-level press test.

Simulator demo: step 1 shows Paste in the header, and after Continue the step-2 header is clean. (Scaffolding: booted straight into restore step 1 with a prefilled test vector so Continue is enabled; that TEMP edit is not part of the pushed commits.)

after

<sub>Demo lives on branch `assets/pr-3890-demo` — deleting it breaks it.</sub>

grimen (approved):

PR #3703, opened 2026-03-10
refactor(icons): modernize icon system with Phosphor icons

designsats:
# Icon system modernization — Phosphor icons
Replaces the legacy custom icon set with [Phosphor Icons](https://phosphoricons.com/) (phosphor-react-native) across the app, establishing a consistent, maintained icon library as the foundation going forward.

### What changed

Visual change: 
https://github.com/user-attachments/assets/0defbee7-3bed-4d42-b005-8c1d54f8b4ec

### Icon system

- Migrated all icons in GaloyIcon from the old custom set to Phosphor equivalents
- Replaced specific icons with better-fit alternatives:

  - transfer → ArrowsDownUp (vertical swap, more intuitive)
  - pencil → PencilSimpleLine (lighter, cleaner stroke)
  - link (external) → ArrowSquareOut (clearer intent for external links)
  - scan (home screen) → Scan (Phosphor native scan icon)

- Added new Phosphor icons: FingerprintSimple, ArrowsDownUp, PencilSimpleLine, ArrowSquareOut

**Security & privacy screen**
- Added left icons to all three rows: fingerprint (biometric), lock (PIN), eye-slash (hide balance)
- Renamed screen title from "Security and Privacy" → "Security and privacy" to match design

**Settings rows**
- Replaced link icon with arrow-square-out on POS, static QR, and API access rows

PR #3898, opened 2026-07-02
fix(transaction-detail): make tx hash copy & mempool link buttons tappable

### LINKED ISSUE #3732: Transaction detail: copy button and Mempool link not working
## Bug Report

**Description:**

**Steps to reproduce:**
1. Open transaction history

**Screenshot:**

- Type: OnChain (received)
- Status: PENDING
- Account: Bitcoin

**Platform:** iOS

Image

designsats:
## Summary

Fixes #3732 — on the transaction detail screen, the copy and external-link (Mempool / Lightning decoder) icons next to the Transaction Hash and other hash rows did not respond to taps.

## Root cause

The icon actions used `TouchableWithoutFeedback` (from `react-native`), which works by cloning its single child and injecting touch-responder props into it. #3703 swapped the child from `Icon` (a `Text` subclass that forwards unknown props to a native view) to `GaloyIcon`, which destructures a fixed prop list and drops everything else — so the injected responder props never reached a native view and the icons went dead on both platforms.

## Fix

- Replace the `TouchableWithoutFeedback` + wrapping `View` icon actions with a small `IconAction` wrapper around `GaloyIconButton` — the repo's standard tappable icon (already used for the close button). It is `Pressable`-based, so taps register correctly inside the gesture-handler `ScrollView`, and it adds pressed-state opacity feedback, `testProps` IDs and disabled support for free.
- This collapses the 8 duplicated `Pressable` + `hitSlop` + `GaloyIcon` blocks (~60 net lines removed) and removes now-dead styles.
- Add an RNTL regression test asserting the icon action fires its handler when tapped.

## Testing

- `yarn jest __tests__/screens/transaction-detail-*` — green (new `IconAction` press test + existing `typeDisplay` spec).
- `yarn tsc:check` / `eslint` — clean.
- Built and installed the debug variant on an Android emulator; opened OnChain / Lightning transaction details and confirmed the copy and external-link icons respond to taps.
- TODO before merge: verify on iOS (the bug was reported there) and attach before/after screenshots of the row icon alignment.


designsats:
Fixed hitslop 
image

designsats:
@blink-claw-bot review please

designsats:
@blink-claw-bot pushed a fix

designsats:
@blink-claw-bot review please

designsats:
@grimen Should be good to go

openoms:
Review of head 837dd91.

## Real strengths first

- The fix is correct and complete within the file: all 8 icon actions (tx hash explorer/copy, description, LN hash, preimage, payment request decoder/copy, internal id) were converted to `Pressable`, not just the two reported in #3732 ([transaction-detail-screen.tsx:360-574](https://github.com/blinkbitcoin/blink-mobile/blob/837dd917243f9c768982aa20f430c01cd0975005/app/screens/transaction-detail-screen/transaction-detail-screen.tsx#L360-L574)).
- Swept the rest of the app: every other `TouchableWithoutFeedback` (11 files) wraps a plain `View`, which forwards responder props fine, so this screen was the only broken site. No follow-up sweep needed.
- Adjacent hitSlops are exactly abutting, not overlapping (`gap: 10` vs `left: 5 + right: 5`), so the explorer and copy targets can't steal each other's taps.
- `description || ""` to `description ?? ""` and the other nullish tidy-ups are harmless and slightly more correct.
- No mode branches added, no generated files touched, custodial path untouched.

## Important

**I1 - The stated root cause is wrong, and it matters for prevention.**
The description says `TouchableWithoutFeedback` inside a gesture-handler `ScrollView` "doesn't reliably register taps". But the RNGH `ScrollView` has wrapped these rows since Nov 2023 (`ac567804a`) and the buttons worked until #3703. What #3703 actually changed was the child: it swapped `Icon` from react-native-vector-icons (a `Text` subclass that forwards props to a native view) for `GaloyIcon`. `TouchableWithoutFeedback` works by cloning its single child and injecting responder props into it, and `GaloyIcon` destructures a fixed prop list and drops everything else ([galoy-icon.tsx:281-293](https://github.com/blinkbitcoin/blink-mobile/blob/837dd917243f9c768982aa20f430c01cd0975005/app/components/atomic/galoy-icon/galoy-icon.tsx#L281-L293)), so the responder props never reached a native view. Taps were dead on both platforms for that reason alone. The fix is still valid (`Pressable` is its own responder host), but since PRs squash-merge, this incorrect diagnosis becomes permanent history and invites the same bug elsewhere (nothing stops the next `TouchableWithoutFeedback` + custom-component pairing).
**Recommendation:** rewrite the Root cause section of the description to state the real mechanism (GaloyIcon dropping the responder props injected by `TouchableWithoutFeedback`, regressed in #3703's icon swap) before merge.

**I2 - Prose doesn't match the diff (two claims).**
- "Also corrects the payment-request copy handler that was referencing the wrong value": the net diff vs main shows no change there; main already copied `initiationVia?.paymentRequest ?? ""` with the right label. (The apparent change in the commit view is diff-alignment noise from the reflowed JSX.)
- "Add an `iconPressable` style plus `alignSelf`/`marginVertical` tweaks on `valueIcons` so each touch target fills 100% of the input row height": the final code has no `alignSelf`/`marginVertical` on `valueIcons` ([transaction-detail-screen.tsx:631-640](https://github.com/blinkbitcoin/blink-mobile/blob/837dd917243f9c768982aa20f430c01cd0975005/app/screens/transaction-detail-screen/transaction-detail-screen.tsx#L631-L640)); vertical reach comes only from `hitSlop`. The description describes an intermediate revision, not what's being merged.
**Recommendation:** drop the payment-request claim and update the styling paragraph to describe the final approach (`iconPressable` centering + `hitSlop`, `gap` on `valueIcons`).

**I3 - No test pins the fix, and a genuine one is possible here.**
The existing spec only covers the pure `typeDisplay` function ([transaction-detail-screen.spec.ts:6](https://github.com/blinkbitcoin/blink-mobile/blob/837dd917243f9c768982aa20f430c01cd0975005/__tests__/screens/transaction-detail-screen.spec.ts#L6)). A real regression test exists for this bug: on main, the handlers injected by `TouchableWithoutFeedback` are dropped by `GaloyIcon`, so nothing in the rendered tree carries a press handler; with `Pressable`, it does.
**Recommendation:** add an RNTL test that renders the icon action (or the screen), does `fireEvent.press(getByTestId("icon-copy-paste"))` (GaloyIcon already sets `icon-${name}` testIDs) and asserts the clipboard mock was called. It fails against the pre-PR code and passes after, which is exactly the "reverting the fix goes red" bar.

**I4 - Reuse `GaloyIconButton` instead of 8 hand-rolled copies.**
The repo's standard tappable icon, already imported in this file for the close button, is `Pressable`-based (so it equally fixes the bug) and gives for free what the raw blocks lack: pressed-state opacity feedback (#3732 explicitly complains there is no visual response), automatic `testProps` IDs for e2e, disabled support, and built-in hitSlop that an explicit `hitSlop` prop can still override (it is spread after the default, [galoy-icon-button.tsx:116-120](https://github.com/blinkbitcoin/blink-mobile/blob/837dd917243f9c768982aa20f430c01cd0975005/app/components/atomic/galoy-icon-button/galoy-icon-button.tsx#L116-L120)). The current diff repeats an identical ~20-line `Pressable` + `hitSlop` + `GaloyIcon` block 8 times, past the DRY threshold applied in past reviews (extract before the 6th copy, #3818).
**Recommendation:** replace each block with `GaloyIconButton` (or one small local `IconAction` wrapper around it taking `{icon, onPress}`), which collapses ~150 lines and makes the I3 test trivial to target.

## Notes (non-gating)

**N1 - The vertical hitSlop is likely clipped on Android.** RN's documented behavior is that hitSlop never extends past the parent view's bounds, and the parent `valueIcons` row is only icon-height (~22px, centered in the 60px field). So `top: 16, bottom: 16` mostly doesn't materialize on Android and the "fills 100% of the row height" goal isn't met. If that target size is wanted, give `valueIcons` `alignSelf: "stretch"` and let the pressables fill the row height instead of relying on hitSlop.

**N2 - No screenshots.** Icon alignment visibly changed (margins removed, `gap` added, icons flush right). Please add before/after captures as for any UI change.

**N3 - Test plan is Android-only; the bug was reported on iOS** (#3732). Please verify on iOS and say so in the test plan. Related: the reporter noted pending-state buttons give no disabled indication; fine to leave unaddressed, but worth a line in the description if intentional.

**N4 - Stray commit in branch history.** `f103fe684` (`feat(default-balance)`) sits in this branch but its content already landed on main via another squash, so the effective PR diff is clean. Harmless under squash-merge; rebasing it away would tidy the commit list.

**N5 - Bare `Linking.openURL`** in `viewInExplorer` / `viewInLightningDecoder` predates this PR, but since the PR touches these call paths: the repo convention is `openExternalUrl` with catch/fallback (#3899). Fine as a follow-up.

**CI:** the two failed checks (Android/iOS e2e, both 24h timeouts) are the known-flaky Detox runners; no detox spec exercises the transaction-detail icons, so they are not relevant failures and should not gate this.

## Suggested landing order

1. I4 first (swap to `GaloyIconButton` / extracted component) since it subsumes most of the diff.
2. I3's regression test against it.
3. I1 + I2 description fixes before merge, since the squash commit inherits the description.

Nothing here requires re-architecting the change; the core fix is right.

designsats:
@openoms did refactor. I need to take another stab at running ios locally. Its been problematic to test.

kngako (approved):
Code is clean.

kngako:
Testing by copying the payment hash of a receive lightning payment.

image

I have a pending onchain transaction to test out.

kngako:
On a pending received onchain transactions we don't have a copy transaction ID. But it does open mempool.space when the hash is clicked. 

@designsats do you think you can we get a redesign for the pending received onchain transaction so we have two buttons. 1 to copy hash and another to view in explorer? 



image


image

kngako:
Although my transaction had 3 confirmations the app says I have an unclaimed deposit without highlightning what I should do to claim the deposit (if I have to wait for more confirmations it should say that).

image

kngako:
After what appears to be 3 confirmations and a random amount of time I got more details on how to claim the deposit.


image


Interestingly onchain deposits don't appear in the transaction history before they get claimed. Should we have an unclaimed section/filter in the transaction history screen.

image

I am apprently also not able to claim the deposit because "network fees" are too high. Guessing it looks at a percentage of the claimed value because 198 sats for an onchain transaction isn't too high by any other measure (unless it's looking to do a sub 1-sat/vb transaction).

image

kngako:
Refund looks straight forward.

image

kngako:
After transaction confirmed seems the transaction hash being passed to the copy and the open in explorer is wrong. Fix required.



image

grimen (commented):
## Test Coverage Gaps

- The new spec exercises `IconAction` in isolation only; nothing asserts the screen-level wiring, e.g. that tapping copy on the hash row puts the `transactionHash` on the clipboard.

## Code Smells

- `IconAction` is exported from the screen file solely so the spec can import it; testing through the screen (or moving the wrapper next to `GaloyIconButton`) would avoid the test-only export.

PR #3899, opened 2026-07-03
feat(card): card onboarding flow improvements


esaugomez31:
## Overview

Improvements to the Blink Card onboarding flow, plus a new in-app Fee Schedule screen, a Terms and Conditions acknowledgement step, a Card row in the Balances overview, and remote-config-driven fee values.

### Subscribe screen
- Split the agreement into two required checkboxes (automatic renewal and fee schedule acknowledgement); Continue stays disabled until both are checked.
- The status row shows "Special offer" in the subscribe variant.
- The "Blink Card Fee Schedule" link now navigates to the new in-app Fee Schedule screen instead of opening a browser.
- Figma styling pass.

### Processing screen
- Subtitle updated to notify-when-ready copy (removed the wait-time wording).

### Terms and Conditions acknowledgement screen (new)
- New step between the final onboarding step and processing.
- Links to the electronic communications consent, the issuer privacy policy, and the Blink Card cardholder agreement (URLs from remote config).

### Personal information
- Fixed the dropdown modal position so the first option is no longer clipped.

### Blink Card Fee Schedule screen (new)
- In-app fee schedule matching Figma: fee sections, BTC conversion note, fee-update notice, and a collapsible "Additional fee details" section.
- Fully translated across all 28 languages.
- Fee amounts are driven by remote config. The annual fee reuses `cardSubscriptionPriceUsd`, so it stays in sync with the subscribe screen (single source of truth, no drift).

### Card row in the Balances overview (new)
- Shows a "Card" row when the account has a card (any status, including frozen), navigating to the card dashboard.
- Uses a new static theme token `_grey4` (#393939) for the chip, consistent in light and dark.

## i18n

New keys added to `en` and all 28 translation files (`CardFlow.CardFeeSchedule`, `common.card`, and the acknowledgement copy), with real per-language diacritics.

## Testing

- New and updated specs: Fee Schedule screen (100% coverage), Balances Card row (100%), acknowledgement screen, processing screen, and subscribe screen.
- tsc and eslint clean.

## Deploy / follow-up (Firebase Remote Config)

The app ships defaults for every key below, so nothing breaks without them. To control them remotely, create:

URLs:
- `cardCardholderAgreementUrl`, `cardFeeScheduleUrl`, `cardESignConsentUrl`, `cardIssuerPrivacyPolicyUrl`

Fee amounts:
- `cardReplacementFeeUsd`, `cardUsdTransactionFeePercent`, `cardForeignTransactionFeePercent`, `cardMaxOverdraftUsd`, `cardLateRepaymentFeeUsd`


esaugomez31:
* **#3899** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3899?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):
## Review — feat(card): card onboarding flow improvements

All findings verified against the PR head (`43205f8`). Findings 1–3 are blocking; the rest are requested but not gating.

### Correctness

**1. [blocking] The subscribe flow no longer requires (or shows) acceptance of Blink's own Terms of Service and Privacy Policy** — `card-subscription-screen.tsx`

On `main`, acceptance has been doubly enforced in both variants since the screen was created in #3706: the checkbox ("…Terms of Service, Privacy Policy, and Cardholder Agreement") rendered unconditionally, the button was `disabled={!isAgreed || (isSubscribeVariant && !isRenew)}`, and the handler guarded `if (!isAgreed) return`. Now:

```ts
const isAcceptDisabled = isSubscribeVariant ? !isSubscribeReady : !isAgreed
// isSubscribeReady = isRenew && isFeeScheduleAgreed
```

and the ToS/Privacy checkbox JSX lives only in the payment branch, unreachable from `cardOnboardingSubscribeScreen`. The Cardholder Agreement was relocated to the new acknowledgement screen, but Blink's own ToS and Privacy Policy are covered nowhere in the subscribe flow — the acknowledgement screen links the e-sign consent and the **issuer's** privacy policy instead. If this is a deliberate compliance restructuring, please confirm explicitly; otherwise the checkbox needs to come back for the subscribe variant.

**2. [blocking] New network-only `card` query on every home-screen mount for every authed user** — `wallet-overview.tsx:136`

`useCardData()` hardcodes `fetchPolicy: "network-only"` with only a `!isAuthed` skip, so every cold start / home remount now fetches the full cards + card-applications payload during the busiest startup window — including for the majority of users with no card — just to compute `Boolean(card)`. The `homeAuthed` query already fetches `me.defaultAccount`; adding `cards { id }` there and passing it down (see finding 15) makes this free.

**3. [blocking] "Cardholder Agreement" on the fee schedule screen is a dead link** — `card-fee-schedule-screen.tsx:142`

The text is styled exactly like the tappable links elsewhere (`fontWeight: "700", color: colors.primary`) but has no `onPress` anywhere in its Text tree, and the screen never imports `cardCardholderAgreementUrl` — on the very screen the new "I have reviewed and agree to the Blink Card Fee Schedule" checkbox points to. Sibling screens wire the same label to `InAppBrowser.open(cardCardholderAgreementUrl)`.

**4. Legal links call `InAppBrowser.open()` bare, with no catch or fallback** — `card-acknowledgement-screen.tsx:88,103,110` (and the new subscription-screen links)

`open()` rejects on Android without a Custom-Tabs browser and on rapid double-tap ("Another InAppBrowser is already being presented"), leaving an unhandled rejection and the legal document unopened — while the user can still check every box. The repo already has the exact helper:

```ts
// app/utils/external.ts
export const openExternalUrl = async (url: string): Promise<void> => {
  try { await InAppBrowser.open(url) } catch { await Linking.openURL(url) }
}
```

Please use it for the new call sites.

**5. Home-screen Card row shows for any card status, including `CANCELED`/`FAILED`** — `wallet-overview.tsx:137`

`hasCard = Boolean(card)` with `card = cards?.[0]` and no status filter. The dashboard does guard defunct cards with a "not currently available" empty screen, so it's not a crash — but if the API returns canceled cards, a user who closed their card account gets a permanent dead-end row. The new tests only cover `ACTIVE`/`LOCKED`/no-card.

**6. `formatDateFromNow({ years: SUBSCRIPTION_RENEWAL_MONTHS / 12 })` truncates fractional years** — `card-subscription-screen.tsx:52`

`setFullYear` coerces via ToIntegerOrInfinity, so with the constant set to e.g. `6`, the shown renewal date is **today** while the checkbox says "renew in 6 months". Latent today (constant is 12), but it defeats the point of parameterizing by months — use an `addMonths`-style computation.

### Test coverage

**7. The rerouting of personal info → acknowledgement screen is never asserted.** The spec only tests the disabled-button path; a regression reverting `navigate` to `cardOnboardingProcessingScreen` (the pre-PR target) would silently skip the entire new legal gate with a green suite.

**8. Subscribe-variant gating is tested asymmetrically.** Only "renew checked, fee-schedule unchecked" is covered; a regression to `isSubscribeReady = isFeeScheduleAgreed` (dropping auto-renewal consent) passes every test.

**9. The all-four-checkboxes invariant on the acknowledgement screen is only tested with 1 of 4 checked.** Dropping any single later checkbox from the "all accepted" computation passes all tests while allowing applications without a required consent.

**10. The dropdown modal fix has zero coverage — no `DropdownComponent` spec exists.** Removing the `ScrollView` or `maxHeight: "80%"` would silently re-introduce the clipped-options bug this PR fixes.

### Code smells

**11. `cardFeeScheduleUrl` is born dead** — `feature-flags-context.tsx:25`. Full 5-point plumbing (key, type, default, fetch, context) but no consumer — the fee schedule shipped as the in-app screen. Its default is even a copy-paste of the *cardholder agreement* URL. Delete it, and drop it from this PR's deploy instructions so the Firebase parameter never gets created.

**12. `cardProcessingWaitTimeHours` became dead in this PR** — `feature-flags-context.tsx:29`. The processing screen dropped its only consumer, leaving a remote-config parameter operators can "tune" with zero effect.

**13. The checkbox+label pattern is hand-rolled three times in the subscription screen while this same PR introduces exactly that abstraction as `CheckboxRow` in the acknowledgement screen** (byte-identical styles, already drifting: label `lineHeight` 16 vs 18). Promote `CheckboxRow` to a shared component.

**14. `_grey4: "#393939"` shadow-names the existing theme-aware `grey4`** (light `#E2E2E4` / dark `#393939`) — `rne-theme/colors.ts:10`. Its single consumer sits 30 lines from a `colors.grey4` usage in the same stylesheet, and a fixed dark pill on the light theme breaks the palette's light/dark convention. Use the existing grey scale or extend CurrencyPill's variants.

**15. Card-domain logic embedded in the generic balances component** — `wallet-overview.tsx:136`. Every other datum arrives via props from the home screen, but the card row self-fetches through a screens-layer hook and hardcodes card navigation. Prop-driving it from the home screen also resolves finding 2.

**16. The card row's value is a hardcoded `****` — the same glyph used as the hideAmount mask — rendered unconditionally** (`wallet-overview.tsx:267`). With balances revealed, BTC/USD show numbers while Card still shows the "hidden" mask. Show real content (e.g. `•••• {lastFour}`) or tie into hideAmount.

**17. `isSubscribeVariant` now forks nearly the whole subscription component** (accept logic, status label, handler, disjoint checkbox block, button title). Two thin screens over a shared subscription-summary component would remove every ternary.

**18. The 25-line CheckBox mock (plus BlurView/LinearGradient shims) is copy-pasted verbatim between the two new specs.** Move it to a shared helper under `__tests__/helpers` before a third copy appears.

esaugomez31:
@grimen, the feedback has been applied. Just keep in mind that the card feature is not available on main yet, only on staging for now, so the calls are temporarily gated to staging only. The code can be merged without any issues

### Correctness

**1. [blocking] The subscribe flow no longer requires (or shows) acceptance of Blink's own Terms of Service and Privacy Policy**
Fixed. It was not intentional. Restored the ToS/Privacy/Cardholder checkbox in the subscribe variant and folded it into the gate: `isSubscribeReady = isRenew && isFeeScheduleAgreed && isAgreed`.

**2. [blocking] New network-only `card` query on every home-screen mount for every authed user**
Fixed, but not via `homeAuthed`. `ConsumerAccount.cards` only exists on the staging backend, so adding it to `homeAuthed` 400s the whole home screen on every other instance. Kept the isolated `useCardData` query, gated it to staging with a TODO, and it is skipped everywhere else, so the cost is gone for non-staging users.

**3. [blocking] "Cardholder Agreement" on the fee schedule screen is a dead link**
Fixed. Wired it to `cardCardholderAgreementUrl` via `openExternalUrl`.

**4. Legal links call `InAppBrowser.open()` bare, with no catch or fallback**
Fixed. Swapped every card-screen call site to `openExternalUrl`.

**5. Home-screen Card row shows for any card status, including `CANCELED`/`FAILED`**
Fixed. Added a shared `isCardUsable` (ACTIVE/LOCKED) predicate reused by the dashboard, and the home row now uses it.

**6. `formatDateFromNow({ years: SUBSCRIPTION_RENEWAL_MONTHS / 12 })` truncates fractional years**
Fixed. Added a `months` param to `formatDateFromNow` and dropped the `/12`.

### Test coverage

**7. The rerouting of personal info to acknowledgement screen is never asserted**
Added. New test selects all fields and asserts navigation to the acknowledgement screen.

**8. Subscribe-variant gating is tested asymmetrically**
Added. Covered the missing cases (fee-only, terms-unchecked) plus the all-accepted happy path.

**9. The all-four-checkboxes invariant on the acknowledgement screen is only tested with 1 of 4 checked**
Added. Loop asserting the button stays disabled when any single one of the four is left unchecked.

**10. The dropdown modal fix has zero coverage — no `DropdownComponent` spec exists**
Not applied. A spec already exists at `__tests__/components/card-screen/dropdown.spec.tsx` (render, modal, options, disabled). The centering fix is style-only with no RTL-observable behavior.

### Code smells

**11. `cardFeeScheduleUrl` is born dead**
Fixed. Deleted all 5 plumbing spots and dropped it from the deploy notes.

**12. `cardProcessingWaitTimeHours` became dead in this PR**
Fixed. Deleted all 5 plumbing spots.

**13. The checkbox+label pattern is hand-rolled three times in the subscription screen**
Fixed. Promoted `CheckboxRow` to a shared component and used it in both screens.

**14. `_grey4: "#393939"` shadow-names the existing theme-aware `grey4`**
Fixed. Renamed the static token to `_cardPill`.

**15. Card-domain logic embedded in the generic balances component**
Fixed. `wallet-overview` no longer imports the screens-layer hook; `home-screen` fetches and prop-drives `hasCard` / `cardLastFour`.

**16. The card row's value is a hardcoded `****`**
Fixed. Renders `•••• {lastFour}` now, and ties into `hideAmount` (shows `••••` when balances are hidden).

**17. `isSubscribeVariant` now forks nearly the whole subscription component**
Deferred. The shared `CheckboxRow` and the now-unconditional agreement checkbox already reduced the forking; the full split into two thin screens is a larger refactor I would rather do on its own.

**18. The 25-line CheckBox mock (plus BlurView/LinearGradient shims) is copy-pasted verbatim between the two new specs**
Fixed. Extracted the CheckBox mock to `__tests__/helpers/card-flow-mocks.tsx` (hoist-safe factory).

grimen (approved):

PR #3902, opened 2026-07-04
fix(navigation): remove default header shadow globally

designsats:
## Problem

Every screen with a navigation header (Settings, Circles, onboarding, phone validation, etc.) renders an unintended drop shadow / elevation line beneath the header bar. It clashes with the app's flat white header design.

## Cause

The shared `screenOptions` on our native-stack navigators in `app/navigation/root-navigator.tsx` don't set `headerShadowVisible: false`, so the native `@react-navigation/native-stack` default elevation shows. Native-stack header options don't inherit across nested stacks, so the flag has to be set on each navigator. Previously only a single per-screen override existed (`profileScreen`).

## Fix

- Add `headerShadowVisible: false` to the shared `screenOptions` of all four stack navigators: `RootStack`, `OnboardingNavigator`, `ContactNavigator`, `PhoneLoginNavigator`.
- Remove the now-redundant per-screen override on `profileScreen`.

## Testing

- [x] Android: no shadow/elevation under any header (Settings, Circles, onboarding), light + dark theme
- [ ] iOS: header hairline gone


designsats:
@blink-claw-bot review

designsats:
shadow removed
image

openoms:
Review of head `5281d4f20ab501774ad026e5c8f0a1fe7dd3a73e`.

## Recommendation

**Approve after one fix: complete the testing checklist with evidence from both platforms.** The code change itself is correct and complete, no changes to the diff are required.

## Strengths

- The diff does exactly what the description claims, completely. There are only four native-stack navigators in the app (all in `app/navigation/root-navigator.tsx`), and the flag is added to the shared `screenOptions` of every one: `RootStack` (L211), `OnboardingNavigator` (L857), `ContactNavigator` (L912), `PhoneLoginNavigator` (L970). The fifth navigator (bottom-tab, L999) needs no change since all of its screens set `headerShown: false` and never render a header.
- No leaks around the fix: every `navigation.setOptions` call in the app (settings, card screens, receive, webview, earns) only sets `title`/`headerRight`, and no screen uses a custom `header:` component that would bypass the flag.
- Removing the now-redundant `profileScreen` per-screen override instead of leaving a dead duplicate is right.
- The root-cause writeup is accurate: native-stack options genuinely don't inherit across nested navigators, so per-navigator placement is the correct mechanism.

## Findings

**I1 (Important, gating): testing checklist unchecked, only one platform evidenced.**
The PR's own test plan (Android light + dark, iOS hairline) is entirely unchecked, and the single screenshot in the comments doesn't state platform or theme. This change touches every header in the app and the two platforms remove different artifacts (Android elevation vs iOS hairline). Please check the boxes and add before/after screenshots per platform, ideally in the description.

**D2 (non-gating, DRY / future-regression risk): fourth copy of the same `screenOptions` block.**
All four navigators now duplicate the identical six-option block (`gestureEnabled`, `headerBackTitle`, `headerStyle`, `headerTitleStyle`, `headerTintColor`, `headerShadowVisible`). The non-inheritance invariant that forces this is documented in the PR description but invisible in the code, so a future fifth navigator will silently regress the shadow. Suggest extracting a shared `baseStackScreenOptions`, or at minimum a one-line comment on each block noting that native-stack options don't inherit across nested navigators.

**D3 (non-gating, test coverage): no test goes red if this is reverted.**
A Jest test can't observe a native shadow, so "pinned by screenshots" is acceptable pushback here. Note the D2 extraction would give a cheap pin: a spec asserting the shared options object contains `headerShadowVisible: false`.

## CI note

The two failed checks are Android/iOS E2E, both dead at the 24h timeout, the known-flaky pattern. Detox asserts on testIDs, not visuals, so a header-shadow change can't plausibly be the cause. Not blocking. All hard gates (Check Code, Tests, Audit, Conventional commits, CodeQL) are green.

grimen (approved):

PR #3903, opened 2026-07-04
fix(self-custodial): claim on-chain deposits at the network-recommended fee

esaugomez31:
## Problem

Self-custodial (Spark) on-chain deposit claims failed with "Network fees are too high to claim (N sats required)", even on large deposits where the required fee was worth a few cents. The app never passed a max fee, so the Breez/Spark SDK applied its conservative default cap (`Rate { satPerVbyte: 1 }`), which sits below the real on-chain claim fee. Both the automatic background claim and the manual "Claim now" fell to this too-low default, leaving funds stuck as unclaimed deposits.

## Fix

Two layers, following the Breez guidance for on-chain claims:

- **Automatic claims** — `createSdkConfig` now sets `config.maxDepositClaimFee = MaxFee.NetworkRecommended({ leewaySatPerVbyte })`, tracking the network-recommended rate plus a small leeway, so background claims succeed at normal fees.
- **Manual "Claim now"** — passes `maxFeeSats: deposit.requiredFeeSats`, so an explicit claim overrides the cap with the exact fee the SDK reported as required.

## Remote-configurable leeway (single source of truth)

The leeway is a Firebase Remote Config flag `selfCustodialDepositClaimLeewayVbyte` (number, default `1`), wired through the standard feature-flags convention and consumed via `useRemoteConfig()`. There is no hardcoded fee constant in the SDK bridge: all three connect paths (live wallet, delete-probe, restore) read the remote value and thread it into `initSdk` (required parameter). The default is registered via `setDefaults`, so it also acts as the fail-safe when the key is absent.

Two robustness details:
- The leeway is clamped to a non-negative integer (`Math.max(0, Math.trunc())`) before `BigInt()`, so a fractional or invalid remote value can never throw and break the wallet connect.
- The live wallet reads the leeway through a ref, so a remote-config change never forces an SDK teardown/reconnect (no balance/tx flash).

## Comparison with Breez `glow-web` (reference wallet)

Cross-checked against Breez's official reference implementation ([breez/glow-web](https://github.com/breez/glow-web)):

- **Manual claim matches exactly** — glow-web uses `{ type: 'fixed', amount: requiredFeeSats }`; we use the same (`MaxFee.Fixed(requiredFeeSats)`).
- **Auto-claim default is deliberately different (better for a consumer app)** — glow-web defaults to `Rate 1` (the same too-low value) and relies on a per-user fee settings UI plus the manual override. Blink Mobile has no such settings screen, so we default the automatic path to `NetworkRecommended`, which tracks the real network fee and lets claims succeed without user intervention. The remote-config flag provides central tuning in place of glow-web's per-user setting.
- **Defensive validation of the fee value** — glow-web validates the (untrusted, persisted) fee before using it; our `Math.trunc`/clamp guard defends the same class of risk on the (untrusted, remote) value.

## Testing

- `initSdk` config shape (`NetworkRecommended` + leeway), provided leeway, fractional truncation, negative clamp.
- Manual claim passes `maxFeeSats`; success and still-exceeds paths.
- Leeway threaded through probe and restore; live wallet does not reconnect when only the leeway changes.
- Full affected suite green; typecheck and lint clean.

## Deploy note

Create the Firebase Remote Config key `selfCustodialDepositClaimLeewayVbyte` (number, default `1`). The app ships the default, so it works without the key; raise it to add fee headroom.

Device-validated: an on-chain deposit that arrived while the app was closed was auto-claimed on the next app open, with no manual action.

### DIFF
--- a/__tests__/screens/self-custodial/onboarding/restore/hooks/use-restore-wallet.spec.ts
+++ b/__tests__/screens/self-custodial/onboarding/restore/hooks/use-restore-wallet.spec.ts
@@ -36,7 +36,11 @@ jest.mock("@app/self-custodial/hooks/use-spark-network", () => ({
 }))
 
 jest.mock("@app/self-custodial/bridge", () => ({
-  selfCustodialRestoreWallet: (...args: string[]) => mockRestore(...args),
+  selfCustodialRestoreWallet: (...args: unknown[]) => mockRestore(...args),
+}))
+
+jest.mock("@app/config/feature-flags-context", () => ({
+  useRemoteConfig: () => ({ selfCustodialDepositClaimLeewayVbyte: 5 }),
 }))
 
 jest.mock("@app/self-custodial/storage/account-index", () => ({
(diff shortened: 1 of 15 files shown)


esaugomez31:
* **#3903** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3903?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3903?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (approved):

PR #3910, opened 2026-07-11
feat(self-custodial): force converting the blocked stable-token balance to bitcoin


esaugomez31:
## Summary
When the Dollar Balance feature is region-blocked and a self-custodial user still holds a stable-token balance, the home screen now auto-opens the same forced-conversion modal custodial users already get, so the blocked funds can be moved instead of silently hidden.

## How it works
- The modal converts the **full** USDB balance through the same quote/execute pipeline as the regular conversion flow, and after a successful conversion deactivates Stable Balance so future receives stay in bitcoin instead of re-trapping.
- The custodial modal was split into a shared presentational component + its unchanged container; the custodial behavior is untouched (its spec passes as-is) and the existing conversion flow is untouched.
- The trigger hook was renamed `useStablesatsForcedConversion` → `useDollarBalanceForcedConversion` since it now drives both account types.

Closes blinkbitcoin/blink-wip#956. Per the ticket, HK/VN/KR can be added to `selfCustodialDollarBalanceBlockedCountries` only after this ships.

image

### DIFF
--- a/__tests__/components/custom-modal.spec.tsx
+++ b/__tests__/components/custom-modal.spec.tsx
@@ -29,10 +29,12 @@ jest.mock("react-native-modal", () => {
     children,
     isVisible,
     onBackdropPress,
+    onBackButtonPress,
   }: {
     children: React.ReactNode
     isVisible: boolean
     onBackdropPress?: () => void
+    onBackButtonPress?: () => void
   }) =>
     isVisible
       ? ReactNs.createElement(
@@ -41,6 +43,10 @@ jest.mock("react-native-modal", () => {
           ReactNs.createElement(RN.Pressable, {
             testID: "backdrop",
             onPress: onBackdropPress,
+          }),
+          ReactNs.createElement(RN.Pressable, {
+            testID: "back-button",
+            onPress: onBackButtonPress,
           }),
           children,
         )
@@ -102,6 +108,28 @@ describe("CustomModal", () => {
     )
 
     fireEvent.press(getByTestId("backdrop"))
+
+    expect(toggleModal).not.toHaveBeenCalled()
+  })
+
+  it("closes on the Android back button by default", () => {
+    const toggleModal = jest.fn()
+    const { getByTestId } = render(
+      <CustomModal {...baseProps} toggleModal={toggleModal} />,
+    )
+
+    fireEvent.press(getByTestId("back-button"))
+
+    expect(toggleModal).toHaveBeenCalledTimes(1)
+  })
+
+  it("ignores the Android back button when not dismissable", () => {
+    const toggleModal = jest.fn()
+    const { getByTestId } = render(
+      <CustomModal {...baseProps} toggleModal={toggleModal} dismissable={false} />,
+    )
+
+    fireEvent.press(getByTestId("back-button"))
 
     expect(toggleModal).not.toHaveBeenCalled()
   })
--- a/__tests__/components/usd-convert-to-btc-modal/convert-to-btc-modal-ui.spec.tsx
+++ b/__tests__/components/usd-convert-to-btc-modal/convert-to-btc-modal-ui.spec.tsx
@@ -1,0 +1,165 @@
+import React from "react"
+import { fireEvent, render } from "@testing-library/react-native"
+
+import TypesafeI18n from "@app/i18n/i18n-react"
+import { loadLocale } from "@app/i18n/i18n-util.sync"
+import { toUsdMoneyAmount } from "@app/types/amounts"
+import { ThemeProvider } from "@rn-vui/themed"
+
+const mockFormatMoneyAmount = jest.fn(
+  ({
+    moneyAmount,
+    isApproximate,
+  }: {
+    moneyAmount: { amount: number; currency: string }
+    isApproximate?: boolean
+  }) => `${isApproximate ? "~ " : ""}${moneyAmount.currency}:${moneyAmount.amount}`,
+)
+
+const mockConvertMoneyAmount = jest.fn((_moneyAmount: unknown, toCurrency: string) => ({
+  amount: 129184,
+  currency: toCurrency,
+  currencyCode: toCurrency,
+}))
+let mockConvertReady = true
+
+jest.mock("@app/hooks/use-display-currency", () => ({
+  useDisplayCurrency: () => ({ formatMoneyAmount: mockFormatMoneyAmount }),
+}))
+
+jest.mock("@app/hooks/use-price-conversion", () => ({
+  usePriceConversion: () => ({
+    convertMoneyAmount: mockConvertReady ? mockConvertMoneyAmount : undefined,
+  }),
+}))
+
+const mockModalRender = jest.fn()
+
+jest.mock("react-native-modal", () => {
+  const ReactNs = jest.requireActual<typeof import("react")>("react")
+  const RN = jest.requireActual<typeof import("react-native")>("react-native")
+  const MockModal = (props: { children: React.ReactNode; isVisible: boolean }) => {
+    mockModalRender(props)
+    return props.isVisible ? ReactNs.createElement(RN.View, null, props.children) : null
+  }
+  return { __esModule: true, default: MockModal }
+})
+
+import { ConvertToBtcModalUI } from "@app/components/usd-convert-to-btc-modal"
+
+loadLocale("en")
+
+const usdBalance = toUsdMoneyAmount(10001) // $100.01
+
+const wrap = (ui: React.ReactElement) => (
+  <ThemeProvider>
+    <TypesafeI18n locale="en">{ui}</TypesafeI18n>
+  </ThemeProvider>
+)
+
+type OptionalProps = {
+  isVisible?: boolean
+  toggleModal?: () => void
+  onConvert?: () => void
+  loading?: boolean
+  dismissable?: boolean
+  errorMessage?: string
+}
+
+const renderUI = (props?: OptionalProps) =>
+  render(
+    wrap(
+      <ConvertToBtcModalUI
+        isVisible={props?.isVisible ?? true}
+        toggleModal={props?.toggleModal ?? jest.fn()}
+        usdWalletBalance={usdBalance}
+        onConvert={props?.onConvert ?? jest.fn()}
+        loading={props?.loading ?? false}
+        dismissable={props?.dismissable}
+        errorMessage={props?.errorMessage}
+      />,
+    ),
+  )
+
+describe("ConvertToBtcModalUI", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockConvertReady = true
+  })
+
+  it("renders the title, body and both amount rows", () => {
+    const { getByText } = renderUI()
+
+    expect(getByText("Dollar Balance is not available in your region")).toBeTruthy()
+    expect(getByText("Transfer from Dollar Balance to Bitcoin Balance")).toBeTruthy()
+    expect(getByText("You have")).toBeTruthy()
+    expect(getByText("You get")).toBeTruthy()
+  })
+
+  it("shows the exact balance and the approximate bitcoin estimate", () => {
+    const { getByText } = renderUI()
+
+    expect(mockFormatMoneyAmount).toHaveBeenCalledWith({ moneyAmount: usdBalance })
+    expect(getByText("USD:10001")).toBeTruthy()
+    expect(getByText("~ BTC:129184")).toBeTruthy()
+  })
+
+  it("hides the bitcoin estimate while the price conversion is unavailable", () => {
+    mockConvertReady = false
+    const { getByText, queryByText } = renderUI()
+
+    expect(getByText("You get")).toBeTruthy()
+    expect(queryByText("~ BTC:129184")).toBeNull()
+  })
+
+  it("shows the warning icon", () => {
+    const { getByTestId } = renderUI()
+
+    expect(getByTestId("icon-warning")).toBeTruthy()
+  })
+
+  it("forwards the visibility to the modal", () => {
+    renderUI({ isVisible: false })
+
+    expect(mockModalRender).toHaveBeenCalledWith(
+      expect.objectContaining({ isVisible: false }),
+    )
+  })
+
+  it("calls onConvert when Transfer is pressed", () => {
+    const onConvert = jest.fn()
+    const { getByText } = renderUI({ onConvert })
+
+    fireEvent.press(getByText("Transfer"))
+
+    expect(onConvert).toHaveBeenCalledTimes(1)
+  })
+
+  it("disables Transfer while loading", () => {
+    const { queryByText, getByRole } = renderUI({ loading: true })
+
+    expect(queryByText("Transfer")).toBeNull()
+    expect(getByRole("button", { disabled: true })).toBeTruthy()
+  })
+
+  it("renders the error box when an error message is provided", () => {
+    const { getByText } = renderUI({ errorMessage: "Insufficient balance" })
+
+    expect(getByText("Insufficient balance")).toBeTruthy()
+  })
+
+  it("is locked by default: no close icon", () => {
+    const { queryByTestId } = renderUI()
+
+    expect(queryByTestId("icon-close")).toBeNull()
+  })
+
+  it("shows the close icon and closes through it when dismissable", () => {
+    const toggleModal = jest.fn()
+    const { getByTestId } = renderUI({ dismissable: true, toggleModal })
+
+    fireEvent.press(getByTestId("icon-close"))
+
+    expect(toggleModal).toHaveBeenCalledTimes(1)
+  })
+})
--- a/__tests__/components/usd-convert-to-btc-modal/usd-convert-to-btc-modal.spec.tsx
+++ b/__tests__/components/usd-convert-to-btc-modal/usd-convert-to-btc-modal.spec.tsx
@@ -8,13 +8,8 @@ import { toUsdMoneyAmount } from "@app/types/amounts"
 import { ThemeProvider } from "@rn-vui/themed"
 
 const mockFormatMoneyAmount = jest.fn(
-  ({
-    moneyAmount,
-    isApproximate,
-  }: {
-    moneyAmount: { amount: number; currency: string }
-    isApproximate?: boolean
-  }) => `${isApproximate ? "~ " : ""}${moneyAmount.currency}:${moneyAmount.amount}`,
+  ({ moneyAmount }: { moneyAmount: { amount: number; currency: string } }) =>
+    `${moneyAmount.currency}:${moneyAmount.amount}`,
 )
 
 const mockConvertMoneyAmount = jest.fn((_moneyAmount: unknown, toCurrency: string) => ({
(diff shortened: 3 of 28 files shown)


esaugomez31:
* **#3910** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):
Thanks — the split into a shared `ConvertToBtcModalUI` is clean and the custodial path is verifiably behavior-preserving. The issues below all cluster around the new non-dismissable modal's failure states; **1–3 are blocking** since each can strand a user's funds or their entire home screen behind a modal that can't be closed.

**1. A deterministic quote failure permanently bricks the home screen** — `app/screens/conversion-flow/stable-token-convert-to-btc-modal.tsx:77`
`useDollarBalanceForcedConversion` opens the modal for any `usdWalletBalance > 0`, but the bridge throws `BelowMinimum` for balances under `minFromAmount` (`app/self-custodial/bridge/convert.ts:140`). For a dust balance the quote therefore errors on every attempt, `canExecute` stays false, and with `dismissable={false}` / `showCloseIconButton={false}` and no retry affordance the user is stuck behind a generic error forever — it re-latches on every app start. There's also a silent variant: while the price feed is unavailable the quote sits Idle (no spinner, no error, disabled button) until the 5-min poll recovers. Suggest a minimum-balance threshold in the trigger hook plus a retry (or escape) path on quote error.

**2. A failed execute retries the same stale quote forever** — `app/screens/conversion-flow/stable-token-convert-to-btc-modal.tsx:55`
`useSelfCustodialConversion` pins `snapshotParams` once a quote resolves and only resets it on `enabled`/`fromCurrency`/`moneyAmount` changes (`hooks/self-custodial/use-conversion.ts:56-68`) — none of which change on execute failure. If the modal sits open past `MAX_SLIPPAGE_BPS`, every Transfer press replays the identical frozen `PrepareSendPaymentResponse`. The regular conversion screen recovers because the user can navigate away and back; this modal can't be dismissed, so there's no path to a fresh quote. Clearing the snapshot when execute fails fixes it.

**3. The "stop re-trapping" invariant isn't enforced, enabling a fee-paying loop** — `app/screens/conversion-flow/stable-token-convert-to-btc-modal.tsx:40`
Two gaps: (a) `deactivateStableBalance` is a network-backed SDK settings write; on failure it's swallowed into `reportError` with no notice and no retry, so the next incoming receive auto-converts back into restricted USDB and the forced modal reopens for another fee-paying conversion. (b) The stable-balance settings toggle has no restriction gate (`use-stable-balance-toggle.ts` never consults `useDollarBalanceRestricted`), so a restricted user can re-activate Stable Balance and deterministically loop: activate → sweep converts BTC→USDB (fee) → forced modal converts back (fee). The restriction should gate activation too.

**4. Success path refreshes wallets before closing the modal** — `app/screens/conversion-flow/stable-token-convert-to-btc-modal.tsx:44`
`stopRetrappingAndClose` awaits the refreshes before `toggleModal()`. `refreshWallets` commits the zeroed balance mid-await, the memoized amount drops to 0 while the modal is still visible/enabled, the snapshot resets, and a guaranteed-useless 0-amount quote fires — which the bridge rejects (`BelowMinimum`), flashing `LL.errors.generic()` right after a *successful* conversion. This inverts the invariant documented in `use-dollar-balance-forced-conversion.ts` ("closes on success immediately, before the balance refetches to zero"), which the custodial path (`onSuccess: toggleModal`) respects. Close first, then refresh.

**5. Always-mounted modal registers two standing price pollers per self-custodial user** — `app/screens/home-screen/home-screen.tsx:607`
`enabled: isVisible` only nulls the quote params; the hidden mount still holds two `usePriceConversion` instances, each an independent `useRealtimePriceUnauthedQuery` with `cache-and-network` + 5-min poll, duplicating the poller `useTotalBalance` already runs — for the common case where no conversion is ever forced. (The custodial twin only registers lazy mutations, so this is a new standing cost.) Gating the mount on `isConvertModalVisible` as well eliminates it; the entrance animation still plays on mount.

**6. Re-entrancy guard belongs inside `useSelfCustodialConversion`** — `app/screens/conversion-flow/stable-token-convert-to-btc-modal.tsx:63`
`useIntraLedgerConversion` embeds `useInFlightGuard` so callers need nothing; this hook doesn't, so the modal wraps `execute` at the call site while the hook's other caller (`conversion-confirmation-screen.tsx:164`) stays unguarded. Relatedly, the hook fires `onSuccess()` un-awaited, which is the only reason `isFinishing`/`isBusy` exist here — `await onSuccess()` inside execute's try would let `loading` span the whole sequence and delete both.

## Test coverage

**7. No test covers recovery from a persistent quote error** — `stable-token-convert-to-btc-modal.spec.tsx:201` only asserts the error string renders, and the "cannot be dismissed" test (line 315) actively pins the trap from finding 1. Once a retry/escape path exists, the spec should assert it.

**8. No execute-failure-then-retry test** — every failure case in `use-conversion.spec.ts` calls `execute()` exactly once; a test calling it twice after a failure and asserting `getConversionQuote` is re-invoked would pin finding 2.

**9. The deactivation-failure test asserts too little** — `stable-token-convert-to-btc-modal.spec.tsx:240` checks only `toggleModal` + `reportError`; it should at least assert the refreshes still run and pin whatever contract addresses finding 3.

**10. The success-path test asserts no ordering** — `stable-token-convert-to-btc-modal.spec.tsx:228` counts calls but nothing asserts close-before-refetch (finding 4) or that no re-quote fires after success.

**11. The Idle-while-enabled state is never exercised** — the only idle test in `use-conversion.spec.ts` uses `enabled: false`, and the modal spec's `mockConvertReady = false` case runs with the conversion hook fully mocked (`canExecute: true`), so the frozen no-feedback state from finding 1 has zero coverage.

**12. The modal spec stubs the entire pipeline** — the wiring tests assert props passed to `mockUseSelfCustodialConversion` rather than product behavior; one integration case (real hook + mocked `getConversionQuote`, USD fromCurrency) would catch contract drift the mock boundary hides.

## Code smells

**13. Unreachable outer try/catch** — `stable-token-convert-to-btc-modal.tsx:48`: both awaited calls already `.catch()`, and `toggleModal` is a setState wrapper that can't throw; the catch exists only to satisfy the spec's throwing mock. Drop it (keep the `finally`).

**14. Modal exported through the screens barrel** — `app/screens/conversion-flow/index.ts:4`: home-screen imports one modal via the barrel, pulling the whole conversion screen stack into its module graph and forcing `home.spec.tsx` to mock the entire barrel. Move it next to `ConvertToBtcModalUI` in `app/components`, or import the file directly.

**15. `toggleModal` is a required prop that can never fire** — `convert-to-btc-modal-ui.tsx:16`: the UI hardcodes `dismissable={false}` and `showCloseIconButton={false}`, the only paths that invoke it. Remove the prop or make dismissability configurable.

**16. Test asserts its own mock** — `stable-token-convert-to-btc-modal.spec.tsx:301`: "renders nothing when isVisible is false" passes because of the spec's own react-native-modal mock, not component logic. Delete or assert the forwarded prop.

**17. Shared UI presentation tested twice, never at the source** — the extracted `ConvertToBtcModalUI` has no spec, while its presentation assertions (and supporting mocks) are duplicated across both container specs. One `convert-to-btc-modal-ui.spec.tsx` for presentation; containers for wiring.

**18. Modal mutual-exclusivity via action at a distance** — `home-screen.tsx:598`: not a bug today (Apollo 3.9 clears `data` synchronously on skip, and the client is rebuilt on account switch), but the exclusivity of two non-dismissable modals rests on a query-skip 380 lines away and on skip semantics Apollo's own source flags as changing in AC 4.0. Mirroring line 607 with `!isSelfCustodial &&` makes it local and upgrade-proof.

esaugomez31:
@grimen  All 18 points are addressed.

**1. A deterministic quote failure permanently bricks the home screen** — Fixed with a single rule: the modal is now locked only while a conversion is actually executable (`dismissable={!canExecute}`). Quote error, quote loading, and the silent price-feed-down idle state all show a spinner or error with an escape hatch, and on a quote error the Transfer button becomes a retry that fetches a fresh quote. Also gated quoting on wallet readiness and a settled positive balance, so the launch-time transient zero balance no longer fires a doomed below-minimum quote. Covered by unit + integration tests.

**2. A failed execute retries the same stale quote forever** — Fixed in the hook: a failed execute now invalidates the pinned quote and re-quotes. Note: clearing the snapshot alone was not enough (the pin stores the same live-params reference, so the identity never changed) — the invalidation clones the params to force a fresh fetch.

**3. The "stop re-trapping" invariant isn't enforced** — Both gaps closed: (a) a failed deactivation now surfaces the existing "Could not update Stable Balance" toast on top of the report; (b) the settings toggle now blocks *activation* for restricted regions with a toast and snaps the switch back (deactivation stays allowed so an active balance can always be freed).

**4. Success path refreshes wallets before closing the modal** — Reordered: close first, then deactivate, then refresh. The ordering is pinned with `invocationCallOrder` in both the unit and integration specs.

**5. Always-mounted modal registers two standing price pollers** — The mount is now gated on `isSelfCustodial && isConvertModalVisible`, so the hidden case costs nothing.

**6. Re-entrancy guard belongs inside the hook** — Moved `useInFlightGuard` into `useSelfCustodialConversion` (the confirmation screen gets it for free) and `onSuccess` is now awaited, which let me delete the container's `isFinishing` state as you suggested.

**7–11. Test coverage** — All added: quote-error recovery (escape + retry, unit and integration), execute-failure-then-retry pinning `getConversionQuote` re-invocation, deactivation-failure asserting toast + refreshes, close-before-refresh ordering, and the idle-while-enabled state in both the hook spec and the container spec.

**12. The modal spec stubs the entire pipeline** — Added an integration spec that runs the real hook + real quote layer with only `getConversionQuote` mocked, covering the success path end-to-end and the full error → escape → retry → convert cycle.

**13. Unreachable outer try/catch** — Removed (the whole function got simpler once `onSuccess` is awaited).

**14. Modal exported through the screens barrel** — Removed from the barrel; home-screen imports the file directly, and home.spec no longer mocks the whole barrel.

**15. `toggleModal` is a required prop that can never fire** — Dismissability is now configurable on the UI (`dismissable`), which finding 1 needed anyway, so the prop has a real path.

**16. Test asserts its own mock** — Replaced with an assertion on the prop forwarded to the modal, in the new UI spec.

**17. Shared UI presentation tested twice, never at the source** — Added `convert-to-btc-modal-ui.spec.tsx` as the presentation owner and slimmed both container specs down to wiring.

**18. Modal mutual-exclusivity via action at a distance** — Made local: the custodial guard now includes `!isSelfCustodial` (via a `custodialConvertWallets` constant that also carries the wallet ids).

Coverage is at 100% on all touched files and the regular conversion flow's suites stay green.

grimen (changes requested):
Thanks — this round fixes most of what the last review raised, and the fixes hold up: retry after a failed conversion, the double-tap guard inside the shared hook, closing the modal before refreshing, the error toast on deactivation failure, the gate on the settings toggle, and the modal exclusivity guards all check out. The remaining problems come from two things the fixes introduced: the modal decides whether it can be closed based on whether a quote is currently loaded, and the "show the modal" flag is now shared between the two account types. **1 is blocking; 2 and 3 need an explicit product decision.**

**1. Switching accounts can show the wrong modal — and it can't be closed** — `app/screens/home-screen/home-screen.tsx:391`
The home screen is not remounted when the user switches accounts, so the "show the conversion modal" flag (a one-way `useState` latch in `use-dollar-balance-forced-conversion.ts:14-19`) survives the switch. Right after switching to a self-custodial account there is a short window where the app still serves the previous custodial account's cached wallets, but already applies the self-custodial restriction (`isSelfCustodial` is still false because the SDK status is `Unavailable`, while `useDollarBalanceRestricted` gates on `accountType`). In that window the **custodial** modal appears: unclosable, and wired to the custodial account — which isn't restricted at all. Pressing Transfer converts that account's dollars. Once the SDK connects, the stale flag then drives the self-custodial modal. Fix: reset the flag whenever the active account id changes (or key the trigger hook by account).

**2. The close rules are wrong in both directions** — `app/screens/conversion-flow/stable-token-convert-to-btc-modal.tsx:81`
The modal can be closed exactly when there is no usable quote (`dismissable={!canExecute}`). Two consequences:
(a) *Too easy to skip:* while the first quote is still loading — a moment that happens on every open — the close button is already visible (`custom-modal.tsx:107` has no loading gate). If the user closes the modal there, it never comes back for the rest of the session (the trigger only fires when the balance goes from zero to positive, and it stays positive). So the "forced" conversion can be skipped with one tap — unlike the custodial version, which can never be closed. The new tests pin exactly this behavior, so it may be intentional — but it weakens the point of the feature and deserves an explicit sign-off.
(b) *Too hard to escape:* when quotes work but the conversion itself keeps failing (say the conversion service is paused), every attempt fetches a fresh quote, which immediately locks the modal again. The user is stuck in a fail–retry–lock loop and can only escape during the roughly one second while the re-quote is in flight. Android back doesn't work either (no `onBackButtonPress`).
Suggestion: allow closing only when the quote failed, and let the trigger re-open the modal later. That fixes both directions.

**3. Tiny balances still cause a nag modal forever** — `app/hooks/use-dollar-balance-forced-conversion.ts:15`
The modal opens for any balance above zero, but the bridge refuses conversions below a minimum (`convert.ts:140-151`). A balance below that minimum — possible through a small incoming payment, or as leftover from this very flow's own rounding correction (`convert.ts:268-275`) — means: modal on every app start, a retry that can never succeed, and no way to ever get rid of the balance. The minimum-balance check requested last round is still missing.

**4. A retry after a failure can convert too little and strand the rest** — `app/screens/conversion-flow/hooks/self-custodial/use-conversion.ts:74`
If the balance changes while a conversion is in flight (SDK sync events refresh wallets in the background) and the conversion then fails, the retry re-uses the old, smaller amount: `execute` holds the `requote` closure from the moment the button was pressed. The retry converts the old amount, reports success, switches Stable Balance off and closes — and the difference stays trapped until the next app start, because the trigger never fires again. Reading the live params through a ref (or clearing the snapshot on failure instead of re-pinning it) avoids this.

**5. The regular conversion screen got worse** — `app/screens/conversion-flow/conversion-confirmation-screen.tsx:255`
A failed conversion now throws away the pinned quote. If the automatic re-fetch also fails (same outage), the swipe slider stays disabled forever — there is no retry button and nothing refetches on its own. The user has to leave the screen and start over. Before this change they could simply swipe again. The modal got a retry button for exactly this case (`hasQuoteError ? requote : execute`); this screen needs the same.

**6. Opening the Stable Balance settings screen now makes up to four internet requests** — `app/screens/stable-balance-settings-screen/hooks/use-stable-balance-toggle.ts:39`
The new restriction check runs an IP-geolocation lookup on every mount — four external services in sequence, 5-second timeout each, no caching (`ip-country-lookup.ts:77-90`), plus an extra GraphQL query. Self-custodial users (the only ones who see this screen) always take that path because they have no phone number on file. But the result is only needed if the user actually flips the switch on. Check it lazily at press time using the values the home screen already keeps warm, or cache the lookup.

## Code smells

**7. "Deactivate Stable Balance and refresh" now exists twice** — `stable-token-convert-to-btc-modal.tsx:38` vs `use-stable-balance-toggle.ts:70`, and the copies already differ (parallel vs sequential refreshes, different error tags, only one shows a toast on refresh failure). Extract one shared routine.

**8. Errors that aren't `Error` objects are still swallowed silently** — `use-conversion.ts:116`: `if (err instanceof Error)` has no else branch, so the user gets no message, no haptic, no retry — the spinner just stops. The existing `reportError` helper handles this case and is used everywhere else in the self-custodial code.

esaugomez31:
@grimen

**1. Switching accounts can show the wrong modal — and it can't be closed** — Fixed. `useDollarBalanceForcedConversion` now takes the registry's `accountId` and resets the latch whenever it changes, re-arming for the new account on the same render cycle. The home screen also resolves the restricted balance strictly per `accountType` (the cached custodial query is never read while a self-custodial account is active), and the custodial modal render is gated on `accountType === Custodial`. A home spec pins the exact switch window (accountType already self-custodial, SDK still `Unavailable`): neither modal shows.

**2. The close rules are wrong in both directions** — Product decision made: we adopted your suggestion as-is. The modal is dismissable exactly when the quote failed (`dismissable={hasQuoteError}`), so it can no longer be skipped during the first quote load, and a dead quote always offers an exit. Closing is per-visit, not per-session: the trigger re-arms on home focus, so a dismissed modal returns on the next visit. `CustomModal` gained `onBackButtonPress`, gated identically to the backdrop press. Known residual your rule accepts: if quotes keep succeeding while executes keep failing, the modal stays locked — but every failed execute now re-quotes, and the moment the outage also breaks quoting, the escape hatch appears.

**3. Tiny balances still cause a nag modal forever** — Product decision made: we gate on the Breez minimum. The home screen fetches the conversion limits (only when self-custodial + restricted + positive balance) and passes `minimumBalance` to the trigger; below the pool minimum — or while the minimum is still unknown — the modal never opens. If the balance shrinks below the minimum while the modal is already open, the quote error shows the honest `Minimum transfer: …` message (existing key) instead of the generic error.

**4. A retry after a failure can convert too little and strand the rest** — Fixed with the ref you suggested: `requote` reads the live params through `liveQuoteParamsRef`, so a failure retry re-pins the balance as it is now, not as it was at press time. A spec pins the exact scenario: balance changes mid-flight, the conversion fails, and the retry quotes the live amount.

**5. The regular conversion screen got worse** — Fixed. The swipe now mirrors the modal: on a failed quote it retries the quote instead of executing, and the slider is disabled only while the quote is genuinely pending. No more permanent dead slider after a double outage; specs cover both branches.

**6. Opening the Stable Balance settings screen now makes up to four internet requests** — Fixed with `resolveIpCountryCodeCached`: one shared lookup per app session (module-level promise; failures are not cached, so a later mount can retry after starting offline). The settings mount now reuses the lookup the home screen already warmed instead of re-hitting the four services.

**7. "Deactivate Stable Balance and refresh" now exists twice** — Fixed. Extracted `deactivateStableBalanceAndRefresh` (`app/self-custodial/stable-balance.ts`) with one unified semantic: a failed deactivation is reported and toasted but the refresh still runs (funds may have just moved), a failed refresh is report-only, and the boolean result lets callers resync their UI. Both the modal and the settings toggle consume it; the helper has its own spec.

**8. Errors that aren't `Error` objects are still swallowed silently** — Fixed. The catch now always goes through `reportError`, falls back to the generic message, fires the haptic, and re-quotes — regardless of what was thrown. Spec added for a thrown string.

---

While re-auditing the full branch we found three more gaps in the new code and fixed them:

**9. A focus re-open could resurrect the modal on a stale balance — and the custodial one can't be closed** — The focus re-arm from #2 races the post-conversion balance refetch: leave home right after a successful conversion and come back before the refetch lands, and the modal re-opens on a balance that no longer exists — unclosable in the custodial variant. The latch now also drops the moment the balance stops being convertible, so the refetch self-heals any stale re-open (this also covers a self-custodial balance draining to zero while the modal is open).

**10. A limits response without a minimum kept the trigger closed forever** — The bridge treats a fetched `minFromAmount: null` as "no minimum, allow" (`checkConversionMinimum`), but the home screen read it as "unknown" and would never open the modal — silently re-hiding the funds this PR is about. A fetched no-minimum response now falls back to any positive cent, mirroring the bridge.

**11. One failed limits fetch muted the trigger for the whole session** — The fetch never retried (its inputs stay stable while the balance is positive), so a transient network failure at launch suppressed the forced conversion until the next app start. The fetch is now keyed per home visit, so every return to the home screen retries it.

grimen (approved):

PR #3911, opened 2026-07-12
feat(self-custodial): save unilateral-exit recovery bundle on payment events

openoms:
> **Stack 3/3** - split per review: #4016 (wire codec + identity + exporter) -> #4017 (storage, encryption, refresh orchestration) -> this PR. This diff is now only the product layer: refresh hooks, the Recovery Backup screen, settings/navigation wiring, translations, and the clipboard auto-clear. Merge bottom-up; GitHub retargets the stack as bases merge.

Product spec: [blink-specs#26 - PRD: Spark Unilateral Exit for Blink Mobile](https://github.com/blinkbitcoin/blink-specs/pull/26) (this PR implements Phase 1, bundle keep-fresh; see the requirement sync table in section 9).

Seed-only recovery is impossible once Spark operators are offline: current leaves cannot be discovered from the seed alone. Keep a unilateral-exit recovery bundle (spark.unilateral-exit-bundle.v1, consumable by blinkbitcoin/spark-unilateral-exit) fresh on the device so the exit path stays available.

The Breez Spark RN SDK exposes no leaves/bundle API, so the bundle is fetched from the Spark operators directly, mirroring the Rust exporter:

- minimal protobuf wire codec and gRPC-web unary client over fetch (no new dependencies); TreeNodes are stored as the exact wire bytes received
- identity key at m/8797555'/{account}'/0' (mainnet account 1, regtest 0), challenge-response auth with sha256+ECDSA-DER, paged query_nodes(include_parents=true) against the coordinator
- missing ancestors are re-fetched by node id (bypasses the operators' root-skip on legacy mainnet trees) and a bundle with open exit chains is refused

Persistence and sync:

- refresh triggers: SDK payment events (send and receive) debounced 15s, plus a 24h staleness check on startup; single-flight per account
- bundle encrypted at rest with AES-128-GCM keyed from the BIP39 seed (HMAC-SHA256(seed, context)), so the seed alone decrypts any backup; stored under Documents/recovery-bundle-<network>/<accountId>.json and removed on account deletion
- cloud sync follows the seed backup: uploaded to the same provider (iCloud/Google Drive app data) only when the user completed a cloud seed backup, always the seed-encrypted payload regardless of any extra seed-backup password; syncs immediately once cloud backup is set up
- new Settings > Security & Privacy > Recovery backup screen: freshness and leaf count, manual refresh, cloud upload, unencrypted JSON export via share sheet or clipboard

Identity derivation is locked by golden vectors from an independent BIP32 implementation, and the exporter flow (auth, root refetch, open-chain refusal, owner filtering) is tested against a byte-level mocked operator.

## Notes (self-review hardening, commits 95724d2 / 21a5e08 / 7ee0c48)

- All 21 new `RecoveryBundleScreen` strings now ship translations for all 28 locales (fixes the locale-parity `Tests` failure), with typesafe-i18n placeholder annotations matching the existing files.
- Hardening from a three-iteration self-review against docs/pr-review.md: the exporter owner-query paging is capped (misbehaving coordinator cannot loop a background refresh forever) and tolerates truncated by-id rounds; refresh state, single-flight, and the hook's timer cleanup are keyed per (account, network) so mainnet/regtest instance switches cannot mix data; account deletion sweeps both networks' bundle files and state; background refresh and cloud-sync failures are recorded to crashlytics; the screen's actions moved into `use-recovery-bundle-actions` with catch+toast on every action and a distinct loading state.
- Behavior change (shared hook, QA should expect it): `useClipboard`'s `clearAfterMs` timer now survives screen unmount and is shared across instances. A copied seed phrase or recovery bundle is cleared from the clipboard 60s after copying even if the user navigates away, and any newer copy cancels a stale pending clear. This also affects the backup-phrase screens.
- Android share export uses internal storage (`useInternalStorage`), keeping the unencrypted bundle out of the world-readable external cache.
- Deferred tradeoffs (reviewed, intentionally not addressed here): the appdata cloud copy is not deleted on account deletion (neither cloud client exposes a delete API); the encrypted envelope's plaintext metadata is not AEAD-bound (the shared AES-GCM util has no AAD support) though bundle contents are seed-encrypted; the cloud filename embeds the wallet identity pubkey, visible to a cloud-account attacker even when the seed backup uses an extra password.

## Update (commit d5afb03): PRD 4.1 toggles

- **Cloud sync is now opt-in, off by default** (PRD D4/D9), replacing the unconditional auto-sync described above: the encrypted bundle is uploaded only when the user turns on the new "Cloud backup" switch, and the switch is offered only when the cloud seed backup was made **with an extra password** (recorded at backup time as `BackupState.cloudPasswordProtected`). The seed-encrypted bundle must never sit next to an unencrypted seed in the same cloud account. Pre-existing cloud seed backups are treated as unprotected until redone with a password. Turning the switch on syncs the existing bundle immediately.
- **Automatic refresh gains an opt-out** ("Automatic refresh" switch, on by default) to save mobile data; the setting is read when a scheduled refresh fires, so no remount is needed, and manual refresh stays available.
- Per-account settings live in AsyncStorage (`recoveryBundleSettings:<accountId>`) and are swept on account deletion. New strings are translated for all locales.

## Update (commits 9334887 / 5c9be39): two silent-failure fixes

Both defects were introduced in the base PRs but are only reachable from here, so they are fixed in this layer rather than by reopening a merged base.

**The cloud-sync gate closed itself after any later backup.** `isCloudSeedBackupCompleted` keyed on `BackupState.method`, which is last-wins. A user who completed a password-protected cloud backup, turned Cloud backup on, and later ran a manual or keychain backup flipped `method` away from `cloud`, so the gate read as "never backed up to the cloud": the switch disappeared, the screen told them to back up to the provider first, the last-synced timestamp vanished, and every upload stopped - with no warning and no way to turn it back on. The app's own security-score card routes users into exactly that second backup (`retriggerable: true`), so the sequence is routine rather than a corner case. The gate now reads `completedMethods`, the accessor this module already documents for the question and that `use-security-signals` already uses. The password half of D9 is unchanged: `use-cloud-backup` writes an explicit boolean on every cloud completion, so redoing the backup without a password still closes the gate.

**A payment landing mid-refresh was dropped from the bundle.** The single-flight coalescing returned the in-flight run to any later caller, so a payment arriving during a slow refresh was handed a run whose operator fetch had already read the leaves - its own leaves never reached the bundle. The save then stamped `savedAt`, so the 24h staleness path skipped them too, and the manual "Refresh now" button shared the same map and reported success over that stale bundle. Callers now declare what they need covered: the payment event key for a payment-triggered refresh, a fresh value for a user-initiated one. Same key still shares the run, a caller that names no coverage (the staleness sweep) still joins whatever is already fetching, and only a caller naming state the in-flight fetch cannot have read earns a second pass - coalesced so a burst costs one extra pass rather than one per payment. `waitForRefreshesToSettle` now also awaits queued reruns, so a rerun cannot write after the account-deletion sweep.

Both are covered at 100% statements/branches/functions/lines in the files that hold them, and both test sets were mutation-checked: reverting each fix fails 4 and 3 of the new tests respectively. The 30 pre-existing single-flight tests pass unchanged.


openoms:
Thanks for the review - all items addressed in `257136e2f` (or answered below). Point-by-point:

**H1** - Done. The "cancel while not ready" effect is gone; the scheduling effect now clears the pending timer itself when the wallet is not Ready (or there is no account), so exactly two effects touch the timer: the `[accountId, network]` identity cleanup and the scheduler. While consolidating we also closed a related pre-existing race: an in-flight staleness read that settles after the render is gone can no longer arm a timer for a stale closure (cancellation flag in the effect cleanup). Pinned by `use-recovery-bundle-refresh.spec.ts` "cancels a pending refresh when the wallet drops out of Ready before the deadline" (verified failing with the cancellation removed); all 12 pre-existing hook tests pass unchanged. One correction to the write-up: within a single commit React runs *all* cleanups before *all* effect bodies, so the specific interleaving described (an effect body running between two cleanups) cannot occur - the real issue was the fragility you pointed at, which the consolidation removes.

**H2** - Pushback, with the contract now documented in code. The arithmetic `offset += PAGE_SIZE` with `response.offset` used only as a more-pages flag is byte-for-byte the behavior of the spark-unilateral-exit reference exporter, which was validated against the real operators (its test suite replays a real exported bundle). The request offset is a skip count; a cursor interpretation would have broken the reference tool on real wallets. Added a comment at the loop stating exactly this, plus that `MAX_OWNER_QUERY_PAGES` bounds the damage if an operator ever returns a short page with a positive offset.

**H3** - Partially done, remainder deferred as you suggest. `decryptBundleBackupPayload` now cross-checks the plaintext envelope fields (`network`, `walletIdentityPublicKey`, `bundleCreatedAt`) against the decrypted (authenticated) bundle contents and throws a new `envelope-mismatch` error, so a tampered envelope cannot go unnoticed by anything that actually uses the bundle. Pinned by `encryption.spec.ts` "rejects a payload whose plaintext envelope disagrees with the encrypted bundle". AAD support in the shared `app/utils/crypto` util remains a follow-up (it changes a shared surface used by the seed backup too).

**M1** - Done. `storage.ts` header now documents the filesystem-over-keychain rationale: the payload is already AES-GCM encrypted with a seed-derived key, the seed itself lives in the keychain, and bundles can reach hundreds of KB, beyond practical keychain item sizes on Android; AsyncStorage holds only non-secret metadata.

**M2** - Done. `deriveIdentityKeyPair` / `deriveBundleEncryptionKeyHex` now use the async `mnemonicToSeed` (bip39 is noble-backed pure JS here, so the async variant yields between PBKDF2 blocks instead of freezing the JS thread); the async signature is threaded through encryption, exporter, refresh, and the screen actions hook, with all specs updated.

**M3** - Already in the code: the read-modify-write caveat is documented on `markCloudSynced` in `refresh.ts` (landed with `21a5e0898`), including why it self-heals and why single-flight bounds it.

**M4** - Acknowledged. Splitting this PR now would lose the review context, but the remaining scope (stale/fresh status UX, the recovery/import flow and its error states, bundle-size measurement) is planned as separate stacked PRs - see the plan in blink-wip#807.

**L1** - Done: comment added at the cast noting a single leaf cannot exceed 2.1e15 sats, well under `Number.MAX_SAFE_INTEGER`, and the total stays a bigint string.

**L2 / L3 / L4** - Agreed, no action: the `finally` covers L2 (and the spec pins the reset), the writer only builds small messages (L3), and the barrel-less module layout is deliberate (L4).

All recovery-related suites green (identity 7, encryption 7, exporter 12, refresh 19, storage 13, hook 13, actions 15, delete-account 13), tsc and eslint clean.

grimen (changes requested):
## Findings

### High

**1. A short owner-query page with a positive offset silently skips nodes (blocking).**

`exporter.ts:148-155` advances `offset += PAGE_SIZE` unconditionally, regardless of how many nodes the page actually returned, so a page of fewer than 100 nodes that still signals "more pages" causes the unreturned nodes to be skipped and the loop to finish normally. This is the one path where an incomplete bundle is persisted as if complete, since `refresh.ts:133-145` saves whatever comes back without an error.

*Example:* the coordinator returns 60 nodes for `offset=0, limit=100` with a positive response offset, the next request uses `offset=100`, and any AVAILABLE leaves among nodes 60-99 are missing from the exit bundle with no warning.

*Proposal:* advancing by `response.nodes.size`, or treating a short page with a positive offset as an error, would preserve the "incomplete means refuse" property the rest of the module maintains.

**2. The bundle file is written in place, so an interrupted write destroys the last good copy (blocking).**

`storage.ts:31-33` calls `RNFS.writeFile` on the target path, which truncates before writing, and the AsyncStorage `savedAt` stamp still reports the result as fresh afterwards.

*Example:* the OS kills the app mid-write during a background refresh, leaving truncated JSON; recovery then needs a successful re-fetch from the operators, which is exactly the connectivity absent in the unilateral-exit scenario this file exists for, and the staleness path won't retry for 24h.

*Proposal:* writing to a `.tmp` sibling and then `RNFS.moveFile` over the target would make the replacement atomic, so a crash leaves the previous bundle intact.

**3. A cycle in parent pointers is classified as a closed chain rather than refused.**

The per-leaf `visited` guard in `findMissingAncestors` (`exporter.ts:203-215`) exits the walk when it revisits a node without recording anything as missing, so a chain that never reaches a root passes both `IncompleteChain` checks.

*Example:* a misbehaving coordinator, the adversary the paging caps at lines 50-58 explicitly defend against, returns `leaf → A`, `A.parent = B`, `B.parent = A`; the walk terminates silently and the bundle is saved despite being unusable for exit.

*Proposal:* throwing `IncompleteChain` when `visited.has(parent.id)` would match the docstring's promise that every leaf's ancestor chain is verified closed.

### Medium

**4. The shared clipboard clear timer can wipe content copied through the direct API.**

Making `pendingClearTimer` module-level (`use-clipboard.ts:20`) lets a pending clear survive unmount, but the callers that write the clipboard directly rather than through the hook, `use-payment-actions.ts:32` and the developer screen, never cancel it. The comment's reasoning that "ANY newer copy makes an older pending clear obsolete" holds only for copies going through the hook.

*Example:* the user copies their seed phrase or the bundle JSON, arming a 60s clear, then navigates to Receive within that minute and copies a Lightning invoice, and the older timer empties the clipboard with the invoice in it.

*Proposal:* exporting a `cancelPendingClipboardClear()` from the hook module and calling it at the direct `Clipboard.setString` sites would restore the invariant the comment describes.

**5. gRPC-web accepts compressed frames and responses with no trailers frame.**

`grpc-web.ts:136` inspects only the trailer bit (`0x80`), so a frame carrying the compressed flag `0x01` reaches the protobuf decoder as if it were plaintext; separately the loop returns successfully when no trailer frame is present, and `parseFrames` (line 37) silently drops a trailing remainder shorter than 5 bytes.

*Example:* a proxy compresses the response and the exporter fails with a cryptic `proto: unsupported wire type`, or the stream is cut after the message frame and a call whose real non-OK `grpc-status` never arrived is treated as success.

*Proposal:* throwing on `flag & 0x01`, requiring a trailers frame, and erroring on a sub-5-byte remainder would each be a couple of lines and keep protocol violations diagnosable.

**6. Cloud upload failures reach Crashlytics only as breadcrumbs, losing the original error.**

`cloud.ts:67-73` discards the caught error and returns only a reason enum, and the sole consumer at `refresh.ts:116` logs it with `crashlytics().log`, visible only when attached to a later recorded error. The PR describes cloud-sync failures as recorded, which currently holds only for the parse-corruption case at `refresh.ts:105`.

*Example:* an iCloud write error starts failing for every user on a new iOS version and produces no standalone Crashlytics events, with the reason reported as "unknown".

*Proposal:* auth failures are genuinely expected and fine as breadcrumbs, but carrying the original error through and recording non-auth reasons would make a systemic outage visible.

**7. An in-flight refresh can recreate files for an account that was just deleted.**

`use-delete-account.ts:96-109` sweeps the bundle files, state, and settings, but a `runRefresh` already past its fetch keeps going and writes at `refresh.ts:135` and `:145`; nothing consults the `inFlight` map or cancels it.

*Example:* a background refresh is mid-fetch when the user deletes the wallet, and seconds later the bundle file and AsyncStorage state reappear under the deleted account's id, which nothing cleans up since the account is gone from the registry.

*Proposal:* checking a deleted-account marker before those writes, or re-running the sweep once in-flight runs settle, would keep the deletion complete; the payload is encrypted with a discarded seed, so this is about the sweep's guarantee rather than exposure.

**8. The immediate cloud sync doesn't fire when a passwordless backup is upgraded to a password-protected one.**

The effect at `use-recovery-bundle-refresh.ts:170-187` depends on `cloudSeedBackupActive`, derived from `isCloudSeedBackupCompleted` (line 56), the gate that ignores password protection, while the upload itself requires `cloudPasswordProtected`.

*Example:* a user with an existing passwordless cloud backup redoes it with a password, the boolean was already `true` so the effect never re-runs, and the promised immediate upload waits for the next payment, the 24h pass, or a manual toggle.

*Proposal:* keying the effect on `isPasswordProtectedCloudSeedBackup(backupState)` would align the trigger with the condition the sync actually checks.

**9. A `Pending` to `Succeeded` transition for the same payment doesn't reschedule the refresh.**

Both events feed `lastReceivedPaymentId` with the same `payment.id` (`sdk-events.ts:12-15`), so the second is a no-op state update, and the ref dedupe at `use-recovery-bundle-refresh.ts:135-141` would skip it regardless.

*Example:* a payment goes pending, the 15s debounce snapshots the pre-settlement leaf state, and settlement a minute later produces no new refresh, so the bundle describes the wallet as it was before settlement until an unrelated event or the 24h fallback.

*Proposal:* including the event tag alongside the id in the dedupe key would let settlement schedule its own refresh.

**10. A failure in the post-save cloud step reports a successful refresh as failed.**

`syncExistingBundleToCloud` runs inside `runRefresh`'s `try` (`refresh.ts:147`), so an AsyncStorage or RNFS rejection there resolves the whole run as `{ success: false }` even though the bundle and state were already written.

*Example:* the backup-state read throws after a good save, the hook records a `recovery-bundle-refresh` error, and the screen toasts "refresh failed" while a fresh valid bundle sits on disk.

*Proposal:* moving the sync call outside the try, or giving it its own catch, would keep the result reflecting what actually determines bundle freshness.

### Low

**11. The bundle file isn't excluded from iOS device backups.**

It lives under `DocumentDirectoryPath` (`storage.ts:20-24`), which iOS includes in iCloud device backups, and no `NSURLIsExcludedFromBackupKey` exists anywhere in the repo; Android is covered by `allowBackup="false"`.

*Example:* a user makes a passwordless iCloud seed backup, permitted for the seed itself, and the seed-decryptable bundle still reaches the same Apple account through the device backup, which is the adjacency D9 exists to prevent.

*Proposal:* for the implementer to confirm whether D9 is meant to cover the implicit device-backup path; if it is, `RNFS.mkdir` accepts the exclusion option for the directory.

**12. Decoded varints aren't masked to 64 bits.**

`wire.ts:81-93` keeps all 7 bits of the tenth byte at shift 63, so a decoded value can exceed 2^64 where canonical parsers truncate.

*Example:* a hostile operator encodes a TreeNode's `value` as a 10-byte varint with high bits set in the last byte, and `Number(leaf.valueSats)` at `exporter.ts:317` yields an absurd figure that flows into the displayed leaf value and the summed total.

*Proposal:* masking with `& 0xffffffffffffffffn` before returning would match what `decodeQueryNodesResponse`'s `BigInt.asIntN(64, …)` already assumes.

**13. The `NoLeaves` message embeds the wallet identity pubkey in an error string.**

`exporter.ts:258-263` bakes the identity into `Error.message`, and only the `isBenignExportError` filter at `use-recovery-bundle-refresh.ts:110` keeps it out of `recordError`, while other export errors are recorded with their messages at line 117.

*Example:* a later refactor narrows or drops that benign filter and the persistent wallet identity pubkey starts flowing to Firebase, linking the on-chain identity to the Crashlytics installation.

*Proposal:* carrying the pubkey as a structured field on the error, consumed by the UI rather than embedded in the message, would remove the footgun.

## Test Coverage Gaps

- `cloud.ts` has no spec at all: the platform branch, the silent Drive sign-in flow, and the `"reason" in err` duck-typed error mapping are untested.
- `storage.spec.ts` mocks react-native-fs down to `DocumentDirectoryPath` only, so `saveEncryptedBundleFile`, `loadEncryptedBundleFile`, and `deleteRecoveryBundleFile` never actually execute.
- The exporter's short-page skip (finding 1), the `visited` cycle guard (finding 3), and the rounds-exhausted throw at `exporter.ts:287-293` are untested; the mock only ever emits full logical pages.
- `grpc-web.spec.ts` covers no compressed flag, missing-trailers, truncated-remainder, or timeout/abort case, and `wire.spec.ts` has no varint-too-long or overflow-bit test.
- `use-recovery-bundle-refresh.spec.ts` has no unmount test and no network-switch test, despite the load-bearing comment at lines 67-69 explaining why the cleanup is keyed that way.
- `encryption.spec.ts` has no direct GCM tamper test (flipping a byte of `data` and asserting `decrypt-failed`); auth failure is covered only indirectly through the wrong-seed case.
- Low-S canonicalization of the auth signature is inherited from a transitive noble default with nothing pinning it; an assertion that S < n/2, or a golden signature vector since signing is deterministic, would survive a dependency bump.
- `use-delete-account.spec.ts` never asserts `removeRecoveryBundleSettings` was called.

## Code Smells

- The comment at `exporter.ts:151-153` says `MAX_OWNER_QUERY_PAGES` "bounds the damage" from a short page with a positive offset, but the cap bounds looping rather than skipping, and a skip terminates normally and silently (finding 1).
- `firstField` (`wire.ts:134-137`) takes the first occurrence of a repeated field where proto3 specifies last-wins; it fails closed since auth simply fails, but the canonical challenge re-encode depends on matching the server's parse.
- `handleCopy` (`use-recovery-bundle-actions.ts:203`) is the only export action without an in-flight guard or loading state, so its button can be tapped repeatedly while the decrypt runs.
- `readRecoveryBundleState` (`storage.ts:80-88`) validates `savedAt`, `leafCount`, and `totalSats` but not `bundleCreatedAt` or `cloudSyncedAt`, so a garbage `cloudSyncedAt` would flow into the UI and the sync effect's truthiness check.
- `sortedByIdAsc` uses `localeCompare` (`exporter.ts:295-296`), making node ordering ICU-dependent; a plain `<` comparison would keep bundles byte-reproducible across devices.

openoms:
All findings addressed in 8ae674e; the branch is also rebased onto main (conflicts resolved: the i18n both-added blocks, MigrationBlockerProvider vs RecoveryBundleListenerMount in app.tsx, and the cloud-backup completion refactor - `cloudPasswordProtected` now travels through `useCompleteBackup`'s new `backupOptions` into both `setBackupCompleted` and the migration path's `markBackupCompletedFor`).

### High

1. **Short owner-query page skips nodes** - the paging loop now advances the request offset by `response.nodes.size` instead of `PAGE_SIZE`, so a short page with a positive offset re-requests the unreturned tail. The misleading "bounds the damage" comment is rewritten. Test: 60-node page with a positive offset, asserting the next request goes out at offset 60 and all 100 leaves land in the bundle.

2. **In-place bundle write** - `saveEncryptedBundleFile` now writes to a `.tmp` sibling and renames over the target (with an unlink first, since iOS `moveFile` refuses to overwrite). A crash in the unlink-to-rename gap is also covered: `loadEncryptedBundleFile` finishes the interrupted rename from the tmp file, and `deleteRecoveryBundleFile` sweeps a stranded tmp. Tests run against an in-memory RNFS with iOS moveFile semantics: round-trip, failed-write-keeps-old-copy, tmp recovery, tmp sweep.

3. **Parent-pointer cycle classified as closed** - `findMissingAncestors` now throws `IncompleteChain` when a walk revisits a node. Test: `leaf -> a`, `a.parent = b`, `b.parent = a` rejects with `incomplete-chain`.

### Medium

4. **Clipboard clear timer wiping direct copies** - `use-clipboard.ts` exports `cancelPendingClipboardClear()`, called at both direct `Clipboard.setString` sites (receive-screen payment actions and the developer screen). The invariant comment now states the obligation for bypassing callers.

5. **gRPC-web laxness** - a frame with the compressed flag throws (never negotiated, so it is a protocol violation), a body whose frames end without a trailers frame throws unless `grpc-status` came in the HTTP headers, and a trailing remainder shorter than a frame header throws instead of being dropped. Tests for all three plus the 30s timeout/abort path.

6. **Cloud failures invisible** - `attemptSilentCloudUpload` now carries the original thrown error in its failure result; `syncExistingBundleToCloud` keeps `Auth` as a breadcrumb (expected: cloud never linked) and `recordError`s everything else with the original error under `recovery-bundle-cloud-sync`. New `cloud.spec.ts` covers the platform branches, silent Drive sign-in, and the reason/error mapping.

7. **Refresh recreating deleted account files** - `refresh.ts` gains a `markAccountDeletedForRefresh` marker (account ids are random UUIDs, never reused, so process-lifetime membership is safe): a marked account fails fast on new refreshes and an in-flight run re-checks it after the fetch, right before persisting. The delete hook marks before sweeping, and after `waitForRefreshesToSettle` resolves it re-sweeps in the background to close the residual mid-write race without blocking the delete UX.

8. **Password upgrade not triggering sync** - the hook effect now keys on `isPasswordProtectedCloudSeedBackup(backupState)`, matching the condition the upload actually checks. Tests: passwordless backup does not sync; flipping to password-protected fires the immediate upload.

9. **Pending -> Succeeded not rescheduling** - the wallet context now exposes `lastPaymentEventKey` (`${tag}:${id}`, built by `paymentEventKey` in sdk-events) for the refresh hook's dedupe, so settlement of an already-seen payment schedules its own refresh. `lastReceivedPaymentId` (the id alone) is still exposed for the id-keyed consumers (`use-payment-request`, `use-auto-convert-listener`), whose behavior is unchanged.

10. **Cloud step flipping a successful refresh to failed** - the post-save cloud sync now runs in its own try/catch: a throw there is recorded under `recovery-bundle-cloud-sync` and the refresh still resolves `{ success: true }` with the persisted state.

### Low

11. **iOS device backups** - the bundle directory is created with `NSURLIsExcludedFromBackupKey: true`. D9's reasoning (seed-decryptable payload next to a possibly-passwordless seed backup in the same Apple account) applies to the implicit device-backup path too, so exclusion is the right default; flag if you want it in backups instead.

12. **Varint overflow** - decoded varints are masked to 64 bits, matching canonical parsers and the `BigInt.asIntN(64, ...)` downstream. Tests: 10-byte varint with overflow bits decodes to `2^64 - 1`; an 11-byte varint still throws.

13. **Identity pubkey in NoLeaves message** - the pubkey moved out of `Error.message` into a structured `details.walletIdentityPublicKey` field on `RecoveryBundleExportError`, so recording the message can never leak it. Test asserts the hex is absent from the message and present in the field.

### Code smells

- `firstField` replaced by `lastField` (proto3 last-wins) everywhere, with a test.
- `handleCopy` now shares the `exporting` in-flight guard with `handleShare`.
- `readRecoveryBundleState` validates `bundleCreatedAt` and `cloudSyncedAt` too (test matrix extended).
- `sortedByIdAsc` uses a plain code-unit comparison instead of `localeCompare`.

### Test coverage gaps

All listed gaps are covered: new `cloud.spec.ts`; storage spec executes the real file functions against an in-memory RNFS; exporter short-page/cycle/rounds-exhausted tests; grpc-web compressed/missing-trailers/truncated-remainder/timeout tests; wire varint-too-long and overflow-mask tests; refresh-hook unmount and network-switch tests; a direct GCM tamper test (bit-flipped ciphertext -> `decrypt-failed`); a determinism + low-S assertion on the auth signature (S <= n/2, parsed from the DER); and `use-delete-account.spec` now asserts `removeRecoveryBundleSettings`, the deletion marker ordering, and the post-settle re-sweep.

openoms:
working on to break this down to multiple PRs

openoms:
All findings addressed, and the PR is now split as proposed.

### PR split (blocking)

Done, using stacked PRs exactly along your A/B/C cut:

1. #4016 - wire codec + identity + exporter (base: `main`)
2. #4017 - storage, encryption, cloud, refresh orchestration (base: #4016's branch)
3. this PR - hooks, screen, navigation, i18n, clipboard (base: #4017's branch)

Each level typechecks and its test suites pass standing alone (A: 54 tests, B: 489 including the onboarding/backup-state consumers), and the top of the stack is byte-identical to the reviewed tree plus the fixes below. Merging bottom-up auto-retargets the next PR.

### High

**H1 (CI)** - two causes, both fixed: the previous push had prettier/`jest/valid-expect` errors in three spec files (fixed), and `check:codegen` fails because the backend schema gained `DepositFeeTier` - `main` is stale against the live schema too, so every PR currently trips it. A `chore(graphql)` regeneration commit rides at the bottom of #4016.

**H2 (clipboard)** - grep documented: `Clipboard.setString` outside the hook exists in exactly two places, `use-payment-actions.ts` (receive screen) and `developer-screen.tsx`, and both now call `cancelPendingClipboardClear()` first (added in the previous review round). The hook's comment states the obligation for any future direct caller, and `use-clipboard.spec.ts` covers the cancel path.

**H3 (deletedAccounts growth)** - fixed with your option (a): `unmarkAccountDeletedForRefresh` drops the id once the post-deletion re-sweep completes (every in-flight run has settled and nothing can start a new refresh - the mnemonic and registry entry are gone), so the set is bounded by concurrent deletions. Covered in `refresh.spec.ts` and the delete-account spec asserts unmark ordering after the re-sweep.

### Medium

**M1** - agreed with your own re-read: single-flight shares the promise, and the tmp recovery covers the crash gap.

**M2 (unbounded total fetch time)** - implemented: `fetchRecoveryBundle` now has a 2-minute wall-clock budget; the remaining budget is threaded into every gRPC call's timeout (capped by the per-call 30s), and an exceeded budget throws a `budget-exceeded` export error. `totalBudgetMs` is injectable, and a test pins the mechanism.

**M3** - agreed with your withdrawal; the existing `varint too long` + mask tests pin both behaviors.

**M4 (filename identity leak)** - still acknowledged as a deferred tradeoff; will open a tracking issue when the stack lands so it references the final file locations.

### Low

**L1** - guard added, and your instinct was better than the proposal: the value decodes as an *unsigned* varint, so a hostile negative arrives as a near-2^64 positive that a `< 0` check misses. The guard rejects anything outside `[0, 21M BTC]` with an `invalid-node` error; the test encodes a negative value and asserts refusal.

**L2** - `formatWhen` memoized on `locale`.

**L3** - yes, deliberate; comment stands.

**L4** - startup delay raised to 15s with a comment about the cold-start contention window.

### Note on the red Android/iOS checks

Those fail in ~20s with exit 127 on `nix develop -c yarn install` - `nix` is missing on the self-hosted E2E runner. Infra, not this branch.

openoms:
Addressed HIGH #1 — Done in `49b60c40e`.

The active debounce now records which payment event owns it. If readiness loss, unmount, account change, network change, or rescheduling cancels that timer before it fires, the event claim is released so the same latest event is retried on the next Ready render. Once the timer fires, the event remains handled and reconnect re-delivery stays deduplicated.

The updated regression test first reproduced the missed refresh and now verifies that a Loading/Ready cycle before the deadline schedules a fresh 15s debounce. It also preserves coverage that an already-fired event is not refreshed twice.

Verification: 147 recovery-bundle and refresh-hook tests pass; targeted ESLint, Prettier, and diff checks pass.

@blink-claw-bot review

grimen:
Pushed four small follow-up commits from the review:

- c6bceff7c35f5fc7e3c0b3ae4606c8f86b34346b — the `lastReceivedPaymentId` derivation moved into `sdk-events.ts` as `paymentIdFromEventKey`, with a round-trip spec pinning ids that themselves contain `:`.
- c112d36fa98e310eee6c836ad3b7ba3779bdf6ef — Copy JSON now has its own loading state: `exporting` became per-action `sharing`/`copying` flags (still mutually exclusive, so no concurrent decrypts), and the screen wires a spinner to each button.
- ab09644b81fb789528297b71ac1496f1c2f9595f — turning the Cloud backup switch on now runs the full `handleCloudUpload` flow (silent attempt, then interactive sign-in fallback, with toasts) instead of the silent-only sync, so an expired Drive/iCloud token no longer makes the opt-in look like a no-op. The D9 gate is still enforced inside via `isCloudSyncAllowedFor`, reading the just-persisted setting; with no bundle saved yet the toggle stays quiet as before.
- 8108d1bf250ff3044b2884ebef2c15e353de009c — a render spec for `RecoveryBackupScreen` covering the three-way cloud-section gating (no cloud seed backup / passwordless / password-protected), upload-button visibility, switch wiring, and the per-button loading cues.

Targeted suites pass locally (89 tests across the five affected spec files); nothing else in the PR was touched.

grimen (dismissed):

grimen:
@openoms implementation PRs seems to be in good shape now, I will create another PR for the UI implementation.

grimen (dismissed):

grimen (dismissed):

esaugomez31 (approved):

grimen (changes requested):
## Findings

**1. Redoing the cloud seed backup without a password leaves the already-uploaded bundle next to the now-plaintext seed.**

D9 is enforced only at upload time: `isCloudSyncAllowedFor` (`refresh.ts:89-95`) gates future uploads, but nothing acts when the gate goes from open to closed. The seed backup overwrites the same file (`getCloudBackupFilename(instance, identityPubkey)`), so a passwordless redo replaces the protected seed with a plaintext one (`use-cloud-backup.ts:131-136`) and only then records `cloudPasswordProtected: false` (`:152-157`), while `blink-spark-recovery-bundle-<net>-<pubkey>.json` stays in the same appdata and `settings.cloudSync` stays `true`. The screen then hides the "Cloud backup: <time>" line and shows `cloudSyncNeedsPassword` ("requires your wallet backup there to be protected with a password"), which reads as if nothing is in the cloud.

*Example:* the user turns Cloud backup on, the bundle uploads, and weeks later they redo the cloud backup with the password toggle off; the cloud account now holds a plaintext seed and the bundle it decrypts, which is exactly the pairing D9 exists to prevent. If they later redo it with a password again, uploads silently resume because `cloudSync` was never reset.

*Proposal:* one option is to overwrite or delete the bundle file in the same provider session before a passwordless seed upload (the session is already open there), and to reset `cloudSync` to `false` on any passwordless cloud completion; if a delete API is out of scope for now, a warning in the passwordless branch plus the `cloudSync` reset would at least make it a user decision. Writing `cloudPasswordProtected: false` before the upload rather than after would also make the flag fail closed if the app dies between the upload and the unawaited `persist`. The deferred-tradeoffs note in the description covers account deletion only, not this path.

**2. The D9 gate ignores a password-manager backup, which syncs the raw mnemonic into the same cloud account.**

`isPasswordProtectedCloudSeedBackup` (`backup-state.tsx:158-159`) checks that Cloud is in `completedMethods` and the cloud flag, but not `BackupMethod.Keychain`, whose Android path is `signUpWithPassword({ username: walletIdentifier, password: mnemonic })` (`use-credential-backup.ts:53`) into Google Password Manager, and whose iOS path is `setInternetCredentials(..., { cloudSync: true })` (`:71`). Commit 652458916 moved the gate to `completedMethods`, so a keychain backup made after a protected cloud backup now keeps the gate open, where the old last-wins `method` closed it; the reverse order was already open.

*Example:* on Android, a user with a password-protected Drive backup follows the security-score card into the password-manager backup (`retriggerable: true`); anyone who gets into the Google account reads the mnemonic at passwords.google.com and the bundle from Drive appdata.

*Proposal:* one option is to exclude `Keychain` from the gate at least on Android and explain why on the screen; iCloud Keychain is end-to-end encrypted, so whether iOS counts as "unencrypted seed in the cloud" is a call for the PRD owner.

**3. A failed background refresh is never retried, and the staleness check hides the gap.**

When `refreshRecoveryBundle` fails, the hook only records the error (`use-recovery-bundle-refresh.ts:133-143`); `lastHandledEventKeyRef` still holds the key so the event is not rescheduled (`:165`), and the staleness path compares against `savedAt` of the last success (`:176-180`) and only runs when the effect deps change.

*Example:* payment A's refresh saves at T; payment B's refresh (or the rerun queued for it) fails on a transient network error; the bundle lacks B's leaves while `savedAt = T` reads as fresh, so nothing refreshes until another payment or an app start more than 24h after T. This is the same silent-staleness class that 5c9be39 fixed for coalescing.

*Proposal:* one option is to persist a "dirty since" marker that the staleness check treats as stale until a successful save, or a bounded retry with backoff that clears the handled-key ref on failure.

**4. Clipboard export of secrets now outlives the screen, and the clear depends on the JS timer running.**

The previous clear-on-unmount is gone (`use-clipboard.ts`), so in-app navigation no longer clears immediately, and the 60s clear is a plain `setTimeout`, which does not fire while the app is backgrounded and never fires if the process is killed. `Clipboard.setString` also sets no `EXTRA_IS_SENSITIVE` on Android 13+ and no `localOnly`/expiration on iOS, so keyboard clipboard history and Universal Clipboard pick it up. This also changes the seed-phrase copy (`use-view-backup-phrase.ts`, `use-backup-phrase.ts`), and the description's "cleared 60s after copying even if the user navigates away" does not hold once the user switches apps.

*Example:* the user taps Copy JSON (or copies the seed phrase), switches to a notes app, and iOS kills Blink in the background; the content stays on the clipboard with no expiry and syncs to their other Apple devices.

*Proposal:* one option is to also clear on `AppState` going to background while a clear is pending, and to set the OS-level flags (`.localOnly` + `.expirationDate`, `EXTRA_IS_SENSITIVE`) through a small native helper; dropping clipboard export for the bundle, which runs to hundreds of KB, is another.

**5. The Recovery Backup screen has no local auth gate, unlike the seed-phrase screen.**

`recovery-backup-screen.tsx` exposes the unencrypted share/copy export and both switches without `useLocalAuthGate`, which `view-backup-phrase-screen.tsx:39` requires, and the only warning before export is the banner at `:162-166`.

*Example:* someone with brief access to an unlocked phone shares the plaintext bundle (balance, leaf structure, identity pubkey) to themselves, or turns off auto-refresh and cloud backup, quietly degrading the exit path.

*Proposal:* gating the screen, or at least the export actions and switches, with `useLocalAuthGate({ required: true })` as the phrase screen does would keep the two exports consistent.

## Test Coverage Gaps

- `refresh.spec.ts:351` "makes waitForRefreshesToSettle await a queued rerun too" still passes with `reruns` removed from the `allSettled`, because the fetch count is already 2 when the wait resolves; holding the rerun's fetch on a deferred and asserting the wait is still pending would make it bite.
- Nothing marks an account deleted while a rerun is queued and asserts no save or state write follows, and `use-delete-account.spec.ts` doesn't assert `waitForRefreshesToSettle` is called after the mark.
- `use-sdk-lifecycle.spec` never asserts `lastPaymentEventKey`; reverting the lifecycle to `extractPaymentId` fails no test, and the hook's Pending → Succeeded test (`:200`) doesn't assert the second call's `coverageKey`.
- No test covers a failed background refresh and its (non-)retry (finding 3), or a passwordless cloud redo after the bundle was synced (finding 1).
- The `cancelPendingClipboardClear` calls in `use-payment-actions.ts` and the developer screen are unasserted.

## Code Smells

- `lastPaymentEventKey` is never reset on account switch (`use-sdk-lifecycle.ts` connect effect) while the hook resets its dedupe ref (`:98-103`), so every switch after any payment schedules a refresh of the new account with the previous account's key as `coverageKey`, bypassing its 24h freshness check; the comment there says the reset is because payment ids are per-account.
- "Direct callers of Clipboard.setString must cancel the timer first" is a comment-only convention; a `no-restricted-syntax` rule on `Clipboard.setString` outside `use-clipboard.ts` would enforce it.
- `handleRefresh` (`use-recovery-bundle-actions.ts:147-156`) toasts a non-benign failure without recording it, while the background path records the same error.
- `syncExistingBundleToCloud` and `markCloudSynced` run outside the in-flight tracking and never consult `deletedAccounts` (hook `:204`, actions `:267`, `:304`); a run past its deleted check can still upload a deleted account's bundle.
- Test names still say 5s ("schedules a refresh 5s after startup", "arms a 5s timer") while `STARTUP_DELAY_MS` is 15s.
- `recovery-backup-screen.spec.tsx` prints 24 `act()` warnings from `AccountRegistryProvider`.

PR #3913, opened 2026-07-15
fix(location): resolve the ipinfo country lookup via the /me endpoint


esaugomez31:
## Summary

Country detection was resolving to El Salvador for effectively every user, which wrongly disabled the non-custodial Dollar Balance (it treats `SV` as a restricted region). The cause is a broken IP geolocation call, fixed here with a one-line endpoint change.

## Root cause

The authenticated ipinfo adapter requested `https://api.ipinfo.io/lite/` (no IP, no `/me`), which is not a real endpoint:

```
{ "error": "Endpoint not found. Did you mean /lite/me or /lite/8.8.8.8?" }
```

In production builds (where `IPINFO_API_KEY` is set) this adapter always failed, so detection fell through to the free fallback services and, when those were rate-limited/unavailable at scale, to `DEFAULT_COUNTRY_CODE = "SV"` in `use-device-location`. Result: El Salvador for everyone, regardless of their real location.

## Fix

`app/utils/ip-country-lookup.ts`: `api.ipinfo.io/lite/` → `api.ipinfo.io/lite/me` (the caller's-own-IP endpoint). The primary adapter now returns the real country directly, so detection no longer depends on the fragile fallback chain.

## Verification

- `curl` against the old URL returns the 404 error above; `/lite/me` returns the correct `country_code`.
- On device with a VPN in France: with the broken URL, detection fell back to El Salvador; with `/lite/me`, it correctly showed France.
- Unit tests assert the adapter hits `https://api.ipinfo.io/lite/me` exactly and never `/lite/` without the `/me` segment.

### DIFF
--- a/__tests__/utils/ip-country-lookup.spec.ts
+++ b/__tests__/utils/ip-country-lookup.spec.ts
@@ -132,7 +132,7 @@ describe("DEFAULT_ADAPTERS key-gated behaviour", () => {
     )
   })
 
-  it("uses api.ipinfo.io/lite with Bearer header when IPINFO_API_KEY is set", async () => {
+  it("calls the api.ipinfo.io/lite/me endpoint with the Bearer header when IPINFO_API_KEY is set", async () => {
     mutableConfig.IPINFO_API_KEY = "test-ipinfo-key"
     // eslint-disable-next-line camelcase
     mockedAxios.get.mockResolvedValue({ data: { country_code: "DE" } })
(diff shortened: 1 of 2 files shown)


esaugomez31:
* **#3913** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3913?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3913?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (approved):

PR #3916, opened 2026-07-15
fix(android): Dollar balance and transfer features unavailable for VPN users on Android v3.0.4

### LINKED ISSUE #3907: fix(android): Dollar balance and transfer features unavailable for VPN users on Android v3.0.4
## Bug Report

**Version:** Android v3.0.4
**Affected users:** All VPN users on Android
**Reporter:** @designsats

### Description

Dollar balance and transfer features are disabled/unavailable for users connected via VPN on Android v3.0.4. The Dollar wallet section shows "not available in your region" even though the user should have access.

### Steps to Reproduce

1. Connect to a VPN on Android
2. Open Blink v3.0.4
3. Observe the home screen — Dollar wallet shows "not available in your region"
4. Transfer functionality is also affected

### Expected Behavior

Dollar balance and transfer features should be available regardless of VPN usage (region restrictions should be based on account registration, not current IP/location).

### Actual Behavior

Dollar wallet shows "not available in your region" and transfer features are disabled when using a VPN.

### Screenshot

Dollar unavailable

_(Screenshot shows account with Bitcoin balance of 2,915 SAT / ~$1.84, Dollar section showing "not available in your region")_

### Priority

High — affects all VPN users on Android.

/cc @esaugomez31

grimen:
Closes #3907

## Problem

Dollar/stable balance and transfer show **"not available in your region"** whenever a VPN is on, and recover when it's off. Reported on Android v3.0.4; affects any VPN whose exit lands in a blocked list (EU exits are the common trigger).

## Root cause

Region gates derived the country from **the current network path**:

- **No-phone (self-custodial) accounts**: country = live IP geolocation per session → follows the VPN exit.
- **Phone accounts**: primary check used the phone country, but the sync hooks did a **second IP lookup** (`useIpCountryCode`) even when the phone country was allowed, and could latch the restriction off a VPN exit.
- **The persisted restriction flags were set-only** — no code path ever cleared them. One bad observation latched forever (currently masked only by the `dollarRestrictionCacheEnabled` remote kill-switch), and via #3910 pushed latched users into force-converting their Dollar balance.

## Solution

Restrictions now follow the **account's registration country**, never the current IP:

1. **New `useRegistrationCountry`** (`app/hooks/use-registration-country.ts`), the single country source for the gates:
   - phone country (registration) — also backfills persistent state so it survives a switch to self-custodial;
   - else persisted **first-seen country** (`firstSeenCountryCode`, new persistent-state module);
   - else **one IP lookup ever**, persisted on success; failure persists nothing (retry next session) and reports `trusted: false` — gates fail open on unknown.
   - **VPN-aware persistence**: alongside the country lookup, proxycheck.io's proxy/VPN detection is consulted (new `detectAnonymizingIp*` in `ip-country-lookup.ts`, session-cached). A country resolved through a **flagged anonymizer is used for the session but never persisted** — persisting retries on a later session, so a VPN-always-on user behaves like today (live) instead of getting a VPN exit country pinned forever. Only a confirmed flag defers; if detection is unavailable, persistence proceeds (an external detector's uptime must not gate registration).
2. **Gates rewired** (`use-dollar-balance-restricted`, `use-transfer-blocked`): read the new hook; the IP double-check is deleted (`useIpCountryCode` removed entirely).
3. **Latch is self-correcting**: sync hooks now *clear* the persisted flag when trusted evidence says the country is allowed (new `without*` setters, all four flags). Clearing requires trustworthy evidence on **both sides of the comparison**: `!loading && trusted` for the country, and the new `remoteConfigLoaded` feature flag for the blocked lists — a failed remote-config fetch leaves hardcoded defaults behind (`remoteConfigReady` flips true even then), and a country missing from the defaults must never clear a flag set from the real config. Clearing intentionally ignores the `dollarRestrictionCacheEnabled` kill-switch so stale flags heal while it's active. Set path unchanged. Account-type-switch double-write preserved; clears mirror it.
4. Settings footer and `usePaymentRequest` read the same source, so the footer can no longer contradict the gate.
5. **Observability + ops brake**: analytics events for every registration-country pin, VPN-deferred pin, and self-heal (`app/utils/analytics.ts`), and a `restrictionSelfHealEnabled` remote-config kill-switch (default on) that freezes the clearing behavior without a release. Details under Risks #2.

**Remediation is automatic**: previously latched users self-heal on next home-screen mount; the forced-conversion prompt stops firing.

## Trade-offs (accepted)

- A wallet **first opened/recovered behind an undetected anonymizer** can bake in that country until app data is reset (first-seen = registration by definition). Detected VPNs/proxies are already deferred (see VPN-aware persistence above); the residual risk is proxycheck misses.
- The #3847 anti-bypass (foreign phone, physically in a blocked region) is removed by design.
- **Custodial with a phone number** (the common case): unaffected by device/reinstall/VPN state — the phone country always wins and is re-fetched from the backend on any device.
- **Email-only custodial accounts** have no phone, so they use the first-seen country like self-custodial wallets: stable against VPN toggling (the fix still applies), but re-derived on reinstall/new device. The API exposes no registration-country field for them; adding one server-side would be the proper long-term fix (follow-up).

## Risks

| # | Risk | Probability | Severity | Mitigation (short) |
|---|------|-------------|----------|--------------------|
| 1 | Upgrade-moment bake-in of a VPN country | ~~High~~ → **Low** (only undetected anonymizers) | **High** | **Mitigated in-PR**: VPN-flagged IPs never persist first-seen |
| 2 | Compliance exposure (bypass by design) | Low | Medium–**High** (regulatory posture dependent) | **Partially mitigated in-PR** (no pinned bypass via detected VPNs; telemetry; self-heal kill-switch); sign-off still required |
| 3 | Device-global registration state crosses accounts | Low | Low–Medium | Accept; per-account state if support cases appear |
| 4 | Creation vs. usage gate divergence | Medium | Low | Support/FAQ note |
| 5 | Operational semantics shifted (footer, kill-switch) | High (will be noticed) | Low | Ops/support changelog note |
| 6 | Version skew (old clients keep live-IP behavior) | Certain | Low–Medium | Organic updates, then `minSupported` bump |

1. **Upgrade-moment bake-in.** Existing self-custodial installs have no persisted country; the first open after updating *is* the first-seen moment. A VPN-always-on user (the #3907 reporter cohort) whose first open has a blocked exit could get the wrong country baked in **with no self-heal**.
   *Mitigated in this PR:* first-seen is never persisted from a proxycheck-flagged VPN/proxy/Tor IP — such sessions use the resolved country live and retry persisting later, so the worst case is "behaves like today until a non-VPN open", not "bricked until data wipe". Residual risk is limited to anonymizers proxycheck misses. *Further options if needed:* seed first-seen from the Apollo-cached `countryCode @client` of previous sessions; support runbook for the data-reset escape hatch.
2. **Compliance exposure.** The #3847 anti-bypass is removed by design. Note the VPN-aware persistence also narrows the bypass: a country resolved through a *detected* VPN is never pinned, so a durable bypass requires an anonymizer proxycheck misses — the same exposure the old live-IP model already had.
   *Mitigated in this PR:* (a) **compliance telemetry** — analytics events for every registration-country pin (`registration_country_persisted`, with source), every VPN-deferred pin (`registration_country_persist_deferred`), and every self-heal (`region_restriction_cleared`, with restriction/account type/country), so the field behavior is observable instead of assumed; (b) **`restrictionSelfHealEnabled` remote kill-switch** (default on, mirroring `dollarRestrictionCacheEnabled`) — ops can freeze the new clearing behavior in minutes without a release, degrading to today's set-only semantics while the rest of the fix keeps working.
   *Still required:* explicit compliance sign-off before merge; long-term, a server-side account registration-country field (also fixes email-only custodial); blocked lists remain remote-config so ops can react without a release.
3. **Device-global registration state.** `firstSeenCountryCode` is per-device, not per-account (matching the existing flags' granularity): wallets on one device share it, and a custodial login's phone country overwrites it for all. Multi-account devices spanning countries are rare.
   *Mitigations:* accept for now; if support cases appear, migrate to the existing `…ByAccountId` persistent-state pattern as a follow-up.
4. **Creation vs. usage divergence.** Signup/creation gates intentionally stay on live location while usage gates use registration country — a traveler can be unable to create an account where their existing wallet works. Confusion, not loss.
   *Mitigations:* support/FAQ note; revisit unifying creation gates on registration country only if product wants it (compliance-conservative as is).
5. **Operational semantics shifted.** Settings footer now shows registration country (not current location), and clearing ignores the `dollarRestrictionCacheEnabled` kill-switch, so flags may have healed during a kill-switch period.
   *Mitigations:* brief support/ops changelog note; both changes are intended behavior — this is a communication risk only.
6. **Version skew.** Older clients keep live-IP behavior until updated — certain to occur, bounded impact (it's the pre-existing bug, not a new one).
   *Mitigations:* let organic updates absorb most of the tail, then bump `mobileVersions.minSupported` (AppUpdateModal). With risk 1 mitigated in-PR, a forced fleet-wide upgrade concentrating everyone's first-seen moment is acceptable.

Mitigated in this PR: clearing on remote-config **defaults** after a failed fetch is prevented by the new `remoteConfigLoaded` gate (probability was low but severity high — a config outage could have unlatched legitimate restrictions).

## Tests

- New: `use-registration-country.spec.ts` (tier priority, backfill, persist-once, VPN-deferred persist, unmount, fail-open), `first-seen-country.spec.ts`, `detectAnonymizingIp*` cases in `ip-country-lookup.spec.ts` (v3/v2 shapes, inconclusive handling, session cache).
- Updated: gate specs (persist/clear/no-clear-on-untrusted/no-clear-while-config-unloaded/no-clear-when-self-heal-disabled/kill-switch semantics/account-switch heal/cleared-telemetry), device-location spec, 4 persistent-state specs, 3 consumer specs re-mocked.
- **100% stmt/branch/func/line coverage on every file containing PR logic**; `tsc` + `eslint` clean; all affected suites green. Full suite → CI.

**Pre-merge checklist:**
- [ ] Verify proxycheck v3 detection-field shape against a live response (parser handles v3 `detections` and v2 flat `proxy` defensively; a mismatch degrades to "no evidence" → persist, never a crash or wrong block) — one manual device test with a VPN on covers it.
- [ ] Compliance sign-off on the registration-based model (see Risks #2).

## Appendix A: why the #3847 anti-bypass is removed *by design*

The anti-bypass was the rule "even when the registration country is allowed, a second lookup of the current IP can override it and restrict the account." Keeping it was considered and rejected, for four stacking reasons:

1. **It is logically incompatible with the model this PR adopts.** The decided behavior (the issue's Expected Behavior) is that restrictions are a property of the *account*, not of today's network path — the exact negation of the anti-bypass. Any rule of the form "current IP can restrict an allowed account" re-admits the entire class of VPN false positives this PR exists to eliminate. Removing it isn't a side effect of the refactor; it *is* the refactor.

2. **It inverted the signal hierarchy.** Phone country is a strong, stable, registration-tied signal; IP geolocation is weak and noisy (VPNs, roaming, corporate egress, carrier NAT). The anti-bypass let the weak signal override the strong one — and, combined with the set-only latch, a single noisy reading overrode a clean registration *permanently*, ultimately feeding users into the forced-conversion flow (#3910). The strongest consequence was attached to the weakest evidence.

3. **It failed its own cost-benefit test.** The person it targeted (living in a blocked region, foreign phone, deliberately evading) defeats it trivially — connect through a VPN with an allowed exit; the evasion tool *is* a VPN. The person it actually caught was the honest user in an allowed country whose VPN exit happened to land in a blocked list — someone with nothing to hide who therefore takes no care what their IP looks like. #3907 is the field evidence: real harm to legitimate users, enforcement only against evaders too careless to use the evasion tool.

4. **Its legitimate function was replaced, not just deleted.** The genuine concern behind #3847 — no laundering past the restriction — is addressed without punishing honest users: (a) no *durable* allow-bypass through a VPN, since a country resolved via a detected anonymizer is never pinned as registration (a permanent bypass requires keeping the VPN on forever, the same exposure the old live-IP model had); (b) telemetry makes bypass-shaped events measurable so re-tightening can be decided on data; (c) the `restrictionSelfHealEnabled` kill-switch freezes the new leniency instantly if the data says it's abused; (d) the correct long-term enforcement point is a server-side registration-country field the client can't spoof (follow-up).

A middle-ground variant exists — latch only on **clean, non-VPN-flagged** IPs, which would catch a blocked-region resident on home wifi without ever touching VPN users. It is deliberately *not* included: re-adding any IP-based restriction contradicts the decided direction and is a compliance/product call, not something to fold into a bugfix. That is what "removed by design" means here — explicit, argued, and reversible by the people with the authority to reverse it (hence the compliance sign-off in the checklist).

## Appendix B: user stories — account type × lifecycle

Where the registration country comes from, per account type and lifecycle phase:

| Account type | Registration (signup / wallet creation) | Post-registration (normal use) | Recovery (new device / reinstall) |
|---|---|---|---|
| **Custodial, phone** | Phone country, from the number itself | Phone country, re-parsed from `me.phone` | Phone country, re-fetched from backend — always correct |
| **Custodial, email-only** | First-seen country pinned around first use | Pinned first-seen country | First-seen **re-derived** (no backend country field) |
| **Self-custodial** | First-seen country pinned at creation | Pinned first-seen country | First-seen **re-derived** at seed import |

VPN rule in every cell: a *detected* VPN/proxy never pins a country (session uses it live, pin retries later); an *undetected* one can (residual risk). Format below: **Given** → **when** → **then**, with *Before:* where behavior changed.

### Custodial, phone number

**Registration**
- **C1 — Signup, any network.** Given a new user signing up with an SV phone, when the account is created, then the registration country is SV from the number itself — VPN state during signup is irrelevant to the Dollar/transfer gates. (The *creation* gates — OFAC/first-signup lists — still check live location at signup; unchanged by this PR, see S-X2.)

**Post-registration**
- **C2 — Allowed country, VPN on.** Given an SV-phone user, when they use the app through an EU/HK-exit VPN, then nothing changes: the gate reads the phone country only. *Before:* the IP double-check could latch `stablesatsRestrictedCustodial` permanently.
- **C3 — Previously latched by a VPN session.** Given the stale flag on disk, when they next open the home screen, then it self-heals (`region_restriction_cleared`) and Dollar/transfer return — the #3910 forced-conversion prompt stops. *Before:* restricted forever short of a data wipe.
- **C4 — Genuinely blocked registration (HK phone).** Given an HK-phone user, when they open the app — VPN or not, anywhere — then the restriction applies and persists. A VPN does not clear it: the evidence is the phone country, and it says HK. Unchanged by this PR.

**Recovery**
- **C5 — New device / reinstall / re-login.** Given any phone user on a fresh install, when they log in, then `me.phone` is re-fetched from the backend and the registration country is identical to before — no IP involved, VPN state at login irrelevant. **Recovery is lossless for phone accounts.**

### Custodial, email-only (no phone)

**Registration**
- **E1 — Email signup.** Given a new email-only account, when the app first resolves a country (no phone signal exists), then the first-seen country is pinned — on a clean IP the real country, behind a detected VPN nothing is pinned yet (retries), behind an undetected one possibly the wrong country (residual). Custodial blocked lists and flags apply throughout.

**Post-registration**
- **E2 — Normal use.** Given a pinned first-seen country, when the user toggles a VPN, then the gates don't move — the #3907 fix applies to email accounts too. *Before:* the gate followed the live IP every session.

**Recovery**
- **E3 — New device / reinstall.** Given an email-only user logging in on a fresh install, when the gates need a country, then first-seen is **re-derived at that moment** — the backend has no registration-country field for email accounts. Clean IP → correct; detected VPN → deferred; undetected → wrong bake-in until data reset. This is the cell a server-side registration country would fix (noted follow-up).

### Self-custodial (no phone, no backend account)

**Registration (wallet creation)**
- **N1 — Creation at home.** Given a new wallet created on a clean SV IP, when first-seen resolves, then SV is pinned — later VPN use never moves the gates (the core #3907 fix).
- **N2 — Creation behind a VPN (incl. the upgrade moment).** Given the first open happens with a VPN active — including existing installs whose *first open after updating* is their first-seen moment — when proxycheck flags the anonymizer, then **nothing is pinned**: the session uses the resolved country live (behaves like the old model), and the first non-VPN open pins the real country. *Without this mitigation:* the VPN exit country would be baked in permanently.
- **N3 — Creation in a blocked region.** Given a new wallet created in HK on a clean IP, when first-seen resolves, then HK is pinned and the restriction applies from day one.
- **N4 — Deliberate bypass at creation.** Given someone in HK creating a wallet behind an SV-exit VPN, when first-seen runs, then the detected VPN defers pinning — unrestricted only while the VPN stays on (the old live-IP model's exact exposure), and the first clean open pins HK. An undetected anonymizer makes the bypass stick (residual; observable via `registration_country_persist_deferred` / `registration_country_persisted` telemetry).

**Post-registration**
- **N5 — Allowed pin, any VPN (the #3907 reporter).** Given SV pinned, when any VPN is used, then stable balance and transfer stay available. *Before:* VPN on → "not available in your region", VPN off → works.
- **N6 — Blocked pin, VPN as escape attempt.** Given HK pinned, when the user turns on an SV-exit VPN, then the restriction stays — the pin is the evidence, not the live IP. **No durable bypass post-registration.**
- **N7 — The traveler.** Given SV pinned, when the user travels to a blocked country, then the gates don't move (registration model, deliberate). Creating a *new* wallet there may still be blocked — creation gates stay on live location (accepted divergence, Risks #4).

**Recovery (seed import)**
- **N8 — Recovery at home.** Given a seed import on a fresh install (persistent state is empty — the pin does *not* travel with the seed), when first-seen re-resolves on a clean IP, then the correct country is re-pinned. Recovery is lossy but self-correcting for honest users.
- **N9 — Recovery behind a VPN.** Given the same import behind a detected VPN, then pinning defers (as N2) until a clean open. Behind an *undetected* anonymizer the wrong country is baked until data reset (residual risk — the sharpest remaining edge, Risks #1).

### Cross-cutting

- **X1 — Account-type switch.** Given a custodial phone user switching to self-custodial, when the switch happens, then the phone country was already backfilled into `firstSeenCountryCode`, so the self-custodial session inherits the same registration country; in a blocked country both flags are written (anti-bypass double-write preserved), in an allowed country both heal.
- **X2 — Creation vs. usage gates.** Given any user anywhere, account/wallet *creation* is still gated on live location (`use-creation-block`, `use-custodial-eligibility` — compliance-conservative, unchanged), while Dollar/transfer *usage* follows registration. The two can disagree by design.

### Failure modes and operations

- **F1 — Offline / lookup failure at first open.** Given no phone, no pin, and all geolocation adapters failing, then the country is unknown (`trusted: false`) — **fail open**, nothing persisted, retry next session.
- **F2 — Remote config outage.** Given a failed Firebase fetch (default lists, `remoteConfigLoaded` false), then clearing is blocked — a country missing from the *defaults* must not clear a flag set from the *real* config. Setting still works (over-restriction self-heals later; the safe direction).
- **F3 — Compliance pulls the self-heal brake.** Given ops flips `restrictionSelfHealEnabled` off, then no flags clear (and no cleared-telemetry fires) — set-only semantics restored in minutes without a release; registration-based gating keeps working.
- **F4 — The existing cache kill-switch.** Given `dollarRestrictionCacheEnabled` off, then persisted flags are ignored on read and none are set — but healing still runs, so stale flags are cleaned instead of frozen for resurrection.
- **F5 — Old app version.** Given a pre-fix release, then live-IP behavior persists until the user updates (Risks #6) — bounded: it's the pre-existing bug, not a new one.

### DIFF
--- a/__tests__/components/version/version.spec.tsx
+++ b/__tests__/components/version/version.spec.tsx
@@ -4,11 +4,10 @@ import { render } from "@testing-library/react-native"
 
 import { VersionComponent } from "@app/components/version"
 
-const mockUseDeviceLocation = jest.fn()
+const mockUseRegistrationCountry = jest.fn()
 
-jest.mock("@app/hooks/use-device-location", () => ({
-  __esModule: true,
-  default: () => mockUseDeviceLocation(),
+jest.mock("@app/hooks/use-registration-country", () => ({
+  useRegistrationCountry: () => mockUseRegistrationCountry(),
 }))
 
 jest.mock("@app/i18n/i18n-react", () => ({
@@ -43,11 +42,11 @@ jest.mock("@rn-vui/themed", () => ({
 
 describe("VersionComponent", () => {
   beforeEach(() => {
-    mockUseDeviceLocation.mockReset()
+    mockUseRegistrationCountry.mockReset()
   })
 
-  it("shows the device-detected country below the version", () => {
-    mockUseDeviceLocation.mockReturnValue({ countryCode: "HK" })
+  it("shows the registration country below the version", () => {
+    mockUseRegistrationCountry.mockReturnValue({ countryCode: "HK" })
 
     const { getByText } = render(<VersionComponent />)
 
@@ -55,7 +54,7 @@ describe("VersionComponent", () => {
   })
 
   it("falls back to the unknown label when the country is not detected", () => {
-    mockUseDeviceLocation.mockReturnValue({ countryCode: undefined })
+    mockUseRegistrationCountry.mockReturnValue({ countryCode: undefined })
 
     const { getByText } = render(<VersionComponent />)
 
--- a/__tests__/hooks/use-device-location.spec.ts
+++ b/__tests__/hooks/use-device-location.spec.ts
@@ -1,9 +1,6 @@
 import { renderHook, act } from "@testing-library/react-hooks"
 
-import useDeviceLocation, {
-  isBlockedCountry,
-  useIpCountryCode,
-} from "@app/hooks/use-device-location"
+import useDeviceLocation, { isBlockedCountry } from "@app/hooks/use-device-location"
 
 const mockLogError = jest.fn()
 const mockUpdateCountryCode = jest.fn()
@@ -273,41 +270,6 @@ describe("useDeviceLocation", () => {
     await act(async () => {})
 
     expect(mockUpdateCountryCode).toHaveBeenCalledWith(expect.anything(), "DE")
-  })
-})
-
-describe("useIpCountryCode", () => {
-  beforeEach(() => {
-    jest.clearAllMocks()
-    mockResolveIpCountryCode.mockResolvedValue(undefined)
-  })
-
-  it("does not call IP lookup while disabled", () => {
-    const { result } = renderHook(() => useIpCountryCode(false))
-
-    expect(mockResolveIpCountryCode).not.toHaveBeenCalled()
-    expect(result.current).toBeUndefined()
-  })
-
-  it("resolves the country from the adapter chain when enabled", async () => {
-    mockResolveIpCountryCode.mockResolvedValue("HK")
-
-    const { result } = renderHook(() => useIpCountryCode(true))
-
-    await act(async () => {})
-
-    expect(mockResolveIpCountryCode).toHaveBeenCalled()
-    expect(result.current).toBe("HK")
-  })
-
-  it("stays undefined when all adapters return nothing", async () => {
-    mockResolveIpCountryCode.mockResolvedValue(undefined)
-
-    const { result } = renderHook(() => useIpCountryCode(true))
-
-    await act(async () => {})
-
-    expect(result.current).toBeUndefined()
   })
 })
 
--- a/__tests__/hooks/use-dollar-balance-restricted.spec.ts
+++ b/__tests__/hooks/use-dollar-balance-restricted.spec.ts
@@ -1,27 +1,31 @@
 import { renderHook } from "@testing-library/react-native"
 
-import { getStableTokenRestricted } from "@app/store/persistent-state/stable-token-restriction"
-import { getStablesatsRestricted } from "@app/store/persistent-state/stablesats-restriction"
+import {
+  getStableTokenRestricted,
+  withStableTokenRestricted,
+} from "@app/store/persistent-state/stable-token-restriction"
+import {
+  getStablesatsRestricted,
+  withStablesatsRestricted,
+} from "@app/store/persistent-state/stablesats-restriction"
 import { PersistentState } from "@app/store/persistent-state/state-migrations"
 import { AccountType } from "@app/types/wallet"
 
-const mockUseDeviceLocation = jest.fn()
+const mockUseRegistrationCountry = jest.fn()
 const mockUseRemoteConfig = jest.fn()
+const mockUseFeatureFlags = jest.fn()
 const mockUseActiveWallet = jest.fn()
 const mockUpdateState = jest.fn()
-const mockUseIpCountryCode = jest.fn()
 
 let mockPersistentState: PersistentState
 
-jest.mock("@app/hooks/use-device-location", () => ({
-  __esModule: true,
-  ...jest.requireActual("@app/hooks/use-device-location"),
-  default: () => mockUseDeviceLocation(),
-  useIpCountryCode: (enabled: boolean) => mockUseIpCountryCode(enabled),
+jest.mock("@app/hooks/use-registration-country", () => ({
+  useRegistrationCountry: () => mockUseRegistrationCountry(),
 }))
 
 jest.mock("@app/config/feature-flags-context", () => ({
   useRemoteConfig: () => mockUseRemoteConfig(),
+  useFeatureFlags: () => mockUseFeatureFlags(),
 }))
 
 jest.mock("@app/hooks/use-active-wallet", () => ({
@@ -33,6 +37,12 @@ jest.mock("@app/store/persistent-state", () => ({
     persistentState: mockPersistentState,
     updateState: mockUpdateState,
   }),
+}))
+
+const mockLogRegionRestrictionCleared = jest.fn()
+jest.mock("@app/utils/analytics", () => ({
+  logRegionRestrictionCleared: (...args: unknown[]) =>
+    mockLogRegionRestrictionCleared(...args),
 }))
 
 import {
@@ -50,14 +60,25 @@ const remoteConfig = {
   custodialDollarBalanceBlockedCountries: ["HK"],
   selfCustodialDollarBalanceBlockedCountries: ["FR"],
   dollarRestrictionCacheEnabled: true,
+  restrictionSelfHealEnabled: true,
 }
+
+const country = (countryCode: string | undefined) => ({
+  countryCode,
+  loading: false,
+  trusted: Boolean(countryCode),
+})
 
 const setup = (accountType: AccountType): void => {
   jest.clearAllMocks()
-  mockUseDeviceLocation.mockReturnValue({ countryCode: undefined, source: undefined })
+  mockUseRegistrationCountry.mockReturnValue({
+    countryCode: undefined,
+    loading: true,
+    trusted: false,
+  })
   mockUseRemoteConfig.mockReturnValue(remoteConfig)
+  mockUseFeatureFlags.mockReturnValue({ remoteConfigLoaded: true })
   mockUseActiveWallet.mockReturnValue({ accountType })
-  mockUseIpCountryCode.mockReturnValue(undefined)
   mockPersistentState = baseState
 }
 
@@ -68,29 +89,29 @@ describe("useDollarBalanceRestricted", () => {
     beforeEach(() => setup(AccountType.Custodial))
 
     it("is restricted in a Stablesats-blocked country", () => {
-      mockUseDeviceLocation.mockReturnValue({ countryCode: "HK" })
+      mockUseRegistrationCountry.mockReturnValue(country("HK"))
       expect(read()).toBe(true)
     })
 
-    it("is case-insensitive on the device country", () => {
-      mockUseDeviceLocation.mockReturnValue({ countryCode: "hk" })
+    it("is case-insensitive on the registration country", () => {
+      mockUseRegistrationCountry.mockReturnValue(country("hk"))
       expect(read()).toBe(true)
     })
 
     it("is not restricted in a country that only the stable-token list blocks", () => {
-      mockUseDeviceLocation.mockReturnValue({ countryCode: "FR" })
+      mockUseRegistrationCountry.mockReturnValue(country("FR"))
       expect(read()).toBe(false)
     })
 
     it("stays restricted from the persisted custodial flag without a country", () => {
       mockPersistentState = { ...baseState, stablesatsRestrictedCustodial: true }
-      mockUseDeviceLocation.mockReturnValue({ countryCode: undefined })
+      mockUseRegistrationCountry.mockReturnValue(country(undefined))
       expect(read()).toBe(true)
     })
 
     it("ignores the self-custodial persisted flag", () => {
       mockPersistentState = { ...baseState, stableTokenRestricted: true }
-      mockUseDeviceLocation.mockReturnValue({ countryCode: "FR" })
+      mockUseRegistrationCountry.mockReturnValue(country("FR"))
       expect(read()).toBe(false)
     })
   })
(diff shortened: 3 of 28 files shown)


ihaveadifferentname:
@grimen @designsats 

After much research I have concluded that we need to approach this subject from a different angle, not just patchwork but a 360 degrees holistic approach. I've created a decision framework on how to handle our determination of user region based on three modes "Custodial Mode", "Enhanced Mode" and "Anon Mode". 

I understand that this is an epic. However I believe it would be best to implement all at once so that we have a cohesive and marketable approach to the problems of user region and feature flags.

Attached is the decision matrix:
[region_determination_three_mode_journey.html](https://github.com/user-attachments/files/30129212/region_determination_three_mode_journey.html)

And two designs:
Screenshot 2026-07-17 at 21 00 55
Screenshot 2026-07-17 at 21 01 03

grimen:
Superseded by #4046, #4144, and #4145, which fixed the Dollar-balance/transfer VPN restriction more thoroughly and merged 2026-08-21.

PR #3917, opened 2026-07-15
feat(migration): integrate the custodial wind-down backend


esaugomez31:
Runs the custodial migration against the real backend. Every mock and simulation in the flow is gone: the wind-down phase, the balance preview, the point of no return, the funds transfer and the lightning-address re-point are all served by the API now.

## What the flow does now

**Wind-down phase** comes from `Query.windDown`, so the bulletin, the migrate-now prompt, the receive block and the armed gate all follow the server instead of computing anything from dates.

**Balance preview** comes from `Query.migration.preview`. The server owns the fee, the de-minimis subsidy and the resulting amount; the client renders the four figures verbatim and never does the arithmetic.

**The point of no return is server-side.** Landing on the commit screen with figures declares the migration started, and `migration.status` is what decides whether the user may leave. It survives a reinstall and a new device, which the local checkpoint could not: that checkpoint is demoted to remembering which screen to resume on, and never decides whether the user can exit. A locked account is taken straight back to where it left off rather than through the intro again.

**The lightning address moves too.** On the commit screen the custodial lightning address(es) are re-pointed onto the new self-custodial wallet with the same signed proof of possession, once per visit. Approve waits on it exactly as it waits on the started migration and the ready figures: a lost connection offers the shared retry, a settled rejection hands over to support, and an address already moved or never registered counts as done. The mutation is idempotent, so a retry after a dropped network never double-registers.

**The transfer is real.** The commit screen collects a no-amount invoice, the wallet's public key and a proof of possession from the provisioned self-custodial wallet in a single connection, commits them, and watches the phase until it settles. A migration the server completed but the device never swapped away from is finished on the next launch.

**Failures are told apart.** A lost connection offers a retry and support never hears about it; an answer the server actually gave is final and hands over to support. Every handover names a stable, untranslated reason code that travels in the support email body, so a ticket says what failed instead of arriving blind.

## Behaviour worth knowing when reviewing

- **A dollar balance blocks every phase, the armed gate included.** The backend rejects `migrationStart` and `migrationCommit` while the USD wallet holds anything and never converts it, so letting a gated account through would only move the refusal to a screen with no way back.
- **A failed transfer offers no retry, by design.** The backend's phase machine has no client-reachable path out of `FAILED`, so a retry button would loop on a refusal. Support owns that case.
- **A clock-skew rejection is recoverable, not a handover.** The backend answers a stale proof and a bad destination with the same code, so the client treats it as recoverable only when it has also seen the device clock running more than ten minutes off the server's, read from each response's `Date` header. That case points the user at their clock and offers a retry; a genuinely bad destination still hands over to support.
- **A rejected start never locks.** The status stays `NOT_STARTED`, which keeps a user the server refuses from being sealed into a flow they cannot finish.
- **A status the app could not read leaves the user free.** Locking every offline launch into a migration is a worse failure than letting a locked user browse until the next successful read.
- Migration queries are `no-cache`: they are argument-less top-level fields whose nodes carry no id, so a cached read would serve one account's state to another after a switch.

## Verification

Every touched file is at 100% statement, branch, function and line coverage.

Verified on device with an account in the wind-down cohort: the migration starts, and the lock survives killing the app, landing straight back on the commit screen.

**The transfer itself has not been verified end to end, and cannot be today.** The staging backend runs on signet while the Breez Spark SDK exposes only mainnet and regtest, so the wallet issues an invoice on a chain the backend's node cannot pay. Everything up to that boundary was confirmed on device: the SDK connects to the provisioned wallet, signs and produces the invoice without error. Closing that gap needs the two environments on the same network.


esaugomez31:
* **#3934** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3917** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3917?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3917?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3863** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3863?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

grimen (changes requested):
## Findings

### High

**1. `migrationCommit` is auto-resent by the global retry link (blocking).**
The `RetryLink` in `app/graphql/client.tsx:184` resends any operation that fails with a non-401 network error unless it is listed in `noRetryOperations` (`app/graphql/client.tsx:48`) — none of the new migration mutations are.
*Example:* a commit lands but its response is lost; the link silently resends it; the backend refuses the second invoice as a state conflict, and `use-migration-transfer.ts:218` hands the user to support while the real transfer completes underneath.
*Proposal:* add `migrationCommit` to `noRetryOperations` (the list's trailing comment invites this); `migrationStart` and `migrationLnAddressTransfer` are idempotent, so including them too is a consistency call.

**2. A settled commit failure does not survive a remount; a re-entered transfer screen spins forever (blocking).**
The commit guard `accountsWithCommitStarted` (`use-migration-transfer.ts:57`) is module-level, but the failure outcome (`failureReason`) is per-mount state.
*Example:* the commit fails, the user goes support, Back, Approve; the fresh transfer screen has no failure to show, the guard blocks a re-commit (`use-migration-transfer.ts:208`), hardware back is swallowed, and the status poll never advances — no exit.
*Proposal:* keep the outcome at module level next to the guard (account id to outcome), so a remount can render the failure the guard already remembers.

**3. The resume swap races registry hydration and can hand a healthy user to support (blocking).**
`use-resume-completed-migration.ts:42` waits only for checkpoint loading, but `useCompleteMigration` checks the destination against the registry's `accounts`, which start empty on a resume launch and fill after an async keystore read.
*Example:* a fast network answers `COMPLETED` before a slow iOS keychain read finishes; `completeMigration()` returns false and the user is terminally routed to support with `self-custodial-account-not-on-device` for a wallet that is on the device.
*Proposal:* include the registry's `loading` (`use-account-registry.tsx:158`) in the `isSwapPending` gate.

**4. The ln-address retry can open two Breez SDK connections on the same storage directory (blocking).**
`retry()` (`use-migration-ln-address-transfer.ts:93`) refuses only settled outcomes, not an attempt still in flight.
*Example:* `migrationStart` fails fast on a blip, so the shared Retry appears while the ln proof's SDK connect is still pending; tapping it re-fires the effect and a second `initSdk` runs against the same `storageDirFor(accountId, network)` — the race `withMigrationSdk`'s own contract warns against.
*Proposal:* make `retry()` a no-op while an attempt is unsettled (in-flight ref), or serialize `withMigrationSdk` per account id, which also covers the remount path.

**5. A transient SDK-connect failure is classified as a settled rejection — an unrecoverable dead end (blocking).**
Network-versus-settled is only distinguished at the Apollo layer; any non-Ok proof result becomes `Rejected` (`use-migration-ln-address-transfer.ts:114`), which `retry()` permanently refuses. The funds commit has the same shape (`use-migration-transfer.ts:152`), softened only by relaunch-resume.
*Example:* the network drops during the Breez connect; the user is escalated to a support screen they are re-routed back to on every focus, for a condition an app restart fixes.
*Proposal:* classify connect-level network errors in the SDK layer and route them to `ConnectionIssue` and the shared retry.

### Medium

**6. The support screen's Back hardcodes the commit screen, which is wrong for the resume handover.**
`contact-support-screen.tsx:62` always navigates to `accountMigrationBalancesOverview`, but the resume hook (`use-resume-completed-migration.ts:67`) hands over from the root navigator.
*Example:* Back pushes a fresh commit screen onto a completed migration; `migrationStart` fires against `COMPLETED` and the user is re-handed to support with a wrong reason, overwriting the right one.
*Proposal:* carry the origin in the route params, or fall back to `goBack()` when the screen was not reached from the commit screen.

**7. The blocker-mode dollar modal is a dead end for users who cannot transfer in-app.**
In blocker mode `exitFlow` is a `goBack()` from the root route (a no-op) and `onTransfer` is undefined for transfer-blocked or dollar-restricted users, while `isVisible={isFocused}` keeps the modal up (`migration-gate.tsx:233`).
*Example:* a user in a transfer-blocked region taps the modal's only button, nothing happens, and the modal covers the gate's support link forever.
*Proposal:* in blocker mode, offer the support path from the modal, or dismiss to the gate screen instead of trying to exit.

**8. An errored lock query silently reads as "unlocked".**
`useMigrationStatus` (`use-migration-status.ts:45`) swallows `error`, so a failed read yields `{status: null, loading: false}`; the gate's readiness guard never notices and `retryGateData` cannot refetch it.
*Example:* a locked user on a flaky connection is re-pitched the intro instead of being resumed — the case the loading guard exists to prevent, arriving via the error path.
*Proposal:* expose `error` and `refetch` from `useMigrationStatus` and fold the lock into `hasGateDataError`.

**9. The commit screen lost its fresh-balance fetch; a comment promises an override that no longer happens.**
`use-migration-balances-preview.ts:49` now calls `useCustodialWalletBalances()` cache-first, while `use-custodial-wallet-balances.ts:20` still documents a commit-screen freshness override (the old hook used `cache-and-network`).
*Example:* a dollar deposit arriving after the last cache write stays invisible until `migrationStart` refuses it — a support handover instead of the empty-your-dollars modal.
*Proposal:* restore the fetch-policy override, or update the comment and drop the unused option if cache-first is intended.

**10. Retry after a lost response can commit a second invoice and route a succeeding transfer to support.**
`retry()` deletes the commit guard immediately (`use-migration-transfer.ts:229`), but the poll needs up to two seconds to observe `TRANSFERRING`.
*Example:* the first commit landed; the user retries inside that window; the duplicate is refused as a state conflict and treated as settled, sending the user to support while the real transfer completes.
*Proposal:* refetch the status once before recommitting, or treat a state-conflict rejection right after a connection issue as "keep watching".

**11. An absent payload counts as a successful start.**
`use-ensure-migration-started.ts:78` derives the rejection from `data?.migrationStart`, so `{ data: undefined }` without a throw sets `isStarted` — against the codebase's own "a null answer is not an answer" rule (`use-migration-status.ts:30`).
*Example:* a transport quirk returns no payload; the commit screen believes the lock is armed on the strength of no answer, and the backend's re-validation surfaces later as a confusing refusal.
*Proposal:* treat missing `data` as a connection issue and hand it to the shared retry.

**12. Two independent blocker instances can disagree, defeating the deeplink reset.**
`navigation-container-wrapper.tsx:99` and `root-navigator.tsx:1064` each mount `useMigrationBlocker` over `no-cache` queries, so each gets independent answers.
*Example:* the navigator's read answers `IN_PROGRESS` and shows the gate while the wrapper's read fails, so `resetToBlocker` never fires and deeplinked screens keep operating over the locked account.
*Proposal:* fetch the lock and wind-down once in a provider and share the answer; this also removes the duplicate reads on every custodial launch.

## Test Coverage Gaps

- Skew recovery is unpinned: no spec observes a large skew, then a corrected clock, and asserts `isDeviceClockSkewed()` returns false.
- No spec re-enters the transfer screen after a settled failure (finding 2) or taps retry mid-flight (finding 4).
- No spec exercises a network-shaped SDK-connect failure in either hook (finding 5).
- `canClose = shouldShowClose && !isExitBlocked` (`migration-required-screen.tsx:135`) is untested at screen level; deleting the clause passes the suite.
- The `expirySecs` passthrough in `bridge/receive.ts` is asserted nowhere; re-hardcoding `undefined` (the stranded-transfer hazard the comments warn about) would be invisible.
- The resume spec mocks `useCompleteMigration` wholesale (so it cannot catch finding 3) and skips the offline case (`status: null`, expecting neither swap nor handover).
- `use-migration-blocker.spec.ts:20` hardwires the lock's `loading: false`, so the documented "ignores the lock's loading" behavior is never asserted.

## Code Smells

- The proof challenge binds neither invoice nor purpose tag: one signature is valid for both migration mutations within the ten-minute window, and the payout destination is safe only if the backend checks the invoice's key against `sparkPubkey`. For the backend to confirm; a comment in `migration-proof.ts` would help if intended.
- `NoMnemonic` maps to `SelfCustodialAccountMissing` on the commit path but to the generic `LnAddressTransferFailed` on the ln path, and a per-identifier `FAILED` result is never reported — both muddy the triage key.
- `is-network-failure.ts:6` claims to catch only requests the server never settled, but truthy `networkError` also covers `ServerError` and `ServerParseError`; the docstring oversells.
- The skew threshold matches the backend's window exactly while the measurement includes latency and the `Date` header's one-second resolution; fail-safe direction, but worth a comment.
- `use-migration-lock.spec.ts:98` uses `"selfCustodial"`, which is not a real `AccountType` value; `AccountType.SelfCustodial` keeps the fixture honest.
- `useEnsureMigrationStarted.retry` does not guard `isStarted`, re-firing an accepted start — harmless but asymmetric with the ln hook, which guards `isTransferred`.
- The PR description's "straight back on the commit screen" after a reinstall overstates: the checkpoint dies with the app, so the flow re-provisions and re-walks terms and backup first.

esaugomez31:
Thanks @grimen for the thorough review. Here is how each finding landed, in the same order, plus a section for the findings from a follow-up review pass.

## Findings

### High

**1. `migrationCommit` auto-resent by the global retry link.**
Resolved: `migrationStart`, `migrationCommit` and `migrationLnAddressTransfer` are in `noRetryOperations`, so the transport layer never resends them; recovery is caller-driven through the shared retry.

**2. A settled commit failure does not survive a remount.**
Resolved: the terminal outcome is kept at module level next to the commit guard (`settledCommitFailures`, account id → reason), so a re-entered transfer screen restores the failure instead of spinning.

**3. The resume swap races registry hydration.**
Resolved: `useCompleteMigration` folds the registry's `loading` into `migrationLoading`, and `useResumeCompletedMigration` gates `isSwapPending` on it.

**4. The ln-address retry can open two Breez SDK connections on the same storage directory.**
Resolved: `withMigrationSdk` serializes runs per storage directory, and the hook claims the attempt before the request (`firedAttemptRef`).

**5. A transient SDK-connect failure is classified as a settled rejection.**
Resolved: `toSdkFailure` maps a network-tagged `SdkError` to `ConnectionError` (→ shared retry) rather than `Failed`, on both the commit and ln paths.

### Medium

**6. The support screen's Back hardcodes the commit screen.**
Resolved: the route carries an `origin` (Commit/Resume); Back returns to the commit screen for the mid-migration handover and `goBack()`s for the resume handover.

**7. The blocker-mode dollar modal is a dead end for users who cannot transfer in-app.**
The modal now surfaces an explicit action (labeled Transfer/Close primary plus a close icon). The remaining transfer-restricted-region case, where in-app conversion is not reachable, is resolved on the follow-up refinements branch, which routes the modal's action into the conversion flow (the only entry point for in-migration conversion).

**8. An errored lock query silently reads as "unlocked".**
Resolved: `useMigrationStatus` exposes `error`/`refetch`, `useMigrationLock` surfaces `hasError`, and the gate folds it into `hasGateDataError`.

**9. The commit screen lost its fresh-balance fetch.**
Resolved: the balances preview reads `cache-and-network` again, so a late dollar deposit is visible before `migrationStart` refuses it.

**10. Retry after a lost response can commit a second invoice.**
Resolved: `retry()` refetches the phase first; if it advanced past `IN_PROGRESS` it watches instead of recommitting.

**11. An absent payload counts as a successful start.**
Resolved: a settled start with no payload is treated as a connection issue and handed to the shared retry, never read as armed.

**12. Two independent blocker instances can disagree, defeating the deeplink reset.**
Resolved: `useMigrationBlocker` computes once in a `MigrationBlockerProvider` wrapping the navigation container in `app.tsx`; both consumers read one shared value, and the duplicate `windDown`/`migration.status` reads per launch are gone.

## Follow-up review pass

### Medium

**13. Invoice-mint network failure misclassified as a settled failure.**
The `createReceiveLightning` adapter flattened the `SdkError` to a string, so a dropped connection while minting the invoice became a settled `Failed` and a support dead-end. Resolved: `buildMigrationTransferRequest` calls `sdk.receivePayment` directly (mirroring `getWalletInfo`), so the raw error reaches `classifySdkError` with its network tag intact.

**14. The lightning-address re-point was not gated on a confirmed start.**
It fired in parallel with `migrationStart`; a rejected start would still re-point `@blink.sv` irreversibly. Resolved: gated on `migrationStart.isStarted`.

### Low

**15. A failed migration returned the user to the app on relaunch.**
`FAILED` released the lock. Resolved: `FAILED` now keeps the account locked and routes it to support (funds may still settle server-side, so the account stays in the flow).

**16. The transfer retry was not re-entrancy-guarded.**
A double-tap could clear the commit guard mid-flight. Resolved: an in-flight ref serializes retries.

**17. Cache-first reads of the argument-less `me` field.**
`useMigrationOwnerQuery` (owner id, which the proof binds to) and `useMigrationSupportDetailsQuery` read cache-first and could serve the previous account after a switch. Resolved: both set `no-cache`.

Also folded in: the point-of-no-return comment corrected, a defensive optional chain on the commit payload, and an em-dash cleanup.

grimen (approved):

PR #3919, opened 2026-07-16
ci: update user references and labels from galoybot to blinkbitcoinbot

openoms:
needs ci repipe after merge

match with deployments


grimen (approved):

openoms:
repiped ok

PR #3921, opened 2026-07-17
fix(auth): re-apply the app lock when resuming past the grace period

esaugomez31:
## Summary

The PIN/biometric lock only ever guarded the cold start. Once the app was running, a backgrounded session stayed open indefinitely: anyone picking up the phone could reopen Blink straight into the home screen without being challenged. The lock is now re-applied when the app returns to the foreground after a grace period, and the raised lock resists being dismissed back out of.

## Problem

`AppStateWrapper` handled the background/foreground transitions (analytics, home-query refetch) but never touched the lock. The lock was raised exactly once, by `authenticationCheck` being the initial route at cold start, and no code path raised it again for the life of the process. The setting the user turned on in Settings > Security therefore protected a killed app, not a backgrounded one.

## Fix

`AppStateWrapper` records when the app leaves the foreground and, on return, re-raises the lock if the trip outlasted the grace period and the user has a PIN or biometrics configured. Both halves are needed:

- `setAppLocked()` flips the flag that defers incoming deep links until the unlock.
- `navigate("authenticationCheck", { isResume: true })` puts the lock in front of the user.

The timestamp lives in a ref rather than state: nothing renders from it, and losing it to a killed process is harmless, since that path already locks through the cold start.

## Grace period (remote config)

`appLockGracePeriodSeconds`, default `60`. Without it, the common hop to another app and straight back (copying an address, checking a message) would demand the PIN every time.

Following the existing flags convention, the default is registered through `setDefaults`, so it doubles as the fail-safe when the key is absent. **The Firebase key does not need to exist for this to ship**; it only needs to exist to tune the value remotely.

RN exposes no monotonic clock, so the trip is measured against the wall clock. Winding it forward only locks sooner; a trip that reports a negative duration is not trusted and locks rather than being let through.

## Resume vs cold start

The three unlock screens share a `useUnlockScreen` hook that holds the resume contract in one place. `isResume` threads from `authenticationCheck` into `authentication`/`pin`, and changes only what follows a successful unlock:

- **Cold start**: `updateDeviceSessionCount` plus the reset to `Primary`.
- **Resume**: `goBack()`, returning the user to the screen they were on, and skips `updateDeviceSessionCount`, since only a cold start opens a device session.

The biometric screen's "Use PIN" fallback forwards `isResume`, so a resume that drops to the PIN steps back the same way.

## Keeping the resume lock from being dismissed

A cold-start lock is the initial route with nothing behind it. A resume lock instead sits on top of the live stack, so back navigation would pop it and reveal the app unchallenged. Blocking that takes one guard per platform, and neither is redundant:

- **iOS**: `gestureEnabled: false` on the unlock screens blocks the edge swipe. It is off only on resume, since a fixed value would trap the header-less PIN-creation flow Settings opens, where the swipe is the only way back.
- **Android**: `useUnlockScreen` swallows the hardware back press via `BackHandler`, gated on `isResume`. This leaves the unlock's own programmatic `goBack` untouched.

## The `isAppLocked` guard

The relock bails when the lock is already up, so a cold-start unlock screen the user has not answered yet never gets a second one stacked over it (which would otherwise demand the PIN twice: once for the stacked screen, then again on the one behind it).

## AppState transitions

The transition checks are explicit: `=== "active" || === "inactive"` on the way out, `=== "background" && === "active"` on the way in. On iOS the trip to background passes through `"inactive"` (`RCTAppState.mm` maps `willResignActive` there), so both are matched by name rather than a loose `/active/` regex. The `AppState` values and grace period read through refs, so the listener does not resubscribe on a lock-state flip, and `appState.current` advances synchronously before the relock's await.

## Verification

35 tests across 5 specs: lock past the grace period, no lock within it, biometrics without a PIN, no lock configured, a remote-config value honored, the iOS background sequence, a backwards clock still locking, no double-lock per trip, the `isAppLocked` guard, the Android back press swallowed on resume but not on cold start nor the Settings PIN flow, the iOS swipe gate, and "Use PIN" forwarding `isResume`.

Device-tested on both platforms: Android (back press and the double-PIN scenario) and iOS (edge swipe on the resume lock).

### DIFF
--- a/__tests__/navigation/app-state.spec.tsx
+++ b/__tests__/navigation/app-state.spec.tsx
@@ -165,6 +165,21 @@ describe("AppStateWrapper", () => {
       expect(mockSetAppLocked).not.toHaveBeenCalled()
     })
 
+    it("locks when the clock was wound back to fake a short trip", async () => {
+      /** The trip is measured against the wall clock for want of a monotonic one, so winding
+       *  the clock back would otherwise report a negative stay and slip under the grace
+       *  period. A clock that moved backwards is not trusted, and locks. */
+      renderWrapper()
+      await flushEffects()
+
+      await emitAppState("background")
+      jest.setSystemTime(START_TIME_MS - 60 * 60 * 1000)
+      await emitAppState("active")
+
+      expect(mockSetAppLocked).toHaveBeenCalledTimes(1)
+      expect(mockNavigate).toHaveBeenCalledWith("authenticationCheck", { isResume: true })
+    })
+
     it("locks across the ios sequence, which reaches the background through inactive", async () => {
       renderWrapper()
       await flushEffects()
--- a/__tests__/navigation/app-state.spec.tsx
+++ b/__tests__/navigation/app-state.spec.tsx
@@ -1,0 +1,255 @@
+import React from "react"
+import { AppState, AppStateStatus } from "react-native"
+import { render } from "@testing-library/react-native"
+
+import { AppStateWrapper } from "@app/navigation/app-state"
+import KeyStoreWrapper from "@app/utils/storage/secureStorage"
+
+import { flushEffects } from "../helpers/flush-effects"
+
+const mockNavigate = jest.fn()
+const mockSetAppLocked = jest.fn()
+const mockRefetchQueries = jest.fn()
+
+let mockGracePeriodSeconds = 60
+let mockIsAuthed = true
+let mockIsAppLocked = false
+
+jest.mock("@react-navigation/native", () => ({
+  ...jest.requireActual("@react-navigation/native"),
+  useNavigation: () => ({ navigate: mockNavigate }),
+}))
+
+jest.mock("@app/navigation/navigation-container-wrapper", () => ({
+  useAuthenticationContext: () => ({
+    isAppLocked: mockIsAppLocked,
+    setAppLocked: mockSetAppLocked,
+  }),
+}))
+
+jest.mock("@app/config/feature-flags-context", () => ({
+  useRemoteConfig: () => ({ appLockGracePeriodSeconds: mockGracePeriodSeconds }),
+}))
+
+jest.mock("@app/graphql/is-authed-context", () => ({
+  useIsAuthed: () => mockIsAuthed,
+}))
+
+jest.mock("@apollo/client", () => ({
+  ...jest.requireActual("@apollo/client"),
+  useApolloClient: () => ({ refetchQueries: mockRefetchQueries }),
+}))
+
+jest.mock("@app/utils/analytics", () => ({
+  logEnterForeground: jest.fn(),
+  logEnterBackground: jest.fn(),
+}))
+
+jest.mock("@app/utils/storage/secureStorage", () => ({
+  __esModule: true,
+  default: {
+    getIsPinEnabled: jest.fn(),
+    getIsBiometricsEnabled: jest.fn(),
+  },
+}))
+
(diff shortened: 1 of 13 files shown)


esaugomez31:
* **#3921** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3921?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3921?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

grimen (changes requested):
## Findings

**1. The resume lock can be dismissed with back navigation (blocking).** At cold start the lock screen is the initial route, so there is nothing behind it to go back to. On resume it is instead *pushed* onto the live stack, and the lock screens inherit `gestureEnabled: true` — so an iOS edge swipe or one Android back press pops the lock and reveals the app without a challenge. One option: `gestureEnabled: false` on the three lock screens plus ignoring pops while `isAppLocked` (e.g. `beforeRemove` or `BackHandler`).

**2. Grace period rides on the wall clock.** `secondsInBackground` comes from `Date.now()`, so rolling the clock back defeats the relock. RN has no monotonic clock, so this may be an accepted trade-off — a comment saying so would help the next reader.

## Test coverage

- No test pins the back-navigation behavior of the pushed lock screen (finding 1); once hardened, a test that a back press/pop while locked does not reveal content would protect it the same way the cold-start guard test does.

## Code smells

- The post-unlock `if (isResume) { navigation.goBack(); return }` branch is now repeated in three screens (`authentication-check-screen`, `authentication-screen`, `pin-screen`); a shared helper (or folding it into `setAppUnlocked`) would keep the resume contract in one place.
- The background-timestamp capture still leans on the `"inactive".match(/active/)` substring accident the PR body itself calls a landmine; the test pins it, but spelling the intent out in code (e.g. `appState.current === "active" || appState.current === "inactive"`) would remove the trap rather than just alarm it.
- `handleAppStateChange`'s dependency chain (`relockIfGracePeriodExpired` → `isAppLocked`, remote config) makes the AppState listener unsubscribe/resubscribe on every lock-state flip; harmless today, but reading those values from refs would keep the subscription stable.

esaugomez31:
@grimen 

## Findings

1. Fixed, though deliberately not to the letter. `gestureEnabled` is turned off only on resume, not fixed off — a fixed value would trap the header-less PIN-creation flow Settings opens, where the swipe is the only way back. The back press is caught with a `BackHandler` gated on `isResume` rather than `isAppLocked`: the latter flips to `false` a tick before the unlock's own `goBack`, so a `beforeRemove` guard would swallow that navigation too. iOS (swipe) and Android (back press) each need their own guard, neither is redundant.

2. Done, and taken one step past the comment: the wall-clock trade-off is now documented, and a clock caught moving backwards locks instead of being trusted.

## Test coverage

Added, covering both halves. A spec pins `gestureEnabled: false` on resume (the iOS swipe), and the pin-screen specs pin the `BackHandler` swallowing the Android back press without revealing content. Both were checked by breaking the guard and watching them fail.

## Code smells

- Resolved via the shared-helper option: the branch now lives in a `useUnlockScreen` hook the three screens call.
- Resolved exactly as suggested: the substring match is gone, replaced with explicit `appState.current === "active" || appState.current === "inactive"`.
- Resolved exactly as suggested: both values are read through refs, so the listener no longer resubscribes on a lock-state flip.

grimen (approved):

PR #3930, opened 2026-07-20
feat(settings): Add Setting items for merchant tools into "Ways to get paid" group in Settings

designsats:
## Summary

Builds out the *Ways to get paid* settings group into a full merchant-tools surface, and
brings self-custodial accounts up to parity with custodial ones.

Three rows are new — **Donation Button**, **BTCPay Server** and **WooCommerce** — and two
existing rows, **Point of Sale** and **Printable static QR**, now work for self-custodial
accounts, where they previously rendered nothing.

| Tranche | Commits | What it does |
| --- | --- | --- |
| Donation button | `0d8542662` `4372f8db8` `1fcdd884f` `cab7f49c8` `7020cc428` | `AccountDonationButton` row; extracts `useSelfCustodialLightningAddress`; label iteration settling on **"Donation Button"** with native translations in all 28 locales |
| Merchant tools | `92053f4bb` | `AccountBtcpay` and `AccountWoocommerce` rows; extracts `usePayLinks`; `AccountPOS` and `AccountStaticQR` refactored onto it |
| Hardening | `b0addf7ab` | URL-encodes the username before interpolating it into any pay link |
| Link host | `fa80aeab2` | Points the POS and printable QR at `terminal.blinkbtc.com` in **both** modes; custodial still used the pay-server host |
| Copy | `2cf60a952` | Corrects the **WooCommerce** brand capitalization |

`562d0ce30` also resolves `tar` to a patched version for GHSA-23hp-3jrh-7fpw. It is
unrelated to the feature and touches only `package.json` / `yarn.lock`.

### Implementation notes

- **`usePayLinks`** (`app/screens/settings-screen/settings/use-pay-links.ts`) resolves the
  identity every pay link is built from, once, for the whole group. Custodial accounts key
  off `data.me.username`; self-custodial accounts have no username of their own, so they
  key off the local part of their lightning address. `AccountPOS` and `AccountStaticQR`
  previously each re-derived username, base URL and loading state inline; that duplication
  is where their net line reduction comes from.

- **The POS and printable QR hosts are not instance-scoped.** Per the spec both resolve
  against `https://terminal.blinkbtc.com` for *every* account, custodial or not, so the
  host is a module constant in `pay-links` next to `DONATION_BUTTON_URL` and `getPosUrl` /
  `getPrintableQrCodeUrl` take only a username. `appConfig.galoyInstance.posUrl` is no
  longer read for these links — the receive screen still uses it for the
  `.well-known/lnurlp` lookup, which is a genuine pay-server concern.

- **`useSelfCustodialLightningAddress`** prefers the live SDK value and falls back to the
  persisted registry entry while the SDK reconnects, so a user who already registered never
  sees the empty-state prompt flash on cold start.

- **`loading` is gated to the custodial path** (`!isSelfCustodial && loading`). The
  self-custodial address resolves synchronously from the registry/SDK, so only the custodial
  username can arrive over the network and justify a skeleton.

- **Every row returns `null` at the top level**, not from a wrapper, so `SettingsGroup`
  filters it out without leaving a phantom divider.

- **The two plugin rows take no username**, but still hide until one exists: the plugin
  pages are only actionable once there is an address to connect, so they follow the rest of
  the group rather than dangling on an unconfigured account.

- **Username encoding.** `getPosUrl`, `getPrintableQrCodeUrl` and `getDonationButtonUrl`
  each dropped the username straight into a path segment. `validateUsername` constrains
  registered usernames to `[0-9a-z_]`, so this is a no-op for every well-formed value — but
  the value we interpolate does not come from that check, it arrives from the server or the
  Breez SDK. A username carrying `/`, `?` or `#` would otherwise silently alter the path,
  query or fragment of the URL handed to `Linking`.

## Acceptance requirements

**Self-custodial account, with a registered lightning address**

- [x] *Ways to get paid* shows **Point of Sale**, **Printable static QR**, **Donation
      Button**, **BTCPay Server** and **WooCommerce** rows, each with an external-link
      chevron.
- [x] Point of Sale opens `https://terminal.blinkbtc.com/<username>`.
- [x] Printable static QR opens `https://terminal.blinkbtc.com/<username>/print`.
- [x] Donation Button opens `https://donation-button.blink.sv/<username>`.
- [x] BTCPay Server opens `https://www.blink.sv/en/btcpay-blink-plugin`.
- [x] WooCommerce opens `https://www.blink.sv/en/woocommerce`.
- [x] `<username>` is the local part of the lightning address, not the custodial username.

**Self-custodial account, no address yet**

- [x] All five rows are hidden — no empty rows, no stray dividers.
- [x] The lightning-address row reads **"Create address"**; tapping it opens the
      set-address modal.

**Custodial account**

- [x] Point of Sale opens `https://terminal.blinkbtc.com/<username>` and Printable static
      QR opens `https://terminal.blinkbtc.com/<username>/print`, built from the account
      username. These previously pointed at `appConfig.galoyInstance.posUrl`.
- [x] Donation Button, BTCPay Server and WooCommerce rows appear and open the URLs above.
- [x] The address screen's "your cash register" and paycode fields show the same two
      terminal URLs.
- [x] The lightning-address empty state copy is unchanged.

**Both modes**

- [x] Cold-starting with a registered address shows the address immediately — no flash of
      the "Create address" prompt while the SDK reconnects.
- [x] The WooCommerce row reads **WooCommerce**, with a capital C.

**Platforms**

- [ ] Tested on iOS.
- [x] Tested on Android.

### Default state

| Before | After |
| --- | --- |
| image | image |

### Username set - Custodial account

| Before | After |
| --- | --- |
|  image | image |

### username set - non-custodial account

| Before | After |
| --- | --- |
| image | image |

`Note: Screenshots are not pixel-perfect. This is understood and not an oversight. `

### Automated

- `__tests__/screens/settings-screen/settings/pay-link-rows.spec.tsx` — new. Asserts the
  destination URL of every row in both account modes: POS and printable QR resolve against
  the terminal in both, keyed off the account username for custodial and the
  lightning-address username for self-custodial. Asserting the terminal host on the
  custodial path is what fails if the pay-server host is reintroduced. Includes a
  regression case asserting that a self-custodial
  account with **no** registered address does not fall back to a custodial username that
  happens to be present in the query cache — all three rows must render nothing.
- `__tests__/utils/pay-links.spec.ts` — extended. Covers `getDonationButtonUrl`, and adds
  path-traversal, query-injection and fragment-injection cases for each helper, plus an
  assertion that a well-formed username still round-trips untouched.
- `__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx` and
  `skip-queries-when-unauthed.spec.tsx` — updated for the extracted hook.
- Full settings + pay-links suites: 20 suites, 171 tests passing. `tsc:check` clean;
  `eslint:check` reports 0 errors.

### Manual QA

1. **Self-custodial, no address** — fresh self-custodial account, open *Settings → Ways to
   get paid*. Expect none of the five rows, and the address row reading "Create address".
2. **Self-custodial, with address** — register a lightning address and return. Expect all
   five rows to appear; tap each and confirm the destination URLs above.
3. **SDK reconnect** — force-quit and relaunch with a registered address. Expect the
   address to render immediately, never the "Create address" prompt.
4. **Custodial** — switch to a custodial account and confirm POS and printable QR now
   resolve against `terminal.blinkbtc.com` with the account username, not `pay.blink.sv`.
   Check the address screen ("Ways to get paid" → address screen) shows the same URLs.
5. **Locale sweep** — switch the device language to ar, el, hy, ja, th and sr, and confirm
   the new strings render in-script and fit their rows without truncation.

## Notes / behavior changes

- **Custodial POS and printable QR links changed host**, from
  `appConfig.galoyInstance.posUrl` (`https://pay.blink.sv`) to
  `https://terminal.blinkbtc.com`, matching the spec and the self-custodial path. This also
  changes the two URLs shown on the address screen, which builds them from the same
  helpers.
- **Point of Sale and Printable static QR are newly exposed to self-custodial accounts.**
  Both previously rendered `null` for them. Existing self-custodial users will see two new
  rows after updating — QA should expect this, and it is the largest user-visible change
  here.
- **The self-custodial empty-state copy changed** from "Set your lightning address" to
  "Create address". Custodial copy is untouched.
- **The WooCommerce label changed** from "Woocommerce" to "WooCommerce" across the source
  locale and all 28 translations. The `woocommerce` message key is unchanged.
- **URL encoding is a behavior change only for malformed usernames.** Every username that
  passed `validateUsername` produces a byte-identical URL to before.
- No migration needed — all changes are presentational and read-only.


designsats:
@blink-claw-bot Review pls

yarn.lock
kngako, line 1: I don't think we should be making an changes to the dependencies to get the new UI updated?

package.json
```diff
@@ -305,7 +305,8 @@
     "shell-quote": "^1.8.4",
     "@react-native/codegen": "0.85.3",
     "handlebars": ">=4.7.9",
-    "protobufjs": ">=7.5.5"
+    "protobufjs": ">=7.5.5",
+    "tar": ">=7.5.19"
```
kngako: Where is this new "tar" dependency being used?

yarn.lock
designsats: reverted

package.json
designsats: reverted

designsats:
What was wrong: only the self-custodial path used the terminal. Custodial accounts built both links from appConfig.galoyInstance.posUrl → https://pay.blink.sv/<username>. Spec ([blink-wip#988](https://github.com/blinkbitcoin/blink-wip/issues/988)) says terminal.blinkbtc.com/USERNAME and terminal.blinkbtc.com/USERNAME/print for both modes.

The fix — since neither link is instance-scoped any more, the host became a module constant in [pay-links.ts](vscode-webview://15fj1ri4nhk5mb03r2lagd0grvdq4drkpcgqu3n5k9du4jij5j6h/app/utils/pay-links.ts) next to DONATION_BUTTON_URL, and getPosUrl / getPrintableQrCodeUrl dropped their host parameter:

[use-pay-links.ts](vscode-webview://15fj1ri4nhk5mb03r2lagd0grvdq4drkpcgqu3n5k9du4jij5j6h/app/screens/settings-screen/settings/use-pay-links.ts) — baseUrl and the useAppConfig call are gone; the hook is now purely identity resolution
[account-pos.tsx](vscode-webview://15fj1ri4nhk5mb03r2lagd0grvdq4drkpcgqu3n5k9du4jij5j6h/app/screens/settings-screen/settings/account-pos.tsx), [account-static-qr.tsx](vscode-webview://15fj1ri4nhk5mb03r2lagd0grvdq4drkpcgqu3n5k9du4jij5j6h/app/screens/settings-screen/settings/account-static-qr.tsx) — single-arg calls
[address-screen.tsx](vscode-webview://15fj1ri4nhk5mb03r2lagd0grvdq4drkpcgqu3n5k9du4jij5j6h/app/screens/galoy-address-screen/address-screen.tsx#L69-L70) — built the same two URLs from the same helpers, so it was wrong in the same way; corrected by the signature change. This is the one thing beyond the reported scope — say the word and I'll pull it back out.
galoyInstance.posUrl itself is untouched; the receive screen still uses it for the .well-known/lnurlp lookup.
Tests now assert the terminal host on the custodial path too, so reverting either helper fails them. 20 suites / 171 tests pass, tsc and eslint clean.

app/screens/settings-screen/settings/account-donation-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const {
+    theme: { colors },
+  } = useTheme()
+  const { LL } = useI18nContext()
+  const { username, loading } = usePayLinks()
+
+  if (!username) return null
+
+  return (
+    <SettingsRow
```
kngako, line 22: Nitpick: I think we can have a `ExternalLinkButton` that accepts text, action, and icon as an argument so that we can have one reusable component instead of multiple buttons in the form of `AccountWoocommerce`, `AccountBtcpay`, `AccountDonationButton` who technically all share functionality.

Just waiting for my build to finish and I'll be giving this an approve regardless of the nitpick.

designsats:
@blink-claw-bot We are happy. Are you happy with the code?

kngako (dismissed):
Does as it says on the tin.

kngako (approved):
Conflict resolved

designsats:
Got a review comment to change "Create address" string to "Set receive address".
Reasoning - not descriptive enough.

kngako:
Running locally and looks good and behaves correctly.


image

kngako:
image


image

image

image

PR #3931, opened 2026-07-21
chore(graphql): update graphql schemes

kngako:
This is the result of running `yarn dev:codegen`


grimen (approved):
This is a constant wack-a-mole yes - it should be automatic but is not. A complete CI/CD overhaul starting in ~August.

PR #3933, opened 2026-07-21
chore(deps): Remove .storybook_server and dependency so we can get a clean audit

kngako:
It was giving us a critically vulnerable dependency (tar). Doesn't seem like it was being used anywhere?


grimen (changes requested):
Verified the motivation and the blast radius: on `main`, `yarn.lock` has exactly one `tar` entry (`tar@^6.0.2`, pulled in via `@storybook/react-native-server` → `core-server` → webpack4/cacache subtree); on this branch it's zero. No CI workflow references storybook, and on-device Storybook (`sb-rn-get-stories`/`sb-rn-watcher` from `@storybook/react-native`, configured by `.storybook/`) is a separate tree and unaffected. So the dep itself is safe to remove — but the PR leaves the scripts that call it behind:

**1. `sb-server` script is left dangling — please remove it in this PR.** `package.json` line 25 (`"sb-server": "... yarn storybook-server"`) invokes the `storybook-server` bin that this exact dep provides. After this change, `yarn sb-server` fails with "command not found", and its config (`.storybook_server/`) is deleted here anyway. This is the one blocking item.

**2. `storybook` script (line 24) is already dead — remove it while you're in here.** It runs `start-storybook`, but nothing installed provides that bin: `@storybook/react` is pinned at 7.0.7, which ships no bins (Storybook 7 replaced `start-storybook` with `storybook dev`), and this dep's bins are `storybook-server`/`react-native-storybook-server`. Not a regression from this PR, but this cleanup is the natural place to drop it.

### Test coverage

Nothing to add — this is a dependency removal with no runtime surface. CI (`yarn install` + existing checks) is the meaningful verification, plus the audit going clean.

### Code smells

- `@storybook/addon-essentials@6.5.16` was referenced only by the deleted `.storybook_server/main.js`, so it's now likely orphaned — candidate for a follow-up removal after checking no story imports it.
- Title typo: `acitons`; scope would better be `chore(deps)` since this touches dependencies, not GitHub Actions.

grimen (approved):

PR #3934, opened 2026-07-22
feat(migration): custodial migration UI refinements

esaugomez31:
## Summary

UI refinements for the custodial to non-custodial migration flow, on top of the backend-integration branch. It routes the mandatory dollar-balance emptying through the in-app convert, hardens the commit screen so it never implies a dollar conversion and never commits with dollars present, locks the convert controls during a migration conversion, polishes the contact-support screen, and makes the active percentage chip visible.

## Dollar balance conversion into the migration

- From the migration gate's Dollar Balance modal, a single action sends every affected user (restricted regions included) into the existing dollar-to-bitcoin convert screen, pre-filled with the USD wallet as source, BTC as destination, and the whole balance already entered (as if the 100% chip were tapped).
- The region restriction that would otherwise bounce a restricted user home is waived only for a genuine migration step, validated by an in-memory flag the gate arms right before navigating, not a deep-linkable route param.
- The convert screen reuses the existing chip loading state while the amount settles, so it does not flash up from zero.
- On success the convert screens are dropped and the flow resumes through the shared migration entry (the same dispatcher the Settings row uses), so a back press lands on Home and the migration continues.

## Commit screen (balances overview)

- The post-deadline exchange-rate line is gone. It was the only figure implying the moved bitcoin came from converting the user's dollars, and the migration never converts dollars, so it could only mislead. It only ever rendered after the wind-down gate, never on the pre-deadline happy path, which is untouched.
- A remaining dollar balance now blocks the commit before the migration is declared started: the start never arms, and the same dollar-balance modal (Transfer only, no dismiss) sends the user to empty the balance through the in-app convert first. This closes the gap where dollars arriving after the gate could reach the commit.
- The New Dollar Balance still reads `$0`, or "not available" for restricted regions, never a positive figure: it can never promise the user dollars the migration will not deliver.

## Convert screen

- The active percentage chip now renders as pressed (filled) and stays visible after its amount settles, so the migration user sees the whole balance is the active selection. It clears when the amount is typed by hand or a wallet toggle recalculates it.
- During a migration conversion the amount is locked to the whole balance: the wallet toggle, the keypad, and every percentage chip except 100% are disabled natively at the component level, so the migration user cannot change what is being converted.

## Contact-support screen

- Account ID and pubKey render in full, never truncated, so support gets the whole string.
- When the migration failed server-side, the screen names the failure reason so both support and the user see why the flow stopped.
- Back moved to the navigator header, reusing this screen's return-to-commit-point path.
- The support address doubles as a copy control (copies just the address), and a new copy button copies the whole diagnostics block as `label: value` lines. The button label is translated across all locales.

## Testing

- New and updated unit tests across the modal, region guards, migration gate, the commit screen, both convert screens, the migration hooks, the percentage selector, the currency keyboard and the contact-support screen.
- 100% coverage on the new units; the convert screens, commit screen and contact-support cover the new behavior end to end.
- Typecheck and lint clean.


esaugomez31:
* **#3934** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3934?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3917** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3863** <a href="https://app.graphite.com/github/pr/blinkbitcoin/blink-mobile/3863?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

grimen (changes requested):
## Findings

**1. Migration waiver lets a restricted user convert in the opposite direction (blocking).**
Consuming the migration flag disables both region guards for the whole life of the convert screen, but the wallet toggle stays enabled (`canToggleWallet` is unconditionally true in `use-convert-money-details.ts`), so nothing holds the conversion to USD→BTC.

*Example:* a restricted-region user taps Transfer in the migration gate, presses the wallet toggle, and converts BTC→USD — acquiring the dollar balance the restriction exists to prevent.

*Proposal:* one option could be disabling the wallet toggle while `isMigrationConversion` is set; if the server already rejects BTC→USD for restricted regions and this is acceptable client-side, a comment on the waiver explaining that would help.

**2. Copied diagnostics block carries less than the support email.**
`copyDiagnostics` copies only the `diagnostics` rows, while `useMigrationSupportEmail` also includes reason, platform, app version, and country.

*Example:* a user whose mail app can't open the mailto copies the block into their own message, and support receives it without the reason and version lines it triages on.

*Proposal:* one option could be extracting the email's `infoLines` builder and sharing it with `copyDiagnostics`.

## Test Coverage Gaps

- No test covers the wallet toggle during a migration conversion — whether the guard waiver persists (or the direction locks) after flipping to BTC→USD is unasserted either way.

## Code Smells

- Plain conversions now navigate with `params: { returnToMigration: undefined }` instead of no params, changing the persisted nav-state shape for every conversion; spreading the params only when `isMigrationConversion` is set would keep the plain path unchanged.
- `isMigrationConversion` rides `conversionConfirmation`'s route params while the details screen deliberately refuses to trust route params for the waiver; a note in `stack-param-lists.ts` that this param is navigation-only (not a privilege) would keep the pattern from being copied for something that matters.

esaugomez31:
@grimen all items addressed.

## Findings

**1. Migration waiver lets a restricted user convert in the opposite direction — fixed.**
The direction is now locked, not just the guards waived. The wallet toggle is disabled for the whole life of a migration conversion, so nothing can flip it to BTC→USD:

```ts
const isWalletToggleDisabled = !canToggleWallet || uiLocked || isMigrationConversion
```

The keypad and every percentage chip except 100% are disabled the same way (`migrationLockedPercentages`, keypad `disabled={isMigrationConversion}`), so a restricted user can only convert the whole USD balance to BTC.

**2. Copied diagnostics block carries less than the support email — fixed.**
The payload is now built once and shared, so the two can't drift. `useMigrationSupportEmail` builds a single `supportDetails` block (reason, identity, platform, app version, country) into `supportDetailsText`, and both the copy control and the email body use that same string. The error card keeps the shorter `cardDetails` (reason + identity); the clipboard and the email now carry the full block.

## Test Coverage Gaps

- Covered. `conversion-details-screen.spec.tsx` > *"locks the amount controls to the full balance during a migration conversion"* asserts the wallet toggle, the 25/50/75% chips and the keypad are all disabled while 100% stays enabled — so the direction lock is pinned either way.

## Code Smells

- Plain conversions no longer carry `returnToMigration: undefined`. The nav params are spread only when the conversion is a migration step, so the plain nav-state shape is unchanged:
  ```ts
  ...(isMigrationConversion ? { params: { returnToMigration: true } } : {})
  ```
- Added the note on `stack-param-lists.ts`: `isMigrationConversion` is marked navigation-only, never a privilege (the waiver comes from the armed in-memory flag, not the route param), so the pattern isn't copied for something that matters.

## Additional (found while wiring this up)

While removing the restriction that let a post-deadline user reach the migration with a dollar balance, we cleaned up two things on the commit screen (balances overview):

- **Removed the post-deadline exchange-rate line.** It implied the moved bitcoin came from converting the user's dollars, but the migration never converts dollars, so it could only mislead. It only ever rendered after the wind-down gate, never on the pre-deadline happy path, which is untouched.
- **Added a USD re-guard before the commit.** A dollar balance still present on landing blocks the start (it never arms) and shows the same dollar-balance modal (Transfer only, no dismiss) that routes to the in-app convert, catching dollars that arrive after the gate. New Dollar Balance still reads `$0` or "not available" for restricted regions, never a positive figure.

100% coverage on the touched units; typecheck and lint clean.

grimen (dismissed):

grimen (dismissed):

openoms (approved):

grimen (approved):

PR #3433, opened 2025-05-19
feat: onboarding without authentication

tiankii:
imagen
imagen

PR #3939, opened 2026-07-22
fix(home): show the transfer button as disabled

### LINKED ISSUE #3923: fix(home): Transfer button hidden instead of disabled when Dollar Balance unavailable in non-custodial mode
## Description

In non-custodial mode, when the Dollar Balance is not available in the user's region, the Transfer button is completely hidden from the home screen action bar (only Receive, Send, Scan are shown).

In custodial mode with the same Dollar Balance restriction, the Transfer button is visible but greyed out/disabled — which is the correct behavior.

The non-custodial mode should match the custodial behavior: show the Transfer button in a disabled state rather than hiding it entirely.

## Screenshots

**Non-custodial (bug — Transfer button missing):**
- Anon user, dark mode — action bar shows only: Receive, Send, Scan

**Custodial (correct — Transfer button disabled):**
- "thelegend" user, light mode — action bar shows: Transfer (greyed out), Receive, Send, Scan

## Expected behavior

Transfer button should be visible but disabled/greyed out when Dollar Balance is restricted, regardless of custodial vs non-custodial mode.

## Actual behavior

Transfer button is completely hidden in non-custodial mode when Dollar Balance is restricted.

## Labels

mobile

kngako:
Have transfer button disabled i dollar balance is restricted in region or transfer is blocked on self custodial)

https://github.com/blinkbitcoin/blink-mobile/issues/3923

### DIFF
--- a/__tests__/screens/home.spec.tsx
+++ b/__tests__/screens/home.spec.tsx
@@ -675,7 +675,9 @@ describe("HomeScreen", () => {
     },
   )
 
-  const testDisabledTransferButton = async (shouldShowDollarModal: boolean = true) => {
+  const testDisabledTransferButton = async (
+    shouldShowDollarTransferDisabledModal: boolean = true,
+  ) => {
     mockTransferBlockedOverride = true
     currentMocks = generateHomeMock({
       level: AccountLevel.Two,
@@ -698,7 +700,7 @@ describe("HomeScreen", () => {
 
     fireEvent.press(getByTestId("transfer"))
 
-    expect(mockDollarBalanceModalVisible).toBe(shouldShowDollarModal)
+    expect(mockDollarBalanceModalVisible).toBe(shouldShowDollarTransferDisabledModal)
   }
 
   it("Disable the transfer button when transfers are blocked and show disabled model when clicked", async () => {
@@ -750,7 +752,7 @@ describe("HomeScreen", () => {
       needsBackendAuth: false,
     }
 
-    await testDisabledTransferButton(false)
+    await testDisabledTransferButton()
 
     mockActiveWalletOverride = null
   })
--- a/__tests__/screens/home.spec.tsx
+++ b/__tests__/screens/home.spec.tsx
@@ -615,6 +615,7 @@ const selfCustodialReadyWalletOverride = (usdBalance: number) => ({
   needsBackendAuth: false,
 })
 
+// eslint-disable-next-line max-lines-per-function
 describe("HomeScreen", () => {
   beforeEach(() => {
     currentMocks = []
@@ -674,7 +675,7 @@ describe("HomeScreen", () => {
     },
   )
 
-  it("Disable the transfer button when transfers are blocked and show disabled model when clicked", async () => {
+  const testDisabledTransferButton = async (shouldShowDollarModal: boolean = true) => {
     mockTransferBlockedOverride = true
     currentMocks = generateHomeMock({
       level: AccountLevel.Two,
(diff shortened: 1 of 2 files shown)


grimen (changes requested):
## Findings

**1. Inverted disable condition (blocking).**
`isTransferDisabled = isDollarBalanceRestricted || (isSelfCustodial && !isTransferBlocked)` negates the wrong flag: `useTransferBlocked()` returns `true` when transfers *are* blocked, so the button disables in the normal case and enables in the blocked case. The issue's restricted-region repro only passes because `isDollarBalanceRestricted` short-circuits first.
*Example:* A self-custodial user in an unrestricted, unblocked country gets a permanently disabled Transfer button whose press opens a dollar-restriction modal that doesn't apply to them.
*Proposal:* `isDollarBalanceRestricted || isTransferBlocked` matches the PR description.

**2. Custodial transfer-block dropped (blocking).**
The old gate applied `!isTransferBlocked` to both account types, and `useTransferBlocked()` already resolves the per-account-type block list (`custodialTransferBlockedCountries` vs `selfCustodialTransferBlockedCountries`); the new expression only consults it inside the self-custodial arm.
*Example:* A custodial account in a country on `custodialTransferBlockedCountries` now gets a fully enabled Transfer button where it previously had none.
*Proposal:* The fix in finding 1 covers this too — dropping the `isSelfCustodial &&` scoping lets the hook's own account-type scoping apply.

**3. iOS gating removal exceeds the issue's scope (possibly blocking — pending confirmation).**
Issue #3923 only asks that the dollar-restricted non-custodial case show a disabled button, but the PR also deletes the `!isIos || network !== "mainnet" || level 2/3 || isIosWithBalance` matrix, which appears to be a deliberate product gate (see Appendix). Whether this blocks depends on intent: if "always show" is a confirmed product decision this is fine as-is; if not, it reintroduces the Transfer button to exactly the profile the gate was built to hide.
*Example:* An iOS mainnet level-0 custodial account with zero balance now sees an enabled Transfer button where it previously saw none.
*Proposal:* For the implementer to confirm with Juan/Lukas (see Appendix) — if intentional, a note in the PR body would help; otherwise one option is keeping the old visibility gate and only converting the hide-on-restriction path to `disabled`.

## Test Coverage Gaps

- `__tests__/screens/home.spec.tsx` still contains `"hides the transfer button when transfers are blocked"` and the `it.each` convert-button visibility matrix; both contradict "always show" and will fail — they need updating whichever way finding 3 resolves.
- No test encodes the intended new behavior (self-custodial + dollar-restricted renders a disabled button that opens the restriction modal; transfer-blocked renders a disabled button) — exactly the test that would have caught finding 1.

## Code Smells

- `onDisabledPress` always opens `DollarBalanceRestrictionModal`, so a user disabled for the transfer-block reason sees dollar-restriction copy from a different feature; one option is branching the handler on the disable reason.
- `unshift` outlived its justification: the prepend was only separate because insertion used to be conditional; the Transfer button could now be the first element of the `buttons` literal, matching declaration order to render order.

## Appendix: history of the iOS gate (needs verification with Juan/Lukas)

The gate deleted here was introduced and then deliberately relaxed twice, so its removal likely needs product sign-off rather than a silent drop:

- 9160d9e65 (Jun 2023, Nicolas): hid the Transfer button for all iOS **mainnet** users — the mainnet check suggests an App Store review gate, though the commit message doesn't state a reason.
- #3433: relaxed to show it for level 2/3 accounts on iOS mainnet.
- #3451: relaxed further to any iOS user with a positive sats balance (`isIosWithBalance`), including the test matrix still in `home.spec.tsx`.

The only profile still hidden before this PR was iOS + mainnet + level 0/1 + zero balance — i.e. a fresh, empty account. Please verify with Juan/Lukas why this gate exists (App Store review?) before removing it.

grimen:
Follow-up on my review: Lukas confirmed the iOS visibility gate (finding 3 / Appendix) can be removed, so "always show" is fine and that finding is resolved — no need to restore the `!isIos || network !== "mainnet" || level 2/3 || isIosWithBalance` matrix.

Still blocking: findings 1 and 2 (the `isTransferDisabled` condition should be `isDollarBalanceRestricted || isTransferBlocked`) plus the test updates — the `it.each` visibility matrix in `home.spec.tsx` can now be dropped outright along with the "hides the transfer button when transfers are blocked" expectation.

grimen:
Correction to my previous comment: the decision was reversed — the iOS visibility gate should **stay for now**; its removal is deferred to a later, separate change. Finding 3 is therefore blocking again.

For this PR that means: keep the `!isIos || network !== "mainnet" || level 2/3 || isIosWithBalance` visibility gate (and its `it.each` test matrix in `home.spec.tsx`), and only convert the hide-on-restriction behavior to a disabled button — i.e. when the gate shows the button, render it with `disabled: isDollarBalanceRestricted || isTransferBlocked` instead of hiding it. Findings 1 and 2 remain blocking as before.

grimen:
One addition regarding the iOS gate: when restoring it, please add a short inline comment along the lines of `// Do not change this condition without checking with Lukas first` above the `shouldShowTransferButton` gate. Intentionally no more than that — the rationale itself should not be spelled out in code; the pointer is just there so nobody removes the gate prematurely again.

kngako:
Is dollar balance not available in South Africa? 

image

image

kngako:
I think something is wrong with configs. I don't have the transfer option using the version of the Blink app from the playstore.

image

kngako:
Seems SA is actually restricted from the transfer functionality.

kngako:
Upon confirming that the transfer button is actually restricted in South Africa... I can confirm that the PR implement the desired changes correctly.

This is how it would look if ZA wasn't restricted.

image

image

esaugomez31:
The bot does not have access to Remote Config, but I confirm that South Africa (ZA) is restricted for transfers only in self-custodial accounts
CC: @kngako @grimen

grimen (approved):

PR #3942, opened 2026-07-22
fix(home): let header grow with font scale so username and settings menu stay usable

grimen:
## Summary

Fixes blinkbitcoin/blink-wip#931 — the username at the top of the home screen was cut off / overlapping the total balance, and with enlarged OS font settings the settings hamburger became untappable.

**Root cause:** the header's `balanceContainer` was hard-capped at `height: 40 / maxHeight: 40`. The username (`p2`, OS-font-scaled) outgrows that box above ~1.4x scale; since RN views don't clip overflow, the text spilled onto the `BalanceHeader` sibling below. `BalanceHeader`'s full-width touchable renders after the header, so the overlap also stole taps from the menu button (which only has a ~17px hitSlop). The bulletin/notification in the report is incidental — bulletins render inside the ScrollView and don't affect the header.

## Changes

- `balanceContainer`: fixed `height`/`maxHeight` → `minHeight: 40` so the row can grow with the font scale (pixel-identical at 1x: 34px icon button + 6px margin = 40).
- `header`: drop `flex: 1` (with a content-driven parent height, `flexBasis: 0` would collapse the row); add `columnGap: 8`.
- Username text: `numberOfLines={1}`, `maxFontSizeMultiplier={1.5}`, `flexShrink: 1` — long names ellipsize instead of crowding the icon buttons, and the account-switcher caret stays visible; the 1.5x cap is defense-in-depth on compact chrome at extreme accessibility sizes.
- Balance text: add `numberOfLines={1}` so the existing `adjustsFontSizeToFit` actually takes effect on iOS.
- `badgeSlot`: `height: 35` → `minHeight: 35` (same bug class, same screen).

**Out of scope:** a global font-scale policy in `app/rne-theme/theme.ts`, the ad-hoc `adjustsFontSizeToFit` in `transaction-limits-screen.tsx`, and `GaloyIconButton` hit-testing. If more custom in-content headers appear, the 1.5 cap should move to a shared constant.

## Verification

- `yarn jest __tests__/screens/home.spec.tsx` — 50/50 pass
- `yarn tsc --noEmit` and eslint on both files — clean
- Manual check recommended on device/sim: iOS `xcrun simctl ui booted content-size accessibility-extra-large` (and XXXL), Android `adb shell settings put system font_scale 2.0` — username never touches the balance; hamburger, graph, username switcher, and balance (hide-amount) all tappable at max scale.

### DIFF
--- a/app/components/balance-header/balance-header.tsx
+++ b/app/components/balance-header/balance-header.tsx
@@ -87,6 +87,7 @@ export const BalanceHeader: React.FC<Props> = ({
                 style={styles.primaryBalanceText}
                 allowFontScaling
                 adjustsFontSizeToFit
+                numberOfLines={1}
               >
                 {formattedBalance}
               </Text>
--- a/app/screens/home-screen/home-screen.tsx
+++ b/app/screens/home-screen/home-screen.tsx
@@ -712,11 +712,18 @@ export const HomeScreen: React.FC = () => {
             iconOnly={true}
             weight="bold"
           />
-          <View>
+          <View style={styles.usernameSlot}>
             {!loading && usernameTitle && (
               <Pressable onPress={canSwitchAccount ? handleSwitchPress : null}>
                 <View style={styles.profileContainer}>
-                  <Text type="p2">{usernameTitle}</Text>
+                  <Text
+                    type="p2"
+                    style={styles.usernameText}
+                    numberOfLines={1}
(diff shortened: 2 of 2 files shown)


esaugomez31:
Superseded by #3955, which now covers both the header and footer font-scaling fixes for blink-wip#931. Closing as duplicate.

PR #3946, opened 2026-07-22
fix(transaction-detail): open spark transactions using sparkscan

### LINKED ISSUE #3732: Transaction detail: copy button and Mempool link not working

kngako:
As mentioned after merging https://github.com/blinkbitcoin/blink-mobile/pull/3898 the transaction "hash" (which is a UUID instead of a transaction hash) doesn't open when queried on mempool.space. It should actually be opened via sparkscan.io

This change just adds sparkscan as a sparkexplorer.


designsats:
I think its a privacy meassure. It cant open in any explorer.

grimen (commented):
## Findings

**1. Persisted Custom instances never get the new `sparkExplorer` field.**
`resolveGaloyInstanceOrDefault` returns a persisted Custom instance as-is, and no migration backfills the new required field. A Custom instance saved by an older build has `undefined` here at runtime despite the type saying `string`.
*Example:* A user with a persisted Custom instance taps the explorer icon on a spark transaction and the app opens `"undefined" + hash`.
*Proposal:* One option is a fallback where the instance is resolved, e.g. defaults for missing fields in `resolveGaloyInstanceOrDefault` — earlier additions like `fiatUrl` share this gap, so that would cover them all.

**2. PR title fails the Conventional-commits check (blocking).**
The "Conventional commits" job rejects the title because it has no type/scope prefix, and the merge is blocked on it.
*Example:* The check is red on the current head for "Open spark transactions using sparkscan".
*Proposal:* A title like `fix(transaction-detail): open spark transactions using sparkscan` would pass.

## Test Coverage Gaps

- The new `viewInExplorer` branching (UUID with dashes → `sparkExplorer`, hex hash → `blockExplorer`) has no test; the spec changes only add the new field to existing fixtures.

## Code Smells

- The `@app/components/...` import rewrites in two screens and the `spendOrReceiveText` type annotation are unrelated to the fix and add diff noise.
- `"https://sparkscan.io/tx/"` is spelled out seven times across config, developer screen, and tests; a single exported constant would keep them in sync.

kngako:
Tests look good and test the new logic correctly. 👌🏿

grimen (approved):

PR #3947, opened 2026-07-22
fix(settings): rename create address string to set receive address

designsats:
## Summary

Follow-up to #3930, which was merged before this commit landed on the branch.

Review feedback on that PR: **"Create address"** is not descriptive enough for the lightning
address row. This renames the message key `SettingsScreen.createAddress` →
`SettingsScreen.setReceiveAddress` and sets the English string to **"Set receive address"**,
with native translations in all 28 locales.

### Implementation notes

- **The key is renamed, not aliased.** `createAddress` has no other call site — the row in
  `account-ln-address.tsx` is its only consumer — so leaving the old key behind would only
  create drift between the source locale and the translations.

- **Translations are capped at 26 characters.** The row title renders single-line with tail
  ellipsis, so the budget is the length of `"Set your lightning address"`, the string this key
  replaced in #3930. Where a language cannot fit both the verb and "receive" inside that,
  the verb is kept and "receive" is dropped rather than letting the title truncate.

- **Two settings suites stub the i18n context with plain object literals**, so a stale
  `createAddress` entry does not fail an assertion — the component calls
  `LL.SettingsScreen.setReceiveAddress()` on an object that has no such property and the row
  throws during render. Both mocks and the two title assertions are updated here.

## Acceptance requirements

**Self-custodial account, no address yet**

- [x] The lightning-address row reads **"Set receive address"**; tapping it still opens the
      set-address modal.
- [x] The rest of the *Ways to get paid* group is unchanged — the five pay-link rows stay
      hidden until an address exists.

**Custodial account**

- [x] Custodial empty-state copy is untouched; this key is only read on the self-custodial
      path.

**Localization**

- [x] All 28 translation files carry the renamed key — no locale falls back to English.
- [x] `createAddress` appears nowhere in `app/` or `__tests__/`.
- [x] Titles render in-script and fit their row without truncation in ar, el, hy, ja, th and sr.

**Platforms**

- [ ] Tested on iOS.
- [x] Tested on Android.

### Automated

- `__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx` — updated. The LL
  mock and the two empty-state title assertions move to `setReceiveAddress` /
  `"Set receive address"`. Reverting the mock alone reproduces the render-time throw.
- `__tests__/screens/settings-screen/skip-queries-when-unauthed.spec.tsx` — updated. Same
  mock rename; the suite renders the whole settings screen, so the stale key takes it down
  the same way.
- Full settings + pay-links suites: 20 suites, 175 tests passing. `tsc:check` clean;
  `eslint:check` reports 0 errors.

### Manual QA

1. **Self-custodial, no address** — fresh self-custodial account, open *Settings → Ways to
   get paid*. Expect the address row to read "Set receive address", and tapping it to open
   the set-address modal.
2. **Locale sweep** — switch the device language to ar, el, hy, ja, th and sr and confirm the
   row renders in-script and is not ellipsized.

## Notes / behavior changes

- **Copy-only change.** No logic, no navigation and no URL is touched — the row's `onPress`
  and visibility rules are exactly as merged in #3930.
- **The message key changed.** Anything outside this repo reading `SettingsScreen.createAddress`
  (translation tooling, for instance) needs to follow the rename.

### DIFF
--- a/__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx
+++ b/__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx
@@ -57,7 +57,7 @@ jest.mock("@app/i18n/i18n-react", () => ({
   useI18nContext: () => ({
     LL: {
       SettingsScreen: {
-        createAddress: () => "Create address",
+        setReceiveAddress: () => "Set receive address",
       },
       GaloyAddressScreen: { copiedLightningAddressToClipboard: () => "Copied" },
     },
@@ -86,7 +86,7 @@ describe("AccountLNAddress (self-custodial)", () => {
 
     render(<AccountLNAddress />)
 
-    expect(lastRowProps().title).toBe("Create address")
+    expect(lastRowProps().title).toBe("Set receive address")
     expect(lastRowProps().rightIcon).toBeUndefined()
     expect(mockScModal.mock.calls.at(-1)?.[0]?.isVisible).toBe(false)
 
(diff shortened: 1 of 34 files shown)


designsats:
@blink-claw-bot review pls

designsats:
stale now

PR #3950, opened 2026-07-23
fix(test): assert settled highlight state in transaction history spec

grimen:
### Problem

`transaction-history-screen.spec.tsx › highlights none when lastSeen ids are missing` fails on Concourse CI but passes on GitHub CI:

```
Expected substring: "no-highlight"
Received string:    "507f1f77bcf86cd799439012:highlight"
```

### Root cause

With empty `lastSeen` ids the screen renders a transient highlight frame. Sampling the tree per event-loop tick shows:

```
tick 0–1:  <none>
tick 2–3:  507f1f77bcf86cd799439012:highlight   ← what Concourse catches
tick 4+:   507f1f77bcf86cd799439012:no-highlight (settled)
```

The chain: the highlight baseline initializes to `{btcId: "", usdId: ""}`, so `shouldHighlightById` falls back to `txId === latestTxIdForCurrency` and highlights the latest tx per currency; then `markTxSeen` writes the cache and the baseline backfill adopts those ids, turning the highlight off. The test waited only for row *existence* and asserted synchronously, so it raced this window.

Concourse (floating `node:24-alpine`, musl) lands the `waitFor` poll inside the window consistently — all `jest.retryTimes(2)` attempts fail. GitHub CI (Node 24.13.0 pinned via `.nvmrc`) polls after it closes.

### Fix

- Move the `no-highlight` assertions inside `waitFor` so the test asserts the settled state.
- Make the "highlights transactions newer than min lastSeen" assertions strict: `toContain("highlight")` also matches `"no-highlight"` and could never fail. Assert `":highlight"` instead; with strict matching the first assertion pair also needs `waitFor` (the first data render briefly shows `no-highlight` before the baseline initializes).

Verified: spec passes 6/6 across repeated local runs; eslint/tsc clean.

### Follow-ups (not in this PR)

- Pin `ci/image/Dockerfile` to `node:24.13.0-alpine` to match `.nvmrc` and eliminate runtime drift between the two CIs.
- The transient highlight flash is real app behavior for never-seen users, and the mapper unit spec ("highlights only the latest when the currency has never been seen") contradicts this screen spec's settled expectation ("highlights none") — needs a product decision.

### DIFF
--- a/__tests__/screens/transaction-history-screen.spec.tsx
+++ b/__tests__/screens/transaction-history-screen.spec.tsx
@@ -397,16 +397,17 @@ describe("TransactionHistoryScreen", () => {
       </ContextForScreen>,
     )
 
+    // The screen transiently highlights the latest tx per currency until the
+    // markTxSeen -> cache -> baseline-backfill roundtrip settles, so wait for
+    // the settled state instead of asserting right after the rows appear.
     await waitFor(() => {
-      expect(screen.getByTestId("transaction-by-index-0")).toBeTruthy()
+      expect(screen.getByTestId("transaction-by-index-0").props.children).toContain(
+        "no-highlight",
+      )
+      expect(screen.getByTestId("transaction-by-index-1").props.children).toContain(
+        "no-highlight",
+      )
     })
-
-    expect(screen.getByTestId("transaction-by-index-0").props.children).toContain(
-      "no-highlight",
-    )
-    expect(screen.getByTestId("transaction-by-index-1").props.children).toContain(
-      "no-highlight",
-    )
   })
 
   it("highlights transactions newer than min lastSeen when ALL", async () => {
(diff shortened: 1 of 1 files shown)


openoms (approved):

openoms:
straightforward changes, good to merge

openoms:
### Review

Traced the claimed race against the implementation (`transaction-history-screen.tsx:268-296`, `transaction-highlight.ts:23-30`) and the root-cause analysis checks out:

- With empty `lastSeen` ids the baseline initializes to `null`, an effect then sets it to `{btcId: "", usdId: ""}`, and `shouldHighlightById` falls back to `txId === latestTxIdForCurrency`, so the latest tx per currency briefly highlights. `markTxSeen` writes the cache, the backfill effect adopts those ids into the baseline, and the highlight turns off. Settled state is `no-highlight`, so the new assertion target is correct.
- The `:highlight` tightening is a genuine fix: `"no-highlight".includes("highlight")` is true, so the old assertions in the ALL test could never fail.
- The ALL test's settled state is stable: both baseline ids are present so the backfill never overwrites them, and `markTxSeen` updating `lastSeen` doesn't touch the baseline, so `:highlight` persists.

**Verdict: approve.** The fix addresses the real race rather than papering over it with retries or timeouts. Non-blocking suggestions below.

#### 1. The "doesn't flip off" `waitFor` no longer verifies anything (minor)

After this change the second `waitFor` block (lines 443-451) is identical to the first. Since `waitFor` resolves on its first successful poll and the condition was already true when the first block resolved, the second block passes immediately and would not catch a flip that happens a tick later. This weakness predates the PR, but the edit makes the two blocks fully redundant. To actually test "doesn't flip off after `markTxSeen`", flush pending work between assertion rounds:

```ts
await act(async () => new Promise((resolve) => setImmediate(resolve)))
expect(screen.getByTestId("transaction-by-index-0").props.children).toContain(":highlight")
```

Alternatively drop the second block, since as written it only adds false confidence.

#### 2. `waitFor` proves "eventually observed", not "settled" (minor, acceptable)

In the "highlights none" test there is also a transient `no-highlight` frame *before* the baseline initializes (the comment in the other test says as much). A poll could in principle match that early frame rather than the settled one. The outcome is the same either way here, but if the settled behavior ever changed to "highlight", the test could still pass by catching the early frame. Not worth blocking on; a deterministic flush-then-assert (as in point 1) would fully close it.

#### 3. Optional strictness

``toBe(`${txId}:no-highlight`)`` instead of `toContain` would also pin which tx renders at which index and give clearer failure output. `toContain(":highlight")` is sufficient for correctness though.

#### Follow-ups

Scope discipline is good: the Dockerfile pin and the mapper-spec vs screen-spec contradiction are correctly deferred. Both are worth tracking as issues so they don't get lost; the floating `node:24-alpine` CI image is the actual environmental root cause and will bite other specs too.

grimen:
Superseded by #4188, which fixed the same flaky `transaction-history-screen.spec.tsx` highlight test more robustly (`waitForStableRender`) and merged 2026-08-20.

PR #3951, opened 2026-07-23
ci(test): move unit tests to self-hosted runners

openoms:


openoms:
The first run of the Test https://github.com/blinkbitcoin/blink-mobile/actions/runs/30012251597/attempts/1?pr=3951 took 17 mins.
Triggered again to check if the cache is working.
Second run was 14 mins.

grimen (approved):

PR #3953, opened 2026-07-23
feat(migration): update migration banner copy

### LINKED ISSUE #3952: feat(migration): Update migration banner copy
## Description

Update the migration banner text to clearly communicate the timeline and consequences.

### Current copy
- **Title:** Important
- **Body:** Migrate your account to new, non-custodial mode before August 31
- **Button:** Migrate

### New copy
- **Title:** Important
- **Body:** Move to a non-custodial account via guided migration, or withdraw, before 31 August. Receiving stops 10 August.
- **Button:** Migrate

### Why
The current text doesn't mention that receiving will be disabled on August 10, or that withdrawal is an alternative to migration. Users need to understand the full timeline and their options.

---
Requested by Lukas via Jonas.

grimen:
Closes #3952

## What

Updates the pre-cutoff migration reminder bulletin body copy to communicate the full wind-down timeline and options:

- **Before:** "Migrate your account to new, non-custodial mode before {date}"
- **After:** "Move to a non-custodial account via guided migration, or withdraw, before {date}. Receiving stops {receiveStopsDate}."

Title ("Important") and button ("Migrate") are unchanged.

## How

Both dates stay server-driven, per the existing convention that wind-down dates follow the backend rather than being hardcoded: the new `{receiveStopsDate}` interpolates `windDown.receiveDisabledAt` through the same `formatDayAndMonth` path the final deadline already uses. `receiveDisabledTimestamp` is threaded from `useMigrationReminderBulletin` through the home screen into the bulletin component.

`yarn update-translations` regenerated `i18n-types.ts` and `raw-i18n/source/en.json`. The new body is machine-translated into all 28 other locales in this PR (following the #3937 pattern), reusing each locale's established wind-down terminology and keeping both placeholders intact. `receiveDisabledAt` is confirmed non-null (`Timestamp!`) on `AccountWindDown` in the production schema.

## Testing

- `yarn tsc --noEmit` passes
- `migration-reminder-bulletin.spec.tsx` updated for the new prop and copy; 3/3 pass locally
- `use-migration-reminder-bulletin.spec.ts` asserts the `receiveDisabledAt` passthrough; `home.spec.tsx` hook mock updated to the new shape — both suites green locally


kngako (approved):

openoms (approved):
Reviewed the copy change and the date threading end to end. Approving.

**Verified:**
- `receiveDisabledAt` is `Timestamp!` in the generated schema, already fetched by the `WindDown` query and mapped in `use-wind-down-status.ts` - no query change needed and no undefined risk.
- The `?? 0` fallback mirrors the existing `finalDeadline ?? 0` pattern; when `windDown` is null the bulletin is not visible, so the epoch-zero date is never rendered.
- The second `formatDayAndMonth` call inherits the malformed-timezone fallback, so a bad backend timezone still cannot crash the home screen.
- `NotificationCardUI` sets no `numberOfLines`, so the longer body wraps instead of clipping.
- If `receiveDisabledAt` passes while the copy says "Receiving stops {date}", the status leaves PreCutoff and the migrate-now modal takes over, so the past-tense case never displays.
- All 28 locale files are updated and each keeps both `{date}` and `{receiveStopsDate}` placeholders (checked the reordered ones: de, nl, hu, tr). Placeholder types regenerated correctly (`RequiredParams<'date' | 'receiveStopsDate'>`).
- Ran the three affected suites locally: 60/60 pass.

**Non-blocking notes:**
- The machine-translated locales could use a native-speaker pass eventually, consistent with the #3937 precedent.
- "or withdraw, before {date}" reads slightly clunky in English; "or withdraw your funds before {date}" would match how most translations render it. Fine as is if the copy is verbatim from #3952.

PR #3958, opened 2026-07-23
fix(android): BitcoinPriceWidget stored REALTIME_PRICE defaults to null

### LINKED ISSUE #3843: fix(android): BitcoinPriceWidget crashes parsing "No Data" default as JSON
## Summary

Production crash (Crashlytics): the Bitcoin price widget crashes its broadcast receiver when no realtime price has been stored yet.

```
Fatal Exception: java.lang.RuntimeException: Unable to start receiver com.galoyapp.BitcoinPriceWidget:
org.json.JSONException: Value No of type java.lang.String cannot be converted to JSONObject
  at com.galoyapp.BitcoinPriceWidgetKt.updateAppWidget(BitcoinPriceWidget.kt:135)
  at com.galoyapp.BitcoinPriceWidget.onUpdate(BitcoinPriceWidget.kt:39)
  at android.appwidget.AppWidgetProvider.onReceive(AppWidgetProvider.java:66)
  ...
```

**Crashlytics:** https://console.firebase.google.com/u/2/project/galoyapp/crashlytics/app/android:com.galoyapp/issues/2b0cd71625c81d731088218fe9ed9e6a?time=7d&types=crash&sessionEventKey=6A29C36E025E00014F6285FE2E5C1D2D_2228305925547538137

## Root cause

`android/app/src/main/java/com/galoyapp/BitcoinPriceWidget.kt:135`:

```kotlin
val realtimePrice = JSONObject(prefs.getString("REALTIME_PRICE", "No Data") ?: "{}")
```

The default passed to `getString` is the literal string `"No Data"`, which is fed straight into `JSONObject(...)`. `"No Data"` is not valid JSON, so the parser fails on the first token — matching the crash message (`Value No ... cannot be converted to JSONObject`).

This triggers whenever `REALTIME_PRICE` has not been written yet — i.e. before `FetchPriceWorker` has ever succeeded. It is reliably reproducible on first widget placement: `onEnabled()` calls `updateAppWidget()` synchronously (line 63) **before** the enqueued worker runs (lines 46–58), so on a fresh install with empty `bitcoinPricePrefs` the bogus default is read and the receiver crashes. It can also occur offline / before the first successful fetch.

Two compounding issues:

1. **Wrong default sentinel.** The no-data path expects valid JSON with a `noData` flag — the HTTP-error path writes `{ "noData": true }` (line 222) and line 136 checks `realtimePrice.has("noData")`. The default should be `{"noData": true}` (→ error branch), not `"No Data"`.
2. **`?: "{}"` is dead code.** `getString` with a non-null default never returns null, so the Elvis guard never fires. Even if it did, `"{}"` has no `noData` and would then crash reading `btcSatPrice`.

## Proposed fix

Use a valid no-data default and treat any unparseable stored value as no-data instead of crashing:

```kotlin
val realtimePriceRaw = prefs.getString("REALTIME_PRICE", null)
val realtimePrice = try {
    if (realtimePriceRaw != null) JSONObject(realtimePriceRaw) else JSONObject("""{"noData": true}""")
} catch (e: org.json.JSONException) {
    JSONObject("""{"noData": true}""")
}
```

`PRICE_ARRAY` (line 133) already defaults to valid `"[]"` so it isn't the crasher, but the same try/catch hardening could wrap it for corrupt-pref robustness.

---
_Backlog: production crash on the price widget; scoped to users who add the widget. Not app-core, but a hard crash._

kngako:
The Bitcoin price widget crashed its broadcast receiver when no realtime price has been stored yet because it used "No data" as a default value and would only provide a empty json object if it found a null on the REALTIME_PRICE key.

Just replacing the default value with null fixes the problem. 

Second problem was that it would check for a `noData` key in the json object, which I guess would be returned from the backend if the backend had no data to provide. Fix for the second problem would be checking if the json had object had the desired `btcSatPrice` key with the price data (might need to confirm the backend doesn't return this when there's errors).  

Replicated the problem and tested the solution by removing the `REALTIME_PRICE` preference in the BitcoinPriceWidget onEnabled.

```
super.onEnabled(context)

val prefs = context.getSharedPreferences("bitcoinPricePrefs", Context.MODE_PRIVATE)
prefs.edit().remove("REALTIME_PRICE").apply()
```

image

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -135,6 +135,12 @@ android {
             }
         }
     }
+
+    testOptions {
+        unitTests {
+            includeAndroidResources true
+        }
+    }
 }
 
 dependencies {
@@ -158,6 +164,10 @@ dependencies {
 
     androidTestImplementation('com.wix:detox:+')
     implementation 'androidx.appcompat:appcompat:1.1.0'
+
+    testImplementation("junit:junit:4.13.2")
+    testImplementation("androidx.test:core:1.6.0")
+    testImplementation("org.robolectric:robolectric:4.16")
 }
 
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -168,6 +168,8 @@ dependencies {
     testImplementation("junit:junit:4.13.2")
     testImplementation("androidx.test:core:1.6.0")
     testImplementation("org.robolectric:robolectric:4.16")
+    testImplementation("androidx.work:work-testing:2.9.0")
+
 }
 
 
--- a/android/app/src/main/java/com/galoyapp/BitcoinPriceWidget.kt
+++ b/android/app/src/main/java/com/galoyapp/BitcoinPriceWidget.kt
@@ -132,9 +132,8 @@ internal fun updateAppWidget(context: Context, appWidgetManager: AppWidgetManage
     val prefs = context.getSharedPreferences("bitcoinPricePrefs", Context.MODE_PRIVATE)
     val priceArray = JSONArray(prefs.getString("PRICE_ARRAY", "[]"))
 
-    val realtimePrice = JSONObject(prefs.getString("REALTIME_PRICE", "No Data") ?: "{}")
-    if (!realtimePrice.has("noData")) {
-
+    val realtimePrice = JSONObject(prefs.getString("REALTIME_PRICE", null) ?: "{}")
+    if (realtimePrice.has("btcSatPrice")) {
         val bitmap = generateChartBitmap(context, priceArray, minWidth, maxHeight)
         views.setImageViewBitmap(R.id.chart_image_view, bitmap)
 
--- a/android/app/src/main/java/com/galoyapp/BitcoinPriceWidget.kt
+++ b/android/app/src/main/java/com/galoyapp/BitcoinPriceWidget.kt
@@ -34,6 +34,10 @@ import org.json.JSONObject
 import kotlin.math.pow
 
 class BitcoinPriceWidget : AppWidgetProvider() {
+    companion object {
+        internal const val FETCH_PRICE_WORK_TAG = "FETCH_PRICE_WORK"
+    }
+
     override fun onUpdate(context: Context, appWidgetManager: AppWidgetManager, appWidgetIds: IntArray) {
         for (appWidgetId in appWidgetIds) {
             updateAppWidget(context, appWidgetManager, appWidgetId)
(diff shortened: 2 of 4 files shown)


kngako:
Need to add:

- Android  native unit test for the widget (probably will also use roboelectric or something like that)
- Action Workflow for automated running of the new android native unit tests

kngako:
The current BitcoinPriceWidgetTest unit test is failing with the following error: 

```
java.lang.NullPointerException
	at java.base/java.io.File.<init>(File.java:278)
	at com.facebook.soloader.ApplicationSoSource.getNativeLibDirFromContext(ApplicationSoSource.java:38)
	at com.facebook.soloader.ApplicationSoSource.<init>(ApplicationSoSource.java:34)
	at com.facebook.soloader.SoLoader.addApplicationSoSource(SoLoader.java:506)
```

Which suggests I need to figure out how to get all the JVM related things that Robolectric depends on linked into the react-native build.

kngako:
Current issue has to do with the us having to configure our own TestWorkManager and also having it process UI updates.

UI updates need to run on the UI thread... and our tester is currently not running on that thread and so we experience issues like the following exception.

```
java.lang.IllegalArgumentException: width and height must be > 0
	at org.robolectric.shadows.ShadowLegacyBitmap.createBitmap(ShadowLegacyBitmap.java:96)
	at org.robolectric.shadows.ShadowLegacyBitmap.createBitmap(ShadowLegacyBitmap.java:85)
	at org.robolectric.shadows.ShadowLegacyBitmap.createBitmap(ShadowLegacyBitmap.java:79)
	at android.graphics.Bitmap.createBitmap(Bitmap.java)
	at com.github.mikephil.charting.charts.Chart.getChartBitmap(Chart.java:1441)
	at com.galoyapp.BitcoinPriceWidgetKt.generateChartBitmap(BitcoinPriceWidget.kt:118)
```

This kinda forces us to either specify a test Executor that runs on the UI thread or... separate the UI updates from the data updates.

grimen:
Pushed two commits to this branch: 654ef93 (deterministic Robolectric setup) and f39ff1b (full widget test suite + three small production fixes). All 8 tests pass locally with no network access. A few findings along the way:

**The `width and height must be > 0` exception is not a UI-thread issue** - `Bitmap.createBitmap(0, 0)` throws on any thread. Under Robolectric no launcher ever sizes the widget, so `getAppWidgetOptions()` returns an empty Bundle and the chart bitmap gets width/height 0. The same can happen in production right after widget placement, so `updateAppWidget` now skips chart generation while dimensions are 0 (the price text still renders).

- The `SynchronousExecutor` config makes `FetchPriceWorker` do a real network fetch inside the unit test, which is why the "while executing work" assertion could not pass deterministically. The worker's network call is now a replaceable function reference (`FetchPriceWorker.fetch`) that tests stub out, and the test WorkManager uses a no-op worker executor.
- `onDisabled` cancelled the tag `FETCH_PRICE_WORK`, but the work requests never carried that tag, so the periodic fetch survived widget removal. Both requests are now tagged.
- The `noData` key you wondered about in the description does not come from the backend - it is `fetchBitcoinPrice`'s own non-200 fallback that the worker stores in prefs. The `btcSatPrice` check handles it correctly.
- Robolectric runs these tests on SDK 35 unless the JVM is Java 21 (required for SDK 36) - worth pinning `@Config(sdk = [35])` or using Java 21 in the CI workflow you mentioned.

Before/after from an API 36 emulator (fresh install, airplane mode):

**Before (main):** widget stays stuck on Loading... and logcat shows the crash this PR fixes (`JSONException: Value No of type java.lang.String cannot be converted to JSONObject` in `updateAppWidget` via `onEnabled`):
before_widget


**After (this branch, offline):** graceful error state, no crash:
after_offline_widget


**After (this branch, online):** live price and chart:
after_online_widget

esaugomez31 (approved):
LGTM, @kngako just try to use conventional commits next time

PR #3959, opened 2026-07-23
test(transaction-detail): cover detail-row copy and explorer icon actions

### LINKED ISSUE #3732: Transaction detail: copy button and Mempool link not working

grimen:
## Summary

Follow-up test coverage for the #3898 icon-row rework (refs #3732), now rebased on #3946 and consolidated with its explorer spec: `transaction-detail-explorer.spec.tsx` is folded into a single `transaction-detail-row-actions.spec.tsx` (GitHub shows this as a rename) so there is one spec, one mock harness, and no duplicated cases.

## What is covered

Renders `TransactionDetailScreen` with onchain, spark and lightning transaction fixtures and presses every row icon:

- hex transaction hash opens on the block explorer (`blockExplorer + hash`)
- spark transaction (UUID hash with dashes) opens on the spark explorer (`sparkExplorer + hash`)
- lightning payment request opens in the lightning decoder (`https://dev.blink.sv/decode?invoice=...`)
- copy actions pass the exact payload for: transaction hash, description, payment hash, preimage, payment request, and Blink internal id

Each wiring was mutation-verified (breaking a handler payload or the explorer branching fails the corresponding assertion).

## Notes

- Test-only change. The #3898 rework shipped with a unit test for the `IconAction` primitive only; none of the per-row handler wirings were exercised, which is also how the spark-UUID-on-mempool mismatch fixed in #3946 went unnoticed.


kngako (approved):
Improved coverage

esaugomez31 (approved):

PR #3960, opened 2026-07-23
fix(info): show registered and detected country in version footer

grimen:
## Summary

The country line under the app version in the settings (and app-update) footer conflated two different things behind one `Country:` value: for custodial accounts it showed only the phone-derived registered country, for non-custodial accounts only the IP-detected one — and on detection failure it could silently fall back to `SV`.

This renders both values explicitly so the footer is accurate for both account types:

```
2.4.60 (build)
Registered: US · Detected: SE
Wallet powered by Blink
```

- **Registered** — parsed from `me.phone` via a new `usePhoneCountryCode` hook (custodial only; `Unknown` for non-custodial/logged-out, which have no registered country).
- **Detected** — existing IP adapter chain via `useIpCountryCode`, showing `Unknown` on failure instead of the `SV` fallback.

`useDeviceLocation` and its consumers (signup gating etc.) are untouched. This is a quick debugging aid until the geolocation ticket lands a more robust detection.

Closes blinkbitcoin/blink-wip#1006

## Remote config placement flag

A new Firebase Remote Config boolean **`footerCountryInfoEnabled`** (default `true`) controls the footer line, in case we change our minds without an app release:

- `true` (default) — the line renders in the version footer, as above.
- `false` — the line is hidden from the footer.

The hidden developer screen **always** renders the line under its own "Country" section header (below the version), regardless of the flag, as a debugging aid. The line is extracted into a shared `CountryInfo` component consumed by both locations. ⚠️ The `footerCountryInfoEnabled` parameter still needs to be created in the Firebase console (baked-in default keeps current behavior until then).

## Test plan

- [x] `__tests__/components/version/version.spec.tsx` — both countries rendered; `Unknown` fallbacks for missing phone country and failed IP lookup
- [x] `__tests__/hooks/use-device-location.spec.ts` — `usePhoneCountryCode`: valid phone → country, no phone → undefined, unparseable phone → undefined
- [x] `__tests__/screens/settings-screen/settings-screen.spec.tsx` still green
- [x] Flag placement: footer hides the line when `footerCountryInfoEnabled` is false; `__tests__/screens/developer-screen-country-info.spec.tsx` locks the developer screen always showing it; default locked to `true` in `feature-flags-default-blocks.spec.ts`
- [x] 100% statement/branch/function/line coverage on `version.tsx`, `country-info.tsx`, and `use-device-location.ts`
- [x] `tsc`, eslint, prettier clean; i18n types regenerated via `yarn update-translations`
- [x] Manual simulator check of the settings footer (both flag states verified on iPhone 16 Pro simulator; screenshots on file)
- [x] Create `footerCountryInfoEnabled` in the Firebase console


kngako:
For my self custodial account in SA this PR does exactly what it says on the tin. 

image

grimen:
Updated it to be hidden/shown via remote config.

01-settings-footer-flag-on
02-settings-footer-flag-off
03-developer-screen-flag-off

esaugomez31 (approved):

PR #3961, opened 2026-07-23
fix(ci): purge build caches and trash in mac worker disk cleanup

grimen:
## Context

Today's prod-build failed on `scaleway-mac-13342` with `ERROR: Only 5.5 GB free ... Required: more than 10 GB` even though `macos-build-disk-cleanup.sh pre` had just run — everything the script knows about was already clean. Manual diagnosis on the worker found ~42 GB in paths the script neither reports nor cleans:

| Path | Size |
|---|---|
| `/private/var/root/Library/Caches` (Yarn/CocoaPods as root) | 20 GB |
| `/private/var/root/.gradle` | 16 GB |
| `/private/var/root/.Trash` | 2.8 GB |
| `/Users/m1/Library/Caches/Yarn` | 3.1 GB |

These were deleted manually to unblock the worker (now ~73 GB free). This PR makes the script handle them itself.

## Changes

- **Cleanup** (`pre`/`scheduled`): purge build cache dirs (Yarn, CocoaPods, node-gyp, Gradle) for root, `$BUILD_HOME` and m1 when a dir exceeds `BUILD_CACHE_MAX_GB` (default 5 GB) or free space is below target — same pattern as the existing `ROOT_DERIVED_DATA_MAX_GB` cap. CI trash dirs are always emptied.
- **Diagnostics**: the disk report now includes cache/trash dir sizes and a "Top consumers in build homes" `du -d1` breakdown of `/private/var/root` and the build homes, so the next invisible consumer shows up in the failure log instead of requiring SSH archaeology.
- **Hint**: the "Manual follow-up" message now points at the top-consumers report (the old stale-live-volumes advice was a dead end in this incident — live volumes were only 1.9 GB).

## Verification

- `bash -n` and `shellcheck` clean.
- `check` mode run locally with worker-like env (`HOME=/Users/m1`, empty `CI_ROOT`): full report prints, missing paths tolerated, exit 0.
- Cache purge logic mirrors the manual cleanup performed on the worker today; the daily `cleanup-mac-build-worker` job will exercise `scheduled` mode after merge.

### DIFF
--- a/ci/tasks/macos-build-disk-cleanup.sh
+++ b/ci/tasks/macos-build-disk-cleanup.sh
@@ -26,6 +26,10 @@ free_gb() {
   awk "BEGIN { printf \"%.1f\", $(free_kb) / 1024 / 1024 }"
 }
 
+kb_to_gb() {
+  awk -v kb="$1" 'BEGIN { printf "%.1f", kb / 1024 / 1024 }'
+}
+
 free_space_below_target() {
   (( $(free_kb) < required_kb ))
 }
@@ -73,6 +77,10 @@ unique_paths() {
 }
 
 build_cache_dirs() {
+  # /private/var/root/Library/Caches is listed whole, not per-tool like the
+  # user homes: root runs of Yarn/CocoaPods/etc. scatter caches across it and
+  # everything under it is droppable on a CI worker (20 GB of mixed root
+  # caches in the 2026-07-23 incident). Keep it whole-dir.
   unique_paths \
     "/private/var/root/Library/Caches" \
     "/private/var/root/.gradle/caches" \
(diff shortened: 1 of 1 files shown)


grimen:
Follow-up from the live incident: after the manual cache purge, the next Android build failed with `Could not read workspace metadata from /private/var/root/.gradle/caches/8.13/groovy-dsl/.../metadata.bin` — a Gradle daemon from before the purge was still alive and referenced the deleted metadata. The script now stops `GradleDaemon` processes and removes the stale `.gradle/daemon` registry whenever it purges a Gradle cache dir, so the automated purge can't reproduce this.

openoms (approved):

openoms:
Piped in and ran the improved script successfully:
image

Ready to be merged.

PR #3962, opened 2026-07-23
fix(android): enable Crashlytics native symbol upload on release builds

### LINKED ISSUE #3925: fix(android): enable Crashlytics native symbol upload — NDK crashes unsymbolicated (SIGABRT in libreactnative.so)
## Problem

Crashlytics is receiving native (NDK) crashes that cannot be symbolicated, e.g. a `SIGABRT` with all frames as raw addresses in `libreactnative.so`:

```
Crashed: Thread: SIGABRT
#00 pc 0x9dd44 libc.so
#01 pc 0x9dd10 libc.so
#02 pc 0x629d88 libreactnative.so (BuildId: 0f886ef5554d247646ab8a9aabc62b4b12cfa24a)
#03 pc 0x628ee8 libreactnative.so
...
```

An `abort()` raised inside `libreactnative.so` is typically an uncaught C++ exception hitting the terminate handler in the RN runtime (RN 0.85.3). Side threads show an active WebView and the ML Kit barcode scanner (`libbarhopper_v3.so`), suggesting the QR-scanning screen — but without symbols this is guesswork and the crash cannot be root-caused.

## Root cause

`android/app/build.gradle` has no `firebaseCrashlytics` config block — native symbol upload is not enabled, so every NDK crash arrives as raw addresses.

## Proposed fix

1. Add to `android/app/build.gradle` (release build type):
   ```groovy
   firebaseCrashlytics {
       nativeSymbolUploadEnabled true
   }
   ```
2. Ensure the release CI pipeline runs `./gradlew app:assembleRelease app:uploadCrashlyticsSymbolFileRelease` (or bundles equivalent) so symbols for `libreactnative.so`, `libhermesvm.so`, and the Breez Spark SDK `.so` are uploaded per release.
3. Once a symbolicated report of the `SIGABRT` above arrives, file a follow-up issue for the actual crash.

## Impact

All native crashes are currently unactionable. This blocks triage of at least one recurring `SIGABRT` in production.

esaugomez31:
## Summary

Enables Crashlytics native (NDK) symbol upload so production crashes stop arriving as raw addresses.

Closes #3925

- `android/app/build.gradle`: adds `firebaseCrashlytics { nativeSymbolUploadEnabled true }` to the release build type, which is what registers the symbol generation and upload tasks.
- `android/fastlane/Fastfile`: the release lane now runs `uploadCrashlyticsSymbolFileRelease` after `assemble`/`bundle`, so symbols are uploaded per release build.

## ⚠️ Required CI variable

The upload authenticates against Firebase, and **the pipeline does not carry that credential today**. A Google service account with permission to upload Crashlytics symbols on the `galoyapp` Firebase project needs to be exposed to the release job as:

```
GOOGLE_APPLICATION_CREDENTIALS=<path to the service account JSON file>
```

**It must be a path to a file on disk, not the JSON value.** Concourse params inject raw values, so `ci/tasks/build.sh` (or the job wiring) has to write the service-account JSON out to a file and export `GOOGLE_APPLICATION_CREDENTIALS` pointing at that path. Passing the JSON straight in would make the Crashlytics plugin treat it as a filename and fail with a confusing file-not-found error.

Until the credential is wired, the upload step is **skipped, not failed**: the release lane skips (with a log line) when `GOOGLE_APPLICATION_CREDENTIALS` is unset or points at a missing file, so ordinary builds don't pay for a doomed upload and keep shipping exactly as they do today, just without symbols. Once it resolves to a real file the upload runs, and a genuine failure (an expired or misconfigured service account) surfaces loudly instead of being swallowed.

For reference, the only Google credentials currently in CI are the GCS artifact-bucket service accounts, which do not grant Crashlytics access.

## Verification

Ran `:app:generateCrashlyticsSymbolFileRelease` locally:

- **134 symbol files generated**, covering the three libraries named in the issue: `libreactnative.so`, `libhermesvm.so` and the Breez Spark SDK.
- The generated `reactnative-arm64-0f886ef5554d247646ab8a9aabc62b4b12cfa24a.sym` matches **exactly** the `BuildId` of the `SIGABRT` reported in the issue, so these symbols would have symbolicated that specific crash.
- Counterfactual: with the `firebaseCrashlytics` block removed, no symbol task exists at all, confirming the config is what enables them.

The upload itself is not exercised here, since no Firebase credential is available locally.

## Follow-up

Step 3 of the issue (filing an issue for the actual `SIGABRT` once a symbolicated report arrives) is intentionally out of scope: it depends on a release going out first.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -133,6 +133,11 @@ android {
             ndk {
                 abiFilters "armeabi-v7a", "arm64-v8a"
             }
+            // Release strips the native libs, so without uploading the symbol files every
+            // NDK crash reaches Crashlytics as raw addresses and cannot be root-caused.
+            firebaseCrashlytics {
+                nativeSymbolUploadEnabled true
+            }
         }
     }
 }
--- a/android/fastlane/Fastfile
+++ b/android/fastlane/Fastfile
@@ -46,6 +46,18 @@ platform :android do
       flags: "--no-daemon --stacktrace --info",
       properties: env_props,
     )
+    # Non-fatal on purpose: the upload needs a Firebase credential the pipeline does not
+    # carry yet, and a failed upload must not stop the release from shipping.
+    begin
+      gradle(
+        task: "uploadCrashlyticsSymbolFile",
+        build_type: "Release",
+        flags: "--no-daemon --stacktrace",
+        properties: env_props,
+      )
+    rescue => e
+      UI.important("Crashlytics native symbol upload failed (non-fatal): #{e.message}")
+    end
   end
 
   desc "Deploy a new version to the Google Play"
--- a/android/fastlane/Fastfile
+++ b/android/fastlane/Fastfile
@@ -46,17 +46,17 @@ platform :android do
       flags: "--no-daemon --stacktrace --info",
       properties: env_props,
     )
-    # Non-fatal on purpose: the upload needs a Firebase credential the pipeline does not
-    # carry yet, and a failed upload must not stop the release from shipping.
-    begin
+    # Skip when the Firebase credential (GOOGLE_APPLICATION_CREDENTIALS) is absent so ordinary
+    # builds don't pay for a doomed upload; when present, a failure surfaces loudly.
+    if ENV["GOOGLE_APPLICATION_CREDENTIALS"].to_s.empty?
+      UI.important("Skipping Crashlytics native symbol upload: GOOGLE_APPLICATION_CREDENTIALS is not set")
+    else
       gradle(
         task: "uploadCrashlyticsSymbolFile",
         build_type: "Release",
-        flags: "--no-daemon --stacktrace",
+        flags: "--no-daemon --stacktrace --info",
         properties: env_props,
       )
-    rescue => e
-      UI.important("Crashlytics native symbol upload failed (non-fatal): #{e.message}")
     end
   end
 
--- a/android/fastlane/Fastfile
+++ b/android/fastlane/Fastfile
@@ -46,10 +46,14 @@ platform :android do
       flags: "--no-daemon --stacktrace --info",
       properties: env_props,
     )
-    # Skip when the Firebase credential (GOOGLE_APPLICATION_CREDENTIALS) is absent so ordinary
-    # builds don't pay for a doomed upload; when present, a failure surfaces loudly.
-    if ENV["GOOGLE_APPLICATION_CREDENTIALS"].to_s.empty?
+    # Skip when GOOGLE_APPLICATION_CREDENTIALS is unset, or set but pointing at a missing
+    # file, so ordinary builds don't pay for a doomed upload; when it resolves, a failure
+    # surfaces loudly.
+    creds = ENV["GOOGLE_APPLICATION_CREDENTIALS"].to_s
+    if creds.empty?
       UI.important("Skipping Crashlytics native symbol upload: GOOGLE_APPLICATION_CREDENTIALS is not set")
+    elsif !File.exist?(creds)
+      UI.important("Skipping Crashlytics native symbol upload: GOOGLE_APPLICATION_CREDENTIALS does not point to an existing file")
     else
       gradle(
         task: "uploadCrashlyticsSymbolFile",


grimen (commented):
## Findings

**1. Symbol upload runs ungated on every build.**
The `build` lane runs in the `dev-build` job on every main merge, so until `GOOGLE_APPLICATION_CREDENTIALS` exists each build pays for a third Gradle invocation plus symbol generation over ~134 `.so` files, only for the upload to fail.
*Example:* a routine merge to main triggers `dev-build`, which spends extra minutes on `generateCrashlyticsSymbolFileRelease` and a doomed upload on the shared `mac-m1` worker that recently had disk-pressure issues.
*Proposal:* one option could be gating the whole block on `ENV["GOOGLE_APPLICATION_CREDENTIALS"]` and logging a skip message when it is absent.

**2. Blanket `rescue` hides a broken credential permanently.**
The `begin`/`rescue` swallows every upload failure, so once the credential is added, a misconfigured or expired service account fails silently and nobody notices symbols stopped arriving.
*Example:* the service-account key is rotated next year, every upload starts failing, and crashes quietly revert to raw addresses with no CI signal.
*Proposal:* if the env gate from finding 1 is added, the rescue could be dropped (or narrowed to transient errors) so the upload fails loudly whenever a credential is actually present.

**3. `GOOGLE_APPLICATION_CREDENTIALS` must be a file path, not the JSON value.**
The PR asks infra to expose the credential as an env var, but Concourse params inject raw values while the Crashlytics plugin expects a path to a service-account file.
*Example:* infra wires `data.values` JSON straight into `GOOGLE_APPLICATION_CREDENTIALS`, and the upload fails with a confusing file-not-found error that the rescue then swallows.
*Proposal:* it could help to note in the PR body or infra ticket that `ci/tasks/build.sh` needs to write the JSON to disk and export the path.

## Code Smells

- The new gradle call uses `--no-daemon --stacktrace` while the neighboring `assemble`/`bundle` calls use `--no-daemon --stacktrace --info`; if dropping `--info` is intentional, aligning or a short comment would help.

grimen:
The required `GOOGLE_APPLICATION_CREDENTIALS` CI variable is now tracked in #3965 (assigned to @openoms).

esaugomez31:
@grimen

**1. Ungated upload — Addressed.** The block now runs only when `GOOGLE_APPLICATION_CREDENTIALS` is set, and logs a skip message when it is absent.

**2. Blanket rescue — Addressed.** Dropped the `begin`/`rescue`; a failure now surfaces loudly whenever a credential is present.

**3. Path, not JSON — Noted.** Tracked in #3965: `ci/tasks/build.sh` will write the JSON to disk and export the path.

**Code smell (`--info`) — Addressed.** Aligned to `--no-daemon --stacktrace --info` to match the neighboring `assemble`/`bundle` calls.

grimen (changes requested):
## Findings

**1. PR description contradicts the failure behavior (blocking).**
The body says upload failures are swallowed with a `begin`/`rescue` so they "cannot break the release", but the code has no rescue. Once `GOOGLE_APPLICATION_CREDENTIALS` is set, an upload failure fails the whole `build` lane, and the inline comment ("a failure surfaces loudly") says this is intentional.
*Example:* Whoever provisions the CI credential reads the body, expects a bad service account to be harmless, and an expired key then blocks the next release build.
*Proposal:* If loud failure is intended, updating the body to match would avoid the surprise; otherwise the described `begin`/`rescue` would restore the documented guarantee.

**2. Credential gate checks presence, not validity.**
`ENV["GOOGLE_APPLICATION_CREDENTIALS"].to_s.empty?` only detects an unset variable. A variable that is set but points to a missing file goes down the upload path and, per finding 1, fails the release lane.
*Example:* Concourse templates the variable to a path before the secret file is actually mounted, and the first release after that plumbing change breaks on the symbol upload instead of skipping it.
*Proposal:* One option could be to also skip (with a distinct warning) when `!File.exist?(creds)`, separating "not configured yet" from "configured but broken".

esaugomez31:
@grimen

## Findings

**1. Symbol upload runs ungated on every build.**
Gated the whole block on `GOOGLE_APPLICATION_CREDENTIALS`. The release lane now skips the upload (with a log line) when it is unset, so ordinary builds no longer pay for a doomed upload. — `53e9947e5`

**2. Blanket `rescue` hides a broken credential permanently.**
Dropped the `begin`/`rescue`. With the env gate in place the upload only runs when a credential is present, so a real failure surfaces loudly instead of being swallowed. — `53e9947e5`

**3. `GOOGLE_APPLICATION_CREDENTIALS` must be a file path, not the JSON value.**
Noted in the PR body that it must be a path to a file on disk and that `ci/tasks/build.sh` (or the job wiring) has to write the service-account JSON out to a file and export the path. Also added a defensive guard: the lane skips when the variable is set but does not resolve to an existing file, so the raw-JSON misconfiguration skips cleanly instead of failing with a confusing file-not-found. The skip log does not echo the value, so a misconfigured JSON credential can't leak into CI logs. — `f26e7c2d5`

## Code Smells

- `--info` on the upload gradle call: aligned with the neighboring `assemble`/`bundle` calls (`--no-daemon --stacktrace --info`). — `53e9947e5`

grimen (approved):

PR #3963, opened 2026-07-24
fix(backup): recover from a declined Drive scope and a revoked access token

### LINKED ISSUE #3927: fix(backup): declined Drive appdata scope yields generic sign-in failure instead of scope re-prompt
## Problem

Crashlytics non-fatal seen in production during Google Drive backup:

```
Drive list query failed (403): {
  "error": {
    "code": 403,
    "message": "Request had insufficient authentication scopes.",
    "status": "PERMISSION_DENIED",
    "details": [{ "reason": "ACCESS_TOKEN_SCOPE_INSUFFICIENT",
                  "metadata": { "service": "drive.googleapis.com",
                                "method": "google.apps.drive.v3.DriveFiles.List" } }]
  }
}
```

`app/screens/self-custodial/onboarding/hooks/use-google-drive-backup.ts` requests the `https://www.googleapis.com/auth/drive.appdata` scope via `GoogleSignin.configure`. Google's granular-consent screen lets users complete sign-in while **unchecking** that scope. When they do, sign-in succeeds but `listAppDataFiles` / `findAppDataFile` fail with the 403 above.

## Current behavior

- The 403 is recorded as a Crashlytics non-fatal (it is an expected user choice, not an app defect).
- `resolveErrorMessage` maps it to the generic "sign in failed" message, giving the user no hint that they need to grant Drive access.

## Proposed fix

- After sign-in, verify the granted scopes (e.g. `GoogleSignin.hasPermissions` / inspect the signed-in user's scopes) and re-prompt with `addScopes` when `drive.appdata` is missing; or catch the `ACCESS_TOKEN_SCOPE_INSUFFICIENT` 403 specifically.
- Show a dedicated "you need to allow Drive access for backups" message with a retry path instead of the generic sign-in failure.
- Stop recording this expected state as a non-fatal (log a breadcrumb instead) — see the companion noise-classification issue.

esaugomez31:
Closes #3927

Two related failures in the Google Drive cloud backup, each in its own commit.

## 1. Declined Drive scope (the ticket)

Google's granular consent lets a user finish signing in with the Drive checkbox unticked. Sign-in reports success, so the app assumed the grant and went straight to Drive, which answered `403 ACCESS_TOKEN_SCOPE_INSUFFICIENT`. That surfaced as a generic "Google Drive sign in failed", which points the user at the wrong thing: they are signed in, they just never granted storage access.

The fix reads the granted scopes back from the sign-in response instead of assuming them. If the appdata scope is missing, the app asks once more for that single scope. If the second answer still does not include it, the flow stops with a dedicated `permission-denied` reason and a message that tells the user what to do:

> Allow Google Drive access to back up your wallet, then try again.

A single re-prompt is deliberate. A first refusal is usually a misread consent screen; a second is a decision, and asking in a loop would trap the user with no way out.

Two supporting changes: the Drive client now classifies a scope-insufficient `403` as `permission-denied` rather than a plain auth error, and a withheld scope is no longer reported to Crashlytics. It is the user's choice, not a defect, and it would otherwise flood the crash reports.

New reason `permission-denied` in `CloudBackupErrorReason`, the new copy in `en/index.ts` and the 28 translation files, and the restore flow's error map updated to route it to the error step.

## 2. Revoked access token (found while testing)

Reproduced on device: remove Blink from the Google account's third party access, then open cloud backup again. Sign-in succeeds and the SDK hands back the access token it still has cached, but Google has already revoked it, so the first Drive call fails with `401` and the user sees the generic sign-in error again. Signing out first does not help, because the dead token stays in the token cache.

Every Drive call now goes through a small wrapper: on a `401` it clears the cached token, asks the SDK for a fresh one and retries the call once. The refreshed token is returned to the caller so the rest of the session uses the working one rather than retrying on every call.

The retry is keyed on the HTTP status, not on the error reason. A `403` shares the same auth reason but means a withheld permission, which a new token cannot fix, so it is not retried. If the refresh itself fails, the original `401` is rethrown: the failure to refresh only says the retry never ran, it is not the diagnosis.

## Testing

Both cases verified on device with a real Google account, including the second one after revoking access from the account settings. Unit tests cover the new paths, with 100% coverage on `use-google-drive-backup.ts` and `google-drive-client.ts`.

## Notes

No config or backend work needed. The iCloud backup hook shares the `CloudBackupHook` interface and is untouched; the new reason falls through to its existing generic message.


grimen (commented):
## Findings

**1. The permission message talks about backing up, but it also shows during restore.**
The new `storageAccessRequired` text says "Allow Google Drive access to back up your wallet, then try again." The restore flow reuses this message through `resolveErrorMessage`, so a user who is trying to *recover* a wallet reads about backing one up.
*Example:* A user reinstalls the app, goes to restore, declines the Drive checkbox twice, and is told to allow access "to back up your wallet" — which is not what they are doing.
*Proposal:* One option is a flow-neutral wording such as "Allow Google Drive access to your wallet backup, then try again", which reads correctly in both flows without needing a second translation key.

**2. The refreshed token is not saved back in `upload` and in the restore flow.**
`callDrive` returns the fresh token after a 401 retry, but `upload` throws it away, so `session.accessToken` keeps the dead one, and the restore hook never updates `accessTokenRef`. Everything still works — each later call just repeats the whole clear-cache-refresh-retry dance instead of doing it once.
*Example:* In the picker with a revoked token, every backup the user taps runs its own refresh round trip before succeeding.
*Proposal:* If this is accepted as a rare-case inefficiency, a short comment would help; otherwise the result types could carry the fresh token back, like `listBackups` already does.

**3. Only one spelling of the scope error is recognized.**
`classifyFailure` looks for the exact string `ACCESS_TOKEN_SCOPE_INSUFFICIENT` in the 403 body. Google's older Drive error format reports the same problem as `insufficientPermissions`, and a body in that shape would fall back to the generic sign-in message instead of the new permission one.
*Example:* If Google serves the legacy error shape, the user sees "Failed to sign in to Google Drive" even though the real problem is the missing permission.
*Proposal:* Since the current string was verified on a real device this may be fine as is — adding `insufficientPermissions` to the check would be a cheap safety net.

**4. A cancelled sign-in still lands in Crashlytics.**
When the user cancels the Google sign-in sheet, the response has `type: "cancelled"`, the code continues to `getTokens()`, and the resulting error is recorded as `Unknown`. This is pre-existing behavior, but it goes against the principle this PR establishes for the declined scope: a user's choice is not a defect.
*Example:* Every user who opens cloud backup, changes their mind, and taps outside the sign-in sheet adds a Crashlytics report.
*Proposal:* Fine to leave for a follow-up — checking for `type !== "success"` before `getTokens()` and returning a non-reported reason would mirror the scope handling.

## Code Smells

- In the multi-file restore path, the downloads run in parallel, so if the token dies right after `listBackups`, each download triggers its own token refresh at the same time. Harmless duplicate work, but worth knowing it can happen.

esaugomez31:
@grimen

## Findings

**1. The permission message talks about backing up, but it also shows during restore.**
Addressed. `storageAccessRequired` is now flow-neutral: "Allow {provider} access to your wallet backup, then try again.", which reads correctly in both flows. Applied across `en` and every translation (each using its own established backup term).

**2. The refreshed token is not saved back in `upload` and in the restore flow.**
Addressed. `CloudBackupUploadResult` and `CloudBackupDownloadResult` now carry the fresh token back like `listBackups` does. The backup flow updates the session token before uploading and the restore hook updates `accessTokenRef` after each download, so a later pick reuses it instead of repeating the refresh.

**3. Only one spelling of the scope error is recognized.**
Addressed. `classifyFailure` now matches the legacy `insufficientPermissions` alongside `ACCESS_TOKEN_SCOPE_INSUFFICIENT`.

**4. A cancelled sign-in still lands in Crashlytics.**
Addressed. Sign-in stops before `getTokens()` when the response is not `success` and returns a non-reported `Cancelled` reason (kept out of the crash reports alongside the withheld scope). It is also dismissed silently now: backup stays put and restore navigates back, instead of the generic sign-in failure.

## Code Smells

- Parallel per-file downloads can each refresh the token at the same time: not addressed, left as-is. The token is now reused across subsequent calls, but the initial parallel batch can still refresh simultaneously; harmless duplicate work, as you noted.

grimen (changes requested):
## Findings

**1. Unknown-reason failures now read "Failed to sign in" on the restore error step.**
The per-file download catch maps a thrown error to `unknown`, and `showFailure` resolves that through `resolveErrorMessage`, whose fallback is `signInFailed` — even though sign-in has already succeeded by that point. The `handlePick` catch keeps the description empty for the same kind of throw, so identical failures get different copy depending on where they happen.

*Example:* a download throws while the picker is being assembled after a successful sign-in; the error step shows "Restore failed" with "Failed to sign in to Google Drive" underneath, pointing the user at the wrong remedy.

*Proposal:* one option is to leave `errorMessage` null for `unknown` (matching the `handlePick` catch), or give `unknown` a neutral fallback instead of the sign-in message.

**2. The new `storageAccessRequired` copy is unreachable once a backup session has started.**
In `use-cloud-backup.ts`, failures from `downloadById` and `upload` always toast the generic `uploadFailed`, ignoring the failure reason — so a scope-insufficient 403 that slips past the proactive check still shows "Failed to upload backup". The proactive check makes this rare, so if the trade-off is intended this is fine as-is — a comment would help.

*Example:* the sign-in response carries no scope list (the defensive skip path), the scope was never granted, and the upload fails with 403; the toast says the upload failed rather than that access must be granted.

*Proposal:* one option is to route non-cancelled upload/download failure reasons through `resolveErrorMessage` the way session-start failures already are.

## Test Coverage Gaps

- The new `permission-denied` and `cancelled` reasons falling through the iCloud hook's `resolveErrorMessage` to the generic message (intentional per the PR notes) have no spec pinning that behavior.

## Code Smells

- `startSession`'s catch reports every failure as `DriveOperation.SignIn`, including `findAppDataFile` errors that now travel through `callDrive` — pre-existing, but the retry work makes the label more misleading in Crashlytics.

esaugomez31:
@grimen

## Findings

**1. Unknown-reason failures reading "Failed to sign in" on the restore error step.**
`showFailure` now leaves the error message empty for the `unknown` reason, matching the `handlePick` catch, so a failure surfacing after sign-in already succeeded no longer points the user at a sign-in remedy.

**2. `storageAccessRequired` unreachable once a backup session has started.**
A `toastFailure(reason)` helper now routes every non-cancelled failure through `resolveErrorMessage`; `startSession`, the existing-file `downloadById`, and `upload` all use it, so a scope-insufficient 403 surfaces `storageAccessRequired` instead of the generic upload-failed toast. The local identity-pubkey path intentionally keeps `uploadFailed`, since that is not a Drive error.

## Test Coverage Gaps

- Added `use-icloud-backup.spec.ts` pinning that the iCloud `resolveErrorMessage` falls through to the generic sign-in message for `permission-denied` and `cancelled`.

## Code Smells

- `startSession` now catches sign-in and the appdata-file lookup separately: sign-in keeps the `sign-in` label, the lookup reports under a new `find` operation, so `findAppDataFile` errors no longer surface as sign-in failures in Crashlytics.

grimen (approved):

PR #3966, opened 2026-07-24
fix(settings): gate username creation behind backup in non-custodial mode

grimen:
Fixes blinkbitcoin/blink-wip#997

## Problem

In self-custodial mode a user can create a username (Lightning address) without having completed the seed-phrase backup. They can then start receiving funds into a wallet that cannot be recovered if the device is lost.

## Changes

- **Settings gate**: `SelfCustodialLightningAddressRow` now renders a new `BackupRequiredModal` instead of the create-address modal when `backupState.status !== Completed`. The modal explains why backup comes first and its primary button leads into the existing backup flow (`selfCustodialBackupMethod`), mirroring `BackupNudgeModal`.
- **Defense in depth**: `useRegisterLightningAddress.register()` refuses with a new `SetUsernameError.BACKUP_REQUIRED` when backup is not completed, mapped to an error message in the shared modal UI.
- New `BackupRequired` i18n strings and `SetAddressModal.Errors.backupRequired` (en source + regenerated types).

The custodial flow is unchanged. The receive-screen paycode prompt needs no gate: in self-custodial mode the invoice type never switches to PayCode without a registered address, and self-custodial onboarding has no username step.

## Tests

- New `backup-required-modal.spec.tsx` (render, copy, navigate-to-backup).
- `account-ln-address.spec.tsx`: gated tap shows the backup prompt and never opens the create modal; existing address still copies; custodial flow unaffected by backup status.
- `use-register-lightning-address.spec.ts`: guard refuses before validation/SDK calls.

## Screenshots

New non-custodial wallet, backup **not** completed ("Your funds are at risk" banner visible), Settings → Create address:

| Before (main) | After (this PR) | Backup flow entry | After backup completed |
|---|---|---|---|
|  |  |  |  |
| Create-address modal opens despite missing backup — "it cannot be changed later!" | Same tap now shows the backup-required prompt | "Back up wallet" lands on the existing backup-method screen | Risk banner gone; Create address opens normally |

### DIFF
--- a/__tests__/components/backup-required-modal.spec.tsx
+++ b/__tests__/components/backup-required-modal.spec.tsx
@@ -1,0 +1,153 @@
+import React from "react"
+import { render, fireEvent } from "@testing-library/react-native"
+
+import { BackupRequiredModal } from "@app/components/backup-required-modal"
+
+const mockNavigate = jest.fn()
+
+jest.mock("@react-navigation/native", () => ({
+  ...jest.requireActual("@react-navigation/native"),
+  useNavigation: () => ({ navigate: mockNavigate }),
+}))
+
+jest.mock("@rn-vui/themed", () => {
+  const colors = {
+    grey0: "#ccc",
+    grey2: "#999",
+    grey5: "#f5f5f5",
+    primary: "#fc5805",
+    black: "#000",
+    white: "#fff",
+  }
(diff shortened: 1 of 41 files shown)


esaugomez31 (approved):
The backup gate works as expected!

PR #3971, opened 2026-07-24
fix(crashlytics): classify expected/transient states as breadcrumbs

### LINKED ISSUE #3928: fix(crashlytics): classify expected/transient states as breadcrumbs — non-fatal noise floods (offline retry loops, fresh wallets, device states)
## Problem

A Crashlytics triage sweep found that nearly all current non-fatals are **expected or transient device states** recorded as errors. They bury real signals and can flood the quota. Confirmed instances, each traced to its source:

**1. Spark SDK connectivity errors (worst offender — unbounded flood)**
`app/self-custodial/logging.ts:42-45` promotes *every* ERROR-level Spark SDK log line to `crashlytics().recordError(new Error(msg))` with no dedup. The SDK logs at ERROR for plain offline conditions, from inside retry loops:
- `Failed to subscribe to server events: ... "dns error"` — [`crates/spark/src/events/server_stream.rs:48`](https://github.com/breez/spark-sdk/blob/main/crates/spark/src/events/server_stream.rs) (retries with backoff)
- `Failed to establish real-time sync update subscription: ... "dns error"` — [`crates/breez-sdk/common/src/sync/background.rs`](https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/common/src/sync/background.rs) (outer loop retries **every 1 second** → an offline device can emit one non-fatal per second)

**2. USDB token absent from a fresh wallet (double-reported, once per session)**
`app/self-custodial/bridge/token-balance.ts` — `getInfo().tokenBalances` is built from the wallet's own token outputs, so a wallet that never held USDB legitimately has no entry. Both `spark-token-not-found` and `spark-token-decimals-missing` fire for every such wallet every session (the `recordErrorOnce` dedup set is in-memory and resets each launch). Caveat to preserve: if the "not present" error ever fires for a user who *does* hold USDB, `getStableBalance` silently shows 0 — that sub-case is worth keeping visible.

**3. Biometrics: expected device states**
`app/utils/biometricAuthentication.ts:14,38` records every rejection from `isSensorAvailable()`/`authenticate()`, including `FingerprintScannerNotEnrolled` ("no enrolled fingers") and plain user cancellations.

**4. No Play services**
`use-google-drive-backup.ts` `reportDriveError` records `Drive sign-in failed: Play services not available` on de-Googled devices — expected device state.

**5. Remote Config fetch failure**
`app/config/feature-flags-context.tsx:391` records `[remoteConfig/failure] fetch() ...` (network/throttling); the app correctly proceeds with defaults.

**6. FCM token timeout**
`app/utils/notifications.ts:31` — `messaging().getToken()` `IOException: TIMEOUT` on flaky network, recorded in the catch.

**7. Lightning address resolution offline**
`app/self-custodial/providers/wallet.tsx:35-43` records `SdkError.NetworkError` for resolve/persist/refresh while offline.

## Proposed fix (systemic, not per-case patches)

Introduce a single error-classification helper at the recording boundary that all these call sites route through:

- **Connectivity-class** errors (match `Unavailable` / `dns error` / `transport error` / `TIMEOUT` / `NetworkError`, or check the connectivity state) → `crashlytics().log()` breadcrumb, never `recordError`.
- **Expected device/user states** (not enrolled, cancelled, no Play services, declined scope, USDB never held) → breadcrumb or dedicated UX, never `recordError`.
- **Actionable defects** → `recordError` as today.
- Route the Spark SDK ERROR log path through `recordErrorOnce` (per-message dedup) so SDK retry loops cannot flood regardless of classification.

## Impact

Signal-to-noise in Crashlytics is currently near zero for non-fatals; real regressions (e.g. the unknown-token-payment guard, genuine SDK failures) are indistinguishable from offline blips.

grimen:
Closes #3928.

## What

Introduces a single error-classification boundary, `app/utils/error-reporting.ts`, and routes **every** `crashlytics().recordError` call site in the app through it (that file is now the only caller — enforced by grep in review):

- **Connectivity-class errors** (gRPC `Unavailable`, `dns error`, `transport error`, timeouts, network failures — pattern list in `CONNECTIVITY_PATTERNS`) become `crashlytics().log()` breadcrumbs tagged `[transient]`, never non-fatals.
- **Expected device/user states** (biometrics not enrolled / cancelled, google-signin cancel / no Play services, iCloud not signed in, USDB token absent on a fresh wallet, offline lightning-address resolution via `classifySdkError`) are declared `expected: true` by small predicates next to each site → `[expected]` breadcrumbs.
- **Actionable defects** record as before; crash-adjacent sites (ErrorBoundary, persistent-state quarantine/rehydrate, account-index reads) pass `alwaysRecord: true` so a defect whose message merely looks connectivity-shaped can never be downgraded.
- The **Spark SDK ERROR log path** gains a digit-normalized per-message dedup key, so the offline sync retry loop (previously one non-fatal **per second**) can flood neither non-fatals nor distinct-issue counts.

The existing wrappers (`reportError`, `logError`, `recordErrorOnce`) keep their signatures and delegate to the boundary, so their ~15 existing importers were reclassified without call-site changes.

## Noteworthy behavior changes

- `spark-token-not-found` / `spark-token-decimals-missing` no longer fire for fresh wallets; decimals-missing stays a defect only when the token entry exists but metadata is malformed. The issue's caveat is preserved by a new guard in `wallet-snapshot.ts`: token absent **while USD transactions exist** records `spark-token-missing-with-usd-history` (the silent-0-balance case).
- `validate-network` and `getUserSettings` failures previously double-reported (once directly, once via the SDK log listener); the direct records are removed and `logSdkEvent` at ERROR level owns recording.
- Wallet-snapshot `withTimeout` expiries ("wallet snapshot timed out") are now breadcrumbs — the UI already retries with backoff.

## Judgment call to review

Payment-flow sites (`use-payment-request`, `send-bitcoin-*`, `use-fee`) use plain pattern classification: a genuine backend regression that presents as timeouts would move from non-fatals to breadcrumbs. This matches the issue's intent (retryable UX exists and offline blips dominated the signal), but flagging it explicitly.

## Verification

- 51 touched test suites / 759 tests green locally (new specs for the boundary, biometric predicate, notifications, iCloud hook; updated specs where behavior intentionally changed).
- `yarn tsc --noEmit` clean; eslint clean (2 pre-existing warnings in `send-bitcoin-destination-screen` verified present on main).
- `git grep "crashlytics().recordError" app` → only `app/utils/error-reporting.ts`.


grimen:
**Before/after evidence** — since this PR has no UI, screenshots can't show the change; instead the table below was generated by replaying the exact scenarios from #3928 (real messages from the triage) through the recording layer and counting what Crashlytics receives. "Before" is origin/main's semantics (every call → `recordError`); "After" is executed against this branch's actual code (temporary replay spec, not committed — the equivalent assertions live in the committed test suites).

| Scenario | Before (non-fatals) | After |
| --- | --- | --- |
| Spark sync retry loop, 60s offline (60 ERROR lines) | 60 | 0 — 60 breadcrumbs |
| Spark server-events subscribe failures (12 lines) | 12 | 0 — 12 breadcrumbs |
| Fresh wallet, USDB never held (per session) | 2 | 0 — 2 breadcrumbs |
| Biometrics: not enrolled + user cancel | 2 | 0 — 2 breadcrumbs |
| Remote Config fetch failure (3 attempts) | 3 | 1 (deduped) — 3 breadcrumbs |
| FCM `getToken` TIMEOUT | 1 | 0 — 1 breadcrumb |
| Lightning address resolve while offline | 1 | 0 — 1 breadcrumb |
| **Control:** genuine SDK defect line | 1 | **1** — still recorded |
| **Control:** novel defect repeated by a retry loop (10 lines) | 10 | **1** (deduped) — 10 breadcrumbs |

Note on the Remote Config row: the generic `[remoteConfig/failure] fetch() …` message is deliberately *not* pattern-matched (it isn't unambiguously connectivity), so it still records — capped at one per session by the dedup key. Offline-shaped variants (timeouts, network failures) downgrade fully.

A dashboard before/after can only come from a released build — the Crashlytics non-fatals view should collapse to roughly the two control rows' worth of signal once this ships.

esaugomez31:
Following up on the before/after evidence above — one more call site needed routing through the boundary.

`app/hooks/use-resolve-transaction-account.ts` (the transaction-detail account resolver) landed on `main` after this PR was opened and called `crashlytics().recordError` directly, breaking the "only `error-reporting.ts` records" invariant. `c8f7ce3` routes its `recordProbeError` through `recordAppError`. Same format, replayed through the recording layer — the equivalent assertions are committed in `__tests__/hooks/use-resolve-transaction-account.spec.ts`:

| Scenario | Before (non-fatals) | After |
| --- | --- | --- |
| Foreign-account probe fails offline (per failed probe) | 1 | **1** — still recorded, + 1 breadcrumb |
| Active-account wallet probe hits a transport error | 1 | **1** — still recorded, + 1 breadcrumb |
| **Control:** empty ownership payload (`ownership probe returned no account`) | 1 | **1** — still recorded, + 1 breadcrumb |

Unlike the rows in the comment above, this site is deliberately **not** downgraded. Every failure is wrapped in a `ProbeNetworkError`, whose class name matches the connectivity heuristic (`/network ?error/i`) regardless of the message. Under plain pattern classification that would sweep the genuine `ownership probe returned no account` defect into a breadcrumb along with the real offline blips — hiding a real bug. So it uses `alwaysRecord: true`, the same escape hatch the crash-adjacent sites use so "a defect whose message merely looks connectivity-shaped can never be downgraded."

Net effect: telemetry counts are identical to `origin/main` (every probe error still records); the only change is that each now also emits a `[defect]` breadcrumb and flows through the single boundary. A message-based predicate could downgrade the genuinely-offline probes later, but that is a behaviour change and out of scope for this rebase.

esaugomez31 (approved):

PR #3978, opened 2026-07-24
fix(send): remove "enjoying this app" feedback popup

grimen:
Fixes blinkbitcoin/blink-wip#923

Removes the "Enjoying the app?" alert that fired 3 seconds after the first completed send, along with everything that existed only to serve it:

- The feedback `Alert` trigger and timer in `send-bitcoin-completed-screen.tsx`
- `SuggestionModal` (free-text feedback → `feedbackSubmit` mutation) — the client no longer uses the mutation; the server schema is untouched
- The `feedbackModalShown` client-only Apollo field (query, setter, cache default, local schema) and regenerated codegen artifacts
- The `app_feedback` analytics event (`logAppFeedback`) and its spec cases
- i18n keys `support.enjoyingApp`, `support.thankYouText`, `SendBitcoinScreen.suggestionInput` from the English source and all 28 translation files (`yarn update-translations` re-run)
- The `react-native-in-app-review` native dependency — its only other consumer was a debug button on the developer screen, which is removed as well (`Podfile.lock` updated via `pod install`; Android is autolinked)
- The e2e step that existed only to dismiss the suggestion modal

Verified with `yarn tsc:check`, eslint on changed files, and the affected jest suites (analytics, send-bitcoin-completed-screen, developer-screen-custom-instance, send-confirmation — 63 tests passing).


grimen:
### Before / after

| Before — popup 3s after a send | Before — "No" opens the suggestion modal | After — same screen, no popup |
|---|---|---|
|  |  |  |

Demo setup: dev build on a throwaway simulator, booted directly into `sendBitcoinCompleted` with mock route params via a temporary (uncommitted) initial-route hook. The popup trigger itself is unmodified `main` code; the after shot is the same flow on this branch, captured well past the 3s mark. Images live on the `pr-3978-assets` branch.

esaugomez31 (approved):
LGTM!

PR #3979, opened 2026-07-24
compliance(mobile): fix limits display 1000 -> 999

### LINKED ISSUE #4033: feat: Expose account limits from backend instead of hardcoding in mobile
## Description

Account limits (e.g. daily send/receive limits per account level) are currently hardcoded in the mobile app copy. These should be exposed by the backend API so the mobile app displays the actual enforced values rather than manually maintained strings.

## Context

This came up during [PR #3979](https://github.com/blinkbitcoin/blink-mobile/pull/3979) (fixing `USD 1,000` → `USD 999` display mismatch with the enforced Level 1 limit). As [noted](https://github.com/blinkbitcoin/blink-mobile/pull/3979#issuecomment-5092539600) by @littledino2112: "Makes sense to expose these limits on the backend."

Hardcoding limits in locale strings creates a maintenance burden and risk of display/enforcement mismatches — the SSF audit draft flagged exactly this discrepancy.

## Suggested approach

1. Backend: expose current account limits (daily send, daily receive, per-transaction, etc.) via the GraphQL API, per account level
2. Mobile: read limits from the API response and interpolate into display strings instead of hardcoding values

---
Filed by Blink-Claw-Bot ⚡ per Jonas's request

grimen:
Fixes blinkbitcoin/blink-wip#739 (cross-repo — close manually on merge).

The SSF audit draft report observed that the Level 0 → Level 1 upgrade copy advertises a **USD 1,000** daily limit while the enforced Level 1 limit is **USD 999** per day.

## The fix is backend-driven

The advertised amount is now read from the value the backend enforces, rather than from a number maintained by hand across 29 locale files. [blinkbitcoin/blink#750](https://github.com/blinkbitcoin/blink/pull/750) (merged 2026-08-14) adds the unauthenticated `Globals.accountLimitsByLevel` field carrying the enforced daily limits per account level, and closes [#4033](https://github.com/blinkbitcoin/blink-mobile/issues/4033).

`useLevel1DailyLimit` (`app/hooks/use-level1-daily-limit.ts`) queries that field, picks the level 1 withdrawal limit, and formats it for the reader's own language. Both surfaces interpolate the result into a `{limit}` placeholder instead of stating an amount in prose, so a limit change on the backend reaches users without a mobile release.

### ⚠️ Blocked on a staging deploy

`Globals.accountLimitsByLevel` is merged but **not deployed** — introspecting both `api.staging.blink.sv` and `api.blink.sv` today returns a `Globals` type without it. Mobile codegen and `graphql-check` both validate against staging (`package.json`), so **Check Code and graphql-check are expected to fail on this branch until that deploy lands**. Nothing in the diff fixes that; it clears itself when staging picks the field up.

Runtime degrades safely in the meantime: an API without the field errors the query, and the hook falls back to `FALLBACK_LEVEL1_DAILY_LIMIT_CENTS` (99900), which renders the audited USD 999 — the same amount the copy stated before this PR.

## Changes

- **`app/hooks/use-level1-daily-limit.ts`** — new hook reading `globals.accountLimitsByLevel`, with the audited USD 999 as fallback, formatted with the active locale.
- **`GetStartedScreen.trialAccountLimits.dailyLimit`** (trial-account limits modal) and **`OnboardingScreen.welcomeLevel1.dailyLimitDescription`** (post-upgrade welcome screen) take a `{limit}` placeholder, in `app/i18n/en/index.ts` (English source of truth) and in all 28 locale files under `app/i18n/raw-i18n/translations/`.
- Armenian carried the English sentence verbatim for both keys; it is now translated.
- `app/graphql/generated.ts`, `app/i18n/raw-i18n/source/en.json` and `app/i18n/i18n-types.ts` regenerated.
- **`PhoneLoginInitiateType` extracted** to `app/screens/phone-auth-screen/phone-login-initiate-type.ts`. It lived inside the phone-login screen, so every consumer pulled firebase app-check in with it and the modal spec had to stub the whole barrel. The barrel re-export stays for existing callers, with a note steering new code at the leaf.
- **The `react-native-modal` stand-in moved to `__mocks__/react-native-modal-mock.tsx`**, behind the `@mocks/` alias that `jest.config.js` already mapped and nothing used; the matching `tsconfig` path is added. The six suites that used the relative helper now point at the alias, `custom-modal.spec`'s inline copy is folded in, and the duplicate under `__tests__/helpers/` is deleted — nine suites now share one stand-in. The remaining inline mocks across the repo are deliberately out of scope.

## Test coverage

- **`__tests__/hooks/use-level1-daily-limit.spec.tsx`** — the hook's edge cases: served value, missing level, query error.
- **`__tests__/components/upgrade-account-modal/trial-account-limits-modal.spec.tsx`** — renders the modal and asserts the *backend-served* amount (USD 1,500 from a mocked response, proving the copy follows the API rather than a constant), the fallback to USD 999 when the API rejects the query, and the benefit items. Two cases render in **Spanish** and **German**, catching a locale that drops the placeholder, reverts to the pre-audit string, or gets English number grouping.
- **`__tests__/screens/onboarding-screen/welcome-level1.spec.tsx`** — the welcome screen locks its own copy, rendering that screen.
- **`__tests__/i18n/locale-parity.spec.ts`** — extended from key parity to **placeholder parity**: every locale must interpolate the same `{...}` argument names as the English source. Key parity alone stays green when a locale keeps the key and loses the value.
- **`__tests__/i18n/limit-copy.spec.ts`** — fails any copy, in English or any locale, that states a daily allowance as a literal amount. Guards the drift class rather than these two keys, so a third surface cannot reintroduce it.
- **`__tests__/mocks/react-native-modal-mock.spec.tsx`** — covers the stand-in's visibility contract, both dismissal handlers and its export shape, so a broken edit fails on one file rather than nine with an opaque "Element type is invalid".
- The upgrade CTA is covered end to end (`beforeSubmit` → `navigate("login", …)` → `closeModal`), asserting the literal wire value `"CreateAccount"` rather than reading `PhoneLoginInitiateType` — an assertion against the constant the component itself reads moves with any change to its value.

## Decisions

- **Backend value over a client constant.** An earlier revision of this branch advertised the amount from a `LEVEL_ONE_DAILY_LIMIT_USD = 999` constant. That fixes the audit instance but leaves the structure that caused it — copy and enforcement can drift again, and pinned tests would assert the stale number is correct. The constant is removed; `FALLBACK_LEVEL1_DAILY_LIMIT_CENTS` covers its role as a last resort. The cost is the deploy dependency above, accepted deliberately.
- **The placeholder-parity check ships with an allowlist, not a clean sheet.** 23 locales carry 47 pre-existing placeholder mismatches unrelated to this PR. They are recorded in a documented `KNOWN_PLACEHOLDER_DRIFT` set that may shrink and never grow, so the check lands green today and still fails all new drift. Fixing those 47 translations is separate work.
- **The shared mock is named `-mock`, not after the package.** A root `__mocks__/react-native-modal.js` would be applied automatically to every suite, silently changing specs that render a real modal today. Opt-in per spec keeps the blast radius at the file that asks for it.
- **Number grouping follows the reader's language, not en-US.** The amount is composed at runtime now, so grouping it is the app's job: German writes 1.500 where English writes 1,500, and a comma is German's decimal separator. Done by reading the active locale in the hook rather than by registering a typesafe-i18n `number` formatter — the latter would mean re-typing the placeholder across all 30 locale files for the same result.
- **`compliance(mobile):` kept as the commit type.** The defect class is the audit finding, not the i18n mechanics; the type carries that, and the Conventional commits check accepts it.

## Verification

- `yarn tsc:check` clean; `eslint` clean across `__mocks__`, `__tests__` and the touched app directories.
- Affected suites green: i18n, mocks, upgrade-account-modal, custom-modal and the other suites sharing the stand-in, conversion-flow, welcome-level1, the hook, and account-type-selection.
- **Mutation-verified**, both new i18n guards, on the exact regression they exist for: reverting `es.json` to the pre-fix English string fails *both* the literal-amount check and placeholder parity; dropping only the placeholder fails parity alone. Restored afterwards.

## Screenshots

Both screens shown on the isolated iOS simulator (iPhone 16 Pro), same build, only the JS bundle switched between `origin/main` (before) and this branch (after). The "after" amount is USD 999 in both the backend-served and fallback paths today.

| Screen | Before (`main`) | After (this branch) |
|---|---|---|
| Trial account limits modal | Modal before: USD 1,000 daily transaction limit | Modal after: USD 999 daily transaction limit |
| Welcome to Level 1 | Welcome screen before: Send up to USD 1,000 per day | Welcome screen after: Send up to USD 999 per day |

_Images hosted on the `pr3979-assets` branch — delete it after merge._


grimen:
**Note on why the amount stays hardcoded (and where $999 comes from):**

- Verified against the deployed backend config: `blink-deployments` sets `withdrawal_level_one = 99900` and `intraledger_level_one = 99900` cents (no per-environment override), i.e. **$999.00/day** for Level 1 on both send paths — matching the SSF observation. The open-source `galoy.yaml` default ($1,000) is not what production runs.
- The value is not fetchable for these screens today. The public API only exposes limits as `me.defaultAccount.limits` (the *current* account's limits at its *current* level). The trial modal advertises Level-1 limits to a Level-0 user, whose query returns Level-0 values ($125); `Globals` exposes no limits and there is no level-parameterized query.
- The Welcome-to-Level-1 screen alone could query live limits, but is kept as static copy for consistency with the modal.
- Guardrail: the new spec pins `USD 999` in both strings so a translation sync can't silently reintroduce 1,000. If the backend value ever changes, the failing spec is the reminder to update copy.

Possible follow-up (separate backend issue): expose per-level account limits on `Globals` (instance config, not account data), then interpolate the amount in both strings to eliminate this drift class entirely.

grimen:
@littledino2112 See comment above, would be most consistent / manageable if this was fetch able from the backend - thoughts?

littledino2112:
Makes sense to expose these limits on the backend

grimen:
**The durable fix here is backend-driven, and as of today it is unblocked.** [blinkbitcoin/blink#750](https://github.com/blinkbitcoin/blink/pull/750) merged this afternoon and adds an unauthenticated `Globals.accountLimitsByLevel` field exposing the enforced daily limits (`withdrawal`, `internalSend`, `convert`, plus the rolling `interval`) for every account level. That is what the trial-account modal and the Level 1 welcome screen should render: the number the backend actually enforces, read at runtime, instead of a number maintained by hand across 29 locale files. It also closes [#4033](https://github.com/blinkbitcoin/blink-mobile/issues/4033), which was filed off this PR after @littledino2112 pointed out that these limits belong on the backend.

**Why this PR still hardcodes `USD 999`.** The field is merged but not deployed. Querying `globals { accountLimitsByLevel { level withdrawal } }` right now returns `parsing error: cannot query field 'accountLimitsByLevel' on type 'Globals'` against both `api.staging.blink.sv` and `api.blink.sv`. Mobile codegen validates against staging, so the query cannot even be generated yet, and shipping it before prod has the field would fail the entire `globals` query the app depends on for supported countries, the lightning address domain and fee information. This PR therefore stays what it was scoped as: the interim compliance patch that corrects the advertised number now, to be replaced by the fetched value once the field is live in both environments.

**What the test commit does in that light.** [`1fabe321`](https://github.com/blinkbitcoin/blink-mobile/commit/1fabe32141d1d94b4cdedce6f2e11f3a430a2f41) touches specs only — no app code and no copy — and exists to keep the interim literal from drifting during that window. The previous guard could not do that job: it asserted `getByText(LL.GetStartedScreen.trialAccountLimits.dailyLimit())`, comparing rendered text against the same source string it renders from, so a regression in `en/index.ts` moved both sides together and stayed green. Each screen now asserts the literal `USD 999` against rendered output, in the spec that renders it, verified by mutation — reverting the copy to `USD 1,000` fails exactly those two assertions, one per screen, independently.

The locks are deliberately per-screen and literal so that the interpolation change removes them cleanly: when the value comes from `accountLimitsByLevel`, each assertion is replaced by one that renders a mocked backend limit and checks the formatted output, and the audit constant disappears from the locale files entirely.

Two smaller corrections rode along in the same commit, both in test-only code. The upgrade CTA is now covered (`beforeSubmit`, then navigation with `{ type: CreateAccount, title, onboarding: true }`, then `closeModal`), which required a module-level `mockNavigate` because the previous `useNavigation: () => ({ navigate: jest.fn() })` returned a fresh spy on every call and could never be asserted against. And the `PhoneLoginInitiateType` stub declared a `SignIn` member that does not exist — the real export at [`phone-login-input.tsx#L106-L109`](https://github.com/blinkbitcoin/blink-mobile/blob/1fabe32141d1d94b4cdedce6f2e11f3a430a2f41/app/screens/phone-auth-screen/phone-login-input.tsx#L106-L109) is `{ Login, CreateAccount }`, which matters now that a test asserts on navigation params.

grimen:
**Mutation proof for the test commit.** Reverting the audited copy in `app/i18n/en/index.ts` back to `USD 1,000` (both strings) and running the same two specs, before and after [`1fabe321`](https://github.com/blinkbitcoin/blink-mobile/commit/1fabe32141d1d94b4cdedce6f2e11f3a430a2f41). Everything else is identical between the two runs.

**Before** — specs as of [`0c854e76`](https://github.com/blinkbitcoin/blink-mobile/commit/0c854e7603769b049c52fc3af5b42f9f73a1a7e2):

```
✓ renders the level 1 benefit items
✕ shows the audited USD 999 daily limit
✓ Renders localized title and description lines
✓ Triggers primary action button with label

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 3 passed, 4 total
```

The regression is caught, so this was never unguarded — but look at *where*. The only failing test renders nothing; it reads the string through `LL` and checks the value. Both tests that actually render a screen passed while that screen displayed `USD 1,000`. And `welcome-level1.spec.tsx` — the spec for the screen now showing "Send up to USD 1,000 per day" — passed in full, because its protection lived in the other file's assertion. That is the circularity: `getByText(LL...dailyLimit())` compares the render against the same source string it renders from, so it cannot notice the value changing.

**After** — same mutation, specs as of `1fabe321`:

```
✓ renders the level 1 benefit items
✕ advertises the audited USD 999 daily limit
✓ renders nothing when isVisible is false
✓ navigates to account creation and closes from the upgrade action
✓ closes without navigating from the secondary action
✓ Renders localized title and description lines
✕ Advertises the audited USD 999 daily limit
✓ Triggers primary action button with label

Test Suites: 2 failed, 2 total
Tests:       2 failed, 6 passed, 8 total
```

Each screen now fails in its own spec, on what the user would actually see:

```
● WelcomeLevel1Screen › Advertises the audited USD 999 daily limit

  Unable to find an element with text: Send up to USD 999 per day
```

Restoring `USD 999` returns both suites to green (8/8), so the failures track the copy and nothing else.

The practical difference is what survives a future edit. Before, deleting the one non-rendering test — the kind that reads as redundant, since it renders nothing and duplicates a string — silently removed the only real guard on both screens. After, each screen's lock sits in the spec that renders it, and deleting it is visibly deleting that screen's coverage.

grimen:
**Non-English rendering.** The existing screenshots cover English, but 28 locale files changed and the risky part is where the number carries grammar or sits inside right-to-left text. Hungarian, Turkish and Arabic below, both screens, same simulator (iPhone 16 Pro, iOS build from 2026-07-24) and same crop box per pair. The **before** side is this branch's base commit `04c292c9` with only `app/i18n` swapped back, so the only difference between the two columns is the audited amount.

### Welcome to Level 1 — `welcomeLevel1.dailyLimitDescription`

| Locale | Before (`04c292c9`) | After (this branch) |
|---|---|---|
| Hungarian | Hungarian before: Küldj naponta akár USD 1,000-t | Hungarian after: Küldj naponta akár USD 999-et |
| Turkish | Turkish before: Günde USD 1,000'a kadar gönderin | Turkish after: Günde USD 999'a kadar gönderin |
| Arabic (RTL) | Arabic before: أرسل حتى USD 1,000 يومياً | Arabic after: أرسل حتى USD 999 يومياً |

Hungarian is the one worth looking at closely: the accusative suffix is bound to how the number is pronounced, so `USD 1,000-t` (*ezret*) becomes `USD 999-et` (*kilencvenkilencet*). The suffix changes with the digits, and it renders correctly. Turkish keeps the dative `'a`, which stays valid for 999. In Arabic the Latin `USD 999` sits inside a right-to-left line and the bidi ordering is unchanged from before.

### Trial account limits modal — `trialAccountLimits.dailyLimit`

| Locale | Before (`04c292c9`) | After (this branch) |
|---|---|---|
| Hungarian | Hungarian modal before: Napi USD 1,000 tranzakciós limit | Hungarian modal after: Napi USD 999 tranzakciós limit |
| Turkish | Turkish modal before: Günlük USD 1,000 işlem limiti | Turkish modal after: Günlük USD 999 işlem limiti |
| Arabic (RTL) | Arabic modal before: حد المعاملات اليومي USD 1,000 | Arabic modal after: حد المعاملات اليومي USD 999 |

**Forced state, stated plainly:** the modal normally opens from an authenticated screen, so for these shots it was TEMP-mounted with `isVisible` over the Level 1 welcome screen — that is why the welcome copy shows dimmed behind the card. Only the mount point was forced; the modal itself is the real component rendering real locale strings. Both TEMP edits (the mount and a boot-route override) were reverted, and the branch contains neither.

<sub>Images live on branch `assets/pr-3979-locales`, embedded by commit SHA — deleting that branch breaks them. The English screenshots in the PR body are on `pr3979-assets` and need to survive too.</sub>

esaugomez31 (changes requested):
Here is what I found in the review (I'm already working on them):

## HIGH

**1. Armenian is the only locale left with the English fallback**

`app/i18n/raw-i18n/translations/hy.json:2065` and `:2639` still read `"USD 999 daily transaction limit"` and `"Send up to USD 999 per day"` in English. Every sibling key in that same block (`modalTitle`, `noBackup`, `sendingLimit`, `recoveryOption`, `onchainReceive`) is in Armenian, and the other 27 locales were translated in this PR. The diff edits exactly those two lines, so this was the moment to translate them. An Armenian-locale user sees English copy in the trial-account-limits modal and on the Welcome to Level 1 screen.

## MEDIUM

**2. The extracted modal mock drops `onBackdropPress` and `onBackButtonPress`**

`__tests__/helpers/react-native-modal-mock.tsx:12` renders only `<View>{children}</View>`. `custom-modal.tsx:90-91` wires both handlers to `toggleModal`, and `__tests__/components/custom-modal.spec.tsx:31-49` deliberately exercises them through `backdrop` / `back-button` testIDs. Migrating that spec (or any spec covering `dismissable`) to the helper would silently drop that coverage. There are still 16 specs mocking `react-native-modal` inline and they are not byte-identical (migrate-now-modal returns `{ __esModule: true, default }` and builds the view through `jest.requireActual`), so as it stands this adds a 17th variant instead of consolidating the existing ones.

**3. The new leaf module is not used by production, so the import cycle stays**

`app/navigation/stack-param-lists.ts:10` still imports `PhoneLoginInitiateType` from the barrel `@app/screens/phone-auth-screen`; the barrel re-exports `phone-login-input.tsx`, which imports back `../../navigation/stack-param-lists`. That is the cycle. The same barrel import is in `trial-account-limits-modal.tsx:9`, `get-started-screen.tsx:28`, `accept-t-and-c.tsx:20`, `account-type-selection.tsx:21`, `login-method-screen.tsx:18` and `root-navigator.tsx:29`. `phone-login-initiate-type.ts` has zero imports and is exactly the leaf that breaks it, but only a test mock benefits today. Pointing at least `stack-param-lists.ts` and `trial-account-limits-modal.tsx` at the leaf removes the cycle and makes the spec's firebase workaround unnecessary.

**4. The navigation-type assertion is self-referential**

`__tests__/components/upgrade-account-modal/trial-account-limits-modal.spec.tsx:19-21` mocks the barrel with `jest.requireActual` of the leaf, and line 24 imports the same leaf directly, so the assertion on line 84 reads `PhoneLoginInitiateType.CreateAccount` from the very module object the component reads. Changing the constant's value from `"CreateAccount"` to `"createAccount"` moves both sides together and the test stays green. Only a key rename is caught, and tsc already catches that. The comment on lines 17-18 ("keep the real wire values so a rename of them breaks the navigation assertion") therefore overstates the guarantee; the literal it replaced did pin the wire value.

**5. Mocking the whole barrel leaves the new re-export covered by nothing**

`trial-account-limits-modal.tsx:9` imports from the barrel, but the spec replaces that barrel wholesale with the leaf module. Deleting `export * from "./phone-login-initiate-type"` from `app/screens/phone-auth-screen/index.ts` keeps the spec green while the real component gets `undefined` and throws on press. Likewise, any future barrel import added to the component resolves to `undefined` inside the spec with no diagnostic. Fixing finding 3 (import the leaf in the component) lets the barrel mock go away entirely.

**6. Bare `require()` in the mock factory**

`trial-account-limits-modal.spec.tsx:10` is `jest.mock("react-native-modal", () => require("../../helpers/react-native-modal-mock"))`. `tsconfig.jest.json` overrides `types` and does not include `"node"`, so `require` is untyped here and the returned module shape is `any`: a rename or default-export change in the helper produces no compile error, only a confusing runtime failure. `jest.requireActual("../../helpers/react-native-modal-mock")` is typed and is what the sibling specs use inside their factories.

**7. The audited amount is baked into the sentence across 29 files**

With `dailyLimit: "USD 999 daily transaction limit"` and `dailyLimitDescription: "Send up to USD 999 per day"` written as literals, the interpolation follow-up this PR describes has to touch all 28 translation files a second time to introduce a placeholder. Until then nothing ties `GetStartedScreen.trialAccountLimits.dailyLimit` to `OnboardingScreen.welcomeLevel1.dailyLimitDescription`, so the two screens can drift apart. Parameterizing now (`"USD {amount} daily transaction limit"` fed from a single constant) confines both this change and the follow-up to one file.

## LOW

**8. The helper uses a top-level `react-native` import where every copy it replaces uses `jest.requireActual`**

`__tests__/helpers/react-native-modal-mock.tsx:2` is a plain `import { View } from "react-native"`. A consuming spec that does `jest.mock("react-native", ...)` to stub `Platform` or `Dimensions` gets the mocked `View` inside the modal body, changing the rendered tree in ways unrelated to the component under test. The inline copies avoid this by resolving `react-native` with `jest.requireActual` inside the factory.

**9. The hand-rolled stub this module was created to eliminate is still in a sibling spec**

`__tests__/screens/account-type-selection/account-type-selection.spec.tsx:110-112` still does `jest.mock("@app/screens/phone-auth-screen", () => ({ PhoneLoginInitiateType: { Login: "Login" } }))`, the exact drift-prone pattern removed from the trial-limits spec. Swapping it to `jest.requireActual("@app/screens/phone-auth-screen/phone-login-initiate-type")` is a one-line change the new module already enables.

**10. The audit spec hardcodes two non-audited copy strings**

`trial-account-limits-modal.spec.tsx:59-60` asserts the literals `"Recover funds by SMS or email"` and `"Receive bitcoin onchain"`, neither of which carries an audit constraint. A copy tweak to `recoveryOption` fails the compliance spec and the failure reads like an audit regression. Only the audited `USD 999` literal needs pinning; the other two should go through `LL...recoveryOption()` since their purpose is presence, not exact wording.

**11. An in-repo doc still quotes the pre-fix copy**

`docs/dollar-vs-stablesats-naming-analysis.md:98-99` and `:104-105` quote "Send up to USD 1,000 per day" and "USD 1,000 daily transaction limit" as the app's copy. The verification step in the PR description greps `app/` only, so it misses these. Anyone grepping the repo root for the audited string after merge hits four live-looking matches and cannot tell they are a historical quote.

---

The wording of the daily-limit copy itself is a separate, optional suggestion left inline on `app/i18n/en/index.ts`.

esaugomez31:
The points from my previous review are now resolved.

app/i18n/en/index.ts
```diff
@@ -2098,7 +2098,7 @@ const en: BaseTranslation = {
       sendingLimit: "Reduced daily sending limit",
       noOnchain: "No receiving bitcoin onchain",
       recoveryOption: "Recover funds by SMS or email",
-      dailyLimit: "USD 1,000 daily transaction limit",
+      dailyLimit: "USD {amount: number} daily transaction limit",
```
esaugomez31: A suggestion on the wording of the daily-limit copy itself, kept out of the main review list.

`app/screens/settings-screen/transaction-limits-screen.tsx:99-109` enforces three separate daily limits (`withdrawal`, `internalSend`, `convert`), and the audited 999 is the withdrawal one. So I would rather these two lines said "withdraw" instead of "send":

**`GetStartedScreen.trialAccountLimits.dailyLimit`**
- current: `USD {amount: number} daily transaction limit`
- suggested: `USD {amount: number} daily withdrawal limit`

**`OnboardingScreen.welcomeLevel1.dailyLimitDescription`**
- current: `Send up to USD {amount: number} per day`
- suggested: `Withdraw up to USD {amount: number} per day`

Only the wording changes, the `amount` placeholder stays as it is.

This is only a suggestion and can be omitted. Your call @grimen.

esaugomez31 (dismissed):
LGTM

kngako (dismissed):
# Branch Review: `fix/level1-limit-display-999`

**Commits reviewed (8):** head `0b7b872d1`, merge-base `04c292c95`

1. `0c854e760` compliance(mobile): fix limits display 1000 -> 999
2. `1fabe3214` test(mobile): lock the audited USD 999 copy per screen
3. `a0a570494` test(mobile): assert the real navigation type in the trial-limits spec
4. `d7244ce8b` fix(i18n): translate the level 1 daily limit copy to Armenian
5. `5bff75eaf` refactor(mobile): import PhoneLoginInitiateType from its leaf module
6. `c08b0ac16` test(mobile): share one react-native-modal mock and drop the barrel stubs
7. `cf19c0b98` refactor(i18n): feed the level 1 daily limit amount from a single constant
8. `0b7b872d1` docs: note the audited amount supersedes the quoted copy

**What the branch does:** corrects the advertised level 1 daily limit from USD 1,000 to the
audited USD 999 (SSF audit, blink-wip#739) on the two surfaces that show it
(`welcome-level1-screen.tsx`, `trial-account-limits-modal.tsx`), feeds the amount from a
single constant (`LEVEL_ONE_DAILY_LIMIT_USD`, `app/config/appinfo.ts:36`) through a new
`{amount: number}` i18n placeholder across all 28 locales, pins the copy with literal
assertions in two specs, extracts `PhoneLoginInitiateType` into a leaf module so tests can
reach it without pulling in firebase app-check, and consolidates five inline
`react-native-modal` mocks into `__tests__/helpers/react-native-modal-mock.ts`.

**Verified during review:**

- All touched suites green: 18 suites / 148 tests pass (`NODE_ENV=test yarn jest` over the
  upgrade-account-modal, onboarding, account-type-selection, all five mock-refactored modal
  specs, and conversion-flow). Note for local repro: a shell with `NODE_ENV=development`
  exported (e.g. via direnv) makes RNGH's `isTestEnv()` false and fails every
  `CustomModal`-based suite with `NativeViewGestureHandler must be used as a descendant of
  GestureHandlerRootView` — that is an environment artifact, not a branch regression; it
  reproduces identically at the merge-base.
- `tsc -p .` clean; `yarn eslint` clean on every file the branch touches (the repo-wide
  `eslint:check` failure is pre-existing noise from `android/app/build/` artifacts and
  untouched files).
- `yarn typesafe-i18n --no-watch` produces zero diff — `app/i18n/i18n-types.ts` was properly
  regenerated, not hand-edited.
- No stray "1,000" copy remains outside `docs/` (where the stale quotes are explicitly
  annotated as superseded). The two changed i18n keys have exactly the two intended
  consumers each.
- `PhoneLoginInitiateType` extraction is safe: the barrel (`phone-auth-screen/index.ts`)
  still re-exports it, no consumer imports it from `phone-login-input` directly, and the
  account-type-selection spec's barrel→leaf mock swap keeps the real wire values.
- Revert test per house norm: changing the constant or the copy back to 1,000 turns both
  pinned-literal specs red. Good.

---

## Findings

### Finding 1 — hand-edited locale JSONs can be silently reverted by the next translation sync

The 27 files under `app/i18n/raw-i18n/translations/` are listed in `docs/pr-review.md` as
generated artifacts ("Only `app/i18n/en/index.ts` is hand-edited; `i18n-types.ts`,
`raw-i18n/`, and all locale folders are generated via `yarn update-translations`"), but this
branch hand-edits every one of them to inject `{amount: number}`. If the translation
management platform (the upstream source these files sync from) still holds the old
"1,000" strings, a future translation import overwrites these edits and re-ships the
non-compliant amount — the exact bug this branch exists to fix.

**When it comes up:** two sprints from now someone lands a routine translation refresh.
`es.json` is regenerated from the platform's stale entry and again reads "Envía hasta USD
1,000 por día". `locale-parity.spec.ts` still passes (it only checks key presence, not
placeholders), no spec renders a non-English locale, CI goes green, and Spanish-speaking
users see the pre-audit amount.

**Proposal:** update both keys (`GetStartedScreen.trialAccountLimits.dailyLimit`,
`OnboardingScreen.welcomeLevel1.dailyLimitDescription`) in the translation platform in the
same change window, and say so in the PR description so the reviewer can confirm. Add a
placeholder-parity assertion to `__tests__/i18n/locale-parity.spec.ts` that extracts the
`{...}` tokens per leaf key from `en` and each locale and fails on mismatch — a reverted or
placeholder-dropping locale then goes red immediately. If the raw JSONs are in fact edited
directly by translators with no upstream platform, downgrade this to a note, but state that
in the PR.

### Finding 2 — non-blocking: the `{amount: number}` formatter is a no-op

`app/i18n/formatters.ts` registers only a `sats` formatter; there is no `number` formatter,
so the typesafe-i18n runtime falls back to the raw value (`node_modules/typesafe-i18n/
runtime/cjs/runtime/src/core.cjs:8-12`, `?? value`). With 999 that renders correctly, which
is why everything is green — but the `: number` annotation implies locale-aware formatting
that never happens, in any of the 28 locales.

**When it comes up:** the level 1 limit is revised to, say, USD 1,500 next year. Someone
bumps `LEVEL_ONE_DAILY_LIMIT_USD`, both pinned specs are updated, CI is green — and every
locale renders "USD 1500" with no grouping separator, where German users should see
"1.500" and the prior English copy style was "1,000".

**Proposal:** register a real formatter in `initFormatters` —
`number: (value) => new Intl.NumberFormat(locale).format(Number(value))` — so the
annotation does what it reads as doing. Alternatively, if unformatted output is the
deliberate choice, drop the `: number` annotation to plain `{amount}` so the no-op is
explicit. Either is a two-line change; the first is the safer default for a compliance
amount.

### Finding 3 — non-blocking: the advertised amount is a client constant while enforcement is server-side

`LEVEL_ONE_DAILY_LIMIT_USD = 999` (`app/config/appinfo.ts:36`) hardcodes in the binary a
number the backend actually enforces. This PR fixes one instance of copy/enforcement drift,
but the structure that caused it is unchanged — the two values can drift again, in either
direction, and the new pinned tests stay green because they pin the constant, not the
backend value.

**When it comes up:** ops tightens the level 1 limit to USD 750 after a fraud review. The
backend ships the change; the app keeps advertising "Send up to USD 999 per day" until a
mobile release catches up — the same over-promise compliance problem, now with tests that
actively assert the stale number is correct.

**Proposal:** if the enforced limit is exposed over GraphQL (the settings/account-limits
payload already carries limit fields, cf. `app/graphql/mocks.ts` `totalLimit`), source the
advertised amount from there and keep the constant only as a fallback default. If it is not
exposed, add a one-line comment at the backend limit's definition site pointing at
`LEVEL_ONE_DAILY_LIMIT_USD`, so whoever changes enforcement sees the mobile copy coupling.
Keep the literal-pinned specs either way — they are the right tripwire for unintended copy
edits.

---

## Test Coverage Gaps

1. **No non-English rendering or placeholder-parity coverage.** Both pinned specs run only
   in `en`, and `locale-parity.spec.ts` compares key sets, not `{placeholder}` tokens. A
   locale that mistypes the placeholder (renders the literal text "{amount: number}" to
   users) or keeps a stale hardcoded "1,000" passes every gate. Fix per Finding 1:
   placeholder-token parity in the parity spec, plus one render assertion in a non-English
   locale (e.g. `es`) that the interpolated amount appears.

2. **No guard against a third surface reintroducing the amount.** The two current consumers
   are pinned, but nothing fails if a future screen hardcodes "1,000" (or "999") instead of
   using `LEVEL_ONE_DAILY_LIMIT_USD` — the drift class repeats per surface. A cheap
   mitigation is a unit test over `app/i18n/en/index.ts` asserting no limit-copy string
   contains a literal four-digit amount; at minimum, mention the constant in the PR so
   future authors know it exists.

3. **The shared modal mock's contract is untested.** Six suites now depend on
   `__tests__/helpers/react-native-modal-mock.ts` exporting a default component with
   `__esModule` interop; an edit that drops the default export breaks all six with an
   opaque "Element type is invalid" error far from the cause. A one-assertion smoke test of
   the helper (renders children when visible, null when not) would localize that failure.

## Code Smells

- **Path-depth-sensitive mock wiring**: `jest.requireActual("../../helpers/react-native-modal-mock")`
  silently breaks if a spec file moves a directory level. The repo already has a root
  `__mocks__/` convention — a root `__mocks__/react-native-modal.js` would let each spec
  keep a bare `jest.mock("react-native-modal")` with no relative path at all.

- **Two import paths for one symbol**: `PhoneLoginInitiateType` is now importable from both
  the barrel and the leaf module, with only the leaf's file comment steering the choice.
  A one-line comment on the barrel re-export (`phone-auth-screen/index.ts:1`) — "prefer the
  leaf import in new code; the barrel pulls in firebase app-check" — would keep the split
  intentional rather than accidental.

- **Off-convention commit type**: `0c854e760` uses `compliance(mobile):`; the house
  convention (`docs/pr-review.md`) maps scopes to feature domains and expects standard
  types (`fix(i18n):` would fit). Squash-merge makes the PR title the real commit, so this
  is cosmetic — but it will mislead anyone filtering history by type.

- **Over-pinned literals (noted, accepted)**: the `DAILY_LIMIT` / "Send up to USD 999 per
  day" literals fail on *any* wording change to those sentences, not just the amount. The
  spec comments say this is deliberate; flagging only so the next copy edit updates both
  files in the same commit rather than treating the failures as noise.

---

## Verdict

The code change itself is correct, minimal, well-tested in English, and the i18n machinery
is properly regenerated — no functional defects found. **One blocking item: Finding 1** —
confirm the translation source of truth carries the new placeholder strings (or that no
such upstream exists), because a silent sync-revert would undo a compliance fix under a
green pipeline. Findings 2 and 3 and the coverage gaps are follow-up material and should
not hold the fix.

grimen:
Thanks — all 11 are addressed, and the branch changed direction while doing it. Summary of where each landed, at d9a6524db.

## Direction change

The interim constant is gone. Finding 7 asked for parameterisation fed from a single source; rather than a client constant, the copy now reads the enforced value from `globals.accountLimitsByLevel` (your blink#750) through a new `useLevel1DailyLimit` hook, with the audited USD 999 as fallback. That also closes the drift class the constant only narrowed.

> [!WARNING]
> **Check Code will fail until `Globals.accountLimitsByLevel` is deployed to staging. This is expected and is not fixable from this diff.**
>
> The field is merged in blink#750 but not deployed — introspecting both `api.staging.blink.sv` and `api.blink.sv` today returns a `Globals` type without it. `check:codegen` and `graphql-check` both validate against the staging schema, so codegen rejects the document:
>
> ```
> [FAILED] GraphQL Document Validation failed with 1 errors;
> [FAILED]   Error 0: Cannot query field "accountLimitsByLevel" on type "Globals".
> [FAILED]     at app/hooks/use-level1-daily-limit.ts:4:7
> ```
>
> That is the only failure — `tsc`, `eslint`, `check:translations` and the i18n regeneration all pass ahead of it, and Android unit tests are green. **The check clears itself once staging picks the field up; please do not merge before then.**
>
> Runtime degrades safely in the meantime: the query errors and the fallback renders the audited USD 999, which is what the copy said before this PR.

Getting there needed a reconcile: this branch had been rebased off the published history, so I merged the published commits back in rather than force-pushing over them. Your commits are intact in the history.

## Findings

| # | Where it landed |
|---|---|
| 1 | Armenian translated for both keys, now with the `{limit}` placeholder. |
| 2 | Fixed and consolidated further — see below. |
| 3 | `stack-param-lists.ts` and `trial-account-limits-modal.tsx` import the leaf, so the cycle is broken and the spec's firebase workaround is gone. |
| 4 | The assertion is back, pinning the literal `"CreateAccount"` instead of reading `PhoneLoginInitiateType` — you were right that reading the constant moves both sides together. |
| 5 | The barrel mock is gone entirely, as you predicted once finding 3 was fixed. |
| 6 | `jest.requireActual`, typed. |
| 7 | `{limit}` placeholder in all 30 locale files, fed from the hook. |
| 8 | The stand-in resolves `react-native` through `jest.requireActual` instead of a top-level import. |
| 9 | `account-type-selection.spec` uses `jest.requireActual` of the leaf. |
| 10 | `recoveryOption` and `onchainReceive` go through `LL`; only the audited amount is pinned as a literal. |
| 11 | Both doc quotes annotated as superseded by the audited amount. |

### On finding 2

You were right on both halves. The stand-in dropped `onBackdropPress` and `onBackButtonPress`; it now exposes them as the `backdrop` / `back-button` pressables the modal specs already query, and its contract is covered directly by `__tests__/mocks/react-native-modal-mock.spec.tsx` so a broken edit fails on one file instead of nine.

On consolidation: it lives at `__mocks__/react-native-modal-mock.tsx` behind the `@mocks/` alias `jest.config.js` already mapped and nothing used, which removes the relative-path fragility. Nine suites share it now, including `custom-modal.spec`, whose inline copy was the one exercising the handlers. **The remaining inline mocks elsewhere in the repo are deliberately left alone** — mechanical, but 21 unrelated files.

## Also changed since your review

- **Placeholder parity in `locale-parity.spec.ts`.** Key parity stays green when a locale keeps the key and loses the interpolated value, which is exactly how a translation-platform sync would silently revert this fix. 47 pre-existing mismatches across 23 locales are recorded in a documented shrink-only allowlist, so this lands green and still fails all new drift.
- **`limit-copy.spec.ts`** fails any copy, in English or any locale, stating a daily allowance as a literal amount — guarding the drift class rather than these two keys.
- **Number grouping follows the reader's language.** The hook was formatting with a hardcoded `en-US`. Invisible at 999, but the backend can raise the limit without a mobile release, and at four digits German writes `1.500` where English writes `1,500` — a comma being German's decimal separator, that reads as one and a half. Covered by rendering the modal in `es` and `de`.

Both new i18n guards are mutation-verified: reverting `es.json` to the pre-fix English string fails both; dropping only the placeholder fails parity alone.

## Still open

Someone with access needs to confirm the translation platform holds the new `{limit}` strings, so a routine sync does not revert them upstream. The parity and literal-amount guards catch such a revert in CI, but the upstream side is a human check.

kngako:
the renderHook wrapper now nests MockedProvider inside a TypesafeI18n locale="en" provider (with loadLocale("en") at the top), so the hook formats against a pinned locale  instead of the host machine's ICU default. A comment records why, so nobody "simplifies" it back out.  Without this change on my machine (LANG=en_ZA.UTF-8) the suite fails: it expects "1,500" and "999.5" but gets "1 500" and "999,5".

kngako:
`useLevel1DailyLimit` now passes `fetchPolicy: "cache-and-network"`, with a new spec test ("refreshes a stale cached value from the network") that pre-seeds a persisted-style cache with 99900, asserts it renders immediately, and asserts the network response corrects it to 1,500. Revert-checked: the test fails if the fetch policy is removed.

kngako (dismissed):
We might need to do a follow up PR checking the `cache-first` fetch policy on configs as users will keep getting persisted/cached values.

esaugomez31 (approved):

PR #3993, opened 2026-07-24
feat: API key generation and sharing from the app (read/receive scopes)

grimen:
Fixes blinkbitcoin/blink-wip#366

## Summary

Custodial users can now create, list and revoke API keys directly in **Settings → Advanced → API integration**, limited to **READ** and **RECEIVE** scopes (WRITE stays dashboard-only). After creation the one-time secret is shown exactly once with tap-to-copy (clipboard auto-clears after 60 s), a QR code and the native share sheet — ready to paste into BTCPay Server or any watch-only integration. Multiple named keys per account are supported (one per integration), which is why in-app listing and revocation are included.

## How it works

- **No backend changes** — uses the existing `apiKeyCreate` / `apiKeyRevoke` mutations and `me.apiKeys` query (same API the dashboard uses); documents live in `app/screens/settings-screen/api/use-api-keys.ts`, mirroring the dashboard's single-file layout.
- **Existing API screen extended**: an "API keys" group (active keys sorted first with scopes + expiry subtitle, revoked/expired dimmed, empty state, "Create API key" row) sits above the unchanged docs/dashboard links.
- **New `apiKeyCreateScreen`** hosts the create form (name; Read/Receive switches — both default on, ≥1 required; expiry presets 30/90 days/Never, default 90) and the one-time secret reveal as two internal states, so **the secret only ever lives in component state** — never in navigation params, which React Navigation persists.
- **Secret reveal hardening**: `useScreenSecurity()` blocks screenshots/recording (see the black capture below); the header back button, swipe-back **and the Android system back button** (via a `beforeRemove` guard) are all blocked so the one-time secret can't be lost accidentally — Done is the only way out; and the clipboard self-clears 60 s after copy or immediately on leaving the screen, whichever comes first, and only when it still holds the key (a newer user copy is left untouched).
- **Cache freshness**: both mutations use `refetchQueries: [ApiKeysDocument]`. Revoke originally relied on Apollo normalization of the returned `ApiKey`, but live staging QA showed the row didn't flip to revoked until an app restart, so revoke refetches too (matching the dashboard, which also refetches after revoke). The list query itself uses `cache-and-network`, so keys revoked or expired outside the app show up on revisit without hiding cached keys behind a loading state.
- WRITE is never offered; a footnote points payment-capable keys to dashboard.blink.sv. The Advanced group is already hidden in self-custodial mode, so the feature is custodial-only by existing gating — intentional, per product: API keys are custodial-account credentials (the migration flow already warns about active keys via `use-active-api-keys.ts`).

## Test plan

- New `__tests__/screens/api-screen.spec.tsx`: active-key subtitle (scopes + expiry/never), revoked/expired rendering without revoke control, empty state, Alert-confirmed revoke calling the mutation with the right id.
- New `__tests__/screens/api-key-create.spec.tsx`: defaults (both scopes on, 90 days, disabled submit on empty name), both-scopes-off blocks submit, mutation variables for scope/expiry combinations, reveal phase (secret + warning shown, copy hits clipboard, share hits the share sheet, screen security enabled).
- New `__tests__/screens/api-key-secret-reveal.spec.tsx`: back-guard blocks removal, Done allows it, header options set; `__tests__/hooks/use-api-keys.spec.ts`: fetch policy, active-first/newest-first sorting, status precedence, scope labels (unknown scopes skipped); `__tests__/hooks/use-clipboard.spec.ts` updated to the clear-on-unmount / only-if-still-held semantics. Error paths covered: create mutation failure and missing-secret response, revoke success/error toasts, list load-error box.
- `yarn tsc:check`, eslint and all targeted suites pass locally; full suite runs in CI.
- **Live staging QA** on an isolated simulator: created keys end-to-end (list refresh, server-computed expiry), revoked them (immediate UI flip + success toast), verified screenshot protection blanks captures of the secret. Both demo keys shown below were revoked afterwards — no live secret appears in any image.

## Screenshots

| Before — this screen today | After — with keys | After — empty state |
| --- | --- | --- |
| API integration screen with only documentation and dashboard links | API keys group listing an active key with scopes and expiry plus a revoke control | API keys group with empty state and Create API key row |

| Create form | Revoke confirmation | Revoked state + toast |
| --- | --- | --- |
| New API key form with name input, Read and Receive toggles and expiry presets | Native alert confirming revocation of a named key | List showing dimmed revoked keys and a success toast |

**Secret reveal — screenshot protection working as designed:** capturing the reveal screen produces a black frame, so the one-time secret cannot leak via screenshots or screen recording. (Reviewers can see the actual layout in `api-key-secret-reveal.tsx`: success icon, warning copy, QR, monospace tap-to-copy row, share button, Done.)

Entirely black capture proving the secret reveal screen blocks screenshots

_Images hosted on the `assets/pr-3993-screenshots` branch (orphan commit, safe to delete after merge)._

## Open questions / follow-ups (out of scope)

- **Level-0 accounts**: whether `apiKeyCreate` requires level ≥ 1 isn't determinable from this repo; QA ran on a leveled test account. If the backend rejects level 0, a `useLevel()` gate on the create row is a one-line follow-up.
- **Latent bug (pre-existing, untouched)**: the "API documentation" row opens `dashboard.blink.sv` instead of the docs site (`api-screen.tsx`).


esaugomez31 (approved):
Works as expected

esaugomez31 (approved):

PR #3994, opened 2026-07-24
fix(deps): sync shell-quote lockfile entry key with resolutions

grimen:
Concourse `check-code`/`test-unit` fail on main with **"Extracting deps has created a diff - deps are not in sync"** since #3936 merged: that PR carried Dependabot's lockfile hunk, which kept the stale `shell-quote@^1.8.4` range key while package.json's resolutions moved to `^1.10.0`. A fresh `yarn install` rewrites the key line, so the pipeline's regenerate-and-diff check trips.

One-line fix: the regenerated key (`shell-quote@^1.10.0, ^1.6.1, ^1.7.2, ^1.7.3`). No version changes.


grimen:
No longer needed — origin/main's yarn.lock already has the shell-quote key ordering this PR produces (landed via a later Dependabot regen).

PR #3999, opened 2026-07-24
fix(send): display canonical lightning address for Blink LNURL destinations

grimen:
Fixes blinkbitcoin/blink-wip#1034

## Problem

Scanning a custodial receive-flow paycode QR with a **self-custodial (Spark)** wallet shows the destination as `user@pay.blink.sv` instead of the canonical `user@blink.sv` shown on the receive screen.

The receive paycode QR encodes a bech32 LNURL of `https://pay.blink.sv/.well-known/lnurlp/<username>`, and the pay-server's LNURL metadata advertises the identifier under its own hostname (`user@pay.blink.sv`). The custodial send path is unaffected (its scan context collapses Blink LNURLs to intraledger via `lnurlDomains`), but the self-custodial path intentionally has `lnurlDomains: []` — it pays Blink users over real Lightning — so `createSelfCustodialLnurlPaymentDetails` displayed the raw identifier verbatim. The same string was copied to clipboard and persisted as the auto-saved contact handle.

## Fix

- New pure helper `canonicalizeLightningAddress` in `app/utils/pay-links.ts`: rewrites addresses whose domain is a Blink alias (the existing `LNURL_DOMAINS` list: `blink.sv`, `pay.blink.sv`, `pay.bbw.sv`) to `user@blink.sv` (`BLINK_DOMAIN`). Non-Blink addresses, bech32 `lnurl1...` strings, and malformed input pass through unchanged; the rewrite is idempotent.
- Applied at the single production call site: the `destination` assignment in `app/self-custodial/payment-details/lnurl.ts`. This also fixes copy-to-clipboard and the saved contact handle, which read `paymentDetail.destination`.

Deliberately untouched: `lnurlParamsToPayRequest` (the Breez SDK payload keeps the server-provided identifier), the shared custodial `resolveLnurlDestination`/`createLnurlPaymentDetails`, and the scan contexts. Populating `lnurlDomains` for self-custodial instead was considered and rejected: the intraledger collapse would produce a custodial GraphQL destination a Spark wallet cannot pay.

## Tests

- `__tests__/self-custodial/payment-details/lnurl.spec.ts`: regression cases for `pay.blink.sv` → `blink.sv`, legacy `pay.bbw.sv` → `blink.sv`, and canonical passthrough (existing non-Blink and bech32-fallback cases unchanged).
- `__tests__/utils/pay-links.spec.ts`: helper coverage — alias mappings, case-insensitive domain match, non-Blink/staging/bech32/empty/multi-`@`/empty-username/empty-domain passthrough.
- A final commit closes pre-existing coverage gaps in the touched files (the three setters that were only asserted as defined, the message/url success-action `decipher` stubs, and the empty-domain / zero-amount fallbacks).

Both changed production files are now at **100% statements, branches, functions and lines**:

```
File       | % Stmts | % Branch | % Funcs | % Lines
lnurl.ts   |     100 |      100 |     100 |     100
pay-links.ts|    100 |      100 |     100 |     100
```

`tsc --noEmit`, `eslint`, and `prettier --check` all clean.

## Out of scope / follow-up

The root advertisement lives in the pay-server's LNURL metadata (`text/identifier`); fixing it there would correct every third-party wallet, not just Blink mobile. Left as a potential backend follow-up.

### DIFF
--- a/__tests__/self-custodial/payment-details/lnurl.spec.ts
+++ b/__tests__/self-custodial/payment-details/lnurl.spec.ts
@@ -173,6 +173,44 @@ describe("createSelfCustodialLnurlPaymentDetails", () => {
     expect(detail.setSuccessAction).toBeDefined()
   })
 
+  describe("setters recreate the detail", () => {
+    it("setConvertMoneyAmount swaps the conversion function", () => {
+      const detail = createSelfCustodialLnurlPaymentDetails(createParams())
+      const newConvert = jest.fn((amount, target) => ({
+        amount: amount.amount,
+        currency: target,
+        currencyCode: target,
+      }))
+      const updated = detail.setConvertMoneyAmount(newConvert)
+      expect(updated.convertMoneyAmount).toBe(newConvert)
+    })
+
+    it("setInvoice recreates an equivalent lnurl detail", () => {
+      const detail = createSelfCustodialLnurlPaymentDetails(createParams())
+      if (detail.paymentType !== PaymentType.Lnurl) throw new Error("expected lnurl")
+      const updated = detail.setInvoice({} as never)
+      expect(updated.paymentType).toBe(PaymentType.Lnurl)
+      expect(updated.destination).toBe(detail.destination)
+    })
+
+    it("setSuccessAction carries the success action onto the new detail", () => {
+      const detail = createSelfCustodialLnurlPaymentDetails(createParams())
+      if (detail.paymentType !== PaymentType.Lnurl) throw new Error("expected lnurl")
+      const successAction = {
+        tag: "message" as const,
+        message: "Thanks!",
+        description: null,
+        url: null,
+        ciphertext: null,
+        iv: null,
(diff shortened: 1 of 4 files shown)


dolcalmi:
this is a blink client responsability

grimen:
Closing in favour of an upstream fix: blinkbitcoin/blink-client#51.

This PR canonicalised the domain on the mobile side, after the LNURL endpoint had already been resolved over HTTP. That treats the symptom — the pay server advertises its own host in the LNURL metadata `text/identifier`, and the app was displaying it verbatim.

The better place is parsing. A Blink pay code QR is the bech32 of `https://pay.blink.sv/.well-known/lnurlp/<username>`, so the payload already identifies the account and no network call is needed to recover it. blink-client now resolves it through the same path as typing that account's lightning address, so both inputs produce the same destination and the pay host never reaches the UI.

See blinkbitcoin/blink-wip#1034 for the follow-up needed on this side once that lands.

PR #4007, opened 2026-07-24
fix(send): treat user@domain input as lightning address, not phone number

### LINKED ISSUE #4021: fix: Lightning addresses with numeric usernames rejected as phone numbers
## Description

When entering a Lightning address whose username is purely numeric (e.g. `2290161757483@bitcoinflash.xyz`), the "Send to" field rejects it with the error *"This field does not accept phone numbers."*

The input validation appears to match the numeric username against phone number patterns (229 is Benin's country code) and incorrectly classifies the entire Lightning address as a phone number, even though it contains `@` and a domain — which makes it unambiguously a Lightning address.

## Steps to reproduce

1. Open Blink mobile
2. Go to Send → "Send to" field
3. Enter `2290161757483@bitcoinflash.xyz`
4. Observe error: "This field does not accept phone numbers"

## Expected behavior

The input should be recognized as a valid Lightning address and proceed to the payment flow.

## Actual behavior

The input is rejected with a phone number validation error.

## Suggested fix

Lightning address validation (`*@*.*`) should take priority over phone number detection. If the input contains `@` followed by a valid domain, it should never be classified as a phone number.

## Context

Reported by Flash (bitcoinflash.xyz), a mobile money ↔ Lightning exchange from Benin. Their Lightning addresses use phone numbers as usernames, which is a common pattern for mobile money integrations. This blocks interoperability with their service.

Screenshot

---
Filed by Blink-Claw-Bot ⚡ on behalf of @pretyflaco (kemal)

grimen:
## Summary

Fixes the Send-to field rejecting valid Lightning Addresses like `0777491011@bitzed.xyz` with *"This field does not accept phone numbers"*.

Fixes blinkbitcoin/blink-wip#917
Fixes #4021

## Root cause

The bug is in the mobile app, not in blink-client — both blink-client 0.6.0 (currently shipped) and 1.1.0 parse the reported addresses as valid LNURL, so blink-client#43 (closed) and blink-client#46 did not address this.

`validateDestination` runs the raw input through `parseValidPhoneNumber` (`app/utils/phone.ts`) before destination parsing, and `libphonenumber-js` **extracts** a phone number out of surrounding text by default. It strips `@bitzed.xyz` — and even letters — so:

- `0777491011@bitzed.xyz` → `+260777491011`, a valid ZM phone → rejected
- `u66474248@rurbit.mooo.com` → `66474248` → `+50366474248`, a valid SV phone → rejected (the second report in the issue)

This also explains why the bug seemed version-specific in the issue thread: the check depends on the user's detected country. For a country where the extracted digits aren't a valid phone number, the same address goes through fine.

## Also fixes #4021 (Benin / Flash)

#4021 is the same defect reported from a different country. Flash (bitcoinflash.xyz), a mobile money ↔ Lightning exchange in Benin, uses phone numbers as Lightning address usernames, so `2290161757483@bitcoinflash.xyz` was rejected. Verified against the installed `libphonenumber-js/mobile`:

| input | detected country | before | after |
| --- | --- | --- | --- |
| `2290161757483@bitcoinflash.xyz` | **BJ** | `+2290161757483` → rejected | `null` → reaches the LNURL parser |
| `2290161757483@bitcoinflash.xyz` | SV / US / ZM / none | `null` | `null` |
| `2290161757483` (bare number) | BJ | `+2290161757483` | `+2290161757483` (unchanged) |

BJ is the only detected country for which that local part parses as a valid number, which is why the issue could not be reproduced from other countries.

## Fix

Pass `extract: false` to `parsePhoneNumber` in `parseValidPhoneNumber` and in the fallback parse of `isPhoneNumber` (same extraction bug, used for contact filtering). Actual phone input still parses — spaces, dashes, parentheses and `+` prefixes are unaffected — but text containing anything else (like `@domain`) is no longer treated as a phone number. Since all input paths (typed, QR scan, deeplink) funnel through `parseValidPhoneNumber`, one fix covers them all.

A second commit fixes a path the first change would have regressed: tapping a **phone contact** relied on libphonenumber extracting the number back out of the `handle@domain` display string. `handleContactPress` now parses the raw handle instead — phone contacts still enter the phone flow, while contacts whose handle is already a lightning address never do, even with a phone-like local part.

## Tests

- `__tests__/utils/phone.spec.ts`: LN addresses with phone-valid local parts return null (`0777491011@bitzed.xyz` ZM, `70000000@bitzed.xyz` SV, `+256777491011@bitzed.xyz`, `+50370000000@blink.sv`, `2290161757483@bitcoinflash.xyz` BJ for #4021); alphanumeric local part (`u66474248@rurbit.mooo.com`); formatting regression guards (spaced/dashed/parenthesized numbers still parse, and the bare `2290161757483` BJ number still parses); `isPhoneNumber` returns false for phone-valid local parts.
- `__tests__/screens/send-destination.spec.tsx`: screen-level regression tests — `70000000@bitzed.xyz` (valid phone local part for the suite's mocked SV country) and `u66474248@rurbit.mooo.com` must not show the phone-not-allowed error and must reach `parseDestination`.
- Coverage for every `parseValidPhone` call site in the screen:
  - typed search input: the two cases above
  - route-param handoff (QR scan/deeplink): a lightning address is validated as a destination; a real phone number still routes to the phone input
  - contact taps: a phone contact enters the phone flow (validated as its international number); a lightning-address contact with a phone-like local part is validated as the full address

All new tests were written first and observed failing against the unfixed code (the QR-path, contact-path and BJ tests verified by temporarily reverting `app/utils/phone.ts` to origin/main). Full suites green locally: `phone.spec.ts` + `send-destination.spec.tsx` 51/51. ESLint + `tsc --noEmit` clean.

## Out of scope

- The "Try Again" crash loop reported in the blink-wip#917 comments is a separate error-recovery bug and should be split into its own issue.
- The `isInt(rawInput)` half of the Search-input guard is unchanged: a bare numeric username with no domain is still rejected, which is intentional.

### DIFF
--- a/__tests__/screens/send-destination.spec.tsx
+++ b/__tests__/screens/send-destination.spec.tsx
@@ -501,6 +501,162 @@ describe("SendBitcoinDestinationScreen", () => {
     await flushEffects()
   })
 
+  it("routes a tapped phone-number contact into the phone flow", async () => {
+    mockedDestinationData = {
+      ...mockedDestinationData,
+      me: {
+        ...mockedDestinationData.me,
+        contacts: [
+          {
+            id: "contact-id",
+            handle: "+50370000000",
+            username: "+50370000000",
+            alias: null,
+            transactionsCount: 1,
+          },
+        ],
+      },
+    }
+    parseDestinationMock.mockResolvedValue({
+      valid: true,
+      destinationDirection: DestinationDirection.Send,
+      validDestination: {
+        valid: true,
+        paymentType: PaymentType.Lnurl,
+        lnurl: "lnurl",
+        isMerchant: false,
+        lnurlParams: createLnurlPayParams("+50370000000"),
+      },
+      createPaymentDetail: jest.fn(),
+    })
+
+    render(
+      <ContextForScreen>
+        <SendBitcoinDestinationScreen route={sendBitcoinDestination} />
+      </ContextForScreen>,
+    )
+    await flushAsync()
+
+    fireEvent.press(screen.getByText("+50370000000@blink.sv"))
+    await flushAsync()
+
+    // the phone contact is validated as its international phone number,
+    // not as handle@domain
+    expect(parseDestinationMock).toHaveBeenCalledWith(
+      expect.objectContaining({ rawInput: "+50370000000" }),
+    )
+  })
+
+  it("treats a tapped lightning-address contact as a lightning address", async () => {
+    mockedDestinationData = {
+      ...mockedDestinationData,
+      me: {
+        ...mockedDestinationData.me,
+        contacts: [
+          {
+            id: "contact-id",
+            handle: "70000000@bitzed.xyz",
+            username: null,
+            alias: null,
+            transactionsCount: 1,
+          },
+        ],
+      },
+    }
+    parseDestinationMock.mockResolvedValue({
+      valid: true,
+      destinationDirection: DestinationDirection.Send,
+      validDestination: {
+        valid: true,
+        paymentType: PaymentType.Lnurl,
+        lnurl: "lnurl",
+        isMerchant: false,
+        lnurlParams: createLnurlPayParams("70000000@bitzed.xyz"),
+      },
+      createPaymentDetail: jest.fn(),
+    })
+
+    render(
+      <ContextForScreen>
+        <SendBitcoinDestinationScreen route={sendBitcoinDestination} />
+      </ContextForScreen>,
+    )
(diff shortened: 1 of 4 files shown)


kngako:
When a phone input is typed in the invoice field we correctly get the error

image


When a phone number is part of a lightning address everything works as expected.

image

image

kngako (dismissed):

esaugomez31 (approved):
LGTM

Issue #3815, opened 2026-06-03
fix(test): Jest tests time out randomly across the suite, causing flaky CI and repeated re-runs


grimen:
## Description

Jest tests **time out randomly** across the suite in CI. The failures are **non-deterministic** — different suites/tests fail on different runs, with no single reliable repro. The default 5000 ms per-test timeout is exceeded intermittently, the run exits non-zero, and the **only "fix" today is re-running CI over and over** until it happens to pass. This is a major source of wasted time and friction.

This run happened to catch it in `__tests__/screens/settings-screen/settings-screen.spec.tsx`, but **that file is not the root cause** — it's just where the dice landed this time. The underlying issue is flaky timeouts that can surface anywhere.

## Symptoms

- A test exceeds the default timeout with the generic Jest message:

```
thrown: "Exceeded timeout of 5000 ms for a test.
Add a timeout value to this test to increase the timeout, if this is a long-running test.
See https://jestjs.io/docs/api#testname-fn-timeout."
```

- The timeout often **cascades** to the *next* test in the same suite, which then fails at its `render(...)` call with:

```
Can't access .root on unmounted test renderer
```

because the timed-out test never cleaned up / unmounted, leaving React Testing Library's renderer in a bad state.

- Jest sometimes warns about leaked async work:

```
Force exiting Jest: Have you considered using `--detectOpenHandles`
to detect async operations that kept running after all tests finished?
```

## Example from a recent run

`__tests__/screens/settings-screen/settings-screen.spec.tsx` — `2 failed`:

```
● Settings Screen › clears the badge after entering the notification history
    thrown: "Exceeded timeout of 5000 ms for a test. ..."
    > 287 |   it("clears the badge after entering the notification history", async () => {

● Settings Screen › hides the badge when the last unread notification is acknowledged
    Can't access .root on unmounted test renderer
    > 344 |     render(
    at react-test-renderer ... at @testing-library/react-native/src/render.tsx
```

```
Test Suites: 1 failed, 400 passed, 401 total
Tests:       2 failed, 4524 passed, 4526 total
Time:        710.862 s
error Command failed with exit code 1.
```

On another run these two would pass and a different test elsewhere would time out instead.

## Impact

- **CI is flaky / red intermittently** — runs exit non-zero (`error Command failed with exit code 1`) with no real regression.
- **Wastes significant time**: the standard workaround is re-triggering CI repeatedly until it goes green, which is slow and demoralizing and burns CI capacity.
[...]

PR #4008, opened 2026-07-24
test: eliminate act() warnings and console.warn noise from jest logs

grimen:
## Summary

Jest output (local and CI) was bloated by two kinds of repeated noise:

1. **`act()` warnings** — most of them trace to `AccountRegistryProvider`'s two async device reads (`listSelfCustodialAccounts()`, `KeyStoreWrapper.getSessionProfiles()`). Their promises settle after the test body finishes, so the resulting `setState`s fire outside `act()`. The affected set is timing-dependent (16 suites in one run, a different set in the next) across the ~110 specs that mount `ContextForScreen`, so per-spec `act()` patching would be whack-a-mole.
2. **`console.warn` noise** — `[ip-country-lookup] No API key configured` at module load in ~27 suites, plus the `InteractionManager`/`SafeAreaView` deprecations that were already suppressed in CI but still printed locally.

Beyond silencing them, the branch adds a guard that **fails** a test when a state update escapes `act()`, so the suite cannot quietly drift back into noise.

## Changes

- `app/hooks/use-account-registry.tsx`: split into "read the device" and "compose the value". `useComposedAccountRegistry` is exported alongside `AccountRegistryContext`, so tests can supply a settled registry without any test-only branch in shipped code. The provider's own hydration is unchanged from `main`.
- `__tests__/screens/helper.tsx`: both `ContextForScreen` variants mount a settled stub registry built from that same hook, with an `accountRegistry` prop for specs that need seeded self-custodial accounts. Specs that exercise real hydration (`use-account-registry.spec`, `wallet.spec`) mount `AccountRegistryProvider` directly. `__tests__/screens/helper-account-registry.spec.tsx` pins the default and the seeded path, so the stub cannot silently diverge from the provider.
- `jest.setup-after-env.ts`: the act() guard. `console.error` is locked against replacement — a spec that switched the guard off with `jest.spyOn` used to disable escaped-update detection for its whole file, silently. `silenceConsoleError()` (`__tests__/helpers/silence-console-error.ts`) mutes and captures output through a sink underneath the guard instead. Each escaped update records the test that emitted it, so an unmount-time leak from test A no longer fails test B under a component name from A.
- `jest.setup.js`: warn suppression is now unconditional (local noise was just as distracting as CI noise) and is limited to the two third-party deprecations. The file documents the rule: this list is for third-party noise only.
- `app/utils/ip-country-lookup.ts`: the no-API-key warning moved off module load and onto the first actual lookup, so it no longer fires in every suite that transitively imports the module — and no longer needs suppressing.
- `__tests__/helpers/flush-effects.ts`: captures the real `setImmediate` at import time, so the helper does not hang in fake-timer suites.
- ~40 specs migrate from `@testing-library/react-hooks` to `@testing-library/react-native`; the dependency and its orphaned lockfile entry are dropped.

## Test plan

- [x] Full parallel suite locally: 523 suites / 5834 tests passing
- [x] Zero `act()` warnings, zero `console.warn`/`console.error` in the full-suite log
- [x] Real-hydration specs (`use-account-registry`, `wallet`) still pass
- [x] `ip-country-lookup.spec.ts` covers quiet-at-load, warn-on-first-lookup, and warn-only-once
- [x] Guard verified to bite: a deliberate escaped update inside a spec that mutes `console.error` fails, naming the right test
- [x] `eslint` and `tsc` clean

### DIFF
--- a/__tests__/components/push-notification.spec.tsx
+++ b/__tests__/components/push-notification.spec.tsx
@@ -6,6 +6,7 @@ import { PushNotificationComponent } from "@app/components/push-notification/pus
 import { BulletinsDocument } from "@app/graphql/generated"
 
 import { flushEffects } from "../helpers/flush-effects"
+import { silenceConsoleError } from "../helpers/silence-console-error"
 
 const mockRefetchQueries = jest.fn()
 jest.mock("@apollo/client", () => ({
@@ -149,9 +150,7 @@ describe("PushNotificationComponent", () => {
     })
 
     it("logs instead of throwing when opening the link fails", () => {
-      const consoleErrorSpy = jest
-        .spyOn(console, "error")
-        .mockImplementation(() => undefined)
+      const consoleErrorSpy = silenceConsoleError()
       openURLSpy.mockImplementationOnce(() => {
         throw new Error("no handler for scheme")
       })
--- a/__tests__/components/revealed-checkbox-list/revealed-checkbox-list.spec.tsx
+++ b/__tests__/components/revealed-checkbox-list/revealed-checkbox-list.spec.tsx
@@ -3,6 +3,7 @@ import { fireEvent, render, screen } from "@testing-library/react-native"
 import { ThemeProvider } from "@rn-vui/themed"
 
 import { RevealedCheckboxList } from "@app/components/revealed-checkbox-list"
+import { silenceConsoleError } from "../../helpers/silence-console-error"
 
 const labels = ["First", "Second", "Third"]
 
@@ -67,7 +68,7 @@ describe("RevealedCheckboxList", () => {
   })
 
   it("gives duplicate labels stable, independent keys (no key collision)", () => {
-    const errorSpy = jest.spyOn(console, "error").mockImplementation(() => {})
+    const errorSpy = silenceConsoleError()
     const onAllCheckedChange = jest.fn()
     render(
       <ThemeProvider>
--- a/__tests__/config/galoy-instances.spec.ts
+++ b/__tests__/config/galoy-instances.spec.ts
@@ -1,4 +1,5 @@
 import { resolveGaloyInstanceOrDefault, GALOY_INSTANCES } from "@app/config"
+import { silenceConsoleError } from "../helpers/silence-console-error"
 
 it("get a full object with BBW", () => {
   const res = resolveGaloyInstanceOrDefault({ id: "Main" })
@@ -19,7 +20,7 @@ it("get a full object with Local", () => {
 })
 
 it("falls back to the Main instance for an unknown standard id", () => {
-  const consoleErrorSpy = jest.spyOn(console, "error").mockImplementation(() => {})
+  const consoleErrorSpy = silenceConsoleError()
 
   const res = resolveGaloyInstanceOrDefault({ id: "Nonexistent" } as never)
 
--- a/__tests__/graphql/network-error-component.spec.tsx
+++ b/__tests__/graphql/network-error-component.spec.tsx
@@ -11,6 +11,7 @@ import KeyStoreWrapper from "@app/utils/storage/secureStorage"
 import { useNetworkError } from "@app/graphql/network-error-context"
 import { NetworkErrorCode } from "@app/graphql/error-code"
 import { NetworkErrorComponent } from "@app/graphql/network-error-component"
+import { silenceConsoleError } from "../helpers/silence-console-error"
 
 jest.mock("@app/graphql/network-error-context")
 jest.mock("@app/i18n/i18n-react")
@@ -265,7 +266,7 @@ describe("NetworkErrorComponent", () => {
     // The component logs the simulated storage failure via console.error;
     // capture it so the expected error doesn't pollute CI logs (and assert it
     // actually happened).
-    const consoleErrorSpy = jest.spyOn(console, "error").mockImplementation(() => {})
+    const consoleErrorSpy = silenceConsoleError()
     ;(KeyStoreWrapper.getSessionProfiles as jest.Mock).mockRejectedValue(
       new Error("Storage error"),
     )
--- a/__tests__/helpers/flush-effects.ts
+++ b/__tests__/helpers/flush-effects.ts
@@ -20,10 +20,18 @@ import { act } from "@testing-library/react-native"
  * Tests that already await `waitFor`/`findBy*` on the resolved state do not need
  * this — the awaited query already settles the effects inside `act()`.
  */
+// Captured at import time, before any spec installs fake timers. Under
+// `jest.useFakeTimers()` the global `setImmediate` is replaced by one that only
+// fires when the spec advances the clock, so a flush scheduled on it would
+// never resolve and the test would time out. The real one yields to the event
+// loop once, which is all this needs: it settles already-pending promises
+// without disturbing the fake clock the spec is driving.
+const realSetImmediate = globalThis.setImmediate
+
 export const flushEffects = async (): Promise<void> => {
   await act(async () => {
     await new Promise<void>((resolve) => {
-      setImmediate(resolve)
+      realSetImmediate(resolve)
     })
   })
 }
--- a/__tests__/helpers/silence-console-error.ts
+++ b/__tests__/helpers/silence-console-error.ts
@@ -1,0 +1,35 @@
+/**
+ * Mute (and capture) `console.error` for the current test without switching off
+ * the act() regression guard.
+ *
+ * `jest.spyOn(console, "error").mockImplementation(...)` replaces the wrapper
+ * that jest.setup-after-env.ts installs, so for the rest of that file React's
+ * "An update to X inside a test was not wrapped in act(...)" warnings are
+ * neither printed nor failed — the guard is off, silently. This helper swaps
+ * only the *sink* underneath the wrapper, so the inspection stays in place.
+ *
+ *   const errors = silenceConsoleError()
+ *   ...
+ *   expect(errors).toHaveBeenCalledWith(expect.stringContaining("boom"))
+ *
+ * The returned mock receives exactly what `console.error` was called with, so
+ * it is a drop-in for the spy in assertions. The sink is restored after every
+ * test by the guard's own `afterEach` — no teardown needed here.
+ */
+export const silenceConsoleError = (): jest.Mock<void, unknown[]> => {
+  const calls = jest.fn<void, unknown[]>()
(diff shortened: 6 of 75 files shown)


grimen:
Rebased onto `main` (`8242efe5c`) — the branch was a month stale.

Verified after the rebase: `__tests__/screens` + `__tests__/hooks`, **323 suites / 3908 tests green**.

Two notes for whoever reviews:

- `skipHydration` removes most of the `act(...)` warning flood, but not all of it — `HomeScreen` still emits some from its own async state, so this narrows the noise rather than eliminating it.
- This will conflict on one hunk with #4188, which touches `reloadSelfCustodialAccounts` in the same file (it drops a stale read that a newer one outran). Whichever lands second needs a trivial rebase; the two changes are complementary — `skipHydration` keeps the reads out of screen tests, the request guard makes the reads themselves safe when they do run.

grimen:
Follow-up to my rebase comment above — I measured the act(...) warnings properly and the conclusion changed, so please don't merge this on the strength of that comment.

- `__tests__/screens` on plain `main`: **301** act warnings, 297 of them from `AccountRegistryProvider`.
- Wrapping the existing `afterEach` microtask drain (`jest.setup-after-env.ts`) in `act()` so those updates settle legitimately: still **301**. No effect. The wrapper does run — `IS_REACT_ACT_ENVIRONMENT` is `true` and no "testing environment is not configured" warnings appear.
- The warnings are **cross-suite**: `api-key-create.spec.tsx` alone emits **0**; run immediately after `card-limits-screen.spec.tsx` it emits **45**. Same family as the leaked promises/timers behind #3815.

That means `skipHydration` does not fix the timing — it removes the async device reads so nothing is left to land late, which quiets a warning that is reporting something real, and costs screen tests their coverage of provider hydration. Jonas's call is that the warnings stay as a signal, so this branch should not land as-is.

Parking it here rather than reworking it. The branch is rebased onto `8242efe5c` and 323 suites / 3908 tests pass on it, if it's useful as a starting point for a cause-side fix later.

grimen:
Extended this PR to finish what its title claims, rather than leaving the residue to a follow-up.

**The residue is gone.** The note above ("`skipHydration` removes most of the flood, but not all of it — `HomeScreen` still emits some") turned out to be one concrete source, not a diffuse tail. With `skipHydration` in place the only remaining warnings in the whole suite came from `home.spec.tsx`, and all of them traced to [`use-pending-deposits.ts:66`](https://github.com/blinkbitcoin/blink-mobile/blob/8b2c2870100b32d41433bfcff6f6f72f57054bf0/app/self-custodial/hooks/use-pending-deposits.ts#L66) — with no override set the spec fell through to the real hook, whose SDK listing commits `setDeposits` after the test body returns. Fixed at the mock boundary in e883a9f58, for the same reason `skipHydration` is a provider-level fix: which suites trip is timing-dependent, so per-spec patching is whack-a-mole.

**Three judgement calls worth flagging:**

1. *Defaulting `usePendingDeposits` to a settled stub costs no coverage.* Every test that asserts on pending deposits already sets `mockPendingDepositsOverride`; the two places assigning `null` are `afterEach` cleanups, not tests exercising the real hook. The hook's own behavior is covered directly in `__tests__/self-custodial/hooks/use-pending-deposits.spec.ts`.

2. *The 42 `@testing-library/react-hooks` specs had act enforcement silently switched off* (2753ac6d6). That package's peers are `react ^16.9 || ^17`, and it renders through `react-test-renderer` without ever setting `IS_REACT_ACT_ENVIRONMENT` — so React never warned there, and the guard below could not have seen them either. The migration is mostly an import swap; none of the specs used the legacy-only APIs. RNTL types options as `RenderHookOptions<NoInfer<Props>>`, so six specs now annotate the callback parameter instead of relying on inference from `initialProps`.

3. *The guard fails the test rather than counting warnings* (8b2c28701). A warning-only counter would not have prevented the regression this PR is cleaning up twice — #3820 did this once already and it came back precisely because nothing failed.

**On the guard's implementation**, two things were verified rather than assumed:
- jest-circus runs `afterEach` hooks in *registration* order, not reverse. Since RNTL registers its cleanup when a spec imports it (after the setup file), unmount-time warnings arrive too late for `afterEach` — hence the `afterAll` sweep as well.
- The guard was proven to fire: a scratch spec rendering a component with a promise-resolved `setState` failed with the component named, and passed once `await flushEffects()` was added. A guard never seen to fail isn't verified.

**Verification.** Full suite locally: **651 suites, 7900 tests, 0 act warnings**. Three suites fail (`map-component`, `use-place-clusters`, `maps-marker-removal-patch`) — these are a local artifact of a stale `node_modules` (`supercluster` not installed, pre-BTC-Map `react-native-maps` source). Confirmed identical on the unmodified branch with the changes stashed, so they are unrelated to this PR; CI installs fresh and should run them.

grimen:
Rebased onto `main` (`9612ad791`) — #4188 landed and produced the conflict predicted above, in `reloadSelfCustodialAccounts`.

The two changes are complementary, so both survive; the only judgement call was ordering. The `skipHydration` bail-out goes **before** the request-number bump, so a provider that never issues a listing never consumes a request number either:

```ts
const reloadSelfCustodialAccounts = useCallback(async () => {
  if (skipHydration) return

  const request = reloadRequestRef.current + 1
  reloadRequestRef.current = request
  ...
```

Putting the bump first would have worked too, but it would leave the counter advancing on calls that do nothing — noise in the exact mechanism #4188 added to reason about ordering.

Re-verified after the rebase: `use-account-registry`, `multi-account`, `wallet` and #4188's own `transaction-history-screen` spec — 4 suites / 85 tests green, no act warnings. Full-suite re-run in progress.

grimen:
## Review round 1 — all seven findings addressed in 99ec3daeaf1b07fbde6ea904e905ec1e20980bc2

### 1 + 2. `skipHydration` removed rather than narrowed

The prop did not skip hydration, it disabled the provider: `reloadSelfCustodialAccounts` early-returned forever and `loading` was pinned `false`, so the reload that [`use-provision-self-custodial-account.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/99ec3daeaf1b07fbde6ea904e905ec1e20980bc2/app/self-custodial/hooks/use-provision-self-custodial-account.ts) and [`use-restore-wallet.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/99ec3daeaf1b07fbde6ea904e905ec1e20980bc2/app/screens/self-custodial/onboarding/restore/hooks/use-restore-wallet.ts) call after creating or restoring a wallet was a no-op for every consumer, and the loading branch in `use-migration-account.ts` was unreachable through the shared wrapper.

**Considered and rejected:** narrowing the prop so it skipped only the two mount-time effects. Six lines, fixes the severe half — but it leaves a test-only branch in code that ships on every screen, which was the reviewer's actual objection.

**Taken:** [`useComposedAccountRegistry`](https://github.com/blinkbitcoin/blink-mobile/blob/99ec3daeaf1b07fbde6ea904e905ec1e20980bc2/app/hooks/use-account-registry.tsx#L81-L149) is split out and `AccountRegistryContext` exported. The provider's hydration is now byte-identical to `main`; the test wrapper builds a settled value from the same hook. The split was the condition for taking this route — had the stub ended up a copy of the provider, the narrowed prop would have been the better trade.

**What this does *not* fix:** finding 1's trap survives. The shared wrapper still hands out an empty registry, so mocking `listSelfCustodialAccounts` / `getSessionProfiles` does not reach it. It is now *visible* (it lives in the test helper, documented on the component) and *escapable* (`accountRegistry={{ entries: [...] }}`), not eliminated. [`helper-account-registry.spec.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/99ec3daeaf1b07fbde6ea904e905ec1e20980bc2/__tests__/screens/helper-account-registry.spec.tsx) pins both paths so the stub cannot drift from the provider unnoticed.

### 3. Real `usePendingDeposits` restored in `home.spec`

Default is the real hook again; the override is kept for the tests that seed deposits and is now typed `Partial<ReturnType<typeof usePendingDeposits>>`, which immediately caught an untyped `sparkDeposit` fixture. Four tests then tripped the guard — fixed with `await flushEffects()` after `fireEvent.press`, which is what the guard's own advice says to do.

### 4. The guard's bypass is closed

An `afterEach` check turned out to be insufficient: specs restore their spy inside the test body, so by hook time `console.error` is back to the wrapper and the bypass is invisible. [`console.error` is locked](https://github.com/blinkbitcoin/blink-mobile/blob/99ec3daeaf1b07fbde6ea904e905ec1e20980bc2/jest.setup-after-env.ts#L159-L184) instead, so `jest.spyOn` fails at the call site. All 20 call sites across 16 files moved to `silenceConsoleError()`, which swaps a sink underneath the guard and returns a mock for specs that assert on the error.

**Scope note:** [`conversion-details-screen.spec.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/99ec3daeaf1b07fbde6ea904e905ec1e20980bc2/__tests__/screens/conversion-details-screen.spec.tsx) had a `beforeAll` spy that swallowed `"not wrapped in act"` outright — on `main`, predating this branch. Removing it exposed 40 failing tests. The root cause was a **synchronous** `act()` around timer advancement, which returns before Apollo's mocked link resolves its delivery promise; making it `await act(async …)` fixed 38 of them, and 13 render-and-assert tests got a closing `flushEffects()`. Verified stable over four consecutive runs. Fixing it here rather than filing it: leaving the spy in place would have left the single largest hole in the guard this PR exists to install.

### 5. `ip-country-lookup` un-suppressed

The warning is app-owned, so suppressing it also hid any regression that would make it fire for a real reason. It now emits once on the first `resolveIpCountryCode` call instead of at module load — no suite that merely imports the module trips it. The suppression list is back to third-party deprecations only, with the rule stated in the file.

### 6. Lockfile

The orphaned `@testing-library/react-hooks` entry (and `react-error-boundary` with it) removed by hand rather than by `yarn install` — a full re-resolve also bumped unrelated packages (`react-is` 19.2.3 → 19.2.6), which does not belong in this PR. `yarn install --frozen-lockfile` verified to pass and to leave the lockfile untouched.

### 7. PR body

Corrected. The claim that `multi-account.spec` "mounts the provider directly and keeps the real path" was wrong — it renders through `ContextForScreen` and mocks `useAccountRegistry`. Its `mockResolvedValueOnce` chain now carries a comment explaining that the component under test consumes both calls.

### Verification

589 suites / 7224 tests pass. The two `map-component` suites fail on `Cannot find module 'supercluster'` — a dependency missing from this machine's shared `node_modules`, unrelated to this diff and untouched by it.

PR #4011, opened 2026-07-26
fix(android): update to @blinkbitcoin/react-native-geetest-module to fix the geetest crash

### LINKED ISSUE #3842: fix(android): NPE in GeeTest tearDown() — destory() on null GT3GeetestUtils
## Summary

Production crash (Crashlytics): `NullPointerException` when tearing down the GeeTest captcha module.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method
'void com.geetest.sdk.GT3GeetestUtils.destory()' on a null object reference
  at com.reactnativegeetestmodule.GeetestModuleModule.lambda$tearDown$0(GeetestModuleModule.java:134)
  at android.os.Handler.handleCallback(Handler.java:959)
  ...
```

**Crashlytics:** https://console.firebase.google.com/u/2/project/galoyapp/crashlytics/app/android:com.galoyapp/issues/89c168d3305c2047ffb2d9c3d89bdcec?time=7d&types=crash&sessionEventKey=6A2A8AA401EB00012AE34EA6F96E871A_2228527837840948781

## Root cause

`tearDown()` in [`@galoymoney/react-native-geetest-module`](https://github.com/GaloyMoney/react-native-geetest-module) posts a runnable to the UI thread that reads `gt3GeetestUtils` **at execution time** and nulls it, with no null guard:

```java
public void tearDown() {
    getCurrentActivity().runOnUiThread(() -> {
        gt3GeetestUtils.destory();   // line 134 — NPE: field already null
        gt3GeetestUtils = null;
        gt3ConfigBean = null;
    });
}
```

The module is a singleton with a single `gt3GeetestUtils` field, but `setUp`/`tearDown` are tied to a component mount lifecycle in `app/hooks/use-geetest-captcha.ts`:

```ts
useEffect(() => { GeetestModule.setUp(); return () => GeetestModule.tearDown() }, [])
```

Any double-teardown — rapid unmount/remount on navigation, or two consumers — queues two runnables: the first destroys + nulls the field, the second dereferences null. The Crashlytics stack confirms the runnable executed (`Handler.handleCallback → lambda$tearDown$0`), so `getCurrentActivity()` was non-null; the field itself was null.

## Library / version status

- Installed **0.1.5**, which is the **latest** published version (npm `latest`, last released Sept 2022, unmaintained).
- Same buggy code exists on upstream `master` — it is **not version-specific**.
- **Not previously reported** (repo issues #4 and #7 are unrelated).
- The package is **our own** (`GaloyMoney/react-native-geetest-module`), so we control the fix.

This is primarily a **library robustness bug** (unconditional async null-deref). JS cannot observe the native field's state, so the authoritative fix is a null guard in the native module.

## Proposed fix

Make `tearDown()` null-safe and idempotent (and guard a null `getCurrentActivity()`, a sibling crash on the same line when backgrounded):

```java
public void tearDown() {
    Activity activity = getCurrentActivity();
    if (activity == null) return;
    activity.runOnUiThread(() -> {
        if (gt3GeetestUtils != null) {
            gt3GeetestUtils.destory();
        }
        gt3GeetestUtils = null;
        gt3ConfigBean = null;
    });
}
```

Ship via **both**:
1. `patch-package` in blink-mobile for the immediate fix (this repo already patches this module: `patches/@galoymoney+react-native-geetest-module+0.1.5.patch`).
2. Upstream commit + 0.1.6 release to `GaloyMoney/react-native-geetest-module` so the patch isn't carried forever.

Optionally also harden `use-geetest-captcha.ts` to skip `tearDown()` if `setUp()` never completed (defense in depth).

---
_Backlog: production crash, low volume; not user-blocking but should be fixed._

kngako:
Replaces `@galoymoney/react-native-geetest-module` with the published [`@blinkbitcoin/react-native-geetest-module@0.1.2`](https://www.npmjs.com/package/@blinkbitcoin/react-native-geetest-module), which ships the fix for the GeeTest crash (NPE calling `destory()` on a null `GT3GeetestUtils` during `tearDown()`) upstream.

- `package.json` / `yarn.lock`: the dependency now comes from npm (`0.1.2`) instead of the temporary dev build from a fork.
- `patches/`: drops the patch-package patch for the old package — the fix lives upstream now.
- `app/hooks/use-geetest-captcha.ts`: import moved to the `@blinkbitcoin` scope.
- `__tests__/hooks/use-geetest-captcha.spec.ts`: adds mocked tests for the captcha hook (registration flow, dialog results, failure events, resets).
- `app/graphql/generated.ts`: regenerated against the current API schema (picks up the new `DepositFeeTier` type).
- `ios/Podfile.lock`: regenerated with `pod install` (CocoaPods 1.16.2, the version pinned in `ios/Gemfile`). Besides the geetest entry, the regeneration also caught up entries that had drifted from `yarn.lock` (`RNVectorIcons` 10.0.3 → 10.3.0, which is what `yarn.lock` already resolves) and refreshed a few spec checksums.

Verified locally: `yarn install`, `tsc -p .` and the geetest hook test suite (19 tests) all pass.


kngako:
This PR should just need to update the repo where the dependency is getting pulled from on the related PR is merged.

Ran on Android

kngako:
Run a android build with the rebase and npm served dependency. Good.

image

esaugomez31 (approved):
I already tested it with the new version of the library and it works fine

PR #4016, opened 2026-07-27
feat(self-custodial): spark wire codec, identity, and exit-bundle exporter

openoms:
**Stack 1/3** for the unilateral-exit recovery bundle (split out of #3911 per review). Next: `feat--recovery-bundle-storage` -> #3911.

The cryptographic core, with no app wiring - nothing outside these modules imports them yet:

- `protocol/wire.ts` - minimal protobuf wire primitives (varint/length-delimited only), 64-bit masking, proto3 last-wins field selection. Hand-rolled so TreeNodes can be stored byte-for-byte as received.
- `protocol/messages.ts` - encoders/decoders for the Spark authn and query_nodes messages, mirroring spark.proto field numbers.
- `protocol/grpc-web.ts` - fetch-based gRPC-web unary client: 30s per-call timeout, rejects compressed frames, requires a trailers frame (or header-carried status), errors on truncated remainders.
- `identity.ts` - seed-derived Spark identity at m/8797555'/{account}'/0' (network-dependent default account number) and DER challenge signing; bundle encryption key derivation.
- `exporter.ts` - pages owner leaves (offset advanced by returned page size), re-fetches missing ancestors by id (bypassing the operators' root-skip bug), and refuses any bundle whose exit chain is not closed: cycles, unresolved ancestors, implausible leaf values, exceeded paging caps or the 2-minute total fetch budget are all errors, not warnings.

Tests: golden-vector identity derivation (independent Python implementation), determinism + low-S signature pin, byte-level mocked-operator exporter flows (short pages, cycles, refetch rounds, budget), wire/grpc-web edge cases.

Also carries a `chore(graphql)` commit regenerating `generated.ts` for the backend's new `DepositFeeTier` field - main is currently stale against the live schema, and every PR fails `check:codegen` without it.

### DIFF
--- a/__tests__/self-custodial/recovery-bundle/exporter.spec.ts
+++ b/__tests__/self-custodial/recovery-bundle/exporter.spec.ts
@@ -30,6 +30,8 @@ import {
 import { RECOVERY_BUNDLE_SCHEMA } from "@app/self-custodial/recovery-bundle/types"
 import type { DecodedTreeNode } from "@app/self-custodial/recovery-bundle/protocol/messages"
 
+import packageJson from "../../../package.json"
+
 const TEST_MNEMONIC =
   "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"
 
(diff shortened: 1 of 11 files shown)


grimen:
Pushed two small follow-up commits from the review:

- 9cd902a — `sparkSdkVersion` now embeds the pinned SDK version (`breez-sdk-spark-react-native@0.15.0`, read from `package.json` so it tracks dep bumps) instead of just the package name.
- 7f2a62c — closes two coverage gaps: a 101-ancestor refetch round pins the `queryNodesByIds` 100+1 request split, and the mock operator now decodes the `verify_challenge` request body so the field numbers in `encodeVerifyChallengeRequest` are actually asserted.

No exporter behavior changes; recovery-bundle suites pass locally (27/27).

grimen (approved):

grimen (changes requested):
## Findings

**1. The owner-query paging skips owned leaves whenever `include_parents` adds ancestors to a page (blocking).**
`queryNodesByOwner` moves the request offset forward by `response.nodes.size` (`exporter.ts:194-195`). That map holds the owned rows plus the ancestors merged in by `include_parents`. The operator counts only the owned rows when it sets the next offset. In `spark/so/handler/tree_query_handler.go`, `QueryNodes` applies `Offset(offset).Limit(limit)` to the owner rows, then merges `resolveAncestorChains` into `protoNodeMap`, and then returns `nextOffset = offset + len(nodes)`, or `-1` when `len(nodes) < limit`. Ancestors are never owned rows: `SPLITTED` is excluded from the owner query and ancestors come back with a NUMS owner. So every ancestor on a page pushes the next request past owned rows that were never returned. The loop still ends normally, and the chain checks cannot notice because they only walk the leaves that were fetched. The JS SDK's `LeafManager.queryNodes` advances by `pageSize`, and so does the reference exporter (`src/operator/exporter.ts:148`). My proposal in finding 1 on #3911 introduced this skip. The short page with a positive offset it guarded against cannot happen, because the operator returns `-1` for any short page.

*Example:* a wallet holds 150 available leaves with one ancestor each. Page 1 returns 100 leaves and 100 ancestors with offset 100. The exporter requests page 2 at offset 200, gets an empty page, and `fetchRecoveryBundle` resolves with 100 leaves and `btcSats: "100000"`. Reproduced locally against a mock page function that follows the handler above: the requested offsets were `[0, 200]`, and 50,000 sats were missing from a bundle saved as complete.

*Proposal:* one option is `offset = Number(response.offset)`, the operator's own resume point, together with a guard that the offset strictly increases. With that change the repro passes. The only existing test that fails is `pages through the owner query until the operator returns offset 0`, because its mock returns offset 100 for a one-node page.

## Test Coverage Gaps

- The mock operator's owner paging (`exporter.spec.ts:122-178`, pages at 427-430 and 461-464) does not behave like the real handler. None of its multi-page owner responses include ancestors, and short pages still return a positive offset (`{ nodes: [pageOneLeaf], offset: 100 }`), which the operator never does. As a result, the two paging tests pin the behaviour behind finding 1. A shared page function that follows `tree_query_handler.go` would have caught it: owned rows plus merged ancestors, and a next offset of `offset + owned` or `-1`.

PR #4017, opened 2026-07-27
feat(self-custodial): recovery-bundle storage, encryption, and refresh orchestration

openoms:
**Stack 2/3** for the unilateral-exit recovery bundle (split out of #3911 per review). Base: `feat--recovery-bundle-core`. Next: #3911.

Persistence and orchestration on top of the exporter - still no UI or hooks:

- `encryption.ts` - AES-128-GCM payload keyed from HMAC-SHA256(seed, context); plaintext envelope limited to freshness/matching metadata and cross-checked against the authenticated contents after decrypt.
- `storage.ts` - bundle file under a per-network directory excluded from iOS device backups (D9), written tmp-then-rename so an interrupted write keeps the last good copy (stranded tmp recovered on load, swept on delete); validated AsyncStorage freshness state.
- `settings.ts` - per-account auto-refresh opt-out and cloud-sync opt-in.
- `cloud.ts` - best-effort silent upload to the seed backup's provider (iCloud, or Drive only when a silent sign-in succeeds - the background path never pops UI); failures carry the original error.
- `refresh.ts` - single-flight per (account, network) refresh: fetch, encrypt, persist, then cloud sync in its own error scope so a cloud failure cannot fail a persisted refresh. Cloud sync is gated on `isCloudSyncAllowedFor`: opted in AND password-protected cloud seed backup (D9). Account-deletion marks block in-flight refreshes from persisting after a delete, with `waitForRefreshesToSettle` + unmark for the caller's re-sweep.
- `backup-state.tsx` - records `cloudPasswordProtected` with a completed backup (new optional field/param, backward compatible), making the D9 gate decidable.

Tests: encrypted-payload round-trips including a GCM bit-flip tamper case, real-execution storage tests on an in-memory RNFS with iOS moveFile semantics, refresh gating/single-flight/deletion-race coverage, and a cloud spec for both platform branches.

### DIFF
--- a/__tests__/self-custodial/providers/backup-state.spec.tsx
+++ b/__tests__/self-custodial/providers/backup-state.spec.tsx
@@ -7,7 +7,10 @@ import {
   BackupMethod,
   BackupStatus,
   completedMethodsOf,
+  isCloudSeedBackupCompleted,
+  isPasswordProtectedCloudSeedBackup,
   markBackupCompletedFor,
+  readBackupStateFor,
   removeBackupStateFor,
 } from "@app/self-custodial/providers/backup-state"
 import { AccountType, AccountStatus } from "@app/types/wallet"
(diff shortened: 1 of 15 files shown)


grimen:
Pushed five commits addressing findings from review ([`3fabf18`](https://github.com/blinkbitcoin/blink-mobile/commit/3fabf18e3ed9853eeea92381b1c574fb6a9da9aa)..[`aa16588`](https://github.com/blinkbitcoin/blink-mobile/commit/aa1658890aac04d6a8ba644fc3f2fe263d3857cb)):

**1. Reporting boundary** ([`3fabf18`](https://github.com/blinkbitcoin/blink-mobile/commit/3fabf18e3ed9853eeea92381b1c574fb6a9da9aa)). `refresh.ts` called `crashlytics().recordError` directly, bypassing the single recording boundary in `error-reporting.ts` (#3928). A transient network failure during the post-payment silent upload would have been filed as a non-fatal defect for every user on a flaky link. All sites now go through `recordAppError`: the reasons the interactive Drive hook already treats as expected (`auth`, `permission-denied`, `cancelled`, `transient`) become breadcrumbs, real defects keep recording with dedup keys.

**2. Dead-token recovery on the silent path** ([`0b1e400`](https://github.com/blinkbitcoin/blink-mobile/commit/0b1e40075774b84b68d55d15740f2d1daa35e6d6)). `getTokens()` keeps returning a token revoked out-of-band (the reason `callDrive` exists in `use-google-drive-backup.ts`), so one revocation permanently killed silent sync, misdiagnosed as "never linked". `callDrive` now lives in `app/utils/google-drive-session.ts` and the silent upload uses it too: on a `DriveError` 401 it clears the cached token and retries once. The hook change is a pure move (its 44 existing tests pass unchanged); the retry behavior on the silent path is covered in `cloud.spec.ts`.

**3. Load no longer renames a stranded tmp** ([`ce3793a`](https://github.com/blinkbitcoin/blink-mobile/commit/ce3793ad7e23326495721e15174a98eaa68a6dac)). `loadEncryptedBundleFile`'s repair-rename could steal a concurrent save's rename out of its unlink-to-move window (relevant once the screen-triggered sync lands in stack 3), making a persisted refresh report failure. The stranded payload is now read in place; repair stays with save/delete.

**4. Coverage for `cloudPasswordProtected`** ([`d70fa14`](https://github.com/blinkbitcoin/blink-mobile/commit/d70fa14a517ff622ef74b1d3c7311a77a312580f)). Nothing failed if a write dropped the flag that decides the D9 gate. Added the `setBackupCompleted` persist-shape test, the `markBackupCompletedFor` round-trip into `isPasswordProtectedCloudSeedBackup`, and the default-closed passwordless case.

**5. Isolated spec for the shared Drive session helper** ([`aa16588`](https://github.com/blinkbitcoin/blink-mobile/commit/aa1658890aac04d6a8ba644fc3f2fe263d3857cb)). The extracted `google-drive-session.ts` was only exercised through its two call sites; it now has its own spec pinning the contract: pass-through, 401 clear-and-retry-once (returning the refreshed token), no second retry, non-401 and non-Drive errors rethrown untouched, and the original 401 kept when the refresh itself fails.

All recovery-bundle + backup-state + drive-hook + session-helper specs pass locally (143 tests); `tsc` and lint are clean on the touched files.

grimen (dismissed):

grimen (dismissed):

grimen (dismissed):

grimen (dismissed):

esaugomez31 (approved):
LGTM

PR #4024, opened 2026-07-27
fix(receive): show the real onchain deposit fee above 1M sats

grimen:
Fixes blinkbitcoin/blink-wip#1013

## Problem

The receive screen and the Fee Rates settings page derived the above-threshold onchain deposit fee from `globals.feesInformation.deposit.ratio` — basis points applied to `minBankFeeThreshold`. Onchain receive fees moved to flat-per-tier pricing, so the server has hardcoded `ratio: "0"` ever since, and both surfaces rendered **0 sats** while deposits over 1M sats are actually charged **5,000 sats** (blink-deployments#8918, 2026-01-13).

## Fix

blinkbitcoin/blink#717 deprecated `ratio` and added `tiers` to `feesInformation.deposit`. Both queries now select `tiers` instead, and `formatDepositFees` takes the over-threshold fee from the unbounded tier (the one with no `maxAmount`) rather than doing the basis-point math. `minBankFee` / `minBankFeeThreshold` are unchanged and still drive the below-threshold amount and the "1M" threshold text, so no copy and no translations change.

The unbounded tier is found by predicate rather than by position, and the existing 5,000 default remains the fallback when it is missing or non-numeric.

## Verification

- `deposit-fees.spec.ts` rewritten: staging-shaped tiers now yield `overFee: "5,000"` (the regression), plus unbounded-tier-anywhere, zero amount, missing tiers, and non-numeric fallbacks.
- Affected suites pass: `deposit-fees`, `contextual-info`, `fee-rates-screen`, `receive`, `use-wallet-resolution`, `self-custodial/use-payment-request` — 125 tests.
- `yarn tsc:check`, `yarn eslint` on changed files, and `yarn graphql-check` all clean. The remaining "1 document with deprecated fields" notice is pre-existing on `main` and unrelated.

## ⚠️ Draft — blocked on a prod deploy

`tiers` is live on staging but **not yet on prod** (verified today):

    curl -s https://api.blink.sv/graphql -H 'Content-Type: application/json' \
      -d '{"query":"{ globals { feesInformation { deposit { tiers { maxAmount amount } } } } }"}'

Requesting an unknown field fails the entire `paymentRequest` query, and `use-wallet-resolution.ts` returns `null` in that case — the receive screen would render blank for everyone. CI cannot catch this, since codegen and `graphql-check` both run against staging. Re-run the curl above and confirm it returns tier data before marking this ready.

### DIFF
--- a/__tests__/components/contextual-info.spec.tsx
+++ b/__tests__/components/contextual-info.spec.tsx
@@ -67,6 +67,11 @@ jest.mock("@app/i18n/i18n-react", () => ({
           overFee: string
         }) =>
           `Deposit fee: ${fee} SAT for amounts under ${threshold} SAT or ${overFee} SAT for deposits over ${threshold} SAT`,
+        depositFeeTiers: ({ tiers }: { tiers: string }) => `Deposit fees: ${tiers}`,
+        depositFeeTierUpTo: ({ fee, max }: { fee: string; max: string }) =>
+          `${fee} SAT up to ${max} SAT`,
+        depositFeeTierAbove: ({ fee, min }: { fee: string; min: string }) =>
+          `${fee} SAT above ${min} SAT`,
         autoConvertMinAmount: ({
           minSats,
           minFiat,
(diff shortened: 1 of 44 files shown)


grimen:
## Before / after

Settings → Fee rates, iOS simulator, both builds pointed at **the same live staging backend** (`api.staging.blink.sv`, which already serves `tiers`). The only difference between the two runs is the app code — `origin/main` on the left, this branch on the right.

| Before (`origin/main`) | After (this PR) |
|---|---|
|  |  |
| Onchain above 1M SAT — **0 SAT** | Onchain above 1M SAT — **5,000 SAT** |

Staging returns `tiers: [{maxAmount: "1000000", amount: "2500"}, {maxAmount: null, amount: "5000"}]` and `ratio: "0"`, so `main` renders `1000000 × 0 / 10000 = 0` while this branch reads 5,000 off the unbounded tier. The below-1M row is unchanged at 2,500 SAT, as expected — it still comes from the first tier.

## A third tier



Each tier gets its own row, with both bounds read from the tier list. Staging only has two tiers today, so this one shot is against a **stubbed three-tier response** rather than live data — the same case is asserted for real in `fee-rates-screen.spec.tsx` and `deposit-fees.spec.ts`.

This matters because the earlier revision of this PR would have rendered a *wrong* number here, not merely an incomplete list: it kept labelling the last row "above 1M SAT" from the deprecated `minBankFeeThreshold` while showing the unbounded tier's fee, so a 2M-sat deposit would have been quoted 5,000 instead of 4,000.

The receive screen's onchain notice is not pictured because it needs a logged-in account. It keeps the existing two-tier sentence (so all 28 translations stay accurate) and composes per-tier fragments otherwise, covered by `contextual-info.spec.tsx`.

<sub>Screenshots live on the `assets/pr-4024-screenshots` branch so they stay out of this diff. That branch has to stay around — deleting it breaks the images above.</sub>

esaugomez31 (approved):

PR #4025, opened 2026-07-27
feat(settings): open the terminal in the user's display currency

### LINKED ISSUE #4019: feat(mobile): Supply users default currency when redirecting them to Blink Terminal
Blink Terminal defaults to USD and users may prefer to transact using a different unit of account. Logic to parse amount, memo, currency parameters was recently implemented and using this would improve the UX.

When a user clicks on the POS/Terminal under ways to get paid the link should have the display query parameter set to the users default currency so the terminal shows that. 

An alternative would be pulling the default currency associated with a username from the backend. But this isn't easy for self custodial users default currency selections.

**Additional context**

https://github.com/blinkbitcoin/blink-terminal/issues/47
https://github.com/blinkbitcoin/blink-terminal/pull/48

grimen:
## Summary

Blink Terminal defaults its point of sale to USD, so a merchant whose app is set to another currency had to change it by hand every time they opened their cash register from **Settings → Ways to get paid → Point of Sale**.

blinkbitcoin/blink-terminal#48 taught the public POS page to parse a `display` query param, so the link can carry the currency the merchant already picked:

```
https://terminal.blinkbtc.com/<username>?display=EUR
```

The terminal falls back to USD for codes it does not know, so passing our own setting through stays safe if the two currency lists ever drift.

## What changed

- **`app/utils/pay-links.ts`** — `getPosUrl` takes an optional display currency and appends `?display=<code>` only when one is given, so every existing caller keeps producing the exact same URL.
- **`app/screens/settings-screen/settings/account-pos.tsx`** — reads `useEffectiveDisplayCurrency()`, which already unifies both account types (custodial reads `me.defaultAccount.displayCurrency`, self-custodial reads persistent state). Its `loading` flag is folded into the row's existing `loading` prop: the hook answers `"USD"` while the custodial query is in flight, and opening the terminal on that fallback would be wrong for every non-USD merchant.

Two notes for reviewers:

- The row waits for the currency before it becomes tappable, because `useEffectiveDisplayCurrency` answers `"USD"` while the custodial query is in flight. The trade-off is that a cold cache behind a hanging network holds the row in its skeleton state; opening a merchant's cash register in the wrong currency is the worse outcome of the two.
- That query is the one settings query that does not pass `skip: !isAuthed`. The hook keeps it unauthed-safe with `fetchPolicy: "cache-only"` instead, so it still never reaches the network for a logged-out user — which is why `skip-queries-when-unauthed.spec.tsx` mocks it rather than asserting the skip flag.

Scope is the settings row only. The printable QR (`/print`) is left alone — blink-terminal parses the params on `/[blinkusername]` only, so a param there would be silently ignored. The older Galoy Address screen keeps its bare URL, which is the one that gets copied and shared.

## Testing

- `__tests__/utils/pay-links.spec.ts` — the param is appended, omitted when no currency is known, and encoded so a stray value cannot append another parameter.
- `__tests__/screens/settings-screen/settings/pay-link-rows.spec.tsx` — custodial and self-custodial POS rows both open the terminal with `?display=EUR`; the row stays in its loading state while the currency resolves, while the printable QR row (which carries no currency) does not.
- `npx jest --runTestsByPath` on the settings-screen suites, `npx tsc --noEmit` and `npx eslint` on the changed files all pass locally.

Fixes #4019

### DIFF
--- a/__tests__/screens/settings-screen/settings/pay-link-rows.spec.tsx
+++ b/__tests__/screens/settings-screen/settings/pay-link-rows.spec.tsx
@@ -192,7 +192,9 @@ describe("ways to get paid rows", () => {
       expect(lastRowProps().loading).toBe(true)
     })
 
-    it("does not gate the rows that carry no currency", () => {
+    // A guard against the gate spreading: the printable QR link has no currency in it,
+    // so it has no reason to wait for one.
+    it("leaves the printable QR row tappable meanwhile", () => {
       mockUseEffectiveDisplayCurrency.mockReturnValue({
         displayCurrency: "USD",
         loading: true,
--- a/__tests__/screens/settings-screen/settings/pay-link-rows.spec.tsx
+++ b/__tests__/screens/settings-screen/settings/pay-link-rows.spec.tsx
@@ -34,6 +34,11 @@ jest.mock("@app/graphql/generated", () => ({
   useSettingsScreenQuery: () => mockSettingsScreenQuery(),
 }))
 
+const mockUseEffectiveDisplayCurrency = jest.fn()
+jest.mock("@app/hooks/use-effective-display-currency", () => ({
+  useEffectiveDisplayCurrency: () => mockUseEffectiveDisplayCurrency(),
+}))
+
 jest.mock("@rn-vui/themed", () => ({
   useTheme: () => ({ theme: { colors: { primary: "#fc5805", black: "#000" } } }),
 }))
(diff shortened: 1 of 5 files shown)


grimen:
## Before / after

What a merchant with **EUR** as their display currency lands on after tapping **Settings → Ways to get paid → Point of Sale**. Captured against the live terminal at phone width, using `pretyflaco` — the same example account as blinkbitcoin/blink-terminal#47.

| Before — `…/pretyflaco` | After — `…/pretyflaco?display=EUR` |
|---|---|
|  |  |
| Opens in USD regardless of the currency set in the app; the merchant switches it by hand every session. | Opens in the currency the merchant already picked. |

The app screen itself is unchanged — same row, same tap — so the only difference to see is the page the link lands on. That the app now emits the param is covered by the tests: `pay-link-rows.spec.tsx` asserts `Linking.openURL` is called with `https://terminal.blinkbtc.com/<username>?display=EUR` for both custodial and self-custodial accounts.

esaugomez31 (approved):
Works well, LGTM!

PR #4026, opened 2026-07-27
fix(config): add China and Myanmar to offboard-only countries

### LINKED ISSUE #3974: feat(migration): Add offboard-only persistent notification for restricted countries
## Context

For users in restricted countries (offboard-only regions), we need a **separate** persistent in-app notification — distinct from the standard migration notification shown in most regions.

The standard notification (shown in most regions) has:
- **Title:** Important
- **Body:** Move to a non-custodial account via guided migration, or withdraw, before August 31. Receiving stops August 10.
- **Button:** Migrate

The offboard-only notification described here is different — it does **not** offer migration, only withdrawal.

This notification should be **non-dismissable** and visible every time the user opens the app, until they have withdrawn their funds or the deadline passes.

## Copy (offboard-only regions)

**Title:** Important

**Body:** Your account is registered in a region where we cannot offer accounts. Withdraw your funds before August 31. Receiving stops August 10.

**Button:** None — no action button on this notification.

## Requirements

- Show this persistent notification for users in offboard-only countries
- Must be different from the standard migration notification shown in other regions
- Notification must not be dismissable
- No button — text-only notification
- Receiving stops August 10
- Deadline: August 31 for full withdrawal
- Offboard-only countries: UAE, Nepal, Algeria, China, Myanmar

## Related

- Non-custodial migration rollout
- Custodial sunset timeline

grimen:
## What

Adds `CN` (China) and `MM` (Myanmar) to the `offboardOnlyCountries` remote config default, which shipped in #3977 as `["AE", "NP", "DZ"]`.

The final offboard-only list for the custodial sunset is now confirmed: **UAE, Nepal, Algeria, China, Myanmar**.

## Why

#3977 landed with a provisional default while the country list was still TBD in #3974. With the list confirmed, the bundled default should match it so CN/MM-registered custodial accounts get the persistent "withdraw your funds" bulletin even before Remote Config is fetched (or if the fetch fails).

## Remote Config

The `offboardOnlyCountries` parameter is now live in Firebase Remote Config with the same five codes, so the confirmed list applies to already-shipped app versions too — this PR only aligns the bundled fallback.

offboardOnlyCountries in the Firebase Remote Config console

The retirement mechanism is unchanged: ops empties the remote array (`[]`, not a blank value) after the withdrawal deadline. The client never compares dates.

## Test plan

- `__tests__/config/feature-flags-default-blocks.spec.ts` — default list assertion updated; canonical-form check (uppercase alpha-2, no duplicates) still passes.
- `__tests__/screens/account-migration/hooks/use-offboard-only-bulletin.spec.ts` — unchanged, passes.

Ref #3974

### DIFF
--- a/__tests__/config/feature-flags-default-blocks.spec.ts
+++ b/__tests__/config/feature-flags-default-blocks.spec.ts
@@ -99,7 +99,13 @@ describe("defaultRemoteConfig: compliance country lists", () => {
     assertCanonical(defaultRemoteConfig.offboardOnlyCountries)
   })
 
-  it("offboardOnlyCountries defaults to the UAE, Nepal and Algeria", () => {
-    expect(defaultRemoteConfig.offboardOnlyCountries).toEqual(["AE", "NP", "DZ"])
+  it("offboardOnlyCountries defaults to the UAE, Nepal, Algeria, China and Myanmar", () => {
+    expect(defaultRemoteConfig.offboardOnlyCountries).toEqual([
+      "AE",
+      "NP",
+      "DZ",
+      "CN",
+      "MM",
+    ])
   })
 })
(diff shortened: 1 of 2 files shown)


littledino2112 (approved):
LGTM

PR #4027, opened 2026-07-27
fix(bulletins): give notification card titles the themed bold preset

### LINKED ISSUE #4012: fix: "Your funds are at risk" bulletin missing the word "risk" on non-custodial accounts
## Description

The bulletin/notification that reads "Your funds are at risk" is missing the word "risk" when displayed on non-custodial accounts. The text appears truncated or incomplete.

## Steps to reproduce

1. Log in with a non-custodial account
2. Observe the bulletin/notification banner
3. Notice the word "risk" is missing from the text

## Expected behavior

The full text "Your funds are at risk" should be displayed.

## Actual behavior

The word "risk" is cut off or missing from the bulletin text on non-custodial accounts.

## Additional context

Reported by @designsats in Slack.

---
Filed by Blink-Claw-Bot ⚡

grimen:
Fixes #4012

## Problem

@designsats reported that the **"Your funds are at risk"** bulletin looks like it is missing the word *risk*.

The copy is intact — `BackupNudge.title` in `app/i18n/en/index.ts` reads `"Your funds are at risk"`, and all 28 locale files carry the full string. Nothing truncates the text either: `NotificationCardUI` sets no `numberOfLines` and no `ellipsizeMode`, so the title wraps rather than ellipsizing.

What it did have was this:

```ts
titleStyle: { fontSize: 14, lineHeight: 20, fontWeight: "400", color: colors.black },
bodyText:   { fontSize: 14, lineHeight: 20, fontWeight: "400", color: colors.grey2 },
```

The title and the body were styled **identically** — same size, same line height, same weight, differing only in colour. Every bulletin rendered its headline as if it were another paragraph, with nothing telling the eye where the title ended and the body began. When the title wraps (narrow device, or a longer locale — `ro` is 39 characters, `ja` and `el` longer still), the wrapped fragment reads as part of the body sentence.

It was also a convention violation: the app has a themed `Text` with size presets (`app/rne-theme/theme.ts`), used in 106 files. This card hardcoded `14/20`, which matches no preset — it had quietly drifted from the design system.

## Change

Adopt the themed presets on both texts:

```diff
-<Text style={styles.titleStyle}>{title}</Text>
-<Text style={styles.bodyText}>{text}</Text>
+<Text type="p3" bold>
+  {title}
+</Text>
+<Text type="p3" style={styles.bodyText}>
+  {text}
+</Text>
```

`titleStyle` disappears entirely — the theme already defaults to `colors.black`. `bodyText` survives only to carry `colors.grey2`, which is the card's own decision and not something a size preset owns.

Net visual delta: title `400 → 600` weight, both texts `lineHeight 20 → 18`. Size is unchanged at 14, so the title is no wider than before.

Scope is deliberately the one card, which covers every bulletin — `backup-nudge-banner`, `self-custodial-info-bulletin`, `migration-reminder-bulletin`, `offboard-only-bulletin` all render through it. `settings-screen/settings-card.tsx` shows the same string with its own hardcoded `14/700` title, but its title is already bold and so does not exhibit the symptom; left alone.

## Tests

Added one assertion in `notification-card-ui.spec.tsx` locking the convention in, so a future hardcode regresses loudly.

## Verification

- `notification-card-ui`, `backup-nudge-banner`, `backup-nudge-modal`, `self-custodial-info-bulletin` — 30/30 pass
- `bulletins`, `migration-reminder-bulletin`, `offboard-only-bulletin` — 26/26 pass
- `tsc --noEmit` clean, `eslint` clean on both changed files

Draft pending a visual check on a simulator with a non-backed-up self-custodial account over `backupNudgeBannerThreshold`.

### DIFF
--- a/__tests__/components/notification-card-ui.spec.tsx
+++ b/__tests__/components/notification-card-ui.spec.tsx
@@ -62,6 +62,14 @@ describe("NotificationCardUI", () => {
     expect(getByText("Test Body")).toBeTruthy()
   })
 
+  it("styles the title with the themed bold preset so it outranks the body", () => {
+    const { getByText } = render(<NotificationCardUI {...defaultProps} />)
+
+    expect(getByText("Test Title").props).toMatchObject({ type: "p3", bold: true })
+    expect(getByText("Test Body").props).toMatchObject({ type: "p3" })
+    expect(getByText("Test Body").props.bold).toBeUndefined()
+  })
+
   it("renders icon when provided", () => {
     const { queryByTestId } = render(
       <NotificationCardUI {...defaultProps} icon="warning" />,
--- a/app/components/notifications/notification-card-ui.tsx
+++ b/app/components/notifications/notification-card-ui.tsx
@@ -116,7 +116,7 @@ const useStyles = makeStyles(({ colors }) => ({
   textColumn: {
     flex: 1,
     flexDirection: "column",
-    alignItems: "flex-start",
+    alignItems: "stretch",
     justifyContent: "center",
   },
   leftIconContainer: {
--- a/app/components/notifications/notification-card-ui.tsx
+++ b/app/components/notifications/notification-card-ui.tsx
@@ -50,8 +50,12 @@ export const NotificationCardUI: React.FC<NotificationCardUIProps> = ({
             </View>
           )}
           <View style={styles.textColumn}>
-            <Text style={styles.titleStyle}>{title}</Text>
-            <Text style={styles.bodyText}>{text}</Text>
+            <Text type="p3" bold>
+              {title}
+            </Text>
+            <Text type="p3" style={styles.bodyText}>
+              {text}
+            </Text>
           </View>
           {dismissAction && (
             <GaloyIconButton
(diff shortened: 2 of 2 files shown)


grimen:
## Before / after

iOS 26.5 simulator (iPhone 16), both bulletins mounted side by side so the change is visible on every card that goes through `NotificationCardUI`.

### Default text size

| Before (`main`) | After |
|---|---|
|  |  |

The title and the body are the same weight on `main` — the headline reads as the first line of the paragraph. After, the title carries its own weight.

### Accessibility text size (Larger Text → Accessibility XL)

| Before (`main`) | After |
|---|---|
|  |  |

This is the reported case. The title wraps to `"Your funds are at"` / `"risk"`, and on `main` that orphaned `"risk"` sits in the same weight as everything around it with no cue that it belongs to the headline — which is what reads as a missing word. After, the wrap is unambiguous.

Worth noting for the reviewer: the string itself is never truncated in either shot. `BackupNudge.title` is intact in `app/i18n/en/index.ts` and all 28 locales, and the card sets no `numberOfLines`/`ellipsizeMode`. This is a hierarchy problem, not a clipping one.

<sub>Screenshots live on the `pr-4027-assets` branch — please don't delete it or the images above will 404.</sub>

designsats:
LGTM

esaugomez31:
This is how it looks:

| Default font size | x2.5 |
|---|---|
| Screenshot from 2026-07-27 16-27-05 | Screenshot from 2026-07-27 16-31-34 |

esaugomez31 (approved):
LGTM

PR #4028, opened 2026-07-27
feat(privacy): warn that USDB token activity is public

### LINKED ISSUE #3831: fix: Warn users that USDB token activity is public
## Context

Spark private mode only applies to Bitcoin payments. Token payments are not affected by private mode and may still be publicly visible.

This creates a privacy expectation risk for USDB users: if someone can connect a user's USDB activity to them, they may be able to see USDB balance, transfers, timing, and amounts. Users should be clearly informed before they receive USDB or convert funds into USDB.

Related: #3830

## Feature

```gherkin
Feature: As a non-custodial Blink user, I want to be clearly warned before receiving or converting funds into USDB, so that I understand USDB/token activity is public and can make an informed privacy decision.

 Scenario: the user attempts to receive USDB for the first time
 Given the user has a non-custodial Blink account
 And the user has not previously acknowledged the USDB public activity warning
 When they tap "Receive" for USDB
 Then a modal is displayed before the receive screen is shown
 And the receive action does not proceed until the user acknowledges the warning.

 Scenario: the user attempts to convert funds into USDB for the first time
 Given the user has a non-custodial Blink account
 And the user has not previously acknowledged the USDB public activity warning
 When they start a conversion from BTC or another balance into USDB
 Then a modal is displayed before the conversion can be confirmed
 And the conversion does not proceed until the user acknowledges the warning.

 Scenario: the warning explains Spark private mode limitations
 Given the USDB public activity warning modal is displayed
 When the user reads the modal
 Then it states that Spark private mode only applies to Bitcoin payments
 And it states that USDB/token balances and transfers are not affected by private mode
 And it states that USDB/token activity may still be publicly visible.

 Scenario: the warning explains the privacy risk
 Given the USDB public activity warning modal is displayed
 When the user reads the modal
 Then it explains that someone who can connect the user's USDB activity to them may be able to see their USDB balance, transfers, timing, and amounts.

 Scenario: the user cancels the warning
 Given the USDB public activity warning modal is displayed
 When the user taps "Cancel" or dismisses the modal
 Then they return to the previous screen
 And no receive address is shown
 And no conversion is created or confirmed.

 Scenario: the user acknowledges the warning
 Given the USDB public activity warning modal is displayed
 When the user taps "I understand, continue"
 Then the acknowledgement is saved
 And the user proceeds to the receive or conversion flow they originally started.

 Scenario: the user has already acknowledged the warning
 Given the user has previously acknowledged the USDB public activity warning
 When they later receive USDB or convert funds into USDB
 Then the modal is not shown again
 And the user proceeds directly to the selected flow.

 Scenario: the user can revisit the warning later
 Given the user has a non-custodial Blink account with USDB available
 When they open the USDB settings, token settings, or privacy help area
 Then they can view the USDB public activity warning again.

 Scenario: the acknowledgement state is specific to this privacy warning
 Given the user has acknowledged other non-custodial or Spark warnings
 When they attempt to receive or convert funds into USDB for the first time
 Then the USDB public activity warning is still shown
 And prior acknowledgements for unrelated warnings do not suppress it.
```

grimen:
Fixes #3831

## Context

Spark's privacy guarantees are asymmetric. Bitcoin balances and transactions are private, but the USDB token backing the non-custodial Dollar balance is not: anyone who knows a user's Lightning address can look up their Dollar balance and history on a public explorer. That leak is tracked separately in #3830 (indexer/explorer side) — this PR is the mobile half, warning the user before they hold value in Dollars.

A self-custodial user now sees a one-time privacy warning the first time they select **Dollar** on Receive, or open the **Transfer** flow. Custodial accounts never see it, and the acknowledgement is stored per account.

## Screenshots

Non-custodial account on an isolated simulator. The dark shot is the like-for-like comparison against the design artboard.

| Before | After (light) | After (dark) |
|---|---|---|
|  |  |  |

Dimensions were matched to the mock by measuring both renders and normalising against card width, rather than by eye:

| metric | mock | before | now |
|---|---|---|---|
| eye icon width | 14.5% | 20.6% | 14.4% |
| title glyph band | 7.2% | 6.1% | 7.0% |
| body line pitch | 7.1% | 6.1% | 6.9% |
| icon → title gap | 12.5% | 11.7% | 12.2% |
| title → body gap | 8.9% | 5.1% | 8.2% |
| card aspect (h/w) | 1.34 | 1.20 | 1.32 |

The one metric left off the mock is button height (15.9% on the artboard, 13.8% here). That is the shared `GaloyPrimaryButton` standard, and diverging would make this CTA taller than every other modal's — flagging it rather than special-casing it.

> **On how these were shot:** the account is a genuine self-custodial one (created locally, no OTP, no auth token), but a locally built app has no `BREEZ_API_KEY`, so the Spark SDK cannot connect and the self-custodial Receive screen renders "Wallet is offline" instead of the currency toggle. The modal is therefore shown mounted over the real non-custodial home screen — real component, real theme and typography, but not reached through its live trigger. The trigger conditions themselves are covered by the screen-level tests below rather than by these images.

## Scope

The issue body's Gherkin describes a *blocking* modal (Cancel / "I understand, continue", receive address withheld). [designsats' later comment](https://github.com/blinkbitcoin/blink-mobile/issues/3831#issuecomment-4670565111) supersedes it with an info-only popup — eye icon, "Privacy warning", a single "Acknowledged" button that closes it. This PR follows the design comment.

Also deliberately **not** included: the body's "revisit the warning later from settings" scenario. There is no designed placement or copy for it, so it is left for @designsats to spec rather than invented here.

## Implementation

- `app/store/persistent-state/usdb-privacy-warning-acknowledged.ts` — per-account flag keyed by `resolveAccountKey`, mirroring the existing `default-account-modal-shown.ts`. The field is optional on `PersistentState_14`, so there is no schema bump and no migration.
- `app/hooks/use-usdb-privacy-warning.ts` — `{ enabled } → { isVisible, acknowledge }`.
- `app/components/usdb-privacy-warning-modal/` — wraps `CustomModal`. Dismissing via backdrop or close icon also acknowledges; otherwise the modal would re-appear on every render of the same screen.
- Triggers in `receive-screen.tsx` and `conversion-details-screen.tsx`.

One design note worth a reviewer's attention: the Receive trigger keys off the **selected currency** (`amountRowCurrency === Usd`), not the toggle press. Hooking the press would miss users whose Receive screen opens in Dollar mode by default, since `useReceiveAssetMode` seeds from stable-balance-active plus the persisted default currency.

The Transfer trigger is direction-agnostic — both directions move value across the Dollar balance, so entering the flow is what warrants the warning. It sits after the existing `useDollarBalanceRestrictionGuard` / `useTransferBlockedGuard` early return, so a bounced user is never warned about a flow they cannot enter.

## Translations

English copy lives in `app/i18n/en/index.ts` (the source of truth). All 28 Crowdin-managed locale files under `app/i18n/raw-i18n/translations/` are also populated, so no locale falls back to English on release — verified by resolving `LL.UsdbPrivacyWarning.*` through `i18nObject` for every registered locale.

Each translation reuses the locale's own established terms for "non-custodial", "Dollar balance" and "Lightning address"; **Bitcoin**, **Spark** and **USDB** are left as proper nouns. The keys were inserted textually rather than by reserializing the JSON — these files are not uniformly formatted, so a rewrite would have touched every line instead of adding three. Crowdin remains free to revise the wording later.

## Testing

New unit tests: `__tests__/components/usdb-privacy-warning-modal/`, `__tests__/hooks/use-usdb-privacy-warning.spec.ts`, `__tests__/store/persistent-state/usdb-privacy-warning-acknowledged.spec.ts` — the copy, single-acknowledge behaviour, per-account isolation, and that an unacknowledged account still sees the warning when another has acknowledged.

New screen-level tests pinning the trigger wiring, which the unit tests above do not reach: `receive-screen-usdb-privacy-warning.spec.tsx` (armed on Dollar, disarmed on Bitcoin, never armed for a custodial account) and a `USDB privacy warning trigger` block in `conversion-details-screen.spec.tsx` (shown for self-custodial, absent for custodial). Each was verified against a deliberately broken trigger to confirm it fails rather than passing vacuously.

Ran locally: the new specs plus `receive-screen.spec.tsx`, `conversion-details-screen.spec.tsx`, `persistent-storage.spec.ts` and the persistent-state suite (153 tests), `tsc --noEmit`, eslint, and `check:translations`.

---

*Please do not delete the `pr-3831-assets` branch — the screenshots above are served from it.*


designsats:
We concluded it doesnt leak spark address so its not necessary.

grimen:
Per https://github.com/blinkbitcoin/blink-mobile/pull/4028#issuecomment-5091857491 this PR  and issue should be closed - it seems the decision was not recorded on the board and thus ticket was not already closed.

PR #4034, opened 2026-07-27
fix(self-custodial): note/label field in receive and send flows

grimen:
Fixes blinkbitcoin/blink-wip#949

## Problem

In non-custodial mode the note/label field was broken on both sides:

- **Send:** typing in the note field did nothing — it read as a disabled field.
- **Receive:** editing the note sometimes never regenerated the invoice; the edit was silently dropped.

Both work in custodial mode. The non-custodial flows are separate implementations that reproduce the shape of the custodial ones but not two mechanisms that make them correct.

## Receive — the edit was dropped while a generation was in flight

`usePaymentRequest` regenerated only when the `generateRequest` callback identity changed, and returned early whenever the request state was `Loading`. Spark invoice creation takes a second or two, so an edit made during that window hit the guard, was dropped, and nothing ever re-triggered it — the in-flight call finished with the old memo.

Generation is now identified by the sdk instance plus a key over the inputs the invoice is derived from (`type`, `memo`, `amount`, asset mode). The sdk is part of the identity because a reconnect or account switch hands back a new instance whose invoices belong to a new session. The key of the attempt is recorded before awaiting, so a failed attempt is not retried in a loop, and a change landing mid-flight leaves the key mismatched.

Picking that change back up is done by the settle handler that clears the busy flag, not by waiting for the state transition to re-run the effect — so the retry does not depend on React re-running that effect after the flag is cleared rather than before. `generateRequest` takes the generation its caller observed, so the key recorded and the memo the closure actually sends come from the same render.

Side effect: this fixes the same race for amount edits and BTC/USDB toggles, not just the memo.

## Send — the destination description kept winning

`NoteInput` on the details screen is controlled by `paymentDetail.memo`, which resolves as `destinationSpecifiedMemo || senderSpecifiedMemo`. The non-custodial LNURL details recreated themselves with only `senderSpecifiedMemo`, and a Lightning address always resolves with a non-empty `lnurlParams.description` — so the value fed back into the input never changed and typing appeared to do nothing. The LUD-12 comment carried the destination's description rather than the user's note for the same reason.

`setMemo` now overrides both memos, exactly as the custodial LNURL details do (`app/screens/send-bitcoin-screen/payment-details/lightning.ts`). Also aligned the spread order in `buildLnurlDetail` with `buildLightningDetail` so an explicit override can't be clobbered by `...params`.

One consequence, shared with custodial: once the note is edited the destination description is gone, so clearing the field sends no comment rather than falling back to it.

## Send — the comment is bounded by what the destination allows

`lnurlParams.commentAllowed` is a maximum length, not a flag, and the note field accepts 500 characters. While the comment carried the destination's own description an over-long value could never reach it; now that the user's note is what fills it, a note longer than the destination allows would have the pay request rejected outright. The comment is sliced to `commentAllowed`.

The custodial path has the same unsliced shape (`app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx`) and is left alone here as out of scope.

## Out of scope

BOLT11, Spark-address and on-chain non-custodial sends are unchanged. `PrepareSendPaymentRequest` in the Breez Spark SDK has no memo/description field, so a sender note cannot be transmitted for those types — it stays a local value. BOLT11 invoices that carry their own description keep the note field disabled, same as custodial.

## Screenshots

The user types "Dinner with Andrej" into the note field while paying the Lightning address `andrej@blink.sv`.

| Before | After |
| --- | --- |
|  |  |

Before, every keystroke is discarded and both the field and the LUD-12 comment stay pinned to the destination's own description. After, the typed note is what the field shows and what is sent as the comment.

Captured on an isolated simulator against a harness that renders the real `NoteInput` and the real `createSelfCustodialLnurlPaymentDetails` with a stubbed LNURL destination — self-custodial needs a provisioned Spark account, which is not reachable without a `BREEZ_API_KEY`. The readout lines under the field print `paymentDetail.memo` and the comment the LNURL send would carry. Only the fix itself differs between the two captures. The note is well under any advertised `commentAllowed`, so the slice added later does not change what these frames show.

The receive-side fix is a timing race — an edit landing while the SDK call is in flight — which a still screenshot cannot show. It is covered by the deterministic regression tests instead, three of which fail against the unpatched source.

## Tests

- `__tests__/self-custodial/hooks/use-payment-request-regeneration.spec.ts` (new): regeneration on memo and amount change, the mid-flight edit race for memo and amount and on the path where the in-flight call rejects, regeneration on a BTC→USDB toggle and when the sdk instance is replaced, no regeneration when nothing changed, once paid, while an auto-convert is settling, or on a PayCode round trip, no retry loop on failure, manual `regenerateInvoice` still forces a new invoice.
- `__tests__/self-custodial/payment-details/lnurl.spec.ts`: `setMemo` overrides a destination-specified memo, survives successive edits and clearing, the typed memo is what reaches `prepareLnurl` as the comment, and a note longer than `commentAllowed` is truncated to it.

Each new test was confirmed to fail against the source it covers. 209 tests across the touched modules and their consumer screens pass; `eslint` and `tsc --noEmit` are clean.

## Verification

Targeted run (the full suite is left to CI):

```
yarn jest --runTestsByPath \
  __tests__/self-custodial/hooks/use-payment-request.spec.ts \
  __tests__/self-custodial/hooks/use-payment-request-regeneration.spec.ts \
  __tests__/self-custodial/payment-details/lnurl.spec.ts \
  __tests__/self-custodial/payment-details/wrap-destination.spec.ts \
  __tests__/screens/receive-bitcoin-screen/receive-screen.spec.tsx \
  __tests__/screens/receive-self-custodial.spec.tsx \
  __tests__/screens/receive.spec.tsx \
  __tests__/receive-bitcoin/hooks/use-receive-flow.spec.ts \
  __tests__/screens/send-details.spec.tsx \
  __tests__/screens/send-confirmation.spec.tsx
```

The receive flow needs a provisioned Spark account to exercise on device, which is not reachable locally without a `BREEZ_API_KEY` — worth a pass from someone who has one.

### DIFF
--- a/__tests__/helpers/self-custodial-payment-request.ts
+++ b/__tests__/helpers/self-custodial-payment-request.ts
@@ -1,0 +1,72 @@
+import { WalletCurrency } from "@app/graphql/generated"
+
+/**
+ * Fixtures and default mock behaviour shared by the self-custodial receive specs.
+ *
+ * The `jest.mock` factories themselves have to stay in each spec file (babel hoists
+ * them above imports), but everything they resolve to lives here so the two suites
+ * cannot drift apart.
+ */
+
+export const btcWallet = {
+  id: "btc-w1",
+  walletCurrency: WalletCurrency.Btc,
+  balance: { amount: 1000, currency: WalletCurrency.Btc, currencyCode: "BTC" },
+  transactions: [],
+}
+
+export const usdWallet = {
+  id: "usd-w1",
+  walletCurrency: WalletCurrency.Usd,
+  balance: { amount: 500, currency: WalletCurrency.Usd, currencyCode: "USD" },
+  transactions: [],
+}
+
+export const mockSdk = { id: "mock-sdk" }
+
+export const btcAmount = (amount: number) => ({
+  amount,
+  currency: WalletCurrency.Btc,
+  currencyCode: "BTC",
+})
+
+export type PaymentRequestMocks = {
+  receiveLightning: jest.Mock
+  receiveOnchain: jest.Mock
+  selfCustodialWallet: jest.Mock
+  activeWallet: jest.Mock
+  convertMoneyAmount: jest.Mock
+  addPendingAutoConvert: jest.Mock
+  fetchAutoConvertMinSats: jest.Mock
+  useReceiveAssetMode: jest.Mock
+  formatMoneyAmount: jest.Mock
+}
+
+/** Puts every mock in the "self-custodial wallet is ready, BTC mode" baseline state. */
+export const applyPaymentRequestDefaults = (mocks: PaymentRequestMocks): void => {
+  mocks.selfCustodialWallet.mockReturnValue({
+    sdk: mockSdk,
+    lastReceivedPaymentId: null,
+  })
+  mocks.activeWallet.mockReturnValue({ wallets: [btcWallet, usdWallet], isReady: true })
+  mocks.receiveLightning.mockResolvedValue({ invoice: "lnbc1test..." })
+  mocks.receiveOnchain.mockResolvedValue({ address: "bc1qtest..." })
+  mocks.convertMoneyAmount.mockImplementation(
+    (amount: { amount: number }, currency: string) => ({
+      amount: amount.amount,
+      currency,
+      currencyCode: currency,
+    }),
+  )
+  mocks.formatMoneyAmount.mockImplementation(
+    ({ moneyAmount }: { moneyAmount: { amount: number } }) => `$${moneyAmount.amount}`,
+  )
+  mocks.addPendingAutoConvert.mockResolvedValue(undefined)
+  mocks.fetchAutoConvertMinSats.mockResolvedValue(undefined)
+  mocks.useReceiveAssetMode.mockReturnValue({
+    assetMode: "bitcoin",
+    setAssetMode: jest.fn(),
+    isToggleDisabled: false,
+    loading: false,
+  })
+}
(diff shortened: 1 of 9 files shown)


esaugomez31:
Pushed a follow-up in `ee707f83d` for a related gap surfaced while device-testing the send flow.

## Send — the sender's own note was missing from their self-custodial transaction history

After typing a note on an LNURL/Lightning-address send, the transaction detail in the sender's own history showed the destination's advertised text (e.g. *"Payment to esaudeveloper"*) rather than the note the user typed.

The payment itself was correct: the note is sliced to `commentAllowed` and sent as the LUD-12 comment, and the SDK persists it on the payment record in `lnurlPayInfo.comment`. The gap was only in the history mapper — `extractMemo` read `payment.details.inner.description`, which for an LNURL send is the invoice description supplied by the destination, not the sender's comment.

`extractMemo` now prefers `lnurlPayInfo.comment` and falls back to the invoice `description` when there is no comment. Since `getTransactionDescription` already prioritises `memo`, the sender's note now surfaces as the history title — matching custodial, where the sender's memo is what shows.

## Scope

- **Sender-side history only.** Whether the comment is rendered on the *recipient's* side is the receiving backend's concern (LUD-12 comment handling), not the app.
- The note only travels when the destination advertises `commentAllowed`; if it does not, no comment is sent and the mapper falls back to the invoice description, unchanged.

## Screenshots

The transaction history now reflects the note the sender typed instead of the destination's advertised description.

| Before | After |
| --- | --- |
| image | image|



## Tests

- `__tests__/self-custodial/mappers/transaction.spec.ts`: comment wins over the invoice description; falls back to the description when no comment is present. The preference case fails against the pre-fix source.
- Full mapper suite (`transaction`, `transaction-description`) green; `tsc --noEmit` clean.

esaugomez31 (approved):

PR #4046, opened 2026-07-28
fix(region): stop latching dollar-balance and transfer restrictions

### LINKED ISSUE #3924: feat(compliance): Region determination (aka Geo-blocking) — three-mode user journey (Custodial / NC Enhanced / NC Anon)
## Overview

|---|---|---|

Core principles:

## Why

## Scope — planned subtickets

## Open questions (from spec)

## References

esaugomez31:
## What

Region restrictions (dollar-balance and transfers) were persisted to a flag and read back on every launch, so a restriction determined once outlived the session that set it and never cleared. A user who moved out of a restricted region stayed restricted with no way back.

This makes both restrictions **live**: re-evaluated from the current country each session, with nothing persisted.

Dropping the latch opened a window the latch used to cover: on every cold start the country is briefly unknown, and reading that as "not restricted" showed a restricted user an unrestricted app. So the hooks also grew a second output, `isRegionPending`, and every gated surface holds on it.

### Removing the latch

- **Schema v17 migration** drops the four persisted latch flags (`stableTokenRestricted`, `stablesatsRestrictedCustodial`, `stableTokenTransferBlocked`, `stablesatsTransferBlocked`). Existing values are discarded, not carried forward.
- `useDollarBalanceRestricted` / `useTransferBlocked` now return only the live `isBlockedCountry(...)` check.
- Removed the write-only sync hooks (`useDollarBalanceRestrictionSync`, `useTransferBlockedSync`) and their mounts in the home screen.
- Removed the now-dead `dollarRestrictionCacheEnabled` remote-config switch and its four store helpers.

### Holding the pending window

`useDollarBalanceRestriction` and `useTransferBlock` return `{ verdict, isRegionPending }`. The verdict itself stays truthful: it still needs a resolved country, so an unresolved region never accuses an unrestricted user. Failing closed instead regressed the surfaces that *seed* state from it, latching Bitcoin for a user who turns out to be unrestricted. Holding latches nothing; it just defers.

Surfaces that hold while the region resolves:

| Surface | Hold |
|---|---|
| `WalletOverview` | Dollar row keeps its loader and stays inert: no balance, no restriction label, no explanation tap |
| `HomeScreen` | Transfer button renders disabled, so a fast tap cannot reach the gated flow |
| Both navigation guards | Screen hides while pending, but only resets once the verdict lands |
| `useReceiveAssetMode` | Reports `loading`, so no Dollar request is issued before the verdict |
| `useTotalBalance` | Folds into the header's `isLoading`, and counts the dollars out of `satsBalance` (read by the backup nudge without consulting `isLoading`) |
| `account-default-wallet` | Row renders in its loading state instead of appearing and then vanishing |
| `use-stable-balance-toggle` | Activation refused and the switch snaps back. No toast: the region has not accused the user yet |
| `use-send-wallets` | List holds instead of offering a dollar source that is about to disappear |
| `use-migration-balances-preview` | Screen holds its spinner instead of promising a Dollar Balance the new account cannot hold |

`receive-screen.tsx` / `use-payment-request.ts` were already covered: the PRCD layout effect returns early on `useDeviceLocation().loading`, which is stricter than `isRegionPending` (it holds even once a country has resolved), and until the PRCD exists the screen renders its loading state. Pinned with a test rather than left incidental.

Home's transfer-button **visibility** deliberately does not hold. Holding would hide the button on every cold start and pop it into the row once the verdict lands, a layout shift for everyone, to spare the narrow set of users in a transfer-blocked but dollar-allowed country a button that disappears. The button is already inert while pending, so nothing is reachable either way.

### Settling the self-custodial prediction

`useIpCountryLookup(enabled)` returns `{ countryCode, isSettled }`, where `isSettled` is true once waiting is pointless: the lookup finished with or without a country, or it is disabled and will never run. `useIpCountryCode` stays as a thin wrapper so its other callers are untouched.

`useRestrictionRegion` folds `isSelfCustodialPrediction && !isIpLookupSettled` into the pending term, so a fast phone parse no longer reports settled-unrestricted and then flips once the IP lands. The custodial path is unaffected, since it never waits on the lookup.

## Behavior / impact

The live block logic, country lists, and account-type selection are unchanged. Only the persistence is removed:

| User | Before | After |
|---|---|---|
| Still in a blocked country | Restricted | Restricted (live check) |
| Moved out of a blocked country | Stuck restricted | **Freed on next launch** |
| Never restricted | Unchanged | Unchanged |
| Cold start, region unresolved | Unrestricted UI, then flip | **Held until the verdict lands** |

Nobody who should be blocked gets unblocked; only stale, no-longer-applicable restrictions clear.

## Testing

- `tsc:check` clean, `eslint` clean
- Pending-window specs on both hooks, both guards, `WalletOverview`, `HomeScreen`, `useReceiveAssetMode`, `useTotalBalance`, `use-send-wallets`, `use-stable-balance-toggle`, `account-default-wallet` (which had no spec at all) and the migration balances overview: each holds while pending and settles correctly in both directions
- `useIpCountryLookup` covered directly, including that it reports settled while disabled and drops a resolved country the moment it is disabled
- Schema migration, both restriction hooks and the home screen green

## Notes

- The `dollarRestrictionCacheEnabled` Firebase Remote Config key is now unused and can be retired.
- Live re-check on travel via IP (cache invalidation) lands in a follow-up branch.


grimen (changes requested):
## Findings

**1. Restricted users see an unrestricted home screen during the launch window.**
`useDeviceLocation` returns `countryCode: undefined` while the settings query and phone parse are still resolving, and both hooks now read that as "not restricted" — the state the persisted latch used to cover at launch. The navigation guards self-heal (their effects reset to Primary once the flag flips), but the home UI briefly shows a visible dollar balance and an enabled transfer button, and a fast tap can enter a gated flow before being bounced.
*Example:* on a cold start in Hong Kong, the transfer button renders enabled for the first render(s), then flips once the country resolves.
*Proposal:* `useDeviceLocation` already exposes `loading`, so one option is to return a tri-state (or treat loading as restricted-pending) so callers can hold the gated UI until the country is known — possibly fine to defer, if the flash is deemed acceptable.

esaugomez31:
@grimen

## Findings

**1. Restricted users see an unrestricted home screen during the launch window.** — fixed in 2726155.

Both hooks now expose `isRegionPending` (from `useDeviceLocation().loading`) alongside the verdict, and the gated surfaces hold on it:

- `WalletOverview`: the dollar row keeps its loader and stays inert — no balance, no restriction label, no explanation tap.
- `HomeScreen`: the transfer button renders disabled, so a fast tap cannot reach the gated flow.
- Both navigation guards hide the screen while pending, but only reset once the verdict lands.
- `useReceiveAssetMode` reports `loading`, so no Dollar request is issued before the verdict.

The verdict itself stays truthful — it still needs a resolved country, so an unresolved region never accuses an unrestricted user. Failing closed instead regressed two surfaces that seed state from it (`useReceiveAssetMode`'s initial mode and the send screen's `paymentDetail`), latching Bitcoin for a user who turns out to be unrestricted.

## Test Coverage

Pending-window specs on both hooks, both guards, `WalletOverview`, `HomeScreen` and `useReceiveAssetMode`: holds while pending, and settles correctly in both directions once the region resolves.

grimen (changes requested):
## Findings

**1. The custodial receive path got neither the pending hold nor any stated rationale, so a restricted user can be issued a dollar payment request during the launch window.**
The 2726155 rationale for keeping the verdict fail-open is sound for surfaces that *seed state* from it — failing closed would latch Bitcoin for a user who turns out to be unrestricted. But holding on `isRegionPending`, as `WalletOverview`, the guards, and `useReceiveAssetMode` now do, latches nothing; it just defers. Several boolean call sites received neither treatment: `receive-screen.tsx` / `use-payment-request.ts` (custodial receive), `use-stable-balance-toggle.ts`, `use-total-balance.ts`, `account-default-wallet.tsx` — all still call `useDollarBalanceRestricted()`, which returns `false` while `useDeviceLocation` is loading. With the latch dropped, that window now exists on every cold start rather than only the first-ever session.

*Example:* a restricted custodial user cold-starts straight into the receive screen and generates a dollar payment request before the region resolves — the exact hole `useReceiveAssetMode.loading` closes for the self-custodial side.

*Proposal:* one option is to gate request generation in the custodial receive flow on `isRegionPending` the same way; for the display-only surfaces (balance header, settings row), a comment noting the flash is accepted would help.

**2. `isRegionPending` does not track the in-flight IP lookup in the self-custodial prediction path.**
In `useRestrictionRegion`, `isPending = isDeviceLocationLoading && !countryCode`, but `useIpCountryCode` exposes no loading state. When the device country settles first (phone parse is fast) to an unblocked country while the IP lookup is still in flight, the prediction reports settled-unrestricted, then flips to restricted once the IP lands.

*Example:* during migration, the balances preview briefly shows a dollar balance the self-custodial account won't be able to hold, then flips — the same class of transient 2726155 set out to eliminate on the home screen.

*Proposal:* one option is to extend `useIpCountryCode` with a `loading` flag and fold it into `isPending` for the prediction path; possibly fine to defer given the documented fallback-to-session-country behavior.

## Test Coverage Gaps

- No test asserts the custodial receive flow cannot issue a dollar request during the pending window; `use-receive-asset-mode.spec.ts` gained exactly this test for the self-custodial side, and finding 1 hides in the untested custodial mirror.
- `use-send-wallets.spec.ts` and `use-stable-balance-toggle.spec.ts` stub `isRegionPending: false` only, baking in the assumption that a pending region never reaches those hooks.

esaugomez31:
@grimen

## Findings

**1. The custodial receive path got neither the pending hold nor any stated rationale.**
Split by call site, because they did not all turn out the same.

`receive-screen.tsx` / `use-payment-request.ts` were already covered. The PRCD layout effect returns early on `useDeviceLocation().loading`, which is stricter than `isRegionPending` (it holds even once a country has resolved), and until the PRCD exists the screen renders its loading state, so the wallet toggle is not reachable either. The cold-start example does not reproduce against the current code. `de717d8a5` adds the test that was missing and it passes against the existing gate, so the behaviour is now pinned rather than incidental.

`use-stable-balance-toggle.ts` was the real hole, fixed in `7c58dc09e`. Activation is refused while the region is pending and the switch snaps back. Deliberately no toast: the region has not accused the user yet, and the restricted copy would claim a verdict that may not hold. Deactivation stays allowed, as before.

`use-total-balance.ts` and `account-default-wallet.tsx` hold rather than accept the flash, also in `7c58dc09e`. The pending region folds into the balance header's existing `isLoading`, so it stays on its loader instead of showing a dollar total that flips to zero; the settings row renders in its loading state until the verdict lands, instead of appearing and then vanishing.

**2. `isRegionPending` does not track the in-flight IP lookup in the self-custodial prediction path.**
Fixed in `a976346c0`. `useIpCountryLookup(enabled)` now returns `{ countryCode, isSettled }`, where `isSettled` is true once waiting is pointless: the lookup finished with or without a country, or it is disabled and will never run. `useIpCountryCode` stays as a thin wrapper so its other callers are untouched. `useRestrictionRegion` adds `isSelfCustodialPrediction && !isIpLookupSettled` to the pending term, so a fast phone parse no longer reports settled-unrestricted and then flips once the IP lands. The custodial path is unaffected, since it never waits on the lookup.

## Test Coverage Gaps

- Custodial receive through the pending window is covered in `de717d8a5`: the test drives a USD default wallet with the region unresolved, asserts no request is created, then settles restricted and asserts the request comes back on bitcoin.
- `use-send-wallets.spec.ts` and `use-stable-balance-toggle.spec.ts` no longer stub `isRegionPending: false` only. Both now drive the pending window, and the stable-balance spec asserts the refusal, the absence of a toast, that deactivation still passes, and that a settled unrestricted region activates normally.
- `use-dollar-balance-restricted.spec.ts` gained a block for the in-flight lookup: the prediction holds while the device country has already resolved, settles restricted when the IP lands on a blocked country, settles on the session country when the lookup finishes without one, and leaves the custodial evaluation settled throughout.
- `useIpCountryLookup` is covered directly in `use-device-location.spec.ts`, including that it reports settled while disabled and drops a resolved country the moment it is disabled.
- `account-default-wallet.tsx` had no spec at all; one was added for the row's hold, its settled render, and the restricted case.

## Additional Findings

**1. The send wallet list kept the same hole once the sibling surfaces were fixed.**
Not in the call-site list above, but holding the other three made it asymmetric: `use-send-wallets.ts` kept offering the dollar wallet through the pending window, while the settled-restricted state removes that wallet from the list entirely.

*Example:* a restricted user cold-starts into send, sees the dollar wallet offered as a source, and it disappears once the region resolves.

*Resolution:* `08299d707` folds `isRegionPending` into the hook's existing `loading` on both the self-custodial and the custodial branch, so the list holds instead of offering a source that is about to vanish. The three surfaces from finding 1 and this one now behave identically through the pending window.

esaugomez31:
@grimen

You were right that the flags were still catching something. I re-verified every prior finding against the branch rather than against my own replies, and one was genuinely half-addressed.

## Findings

**1. Finding 2 was fixed in the hook but never reached the only surface that consumes the prediction.**

The `a976346c0` work was correct as far as it went. What it did not do was reach `use-migration-balances-preview.ts`, the only caller of the self-custodial prediction in the app: it still read `useDollarBalanceRestricted(AccountType.SelfCustodial)` and never looked at the pending flag, and `isReady` did not fold it in either. So the migration balances screen could render the new Dollar Balance at `$0.00` and swap it for "not available" once the IP landed. That is the exact example you gave, still live.

Worse, the spec baked in the assumption you objected to elsewhere: `balances-overview-screen.spec.tsx` stubbed `isRegionPending: false` outright.

*Resolution:* `2a8df0c7d` moves both sides to `useDollarBalanceRestriction` and folds the combined pending into `isLoading` and `isReady`. The screen holds its spinner instead of promising a balance the new account cannot hold, in the one step the user cannot take back. A pending region does not count as a settled-empty source, so it never reaches the support handover.

## Test Coverage Gaps

- `balances-overview-screen.spec.tsx` no longer hardcodes `isRegionPending: false`. It drives the pending window in its own describe: the spinner holds with Approve disabled and no figures, no support handover fires, and the restricted label appears once the region settles.

## Additional Findings

**1. The balance header held, but `satsBalance` did not, and one consumer never reads `isLoading`.**

The `7c58dc09e` hold covered the loader; the figures behind it were still computed with the dollar balance included. `use-backup-nudge-state.ts` reads `satsBalance` straight against its thresholds without consulting `isLoading`, so a restricted self-custodial user could have the backup modal armed on a total that drops below the threshold a moment later.

*Resolution:* `fb8e2d36d` counts the dollars out of every figure while the region is unresolved, not just out of the loader.

**2. The transfer-blocked verdict did not hold, unlike the dollar one on the same screen.**

`home-screen.tsx` read the plain `useTransferBlocked()` for the button's visibility. While the region is unresolved that reads `false`, so the button renders (disabled), then disappears once the region settles to a transfer-blocked but dollar-allowed country. The last surface where a region verdict did not hold.

*Resolution:* `51b0a0646` switches to `useTransferGate()`, which already exposes `isRegionPending`, and folds it into visibility. The hold costs a frame in practice: the settings query behind the country is cache-first and the phone parse is synchronous.

## Verification

`tsc:check` and `eslint` clean; the affected suites green. The three fixes are restacked through `#4051`, `#4054` and `#4055`, where the anon-mode rename (`useTransferGate` / `useDollarBalanceGate`) absorbs them.

One deliberate exception, worth stating rather than leaving for the flags: in `#4054` the balance header keeps counting held dollars regardless of the verdict ("gates limit actions, not existence"). That makes the `satsBalance` fix moot upstack by construction rather than contradicting it, since the total no longer moves when the verdict lands. The spec there pins that instead.

grimen (changes requested):
## Findings

**1. The pending hold blanks the whole balance header, not just the dollar row (blocking).**

`use-total-balance.ts:46` folds `isRegionPending` into `isLoading`, and `home-screen.tsx:337` passes that same `loading` on to `BalanceHeader`, the username block and `WalletOverview` — whose Bitcoin row renders on it. Since a self-custodial user has no phone number, `useDeviceLocation` falls back to the IP branch, which can take around 20 seconds across four adapters at a 5s timeout each.

*Example:* On a cold start with a slow network, a self-custodial user sees their total balance, username and BTC balance all replaced by loaders for up to ~20s, even though only the dollar figure depends on the region.

*Proposal:* One option could be to hold just the dollar figure through `isDollarBalanceUnavailable` and leave `isLoading` tied to price conversion.

**2. The send-wallets pending flag has no consumer (blocking).**

`use-send-wallets.ts:83` returns `loading: isRegionPending`, but the only consumer, `send-bitcoin-details-screen.tsx:130`, destructures `wallets`, `defaultWallet`, `btcWallet`, `usdWallet`, `network` and `isSelfCustodial` and has no loading branch. The wallet list therefore does not hold during the pending window, contrary to the PR description.

*Example:* A restricted user opens Send while the region is still resolving; the seeding effect at `:203` builds `paymentDetail` from the USD `defaultWallet` and then returns early on every later run, so when the verdict lands restricted and `restrictToBitcoinWhenBlocked` drops `usdWallet`, the payment detail is still pointing at the dollar wallet.

*Proposal:* One option could be to apply the gate at the screen, so the details screen holds rendering until the hook reports a resolved region.

**3. The stable-balance toggle refuses silently while the region is pending.**

`use-stable-balance-toggle.ts:65` uses `isActivationUnresolved` to snap the switch back, but `stable-balance-settings-screen.tsx:117` disables the control only on `busy || !sdk`, so there is no spinner, toast or disabled state. This screen is self-custodial-only, which is exactly the group whose region resolves over the network.

*Example:* In the first seconds after opening the screen, a user taps "activate", watches the switch flip on and immediately flip back, with no indication of why.

*Proposal:* Surfacing pending as `disabled` or `busy` instead of a silent refusal would be one way to make the wait legible.

**4. The migration balances screen waits on the IP lookup.**

`use-migration-balances-preview.ts:83` includes the `AccountType.SelfCustodial` prediction in `isDollarRegionPending`, and that prediction waits on `isIpLookupSettled`, i.e. on `resolveIpCountryCode` walking up to four adapters. Previously this screen rendered immediately from the still-custodial session's phone country.

*Example:* On a slow or rate-limited network, the balances preview for an irreversible migration step shows a spinner for up to ~20s where it used to render at once.

*Proposal:* One option could be to hold only the dollar row rather than the whole screen, or to bound the IP lookup with a shorter overall deadline.

**5. Pending and restricted collapse into one boolean, so callers render a blank screen.**

`use-dollar-balance-restriction-guard.ts:29` (and the same shape in `use-transfer-blocked-guard.ts:29`) returns `true` while the region is pending, which makes `conversion-details-screen.tsx:103` `return null`. That is an empty screen under the header rather than a loading state, and the screen is reachable by routes other than the home transfer button, which does hide itself while pending.

*Example:* A user deep-links or navigates into conversion details during the pending window and sits on a blank area with no spinner and no explanation until the region resolves.

*Proposal:* Separating the two signals would let each caller render a loader for pending and `null` only for a resolved restriction.

## Test Coverage Gaps

- The new `use-send-wallets` specs assert on the hook's `loading` flag, which no consumer reads; a screen-level test that mounts `send-bitcoin-details-screen` during the pending window would catch the missing gate.
- No test covers the seeded-`paymentDetail` path where the region verdict lands restricted after `defaultWallet` was already USD.
- The pending branches of `use-dollar-balance-restriction-guard` and `use-transfer-blocked-guard` are not distinguished from the restricted branches in the specs, so the blank-screen behaviour in finding 5 is untested.

esaugomez31:
## Resolutions

**1. The pending hold blanks the whole balance header — fixed.**
`isLoading` tracks only the price conversion again, so the total, the username and the Bitcoin row render at once. The dollar figure still waits, on its own: `WalletOverview` already loads that row off `isRegionPending`. `satsBalance` is unchanged, so the backup nudge still reads a steady total.
*Commit:* `8bd829971`

**2. The send-wallets pending flag has no consumer — fixed.**
The details screen now reads `loading`, and the seeding effect waits on it. That is the part that mattered: the effect runs once, so seeding early left the payment pointing at the dollar wallet `restrictToBitcoinWhenBlocked` had just dropped. It holds with a loader rather than a blank screen.
*Commit:* `28447162c`

**3. The stable-balance toggle refuses silently — fixed.**
The hook reports `isRegionPending`; the screen disables the switch and shows its spinner on it. The refusal stays silent by design, since the region has not accused the user yet: what changed is that the wait is legible instead of a switch that flips on and snaps back.
*Commit:* `503157874`

**4. The migration balances screen waits on the IP lookup — fixed.**
The region no longer gates the screen, so the bitcoin figures render immediately and only the dollar rows wait, through a new `isDollarValuePending` on `BalancePairCard`. Approve stays disabled until the verdict lands, so rendering sooner never means approving sooner. The checkpoint gate is untouched.
*Commit:* `e1c1a6429`

**5. Pending and restricted collapse into one boolean — fixed.**
Both guards return two signals instead of one. `conversion-details-screen` and `default-wallet` now render a loader for the wait and `null` only for a resolved refusal, which is what the deep-linked user saw as a blank area.
*Commit:* `04b3b2704`

## Test Coverage Gaps

- `send-bitcoin-details-screen` is now mounted during the pending window: it asserts the loader and that no sending wallet is chosen yet.
- The seeded-`paymentDetail` path is covered: rendered pending with the dollar default, then re-rendered restricted, asserting the payment is built once and never from the withdrawn wallet.
- The guards' pending branches are distinguished from their refused branches, in the hook specs and at both call sites.

## Notes

- Every fix is mutation-tested: reverting each one fails the tests written for it.
- `DefaultWalletScreen` had no spec; one was added for its three branches.
- Not device-tested: the stable-balance and migration screens are gated for a restricted account and mid-migration respectively. Home, Settings and the Send entry were exercised on device and are unaffected.
- Rebased onto `8e01f5597`. #4054 renames these hooks for Anon Mode (`useDollarBalanceGate`, `isGated`), and the guards adopt that vocabulary as the stack replays.

grimen (approved):

PR #4047, opened 2026-07-28
feat(learn): self-custodial version of l-earn without rewards

kngako:
The change to add the learn function for self custodial has been split in two. First step will be to make the learn content available without giving rewards (as the content is fully local on the users device). This change also tracks the quizzes the user completed so far.

This change will be followed up by a PR to add rewards to the with anti-abuse functionality. Why may enrich the data stored when a user completes a quiz to ensure anti-abuse/cheats work well.

image

As it currently stands a user with self custodial wallet well see the above screenshot. With a distrintive "Dig your way through these lessons and learn about Bitcoin".

While user who still access through custodial wallet will see the below message "Dig your way through these lessons and earn as you learn".

image


esaugomez31 (commented):
## Review: findings by severity

Reviewed at `260bb8121`. I checked out the branch, ran the 7 new/updated test suites locally (66/66 passing), typechecked, and traced the custodial call sites touched by the diff to confirm existing behavior is preserved.

### High

None found.

### Medium

None found.

### Low

1. **`useQuizServer` rebuilds the self-custodial quiz list on every render** (`app/screens/earns-map-screen/use-quiz-server.tsx:51`). The `allQuizIds.map(...)` runs per render and `completedQuizIds.includes(id)` inside it makes the build O(quizzes x completions). Wrapping the branch in `useMemo` keyed on `completedQuizIds` (with a `Set` for the lookup) would keep the array identity stable and drop the repeated scans. Cosmetic at roughly 120 quizzes, and the pre-existing signed-out branch behaves the same way, hence Low.

2. **Multi-line `//` comment clusters would read better as single-paragraph `/** ... */` blocks**, which is the style the new files themselves use everywhere else: `app/screens/earns-map-screen/earns-map-screen.tsx:99`, `app/screens/earns-screen/earns-section.tsx:215`, and the level-one note in the `beforeEach` of `earns-self-custodial-learn.spec.tsx`.

3. **The screen spec mocks both collaborating hooks, so the full completion loop never runs unmocked.** `earns-self-custodial-learn.spec.tsx` stubs `useQuizServer` and `useQuizCompletions` together, so "answer correctly, persist, completed state re-renders" is only asserted with hand-injected values. Every pairwise seam is covered (store spec, hook spec, hook-consumer specs), so the residual risk is just the context re-render wiring; one `EarnQuiz` test against the real persistent-state provider would close the gap.

4. **`jest.config.js` now pins `process.env.NODE_ENV = "test"` for every suite in the repo.** The inline justification is solid (a shell exporting `NODE_ENV=development` makes the gesture-handler dev guard throw locally while CI stays green) and it ships as its own commit, but it is repo-wide test infra riding in a feature PR, so flagging it for visibility.

### Non-blocking notes

- Quiz progress is device-only: lost on reinstall and not part of any backup. That matches the staged plan in the PR description (the rewards follow-up enriches the stored record), just making the trade-off explicit.
- Verified along the way: the v14 to v15 migration is additive with the whole chain and specs updated (no stale `schemaVersion: 14` left anywhere); per-account isolation goes through the shared `resolveAccountKey` primitive with dedupe and immutability tested; `AccountRegistryProvider` seeds the active self-custodial id synchronously, so there is no first-frame flash of the custodial gate on the Earn tab; all 28 translation files received real per-language translations with correct diacritics and preserved indentation; no duplicate quiz ids across sections, so the synthetic list is sound; and custodial behavior is unchanged at every touched call site (query skip condition, `isAvailable` stays true, level-one phone gate, badger copy).

esaugomez31:
Pushed three commits on top of this branch today, plus a rebase onto main:

**Review feedback applied**

- f676b0826 — `useQuizServer` now builds the self-custodial quiz list inside `useMemo` with a `Set` lookup, and `getCompletedQuizIds` returns a shared module-level empty array so the memo can actually key on a stable identity. Also normalized the remaining `//` comment clusters to single-paragraph `/** ... */` blocks.
- c242a7e27 — added an integration spec (`earns-self-custodial-quiz-completion.spec.tsx`) that renders `EarnQuiz` against the real `PersistentStateProvider` with `useQuizServer`/`useQuizCompletions` unmocked: it answers correctly, persists through the provider, and asserts the screen re-renders as completed without calling `quizClaim`. `findPressableParent` moved to the shared test helper since three specs now use it.

**Rebase onto main** — clean, no conflicts. It also picks up the `@blinkbitcoin/react-native-geetest-module` switch from #4011.

**Navigation bug found while device-testing this branch** (bdd312ed0)

After completing a section, the `sectionCompleted` screen's "Keep digging!" button and close cross did nothing. Both called `navigation.navigate("Earn")`, but "Earn" is a tab inside the nested Primary tab navigator, so the action is unresolvable from the root stack under React Navigation v7 (navigate only searches the current navigator and its ancestors, and `navigate("Primary")` would push a fresh tab navigator that lands on Home). Fixed with `navigation.popTo("Primary")`, which returns to the existing Primary entry with the Earn tab still focused. Same fix applied to the quiz error-modal close in `earns-quiz.tsx`.

Note this bug is pre-existing on main and affects custodial accounts too — the screen and handlers are unchanged there — but self-custodial users are the first to hit it consistently now that quiz completion works locally. Coverage: new `section-completed.spec.tsx` (100% on the file) plus an error-modal close test in `earns-quiz.spec.tsx`.

Verification after all of the above: `yarn tsc:check` and eslint clean, all touched suites green.

esaugomez31 (approved):
LGTM

PR #4048, opened 2026-07-28
fix(test): run the number-pad specs as several device locales

kngako:
formatNumberPadNumber formats the amount being typed with a bare Number#toLocaleString() on purpose: the number pad groups digits the way the user's own device does, while converted amounts pin "en-US" through formatCurrencyHelper. Which locale that bare call resolves to is decided by the runtime, so specs hardcoding "1,500" were really asserting the locale of whichever machine ran them, and went red on a shell set to en_ZA (which groups with a no-break space).

withDeviceLocale substitutes that default for a describe block, so a spec states the locale it means instead of inheriting one, and can run the same code as several devices. Calls that name a locale pass through untouched, which is what lets a spec tell the two formatters apart.

  - number-pad-reducer: en-US/de-DE/hi-IN assert literals, hi-IN grouping in lakhs so a swapped separator that kept en-US's pattern still fails; fr-FR and en-ZA read the separator from the same ICU data instead of hardcoding it, since CLDR moves those (fr-FR went U+00A0 -> U+202F, en-ZA is still on U+00A0)
  - conversion-details-screen: typing on a de-DE/hi-IN device pins down the split that is invisible on en-US — the focused field follows the device, the converted field stays "en-US"

Both suites pass under LC_ALL of en_ZA, de_DE, fr_FR and C. Pinning "en-US" in formatNumberPadNumber — the change this replaces — fails 8 reducer tests and both screen tests, so the new coverage bites.

### DIFF
--- a/__tests__/components/number-pad-reducer.spec.ts
+++ b/__tests__/components/number-pad-reducer.spec.ts
@@ -1,5 +1,7 @@
-// The @wdio/mocha-framework types in tsconfig.jest.json shadow jest's, and
-// mocha's `describe` has no `.each` — same import the screen specs use.
+/**
+ * The @wdio/mocha-framework types in tsconfig.jest.json shadow jest's, and
+ * mocha's `describe` has no `.each`; same import the screen specs use.
+ */
 import { describe } from "@jest/globals"
 
 import { WalletCurrency } from "@app/graphql/generated"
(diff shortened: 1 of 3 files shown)


esaugomez31 (approved):
LGTM!

PR #4049, opened 2026-07-28
refactor(schema-version): Introduce CURRENT_SCHEMA_VERSION instead of MAGIC Number.

kngako:
The current schema version was spelled out as a literal in 25 test files, so every bump meant editing all of them. Nothing forced them to keep up, and several had already drifted: mocked persistent states were still claiming schema 8, 9, and 12 while the real one was 14.

Export the version once, typed as PersistentState["schemaVersion"], so moving the PersistentState alias without bumping the constant is a compile error. Tests keep their own literal fixtures and take only the number from it. The jest.mock factories reach it via requireActual, since a hoisted factory cannot safely close over a module-scope import.

Also replaces the hand-written 3 | 4 | ... | 14 union in migratePersistentState with keyof StateMigrations, leaving the migrations map as the only list to maintain.

Historical inputs in the migration specs stay literal — those versions are the thing under test.

### DIFF
--- a/__tests__/hooks/use-default-account-modal-shown.spec.ts
+++ b/__tests__/hooks/use-default-account-modal-shown.spec.ts
@@ -1,7 +1,10 @@
 import { act, renderHook } from "@testing-library/react-native"
 
 import { useDefaultAccountModalShown } from "@app/hooks/use-default-account-modal-shown"
-import { PersistentState } from "@app/store/persistent-state/state-migrations"
+import {
+  CURRENT_SCHEMA_VERSION,
+  PersistentState,
+} from "@app/store/persistent-state/state-migrations"
 
 const mockUpdateState = jest.fn()
 let mockPersistentState: PersistentState
@@ -14,7 +17,7 @@ jest.mock("@app/store/persistent-state", () => ({
 }))
 
 const baseState: PersistentState = {
-  schemaVersion: 14,
+  schemaVersion: CURRENT_SCHEMA_VERSION,
   galoyInstance: { id: "Main" },
   galoyAuthToken: "",
 }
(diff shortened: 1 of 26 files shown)


grimen (changes requested):
## Findings

**1. Nothing ties `CURRENT_SCHEMA_VERSION` to the newest registered migration (blocking).**
The type annotation only fires when the `PersistentState` alias moves, so the guard covers one of the two ways a bump can go wrong.

*Example:* Someone adds `PersistentState_15`, writes `migrate14ToCurrent` to emit 15, registers key `15` in `StateMigrations`, but leaves `export type PersistentState = PersistentState_14` — everything compiles, `CURRENT_SCHEMA_VERSION` stays 14, and `defaultPersistentState` keeps writing the old version.

*Proposal:* One option could be exporting `stateMigrations` (or a derived max key) and asserting in `state-migrations.spec.ts` that the highest key equals `CURRENT_SCHEMA_VERSION`, which closes the direction the compiler can't see.

**2. The doc comment claims more than the file delivers.**
It says nothing "app code or tests" should spell the number out, but the migration chain a few lines above still does.

*Example:* `migrate13ToCurrent` is `migrate14ToCurrent({ ...state, schemaVersion: 14 })` — a literal 14 in the same module, and correctly so, since each step targets a specific version rather than "current".

*Proposal:* Softening the wording to something like "outside the migration chain, nothing should spell the number out" would keep the comment true as versions climb.

**3. `wallet.spec.tsx`'s mock factory gained a real dependency it didn't have.**
That factory previously replaced `@app/store/persistent-state` wholesale with no `requireActual` at all; it now loads the real `state-migrations` module just to read one number, and that module imports `@app/config`.

*Example:* If `@app/config` later grows a native or env-dependent import, this spec starts failing at module load for a reason unrelated to what it tests — the other five factories already pulled in the real module, so only this one changes shape.

*Proposal:* If the constant lived in a tiny leaf module with no imports, every factory could read it via `requireActual` without dragging config in; alternatively this one spec could keep its literal, since its assertion doesn't depend on the version.

## Test Coverage Gaps

- No test asserts that `CURRENT_SCHEMA_VERSION` matches the highest key in `stateMigrations` (see finding 1) — the only protection is the type annotation, which misses the forgot-to-move-the-alias case.

## Code Smells

- The `jest.mock("@app/store/persistent-state", () => { const { CURRENT_SCHEMA_VERSION } = jest.requireActual<...>(...) ... })` block is byte-identical across six spec files; a shared helper returning the mock object would keep the next schema-shaped change to one edit.

PR #4051, opened 2026-07-29
feat(region): add self-custodial mode-selection screen

### LINKED ISSUE #3924: feat(compliance): Region determination (aka Geo-blocking) — three-mode user journey (Custodial / NC Enhanced / NC Anon)

esaugomez31:
## What

Adds the **"Select non-custodial mode"** screen where a self-custodial user picks their region posture — **Enhanced Mode** (we use your region to unlock valuable services available where you are) or **Incognito Mode** (maximum privacy, fewer features) — and persists the choice per account.

> Stacked on #4046 — review/merge that first.

## Where it appears

One screen serves three onboarding flows, forwarding to a caller-provided destination:

- **Creation:** `AccountTypeSelection → ChooseExperience → terms → WalletCreation` (before terms; the mode rides through terms to wallet creation, which stores it against the provisioned account).
- **Restore:** after the wallet is restored, before the success screen.
- **Migration:** after backup, before the balances overview.

## Per-account mode persistence

- New `persistentState.selfCustodialAccountModeByAccountId[id]`, schema bumped to **v18**.
- Stored against an explicit account id (like the mnemonic / backup state), so a migration that provisions an account while custodial is still active still records the right account's mode.
- **Storage-ready, no consumer yet** — nothing reads the mode. Server-side mode/region persistence (LNURL server) stays future work per the spec.
- Extracted the shared `resolveActiveSelfCustodialId` helper reused by the existing currency/language stores and the new mode store.

## Design review follow-ups

- Screen carries a **back arrow** (route shows the header with an empty title, matching its sibling onboarding screens). The swipe-back gesture stays blocked, so the arrow is the one deliberate way out.
- Card icons render at 22px rather than the group's 20px default, applied per card so `AccountTypeSelection` keeps the default.

## Reuse

- New reusable `OptionCardGroup` single-select card component; `AccountTypeSelection` now renders its Custodial / Non-custodial cards through it (UI unchanged — its existing tests pass untouched).

## Namespace

- `useSelfCustodialAccountMode` lives in `app/self-custodial/hooks/`, not `app/hooks/`: the Anon/Enhanced mode exists only for self-custodial accounts, so it belongs in that namespace rather than the universal one.

## Migration resume

- After backup, the migration checkpoint is `ChooseExperience`, so an app-kill on the mode screen resumes **onto the mode screen** (the choice is still captured) without re-running the backup. Once the user continues to the balances overview, that screen commits its own checkpoint, so a later kill there resumes at the balances overview.

## i18n

- New `ChooseExperienceScreen` keys across all 28 languages, with real per-language diacritics. `Enhanced` / `Incognito` are kept as product mode names (per the PRD); only the "Mode" descriptor is localized.

## Testing

- Typecheck + lint clean on touched files.
- Full migration + self-custodial onboarding suites green (885 tests), plus new specs for `OptionCardGroup`, the mode store/hook, and the checkpoint resume path.


kngako:
Enhanced Mode Setup

image


image


View terms and prohibited countries

image


image

====

I believe it should be possible to 

1. See the setting the self custodial wallet is currently in.
2. Update the active setting to an alternative from somewhere in settings.

Not sure if this is planned in a future PR to keep this PR concise?

grimen (changes requested):
## Findings

**1. Re-entering the mode screen ignores an already-stored mode — for the implementer to confirm/decide.**
`ChooseExperienceScreen` always initializes `selected` to `AccountMode.Enhanced` and never reads `selfCustodialAccountModeByAccountId`, even when `onContinue` carries an `accountId` that already has a mode recorded.

*Example:* A user restores a wallet and picks Anon; later they re-run the restore flow with the same mnemonic — `useRestoreWallet` takes the "already imported" path back to `selfCustodialChooseExperience`, which shows Enhanced pre-selected, and one Continue tap flips the stored mode from `anon` to `enhanced`.

*Proposal:* If a fresh choice on every pass is not intended, one option could be to seed `selected` from `persistentState.selfCustodialAccountModeByAccountId[onContinue.accountId]` when the destination carries an account id, falling back to Enhanced only when no mode is stored.

**2. A restore interrupted on the mode screen leaves the account onboarded with no mode recorded and no way to be re-prompted.**
A sibling of finding 1 — both reduce to the same open design question: what should the mode screen do when it's reached, or abandoned, for an account that's already active? `useRestoreWallet` calls `activateAccount(...)` and `reinitSdk()` before navigating to `selfCustodialChooseExperience`, so by the time the mode screen shows, the account is already fully active. Unlike the migration path — which commits the `MigrationCheckpoint.ChooseExperience` checkpoint in `useCompleteBackup` precisely so a kill resumes onto the mode screen — restore has no resume mechanism: a kill on that screen relaunches into the restored account and `selfCustodialAccountModeByAccountId` never gets an entry.

*Example:* A user restores their wallet, the app is killed while "Choose your experience" is showing, and on relaunch they're in the app with no mode chosen and no screen that ever asks again.

*Proposal:* Benign today since nothing reads the mode, but it becomes the future consumer's problem; one option is to document it where `AccountMode` says "Absent = not yet chosen" so the consumer is built to handle unset modes on fully-onboarded accounts, another is to re-prompt on next launch when the active self-custodial account has no entry.

**3. The mode screen removes the header back button but not the iOS swipe-back gesture, so the migration checkpoint and the visible screen can disagree.**
`headerless` on `OnboardingScreenLayout` sets `headerShown: false`, but the native-stack swipe-back gesture stays enabled. In the migration path `useCompleteBackup` saves the `ChooseExperience` checkpoint *before* navigating, so a swipe back lands on the backup-confirm screen while the persisted checkpoint says the user is past backup.

*Example:* A migrating user on the mode screen swipes back to the backup screen, kills the app there, and resume drops them on the mode screen — skipping the backup UI they were just looking at.

*Proposal:* The checkpoint semantics actually tolerate this (resume is correct either way), so possibly fine to defer; if the intent is "no way back", one option is `gestureEnabled: false` on the `selfCustodialChooseExperience` route, and if back is deliberate, a comment noting so would help.

## Test Coverage Gaps

- The retry test on `WalletCreationScreen` ("calls create on retry press") only asserts the call count, never that the second call carries the mode — a regression from `onPress={() => create(mode)}` back to `onPress={create}` would create the account with no mode recorded and still pass.
- No spec pins what the mode screen preselects when the account already has a stored mode (finding 1's scenario), so whichever behavior is intended is currently unlocked by tests.
- The new `headerless` prop on `OnboardingScreenLayout` runs during the `ChooseExperienceScreen` specs but nothing asserts `Screen` receives `headerShown={false}` — silently dropping the prop would pass.
- The former "v16 identity migration" test was repurposed into a v16→17 test without adding a v17 counterpart, so `migrateAndGetPersistentState({ schemaVersion: 17, ... })` is never exercised as input, breaking the suite's convention of testing the current version's identity migration.
- `replaceToCheckpoint` with a `ChooseExperience` checkpoint is unasserted — the replace-path specs cover `acceptTermsAndConditions` and `selfCustodialBackupMethod` only, so `StackActions.replace("selfCustodialChooseExperience", { onContinue: … })` never runs in tests.
- The resume story depends on the balances-overview screen committing its own checkpoint after the mode screen (per the PR description), but nothing in this diff asserts that hand-off; a regression there would silently re-show the mode screen on every resume past it.

## Code Smells

- `choose-experience-screen.spec.tsx` hand-redeclares the `OnContinue` union instead of deriving it from `RootStackParamList["selfCustodialChooseExperience"]` — it will drift silently when a fourth destination is added.

grimen (commented):
Consolidated into the changes-requested review above (findings 2–3 and the last two test coverage gaps originated here).

esaugomez31:
@grimen

## Findings

**1. Re-entering the mode screen ignores an already-stored mode.**
Confirmed, and the answer to the open question is that a fresh choice on every pass was *not* intended. Fixed on this branch, exactly as proposed: `useSelfCustodialAccountMode` gains a `getAccountMode(accountId)` read, and `ChooseExperienceScreen` seeds its selection from it whenever the destination carries an account id, falling back to Enhanced only when nothing is stored.

```ts
const isAccountPending = onContinue.route === ChooseExperienceContinueRoute.AcceptTerms
const storedMode = isAccountPending ? undefined : getAccountMode(onContinue.accountId)

const [selected, setSelected] = useState<AccountMode>(storedMode ?? AccountMode.Enhanced)
```

Creation is the one entry that reads nothing, because no account exists yet to read from. Your restore re-entry scenario no longer flips `anon` to `enhanced` on a single Continue.

**2. A restore interrupted on the mode screen leaves the account onboarded with no mode recorded.**
Took the first option. The `AccountMode` docblock in `app/types/account.ts` now states that absent is not only an in-progress onboarding state: it persists on fully onboarded accounts, because restore activates the account before showing the mode screen and has no resume back onto it, so a kill there leaves no entry and nothing asks again. Consumers are told to handle an unset mode rather than assume onboarding recorded one. A store test pins the contract the note describes: an account that never passed the screen stays absent from the map, and writing another account's mode does not invent an entry for it.

**3. The mode screen removes the header back button but not the iOS swipe-back gesture.**
Took `gestureEnabled: false`. The intent is "no way back": the screen opts into `headerless` deliberately, and the neighbouring `accountMigrationBalancesOverview` route already ships the same `headerShown: false, gestureEnabled: false` pair, so the two routes now match.

## Test Coverage Gaps

- The `WalletCreationScreen` retry now asserts the mode rides along, not just the call count: `toHaveBeenNthCalledWith(2, AccountMode.Anon)`. It drives Anon rather than the default, so a regression to `onPress={create}` falling back to Enhanced fails instead of passing.
- The preselect is pinned. It was first locked as the behaviour then stood (Enhanced regardless of what was stored, with a note that the expectation would flip once finding 1 was answered), and flipped along with the fix. `choose-experience-screen.spec.tsx` now asserts the read reaches the account the caller passed, that a stored Anon survives a Continue, that the user can still switch away from it, and that a creation entry reads nothing at all.
- `headerShown={false}` is asserted in a new spec for `OnboardingScreenLayout`, covering both the `headerless` case and the default where the header is left to the navigator.
- The v17 identity migration test is restored: it feeds `{ schemaVersion: 17, … }` and asserts the stored account modes pass through untouched, back in line with the suite's convention.
- `replaceToCheckpoint` with a `ChooseExperience` checkpoint is asserted, covering `StackActions.replace("selfCustodialChooseExperience", { onContinue: … })` with the provisioned account id.
- The balances-overview hand-off turned out to be covered already, in a chain across three specs: the mode screen navigates to `accountMigrationBalancesOverview` (`choose-experience-screen.spec.tsx`), that screen commits its own checkpoint on landing (`balances-overview-screen.spec.tsx` asserts `saveCheckpoint` is called with `MigrationCheckpoint.BalancesOverview`, and four sibling cases assert it is *not* called when the screen has not settled), and that checkpoint resolves forward rather than back to the mode screen (`use-migration-checkpoint.spec.ts`). Dropping the `saveCheckpoint` call fails the second, so the regression you describe is caught. A fourth assertion would only duplicate one of those three.

## Code Smells

- `choose-experience-screen.spec.tsx` now derives its route params from `RootStackParamList["selfCustodialChooseExperience"]["onContinue"]` instead of hand-redeclaring the union. The literal route strings in the spec are compile-checked against `ChooseExperienceContinueRoute` as a result, so a fourth destination cannot drift silently.

---

*Edited: when this was first written, finding 1 was answered in the follow-up PR and the commit bringing it here was still unpushed. It has since landed on this branch, so the finding and its test note now describe what this PR does. Commit SHAs were dropped rather than updated, since the branch has been restacked onto `#4046` and they would go stale again.*

grimen (changes requested):
## Findings

**1. Get Started reaches account creation without passing through the mode screen (blocking).**
[`handleCreateAccount`](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/screens/get-started-screen/get-started-screen.tsx#L75) navigates straight to `acceptTermsAndConditions` whenever `useAccountTypeOptions` resolves a `defaultSelected`, and that navigation carries no `mode` param. The mode screen is only wired into the flow that goes through `AccountTypeSelectionScreen`, so the shortcut skips it entirely and `useCreateWallet` ends up calling `create(undefined)`.
*Example:* in a region where non-custodial is enabled and custodial signup is not allowed, `options` collapses to the single self-custodial entry, so every account created from Get Started is stored with no mode — and per the doc comment on [`AccountMode`](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/types/account.ts#L16) nothing asks again.
*Proposal:* one option could be to route the `defaultSelected` branch through `selfCustodialChooseExperience` when the resolved flow is self-custodial, so both entry paths capture a choice.

**2. The back arrow on the mode screen leaves an already-activated account behind in the restore flow (blocking).**
The screen is registered with the navigator's [default header control](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/navigation/root-navigator.tsx#L885), and the accompanying comment treats that arrow as the deliberate way out. That reads well for the creation entry, but the restore entry arrives here after [`useRestoreWallet`](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/screens/self-custodial/onboarding/restore/hooks/use-restore-wallet.ts#L79) has already called `activateAccount()` and `reinitSdk()`, and only `BackupSuccessScreen` performs the `reset` to `Primary`.
*Example:* a user restoring a wallet taps back on the mode screen and lands on the restore-phrase screen with a fully restored, active account, no reset to `Primary`, and no mode recorded.
*Proposal:* one option could be to make the header control depend on the entry route, hiding it for the restore and migration entries where there is nothing coherent to go back to.

**3. The new checkpoint write in the migration backup flow can only block the user, never help them resume.**
[`use-complete-backup.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/screens/self-custodial/onboarding/hooks/use-complete-backup.ts#L85) now saves `MigrationCheckpoint.ChooseExperience` and aborts with `LL.errors.generic()` if the write fails. Since [`IS_COMMIT_POINT_BY_CHECKPOINT`](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/screens/account-migration/utils/migration-checkpoint-storage.ts#L44) marks that step `false`, a resume from it routes back to the migration explainer exactly as the previous `BackupAlerts` checkpoint did, so the write buys no resume behavior while adding a way to fail.
*Example:* [`saveCheckpoint`](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/screens/account-migration/hooks/use-migration-checkpoint-state.ts#L111) returns `false` while `ownerId` is still resolving, and the user who just finished their migration backup is shown a generic error on a screen with no remaining action.
*Proposal:* one option could be to drop the gate and navigate regardless; if the checkpoint is wanted for later use, treating a failed write as non-fatal would keep the flow moving.

**4. The subtitle promises a mode switch the app does not offer.**
The new copy at [`en/index.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/38e6bc4b03b9f4f48d891111b55a74ac693dbf5e/app/i18n/en/index.ts#L4038) says "You can switch whenever you want", but `AccountMode` is written in only two places — this screen and `useCreateWallet` — and there is no settings entry that changes it afterwards.
*Proposal:* one option could be to soften the sentence until a switcher exists, so the string does not have to be walked back in 28 locales later.

esaugomez31:
## Resolutions

**1. Get Started reaches account creation without passing through the mode screen — fixed.**
`handleCreateAccount` now routes the `defaultSelected` branch through `selfCustodialChooseExperience` when the resolved flow is self-custodial, so both entry paths capture a choice and `useCreateWallet` no longer receives `undefined`. Custodial is untouched and still goes straight to terms.
*Verified:* on device, with only the self-custodial option offered, pressing Create new account now lands on the mode screen and continues to terms; the account type screen is correctly skipped.
*Commit:* `db30d1ea5`

**2. The back arrow on the mode screen leaves an already-activated account behind — fixed.**
The header control now depends on the entry: creation keeps its arrow, restore and migration do not. The decision lives in one predicate, `canGoBackFromChooseExperience`, read by both the navigator and the screen.
One addition to the proposal: hiding the arrow alone would have left the Android hardware back untouched, which strands the same restored account by a different button. The screen therefore also guards `beforeRemove`, which catches the header tap, the hardware back and the gesture. Only `GO_BACK` and `POP` are refused, so an app-lock or migration-gate reset still removes the screen normally.
*Commit:* `0cf28d8b8`

**3. The new checkpoint write in the migration backup flow can only block the user — fixed.**
The gate is gone; the write stays so the stored step still reflects where the user is, but a failed write no longer stops the flow. The comment above it claimed a resume onto the mode screen, which `IS_COMMIT_POINT_BY_CHECKPOINT` contradicts, so it was corrected to say what actually happens.
*Commit:* `738aaf172`

**4. The subtitle promises a mode switch the app does not offer — not changed.**
The copy is the specification, so it stays. Worth noting for the record: the switcher does exist one PR up — `app/screens/settings-screen/settings/account-mode.tsx` lands in #4054 and opens this same screen with `ChooseExperienceEntry.Settings` — so the sentence is accurate from that point on, and softening it here would mean re-translating 28 locales twice.

## Notes

- Findings 1 and 4 are already handled in #4054. Each PR should stand on its own, so the fix for 1 lands here; the resulting duplicate in #4054 was removed during the restack, so the routing exists exactly once at the tip.
- Coverage added: the entry predicate across all entries including settings, the navigator wiring, the `beforeRemove` guard (hardware back and swipe refused for restore and migration, no guard on creation or settings, reset and forward navigation allowed through), and the checkpoint write no longer blocking.
- `yarn check-code` passes on this branch, and the stack was rebased onto `f2692a812`.

grimen (approved):

PR #4054, opened 2026-07-30
feat(region): add anon mode UI and availability gating

### LINKED ISSUE #3924: feat(compliance): Region determination (aka Geo-blocking) — three-mode user journey (Custodial / NC Enhanced / NC Anon)

esaugomez31:
## What

Makes **Anon Mode** real across the app: while the active self-custodial account is in Anon, the app never resolves a location (no IP or phone-country lookups) and every region-dependent surface greys out behind a single prompt that offers the switch to Enhanced.

> Stacked on #4051 — review/merge that first.

## Availability gating

- New `useDollarBalanceGated` / `useTransferGated` (Anon OR region) feed every availability surface and navigation guard. The region-only hooks stay where region alone must decide (forced conversion, migration preview).
- Home: Dollar row and Transfer button grey out, never hide. A gated balance still shows its **amount** when it isn't empty — the "not available" label only stands in when there is nothing to show — and the USD balance always counts toward the header total.
- Receive and Send lock to Bitcoin in Anon; Settings greys **Ways to get paid**, **Stable Balance** and contact support; the default-wallet row keeps its existing restriction behavior.

## No location resolution in Anon

- The guard lives inside `useDeviceLocation` / `useIpCountryCode` / `usePhoneCountryCode`, so no consumer can leak a lookup; in-flight lookups are cancelled and prior results are discarded on entry, so leaving Anon never exposes a stale country.
- Custodial flows are exempt (`isCustodialFlow`): phone login/registration opened from an Anon session keeps detecting normally, since that terrain belongs to the custodial account. A status guard also stops a late detection re-settle from yanking the user out of the captcha.
- The version footer reports only what the session **detected** for a self-custodial account — `Detected: Unknown` in Incognito — since registration is a custodial compliance fact and a self-custodial account never registers a region. Custodial keeps both `Registered` and `Detected`. The map falls back to the last viewed region or the default coords.

## Enhanced Mode prompt

- One app-level provider + lazily mounted modal; every gated tap opens it and it can switch the account to Enhanced on the spot.
- Account creation from an Anon session prompts it too: creation needs a region determination, which Anon refuses.

## Mode switching from Settings

- New **Mode** row (Account section) opens the mode-selection screen with a settings entry that edits the active account in place; both directions are allowed.
- Switching Enhanced → Anon with a remaining (or still unknown) dollar balance demands the conversion first: the migration's drain-conversion flag is generalized with a return destination, so the convert screen waives the region guard, prefills 100% USD → BTC, and the success screen returns to the mode selection. Continue holds while the wallet is still syncing, so a cold start cannot skip the gate.

## Stable Balance in Anon

- The app-side auto-convert listener pauses in Anon (live events and replay) and resumes on exit.
- The SDK stable-balance setting is dropped only when the dollar balance is known and empty; a remaining balance is never converted silently.

## Design review follow-ups

- **Anon is "Incognito" everywhere the user reads it** — the settings row, the mode cards, the gate prompts and the switch confirmation. The stored value stays `anon`: only the label changed, so no state migration was needed. Renamed across all 28 languages.
- The Lightning address carries an explicit **"(disabled)"** suffix everywhere it appears — the settings banner, the Ways-to-get-paid row and the account screen — shown only in Incognito.
- The banner **drops its copy affordance** in Incognito: no copy icon, and the tap opens the Enhanced prompt instead. The suffix is a label, never part of the address, so nothing can copy a string no wallet could pay.
- The switch confirmation breaks into two balanced lines through the copy itself rather than by width, so every locale breaks at its own phrase boundary instead of wherever the font happens to overflow.

## Lightning address in Incognito

One hole was left: the receive screen still **opened on the Lightning address method** and showed its QR, even though an Incognito account cannot be paid at that address.

- A single `useLightningAddressGated` now answers whether the address is withheld. Every surface that offers or labels it reads that one hook — receive, the settings row, the profile banner, the account screen — so the next reason to withhold it (the server-side deactivation) lands in one place instead of being restated per screen.
- **Receive**: the gate feeds `canUsePaycode`, which is the single choke point for the method — the initial snap to it, every later transition back, the identifier under the QR and the QR itself. Incognito opens on a Lightning invoice instead, so the wallet still receives, just not by address.
- The fallback keys on the **policy**, not on whether the address has loaded. The wallet provider clears the address on every account switch and re-resolves it, so `canUsePaycode` legitimately blips false; reacting to that would strand the screen on invoices for the rest of the session once the address came back.
- **Settings > Account** showed the address bare, without the "(disabled)" labelling the other two surfaces already carried. It now matches them.
- **Registration is refused** in Incognito, not merely gated by its parent group: publishing an address is exactly what the mode exists to keep back, so the row no longer opens the create modal on its own.

## Namespace

- `account-mode.tsx` moved to `app/screens/settings-screen/self-custodial/` (it opens with `if (!isSelfCustodial) return null`) and `anon-mode-convert-modal` to `app/self-custodial/components/` (its only consumer is the mode screen). Both are self-custodial-only and no longer sit in shared folders.

## Quality

- `DisabledFeature` keeps the same tree shape in both states, so a runtime toggle never unmounts its children.
- The migration balances preview reads the provisioned account from the screen's checkpoint instance (one query instead of two, no label flash).
- `AnonModeConvertModal` reuses the migration modal shell; `useTransferBlocked` is no longer exported.

## i18n

- New keys across all 28 languages with real per-language diacritics; `notNow` deduped into `common`.

## Testing

- Typecheck + lint clean; full coverage on the new code.
- All impacted suites green: choose-experience, home, wallet-overview, conversion flow, phone auth, settings, auto-convert, migration overview.
- The Lightning-address gate has its own spec driving the real persistent state, and the receive-side PayCode cases moved into `use-payment-request-paycode.spec.ts` — including the two that pin the distinction between an address that is **withheld** and one that is **still resolving**. The account screen's banner had no spec at all and now has one.


kngako:
Anon/enhanced mode UI is working great.  👌🏿

---

When the mode hasn't been set the UI doesn't show a set mode (but I am assuming the default behavour for users who haven't selected the setting would be enhanced mode)?

image

The icon for the mode being the location pin drop is a little confusing though. The standard "Security" or "AdminPanelSettings" material icon might be an interesting candidate for this mode. 


image

image


image


The "switch to enhanced" mode convenience function is working as expected as well.
image

esaugomez31:
@kngako Perfect, you were right: for existing accounts with no mode set, the row wasn't showing Enhanced. That's fixed now. As for the icon, that was @designsats's decision, maybe they can weigh in on your suggestion

grimen (changes requested):
## Findings

**1. Anon fails open on the self-custodial creation region block (blocking).**
`useCreationBlock` reads `useDeviceLocation()` with no custodial-flow exemption, so in Anon it sees `countryCode: undefined` and `isBlockedCountry(undefined, …)` returns false, letting creation pass unchecked. The PR description says account creation from an Anon session prompts Enhanced Mode, but no creation path calls `promptEnhancedMode`.

*Example:* With an Anon account active, "add account" → self-custodial creation passes `selfCustodialCreationBlockedCountries` without any region determination.

*Proposal:* One option is to fail closed on an unresolved country, as `decideCustodialEligibility` already does; another is to open the Enhanced Mode prompt on the creation path as the description states.

**2. Stable-balance deactivation is one-way.**
`useAnonStableBalanceDeactivation` drops the SDK stable-balance setting on entering Anon, but nothing restores it when the account switches back to Enhanced, so a prior opt-in is silently lost.

*Example:* A user with Stable Balance on switches to Anon and back; incoming payments stay bitcoin even though the auto-convert listener itself resumes.

*Proposal:* One option is to remember the pre-Anon setting and re-activate it on exit; if the permanent drop is intended, surfacing it in the UI would help.

**3. Post-drain stale-balance race on the settings entry.**
After the drain conversion, the success screen resets into `selfCustodialChooseExperience`, which fires `refreshWallets()` fire-and-forget; `isWalletReady` can already be true with the pre-conversion balance still cached, so the conversion gate re-checks against stale data.

*Example:* The user completes the 100% conversion, presses Continue as the screen resumes, and the convert modal demands the already-completed conversion again.

*Proposal:* One option is to fold the in-flight refresh into `isContinueWaiting` so Continue holds until the refreshed balance lands.

**4. DisabledFeature can be activated through screen readers.**
`pointerEvents="none"` blocks touches but leaves children in the accessibility tree, and the wrapper sets `accessible={false}`, so VoiceOver/TalkBack can still focus a gated row and trigger its `onPress`.

*Example:* In Anon, a screen-reader user activates the greyed Stable Balance row and navigates past the gate.

*Proposal:* One option is `accessibilityElementsHidden` / `importantForAccessibility="no-hide-descendants"` on the content view while disabled, plus `accessibilityState={{ disabled: true }}`.

## Test Coverage Gaps

- `useDeviceLocation`: no test toggles Anon mid-lookup, so the in-flight discard (`active = false` cleanup) and the loading re-arm on exit — the cancellation behavior the PR description claims — are unverified.
- `useIpCountryCode`: the new Anon branch that clears state and skips the lookup has no direct test.
- `settings-screen`: nothing asserts the Ways-to-get-paid group actually receives `disabled={isAnonMode}` / `onDisabledPress={promptEnhancedMode}` (the group component is only tested in isolation).
- `home-screen`: `isEnhancedModePromptVisible` folded into the modal-suppression logic is untested (the spec pins it to false).
- `map-screen`: dropping the `countryCode` requirement from the last-region path also changes non-Anon detection-failure behavior and has no spec.

designsats:
@esaugomez31 Changed all 3 icons on this screen. 


image

esaugomez31:
> @esaugomez31 Changed all 3 icons on this screen.
> image

@designsats Thanks! I'll resolve this in #4051, since that's where this screen was added:
https://github.com/blinkbitcoin/blink-mobile/pull/4051

esaugomez31:
@grimen

## Findings

**1. Anon fails open on the self-custodial creation region block.**
Confirmed, and the cause sits one level below the symptom: `useCreationBlock` was inheriting the *active* account's Anon. The account being created has no mode of its own yet, and a custodial one never will, so neither should inherit another account's. Custodial creation was reading the same unresolved country, so a self-custodial account in Anon was quietly blinding that check too.

It now resolves the region regardless of the active account's mode, and the check decides on a real country again. That closes the fail-open without failing closed: nobody is blocked by a slow lookup, and Anon still resolves nothing for anything that account does itself.

Moving the check after the mode selection — so Enhanced checks and Anon does not — plus the Enhanced prompt the description mentions, is a flow change on the product side and goes in its own ticket. The PR description is corrected accordingly.

**2. Stable-balance deactivation is one-way.**
Took the first option, with one change forced by the SDK: the setting is remembered, but reactivation cannot be unconditional. Per the Breez docs, *"activation immediately converts excess sats to the specified token"* and *"activating with BTC balance is not free — it sweeps"*. Leaving Anon is exactly when the user holds bitcoin, so a blind restore would convert their balance and charge a fee for it.

The fix separates what was collapsed into one thing. The SDK was the only record of the setting, so switching it off erased an opt-in the user never withdrew. The preference now lives in `persistentState`, keyed per account, while the SDK setting is still dropped on entering Anon. Leaving Anon restores from the marker and clears it.

Only what Anon switched off comes back: a user who turned it off themselves leaves no marker, so nothing reactivates behind them. Deactivation still requires an empty dollar balance, which the mode switch already guarantees by draining first.

**3. Post-drain stale-balance race on the settings entry.**
Fixed as proposed. The entry refresh is tracked rather than fired and forgotten, and folded into `isContinueWaiting`, so Continue holds until the refreshed balance lands instead of re-checking the pre-conversion figure. A failed refresh releases the gate on the balance it already has, rather than stranding the screen on a spinner.

**4. DisabledFeature can be activated through screen readers.**
Fixed. While disabled, the children leave the accessibility tree (`accessibilityElementsHidden` plus `importantForAccessibility="no-hide-descendants"`) and the wrapper stands in for them as a single named button.

Two notes on the details:

`accessibilityState={{ disabled: true }}` is not applicable here. `Pressable` derives that field from its own `disabled` prop and overrides whatever the caller passes, and setting it true would make the wrapper inert — which defeats the gate, since activating it is what explains the block.

Hiding the children takes their labels with them, so the gate needs its own. There were six call sites, not three: `SettingsGroup`, `stable-balance`, `community-need-help`, the home action buttons, the wallet-overview dollar row, and the settings mode row upstack. Every one of them now names the gate, and the dollar row carries the reason too, since gating it hides the label that stated it.

## Test Coverage Gaps

- `useDeviceLocation` now covers the Anon toggle mid-lookup: an in-flight answer is discarded when Anon switches on, loading re-arms for a fresh resolve when it switches off, and a country resolved before the switch is dropped.
- `useIpCountryLookup` gained its own Anon block: the lookup is skipped even when the caller enables it, an already-resolved country is cleared, and it runs again once Anon is off.
- `settings-screen` asserts the Ways-to-get-paid group is gated in Anon and that a tap routes to the Enhanced prompt. Verified by mutation: dropping `disabled`/`onDisabledPress` fails them.
- `home-screen` covers `isEnhancedModePromptVisible` in the modal suppression, in both directions.
- `map-screen` covers the initial region: stored region reused, detected country on a first visit, and the default coords both without a resolved country and for an unrecognised one.

Coverage is 100% on the files added or rewritten here (`disabled-feature`, `stable-balance-anon-pause`, `use-anon-stable-balance-deactivation`, `use-creation-block`, `SettingsGroup`). Two pre-existing spec leaks turned up on the way and are fixed: `mockIsAnonMode` was never reset in the `useIpCountryLookup` block, and the settings tests only passed because an earlier `describe` had loaded the locale.

designsats:
@esaugomez31 

Additonal changes:
- Adding back arrow to this screen [Figma](https://www.figma.com/design/9MQuQi8ZhXVvDibWSI3C4c/Blink?m=auto&node-id=19656-36689&t=YnvKuiYaUHpUszCj-1)
- Renaming it on this screen and Settings item from "Anon" to "Incognito" 

image

### Settings screen

- we add "(disabled)" to two places in Settings screen to indicate LN address is disabled (cant be receiving)
- copy icon from account is hidden
- we change "Mode: Anon" to "Mode: Incognito"

image


Success screen
- text shouldnt use full screen width

image

designsats:
@esaugomez31 @grimen Tested in android emulator, works as expected. Correct copy and icons

image

Screenshot_1787153276

grimen (changes requested):
## Findings

**1. Continue button can spin forever, and blocks the ungated direction (blocking).**
`isContinueWaiting` in `choose-experience-screen.tsx` is `isSettingsEntry && (!isWalletReady || isRefreshingWallets)`, and `useActiveWallet().isReady` is only true for `Ready` and `Degraded`. When the Spark SDK settles on `Offline` or `Error` the button stays disabled with a spinner and nothing tells the user why. The same condition also applies to the Incognito to Enhanced direction, which has no balance gate to wait for.

*Example:* An Incognito user with no connectivity opens Settings, picks Enhanced, presses Continue, and can never leave Incognito.

*Proposal:* One option could be to scope the wait to the Anon-bound switch only, and to surface an error (or fail open) once the wallet status is non-transient rather than treating it as still loading.

**2. The gated account banner tap does nothing (blocking).**
In Incognito, `banner.tsx` sets `onPress={isLightningAddressGated ? undefined : handleCopy}`, but `AccountBanner` is rendered directly in `settings-screen.tsx` rather than inside a `SettingsGroup` or `DisabledFeature`. The `TouchableOpacity` still plays its press animation, so the row looks interactive while nothing happens, and unlike every other gated surface it never opens the Enhanced-mode prompt.

*Example:* An Incognito user taps the disabled `user@blink.sv` row, sees the press feedback, and gets no explanation of why the address is unavailable.

*Proposal:* Routing this row through the same gated wrapper the other surfaces use would make the tap open the Enhanced prompt, which is also what the PR description says it does today.

**3. Add-account from an Incognito session resolves a region (blocking).**
`use-creation-block.ts` calls `useDeviceLocation({ isCustodialFlow: true })` unconditionally, and `switch-account.tsx` navigates to `getStarted` from Settings, Switch account, Add account while an Incognito account is active. That path therefore runs `useCountryCodeQuery` and `resolveIpCountryCodeCached()` from inside Incognito, which is the one thing the PR description says never happens.

*Example:* An Incognito user opens Switch account and taps Add account, and the app performs an IP country lookup before any new account exists.

*Proposal:* If the exemption is deliberate because creation itself must stay gated, a flag named for creation rather than `isCustodialFlow` plus a short comment would make that clear; otherwise the lookup could be deferred until the user has actually chosen a custodial account type.

**4. Stable balance can be reactivated after the feature flag was turned off.**
`use-anon-stable-balance-deactivation.ts` treats `isStableBalanceActive === false` as "the user had it off", but the wallet provider computes that value as `stableBalanceEnabled && sdkStableBalanceActive`, so a disabled remote flag produces exactly the same `false`.

*Example:* A user enters Incognito while the flag is on, the flag is turned off remotely, and on leaving Incognito the reactivation effect calls `activateStableBalance` and switches the feature back on.

*Proposal:* Reading the raw `sdkStableBalanceActive` (or checking `stableBalanceEnabled` before reactivating) would keep the two meanings of `false` apart.

## Test Coverage Gaps

- No test covers `isContinueWaiting` when the wallet status is `Offline` or `Error`, so the permanently disabled Continue button is not caught.
- No test covers the Incognito to Enhanced direction from the settings entry, only the Anon-bound switch.
- The gated `AccountBanner` has no assertion that a tap reaches the Enhanced-mode prompt, which is why the missing wrapper went unnoticed.
- `useCreationBlock` is not exercised from an active Incognito session, so the location lookup on the add-account path is untested.
- The reactivation path in `use-anon-stable-balance-deactivation` is not tested with `stableBalanceEnabled` false.

esaugomez31:
## Resolutions

**1. Continue button can spin forever, and blocks the ungated direction — fixed.**
Both halves of the proposal. The wait is now scoped to the Anon-bound switch, since that is the only direction that reads the balance; leaving Incognito decides nothing and no longer waits at all, which is the case in the example. And `Offline` and `Error` are treated as the answers they are rather than as loading: Continue releases and the refusal is spoken through `errors.network.connection` instead of leaving a dead button. Failing open was the other option and was left alone on purpose, since it would skip the dollar-balance check that exists to avoid an unasked conversion.
*Commit:* `261f1de35`

**2. The gated account banner tap does nothing — fixed.**
The row routes its tap like every other gated surface instead of dropping the handler, so the press reaches the Enhanced-mode prompt and the animation has something to answer with.
*Commits:* `35425637f`, and `6790286da` on #4055, which extends the same row to the restricted-region case

**3. Add-account from an Incognito session resolves a region — already resolved in #4144.**
`useCreationBlock` no longer reads the device location at all there: the rules resolve when an option is actually submitted rather than on mount, so opening Add account from Incognito, or arriving to restore, locates nobody. Duplicating it here would mean writing code that PR deletes, so this one is left to #4144.

**4. Stable balance can be reactivated after the feature flag was turned off — fixed.**
The reactivation reads `stableBalanceEnabled` on its own rather than through `isStableBalanceActive`, which folds the flag into the SDK's setting and makes a disabled flag indistinguishable from a user who turned the feature off. The pause marker is deliberately kept in that case, so a flag that returns still restores the setting the user never withdrew.
*Commit:* `c57188f88`

## Test Coverage Gaps

- `isContinueWaiting` is covered for `Offline` and for `Error`: Continue releases, the refusal is surfaced, and no switch is made.
- The Incognito to Enhanced direction is covered from the settings entry, both with an unreachable wallet and while the wallet is still syncing.
- The gated `AccountBanner` asserts that a tap reaches the Enhanced-mode prompt, and that nothing is prompted outside Incognito.
- The reactivation path is covered with `stableBalanceEnabled` false, including that the marker survives.
- `useCreationBlock` from an active Incognito session is covered in #4144, where the lookup now lives.

## Notes

- Each fix was mutation-tested: reverting it fails the tests written for it.
- One follow-up commit, `90e4752ba`, carries the guard rename into a spec that was written on #4046 before the Anon Mode vocabulary landed.
- The stack was rebased onto `8e01f5597`.

grimen (approved):

PR #4055, opened 2026-07-31
feat(region): add restricted-region sanctions UI

### LINKED ISSUE #3924: feat(compliance): Region determination (aka Geo-blocking) — three-mode user journey (Custodial / NC Enhanced / NC Anon)

esaugomez31:
## What

Adds the **sanctions layer UI**: when the session IP resolves to a restricted jurisdiction, the app presents **"Restricted region detected"** and greys every Blink-served surface, while Bitcoin send/receive, the local wallet and recovery stay fully operable.

> Stacked on #4054 — review/merge that first.

## Evaluation

- Reads the **existing** remote-config lists — no new list mechanism: `custodialCreationBlockedCountries` / `selfCustodialCreationBlockedCountries`, selected by the active account type.
- Sanctions ride the **live session IP, never the phone country**. Anon never evaluates (no lookup runs), and without an active account nothing is evaluated (creation has its own gate).
- Nothing persists: a session that resolves to a clean region releases everything on its own.

## Blocker — one surface, two variants

- **Custodial (full-screen, no way out):** every custodial function is Blink-served, so the block covers the whole session. Rendered as a **native modal**: nothing can present above it, the Android back button is inert, and screen readers treat it as modal. Shows the account balances (hidden until the query resolves), **Contact support** (support inbox) and **Learn more**.
- **Non-custodial (dismissible modal):** the local wallet stays usable behind it and closing never restores Blink-served features. Presents **once per restricted session**, re-arms when the region clears, and any gated tap reopens it. Kept mounted through the restricted session so the dismiss animation plays.
- Copy is verbatim from the ticket tables. **Learn more** opens the official blocked-countries FAQ page, via a constant now shared with the terms screen that already linked it.

## Namespace

The two variants are not one component with a flag — they are two components, each serving exactly one account type, so each lives in its own namespace:

| Surface | Lives in | Serves |
|---|---|---|
| `restricted-region-screen.tsx` | `app/custodial/components/` | Custodial only |
| `restricted-region-modal.tsx` | `app/self-custodial/components/` | Self-custodial only |
| provider, body, banner | `app/components/restricted-region/` | Both — the provider picks between them |

The shared copy already lived in `restricted-region-body.tsx`, so the split cost no duplication.

## Design review follow-ups

- The modal's warning icon uses **`primary`**, matching the custodial full-screen block, which already did. The two sanctions surfaces no longer disagree on their tint.
- The settings banner matches the backup nudge it sits beside: `primary` border and title over a `grey5` fill, with the same typography. It previously used `warning` and had no fill at all.

## Home (non-custodial)

- Transfer renders disabled; the Dollar row stays muted with the **amount visible**. Taps route by precedence: Anon prompt, then the sanctions modal, then the region-restriction modal.
- Auto-presenting modals (migrate-now, trial upgrade, forced conversion) are suppressed while sanctioned, so nothing floats over the restriction surfaces.

## Settings (non-custodial)

- Top banner with the fixed copy, **Mode** row disabled (switching modes is blocked while restricted) and **Ways to get paid** disabled — every gated tap opens the sanctions modal, mirroring the Anon pattern with its own surface.

## i18n

- Five new keys across all 28 languages, reusing each language's existing region terminology and register.

## Testing

- New module at **100% coverage** (21 tests: provider session semantics, both blocker variants, banner).
- Impacted suites green: home, wallet-overview, settings, account-mode, accept-terms.
- Typecheck + lint clean.

## Notes for review

- The spec mentions the custodial blocker balances as "cached, timestamped"; the mockup carries no stamp, so none is rendered — flagged for design.
- Session-start re-resolution (lookup retry and the re-check route) belongs to the upcoming session-start phase; this PR is the presentation layer over the current lookup.

- The app-side block is presentation only: screens and actions are disabled in the client, but the backend still serves any request that reaches it directly. Server-side enforcement of gated/sanctioned requests is separate backend work and remains pending.
- `custodialCreationBlockedCountries` and `selfCustodialCreationBlockedCountries` are now also used for existing accounts: the session blocker evaluates them for the active account, not only at signup.


grimen (changes requested):
## Findings

**1. Upgrade modal is not suppressed during the pending evaluation window (blocking).**
`triggerUpgradeModal` in `home-screen.tsx` returns early on `isRestrictedRegion` but not on `isRestrictedRegionEvaluationPending`, while the migrate-now prompt waits on both flags. The PR states that no auto-presenting modal may float over the restriction surfaces, and this asymmetry leaves one path open.

*Example:* on a custodial level-0 cold start where the IP lookup outlasts the 2 s splash cap, the app is revealed, the trial-upgrade modal auto-presents, and the late restricted verdict then mounts the full-screen block over it — with the modal still open behind it.

*Proposal:* one option is to gate `triggerUpgradeModal` on the pending flag as well, mirroring what `isAnotherHomeModalVisible` already does for migrate-now.

## Test Coverage Gaps

- `navigation-container-wrapper.tsx`: the deferral of `RNBootSplash.hide` behind `bootSplashGate.whenReleased()` executes in the existing wrapper spec but is never asserted — no test would fail if the hide stopped waiting on the gate.
- Settings screen: the "Ways to get paid" group being disabled and routing its tap to the sanctions modal while restricted (non-Anon) is unasserted; only the banner and the Mode row are covered.
- Home screen: neither the upgrade-modal early return on `isRestrictedRegion` nor the migrate-now suppression through `isAnotherHomeModalVisible` is asserted in a sanctioned session — the guard lines run in the new restricted tests, but nothing fails if either modal presents.

## Code Smells

- The `body + "\n\n" + bodyReturn` copy composition is repeated in `restricted-region-banner.tsx`, `restricted-region-modal.tsx` and `restricted-region-screen.tsx`; a small shared component (or a single composed key) would keep the three surfaces in lockstep when the copy changes.
- The FAQ link opens via `Linking.openURL` in the modal and screen but via `InAppBrowser.open` on the terms screen that shares `BLOCKED_COUNTRIES_FAQ_LINK`. For the custodial screen the external browser is likely deliberate (nothing can present above a native modal), but the non-custodial modal could use the in-app browser like the rest of the app — a comment would help if intended.

esaugomez31:
@grimen all six are addressed. Details per item, in your order.

## Findings

**1. Upgrade modal is not suppressed during the pending evaluation window (blocking).** Fixed. `triggerUpgradeModal` now returns early on `isRestrictedRegionEvaluationPending` as well, with the flag added to its dependency list.

Worth noting why the early return does not strand the modal: the callback is a dependency of the focus effect that arms it, so when the verdict settles its identity changes, the effect re-runs and re-arms the timer. The modal is postponed, not lost. `isSettled` is also guaranteed to flip (`resolveIpCountryCodeCached` never rejects), and the splash gate has its own cap, so there is no path where the pending flag sticks.

Covered by three tests in `home.spec.tsx`: the modal stays away while restricted, stays away while pending, and — as a control — still auto-presents on an unrestricted level-zero account. The control mattered: without it the first version of the two guard tests passed for the wrong reason, because the focus effect re-arms its timer as the queries resolve and a single `advanceTimersByTime` fired before the final timer existed. Removing the pending guard now fails the pending test.

## Test Coverage Gaps

- **`navigation-container-wrapper.tsx` boot-splash deferral.** Asserted now. One correction: the deferral was not executing in the existing spec at all — the `NavigationContainer` mock returned `children` and dropped every prop, `onReady` included, so that line never ran. The mock now fires `onReady` once, and a hand-driven gate lets the test hold the splash and release it: `RNBootSplash.hide` must not have been called before the release, and must have been after. Verified by replacing the deferral with a direct `hide` call, which fails the test. The other 20 tests in the file are unaffected.
- **Settings "Ways to get paid" while restricted.** Asserted: with a restricted, non-Anon session the group renders behind `DisabledFeature` and a tap routes to `presentRestrictedRegionModal`, not to the Enhanced prompt. The Anon side of that same gate was already covered, so only the sanctions branch was missing.
- **Home upgrade modal and migrate-now in a sanctioned session.** Both asserted, for `isRestrictedRegion` and for the pending flag.

## Code Smells

- **Repeated copy composition.** Extracted to `RestrictedRegionBody`, used by the banner, the modal and the screen. Each surface keeps its own `type` and style, so nothing moves visually; the composition lives in one place.
- **FAQ link inconsistency.** Your read was right, and the reason is concrete: `restricted-region-screen.tsx` renders inside a React Native `Modal`, which nothing can present above, so `Linking.openURL` is required there and now carries a comment saying so. The non-custodial modal renders through `CustomModal`, i.e. `react-native-modal`, a JS overlay — so it switched to `InAppBrowser.open`, matching `accept-t-and-c` which shares the same constant.

## Verification

`tsc` and `eslint` clean, 370 tests across the 35 affected suites green. The three behavioural fixes were each validated by mutating the source and confirming the new tests fail.

One thing not covered by tests: `InAppBrowser.open` has no `isAvailable()` guard, matching the five existing call sites in the app. Worth a look on device before this merges.

grimen (dismissed):

esaugomez31:
@grimen
This branch is already up to date. Every point from your 08-04 review was resolved here, that review was dismissed today, and nothing has changed since

grimen (dismissed):

grimen (approved):

PR #4057, opened 2026-07-31
fix(notifications): wrap notification history text inside screen bounds

esaugomez31:
## Summary

Notification history items render an icon and a text column inside a `flexDirection: "row"` container, but the text column had no width constraint. Long titles and bodies (like the "Time to move to non-custodial" and "Self-custodial accounts have arrived" bulletins) sized to their content and ran past the right edge of the screen, getting clipped instead of wrapping.

This constrains the text column with `flex: 1` so it takes the remaining row width and long text wraps within screen bounds.

Closes blinkbitcoin/blink-wip#946

## Changes

- `app/screens/notification-history-screen/notification.tsx`: add `flex: 1` to the text column, plus a `testID` used by the regression test, and drop an optional chain on `icon` that is unreachable inside the icon-present branch.
- `app/screens/notification-history-screen/notification.tsx`: map icon enum values with `replace(/_/g, "-")` instead of single-occurrence `replace("_", "-")` — multi-underscore values (`CLOSE_CROSS_WITH_BACKGROUND`, `WARNING_WITH_BACKGROUND`) previously produced names missing from the GaloyIcon map and rendered no icon.
- `__tests__/screens/notification-history-screen/notification.spec.tsx`: new spec with 100% statement/branch/function/line coverage of the component — wrap regression, galoy-icon and ionicon fallback branches, acknowledged styling (initial and transition after mount), and deep link / external link / no-action press behavior.

## Test plan

- `yarn jest __tests__/screens/notification-history-screen/notification.spec.tsx --coverage`: 12 passed, 100% coverage on `notification.tsx`, no console noise

## Screenshots

| Before | After |
| :---: | :---: |
| Screenshot from 2026-07-31 15-32-42 | Screenshot from 2026-07-31 15-34-11 |

### DIFF
--- a/__tests__/screens/notification-history-screen/helpers.ts
+++ b/__tests__/screens/notification-history-screen/helpers.ts
@@ -1,0 +1,10 @@
+import { Icon } from "@app/graphql/generated"
+
+/**
+ * An icon the backend can send but the app ships no asset for.
+ *
+ * Declared once so the guard in `utils.spec.ts`, which asserts it really is unmapped,
+ * keeps every spec that renders it honest. Two copies could drift, and the day a rocket
+ * asset lands only the copy next to the guard would notice.
+ */
+export const UNMAPPED_ICON = "ROCKET" as Icon
--- a/__tests__/screens/notification-history-screen/notification.spec.tsx
+++ b/__tests__/screens/notification-history-screen/notification.spec.tsx
@@ -12,6 +12,8 @@ import { Notification } from "@app/screens/notification-history-screen/notificat
 
 import { findPressableParent } from "../helper"
 
+import { UNMAPPED_ICON } from "./helpers"
+
 /**
  * `timeAgo` reads the wall clock, so the relative-time assertion only stays
  * stable if both the clock and the notification timestamp are pinned.
@@ -19,9 +21,6 @@ import { findPressableParent } from "../helper"
 const FIXED_NOW_MS = Date.UTC(2026, 0, 15, 12, 0, 0)
 const THIRTY_SECONDS_AGO = Math.floor(FIXED_NOW_MS / 1000) - 30
 const ACKNOWLEDGED_AT = Math.floor(FIXED_NOW_MS / 1000) - 60
-
-/** An icon the backend can send but the app ships no asset for. */
-const UNMAPPED_ICON = "ROCKET" as Icon
 
 const makeNotification = (
   overrides: Partial<StatefulNotification> = {},
@@ -100,7 +99,7 @@ describe("Notification", () => {
     )
 
     expect(getByTestId("notification-default-icon")).toBeTruthy()
-    expect(queryByTestId("icon-rocket")).toBeNull()
+    expect(queryByTestId(`icon-${UNMAPPED_ICON.toLowerCase()}`)).toBeNull()
   })
 
   it("renders unacknowledged text and icon in the primary color", () => {
--- a/__tests__/screens/notification-history-screen/notification.spec.tsx
+++ b/__tests__/screens/notification-history-screen/notification.spec.tsx
@@ -68,6 +68,14 @@ describe("Notification", () => {
     expect(getByTestId("galoy-icon-bell")).toBeTruthy()
   })
 
+  it("maps every underscore of the icon enum to the dashed icon name", () => {
+    const { getByTestId } = renderNotification(
+      makeNotification({ icon: Icon.WarningWithBackground }),
+    )
+
+    expect(getByTestId("galoy-icon-warning-with-background")).toBeTruthy()
+  })
+
   it("falls back to the default ionicon when the notification has no icon", () => {
     const { queryAllByTestId } = renderNotification(makeNotification())
 
--- a/__tests__/screens/notification-history-screen/notification.spec.tsx
+++ b/__tests__/screens/notification-history-screen/notification.spec.tsx
@@ -1,5 +1,5 @@
 import * as React from "react"
-import { Linking, StyleSheet, View } from "react-native"
+import { Linking, StyleSheet } from "react-native"
 import { fireEvent, render } from "@testing-library/react-native"
 
 import { ThemeProvider } from "@rn-vui/themed"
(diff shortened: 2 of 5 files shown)


grimen (changes requested):
## Findings

**1. The new spec mocks `GaloyIcon`, so the icon-name regression it guards can't actually fail.**
The mock renders `galoy-icon-${name}` from whatever string the component passes, so the assertions confirm the `replace(/_/g, "-")` output but not that the resulting name exists in the icon map — which is precisely the bug being fixed. The real `GaloyIcon` already emits `testID={\`icon-${name}\`}` on both the phosphor and custom-SVG branches, and returns `null` for an unmapped name.

*Example:* If `WARNING_WITH_BACKGROUND` were mapped to `warning_with-background`, the mocked test would still pass with `getByTestId("galoy-icon-warning_with-background")`, while the app renders blank space.

*Proposal:* One option could be to drop the mock and assert `getByTestId("icon-warning-with-background")` against the real component, so the test fails whenever a derived name has no asset behind it.

**2. The `as IconNamesType` cast still fails silently for enum values with no matching asset.**
All 40 current `Icon` enum values resolve after this fix (verified against `icons` in `galoy-icon.tsx`), but the cast means any value the backend adds later reaches `GaloyIcon`, misses both maps, and returns `null` — an empty gap where the icon should be, with no fallback and no warning.

*Example:* The backend ships `Icon.Rocket` before the SVG asset lands; the notification row renders with a blank leading slot instead of the default bell.

*Proposal:* One option could be to check membership before rendering (`derivedName in icons`) and fall through to the existing `notifications-outline` ionicon branch otherwise, which would also make the fallback the single place unknown icons are handled.

## Test Coverage Gaps

- The acknowledged-icon path is not really covered: `greys out the icon variant once acknowledged` asserts the same body-text color as the test directly above it, and the mock discards the `color` prop, so `colors.grey2` never reaches an assertion. Forwarding `color` into the mock (or using the real component) would make that test distinct from its neighbour.

## Code Smells

- `makeNotification` derives `createdAt` from `Date.now()` and the spec asserts the literal string `"a few seconds ago"`; a fixed timestamp plus fake timers would keep the relative-time assertion independent of wall-clock.

The `Android` and `iOS` checks are red, but they fail identically on every currently open PR (#4061–#4063), so they aren't attributable to this change; `Tests`, `Check Code`, and `qlty check` are green.

esaugomez31:
@grimen

## Findings

**1. Mock dropped, the icon assertions now run against the real component.**
`notification.spec.tsx` no longer mocks `@app/components/atomic/galoy-icon`, so `getByTestId("icon-bell")` and `getByTestId("icon-warning-with-background")` resolve through the real `icons` map and fail when a derived name has no asset behind it. Both branches render fine under jest: `phosphor-react-native` ships CommonJS, and the SVG assets stubbed by `jest-transform-stub` still forward `testID` and `color`.

The per-name assertion only covers the two values the spec exercises, so the new `utils.spec.ts` adds the net that actually guards the mapping: it iterates every value of the backend `Icon` enum and fails if any of them resolves to nothing. A future enum value without an asset breaks the suite instead of shipping a blank slot.

**2. The cast is gone, unmapped names fall through to the ionicon.**
Name resolution moved to `toGaloyIconName` in the screen's `utils.ts`, where a `name is IconNamesType` type guard checks membership in `icons` before the name reaches `GaloyIcon`, so the `as IconNamesType` cast is no longer needed. An unmapped value returns `undefined` and the component renders the existing `notifications-outline` ionicon, which is now the single place unknown icons are handled. The component no longer owns the enum-to-asset mapping, it only picks a branch.

No visible change today: all 40 current enum values resolve after the underscore fix, as you verified. This only covers what the backend adds later.

## Test Coverage Gaps

- The acknowledged-icon test now asserts the icon node itself (`getByTestId("icon-bell").props.color` is `colors.grey2`), which is distinct from the body-text assertion above it. Added the unacknowledged counterpart asserting `colors.black` on the same node, plus a case where an enum value with no asset renders the default ionicon.

## Code Smells

- Both specs pin the clock with `jest.useFakeTimers({ now: FIXED_NOW_MS })` and derive `createdAt` from that fixed instant, so the relative-time assertions no longer depend on wall-clock time. `utils.spec.ts` also covers the remaining `timeAgo` branches (minutes, hours, days, singular and plural) that the file had no test for.

## Verification

- `yarn jest __tests__/screens/notification-history-screen --coverage`: 22 passed, 100% statements/branches/functions/lines on both `notification.tsx` and `utils.ts`, no console output
- `yarn jest __tests__/screens/settings-screen/settings-screen.spec.tsx` (transitive consumer): 14 passed

Commits: `9b35c8dad` (fallback and mapping extraction), `e775327b6` (real-component icon assertions and pinned clock), `192a0a79b` (utils spec).

grimen (changes requested):
## Findings

**1. The all-icons contract test couples schema codegen to shipped assets.**
The docstring on `toGaloyIconName` says falling back to the default ionicon is the designed behavior for enum values the app has no asset for, but `resolves every backend icon to an asset the app ships` in `utils.spec.ts` fails the suite whenever `generated.ts` gains an `Icon` value with no matching entry in the icon map.

*Example:* the backend adds `ROCKET` to the schema, the routine `generated.ts` regeneration after a staging deploy picks it up, and Tests goes red repo-wide until an asset ships — even though the component would gracefully render the default bell.

*Proposal:* if this is an intentional drift guard (fail fast so a designed icon is never silently swapped for the bell), a one-line comment saying so would help; otherwise the two unmapped-icon fallback tests already lock in the designed behavior and this assertion could be dropped.

## Code Smells

- `UNMAPPED_ICON = "ROCKET" as Icon` is declared in both spec files, but only `utils.spec.ts` carries the guard assertion (`"rocket" in icons === false`) that keeps the fixture honest if a rocket asset is ever added.

esaugomez31:
@grimen

## Findings

**1. The all-icons contract test couples schema codegen to shipped assets.**
Dropped it. Not an intentional drift guard — it contradicted the docstring sitting right above it, which calls the default-ionicon fallback the designed behavior for enum values the app has no asset for. A routine `generated.ts` regeneration turning the suite red repo-wide is a cost every branch pays for a gap the component already handles gracefully, and "a designed icon silently became a bell" is a review signal, not a build failure. The two fallback tests you pointed at keep the designed behavior locked in, and the underscore mapping this PR fixes is still covered on both sides — `replaces every underscore of the enum value` in `utils.spec.ts` and `maps every underscore of the icon enum to an icon that has an asset` against the real component.

## Code Smells

- `UNMAPPED_ICON` now lives once, in `__tests__/screens/notification-history-screen/helpers.ts`, following the per-folder helper convention already used by `account-migration` and `payment-details`. Both specs import it, so the guard in `utils.spec.ts` is what keeps the fixture honest for every spec that renders it rather than only for its own copy. `notification.spec.tsx` also derives its negative testID from the constant instead of hardcoding `icon-rocket`, so a fixture change cannot leave a stale string asserting nothing.

Two atomic commits, and the branch is rebased onto current main (it was 3 behind). `tsc` and eslint clean; the notification-history and galoy-icon suites are green.

grimen (approved):

PR #4062, opened 2026-08-01
fix(splash): show a header back arrow when reached from Add account

### LINKED ISSUE #4056: fix(splash): Add arrow back to header on Splash screen
## Description

Add a back arrow to the header on the Splash screen when the user already has an existing account.

## Expected Behavior

- **Initial install (no existing account):** No back arrow shown on Splash screen
- **"Add Account" flow (existing account):** Back arrow displayed in header so user can navigate back

## Context

Small UI fix — the Splash screen currently doesn't show a way to go back when accessed through "Add Account", even though the user has an existing account to return to.

## Size

Small task

designsats:
## Summary

The Splash (`getStarted`) screen renders with `headerShown: false`, which is right on a
first install but wrong in the one flow that pushes it on top of an existing session:
**Settings → Switch account → Add account** ([`switch-account.tsx:64`](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/screens/settings-screen/account/multi-account/switch-account.tsx#L64)
calls `navigation.navigate("getStarted")`). The user lands on Splash with no way back to
the account list short of the OS back gesture.

This drives the header off `navigation.canGoBack()` rather than a hard-coded `false`.

`canGoBack()` is the correct discriminator here — not a "user has an account" flag —
because it distinguishes by how the screen was *entered*, which is exactly what decides
whether a back arrow has a destination:

| Entry point | Navigation call | `canGoBack()` | Header |
| --- | --- | --- | --- |
| First install | `initialRouteName="getStarted"` | `false` | hidden |
| Add account | `navigate("getStarted")` | `true` | **arrow shown** |
| Logout | `reset({ index: 0, … })` | `false` | hidden |
| Account deletion | `reset({ index: 0, … })` | `false` | hidden |
| Biometrics logout | `reset({ index: 0, … })` | `false` | hidden |
| Network-error recovery | `reset({ index: 0, … })` | `false` | hidden |

The arrow itself needs no new component — the root navigator's shared `screenOptions`
already supply `headerLeft: headerBackControl()`, so revealing the header is enough to get
the same arrow used on the adjacent Account-type and Terms screens.

## Changes

- **`app/screens/get-started-screen/get-started-screen.tsx`** — compute
  `canGoBack = navigation.canGoBack()` and set `headerShown` from it in a
  `useLayoutEffect` (before paint, so there is no flash of a headerless screen). The same
  flag is passed to `<Screen headerShown={…}>` so the top safe-area inset comes from the
  header instead of being applied a second time by the `SafeAreaView`.
- **`app/navigation/root-navigator.tsx`** — add `title: ""` to the `getStarted` route
  options. Without it the revealed header would render the route name; the static
  `headerShown: false` stays as the default for the first-install case.
- **`app/screens/authentication-screen/authentication-screen.tsx`** — the unlock screen's
  logout now runs `reset({ index: 0, routes: [{ name: "getStarted" }] })` instead of
  `replace("getStarted")`, matching every other logout/deletion path. A resume relock
  *pushes* the unlock screen on top of the live stack, so the replace left the stale
  session's routes beneath the splash — reachable by the OS back gesture before this PR,
  and advertised with a back arrow once the header follows `canGoBack()`.

## Acceptance requirements

- [ ] Fresh install, no account: Splash shows the logo and both buttons with **no header
      and no back arrow** — layout pixel-identical to before this change.
- [ ] Existing account, Settings → Switch account → **Add account**: Splash shows a header
      with a **back arrow**, matching the arrow style on the Account-type and Terms and
      Conditions screens.
- [ ] Tapping that back arrow returns to the **Accounts** list with the existing account
      still selected.
- [ ] After **logout** (Settings → Account → Logout): Splash shows **no** back arrow.
- [ ] After **account deletion**: Splash shows **no** back arrow.
- [ ] After logging out from the biometrics/PIN unlock screen: Splash shows **no** back
      arrow.
- [ ] Header background matches the screen background in both light and dark theme (it
      inherits the shared `headerStyle`, `colors.white`), with no shadow.
- [ ] The hidden developer menu (three taps on the logo) still opens in both states.

## Testing specs

| Spec | What it pins |
| --- | --- |
| `__tests__/screens/get-started-screen/get-started-screen.spec.tsx` › *hides the header on a first install, where there is no account to return to* | `canGoBack() === false` ⇒ `setOptions({ headerShown: false })` and `<Screen headerShown={false}>` |
| `__tests__/screens/get-started-screen/get-started-screen.spec.tsx` › *shows the header back arrow when opened over an existing account* | `canGoBack() === true` ⇒ `setOptions({ headerShown: true })` and `<Screen headerShown={true}>` |
| `__tests__/screens/authentication-screen/authentication-screen.spec.tsx` › *resets the stack to the splash, which then cannot lead back into the stale session* | unlock-screen logout ⇒ `reset({ index: 0, routes: [{ name: "getStarted" }] })`, never `replace` |
| `__tests__/screens/authentication-screen/authentication-screen.spec.tsx` › *stays put until the confirmation is accepted* | declining the logout confirmation neither logs out nor navigates |

```
yarn jest __tests__/screens/get-started-screen __tests__/screens/authentication-screen
```

Each new spec fails when its change is reverted — on `main` `setOptions` is never called
from the splash screen, and the unlock-screen logout ran `replace` instead of `reset`.
42/42 pass across both suites on this branch, with
`app/screens/authentication-screen/authentication-screen.tsx` at 100%
statements/branches/functions/lines.

Manual QA (run on the Android emulator, debug build):

1. Uninstall the app, install fresh, launch → Splash. **Expect no header, no back arrow.**
2. Create an account (any type) and land on Home.
3. Open the hamburger menu → **Switch account** → **Add account** → Splash.
   **Expect a back arrow in the header.**
4. Tap the back arrow. **Expect to return to the Accounts list.**
5. Go back to Home → Settings → Account → **Logout** → Splash.
   **Expect no back arrow.**
6. Repeat steps 1–4 with the theme set to light.

> No new component ships here, so there is no new component spec — the change is two
> options on an existing screen, covered by the two specs above.

## Notes / behavior changes

- Existing users see no change on app launch: the first-install and post-logout paths all
  reach this screen at stack index 0, where `canGoBack()` is `false` and the screen renders
  exactly as it does today.
- When the header *is* shown, the logo and buttons shift down by the header height. That is
  intended — it is the same inset every other headered screen in the app uses, and the
  logo's absolute positioning keeps it centred within the remaining space.
- No migration needed; no persisted state or protocol format is touched.

Closes #4056


grimen (changes requested):
## Findings

**1. Resume-lock logout shows the back arrow into the logged-out session (blocking).**
The entry-point table assumes the biometrics screen always sits at stack index 0, but that only holds for a cold start. When the app relocks on resume, `app-state.tsx:70` *pushes* `authenticationCheck` on top of the live stack (which is why `completeUnlock` uses `goBack()` for the resume case), so the Logout path at `authentication-screen.tsx:73` runs `navigation.replace("getStarted")` at index > 0 — `canGoBack()` is `true` and the splash advertises a back arrow whose destination is the previous account's stale, logged-out screens. This contradicts the acceptance requirement "After logging out from the biometrics/PIN unlock screen: Splash shows no back arrow".

*Example:* Log in with PIN/biometrics enabled, open Settings, background the app past the grace period, resume, tap Logout on the unlock screen and confirm — the splash renders with a back arrow, and tapping it returns to the Settings screen of the account that was just logged out.

*Proposal:* One option is to change that logout path to `navigation.reset({ index: 0, routes: [{ name: "getStarted" }] })`, matching every other logout/deletion path in the app (`logout.tsx`, `delete.tsx`, `profile.tsx`), which also removes the stale routes the OS back gesture could already reach before this PR.

grimen:
Addressed the blocking finding in aa6cb2063: the unlock screen's logout now runs `navigation.reset({ index: 0, routes: [{ name: "getStarted" }] })` instead of `replace`, matching every other logout/deletion path. On a resume relock — where the unlock screen is *pushed* on top of the live stack — this clears the stale session's routes, so `canGoBack()` is `false` and the splash shows no back arrow. It also removes the routes the OS back gesture could already reach before this PR.

The commit extends `authentication-screen.spec.tsx` with a spec pinning the reset (red on the previous head: `reset` was never called) plus a declined-confirmation guard, and backfills the screen's remaining uncovered paths (setup-run purpose, no-pin lock, failed prompt, dark logo) — the file now sits at 100% statements/branches/functions/lines. 42/42 tests pass across the auth and get-started suites; `tsc` and `eslint` clean.

grimen:
Before/after demo — Settings → Switch account → **Add account**, recorded on the iOS simulator against staging:

| --- | --- |
| Before: Add account lands on a splash with no header and no way back | After: the splash shows a back arrow that returns to the Accounts list |

In the before recording the taps in the top-left corner are deliberate: there is nothing there to tap — the OS back gesture is the only way out. In the after recording the same spot carries the back arrow, and tapping it returns to the Accounts list with the existing account still selected.

grimen (approved):

PR #4063, opened 2026-08-01
fix(transaction-detail): stop the amount heading dropping its last word

### LINKED ISSUE #4058: Bug: "Spent" gets hidden
**Describe the bug**

Navigating to a transaction detail screen twice hides the "spent" text.

First visit shows the "You spent" message in the header: 

Image

Second visit show the "You" message in the header with a blank space where "spent" would've been.

Image

This isn't the same on transaction that have been received.

**To Reproduce**

1. Go to 'home screen'
2. Click on 'swipe up to show transaction history'
3. Click a transaction where the wallet was spending
4. Check "You spent" message
5. Go back to transaction history list
6. Click the same transaction to view it's details again
7. Check for "You" message or "You spent" message.

**Expected behavior**
Transaction should show "You spent" consistently.

Android (3.0.16)

kngako:
The heading intermittently rendered "You spent" as "You". It was not a copy problem: every locale returns the full string through the runtime, and the surrounding labels stayed intact in the broken frame, so LL was healthy rather than falling back.

The Text was measured as a single line, then re-broken after the first word on a later layout pass. Since the container height was already fixed by the first measure, the wrapped word landed outside it and was clipped — same height, missing word. It only hit this element because it was the one unbounded-line Text in the header.

Pin the heading to one line so the re-break cannot happen. The longest heading in any locale is 19 characters, so it fits without ellipsizing.

### DIFF
--- a/__tests__/helpers/transaction-detail-mocks.tsx
+++ b/__tests__/helpers/transaction-detail-mocks.tsx
@@ -1,0 +1,128 @@
+import * as React from "react"
+
+/**
+ * Shared mocks for the transaction-detail-screen specs.
+ *
+ * The screen pulls in the whole themed/hook stack, so every spec that renders
+ * it needs the same stubs. Keeping them here means a change to the screen's
+ * hooks is a one-line change instead of the same edit in three files.
+ *
+ * Pull each one in from inside the factory, the way card-flow-mocks is used:
+ *
+ *   jest.mock("@rn-vui/themed", () =>
+ *     jest.requireActual("../helpers/transaction-detail-mocks").mockThemedText(),
+ *   )
+ *
+ * Importing these at the top of a spec instead looks tidier but breaks. A
+ * factory fires the first time its module is required, which happens midway
+ * through importing the screen under test — so any binding the spec declares
+ * after that import is still in its TDZ. That is also why mockAppHooks takes
+ * a thunk rather than a spy.
+ */
+
+export const TEST_GALOY_INSTANCE = {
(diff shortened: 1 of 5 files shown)


grimen (commented):
Superseded by the request-changes review below (same findings).

grimen (changes requested):
## Findings

**1. Long locale strings can still ellipsize under font scaling (blocking).**
`numberOfLines={1}` removes the silent word-drop, but when the line genuinely doesn't fit the default behavior is tail ellipsis, so the heading can still lose characters visibly.
*Example:* the Malay heading `Anda dah belanjakan` (19 chars, the longest across all 28 locales) at a large accessibility font scale on a narrow device renders as `Anda dah belanja…`.
*Proposal:* one option could be adding `adjustsFontSizeToFit` (or a `maxFontSizeMultiplier`) so the text shrinks instead of ellipsizing.

## Test Coverage Gaps

- The third heading branch, `sending` (on-chain transaction not yet broadcast), has no assertion — one more case with the pending status would cover all three copies.

## Code Smells

- The ~100-line mock preamble is now copy-pasted across three transaction-detail specs (`transaction-detail-fallback`, `transaction-detail-row-actions`, and this one); a shared local helper would keep the next change to the screen's hooks from breaking all three in triplicate.

kngako:
Result of PR before review suggestions (using dummy text): 

image

With the suggested changes:

image

---

The real change looks like this:

image

grimen (approved):

PR #4065, opened 2026-08-03
fix(send): show the send-completed time in the device timezone

### LINKED ISSUE #3896: fix(mobile): show send-completed transaction time in local timezone
## Summary

A user reported that transaction times should account for the country/local timezone. They shared a screenshot and wrote: "some changes on difference in time based on a country would be nice".

From code inspection, most transaction timestamps already render in the device's local timezone:

- `app/components/transaction-date/transaction-date.tsx` uses `new Date(createdAt * 1000).toLocaleString(locale, options)` with no forced timezone when rendering transaction detail times.
- `app/components/transaction-date/transaction-date.tsx` / `app/utils/date.ts` also use local `Date` methods for transaction list relative dates and short dates unless a timezone is explicitly injected for tests.

However, the send-completed receipt screen still formats the time through `formatUnixTimestampYMDHM`, which uses `toISOString()` and therefore forces UTC:

- `app/utils/date.ts`: `formatUnixTimestampYMDHM(timestampInSeconds)` returns `new Date(...).toISOString().slice(0, 16).replace("T", " ")`
- `app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx` displays that value in the `Time` row.

## Current behavior

On the send-completed screen, the `Time` row is displayed in UTC because `toISOString()` is used.

## Expected behavior

Display the send-completed `Time` row in the user's local/device timezone, consistent with transaction detail/list behavior. Ideally use locale-aware formatting, or add an explicit timezone label if product wants a fixed timezone.

## Notes

- Older related issues exist but are closed/broader: #2938 and #1106.
- Add/adjust tests around `formatUnixTimestampYMDHM` or replace it with a local-time formatter that can accept an injected timezone for deterministic tests.

esaugomez31:
## Summary

The send-completed receipt formatted its `Time` row through `formatUnixTimestampYMDHM`, which built the string from `toISOString()` and therefore always rendered UTC. Every other transaction surface already renders in the device timezone (`transaction-date.tsx` and `formatShortDate` both pass through `toLocaleString`/`toLocaleDateString` with no forced zone), so the receipt was the only screen disagreeing with the phone clock: a payment made at 13:10 local showed as 19:10.

This renders the row in the device timezone while keeping the `YYYY-MM-DD HH:mm` shape.

Closes #3896

## Changes

- `app/utils/date.ts`: `formatUnixTimestampYMDHM` now takes `{ timestampSeconds, timezone? }` and builds the string from two halves, `formatShortDate` for the date and `toLocaleTimeString` for the time, instead of slicing an ISO string. Reusing `formatShortDate` keeps the receipt on the exact date formatting the transaction list already ships, so the function moved below it in the file. Each half borrows the locale whose shape it needs, `en-CA` for the `YYYY-MM-DD` order and `en-GB` for the 24 hour clock, and the two are joined by hand so the result does not depend on the separator a locale would pick between date and time. `hourCycle: "h23"` keeps midnight as `00` rather than `24`. The optional `timezone` follows the convention already used by `formatShortDate` and `formatDayAndMonth` in the same module, so tests can pin a zone; production callers omit it and get the device zone. Separately, the unreachable `totalSeconds < 0` guard was dropped from `secondsToDDMMSS`: its only caller `getTimeLeft` already returns `""` whenever the window is closed, so the branch could never run and was the last thing keeping the module short of full coverage.
- `app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx`: updated the single call site to the object signature and pulled the ternary out of the JSX prop into a named `formattedTime`.
- `jest.timezone-environment.js`: a `jest-environment-node` subclass that pins the process timezone for a single spec file and restores the runner's zone on teardown. Jest hands every spec a proxy over a copy of `process.env`, so a spec that assigns `TZ` itself never reaches Node's real environment and neither `Date` nor `Intl` changes zone; pinning it out here, in the worker process that owns the real env, is what makes a device-timezone assertion fail on a UTC runner. `jest-environment-node` is declared in `package.json` because the environment imports it directly, at the `^29.7.0` already keyed in the lockfile, so `yarn.lock` is unchanged.
- `__tests__/utils/date-device-timezone.spec.ts`: the default path, opted into that environment through a `@jest-environment` docblock, so a `19:04Z` timestamp has to render as `13:04`. Keeping it in its own file leaves the pinned zone off the rest of the date suite, and doing it per file rather than through a `TZ=` prefix in the test script means the guard holds under a plain `yarn test`, which is what CI runs.
- `__tests__/utils/date.spec.ts`: seven cases for the formatter with the zone injected (UTC, negative and positive offsets, a day rollover in each direction, midnight and zero padding), plus the cases needed to cover the rest of the module: `getTimeLeft` across its countdown segments and both closed-window paths, the default mask character, and the default locale.
- `__tests__/screens/send-bitcoin-completed-screen.spec.tsx`: the time row renders a wall-clock date, the time and type rows disappear without their values, the close and share buttons fire, and the header drops out of the capture while sharing.

The regression guard is split across the two date specs on purpose. `date.spec.ts` pins a zone through the formatter's own argument, which covers the shape of the output but leaves the default path unguarded: CI resolves the process zone to UTC, so any expectation derived from it compares UTC against UTC and would keep passing even if the default regressed to a hardcoded UTC. `date-device-timezone.spec.ts` closes that hole by pinning a non-UTC process zone instead. With the default path regressed to a hardcoded UTC, it fails under `TZ=UTC` with `Expected "2026-08-03 13:04", Received "2026-08-03 19:04"` while all 75 injected-zone cases still pass, and a spec running after it in the same `--runInBand` process still sees `UTC`, so the pin does not leak. The screen spec asserts the rendered shape only, since a screen-level test cannot pin the process timezone.

## Test plan

- `yarn jest __tests__/utils/date.spec.ts __tests__/utils/date-device-timezone.spec.ts __tests__/screens/send-bitcoin-completed-screen.spec.tsx`: 96 passed, no console output, `date.ts` at 100% statements, branches, functions and lines
- `yarn jest __tests__/utils __tests__/screens/send-bitcoin-completed-screen.spec.tsx`: 563 passed across 35 suites, the wider run that covers every consumer of the module
- The same runs under `TZ=UTC`, to reproduce what CI resolves the process zone to
- `yarn tsc:check` and `eslint` on the changed files: clean. `yarn install --frozen-lockfile`: passes, `yarn.lock` untouched
- Device (Android emulator): sent a 100 sat Lightning payment before and after the change. The status bar clock in both screenshots is the ground truth.

## Screenshots

| Before | After |
| :---: | :---: |
| 3896-1-before | 3896-2-after |

Compare each receipt against the status bar clock of that same screenshot. Before, the phone reads `1:04` while the receipt says `19:04`, six hours ahead because the emulator sits at UTC-6. After, the phone reads `1:10` and the receipt says `13:10`, the same wall clock.

The two columns are separate payments, so the minutes differ between them; the clock in each screenshot is the reference, not the other column.

### DIFF
--- a/__tests__/screens/send-bitcoin-completed-screen.spec.tsx
+++ b/__tests__/screens/send-bitcoin-completed-screen.spec.tsx
@@ -22,6 +22,23 @@ import { ContextForScreen, ContextForScreenWithTheme } from "./helper"
 import { Linking, View, ViewStyle } from "react-native"
 import { light, dark } from "@app/rne-theme/colors"
 
+const screenshotState = { isTakingScreenshot: false }
+const mockCaptureAndShare = jest.fn()
+const mockNavigate = jest.fn()
+
+jest.mock("@app/hooks", () => ({
+  ...jest.requireActual("@app/hooks"),
+  useScreenshot: () => ({
+    isTakingScreenshot: screenshotState.isTakingScreenshot,
+    captureAndShare: mockCaptureAndShare,
+  }),
+}))
+
+jest.mock("@react-navigation/native", () => ({
+  ...jest.requireActual("@react-navigation/native"),
+  useNavigation: () => ({ navigate: mockNavigate }),
+}))
+
 jest.mock("react-native-view-shot", () => {
   return {
     __esModule: true,
@@ -92,6 +109,9 @@ describe("SendBitcoinCompletedScreen", () => {
   beforeEach(() => {
     loadLocale("en")
     LL = i18nObject("en")
+    screenshotState.isTakingScreenshot = false
+    mockCaptureAndShare.mockClear()
+    mockNavigate.mockClear()
   })
 
   it("renders the Success state correctly", async () => {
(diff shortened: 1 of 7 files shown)


grimen (changes requested):
## Findings

**1. The device-zone default test can't fail on a UTC runner.**
The default-path case compares `formatUnixTimestampYMDHM({ timestampSeconds })` against the same call with the resolved device zone injected, and CI resolves the process zone to UTC.
*Example:* if the default path later regresses to a hardcoded UTC fallback while the `timezone` parameter keeps working, every case in `date.spec.ts` still passes on CI, because "device zone" and "UTC" are the same string there.
*Proposal:* one option could be pinning a non-UTC zone for this one case — `process.env.TZ = "America/El_Salvador"` at the very top of a small separate spec file, before any `Date`/`Intl` use, or running the date spec under `TZ=...` in the test script.

## Code Smells

- `date.ts:36`: the `totalSeconds < 0` guard in `secondsToDDMMSS` is unreachable, since its only caller `getTimeLeft` already returns `""` whenever `Date.now() > until`, and it's the single branch keeping the module short of the full coverage this PR otherwise reaches. It predates the PR, so deleting it, or moving the check into `getTimeLeft`, is fair to call out of scope.

esaugomez31:
Both addressed, in `d8368437f` and `622af5bcd`.

## 1. The device-zone default test

You were right that the case was a tautology on CI, and fixing it turned out to need a different mechanism than the one in the proposal: **jest 29 hands every spec a proxy over a *copy* of `process.env`** (`jest-util/build/createProcessObject.js`), so an assignment inside a test file never reaches Node's real environment and neither `Date` nor `Intl` changes zone. I wrote your version first, pinning `America/El_Salvador` at the top of the case, and it still returned `2026-08-03 19:04` under `TZ=UTC` — it would have kept passing just as silently as the old one.

The zone has to be pinned in the worker process, before the sandbox is built. So:

- `jest.timezone-environment.js` extends `jest-environment-node`, pins `America/El_Salvador` in the constructor, and restores the runner's zone in `teardown`.
- `__tests__/utils/date-device-timezone.spec.ts` opts in through the `@jest-environment` docblock and expects `2026-08-03 13:04` for a `19:04Z` timestamp.

Doing it per file rather than through `TZ=` in the test script means the guard holds under a plain `yarn test`, which is what CI runs, instead of depending on the runner exporting the right zone.

Verified three ways:

| Scenario | Result |
| --- | --- |
| Under `TZ=UTC`, as on CI | passes, so the pin really applies |
| Default path regressed to a hardcoded UTC, under `TZ=UTC` | **fails** with `Expected "2026-08-03 13:04", Received "2026-08-03 19:04"`, while all 75 injected-zone cases still pass |
| A spec running after it in the same `--runInBand` process | still sees `UTC`, so the pin does not leak |

The middle row is the one you asked for: that regression is invisible to every other case in the suite.

`jest-environment-node` is now declared in devDependencies, since the environment imports it directly and `import/no-extraneous-dependencies` requires it. `^29.7.0` was already a key in the lockfile, so `yarn.lock` is unchanged and `yarn install --frozen-lockfile` passes.

## 2. The unreachable `secondsToDDMMSS` guard

Deleted, in its own commit. Moving the check into `getTimeLeft` would only duplicate the `dateNow > until` that is already there: `getTimeLeft` owns window validity, `secondsToDDMMSS` owns formatting, and both callers (`may-challenge/card.tsx`, `june-challenge/card.tsx`) reach the helper only through `getTimeLeft`. `date.ts` is now at 100% branches, so the module is fully covered.

## Checks

563 tests across all of `__tests__/utils` plus the send-completed screen pass, `date.ts` at 100% statements, branches, functions and lines, `tsc:check` and eslint clean.

grimen (approved):

PR #4066, opened 2026-08-03
fix(self-custodial): surface why a fee quote failed instead of a dead end

### LINKED ISSUE #4059: Bug: Inconsistent "Unable to calculate fee" messaging
**Describe the bug**
When sending transactions to some destinations the app shows an "unable to calculate fee" error message... which would require the user to go back and try again for another attempt to go through.

Not sure if the message also means unable to find route?

**To Reproduce**

Seems it happens more often for transactions over 150 000 sats

**Expected behavior**
It should be able 

**Screenshots**

Image

Android (3.0.16)

kngako:
Every self-custodial `getFee` swallowed the SDK error in a bare `catch` and
returned `{ amount: undefined }`, which the confirmation screen renders as the
generic "Unable to calculate fee" with the slide-to-confirm disabled. The most
common trigger is benign and fixable by the user: the SDK defaults to
`FeesExcluded`, so a send at or near the full balance throws InsufficientFunds
during `prepareSendPayment`. The user was told nothing and had no way forward.

The failures were also invisible in error reporting. `reportError` in `use-fee`
only fires on a thrown error, and self-custodial `getFee` never throws, so none
of this reached Crashlytics. The send path already reported via
`reportSendFailure`; only the fee path was silent.

`classifySdkError` and the `SelfCustodialError` strings already existed and are
used for the same SDK errors on the details screen — this routes the fee path
through them too.

- classify and report in `createGetFee`, `createGetFeeOnchain` and the LNURL
  fee quote, returning the code in `errors`. Covers the `extractOnchainFees`
  null branch, which previously returned no signal at all.
- thread `errors` through `useFee` into the "error" fee state.
- translate the code on the confirmation screen via `useTranslateSdkError`,
  falling back to the generic string.

Custodial behaviour is unchanged. Custodial `errors` carry raw GraphQL text
that is not fit to show, and `useTranslateSdkError` falls through to `return
raw` on unknown input, so `isSelfCustodialErrorCode` gates the substitution to
classified self-custodial codes only.

### DIFF
--- a/__tests__/screens/send-bitcoin-screen/use-fee.spec.ts
+++ b/__tests__/screens/send-bitcoin-screen/use-fee.spec.ts
@@ -1,0 +1,134 @@
+import { renderHook, waitFor } from "@testing-library/react-native"
+
+import { WalletCurrency } from "@app/graphql/generated"
+import useFee from "@app/screens/send-bitcoin-screen/use-fee"
+import { SelfCustodialErrorCode } from "@app/self-custodial/sdk-error"
+import { toBtcMoneyAmount } from "@app/types/amounts"
+
+// use-fee only calls these to hand them to getFee; the self-custodial quotes ignore them.
+// The tuple is built once and shared: it lands in the fee effect's dependency array, so a
+// fresh identity per render would re-fire the effect forever (Apollo's is stable for real).
+jest.mock("@app/graphql/generated", () => {
+  const probe = [jest.fn(), {}]
+  return {
+    WalletCurrency: { Btc: "BTC", Usd: "USD" },
+    useLnInvoiceFeeProbeMutation: () => probe,
+    useLnNoAmountInvoiceFeeProbeMutation: () => probe,
+    useLnUsdInvoiceFeeProbeMutation: () => probe,
+    useLnNoAmountUsdInvoiceFeeProbeMutation: () => probe,
+    useOnChainTxFeeLazyQuery: () => probe,
+    useOnChainUsdTxFeeLazyQuery: () => probe,
+    useOnChainUsdTxFeeAsBtcDenominatedLazyQuery: () => probe,
+  }
+})
+
+const mockRecordError = jest.fn()
+jest.mock("@react-native-firebase/crashlytics", () => () => ({
+  recordError: (...args: unknown[]) => mockRecordError(...args),
+  log: jest.fn(),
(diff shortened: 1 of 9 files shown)


grimen (changes requested):
## Findings

**1. Every classified fee failure records a Crashlytics non-fatal, including the expected full-balance case.**
`feeFailure` calls `reportError` with no options, and `recordAppError` classifies any non-connectivity error as a Defect, so each failed quote fires `crashlytics().recordError`. The PR body itself calls the most common trigger — `InsufficientFunds` on a full-balance send — benign and user-fixable, which is exactly the "expected user state" the `error-reporting.ts` boundary doc says should stay a breadcrumb rather than non-fatal noise.

*Example:* a user tries to send their full balance, reads the new message, lowers the amount and succeeds — yet a non-fatal defect lands in Crashlytics for every such attempt, from every user, on a flow that worked as intended.

*Proposal:* one option is to pass `expected: true` (or at least a `dedupKey`) when the classified code is `InsufficientFunds`/`BelowMinimum` and keep plain `recordError` for the unexplained codes; if full visibility is deliberate, a note on `feeFailure` saying so would help.

## Code Smells

- The test name "records fee failures to crashlytics with a scoped Error" in `send-helpers.spec.ts` asserts the *unscoped* message `"prepare refused"` — the scope only survives for non-`Error` throws, so the name oversells what is verified.
- `errors?` was added to the `"loading" | "error" | "unset"` arm of `FeeType` even though loading/unset never carry errors; it's needed only because `"error"` appears in two union arms. Narrowing the first arm to `"loading" | "unset"` and letting the existing error arm own `errors` would keep the states honest (pre-existing structure, but this PR widens it).

kngako:
The fee errors are a little hard to test. But managed to test a network error. 

image

Visually it looks good. 

---


For the fee path after the change, that splits four ways:

| Classified code | `expected` | Class | `recordError`? |
|---|---|---|---|
| `InsufficientFunds` | `true` | Expected | **no** — breadcrumb |
| `BelowMinimum` | `true` | Expected | **no** — breadcrumb |
| `InvalidInput` | `false` | Defect | yes |
| `Generic` | `false` | Defect | yes |
| `NetworkError` | `false` | Transient | no — message matches `/network ?error/i` |

grimen (approved):

PR #4067, opened 2026-08-03
fix(contacts): list self-custodial transactions on the contact screen

esaugomez31:
## Summary

The contact detail screen fetched its transactions through `me { contactByUsername { transactions } }` with no account-type guard. A self-custodial account has no session behind `me`, so the query answered with nothing, the component returned early on `!transactions`, and the "Transactions with …" section rendered as an empty hole with no list and no empty state.

The contact adapter already had the answer: `ContactAdapter.getTransactions` is implemented in `use-self-custodial-contacts` and matches payments by the counterparty lightning address. Nothing was calling it. This wires the screen to the adapter, the same way `all-contacts.tsx` already resolves its list per account type, and gives that path real pagination instead of a single truncated read.

The screen itself was then redesigned to the current spec: the stack's own back arrow instead of the close cross, the shared `IconHero` carrying the lightning address, and the history as a rounded date group with the send button under it.

Refs blinkbitcoin/blink-wip#936

The report suggests hiding the People tab instead. This repairs it rather than hiding it, so the tab keeps working while that call is made: whether to rework the UI can be decided later, on a screen that is no longer broken.

## Changes

- `app/self-custodial/providers/payments-page.ts` (new): the raw page primitive — `TRANSACTIONS_PER_PAGE` and `fetchAndMapPayments` — lifted out of `wallet-snapshot.ts`, which kept it private, so the wallet snapshot and the new contact paging both depend on it and neither depends on the other. Pure move, no behaviour change.
- `app/self-custodial/providers/contact-payments.ts` (new): reads a contact's payments a page at a time. The SDK cannot filter by counterparty, so matching happens on the client over the `lnAddress` the mapper already carries, in either direction, and the loop keeps pulling raw pages until it fills one, exhausts the history, or reaches a page budget.
- `app/types/contact.ts`: `getTransactions` is keyed by payment identifier rather than contact id, and returns a page whose cursor is opaque to the caller.
- `app/hooks/use-contact-transactions.ts` (new): owns the paging state. A page from a superseded read is dropped rather than appended, repeats are deduped by id, and a failed page leaves the list intact.
- `app/screens/people-screen/contacts/contact-transactions.tsx`: pages through the adapter for a self-custodial account, with `onEndReached` serving both paths. The list always renders, so an account with no matching payments gets the empty state instead of a blank area. Restyled to the design: the date is plain text above a rounded grey group rather than a bar inside a bordered box, and each row carries its relative time. It also holds the room between the header and the send button whatever the history holds — one payment or twenty — so the button keeps its place instead of riding up a short list, including when the history cannot be read at all.
- `app/screens/people-screen/contacts/contacts-detail.tsx`: rebuilt around `IconHero` with the contact's lightning address, the list, and the send button. The 86pt icon, the alias input, the "Username: …" line and the "Transactions with …" title are gone, and the close cross gives way to the header back arrow the rest of the stack uses.
- `app/components/icon-hero/icon-hero.tsx`: an optional `titleLines`, which the contact screen sets to 1. An address is one unbreakable thing, so it is cut short with the middle ellipsis the app already uses for a payment destination rather than wrapped across two lines. A heading with no cap still wraps, so the onboarding screens are untouched.
- `app/navigation/root-navigator.tsx`: `contactDetail` shows the stack header with an empty title, which is where that back arrow comes from.
- `app/custodial/contact-adapter.ts`: takes over the `userContactUpdateAlias` document, its only remaining caller now that the screen no longer edits aliases. `generated.ts` is unchanged.
- `app/components/transaction-item/transaction-item.tsx`: the date under a description is `grey2`, the colour `card-transaction-item` already uses for the same line. Received amounts were already `_green`.
- `app/self-custodial/hooks/use-self-custodial-transaction-fragments.ts` (new): the mapping into `TransactionFragment` plus the Apollo cache priming, extracted from the transaction history screen because both screens need both halves. `transaction-history-screen.tsx` now consumes it, unchanged in behaviour.

## Notes for review

The paging cursor is an offset into a live wallet history, so returning to the screen restarts from the first page rather than merging across a boundary a new payment may have shifted.

Contact matching runs in both directions, since a conversation with a contact is what they were paid and what they paid back. In practice only sends surface today: verified on device, an incoming Lightning payment carries no `lnAddress` at all — the SDK's receive-side record (`LnurlReceiveMetadata`) holds a nostr zap request and a sender comment, and no identity — so a receive has no counterparty to match against. The matcher is ready for the day one arrives; nothing client-side can conjure it.

A separate, pre-existing backend limit affects the custodial path, tracked in #4193. `contactByUsername` is the only lookup the schema offers, and it resolves only contacts stored under a username. A contact stored as an address answers `NoContactForUsernameError`, and passing the address itself fails the `Username` scalar, so that contact's history cannot be listed at all. It needs a handle-keyed lookup on the backend; once that ships, the app side is a one-field change — query `contactByHandle(handle: $handle)` and pass `contact.handle` — which belongs in its own PR rather than here.

## Test plan

- Coverage on every file this touches: **100% statements, branches, functions and lines** across `contacts-detail.tsx`, `contact-transactions.tsx`, `icon-hero.tsx`, `use-contact-transactions.ts`, `contact-payments.ts`, `payments-page.ts`, `use-self-custodial-contacts.ts`, `use-self-custodial-transaction-fragments.ts` and `contact-adapter.ts` — 288 tests in 16 suites, no console output
- Every guard in the paging hook was verified by mutation, each broken on purpose to confirm a test fails
- `yarn jest __tests__/hooks/ __tests__/custodial/ __tests__/self-custodial/ __tests__/screens/people-screen/`: 2085 passed in 161 suites
- `yarn jest` on the transaction-history and wallet-snapshot suites, the ones the extraction could reach: 47 passed
- `yarn check-code` (the CI command: tsc, eslint, translations, codegen, graphql validation): clean
- Device (Android emulator): custodial and self-custodial contacts, pagination exercised by temporarily lowering the page size so a short history spans several pages, and both design notes checked against a synthetic long address — the address truncates on one line, and the send button holds the same position with one payment and with a full month

## Screenshots

| Before | After — custodial | After — self-custodial |
| :---: | :---: | :---: |
| 936-1-before | Screenshot from 2026-08-21 11-26-27 | Screenshot from 2026-08-21 11-27-12 |

Before: the section renders its title and then nothing at all. After: the contact's payments are listed and grouped by date, under the redesigned header. Amounts are masked by the app's hide-balance toggle.

### DIFF
--- a/__tests__/components/icon-hero.spec.tsx
+++ b/__tests__/components/icon-hero.spec.tsx
@@ -26,6 +26,37 @@ describe("IconHero", () => {
     expect(getByText("Test Title")).toBeTruthy()
   })
 
+  it("lets a title wrap by default, because a heading is a sentence", async () => {
+    const { getByText } = render(
+      <ContextForScreen>
+        <IconHero icon="cloud-arrow-up" iconColor="green" title="Test Title" />
+      </ContextForScreen>,
+    )
+    await flushEffects()
+
+    expect(getByText("Test Title").props.numberOfLines).toBeUndefined()
+  })
+
+  it("cuts a title short when the caller caps its lines", async () => {
+    // A lightning address is one unbreakable thing: wrapping it mid-address reads as two
+    // addresses, so a screen showing one asks for a single line instead.
+    const { getByText } = render(
+      <ContextForScreen>
+        <IconHero
+          icon="cloud-arrow-up"
+          iconColor="green"
+          title="deepbassoon958@walletofsatoshi.com"
+          titleLines={1}
+        />
+      </ContextForScreen>,
+    )
+    await flushEffects()
+
+    const title = getByText("deepbassoon958@walletofsatoshi.com")
+    expect(title.props.numberOfLines).toBe(1)
+    expect(title.props.ellipsizeMode).toBe("middle")
+  })
+
   it("renders subtitle when provided", async () => {
     const { getByText } = render(
       <ContextForScreen>
--- a/__tests__/custodial/contact-adapter.spec.ts
+++ b/__tests__/custodial/contact-adapter.spec.ts
@@ -128,12 +128,12 @@ describe("useCustodialContactAdapter", () => {
     )
   })
 
-  it("returns an empty transactions list (custodial does not back this)", async () => {
+  it("returns an exhausted empty page (the contact query backs this instead)", async () => {
     const { result } = renderHook(() => useCustodialContactAdapter())
 
-    const txs = await result.current.getTransactions("c1")
+    const page = await result.current.getTransactions("c1")
 
-    expect(txs).toEqual([])
+    expect(page).toEqual({ transactions: [], nextCursor: null })
   })
 
   it("propagates the loading flag from the underlying query", () => {
--- a/__tests__/custodial/contact-adapter.spec.ts
+++ b/__tests__/custodial/contact-adapter.spec.ts
@@ -120,6 +120,20 @@ describe("useCustodialContactAdapter", () => {
     expect(mockRefetch).toHaveBeenCalled()
   })
 
+  it("answers with an empty list when the refetch comes back without contacts", async () => {
+    // A rename whose refetch answers with nothing leaves the caller an empty list to
+    // render rather than reading `contacts` off undefined data.
+    mockRefetch.mockResolvedValue({ data: undefined })
+    const { result } = renderHook(() => useCustodialContactAdapter())
+
+    let updated: Awaited<ReturnType<typeof result.current.update>> | undefined
+    await act(async () => {
+      updated = await result.current.update("c1", { displayName: "New Alice" })
+    })
+
+    expect(updated).toEqual({ contacts: [] })
+  })
+
   it("rejects updating an unknown contact id", async () => {
     const { result } = renderHook(() => useCustodialContactAdapter())
 
--- a/__tests__/hooks/use-contact-transactions.spec.ts
+++ b/__tests__/hooks/use-contact-transactions.spec.ts
@@ -80,6 +80,55 @@ describe("useContactTransactions", () => {
     expect(result.current.isLoading).toBe(true)
   })
 
+  describe("a page that holds nothing for this contact", () => {
+    it("keeps reading, because an empty list has no scroll to ask with", async () => {
+      mockGetTransactions
+        .mockResolvedValueOnce(page([], "20"))
+        .mockResolvedValueOnce(page([], "40"))
+        .mockResolvedValueOnce(page(["tx-1"], null))
+
+      const { result } = renderContactTransactions()
+      await flushEffects()
+
+      expect(mockGetTransactions).toHaveBeenCalledTimes(3)
+      expect(mockGetTransactions).toHaveBeenLastCalledWith(IDENTIFIER, "40")
+      expect(result.current.transactions).toEqual([{ id: "tx-1" }])
+    })
+
+    it("reads as loading while it is still searching", async () => {
+      mockGetTransactions.mockResolvedValueOnce(page([], "20"))
+      mockGetTransactions.mockReturnValue(new Promise(() => {}))
+
+      const { result } = renderContactTransactions()
+      await flushEffects()
+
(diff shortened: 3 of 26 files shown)


grimen (commented):
## Findings

**1. The per-request page budget is defeated for contacts with no matches — the hook auto-walks the entire wallet history.**
`MAX_PAGES_PER_REQUEST` caps one `fetchContactPaymentsPage` call at 5 raw pages, but the `hasNothingToScroll` effect in `use-contact-transactions.ts` immediately issues the next request whenever the list is still empty and a cursor remains, so the budget only splits the walk into rounds rather than bounding it.

*Example:* a self-custodial wallet with 10,000 payments and a contact the user never paid triggers roughly 500 sequential `listPayments` bridge calls (plus mapping) behind a spinner before the empty state appears.

*Proposal:* this may be the intended trade-off ("settles on the empty state once the history runs out") — if so a comment noting the worst case would help; otherwise one option is a total auto-read budget in the hook, after which the empty state renders even though `hasMore` was still true.

**2. The error toast fires again on every re-render while the error state persists.**
`contact-transactions.tsx` calls `toastShow` in the render body when `hasTransactionsError` is set, and the new `hasError` from `useContactTransactions` stays true until the next focus, so any parent re-render (alias edit on the detail screen, theme change) queues another toast. The component test only proves a single toast because it renders once.

*Example:* the first self-custodial page fails, the user taps into the alias edit field on the same screen, and each keystroke's re-render shows a fresh "Error loading transactions" toast.

*Proposal:* the Apollo path had this pattern before the PR, so it may be acceptable to keep — but since the PR already restructures this block, one option is moving the toast into a `useEffect` keyed on the error flag so it fires once per failure.

## Code Smells

- `useContactTransactions` gates its first read on `contactsLoading`, but after this PR `getTransactions` is keyed by payment identifier and never touches the contact list — the wait is really a proxy for SDK readiness, and a slow or failed contact-list fetch now delays a page that doesn't need it. Since the self-custodial `getTransactions` identity changes when `sdk` does (and the focus effect re-runs on it), the gate may be droppable.
- `fetchContactPaymentsPage` appends all matches from the raw page that crosses the threshold, so a "page" can hold up to ~2× `TRANSACTIONS_PER_PAGE` items (e.g. 19 matched + a fully-matching final raw page = 39). Harmless to the consumer, but worth a comment since the type reads like a fixed-size page.

grimen (approved):

grimen (changes requested):
See: https://github.com/blinkbitcoin/blink-wip/issues/936#issuecomment-5181233533

esaugomez31:
@grimen

## Findings

**1. The per-request page budget is defeated for contacts with no matches — the hook auto-walks the entire wallet history.**
Kept, documented. It is the intended trade-off: stopping early would claim the contact has no transactions while history is still unread, and an empty list leaves no scroll to resume the search from, so a total budget trades a slow correct answer for a fast wrong one. The comment on the auto-read effect now states the worst case in your terms — one round per 100 raw payments, some 100 rounds behind the spinner for a 10,000-payment wallet.

**2. The error toast fires again on every re-render while the error state persists.**
Fixed. It moved into a `useEffect` keyed on the error flag, matching `use-shipping-address-data.ts`, which already does this. New test renders three times with the flag up and asserts a single toast; confirmed it fails against the old render-body version. Note the previous test could not have caught this even if written: re-rendering the same element makes React bail out, so the spec now builds a fresh one per render, and `LL` is resolved once in the mock so it carries the stable identity the real context hands out.

## Code Smells

- Dropped the `contactsLoading` gate. You were right that it is a proxy that no longer holds: it does not even cover the case it documented, since the adapter reports `loading: false` immediately when there is no SDK. Readiness now rides entirely on `getTransactions` changing identity with `sdk`, which re-runs the focus effect; there is a test for that reconnect path.
- Documented the overshoot on `fetchContactPaymentsPage`: `TRANSACTIONS_PER_PAGE` is a floor, not a cap, and callers append and dedupe by id.

Four atomic commits. `tsc` and eslint clean; the contacts and self-custodial provider suites are green.

grimen (approved):

PR #4068, opened 2026-08-03
fix(home): show the unseen transaction badge for self-custodial accounts

esaugomez31:
> ### Stacked on the contacts PR
>
> This branch sits on top of `fix--contact-transactions-self-custodial`, because it reuses the `useSelfCustodialTransactionFragments` hook introduced there. Review that one first.

## Summary

The home screen only ever fed the unseen-transaction badge with custodial data: `transactions` came from `me.defaultAccount` and the seen state was keyed by the custodial account id. A self-custodial account has no `me` behind those, so `useTransactionSeenState` saw an empty list, `hasUnseenBtcTx` and `hasUnseenUsdTx` stayed false, and a received payment produced no bubble on the currency pill and no amount under the balance.

This sources both inputs per account type, and stores the seen state where a self-custodial account can actually keep it.

It also fixes the badge announcing the same payment again on every launch, reported on this PR after the first round of review. The seen state was only written once the display window had elapsed, so closing the app before that lost the write and the next launch announced the same transaction again.

Closes blinkbitcoin/blink-wip#925

## Changes

**Feeding the badge**

- `app/components/unseen-tx-amount-badge/use-unseen-tx-badge-state.ts` (new): owns everything the home screen needs for the badge. It builds the custodial list from the home query, maps the wallet's transactions through `useSelfCustodialTransactionFragments` for a self-custodial account, and returns the badge, the auto-seen wiring and the visibility flags. Mapping also primes the Apollo cache, which is what lets the transaction detail behind the badge load when it is opened from home.
- `app/screens/home-screen/home-screen.tsx`: calls that hook once instead of assembling the badge from five separate hooks inline. The screen was already at the `max-statements` ceiling, and this keeps it under it.

**Persisting the seen state**

`txLastSeen` is a client-only Apollo field, and the cache persistor only restores from disk when there is an auth token (`app/graphql/client.tsx`). A self-custodial account resolves that token to `""` by design, so its seen state was never restored: on every cold start `lastSeenBtcId` was `""`, the newest transaction always compared as unseen, and the badge fired again for a payment that could be weeks old.

- `app/store/persistent-state/tx-last-seen.ts` (new) plus schema 16 (`txLastSeenByAccountId`): the same per-account device store already used for display currency, language and quiz progress.
- `app/hooks/use-transaction-seen-state.ts`: custodial keeps using the Apollo cache; self-custodial reads and writes the device store. It also stops falling back to the cached custodial home query, which for a self-custodial account is another account's data.
- The hook now takes an explicit `isSelfCustodial` from its caller rather than deriving it, so the flag that picks which transactions are fed in is the same one that picks where they are stored. Deriving it internally meant a second, looser predicate (`accountType` only) that could only agree with the caller's `useActiveWallet().isSelfCustodial` (`accountType` + wallet status) by accident.
- `transaction-history-screen.tsx` and `transaction-detail-screen.tsx` move to the same signature. Both derived their `accountId` from `me.defaultAccount.id`, which is `""` for a self-custodial account, so without this they would have keyed the seen state differently from home and marking a transaction seen in the history would not have cleared the badge.

**Announcing a transaction once**

Marking seen used to be deferred to the end of the display window, 5000ms plus the 180ms exit animation, and the effect cleanup cleared both timers. Any exit before that point cancelled the write: losing focus, unmounting, or closing the app. On a self-custodial account the window effectively started later still, because the transaction list only arrives once the SDK has synced, so a short visit was enough to never persist anything.

- `use-incoming-badge-auto-seen.ts`: marks the transaction seen the moment the badge reaches the screen, and holds what it announced until the exit animation has played. Marking seen empties the unseen state the badge is derived from, so the announcement is what the screen renders from while it is up. The guard is keyed by transaction rather than by currency, so the one already shown never comes back while a second receive in the same wallet is still announced, and a live announcement is never overwritten: with both wallets unseen, each waits its turn instead of the second replacing the first a frame later.
- `use-unseen-tx-badge-state.ts`: resolves the badge amount, id and direction from the announcement while one is live, keeps the announced currency reading unseen so the wallet row dot stays lit for the whole window, and parks the outgoing badge behind an incoming one, which would otherwise run its cycle unpainted and still mark its own transaction seen when it hid.
- `use-unseen-tx-amount-badge.ts`: exposes `navigateToTransaction(txid)` instead of a press handler bound to the newest unseen transaction, which by the time the badge can be pressed no longer exists.

## Test plan

- `yarn jest` on the badge hooks plus the home screen: 117 passed, **100% statements, branches, functions and lines** on the two hooks this rewrites
- Self-custodial coverage exercises the real logic, not mocks: cold start with an empty store, a stored id treated as already seen, marking seen on device instead of the cache, and never reading the custodial cache
- Each fix is pinned by a test that fails when the fix alone is reverted, so none of them pass by coincidence
- `__tests__/screens/home.spec.tsx`: integration cases that differ only by `allTransactions`, asserting the badge appears for a received self-custodial transaction and stays absent when the wallet has none
- Scoped run over the transitive consumers (hooks, home, transaction history, transaction detail, persistent state, contacts): 14 suites, 227 passed
- Device (Android), against the exact scenario reported here: received 100 sats on a self-custodial account, the badge appeared with `+$0.06` and the white dot on the Bitcoin pill, the app was killed within 2 seconds of it appearing, well inside the window where the previous code had not yet written anything, and reopening did not announce it again

## Screenshots

| Before | After |
| :---: | :---: |
| 925-1-before | 925-2-after |

Before: the account has just received a payment and the home shows no bubble and no amount. After: the same moment shows `+$0.06` under the balance and the white dot on the Bitcoin pill. The grey dot on the Dollar pill is an older unseen transaction that the badge cycles to once the newest one is marked as seen. Balances are masked by the app's hide-balance toggle.

## Known follow-ups

Not addressed here, worth their own tickets:

- The announcement holds the amount text it was given, and that text is formatted from raw minor units while the price conversion has not resolved yet, so a receive can read as `$500.00` instead of `$5.00` for the length of the display window. The formatting is pre-existing; holding it is not, and the fix is to resolve the announced transaction from the full list and format it live rather than snapshot the string.
- The announcement is gated on navigation focus rather than app state, so a payment landing while the app is backgrounded is announced and marked seen without anyone having seen it. Pre-existing, and gating on `AppState` is a behaviour change of its own.
- A self-custodial send marked seen while pending and then failed is filtered out of the list, so the latest id falls back to an older transaction and the badge resurfaces for it. Pre-existing, but the id now survives restarts.
- The migration does not seed `txLastSeenByAccountId`, so an existing self-custodial user sees the badge once on the first launch after upgrading. Still strictly better than the current every-launch behavior.
- Home maps the full self-custodial history to fragments and primes the cache for it on every price tick, to render a single badge that only needs the newest transaction per currency.

### DIFF
--- a/__tests__/components/use-unseen-tx-badge-state.spec.ts
+++ b/__tests__/components/use-unseen-tx-badge-state.spec.ts
@@ -1,0 +1,266 @@
+import { renderHook } from "@testing-library/react-native"
+
+import {
+  TransactionFragment,
+  TxDirection,
+  TxStatus,
+  WalletCurrency,
+} from "@app/graphql/generated"
+import { useUnseenTxBadgeState } from "@app/components/unseen-tx-amount-badge/use-unseen-tx-badge-state"
+import {
+  NormalizedTransaction,
+  PaymentType,
+  TransactionDirection,
+  TransactionStatus,
+} from "@app/types/transaction"
+
+const mockSeenState = jest.fn()
+const mockAmountBadge = jest.fn()
+const mockFragments = jest.fn()
+const mockOutgoingVisibility = jest.fn()
+const mockIncomingVisibility = jest.fn()
+const mockMarkTxSeen = jest.fn()
+
+jest.mock("@app/hooks/use-transaction-seen-state", () => ({
+  useTransactionSeenState: (accountId: string, transactions: unknown) =>
+    mockSeenState(accountId, transactions),
+}))
+
+jest.mock("@app/self-custodial/hooks/use-self-custodial-transaction-fragments", () => ({
+  useSelfCustodialTransactionFragments: (transactions: unknown) =>
+    mockFragments(transactions),
+}))
+
+jest.mock("@app/components/unseen-tx-amount-badge/use-unseen-tx-amount-badge", () => ({
+  useUnseenTxAmountBadge: (params: unknown) => mockAmountBadge(params),
+}))
+
+jest.mock("@app/components/unseen-tx-amount-badge/use-outgoing-badge-visibility", () => ({
+  useOutgoingBadgeVisibility: (params: unknown) => mockOutgoingVisibility(params),
+}))
+
+jest.mock("@app/components/unseen-tx-amount-badge/use-incoming-badge-auto-seen", () => ({
+  useIncomingBadgeAutoSeen: (params: unknown) => mockIncomingVisibility(params),
+}))
+
+const CUSTODIAL_ACCOUNT_ID = "custodial-account"
+const SELF_CUSTODIAL_ACCOUNT_ID = "self-custodial-account"
+
+const makeSelfCustodialTransaction = (): NormalizedTransaction => ({
+  id: "sc-tx",
+  amount: { amount: 1000, currency: WalletCurrency.Btc, currencyCode: "BTC" },
+  direction: TransactionDirection.Receive,
+  status: TransactionStatus.Completed,
+  timestamp: 1747691078,
+  paymentType: PaymentType.Lightning,
+})
+
+const makeCustodialFragment = (
+  id: string,
+  overrides: Partial<TransactionFragment> = {},
+): TransactionFragment => ({
+  __typename: "Transaction",
+  id,
+  status: TxStatus.Success,
+  direction: TxDirection.Receive,
+  memo: null,
+  createdAt: 1747691078,
+  settlementAmount: 100,
+  settlementFee: 0,
+  settlementDisplayFee: "0",
+  settlementCurrency: WalletCurrency.Btc,
+  settlementDisplayAmount: "0.06",
+  settlementDisplayCurrency: "USD",
+  settlementPrice: {
+    __typename: "PriceOfOneSettlementMinorUnitInDisplayMinorUnit",
+    base: 0,
+    offset: 0,
+    currencyUnit: "USD",
+    formattedAmount: "0",
+  },
+  initiationVia: {
+    __typename: "InitiationViaLn",
+    paymentHash: `hash-${id}`,
+    paymentRequest: "",
+  },
+  settlementVia: { __typename: "SettlementViaLn", preImage: null },
+  ...overrides,
+})
+
+const renderBadgeState = (
+  overrides: Partial<Parameters<typeof useUnseenTxBadgeState>[0]> = {},
+) =>
+  renderHook(() =>
+    useUnseenTxBadgeState({
+      isSelfCustodial: false,
+      isFocused: true,
+      custodialAccountId: CUSTODIAL_ACCOUNT_ID,
+      selfCustodialAccountId: SELF_CUSTODIAL_ACCOUNT_ID,
+      selfCustodialTransactions: [],
+      pendingIncomingTransactions: null,
+      transactionEdges: null,
+      ...overrides,
+    }),
+  )
+
+describe("useUnseenTxBadgeState", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockFragments.mockReturnValue([])
+    mockSeenState.mockReturnValue({
+      hasUnseenBtcTx: false,
+      hasUnseenUsdTx: false,
+      markTxSeen: mockMarkTxSeen,
+    })
+    mockAmountBadge.mockReturnValue({
+      latestUnseenTx: undefined,
+      unseenAmountText: null,
+      handleUnseenBadgePress: jest.fn(),
+      isOutgoing: false,
+    })
+    mockOutgoingVisibility.mockReturnValue(false)
+    mockIncomingVisibility.mockReturnValue(true)
+  })
+
+  describe("custodial account", () => {
+    it("feeds the badge with the transactions from the home query", () => {
+      renderBadgeState({
+        pendingIncomingTransactions: [
+          makeCustodialFragment("pending", { status: TxStatus.Pending }),
+        ],
+        transactionEdges: [{ node: makeCustodialFragment("settled") }],
+      })
+
+      const [, transactions] = mockSeenState.mock.calls[0]
+      expect(transactions.map((tx: { id: string }) => tx.id)).toEqual([
+        "pending",
+        "settled",
+      ])
+    })
+
+    it("leaves out an incoming transaction that is still pending", () => {
+      renderBadgeState({
+        transactionEdges: [
+          { node: makeCustodialFragment("settled") },
+          {
+            node: makeCustodialFragment("pending-incoming", { status: TxStatus.Pending }),
+          },
+        ],
+      })
+
+      const [, transactions] = mockSeenState.mock.calls[0]
+      expect(transactions.map((tx: { id: string }) => tx.id)).toEqual(["settled"])
+    })
+
+    it("keeps an outgoing transaction that is still pending", () => {
+      const pendingSend = makeCustodialFragment("pending-send", {
+        status: TxStatus.Pending,
+        direction: TxDirection.Send,
+      })
+
+      renderBadgeState({ transactionEdges: [{ node: pendingSend }] })
+
+      const [, transactions] = mockSeenState.mock.calls[0]
+      expect(transactions.map((tx: { id: string }) => tx.id)).toEqual(["pending-send"])
+    })
+
+    it("keys the seen state by the custodial account", () => {
+      renderBadgeState()
+
+      expect(mockSeenState).toHaveBeenCalledWith(CUSTODIAL_ACCOUNT_ID, [])
+    })
+
+    it("never maps self-custodial transactions", () => {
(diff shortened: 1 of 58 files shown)


grimen (changes requested):
## Findings

**1. Opening the newest self-custodial transaction's detail does not mark it seen.**
The detail screen calls `useTransactionSeenState` without `transactions`, and the self-custodial path never falls back to the cache, so `latestBtcTxId`/`latestUsdTxId` are always `""` there and the mark-seen effect (`latestId === txid`) can never fire. The badge press itself only navigates (`handleUnseenBadgePress`), and `useIncomingBadgeAutoSeen` cancels its pending mark when home loses focus.

*Example:* A self-custodial user taps the badge within the 5-second window, reads the transaction detail, and returns home — the badge and amount reappear and run the full auto-seen cycle again, whereas a custodial user returning from the same tap sees no badge because the detail screen marked it.

*Proposal:* One option is to feed the detail screen's `useTransactionSeenState` with the mapped wallet fragments (the screen already has `useActiveWallet`, and `useSelfCustodialTransactionFragments` exists for exactly this shape); if the lingering badge is considered acceptable, a comment on the detail-screen call noting the intentional self-custodial no-op would help.

**2. `useTxLastSeenQuery` still runs for self-custodial accounts with its result discarded.**
`use-transaction-seen-state.ts` executes the query unconditionally and then ignores `cachedLastSeenBtcId`/`cachedLastSeenUsdId` whenever `isSelfCustodial` is true, keeping a cache subscription alive with `accountId: ""` for no consumer.

*Example:* Every self-custodial render of home, history, and detail sets up a `txLastSeen` cache watch whose data can never be used.

*Proposal:* Passing `skip: isSelfCustodial` to the query would make the storage split explicit at the query site as well.

## Test Coverage Gaps

- The cross-screen key agreement the PR description leans on (marking seen in history/detail clears the home badge for a self-custodial account) is only asserted through mock call shapes (`leaves the self-custodial seen-state key to the seen-state hook`); there is no test exercising the history or detail screen's seen flow under `isSelfCustodial: true`, which is also where finding 1 hides.

## Code Smells

- The `hasProvidedTransactions`/`canReadCachedTransactions` block, including its identical doc comment, now appears verbatim in both `use-transaction-seen-state.ts` and `use-unseen-tx-amount-badge.ts`; a small shared helper would keep the "self-custodial never falls back to the custodial cache" rule in one place.

esaugomez31:
@grimen

## Findings

**1. Opening the newest self-custodial transaction's detail does not mark it seen.**
Fixed. The detail screen now feeds `useTransactionSeenState` with the mapped fragments, as proposed. One deviation: it sources them from `useSelfCustodialWallet().allTransactions` instead of flattening `useActiveWallet().wallets` — after `loadMore` the per-wallet lists are appended unsorted while `allTransactions` is re-sorted, so on a timestamp tie the two could resolve to a different "latest" id than home's. Same source as home means the cross-screen key agreement is structural now, not incidental.

**2. `useTxLastSeenQuery` still runs for self-custodial accounts with its result discarded.**
Fixed: `skip: isSelfCustodial`, asserted for both account kinds.

## Test Coverage Gaps

- New `transaction-detail-seen-state.spec.tsx` drives the detail screen with the real seen-state hook and the real device store under `isSelfCustodial: true`, including the cross-screen assertion you were after: after the detail marks, a home-shaped read of the seen state reports no unseen BTC. Confirmed the three relevant cases fail without the fix.
- `transaction-history-self-custodial.spec.tsx` gained a `seen state` block on the same real hook: newest per currency, wallet-filter scoping, the account key, and the custodial contrast (Apollo, not device).

## Code Smells

- The `hasProvidedTransactions`/`canReadCachedTransactions` block and `readCachedTransactions` now live once, in `app/hooks/use-account-transactions.ts`, used by both hooks.
- Two more copies found while there: the pending-vs-settled predicate was duplicated between that hook and `use-unseen-tx-badge-state`, and the empty-list constant with its identical doc comment sat in three files. Consolidated into `toCustodialTransactions` and `NO_TRANSACTIONS`.

Four atomic commits; `tsc` and eslint clean, related suites green.

grimen (dismissed):
All findings from the previous review are addressed.

grimen (dismissed):

designsats:
Seem to work well, but I keep seens same notification when re-openning the app. Each transaction should be only shown once.

esaugomez31:
> @designsats: Seem to work well, but I keep seens same notification when re-openning the app. Each transaction should be only shown once.

Fixed.

designsats:
works on my end, tested on android

grimen (changes requested):
## Findings

**1. Interrupted outgoing badge never marks its transaction seen (blocking).**

`useOutgoingBadgeVisibility`'s effect keys off `txId`/`isOutgoing`/`amountText`, and its cleanup only clears the pending `hideTimeout` — it never calls `onHide`. If a newer outgoing transaction replaces `outgoingTxId` before the 5s hide timer for the current one fires, the cleanup runs but `onHide` (and therefore `markTxSeen`) never fires for the interrupted transaction.

*Example:* User sends BTC, then sends USD a couple seconds later before the BTC badge's hide timer completes; `hasUnseenBtcTx` stays stuck `true` even though the badge was shown, until the user opens that transaction's detail screen directly.

*Proposal:* The cleanup could call `onHide` for the outgoing transaction being replaced, not just clear the timer, so a superseded transaction still gets marked seen.

**2. The "incoming suppresses outgoing" rule is encoded twice (blocking).**

`isOutgoingBadge` and `outgoingTxId`/`outgoingAmountText` are both derived independently from the same `isAnnouncingIncoming` condition, one via boolean and one via undefined/null.

*Example:* A future edit to the suppression condition that updates only one of the two encodings reintroduces the exact race this PR fixes, since `useOutgoingBadgeVisibility` keys off `txId`/`amountText`/`isOutgoing` independently.

*Proposal:* Deriving both from a single source (e.g. one memoized value gating both) would keep them from drifting apart.

## Test Coverage Gaps

- No test covers a newer outgoing transaction (different currency) superseding an in-flight outgoing badge before its hide timer fires — this is the scenario finding 1 describes.

## Code Smells

- The unseen-eligibility predicate (`settlementAmount !== 0`, excluding the fee-reimbursement memo) is duplicated between `getLatestTransactionId` in `use-transaction-seen-state.ts` and the filter in `use-unseen-tx-amount-badge.ts`, instead of one shared predicate.

esaugomez31:
@grimen all findings addressed, in abc8ba2f4, 7400f1a4e and a0391f70d.

## Findings

**1. Interrupted outgoing badge never marks its transaction seen.** Fixed in abc8ba2f4.

Confirmed, and reproduced as a failing test before touching the fix.

I did not take the proposal literally, because a cleanup that always calls `onHide` breaks two other things. It fires when the badge is parked behind an incoming announcement, which would mark a transaction the user never saw and reintroduce exactly what this PR fixes. And it fires again after the hide timer already ran, marking whatever transaction is latest by then.

So the cleanup is guarded by an `isAwaitingHide` flag that is true only between the paint and the hide. A badge cut short while it was on screen owes the same mark-seen its own timer would have done; one that never painted owes nothing.

*Verified:* the existing `restarts timers when txId changes` asserted `onHide` was called once. It was pinning the bug, and now asserts twice.

**2. The "incoming suppresses outgoing" rule is encoded twice.** Fixed in 7400f1a4e.

All three inputs now come from one `outgoingBadge` value, so the suppression cannot be changed for some of them and not the others.

## Test Coverage Gaps

Covered at two levels, since the screen spec mocks the visibility hook and the hook spec mocks the wiring:

- In `use-outgoing-badge-visibility.spec.ts`, the mechanism: superseded after painting marks seen, superseded before painting does not, parked before painting does not, parked after painting does, and a hide already paid for is not paid twice.
- In `use-unseen-tx-badge-state.spec.ts`, the wiring: a superseded badge marks **its own** currency, never the one that replaced it. The cleanup runs with the render that armed it, and that was the part of the fix most worth pinning.

## Code Smells

Fixed in a0391f70d. The predicate is now `isAnnounceableTransaction` in `use-account-transactions.ts`, which both consumers already import and whose docstring already says these shared reading rules live there rather than at each call site.

## Notes

Three commits, one per finding, each state verified green and bisectable on its own at 31, 18 and 34 tests. 100% coverage (statements, branches, functions, lines) on all four files touched. Typecheck and lint clean.

Worth flagging explicitly, since it goes past the letter of finding 1: three cases now mark a transaction seen where they previously did not, all the same rule of "it was painted, so it counts as seen". Superseded by a newer transaction, parked behind an incoming badge after painting, and unmounted while showing. A badge that never painted still marks nothing, which is what keeps this PR's original fix intact.

Also exercised on the Android emulator against staging: a real self-custodial Lightning send between two accounts, 621 sats plus a 3 sat fee, with the balance and the seen state both landing correctly afterwards.

grimen (dismissed):

grimen (approved):

PR #4074, opened 2026-08-04
fix(send): trim whitespace from destination input before parsing

### LINKED ISSUE #4073: bug: Trim destination input text
**Describe the bug**
When I user inputs the destination of a transaction with a leading space it is seen as an invalid destination.

An experienced user may be able to figure out 

**To Reproduce**

1. Go to Send
2. Input a space followed by a valid lightning address (maybe even works for invoices)
3. Try go to next screen

**Expected behavior**
The input should be trimmed of leading and trailing whitespaces before being processed

**Screenshots**

Image

Android (dev build)

grimen:
Closes #4073

## What

A destination entered with a leading or trailing space — ` user@blink.sv ` instead of `user@blink.sv` — was rejected as an invalid destination, with no hint about why. This happens easily when copying an address from a chat app that appends whitespace. Reported for both custodial and self-custodial sends.

## Why it broke

Every destination (typed, pasted, QR-scanned, deep-linked, NFC) funnels into `parsePaymentDestination()` from `@blinkbitcoin/blink-client`, which does not trim its input: a stray space makes payment-type detection fall through to `Unknown`. The self-custodial Spark path happened to trim internally, and trimming at the parse boundary is the existing convention elsewhere (`phone.ts`, `card-screen/utils/address.ts`) — this chain was the one boundary that missed it.

## How

- `parseDestination` trims `rawInput` before handing it to `parsePaymentDestination` (and to the BIP-21 unified-URI lookup). This single point also covers NFC, which calls `parseDestination` directly.
- `resolveDestination` normalizes `rawInput` at entry, so the Spark pre-check and the username re-parse callback see trimmed input by contract rather than by accident.
- `handlePaste` trims clipboard content before storing it, so the input box shows the cleaned value.

Typed input is deliberately **not** trimmed in state: the reducer drops validation results whose `unparsedDestination` echo doesn't match state (so the trim must live inside the resolvers), and the search field doubles as a multi-word contact search where trimming as you type would break entering a second word.

## Tests

- `resolve-destination.spec.ts`: trim assertions for both the custodial and self-custodial paths.
- `resolve-destination.integration.spec.ts` (real parser): `"  esaudeveloper@blink.sv  "` resolves as valid, plus a direct `parseDestination` case pinning the NFC path. Both fail without the fix.
- `send-destination.spec.tsx`: pasting `"  clipboard  "` validates the trimmed value and displays the cleaned string; typing a padded destination and pressing Next validates the trimmed value.

### DIFF
--- a/__tests__/payment-destination/resolve-destination.integration.spec.ts
+++ b/__tests__/payment-destination/resolve-destination.integration.spec.ts
@@ -3,6 +3,7 @@ import { requestPayServiceParams } from "lnurl-pay"
 import { Network as mockSparkNetwork } from "@breeztech/breez-sdk-spark-react-native"
 
 import { Network } from "@app/graphql/generated"
+import { parseDestination } from "@app/screens/send-bitcoin-screen/payment-destination"
 import { resolveDestination } from "@app/screens/send-bitcoin-screen/payment-destination/resolve-destination"
 import { PaymentType } from "@blinkbitcoin/blink-client"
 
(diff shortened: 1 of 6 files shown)


grimen:
Demo on the iOS simulator — the clipboard holds `"  dolcalmi@blink.sv  "` (two leading and two trailing spaces) and is pasted via the in-app Paste button in both runs.

|---|---|
|  |  |

Before: the padded value is stored verbatim and rejected — red border, *Enter a valid destination*, Next disabled. After: the input is trimmed on paste, shows the clean address, validates green and Next is enabled.

<sub>Images live on branch `assets/pr-4074-screenshots` — deleting it breaks them.</sub>

kngako:
Code runs does just as it says on the tin. 👌🏿

image

kngako (approved):
Code and tests are clean.

PR #4077, opened 2026-08-04
feat(spark): transaction history CSV download for non-custodial accounts

### LINKED ISSUE #4042: feat(spark): Transaction history CSV download
## Description

Allow users on non-custodial (Spark) accounts to download their full transaction history as a CSV file, in the same format as currently available in custodial mode.

## Acceptance criteria

- Non-custodial users can export transaction history as CSV
- CSV format matches the existing custodial export format
- All transaction types (send, receive, on-chain, Lightning) are included

---
Filed by Blink-Claw-Bot ⚡ on behalf of @designsats

grimen:
Fixes #4042

## Overview

Non-custodial (Spark) accounts can now export their full transaction history as a CSV from Settings → Advanced → **Export all transactions** — the same entry point custodial accounts use. The Spark history lives in the SDK on the device, so the CSV is generated client-side from the full `listPayments` history and handed to the exact same share-sheet flow as the custodial export. No backend change, no new native dependency.

## Format

The CSV replicates the backend ledger export (`csv-wallet-export.ts`) byte-for-byte in structure: identical 24-column header, LF records with a trailing newline, csv-writer's minimal quoting, unsigned credit/debit in minor units with the fee folded into the debit, and JS `Date#toString` timestamps. Existing spreadsheets and tools built for the custodial CSV keep working.

Deliberate choices where a self-custodial client cannot match the ledger:

- **Fiat columns**: `usd`/`feeUsd` are filled *exactly* for USDB rows (the token amount is the historical dollar value); all fiat columns stay blank for BTC rows. The SDK only knows today's price — a plausible-but-wrong historical rate is worse for accounting than an honest blank.
- **`type` vocabulary**: Lightning and on-chain rows reuse the custodial values (`invoice`, `payment`, `onchain_receipt`, `onchain_payment`). Spark-native operations default to descriptive values (`spark_transfer`, `conversion`); a `CsvTypeVocabulary.CustodialCompat` option maps them to `on_us`/`self_trade` instead, switchable via one constant (`DEFAULT_CSV_TYPE_VOCABULARY`).
- **Row scope**: completed + pending (`pendingConfirmation=true`); failed payments are excluded, matching both the history screen and what a custodial CSV can contain.
- Ledger-only columns (`journalId`, `recipientWalletId`, `username`, `memoFromPayer`, `feeKnownInAdvance`, `address`) stay empty.

## Implementation — custodial and self-custodial fully separated

The two implementations never share export logic: each account mode has its own adapter and hook under its own namespace, following the `usePayments` adapter convention. Neither side imports from the other; the only shared code is the mode-agnostic share-sheet helper and the settings row that selects a side.

**Custodial** (`app/custodial/`) — the backend renders the CSV, the client relays it:
- `adapters/export-csv.ts` — `createCustodialExportCsv`: empty export → false, missing field → throw, otherwise hand the backend's base64 to the share sheet.
- `hooks/use-export-transactions-csv.ts` — wires the `csvTransactions` Apollo lazy query into the adapter. Moved here from the generic `app/hooks/` (it was custodial-specific all along); public contract unchanged, so the migration "Download history" screen only changed an import path.

**Self-custodial** (`app/self-custodial/`) — the phone is the source of truth, the CSV is built on-device:
- `adapters/export-csv.ts` — `createBuildTransactionsCsvBase64`: full history via `listAllPayments`, drops failed and unknown-token payments, returns the CSV base64-encoded (`null` when nothing is exportable). Deliberately share-free — see the review follow-ups below.
- `hooks/use-export-transactions-csv.ts` — wires the connected Spark SDK into the builder and hands the result to the share sheet; the settings-row spinner covers CSV generation only.
- `mappers/transaction-csv.ts` — pure `Payment[] → CSV` builder; sats arithmetic stays in bigint end-to-end.
- `bridge/wallet.ts` — `listAllPayments` pages until a short page, dedupes ids across page boundaries, and aborts with a descriptive error on a runaway loop.

**Shared, mode-agnostic:**
- `app/utils/share-csv.ts` — share-sheet mechanics extracted from the custodial hook (data URL, per-platform filename, dismissal → false). Knows nothing about either account mode.
- Settings: `ExportCsvSetting` is the single selection point — it branches by account type into a custodial and a self-custodial row component, each consuming only its namespace's hook (the `AccountLNAddress` pattern). `ApiAccessSetting` returns null for self-custodial (the API screen documents the custodial GraphQL API); the Advanced group renders in both modes and collapses when every row is null.

## Testing

The 11 suites this PR touches pass with 90 tests: exact header/row assertions per payment variant (LN send/receive, deposit, withdraw, spark transfer, USDB token, conversion legs, missing details), CSV escaping, bigint exactness, both type vocabularies, pagination/dedupe/runaway-cap, hook filtering (failed + unknown-token), settings gating in both modes, and share-sheet regression coverage from review (below). Every file the PR owns sits at 100% statement/branch/line coverage. The previously untested "Advanced group hidden in NC mode" behavior is now locked as "visible with export row, without API access".

Before/after screenshots from a simulator run are attached in the first comment.

## Review follow-ups

- **Settings spinner no longer waits on the share-sheet result** ([`d0979b3d9`](https://github.com/blinkbitcoin/blink-mobile/commit/d0979b3d9370826e0f521978630ba0362ead3d50)) — @kngako hit a share target (Google Drive with an unsynced account) holding the intent open, which pinned the row in its spinner state. The spinner is now scoped to CSV generation, matching the custodial row where it covers the fetch query. The `exportCsv` promise still resolves with the share outcome because the [migration download-history screen consumes it](https://github.com/blinkbitcoin/blink-mobile/blob/d0979b3d9370826e0f521978630ba0362ead3d50/app/screens/account-migration/to-non-custodial/download-history-screen.tsx#L44-L47) to flip Skip → Continue. Proven by a [regression test that simulates a never-resolving share intent](https://github.com/blinkbitcoin/blink-mobile/blob/d0979b3d9370826e0f521978630ba0362ead3d50/__tests__/self-custodial/hooks/use-export-transactions-csv.spec.ts#L169-L189) and asserts the spinner clears anyway.
- **Podfile.lock checksum churn reverted** ([`d2e2c5ca4`](https://github.com/blinkbitcoin/blink-mobile/commit/d2e2c5ca413632ea8038c403d1d7d28aea24f6af)) — `383b05e2` carried three spec-checksum lines from a local `pod install` with no dependency change; the branch no longer touches `ios/Podfile.lock`.


grimen:
Simulator run on a fresh self-custodial account (iPhone simulator, local dev build):

|---|---|
|  |  |

On `main`, Settings in non-custodial mode jumps from *Security and Privacy* straight to *Support* — no export at all. With this PR the *Advanced* group appears with exactly one row, **Export all transactions** (no API integration row, which stays custodial-only).

One honesty note: local dev builds have no `BREEZ_API_KEY`, so the Spark SDK never connects and the row would sit in its loading-skeleton state; for the screenshot the row's SDK-wait state was temporarily stubbed as connected (screenshot-only edit, not in the branch). The share-sheet flow itself is exercised by the hook specs instead.

<sub>Images live on branch `assets/pr-4077-screenshots` — deleting it breaks them.</sub>

grimen:
Example CSV files, generated by the shipped `buildTransactionsCsv` over a representative payment set (LN send/receive, on-chain deposit + pending withdrawal, spark transfer, USDB token send, a BTC↔USDB conversion pair, and a memo that exercises quoting):

- [blink-transactions-sample.csv](https://github.com/blinkbitcoin/blink-mobile/blob/assets/pr-4077-screenshots/blink-transactions-sample.csv) — default descriptive vocabulary
- [blink-transactions-sample-custodial-compat.csv](https://github.com/blinkbitcoin/blink-mobile/blob/assets/pr-4077-screenshots/blink-transactions-sample-custodial-compat.csv) — same payments with `CsvTypeVocabulary.CustodialCompat` (`on_us`/`self_trade`)

Worth eyeballing in the sample: the LN send's `debit` is 50012 with `fee` 12 (debit includes the fee, like the ledger export); the USDB row is in cents (2600) with exact `usd`/`feeUsd` (25.99/0.01) while BTC rows leave fiat blank; the pending withdrawal carries `pendingConfirmation=true`; and the two conversion legs share a timestamp and appear once per wallet, mirroring custodial `self_trade`.

Generated by running the production builder directly — a local dev build has no Breez API key, so an in-app export can't produce populated rows on the simulator. The timestamps show the generating machine's timezone, exactly as the backend export shows the server's.

kngako:
I ran into a small hiccup (trying to upload to Google Drive from the simulator with what appears to be an unsynced Google account). 

The account hadn't finished syncing, but seems the Blink app is waiting for a response from the ShareIntent. This does go away if the users leaves the settting screen, but unless the export is going to use the result of the share action I don't think it needs to wait for a response... or even show a loading action like this (regardless of whether or not the share action was succeessful or hanging).

image

Doesn't seem to be something that always happens though. So we might be okay let it be.

kngako:
Not sure if the 383b05e27bba804597f5d4abc35d4e405caac257 was supposed to touch any dependency related code... I saw the Podfile is updated but I don't see anything in the commit message specifically about it.

grimen:
Both items addressed — details and proof per item:

> seems the Blink app is waiting for a response from the ShareIntent. [...] unless the export is going to use the result of the share action I don't think it needs to wait for a response... or even show a loading action like this

Fixed in [`d0979b3d9`](https://github.com/blinkbitcoin/blink-mobile/commit/d0979b3d9370826e0f521978630ba0362ead3d50). The row spinner was tied to the full export promise, which awaited the share intent — so a target app holding the intent open (as Drive did for you) kept the row loading indefinitely. The adapter is now a share-free CSV builder and [the hook clears the spinner as soon as the CSV is built](https://github.com/blinkbitcoin/blink-mobile/blob/d0979b3d9370826e0f521978630ba0362ead3d50/app/self-custodial/hooks/use-export-transactions-csv.ts#L22-L34), before handing off to the share sheet — the same semantics as the custodial row, whose spinner covers only the fetch query.

One nuance on "doesn't need to wait for a response": the settings row indeed ignores the result, but the returned promise still resolves with the share outcome because the [migration download-history screen consumes it](https://github.com/blinkbitcoin/blink-mobile/blob/d0979b3d9370826e0f521978630ba0362ead3d50/app/screens/account-migration/to-non-custodial/download-history-screen.tsx#L44-L47) to flip Skip → Continue. So the wait is invisible (no spinner, screen fully usable), not removed — and share errors still surface through the row's toast path rather than being fire-and-forgotten.

**Proof**: a [regression test simulating your exact scenario](https://github.com/blinkbitcoin/blink-mobile/blob/d0979b3d9370826e0f521978630ba0362ead3d50/__tests__/self-custodial/hooks/use-export-transactions-csv.spec.ts#L169-L189) — the share intent is a promise that never resolves, and the test asserts the spinner still clears once the CSV is handed off. It fails against the previous hook, so the hang can't regress silently. Error paths (generation failure, share failure) are covered alongside, and both touched files stay at 100% statement/branch/line coverage.

> Not sure if the 383b05e2 was supposed to touch any dependency related code... I saw the Podfile is updated but I don't see anything in the commit message specifically about it.

It wasn't — those were three spec-checksum lines (hermes-engine, ReactNativeDependencies, RNPermissions) from a local `pod install` with no dependency change behind them. Reverted in [`d2e2c5ca4`](https://github.com/blinkbitcoin/blink-mobile/commit/d2e2c5ca413632ea8038c403d1d7d28aea24f6af); the branch no longer touches `ios/Podfile.lock` at all (`git diff main...HEAD -- ios/Podfile.lock` is empty).

grimen:
Formatted in [`ca303946f`](https://github.com/blinkbitcoin/blink-mobile/commit/ca303946f) — `prettier --write` on the spec, `eslint:check` clean locally. Only the multiline call formatting changed, no test behavior.

grimen:
`Check Code` had a second, previously masked failure behind the Prettier one: with eslint fixed, the chain reached `check:codegen`, which found `app/graphql/generated.ts` out of sync with what `yarn dev:codegen` produces. Cause: moving the custodial export hook from `app/hooks/` to `app/custodial/hooks/` changed codegen's document scan order, so the `ExportCsvSetting` query blocks belong at a different position in the generated file.

Fixed in [`d07db34eb`](https://github.com/blinkbitcoin/blink-mobile/commit/d07db34eb462753491fa9987f7690c485fcf157a) by re-running `yarn dev:codegen` and committing the result. The diff is purely the relocated `ExportCsvSetting` blocks (51 lines removed, the same 51 reinserted elsewhere) — no semantic change. The full chain (`tsc:check`, `eslint:check`, `check:translations`, `check:codegen`, `graphql-check`) passes locally, and [`Check Code` is now green on this commit](https://github.com/blinkbitcoin/blink-mobile/commits/feat/spark-csv-export).

kngako (approved):
New behaviour works as expected.

PR #4078, opened 2026-08-04
feat(self-custodial): hand over the recovery backup during backup onboarding

grimen:
**Stack 4/4** for the unilateral-exit recovery bundle. Base: `feat--save-refresh-bundle` (#3911). Merge bottom-up; GitHub retargets as bases merge.

Implements the backup-onboarding half of blink-wip#807, per the UI/UX flows in [this comment](https://github.com/blinkbitcoin/blink-wip/issues/807#issuecomment-4959268940) and Phase 2 of [blink-specs#26](https://github.com/blinkbitcoin/blink-specs/pull/26).

## Problem

#3911 made the recovery bundle real — encrypted on device, refreshed after every payment, exportable from Settings — but nothing in the backup flow mentions it. A user writes down twelve words, believes they are safe, and owns a second artifact they have never seen.

## What this adds

**Manual path** — a step between three-word confirmation and the success screen that hands the bundle over. Exactly two actions, download and copy, behind a "Sensitive" confirmation, and the user cannot continue until one has been used.

**Cloud path** — the "Automatic backups" toggle from the design, beside the existing password checkbox.

Export reuses `useRecoveryBundleActions` rather than re-implementing it, so the 60s clipboard auto-clear (R10) still applies.

## Decisions worth reviewing

Two places where the prototype and the PRD disagree. I followed the PRD; both are cheap to reverse if the design review concluded otherwise.

| Prototype | PRD | Here |
| --- | --- | --- |
| "Automatic backups" on by default | D4 — opt-in, off by default | **off** |
| manual path also offers cloud | D3/R7 — cloud never appears in the manual flow | **no cloud button** |

The prototype's "Blink hosted" option is dropped entirely: no bundle is stored with Blink, only the user's own cloud or on device.

The two toggles are also **coupled**, which the prototype draws as independent: a seed-encrypted bundle stored next to an unencrypted seed is decrypted by that seed on the spot (D9), so sync is only offered once the seed backup carries a password, and the opt-in reads as off whenever a password-protected seed backup is not on offer, so the row never renders a checked box inside a disabled control.

## A wallet has no recovery backup until it has funds

`exporter.ts` throws `NoLeaves` rather than producing an empty bundle, and nothing creates one at wallet creation. So at signup — and during migration, where backup runs before the transfer commits — there is nothing to download.

The screen therefore has two states: a funded wallet gets the export, an empty one is told the backup is created automatically once funds arrive. Handing over a file that would recover nothing seemed worse than saying "not yet".

**Consequence:** R6's "every backup path ends with the bundle saved" is not satisfied at onboarding for a new user. It can only be met once a bundle exists, which makes the stale/missing nudge (Phase 2, not in this PR) load-bearing rather than a nicety.

## Tests

`__tests__/screens/self-custodial/` — 402 passing. New coverage for the export screen's two states and the export gate, and for the four states of the sync gate (off by default, unavailable without a password, can be enabled, cleared when the password goes).

Changed files are 97–100% covered; what remains is three one-line branches.

## Notes for review

- Two defects were found by screenshotting, not by tests: the warning modal rendered with no body (React Native drops a bare string inside a `View` — `CustomModal` needs a `<Text>`), and the gated toggle sat above the password checkbox while its hint read "add a password above". Both fixed in `33623f2`.
- The enabled hint reads "Enables unilateral exit in case Spark operators go offline" — that is the prototype's wording, but PRD 7.3 bans protocol vocabulary in user-facing strings. Worth a copy pass.
- The empty state reuses the "Before you start" title, which reads oddly above copy saying the backup comes later. See the first screenshot.
- The zip archive from PRD 6.2 is **not** included: there is no zip library in the tree and adding one is a native dependency, which does not belong in an otherwise pure-JS PR.

## Update: review fixes

Eight commits addressing a self-review pass. Each carries its own tests, and every fix with real logic was mutation-checked - reverting it fails the new tests, so none of them pass green either way.

**The account being backed up.** The bundle cloud-sync opt-in was recorded against the *active* account. Mid-migration that is still the custodial one, while the phrase being uploaded belongs to the provisioned self-custodial account - so the write was skipped and the user's ticked box was silently discarded, with the upload reporting success. The two answers are now one: `useBackupTargetAccountId` is the single source of truth for which account a backup flow acts on, and `useLoadWalletMnemonic` consumes it too. Note this restores the intent already documented in the migration PR ("the shared backup screens show its recovery phrase") - it was `useActiveWallet().isSelfCustodial` that broke it, because that flag also encodes SDK availability and reads false during startup. Three other backup hooks therefore also stop reading the wrong account in that window.

**A dismissed share no longer confirms a saved bundle.** `handleShare` swallowed the cancellation - correctly, a dismissal is not an error - but returned nothing, so both export paths advanced to "Emergency bundle saved" and completed the backup with no file having left the device. It now resolves false on a dismissal, a missing bundle, or an already-reported failure, and both call sites gate on it. Skip is also disabled while an export is in flight, so it can no longer tear the flow down mid-share.

**The bundle-saved dwell fires once.** Its timer was keyed on the identity of `completeBackup`, which the completion itself changes - so it re-armed after every firing: duplicated analytics and storage writes, and the success copy flipping while the user read it. The dwell is now focus-scoped rather than mount-scoped, because the success screen is swipeable back on iOS and this screen has no action of its own; a mount-only timer would have stranded the user there.

**Two smaller ones.** A failing settings read no longer skips the backup completion - it was inlined in the write's arguments, where a rejection aborts the call before `.catch` can attach, leaving the seed uploaded and the backup unrecorded. And a custom icon sized only by `sizeVariant` rendered at zero pixels, because this PR defaulted `width`/`height` to `0` while that branch never consulted the variant; no caller uses that form today, so it was a latent trap rather than a live regression.

The conditional props in both backup screens were pulled into named constants in a separate refactor commit.


grimen:
## Screenshots

Shot on an isolated simulator (iOS, light mode). The funded and enabled states are forced with a TEMP stub, since a demo simulator has no funded wallet — the copy and layout are real, the data behind them is not.

| Manual path — new wallet | Manual path — funded wallet | Download confirmation |
|---|---|---|
|  |  |  |

The first is the common case at signup: there is no bundle yet, so the screen explains rather than offering a file. Note the title reads "Before you start" above copy about something that happens later — flagged in the description as worth a copy pass.

| Cloud path — no password yet | Cloud path — password set |
|---|---|
|  |  |

Sync is gated on the seed backup carrying a password (D9), so the row is disabled until the checkbox above it is ticked. Provider name is platform-dispatched — Apple iCloud here, Google Drive on Android.

<sub>Images live on branch `assets/pr-4078-screenshots` — deleting it breaks them.</sub>

grimen:
## Redesigned against the mocks

The screen was built from video frames before the mocks existed. Comparing the two dated them: the video's version of this step still offers a **Cloud backup** button, which the mock has deleted — that is the no-hybrid rule (R7) applied, so the mocks carry the 2026-07-14 review's decisions and supersede the video.

| Light | Dark (matches the mock) |
|---|---|
|  |  |

**"One last thing"**, green emergency-kit hero, and copy that drops "or Blink" — a self-custodial exit depends on Spark operators, not on Blink — and fixes "use this file exit onchain".

| Sensitive warning | Saved (dark) | Saved (light) |
|---|---|---|
|  |  |  |

The warning's body regains the word the implementation had dropped: "emergency **bundle** recovery". **"Emergency bundle saved"** is new — the flow previously fell through to the generic backup-success screen.

### Skipping is now allowed

The mock offers **Skip** where this PR forced an export. A mock that applied R7 in the same screen did not forget R6 in it, so I took it as deliberate rather than an oversight.

Nobody is left unprotected: the encrypted on-device copy is written automatically for every self-custodial account (#3911), and the home nudge (#4080) returns once a bundle exists. What is mandatory is that a backup **exists**, not that the user exports it here.

**Copy** stays alongside Download — R10 offers it as a one-shot export with the 60s clipboard clear, and the mock does not argue against a second route.

### Both themes

The mocks are dark-mode only. Every colour is a theme token (`colors.primary`, `colors._green`, `colors.grey5`) and the new icon is authored so the transformer's `currentColor` substitution tints it, so light mode works by construction — shown above rather than assumed.

### Coverage

Export screen **100% statements / 100% lines**, saved screen **100%**. `CustomModal` gains a testID on its close button: modal dismissal was previously unreachable from tests, which is why two dismiss paths were uncovered.

<sub>Images live on branch `assets/pr-4078-redesign` — deleting it breaks them.</sub>

esaugomez31 (approved):

PR #4079, opened 2026-08-04
fix(home): keep the pending deposit amount visible beside the balance

grimen:
Fixes https://github.com/blinkbitcoin/blink-wip/issues/937 (cross-repo — close the issue manually on merge).

## Problem

When an onchain deposit arrives, a green `+$X` flashes under the home balance for about a second and disappears, while the transaction list below keeps showing the pending deposit. That flash was never a pending indicator: it is the unseen-transaction badge, which auto-dismisses after ~5s by design (#3655), and its timer often starts before the screen is actually visible. Nothing in the balance area is pending-aware, so once it dismisses the top of the screen stops reflecting the unconfirmed deposit.

## Fix

A state-driven pending pill on `BalanceHeader`'s existing (previously unused) `statusBadge` slot: **+$X pending**, amber, visible exactly while pending incoming funds exist, independent of the seen/unseen mechanism.

- **Custodial**: sums the home query's `pendingIncomingTransactions` (each converted by its own settlement currency, so USD-wallet pending works too).
- **Self-custodial**: sums immature (unconfirmed) Spark deposits via a new `usePendingDeposits` hook extracted verbatim from `UnclaimedDepositBanner`. Claimable/errored deposits stay with the banner, which carries the action to resolve them — the header pill only covers "waiting for confirmations", so the two never double-signal.
- The unseen badge keeps its current behavior; hide-amounts mode already suppresses the pill via `BalanceHeader`'s existing guard.
- New i18n key `HomeScreen.pendingReceiveBadge` in EN + all 28 locales (each reusing its existing "pending" term).

## Rides along: frozen pull-to-refresh spinner

Surfaced while recording the demo: the refresh control was bound to background query loading, and iOS renders a programmatically-pinned `UIRefreshControl` as a frozen, non-spinning spinner — on every home mount until the first response, and indefinitely while a self-custodial wallet connects. It now reflects only user-initiated pulls (which the gesture animates natively), and a failed pull unpins the spinner without escaping as an unhandled rejection.

## Tests

- `use-pending-receive-amount.spec.tsx` — both modes, currency mix, empty/loading/zero/negative/outgoing edges, non-immature exclusion, conversion rounding to zero.
- `use-pending-deposits.spec.ts` — the setState identity guard (status-only changes commit; unchanged refetches keep array identity), stale-fetch coordination, custodial no-op.
- `home.spec.tsx` — pill visible with a pending deposit, **still visible past the 5s unseen-badge auto-dismiss** (the regression itself), absent when nothing is pending; self-custodial wiring (immature shows, claimable stays with the banner); refresh spinner not pinned while queries load, spins only during a user pull, recovers on a failed pull.
- `balance-header.spec.tsx` — pill hidden while amounts are hidden.
- `unclaimed-deposit-banner.spec.tsx` passes unchanged, proving the fetch extraction preserved banner behavior.

Screenshots and a video demo are in the comments below (simulated pending state, captioned as such).

### DIFF
--- a/__tests__/components/balance-header/balance-header.spec.tsx
+++ b/__tests__/components/balance-header/balance-header.spec.tsx
@@ -145,4 +145,14 @@ describe("BalanceHeader", () => {
 
     expect(queryByTestId("balance-status-badge")).toBeNull()
   })
+
+  it("does not render the status badge while amounts are hidden", () => {
+    mockHideAmount = true
+
+    const { queryByTestId } = renderHeader({
+      statusBadge: { label: "+$1.00 pending", status: "warning" },
+    })
+
+    expect(queryByTestId("balance-status-badge")).toBeNull()
+  })
 })
--- a/__tests__/components/balance-header/balance-header.spec.tsx
+++ b/__tests__/components/balance-header/balance-header.spec.tsx
@@ -1,9 +1,11 @@
 import React from "react"
+import { StyleSheet } from "react-native"
 import { fireEvent, render } from "@testing-library/react-native"
 import { ThemeProvider } from "@rn-vui/themed"
 
 import theme from "@app/rne-theme/theme"
 import { BalanceMode } from "@app/hooks/use-balance-mode"
+import { StatusPill } from "@app/components/status-pill"
 
 import { BalanceHeader } from "@app/components/balance-header/balance-header"
 
@@ -144,6 +146,24 @@ describe("BalanceHeader", () => {
     })
 
     expect(queryByTestId("balance-status-badge")).toBeNull()
+  })
+
+  it("caps and shrinks the real pill and the centering ghost identically", () => {
+    // eslint-disable-next-line camelcase -- testing-library exposes this API verbatim
+    const { UNSAFE_getAllByType } = renderHeader({
+      statusBadge: { label: "+$1,234,567.89 pending", status: "warning" },
+    })
+
+    const pills = UNSAFE_getAllByType(StatusPill)
+    expect(pills).toHaveLength(2)
+
+    const [ghostStyle, realStyle] = pills.map((pill) =>
+      StyleSheet.flatten(pill.props.style),
+    )
+    expect(ghostStyle.flexShrink).toBe(1)
+    expect(realStyle.flexShrink).toBe(1)
+    expect(ghostStyle.maxWidth).toBeDefined()
+    expect(ghostStyle.maxWidth).toBe(realStyle.maxWidth)
   })
 
   it("does not render the status badge while amounts are hidden", () => {
--- a/__tests__/components/balance-header/balance-header.spec.tsx
+++ b/__tests__/components/balance-header/balance-header.spec.tsx
@@ -166,6 +166,18 @@ describe("BalanceHeader", () => {
     expect(ghostStyle.maxWidth).toBe(realStyle.maxWidth)
   })
 
+  it("forwards the badge onPress so the pill can carry an action", () => {
+    const onPress = jest.fn()
+    const { getByTestId } = renderHeader({
+      statusBadge: { label: "+$1.00 pending", status: "warning", onPress },
+    })
+
+    fireEvent.press(getByTestId("balance-status-badge"))
+
+    expect(onPress).toHaveBeenCalledTimes(1)
+    expect(mockSwitchMemoryHideAmount).not.toHaveBeenCalled()
+  })
+
   it("does not render the status badge while amounts are hidden", () => {
     mockHideAmount = true
 
--- a/__tests__/components/balance-header/use-pending-receive-amount.spec.tsx
+++ b/__tests__/components/balance-header/use-pending-receive-amount.spec.tsx
@@ -2,6 +2,7 @@ import { renderHook } from "@testing-library/react-native"
 
 import { usePendingReceiveAmount } from "@app/components/balance-header/use-pending-receive-amount"
 import { TransactionFragment, TxDirection, WalletCurrency } from "@app/graphql/generated"
+import { DisplayCurrency } from "@app/types/amounts"
 import { DepositStatus, PendingDeposit } from "@app/types/payment"
 
 const mockConvertMoneyAmount = jest.fn()
@@ -10,9 +11,9 @@ const mockFormatMoneyAmount = jest.fn(
     `$${(moneyAmount.amount / 100).toFixed(2)}`,
 )
 const mockUseActiveWallet = jest.fn()
-const mockUsePendingDeposits = jest.fn()
 
 jest.mock("@app/hooks", () => ({
+  ...jest.requireActual("@app/hooks"),
   usePriceConversion: () => ({ convertMoneyAmount: mockConvertMoneyAmount() }),
 }))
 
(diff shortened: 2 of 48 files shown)


grimen:
Screenshots (iOS simulator, self-custodial account). The pending state is simulated — the badge was TEMP-forced with a fixed `$102.30` amount to capture the render; the data wiring is covered by the jest specs.

### Pending badge

| While a deposit is pending | No pending deposit |
|---|---|
|  |  |

### Frozen refresh spinner (ride-along fix)

Same screen state (queries loading in the background), same crop. Before: the refresh control is pinned by the loading state and iOS renders it as a frozen, non-spinning spinner. After: it only appears for actual pull gestures.

| Before | After |
|---|---|
|  |  |

<sub>Images live on branch `assets/pr-4079-screenshots` — deleting it breaks them.</sub>

grimen:
Demo (iOS simulator, 12s video — replaces the earlier GIF). Fresh app open onto the home screen with a pending deposit: the green unseen-transaction badge appears and **dismisses itself after ~5 s** (that transient badge was the only top-of-screen signal before this PR — the reported "flash"), while the amber **+$102.30 pending** pill stays for as long as the deposit is unconfirmed.

The clip also shows the second fix in this PR: no more frozen, pinned pull-to-refresh spinner while queries load — the refresh control now only appears for actual pulls.

Simulated state, captioned honestly: the pill and the unseen badge were TEMP-forced with a fixed amount (the unseen badge's production 5 s auto-dismiss re-enacted with a timer), since a real unconfirmed onchain deposit can't be conjured on demand in the sim. The real data wiring is what the jest specs cover.

https://github.com/user-attachments/assets/3943503c-bd48-4c11-a5b6-836e300f6b35

kngako:
Can the pending transaction not be above the balance text instead of being to the top right? 

Something like this?

image psd(3)

kngako (changes requested):
Pending transaction might be better centered above balance line.

esaugomez31:
Reviewed the branch and pushed the fixes for everything I found. Summary of what changed and how it was verified.

## Fixed

**1. Duplicate deposit fetching in self-custodial mode** — `usePendingDeposits` kept state per hook instance, and two instances were live on the SC home screen (`UnclaimedDepositBanner` and `usePendingReceiveAmount`). Each fired on mount through `useFocusEffect` *and* through the `wallets` effect, so `sdk.listUnclaimedDeposits()` ran 4 times per mount instead of 2, and doubled again on every wallet refresh. Worse, the two copies could disagree: `createListPendingDeposits` swallows SDK errors and returns `{ deposits: [] }`, so one instance failing while the other succeeded left the banner and the header pill showing different things.

The screen now owns the fetch and feeds both consumers, `usePendingReceiveAmount({ deposits })` and `<UnclaimedDepositBanner deposits={...} />`. They read the same array, so the divergence is gone by construction. A shared provider was not an option: `usePendingDeposits` relies on `useFocusEffect`, which needs a screen context.

**2. The hook dropped the adapter's `errors` and committed an empty list** — `createListPendingDeposits` returns `{ deposits: [], errors: [...] }` when the SDK fails, and the hook only read `deposits`. A transient failure therefore cleared both the badge and the banner, which reads to the user as "your deposit confirmed". The hook now keeps the last known deposits when the listing reports errors. A legitimate empty result still commits, so a deposit that actually confirmed still clears.

**3. The Arabic string said "being sent" on an incoming badge** — `ar.json` reused `common.pending` = `يتم الارسال`, literally *being sent*, on a pending receive. Now `قيد الانتظار`, the term the same file already uses for transaction status. Same pattern fixed in `vi` (`đang chờ` instead of `Đang xử lí`, "processing") and `ja` (`保留中` instead of `保留`). The other 25 locales already matched their own in-file term.

**4. A test that could not fail** — `"sums pending transactions across BTC and USD wallets"` used a converter that discarded `currency`, so it passed identically whether the code converted by `tx.settlementCurrency` or hardcoded BTC. The converter is now currency-aware and the call args are asserted. Confirmed by mutating the source to hardcode BTC: the test now fails where it used to pass.

**5.** `use-pending-receive-amount.spec.tsx` mocked the whole `@app/hooks` barrel with no `requireActual`.

**6.** The `useFocusEffect` mock invoked its callback during render instead of from an effect, so the focus-effect / wallet-effect ordering under test was not the app's. It now runs through a real `useEffect` that honours the cleanup. Worth noting the cleanup itself stays uncovered: it only guards a `setState` after unmount, which React already drops, so there is nothing a test can observe.

**7.** `sameDeposits` compares only `id` + `status`, so `requiredFeeSats` and `errorMessage` changes never commit. Fine for both current consumers, and the guard is deliberate, but the doc comment promised more freshness than it delivers. It now says so, and points at widening the guard before rendering `requiredFeeSats` off this hook.

**8.** Removed an unused `// eslint-disable-next-line prefer-const`.

## Verification

`tsc` and `eslint` clean, full affected suites green (1077 tests / 70 suites).

Device-tested on an Android emulator against an empty self-custodial account, with deposits driven by a temporary stub that was reverted afterwards (md5 checked against the original). Covered: the pill appearing for an immature deposit and **still being there past the 5s unseen-badge window**; a claimable deposit showing the banner and no pill; immature plus claimable rendering consistently across both; hide-amounts suppressing the pill; the three corrected locales on screen; and the refresh spinner not pinning on mount. For the error guard I ran it both ways, with the guard the pill and banner survive a failing listing, without it they both vanish.

## Needs a decision before it can be fixed

The banner counts **immature** deposits alongside claimable ones and labels the total *"available to claim"*. With an immature 45k and a claimable 20k it announces 65,000 sats claimable when only 20,000 are, and the same immature deposit appears in the pill above and in the banner below.

04-pill-pendiente-deposito-immature


I left it as is because the fix depends on which behaviour we want, and that is a call to make together rather than to assume:

- the banner stops counting immature deposits, leaving the pill as the single "waiting for confirmations" signal, which is what this PR describes; or
- the banner keeps counting them but splits the wording, so the claimable total stays accurate and the immature part reads as pending.

Either way it is a small change in `unclaimed-deposit-banner.tsx`, which today filters `Refunded` only. Happy to implement whichever we settle on.

grimen:
Went with **option 1** — the banner stops counting immature deposits (38b56e2), so its "available to claim" total is accurate by construction and the pill is the single "waiting for confirmations" signal, matching what this PR describes.

One consequence your screenshot hints at: with immature-only deposits the banner disappears, and it was the sole route into the unclaimed-deposits screen — the deposit's txid / mempool link became unreachable while it confirms. Covered that by making the pill itself tappable when immature deposits exist: it opens the same screen (b978288). Custodial pending receives keep an inert pill.

Implemented with tests: the banner filter (Immature and Refunded excluded, Claimable / FeeExceeded / Error kept), the pill navigation, and regression tests for the pill surviving the unseen-badge window and reading the server-locked `settlementDisplayAmount` — so the pill and the unseen-tx badge can never show different amounts for the same receive. The touched files hold 100% coverage; the rest of the pushed commits address the remaining review findings (predicate on account type, deposit reset on account switch, bounded pull-to-refresh wait, pill truncation/font-scaling caps).

Simulator demo (staging; the pending state was driven by a temporary input stub, reverted afterwards — the pill, navigation, and screen logic are the real branch code):

pending pill demo

<sub>Demo lives on branch `assets/pr-4079-demo` — deleting it breaks the embed.</sub>

kngako (approved):
image

image

PR #4080, opened 2026-08-04
feat(self-custodial): surface recovery-backup freshness in settings and on home

grimen:
**Stack 5/5** for the unilateral-exit recovery bundle. Base: `feat--bundle-backup-onboarding` (#4078). Merge bottom-up.

Phase 2 preparedness surfaces from [blink-specs#26](https://github.com/blinkbitcoin/blink-specs/pull/26) §6.4.

## Problem

A recovery backup that has fallen behind the wallet is worth as much as no backup at all for whatever arrived since. Nothing surfaced that: freshness lived on a settings screen the user had no reason to open.

## What this adds

- **Status chip** on the Settings → Recovery backup row: Backed up / Out of date / Not set up.
- **Home-screen banner**, two variants (below).
- **"How recovery works"** explainer on the Recovery Backup screen — what it costs, how long it takes, that small amounts may not be worth recovering.
- **Dollar-balance note** (R3): on-chain recovery moves Bitcoin only, so a user holding Dollars is told plainly they are not covered.

## The banner's two variants differ in kind

**Funds with no recovery backup at all** is a condition, not a reminder — automatic refresh should prevent it, so reaching it means refresh is off or failing and the user has money with no recovery path. **Not dismissible.**

**Backup out of date** is a reminder, so it can be dismissed. The dismissal covers the staleness the user actually saw: once the bundle refreshes and falls behind again, the reminder returns rather than staying silent forever.

An empty wallet with no backup shows nothing — there is nothing that could be backed up yet.

## This is where R6 is actually met

"Every backup path ends with the bundle saved" cannot be satisfied at onboarding: the wallet is empty, the exporter refuses to build a bundle with no outputs, and there is nothing to hand over (see #4078). The first real opportunity to put a backup in front of the user is here, once funds exist. That makes this PR the other half of #4078 rather than a follow-up.

## Notes

- Chip and banner read one shared status hook, so the two surfaces cannot disagree about whether a backup is current.
- The staleness window is the refresh scheduler's own 24h fallback. A second threshold would show "up to date" while a refresh was already pending; the constant is exported and asserted against the scheduler's value.
- A backwards device clock is treated as stale rather than fresh, so a wound-back date cannot hide an out-of-date backup.

## Tests

2247 passing across the touched areas. New: the freshness predicate including the boundary and the backwards-clock case, all four banner outcomes plus the dismissal-vs-refresh interaction, and render assertions for the Dollar note and the explainer.


grimen:
## Screenshots

Shot on an isolated simulator (iOS, light mode).

| Recovery Backup screen | How recovery works | Settings row chip |
|---|---|---|
|  |  |  |

The Dollar note and the chip's "Out of date" state are forced with a TEMP stub — a demo simulator has no Dollar balance and no saved bundle to age. The copy, layout and pill styling are real.

The home banner is not shown: it needs a self-custodial account with a balance, which a demo simulator cannot produce. Its logic is covered by tests instead — all four outcomes plus the dismissal-vs-refresh interaction.

<sub>Images live on branch `assets/pr-4080-screenshots` — deleting it breaks them.</sub>

grimen:
## Freshness now means "the wallet changed", not "the clock moved"

Pushed in `4bace2b`, after a good question on the screenshot: if the backup syncs automatically, why would it ever say "out of date"?

It shouldn't have. The chip was comparing timestamps — 24 hours since the last save — but the bundle records **which outputs the wallet owns**, and those only change when the user transacts, which is exactly when the refresh scheduler rebuilds it. A wallet left untouched for a week had a backup that described it perfectly and got a warning anyway.

That is a false alarm, and false alarms are how users learn to ignore the real one.

**What it does now:** compares the balance recorded in the bundle against the wallet's current balance — a local read, no network call. Same balance means up to date whatever its age. Different balance means out of date even if the backup is a minute old.

**Age survives only as a backstop,** for what a balance comparison cannot see: a swap or consolidation that moves the underlying outputs while the total stays put. Thirty days, deliberately *not* the scheduler's 24h fallback — that number is a good trigger to re-fetch and a bad claim to make to the user.

A Dollar balance is excluded from the comparison, since only Bitcoin is covered by on-chain recovery (R3).

**Not done, and worth deciding separately:** true background sync on a timer, the way messaging apps back up. The refresh only runs while the app is open, so a payment whose refresh failed stays unfixed until next launch. That needs platform background scheduling and is a bigger change than this PR.

Coverage: 19 tests on the freshness module (97% statements, 96% branches), including the unchanged-wallet-stays-fresh case that motivated this, the backstop boundary, a backwards clock, and the Dollar-wallet exclusion.

grimen:
## Correcting a claim #4078 makes

#4078 allows skipping the export and justifies it by saying "the home nudge returns once a bundle exists". **It didn't.** The nudge fired only on `Missing` or `Stale`; a user who skips has a *fresh* bundle, so it stayed silent — and for a manual-path user that bundle then exists only on the phone that might be lost.

That is exactly the risk [raised on blink-wip#807](https://github.com/blinkbitcoin/blink-wip/issues/807#issuecomment-5182691347). Rather than take Skip away, this makes the justification true.

**What changed**

- The bundle records when it last left the device by the user's own hand — written only after a share sheet actually **resolves**. A cancelled share is not a copy of anything, and counting it would silence the warning on a lie. A clipboard copy counts too (R10 export).
- A third nudge variant fires when a bundle exists, has never been exported, and has never reached the user's cloud: *"Your emergency bundle is only on this phone."*
- It is checked **before** staleness. Having no copy off the device is a bigger problem than having an out-of-date one, and saying both at once says neither well.
- Cloud-sync users never see it — their copy is already elsewhere.
- Dismissal matches the stale variant: it covers the bundle the user saw, and returns once the bundle is refreshed and still unexported.

**Its copy points at saving a file, never at cloud.** A cloud offer in this surface would rebuild the manual/cloud hybrid through the back door, in the place most likely to be seen — and that hybrid was ruled out deliberately.

**Coverage:** 2120 tests green across the touched areas. New cases name what they protect — a fresh-but-unexported bundle still warning, precedence over staleness, dismissal returning after a refresh, the cloud-synced case staying quiet, and a cancelled share **not** counting as an export.

### Worth flagging separately

The PRD's **R8** still reads *"Multiple backups are permitted… a user who finished the manual path can later come back and complete the cloud path too"*, and #3911's status table lists it as implemented. That is the opposite of the decision that manual and cloud are one-or-the-other. The code already matches the decision — `shouldShowSettingsBanner = !isBackedUp && isSelfCustodial` hides the method picker once a backup exists — but the written rule should be revised in `blink-specs#26`, or the next implementer builds the hybrid from it and is right to.

grimen:
## Findings

**1. Missing `status-pill` import breaks the build (blocking).**

The settings chip imports `StatusPill` from `@app/components/status-pill`, but no such module exists anywhere in the repo, on this branch or the base branch.

*Example:*
`app/screens/settings-screen/settings/recovery-backup.tsx:4` imports `{ StatusPill, type StatusPillVariant } from "@app/components/status-pill"`. The path resolves nowhere, so `yarn tsc:check` fails with `TS2307`, and the new `__tests__/screens/settings-screen/recovery-backup-setting.spec.tsx` (which renders this component directly, unmocked) fails to load for the same reason.

*Proposal:*
Add the missing `StatusPill` component, or point this at an existing status-chip component if one already covers this case.

**2. Export/cloud-sync timestamps aren't scoped to the current bundle (blocking).**

`isOnlyOnThisDevice` treats `exportedAt`/`cloudSyncedAt` as permanent once set, even though a new bundle can be saved afterward without ever being re-exported or re-synced.

*Example:*
`use-recovery-bundle-status.ts:184-188` computes `isOnlyOnThisDevice` from `saved.exportedAt`/`saved.cloudSyncedAt`, but `exportedAt` is stored per-account with no link to a specific bundle version, and `cloudSyncedAt` is explicitly carried forward on every refresh (`refresh.ts:214`). A user who exports once, then later gets a fresh unexported bundle after a payment, will permanently see `isOnlyOnThisDevice: false` — the nudge meant to catch this never fires again.

*Proposal:*
Scope both timestamps to the bundle they actually cover (e.g. compare against the current `savedAt`/bundle id) instead of carrying them forward across refreshes.

## Test Coverage Gaps

- `use-recovery-bundle-status-hook.spec.ts` only checks `isOnlyOnThisDevice` against a single bundle snapshot (exported-once vs. never-exported). No test covers a bundle refresh that changes `savedAt`/`totalSats` while `exportedAt`/`cloudSyncedAt` carry over from the prior bundle — the exact scenario in finding 2.

grimen (commented):
Invalid

esaugomez31 (approved):

PR #4083, opened 2026-08-05
feat(send): let custodial onchain payments choose a payout speed

### LINKED ISSUE #4053: feat(fees): UI/UX for onchain fee tier selection (slow/normal/fast)
## Description (Updated)

**Each payout speed needs to show the user what the expected payout fee will be.** 

There also needs to be an update on calling the correct mutation to match the users choice. 

The onchain fee tier selector already exists in the live release, but the label "Fee speed" is confusing/weird in the current UI.

Refine the label/copy for the slow / normal / fast fee tier selection so users understand they are choosing onchain confirmation speed and fee tradeoff.

## Requirements

- Replace or improve the "Fee speed" label with clearer wording
- Keep the existing slow / normal / fast fee tier selector behavior
- Preserve fee amount and estimated confirmation-time display for each tier
- Confirm the final copy with design/product before release

## Related

- https://github.com/blinkbitcoin/blink-wip/issues/899 — Fees: On-chain fee restructure (3-tier)
- https://github.com/blinkbitcoin/blink-mobile/issues/2821 — Inaccurate onchain fees (related but separate)
- https://github.com/blinkbitcoin/blink-mobile/issues/4053#issuecomment-5178554359 — design feedback that the selector exists but the label is awkward

esaugomez31:
Closes #4053

## What

Custodial on-chain payments can now choose a payout speed, and each speed shows the fee it would actually cost.

Until now the selector only existed for self-custodial sends. On the custodial rail the app never sent `speed`, so every on-chain payment silently went out on the schema default (`FAST`) with no way for the sender to choose, and no fee shown per option.

## Changes

**The payment carries the chosen speed.** All four send mutations (`onChainPaymentSend`, `onChainPaymentSendAll`, `onChainUsdPaymentSend`, `onChainUsdPaymentSendAsBtcDenominated`) and the three fee queries now pass `speed`. The payment detail owns it, so it survives amount edits and reaches the mutation that finally broadcasts.

**A fee per speed, in one round trip.** The three tiers are quoted through a single aliased query rather than three requests. A payment detail also declares which of the three fee endpoints answers for it, and that declaration is written beside the branch that picks the endpoint, so the selector cannot drift from what the send will use.

**The selector shows for custodial too.** It was gated on `isSelfCustodial`; it now depends only on the payment being on-chain. A spinner replaces the caret while a quote is in flight, so the fee on screen always belongs to the amount on screen.

**Next waits for the quote.** The high-fee warning is judged by the fee the selector quoted, so leaving the screen before the quote lands is leaving without the warning. Next is held while a custodial quote is in flight, and released whatever the outcome: a quote that fails still lets the sender continue, since a custodial quote is only an estimate and the confirmation screen fetches its own. One box, one message, so a fee error never displaces a message about the amount itself — an amount over the daily limit keeps both its message and its "increase limits" route out.

**Copy.** The field title moves from "Fee speed" to "Transaction priority" across the 28 translated languages, matching the Figma design. `Fee speed` reads as a compound that does not mean anything: a fee has no speed. What the sender picks is how the transaction is prioritised, and the fee follows from that. The tier names (Fast / Medium / Slow) are unchanged. The fee-error string both send screens show now lives in `common` rather than the details screen reaching into the confirmation screen's namespace; every locale kept the translation it already had.

## Fixes found along the way

**The confirmation screen served stale fees.** The three on-chain fee queries ran under Apollo's default `cache-first`, so revisiting the same amount replayed a quote from an earlier visit. On device this showed as the details screen and the confirmation screen disagreeing: 5,765 sats against 5,013 sats for the same transaction, the second answering in 44 ms because it never hit the network. On-chain fees move with the mempool, so they are now fetched `no-cache`. This one predates this branch and is isolated in its own commit.

**The high-fee warning measured the wrong tier.** It probed the network rate on the `FAST` queue regardless of what the sender picked, so choosing the cheap tier could still trigger a warning about a fee that was not being paid. It now follows the selection, and discards out-of-order probe responses so the last request wins rather than the last to return.

**Fees are formatted in the unit they came in.** A tier now carries its unit (sats, cents, or a sat/vB rate) instead of leaving the consumer to infer it from the rail. A USD wallet's cents fee can no longer be rendered through the sats formatter.

**A fee of zero is told apart from one that was never quoted.** The label used to drop the fee whenever the amount was zero, which made "still loading" and "this really is free" render identically. It now asks whether a quote exists for the inputs on screen, so an on-chain address that resolves intraledger reads as `Fast (0 sats)` while a tier nobody has priced yet reads as a bare `Fast`. Both rails hide the fee until a quote lands; the spinner stays custodial-only, since the SDK answers fast enough that one would merely flicker.

**Only unexpected fee-quote failures are reported.** Apollo's lazy-query `execute` resolves on failure rather than rejecting, so the `catch` around a quote only ever saw a synthetic "missing data" error: the backend's cause never reached Crashlytics, and because that string matches no connectivity pattern, transport blips were filed as defects instead of being downgraded. The quote now classifies from the resolved `error` — a rejected amount or address is an expected outcome, a transport failure carries its own cause, and only an unexplained empty response is recorded.

**The by-speed quotes are off the retry link.** Their single-speed counterparts already sit in `noRetryOperations`; the three new ones did not, so the global `RetryLink` resent them up to five times with backoff behind the sender's back.

**A tier change keeps the sender's note.** Rebuilding a self-custodial payment starts from the destination, which carries no note, so switching speed dropped whatever had been typed. The custodial rail rebuilds from its own params and never lost it, so the same action behaved differently per rail.

## Screenshots

Custodial account on staging, USD sending wallet.

| Scenario | What it shows | Screenshot |
|---|---|---|
| On-chain address, no amount yet | Tiers read `Fast` / `Medium` / `Slow` with no fee attached | 01-no-amount-selector |
| Speed list open, still no amount | All three bare, and Standard's `~ 4h` window | 02-no-amount-list |
| Amount entered | The quote lands and each tier carries its fee | 03-amount-quoted |
| Speed list open with fees | Every tier priced, ETAs alongside | 04-list-quoted |
| Confirmation screen | The same fee the details screen showed | 05-confirmation |

## Testing

Every fee-tier module sits at 100% statements, branches, functions and lines. Type-check, lint and translation checks are clean.

Device-verified on a custodial account against staging: the tiers carry no fee until one is quoted, the picked speed reaches the payment, and the details and confirmation screens quote the same fee for the same transaction.

> [!NOTE]
> The queue work this depended on has landed: blink-deployments#10052 added the dedicated `slow` queue and retimed `medium`, and #10053 pointed `SLOW` at it. `payoutSpeeds` on staging now advertises Priority ~10 minutes, Standard ~4 hours and Flexible ~24 hours, and the client's windows match.
>
> Two things to keep in mind. All three speeds price against the mempool's `fastestFee`, so their quotes coincide whenever the mempool is at the floor; the cheaper tiers earn their discount from batch amortisation, not from a lower rate. And `payoutSpeeds` exposes only a prose `description`, so the client mirrors these windows in a constant rather than reading them — a numeric field would remove that duplication.


kngako (changes requested):
Seems the fee estimation on self custodial is a little more reasonable.


image

I have no beef with the code in it's current form.


Opus 5 review follows:

---



## 1. The selector shows "(0 sats)" before an amount is entered

I confirmed this by running the real hook: with no amount entered, `isQuoting` is `false` and the tiers are zeros, so [[fee-tier-options.ts:49](https://claude.ai/epitaxy/app/screens/send-bitcoin-screen/fee-tier-options.ts:49)](app/screens/send-bitcoin-screen/fee-tier-options.ts:49) takes the non-quoting branch and renders:

```
Fast (0 sats)      ~ 10m
Medium (0 sats)    ~ 60m
Slow (0 sats)      ~ 24h
```

`paymentDetail` is seeded with `zeroDisplayAmount` in the mount effect, and the selector renders on `isOnchain` alone with no amount guard, so this is the *default* state of the send screen for a custodial user with an on-chain address. It reads as "this send is free."

This existed on `main` for self-custodial, but the branch takes it from a niche rail to every user — which makes it this branch's problem now. Commit 98da6a7 already established the right principle (an unquoted fee ≠ a zero fee); the gap is that "no amount yet" is a third state the `isQuoting` flag doesn't cover.

## 2. The first-render `isQuoting` fix never fires in production

[[use-custodial-onchain-fee-tiers.ts:209](https://claude.ai/epitaxy/app/screens/send-bitcoin-screen/hooks/use-custodial-onchain-fee-tiers.ts:209)](app/screens/send-bitcoin-screen/hooks/use-custodial-onchain-fee-tiers.ts:209) starts `isQuoting` true when the inputs are already in hand. But the send screen builds `paymentDetail` in an effect, so the hook always mounts with `paymentDetail === null` and the initializer always evaluates to `false`. Same for [[use-recommended-fee-tiers.ts:35](https://claude.ai/epitaxy/app/screens/unclaimed-deposits/hooks/use-recommended-fee-tiers.ts:35)](app/screens/unclaimed-deposits/hooks/use-recommended-fee-tiers.ts:35) — `enabled` is `refundDepositId !== null`, which is always `false` at mount.

The tests that cover this (`"quotes from the very first render when the params are already in hand"`, and the sibling in the recommended-tiers spec) mount with the params ready — a path neither screen takes. So they pass while the production path still has a render of `(0 sats)` when the user commits an amount, before the effect sets the flag.

Both this and #1 are the same root cause: `isQuoting` is stored in state and set from an effect, so it's a lagging indicator. Deriving it during render fixes both at once — track which inputs the current tiers were quoted *for*, and show the fee only when they match:

```js
const inputsKey = walletId && address && amount && quote
  ? `${walletId}|${address}|${amount}|${quote}` : null
const [quotedKey, setQuotedKey] = useState(null)  // set alongside setTiers on success
const hasQuote = inputsKey !== null && quotedKey === inputsKey
```

Then `buildFeeTierOptions` keys the label on `hasQuote` instead of `!isQuoting`: no amount → bare tier names, quote in flight → bare names + spinner, quote landed → names + fee. It's correct on the very render the amount changes, with no effect lag.

## Suggestion for 1 & 2 

Store a key identifying which inputs the current tiers were quoted for, and derive both hasQuote and isQuoting from it during render — on the render where the amount changes, the stored key still names the old amount, so it reads correctly with no effect lag. Then flip fee-tier-options.ts:49 to key the label on hasQuote rather than !isQuoting, which covers "no amount yet" and "quote in flight" with the same predicate. The three "quotes from the very first render…" tests need rewriting to mount without the inputs and rerender with them, since that's the only sequence the screens actually perform.

## 3. Self-custodial tiers go stale on error

[[use-onchain-fee-tiers.ts:106-109](https://claude.ai/epitaxy/app/screens/send-bitcoin-screen/hooks/use-onchain-fee-tiers.ts:106)](app/screens/send-bitcoin-screen/hooks/use-onchain-fee-tiers.ts:106) sets the error but never resets `tiers`, and unlike the custodial hook it doesn't clear them when a new request starts either. So a re-quote that fails leaves the *previous* amount's fees labeled on screen next to the error message. The custodial hook does `setTiers(EMPTY_TIERS)` in both places — worth making the two rails consistent.

**Suggestion:** Reset to DEFAULT_TIERS on both failure paths, matching what the custodial hook already does. Fixing #1 masks most of the damage, but the two rails shouldn't disagree about what an error leaves on screen.

## 4. `useRecommendedFeeTiers` lost its sibling's stale guard

[[use-recommended-fee-tiers.ts:68-69](https://claude.ai/epitaxy/app/screens/unclaimed-deposits/hooks/use-recommended-fee-tiers.ts:68)](app/screens/unclaimed-deposits/hooks/use-recommended-fee-tiers.ts:68) — both the `catch` and the `finally { setIsQuoting(false) }` run unconditionally; there's no `requestTokenRef` like the other two hooks have. Toggling refund mode off and on while a request is in flight lets the old resolution clear the new one's spinner. Narrow window, but the fix is the pattern already used twice next door.

**Suggestion:** Port the requestTokenRef pattern verbatim from the two neighbours; it's a narrow window but the fix already exists twice in the codebase. Its lazy initializer has the same defect as #2, since refundDepositId is always null at mount, so the derived approach applies here too with enabled as the whole key.

grimen (changes requested):
## Findings

**1. The pre-amount state renders a fee that was never quoted, which is the ambiguity `isQuoting` was added to remove.**
`buildFeeTierOptions` only omits the fee from the label while `isQuoting` is true. When no amount has been entered yet, `useCustodialOnchainFeeTiers.fetchFees` hits the `!walletId || !address || !amount || !quote` gate, sets `tiers` to `EMPTY_TIERS` and `isQuoting` to false — so the selector, which `send-bitcoin-details-screen.tsx` renders whenever `isOnchain`, shows `Fast (0 SAT)` from the zeroed placeholder. That state is indistinguishable from the genuine intraledger zero the PR description cites as the reason a settled zero must render as `(0 sats)`.

*Example:* reproduced on this branch on a staging custodial account — on-chain address entered, amount not yet set:

| Details screen, no amount set | Speed list open, same state |
|---|---|
|  |  |

Every tier claims `(0 SAT)` before anything was quoted — the same rendering an intraledger-resolving address would legitimately earn.

*Proposal:* one option is a third state alongside `isQuoting` (e.g. `hasQuoted`, or treating the gated-off case as "hide the fee"), so the label shows a fee only when a quote is in flight or has settled; if showing zero before an amount is deliberate, a comment on the gate noting so would help.

**2. The custodial quote error promises a retry that does not exist.**
`feeEstimateError` reads "Unable to estimate fees. Please try again," but nothing re-invokes `fetchFees` in `useCustodialOnchainFeeTiers` except a change to `walletId`, `address`, `amount`, or `quote`. A user who keeps the same amount has no action that triggers the retry the copy asks for; the error persists until they alter the payment.

*Example:* a transient network failure during the quote leaves the error on screen; reopening the tier list re-renders the same failed state.

*Proposal:* one option is re-fetching when the selector opens (or on an explicit tap of the error); alternatively drop "Please try again" from the copy, which is possibly fine to defer since the error is non-blocking.

## Test Coverage Gaps

- The details screen's changed gating is asserted only at the hook level. `isNextDisabled` now blocks solely on `isFeeTierErrorBlocking` (custodial quote failures no longer disable Next), and the selector's render condition widened from `isSelfCustodial && PaymentType.Onchain` to `isOnchain` — neither is exercised by a `send-bitcoin-details-screen` test, which is also where finding 1's `(0 SAT)` label would surface.

## Code Smells

- `useOnchainFeeTierOptions` now carries both rails inline: six parallel locals (`selfCustodialTiers`/`custodialTiers`, two error messages, two quoting flags) merged through repeated `isSelfCustodial ?` ternaries. Extracting each rail's quoting into its own hook and selecting once at the top would localize the next rail-specific divergence.

<sub>Images live on branch `assets/pr-4083-review-screenshots` — deleting it breaks them.</sub>

designsats:
@esaugomez31 

Figma screen: https://www.figma.com/design/9MQuQi8ZhXVvDibWSI3C4c/Blink?m=auto&node-id=1248-14631&t=YhJoSqggisNoc4AR-1

Change label to "Transaction priority" 

send - step2-onchain

kngako:
@esaugomez31, @littledino2112 has merged the changes on required in the backend for payout speed.

https://github.com/blinkbitcoin/blink-mobile/issues/4053#issuecomment-5235566116

esaugomez31:
@grimen @kngako — every point from both reviews is addressed. Here is what the code does now.

**A tier label carries a fee only once one has been quoted for it.** The label asks whether a quote exists for the inputs on screen rather than whether a request is running, so both leaking states are covered by one predicate: with no amount entered the tiers read `Fast` / `Medium` / `Slow` with no fee, and the same while a quote is in flight. A settled zero still renders as `(0 sats)`, so an address that resolves intraledger is unchanged.

**That state is derived during render.** It comes from a key naming which inputs the tiers on hand were quoted for. On the render the amount changes, the stored key still names the previous amount, so the fee disappears with no lag. The flag it replaces was written from an effect and always trailed by a render.

**Both rails agree on what a failed quote leaves behind.** A self-custodial failure no longer leaves the previous amount's fees in `tiers`; only the success path puts numbers back, matching the custodial rail. This mattered past the label, since the refund screen reads `tiers` directly.

**Stale resolutions are discarded on every rail.** The refund screen's recommended-fee hook now carries the request guard its two siblings already had, so leaving and re-entering refund mode mid-flight cannot let an older resolution clear a newer one's state.

**The fee error no longer promises a retry that does not exist.** Nothing re-invokes the quote unless the payment changes, so "Please try again" asked for something the user could not do. It now reuses the confirmation screen's existing wording — *"Unable to calculate fee"* — and the key this branch had added is gone, along with its 28 translations.

**The fee reads the same on both screens.** The details screen formats through the display currency the way the confirmation screen already did. Formatting the wallet amount directly rendered a USD-wallet fee in dollars and a BTC-wallet fee in sats whatever the user's display currency was, and landed a cent off the confirmation screen for the same quote.

**Standard's ETA matches its payout queue.** `MEDIUM` moved off Bria's half-hour queue in blink-deployments#10052, so the advertised `~1 hour` was under-promising. It now reads `~4h`, matching `payoutSpeeds` on staging. `FAST` and `SLOW` were already right.

**The field is titled from the design.** The label reads **Transaction priority**, per the Figma screen, across all 28 translated languages.

**Each rail's quoting lives in its own hook.** Both return one shape and the rail is selected once at the top, so the consumer no longer pairs six parallel locals with repeated `isSelfCustodial` ternaries. The single field the rails deliberately answer differently is the spinner, which stays custodial-only and says so.

**The details screen's gating is exercised end to end.** The selector's render condition for a custodial on-chain send, the tier list carrying no fee before an amount, the picked tier being adopted, and Next staying enabled when a custodial quote fails.

---

Every fee-tier module is at 100% statements, branches, functions and lines. Type-check, lint and translation checks are clean. Verified on device against staging: the tiers carry no fee until one is quoted, the picked speed reaches the payment, and both screens quote the same figure.

One follow-up worth filing: `payoutSpeeds` exposes only a prose `description`, so the broadcast windows stay mirrored in a client constant. A numeric field would let the app read them instead.

kngako:
I'm also seeing the same issue. So the maybe the backend deployment isn't rolled out yet?  

image

kngako (changes requested):
## Findings

### 1 — Important: the high-fee alert probe is not keyed to `payoutSpeed`, so tier switches leave it measuring the wrong queue

`app/screens/send-bitcoin-screen/hooks/use-onchain-fee-alert.ts:52,77-78`

```ts
const [onChainTxFee, setOnChainTxFee] = useState(0)
...
if (fees) {
  setOnChainTxFee(fees)
```

The effect correctly re-probes when `payoutSpeed` changes and guards with `isStale`,
but the *stored* probe value is never invalidated. Between the tier switch and the new
probe resolving — and permanently, if the new probe returns no data — the comparison
uses the previous speed's rate.

**When it comes up:** a user selects Flexible (low probe fee), then switches to
Priority and immediately taps Next on a payment whose Priority fee exceeds 50% of the
amount — no high-fee modal appears, because the check still holds the Flexible probe.
The reverse also occurs: a stale-high probe can spuriously trigger (or suppress) the
modal after switching to Flexible. Before this PR the probe didn't follow the selector,
so it couldn't go stale in this dimension; now that the hook opts into per-speed
probing, it needs the same keying the tier hooks received.

**Proposal:** reset `setOnChainTxFee(0)` when `payoutSpeed` changes (or key the state
by speed, e.g. store `{ speed, fee }` and only compare when `speed === payoutSpeed`).
Also decide whether the alert should suppress or fail open while the probe for the
current speed is unresolved — suppressing matches today's behavior, but the choice
should be explicit and commented. Pin it with a test that switches tier after a
resolved probe and asserts the modal reflects the new speed's rate.

### M1 — Minor: the reworked alert effect still has an unhandled rejection and an unreported failure path

`app/screens/send-bitcoin-screen/hooks/use-onchain-fee-alert.ts:64-82`

The `await getOnChainTxFee({...})` sits in an IIFE with no `try/catch`. With Apollo's
default error policy a network/GraphQL error rejects the promise: unhandled rejection,
no `reportError` (only the missing-data branch reports), and the stale probe value from
I2 stays in effect.

**When it comes up:** a user on flaky mobile data opens the onchain details screen; the
probe query rejects at the network layer. Nothing is logged to Crashlytics, so the team
sees no signal that high-fee warnings are silently not firing in the wild, and the
comparison falls back to whatever the previous probe left behind.

**Proposal:** wrap the probe in `try/catch` and call `reportError` in the catch branch,
matching the observability standard applied elsewhere in this PR. Note also that
`if (fees)` treats a legitimately-zero probe fee as an error — unlikely on mainnet but
possible on signet/regtest; compare against `undefined`/`null` explicitly. This is a
pre-existing flaw the rewrite half-addressed, so it is non-blocking.

### Questions for the author (non-blocking)

- **Q1:** The confirmation screen shows only the fee, never the chosen speed or its
  ETA. A user who picked "Flexible (~24h)" gets no reminder of that tradeoff on the
  slide-to-confirm screen. For a feature whose entire purpose is trading fee for
  delay, surfacing the selected priority at confirmation seems worth a deliberate
  decision.

Image


- **Q2:** `PaymentDetail` now carries optional custodial-only fields (`setPayoutSpeed`,
  `payoutSpeed`, `feeQuote` — `payment-details/index.types.ts:174-196`). Cleanly done
  as an optional capability and the screen-level branching is gone, but it is a
  custodial-flavored knob on the shared type; if a second custodial-only concern
  appears, reconsider whether the type should stay flat.

---

## Test Coverage Gaps

Ordered by priority. G1 and G2 are the two to fix first.

### G1 — `4dd2541f7` ("stop serving a cached onchain fee on the confirmation screen") has no pinning test

The fix is `{ fetchPolicy: "no-cache" }` on the three onchain lazy queries in
`use-fee.ts:135-139`, but `__tests__/screens/send-bitcoin-screen/use-fee.spec.ts:19-21`
mocks the query hooks as `() => probe`, discarding the options argument. Reverting the
fix passes the entire suite. Capture the options each lazy-query hook is called with
and assert the onchain ones receive `{ fetchPolicy: "no-cache" }`.

### G2 — `send-details.spec.tsx` "leaves Next enabled when a custodial quote fails" is a Tier-2 false-confidence test

Pre-PR, Next's disabled condition was `Boolean(feeTierErrorMessage)`, which was always
undefined for custodial sends — so Next was enabled pre-PR too, and this test would
pass against the old code. The quote "fails" only incidentally (no
`onChainTxFeeBySpeed` mock exists in `app/graphql/mocks.ts`). Assert that the fee
error message text is *visible while* Next stays enabled; that combination
distinguishes the new behavior from both the pre-PR behavior and a wrongly-blocking
implementation.

### G3 — `use-fee-rail.ts` self-custodial rail argument gating is unobservable

`useSelfCustodialFeeRail` (`use-fee-rail.ts:44-60`) gates its quote inputs to
`undefined` when the rail is inactive, but there is no `use-fee-rail.spec.ts`, and in
`use-onchain-fee-tier-options.spec.ts:55-67` the `useOnchainFeeTiers` mock ignores its
arguments (unlike the custodial mock, which captures them). A regression dropping the
`isActive` gate — the SDK quoting a custodial payment's address/amount — passes all
tests. Capture the mock's args and assert the inactive rail receives
`address: undefined, amountSats: undefined`.

### G4 — `FEE_UNIT_BY_QUOTE` mapping never asserted in the custodial hook spec

`use-custodial-onchain-fee-tiers.spec.ts` asserts `feeAmount`, `etaMinutes`,
`hasError`, `hasQuote`, `isQuoting` — but never `feeUnit`. The `Usd → FeeUnit.Cents`
mapping (`use-custodial-onchain-fee-tiers.ts:146-150`) is what keeps a cents fee out
of the sats formatter — the exact bug class `3e9b4d16b` fixed one layer up. Assert
`FeeUnit.Cents` after a `Usd` quote and `FeeUnit.Sats` after `Btc` /
`UsdAsBtcDenominated` quotes.

### G5 — No-amount factory: `feeQuote === undefined` at zero amount is untested

`createNoAmountOnchainPaymentDetails` leaves `feeQuote` undefined when the settlement
amount is zero and not send-max — that undefined is what gates the custodial quote.
Add a case to `no-amount-onchain.spec.ts` asserting `feeQuote === undefined` (and
`canGetFee === false`) for a zero-amount detail.

### G6 — `7f86aeef3` (frozen placeholder tiers) unpinned

Per-tier `Object.freeze` on `EMPTY_TIERS` has no test; a selector-side mutation of a
tier would silently corrupt the shared placeholder. Assert
`Object.isFrozen(result.current.tiers[FeeTierOption.Fast])` in the custodial hook spec.

### G7 — Dead scaffolding: `mockHasSelfCustodialQuote` / `mockHasCustodialQuote` never toggled

In `use-onchain-fee-tier-options.spec.ts:23,32` both default to `true` and no test sets
them `false`, so the `feeRail.hasQuote → buildFeeTierOptions` wiring (labels without a
fee until quoted) is untested at hook level — only incidentally at screen level. Set
`mockHasCustodialQuote = false` and assert labels render without fee suffixes despite
nonzero mocked tier amounts.

### Minor / informational

- No screen-level test disables Next on a *self-custodial* quote failure (hook level
  covers `isFeeTierErrorBlocking: true`; the screen suite lacks SDK mocking for it).
- Screen-unmount mid-quote is guarded by the token/`isStale` patterns but never tested
  with an actual `unmount()`. Low value.
- The custodial hook spec replaces the generated lazy-query hooks wholesale, so the
  aliased `fast`/`medium`/`slow` GraphQL document is never executed; a wrong field
  inside the query doc would only be caught by codegen. Consistent with repo
  conventions; noted for completeness.

All other `fix` commits are pinned: `6966123a4`, `81a90ad7a`, `210b3f514`,
`c21f7fdb4`, `bd0334dd9`, `202055cb8`, `6fda41084`, `5a0992780`.

---

## Code Smells

Ordered by impact.

### S1 — The rail abstraction leaks mode one line below its own "single mode-picking point" comment

`use-onchain-fee-tier-options.ts:74-80` picks `feeRail = isSelfCustodial ?
selfCustodialRail : custodialRail` under a comment saying "everything below reads the
winner plainly" — then immediately re-checks `isSelfCustodial` and reaches past
`feeRail` into `custodialRail.isQuoting`. This is forced by `use-fee-rail.ts:65`
widening the custodial return to `FeeRail & { isQuoting: boolean }` instead of putting
the field on the interface; `defaultFeeTier` (:59) and `rebuildForTier` (:115-133)
branch on mode too, making four mode checks, not one. Put `isQuoting` on `FeeRail`
(self-custodial returns `false`) and give each rail a `defaultTier` and an
`applyTier(tier, detail)` strategy so the mode check at line 74 becomes genuinely the
only one.

### S2 — Tier-record construction is at its 3rd copy, past the repo's extraction threshold

`use-custodial-onchain-fee-tiers.ts:165-184`, `use-onchain-fee-tiers.ts:97-113`, and
`use-recommended-fee-tiers.ts:61-77` each repeat the same ~15-line "build
`Record<FeeTierOption, FeeTierInfo>` from a fast/medium/slow triple + ETA map + unit"
block. `buildZeroTiers` was already extracted for the empty case; add a
`buildTiers({ fast, medium, slow }, etaMinutes, feeUnit)` next to it in
`fee-tiers.types.ts` and call it from all three hooks.

### S3 — The keyed-quote fetch scaffolding is also at its 3rd copy

All three quote hooks repeat the same ~35-line skeleton: `useState(tiers)` →
`requestTokenRef` → `useCallback(async …)` with token bump, gate, fetch, token check,
`setTiers`, `markQuoted`/`markFailed` → `useEffect(fetchFees)`. Only the fetcher and
error classifier actually differ — a pure async function, which fits the repo's "pure
core, thin hook" standard. Extract a `useFeeTierQuote(inputsKey, zeroTiers,
fetchTriple)` primitive; each rail supplies only its fetcher and classifier.

### S4 — The quote-gate predicate is written twice per hook — the drift the PR itself warns about

`use-custodial-onchain-fee-tiers.ts:209-211` computes `inputsKey` from
`walletId && address && amount && quote`, then `fetchFees` re-evaluates the same
decision at :229; same pattern in the other two hooks. Meanwhile `onchain.ts:55-59`
adds a comment explaining `feeQuote` is "assigned beside the getFee it stands for… A
second predicate for the same decision is free to drift" — the new hooks do exactly
that. Normalize inputs once (`const inputs = … ? { walletId, address, amount, quote } :
null`), derive `inputsKey` from it, and branch on `if (!inputs) return`.

### S5 — `PaymentDetailSetPayoutSpeed`'s two optionals are always present together

`index.types.ts:175-178` declares `setPayoutSpeed?` / `payoutSpeed?` as independent
optionals, but both factories set the pair atomically and nothing constructs one
without the other; consumers already defensively check both. The same file models this
exact situation as a discriminated union twice (`PaymentDetailSetMemo` :147-155,
`PaymentDetailSetAmount` :158-167, each with a "defined together" comment). Match
them: `{ setPayoutSpeed; payoutSpeed } | { setPayoutSpeed?: undefined; payoutSpeed?: undefined }`.

### S6 — `feeQuote?: OnchainFeeQuote` overloads `undefined` with two meanings

`undefined` means both "onchain detail with no quotable amount yet" and "not an onchain
detail", and consumers can't tell which without re-checking `paymentType` — a smaller
version of the re-derivation hole this PR set out to close. Fold `feeQuote` into the
discriminated union from S5, or give the zero-amount case an explicit sentinel so
`undefined` only means "not onchain".

### S7 — The `Fast` default is hardcoded in four places

`onchain.ts:41`, `onchain.ts:399`, `use-onchain-fee-tier-options.ts:59`, and
`use-onchain-fee-alert.ts:46` (`?? PayoutSpeed.Fast`, which is fully redundant — both
factories always set `payoutSpeed`). If the schema default ever moves, four spots must
change in lockstep. Export one `DEFAULT_PAYOUT_SPEED` (or let the custodial rail own
`defaultTier` per S1) and drop the alert's fallback.

### S8 — Shared fee primitives live under a screen directory, and a second feature now imports five of them

`use-recommended-fee-tiers.ts:7-19` imports `classifySdkFeeError`, `ETA_MINUTES`,
`SdkFeeError`, `buildZeroTiers`, `FeeUnit`, `FeeTierOption`, and `useQuoteStatus` from
`send-bitcoin-screen/hooks/`. These are cross-feature fee primitives now; move
`fee-tiers.types.ts`, `use-quote-status.ts`, and the `SdkFeeError` pair to a shared
home. Also drop the `ETA_MINUTES` re-export (`use-onchain-fee-tiers.ts:20`), a pure
alias of `FEE_TIER_ETA_MINUTES` (`app/types/payment.ts:37`) that exists only to give
the same value a second name.

### S9 — Inconsistent placeholder-tier handling: frozen vs not, `EMPTY` vs `DEFAULT`

`use-custodial-onchain-fee-tiers.ts:152-157` freezes `EMPTY_TIERS` with a `forEach` and
two comments; the other two hooks hand out the same module-level mutable object
unfrozen under the name `DEFAULT_TIERS`. Freeze inside `buildZeroTiers` once and
standardize on one name.

### S10 — `FeeUnit.SatPerVbyte` makes `FeeTierInfo.feeAmount` not an amount

The enum member carries a "a fee *rate*, not an absolute fee" comment, but the field is
still `feeAmount: number`, and the refund screen reads a rate out of it. Nothing in the
type stops an absolute-fee consumer from treating a rate as sats. Low urgency (one rate
consumer exists); consider renaming to `feeValue` or splitting amount-tier vs rate-tier
variants.

### S11 — `SdkFeeError` imported twice under two names

`use-fee-rail.ts:9-13` imports both `type SdkFeeError` and `SdkFeeError as FeeError` in
one block. TypeScript handles a type and a const sharing an identifier natively; the
alias reads as if `FeeError` were a distinct, broader error type. Drop it and use
`SdkFeeError.InsufficientFunds` etc.

### S12 — Vestigial export and a third parallel tier enum

`CUSTODIAL_PAYOUT_ETA_MINUTES` is exported but used only in its own file — drop the
`export`. The PR also adds `PAYOUT_SPEED_BY_FEE_TIER`, an identity bridge between
`FeeTierOption` and the generated `PayoutSpeed`, while a third same-valued enum
`FeeTier` already exists at `app/types/payment.ts:29-35`. Not introduced here, but
deepened; consolidating `FeeTierOption`/`FeeTier` would shrink the bridge surface.

### S13 — Pre-existing, aggravated: `payment-details/onchain.ts` is 582 lines

Well past the ~350-line repo standard, and this PR added ~65 more (`setPayoutSpeed`
plumbing ×2 factories, `feeQuote` assignments, `speed:` in every mutation/query). The
copy-pasted `typeof rawAmount === "number" // FIXME: this branch is never taken?`
block now appears in all five `getFee` bodies. Not a new smell, but each edit makes the
eventual extraction wider.

---

## Suggested landing order

1. **I1** — one-line `setTiers(DEFAULT_TIERS)` plus a regression test; prevents a
   refund broadcasting at an undisplayed rate.
2. **G1** — pin the `no-cache` fix; currently zero protection on a real bug fix.
3. **I2** — key the alert probe to `payoutSpeed`.
4. **G2** — make the custodial-failure screen test able to fail.
5. M1, G3–G7, and the smells as follow-ups; S1–S4 are cheap and shrink the file before
   the next rail lands.

kngako:
This is a finding to track on the self-custodial side in a different PR/issue.

---

### I1 — Important: refund flow can submit with a stale fee rate the UI is not showing

`app/screens/unclaimed-deposits/hooks/use-recommended-fee-tiers.ts:50-55`

```ts
const fetchFees = useCallback(async () => {
  requestTokenRef.current += 1
  const token = requestTokenRef.current
  discardQuote()
  if (!sdk || !enabled) return   // tiers are never reset to DEFAULT_TIERS
```

Both sibling hooks added in this PR reset `tiers` on every fetch entry
(`use-onchain-fee-tiers.ts:79`, `use-custodial-onchain-fee-tiers.ts`), but the refund
hook discards only the quote *status* while keeping the old rates in `tiers`. The
screen reads tiers directly (`unclaimed-deposits-screen.tsx:62-63`:
`canSubmitRefund = hasAddress && feeTiersError === null && selectedFeeRate > 0`).

**When it comes up:** a user opens the refund panel during fee congestion and rates
load (say Fast = 100 sat/vB), cancels, then re-enters refund mode in the same screen
visit. A refetch starts; `hasQuote` is false so the checkbox labels correctly hide the
rates, but `selectedFeeRate` still holds the stale 100 and `feeTiersError` is null
(`discardQuote` cleared it), so pasting an address enables "Refund now" while the fresh
quote is still in flight. The refund broadcasts at a rate the UI never displayed — the
exact bug class commit `210b3f514` set out to fix, except the fix discarded the flag
and not the data.

**Proposal:** reset `setTiers(DEFAULT_TIERS)` at the top of `fetchFees` (before the
early return), mirroring the other two hooks, so a caller reading `tiers` directly can
never observe a rate from a discarded quote. Alternatively — and better long-term —
have the screen gate `canSubmitRefund` on `hasQuote` rather than on
`selectedFeeRate > 0`, so the status object is the single source of truth. Add a
regression test: quote once, toggle refund mode off/on, assert `tiers` are the zero
placeholders and the submit button is disabled until the new quote resolves.

kngako:
AI assisted fix for the fees returned by the backend. 

https://github.com/blinkbitcoin/blink/pull/764

esaugomez31:
> AI assisted fix for the fees returned by the backend.
> 
> [blinkbitcoin/blink#764](https://github.com/blinkbitcoin/blink/pull/764)

Yes, the backend is indeed not using the fee strategy that was built. I worked on part of the logic used for these queues, which was applied in this PR:
https://github.com/blinkbitcoin/blink/pull/355

We just need to validate that this logic is still valid as-is before merging this change
CC: @grimen @kngako @littledino2112

esaugomez31:
@kngako @grimen — everything actionable from the review is resolved, along with a few things a pass over the result turned up. Each fix was checked by reverting it and confirming a test fails, so none of them is a test that would have passed anyway.

### The high-fee warning reads the quote the selector already has

It no longer runs a probe of its own. The three tiers arrive together, so switching speed leaves no window where the warning judges one queue by another's rate, and the payment is now measured by its own fee rather than by a fixed 1000-sat proxy. The hook lost its query, its state and its stale-request guard in the process — it is a plain predicate now, renamed `shouldWarnAboutHighFee` since it no longer contains a React hook.

Worth flagging: the warning will now fire on different payments than before, because it compares against the real fee instead of a reference rate.

### The USD on-chain fee was read in the wrong unit

`onChainUsdTxFeeAsBtcDenominated` resolves to `OnChainUsdTxFee`, whose `amount` is a `CentAmount`; only the amount handed *to* it is sat-denominated. It was mapped to `FeeUnit.Sats`, so a USD wallet paying an address that carries its own amount saw a fee roughly two orders of magnitude too small, and the high-fee warning never fired on exactly the sends where the fee is largest.

### Failed probes are reported with the reason they carry

Apollo resolves a failed query rather than rejecting it, so the reason arrives on `error`. It is read from there and passed to `reportError`; the `catch` remains as a guard for anything that throws outside that contract, and its test says so rather than claiming to cover network failure. A rate of zero is read as a rate rather than as a failed probe.

### Both rails agree on what a failed quote leaves behind

A self-custodial failure no longer leaves the previous amount's fees in `tiers`; only the success path puts numbers back. The refund hook now drops its rates alongside the quote for the same reason, and the refund button is gated on the quote itself rather than on a rate left in memory — so it can no longer submit at a rate the labels stopped showing. It still checks the rate is one the network would accept, which the quote gate alone does not cover.

### Stale resolutions are discarded on every rail

The recommended-fee hook carries the request guard its two siblings already had, so leaving and re-entering refund mode mid-flight cannot let an older resolution clear a newer one's state.

### The fee error no longer promises a retry that does not exist

It reuses the confirmation screen's wording — *"Unable to calculate fee"* — so both screens say the same thing. The key this branch had added is gone, along with its 28 translations.

### The fee reads the same on both screens

The details screen formats through the display currency the way the confirmation screen already did. Formatting the wallet amount directly rendered a USD-wallet fee in dollars and a BTC-wallet fee in sats whatever the user's display currency was, and landed a cent off the confirmation screen for the same quote.

### Standard's ETA matches its payout queue

`MEDIUM` moved off Bria's half-hour queue in blink-deployments#10052, so the advertised `~1 hour` was under-promising. It now reads `~4h`, matching `payoutSpeeds` on staging.

### The field is titled from the design

The label reads **Transaction priority**, per the Figma screen, across all 28 translated languages.

### Structure and naming

Each rail's quoting lives in its own hook returning one shape, selected once at the top, so the consumer no longer pairs six parallel locals with repeated `isSelfCustodial` ternaries. One conversion now serves both the tier label and the warning, so they cannot disagree about a fee's unit. The refund screen's raw comparisons became named predicates and its compound `disabled` props were extracted.

### Coverage

Pinned this round: the `no-cache` fetch policy on all three on-chain queries, the details screen's gating (the selector's render condition, and the fee error being visible *while* Next stays enabled), the self-custodial rail receiving nothing to quote on a custodial payment, the fee-unit mapping per quote endpoint, `feeQuote` being absent at a zero amount, and the refund hook zeroing its rates while a re-entered panel is quoting.

Every fee-tier module is at 100% statements, branches, functions and lines. Type-check, lint and translation checks are clean.

Device-verified on staging, custodial account: with a $5.00 send at a $3.40 fee the warning fires; with $50.00 at the same fee it does not and the flow continues; both screens quote the same figure.

---

**Q1 — showing the chosen speed at confirmation.** The first version of this design, drafted last year, did show it; the current design does not. Since this is a design call rather than an implementation one, let's check with @designsats — the decision is theirs.

**Q2 — custodial-only fields on `PaymentDetail`.** Agreed. This type is special in that it behaves differently from self-custodial, so keeping it flat is the cheaper option for a single case. If a second custodial-only concern shows up, splitting it is worth considering.

---

Two things deliberately left out of this PR, both self-custodial-side and neither a defect introduced here:

- The "tiers travel with the quote" rule is spelled out in each fee hook rather than owned by the shared quote-status mechanism. Folding it in touches all three rails at once, which is its own change.
- One inconsistency this surfaced but predates the branch (#3024): the details screen warns when the fee passes 50% of the amount, while the confirmation screen warns at 2%, so a send can clear one and trip the other. Worth unifying separately.

esaugomez31:
## Fees with the backend's exponential decay strategy applied, for a $10 amount
image

CC @grimen @kngako

kngako (dismissed):

grimen (changes requested):
## Findings

**1. A fee-quote failure hides the real amount error (blocking).**

When the custodial quote fails, the rail sets the generic `LL.SendBitcoinConfirmationScreen.feeError()` message at [use-fee-rail.ts:83](https://github.com/blinkbitcoin/blink-mobile/blob/55252be69feea82bec07095addd2100517df51c5/app/screens/send-bitcoin-screen/hooks/use-fee-rail.ts#L83). `SendBitcoinDetailsExtraInfo` returns early on `errorMessage` ([line 47](https://github.com/blinkbitcoin/blink-mobile/blob/55252be69feea82bec07095addd2100517df51c5/app/screens/send-bitcoin-screen/send-bitcoin-details-extra-info.tsx#L47)) before it ever reaches `amountStatus`, so the quote failure replaces the validation message the user needs. Before this PR the custodial rail produced no error message at all, so nothing could mask it.

*Example:* A custodial BTC wallet holds 3,000 sats and the user enters 4,000 sats, which the backend rejects as below the dust threshold, so the screen reads "failed to calculate fee, please try again" instead of "insufficient balance" — with Next still enabled, since `isErrorBlocking` is false.

*Proposal:* One option could be to let `amountStatus` win over a non-blocking rail error in the extra-info box, so a validation message is never displaced by a quote failure.

**2. Ordinary input rejections are recorded as app errors.**

`reportError("use-custodial-onchain-fee-tiers", err)` at [line 257](https://github.com/blinkbitcoin/blink-mobile/blob/55252be69feea82bec07095addd2100517df51c5/app/screens/send-bitcoin-screen/hooks/use-custodial-onchain-fee-tiers.ts#L257) fires on every quote rejection, including the user-input rejections above, which are expected outcomes rather than defects.

*Example:* Each keystroke that leaves the amount below the dust threshold records a fresh error, so one hesitant amount entry can log several and dilute the real error signal.

*Proposal:* One option could be to classify the known deterministic rejections and report only the unexpected ones.

## Test Coverage Gaps

- `send-details.spec.tsx:416` asserts Next stays enabled when a custodial quote fails, but nothing asserts *which* message renders — no test covers a failed quote arriving alongside an invalid amount, which is how finding 1 slipped through.
- `use-custodial-onchain-fee-tiers.spec.ts` never mocks or asserts `reportError`, so nothing pins down which failures are supposed to reach Crashlytics.

## Code Smells

- The custodial rail pulls its user-facing string from another screen's namespace (`LL.SendBitcoinConfirmationScreen.feeError()` on the details screen). The comment explains the intent — keeping both screens' wording identical — but a shared key would express that without the details screen depending on the confirmation screen's translations.

esaugomez31:
@grimen all the feedback has been applied.

## Findings

**1. A fee-quote failure hides the real amount error (blocking).** Fixed in `af80b9b21`.

`SendBitcoinDetailsExtraInfo` still returns early on `errorMessage`; what changed is that the details screen no longer hands it one that would displace a validation message. A rail error reaches the box only when the amount is valid, or when it blocks the send outright and there is nothing to continue to anyway. Gating on `isErrorBlocking` rather than letting `amountStatus` win unconditionally keeps the self-custodial rail exactly as it was, where the SDK error *is* blocking and must still win.

*On the example:* with an over-balance amount `amountStatus.validAmount` is false, and `isNextDisabled` already carries that term, so Next was disabled throughout — the defect was the message alone, never the gate. That case also turns out to be unreachable from the UI: `AmountInputScreen` disables **Set Amount** once the amount passes `maxAmount`, which is the balance. What the input does *not* clamp is the daily withdrawal limit, so that is the reachable collision, and it is the worse one: the displaced branch renders the remaining-limit message *and* the upgrade / "increase limits" CTA, so a failed quote was hiding the way past the limit, not just the text. That is the case the new test drives.

**2. Ordinary input rejections are recorded as app errors.** Fixed in `81fb5c986`.

Worth recording what the fix had to work around, because the proposal is not reachable as written: Apollo's lazy-query `execute` never rejects — `executeQuery` resolves on its error path too (`useQuery.js:133-149`) — so the `catch` only ever saw the synthetic `new Error("Missing on-chain fee data")`. The backend's cause never reached Crashlytics at all, and since that string matches no `CONNECTIVITY_PATTERNS` entry, genuine transport failures were filed as `Defect` instead of being downgraded to `Transient`. The spec hid this by mocking rejections with `"network down"`, which *does* match.

The hook now reads `{ data, error }` and classifies from it: GraphQL errors are reported `expected: true` (breadcrumb only), transport failures go through carrying their real cause so `isConnectivityError` can do its job, and an empty response with no cause at all is still recorded as a defect. `use-fee.ts:162` was the in-repo precedent — it already treats GraphQL errors as an error state without reporting them.

*One correction:* the amount reaches this screen from a confirm-on-close modal, not a live input, so it is one quote per committed amount rather than one per keystroke. The classification is still right on its merits; the volume was just lower than the example suggests.

## Test Coverage Gaps

- **Which message renders.** `send-details.spec.tsx` does assert the message today (`getByText(feeError())`), so what was missing was the collision. Added: a failed quote arriving alongside an over-limit amount now asserts the limit message renders, that `feeError` does **not**, and that Next is disabled.
- **`reportError` in the hook spec.** Now mocked and asserted across all four outcomes: a backend rejection reported `expected: true`, a transport failure carrying its own cause, an empty response with no cause recorded as a defect, a thrown rejection recorded as a defect, and a successful quote reported not at all.

## Code Smells

- **Cross-namespace string.** Fixed in `3b8f75445`. `feeError` moved from `SendBitcoinConfirmationScreen` to `common`, and both screens read `LL.common.feeError()`. The value travelled verbatim, so all 28 locales keep the translation they already had, diacritics intact — one line moved per file, nothing retranslated.

## Fixed along the way

Four more surfaced while working through the above.

- **The high-fee modal could be skipped entirely** (`2f24fc766`). `shouldWarnAboutHighFee` returns false without a quote, and Next did not account for `isQuotingFees`, so tapping Next during the round trip walked past the warning. The hook this PR replaced probed on mount against a dummy address, so it always had a value by the time the sender could act; the new one only starts quoting when the amount is set, which opened the window. Next is now held while the quote is out.
- **The `*BySpeed` queries were being retried behind the sender's back** (`2ffca4d72`). Their single-speed counterparts are in `noRetryOperations`; the three new ones were not, so the global `RetryLink` resent them up to 5× with backoff. Together with the gate above that would have left Next dead on a spinner for the whole window.
- **A self-custodial tier change wiped the note** (`ebd98af19`). `rebuildForTier` rebuilds from the destination, which carries no memo. Pre-existing, but this PR's custodial branch spreads `...params` and keeps it, so the same action behaved differently per rail. The note is now carried across.
- **The fee-tier tests ran off the end of the mocked responses** (`1969793bc`). Three `console.warn` blocks per run, and no screen-level coverage of a quote succeeding. The quote is now answered per test, and a successful one asserts the fee lands in the tier label.

## Two left open, deliberately

Both are design tensions rather than defects, so they seemed worth raising rather than deciding unilaterally.

- `onchain-fee-alert.ts:35` — when the quote fails there is no `hasQuote`, so the high-fee modal does not show, while `isErrorBlocking: false` still lets the send through. Closing it means either blocking on a failed estimate, which contradicts finding 1's own reasoning, or restoring an independent probe, which is what this PR removed. The confirmation screen still fetches and shows the real fee before the slider.
- `use-onchain-fee-tiers.ts:67` — on the self-custodial rail the blocking error clears on the very render the amount changes, and `isQuotingFees` is custodial-only, so Next is re-enabled for the duration of the new `prepareSend`.

grimen (approved):

Issue #4005, opened 2026-07-24
ci(pipeline): deprecate Concourse — modern concurrent CI/CD from PR to app store, CI-first MVP


grimen:
## Problem

Mobile CI/CD is split across two systems with duplicated work and heavy operational drag:

- **GitHub Actions** runs PR checks (`check-code`, `test`, `audit`, `codeql`, `spelling`, `conventional`, `update_pods`, plus a dead `e2e` workflow).
- **Concourse** (`ci/pipeline.yml`, ytt/vendir-templated) runs its own `install-deps` → `test-unit` / `check-code` (duplicating GHA) → `dev-build` → `prerelease` → `prod-build`, with fastlane match signing, a `testflight-version` resource for version tracking, and GH prerelease notes.

The Concourse side is serial where it could be concurrent, opaque to contributors, and operationally expensive: a dedicated mac build worker (builds run as root; disk-full incidents forced a `cleanup-mac-build-worker` job *inside the pipeline*), Vault-managed secrets, vendored ytt templating few people can maintain, and infra churn (GKE→zrh1 migration) causing transient CI weirdness. Wall-clock times from push to signal — and from commit to store — are far from what modern tooling can deliver.

## Goal

**One modern, concurrency-optimized CI/CD system covering source (PR) to drain (App Store / Play Store) — a React Native pipeline worthy of 2027.**

- Minimal validation/test/build/release wall-clock via maximum concurrency wherever applicable: parallel check fan-out, parallel iOS/Android lanes, aggressive warm caching (yarn, Metro, Gradle, CocoaPods, DerivedData), fast-fail.
- Preserve the staged release flow we have today: dev build → **beta** (TestFlight / Play internal) → **production release**, with same-commit guarantees between stages (the `choose-commit` / `check-same-commit` semantics).
- Changelogs and release logs are **generated** (conventional-commit-driven) but **human-modifiable before release** — a draft/edit gate, not auto-publish.
- Concourse and its mac-worker infra are deprecated once parity is reached.

## Phasing

**Phase 1 — CI (MVP).** All PR validation on the new system: lint, typecheck, unit tests, audit, codegen/i18n drift checks (and later the rebuilt E2E suite from #4003). Measure the current baseline first, then target maximum fan-out with shared caches. Retire the duplicated Concourse validation jobs. CI is the MVP because it's exercised on every PR and proves the platform before release trust is moved onto it.

[...]

PR #4087, opened 2026-08-05
fix(migration): hand a locked migration with no resumable state to support

### LINKED ISSUE #4070: bug(migration): server-side migration lock + crash at phrase entry locks user out permanently across reinstalls
## Report

Support report (2026-08-04): Android user in Portugal. Revoked API keys per our guidance, started the custodial→non-custodial migration; the app crashed **before funds moved**. Since then the app crashes at recovery-phrase entry on **every reinstall** — the user is fully locked out. Server side: account ACTIVE, 24,860 sats intact, zero sends on the ledger. A crash trace was pulled and will be attached once we have the artifact.

Funds are not at risk: the balance is custodial and untouched (zero sends confirms the transfer never committed). This is a lockout loop, not a loss.

## Why the lockout survives a reinstall

The only durable client-visible state is the **server-side migration lock**:

- `app/screens/account-migration/hooks/use-migration-lock.ts` locks the app for `InProgress | Transferring | Failed`, and its doc comment states it "survives a reinstall, which the local checkpoint cannot".
- Android Auto Backup is fully disabled (`android/app/src/main/AndroidManifest.xml` — `allowBackup="false"`, targetSdk 36, no `dataExtractionRules`; confirmed in the merged manifest), so **nothing on-device survives uninstall** — including the local migration checkpoint (`migrationCheckpoint_<env>`, AsyncStorage, 48 h expiry).

The resulting loop:

1. Migration starts → `migrationStatus` becomes `InProgress` server-side.
2. App crashes mid-flow; user reinstalls → checkpoint and account index are gone, the server lock is not.
3. On login the lock is re-derived → `MigrationGate` replaces the whole app (`app/navigation/root-navigator.tsx` `useResumeCompletedMigration` / gate takeover; nav resets in `navigation-container-wrapper.tsx`). With no checkpoint, `resolveCheckpointRoute(null)` restarts from the explainer and **provisions a brand-new wallet + phrase every cycle** (each previous provisioned account's keystore mnemonic is orphaned, since the AsyncStorage account index that referenced it died with the uninstall).
4. User reaches the phrase entry/confirm step → same crash → reinstall → back to 3.

There is currently no escape hatch when the server lock is set but no local checkpoint exists — a crash mid-migration plus a reinstall is a terminal loop.

### API-key revocation is a red herring

The app stores no Blink API key — auth is a session token (`app/graphql/client.tsx`). The migration flow's only API-key use is the read-only `migrationApiKeys` query that decides whether to show the "active API keys" warning screen (`migration-gate.tsx`). Revoking keys merely skips that screen.

## Crash-site candidates (trace pending, ranked)

Note: the nav stack sits inside an `ErrorBoundary` (`app/app.tsx`), so a render throw shows the ErrorScreen — which users report as a "crash". A true native crash points at the SDK layer.

1. **Breez Spark SDK `connect()` during restore/provisioning** — `selfCustodialRestoreWallet` (`app/self-custodial/bridge/lifecycle.ts`) writes the mnemonic then connects; a native crash there is deterministic for a given wallet/backend state, matching "same crash every reinstall" and the original "crashed before funds moved". Prior art: #3838 ("Crash after manual recovery phrase restore", closed as a non-reproduced one-off) — this may be its second occurrence.
2. **Unguarded route-params destructures on the phrase screens** — `restore-phrase-screen.tsx` (`const { step, words } = useRoute().params`, no default; plus the `stepContent[step]` lookup that throws for any step outside {1,2}); same pattern on `backup-phrase-screen.tsx` and `backup-phrase-confirm-screen.tsx`.
3. **bip39 wordlist module evaluation at first render** — `use-bip39-input.ts` builds `new Set(BIP39_WORDLIST_EN)` at module scope; with Metro `inlineRequires: true` this evaluates at first render of the phrase screen, so a module-eval failure surfaces exactly as "crash on recovery phrase entry".
4. **Unhandled async rejections** — `Clipboard.getString()` un-caught (`use-restore-phrase.ts`), `throw new Error("Unknown credential error")` in an async handler (`restore-method-screen.tsx`), `backup-state.tsx` AsyncStorage read outside the try with an un-caught `load()`.

## Suggested remediation

**Immediate (support, no app change):** reset/rollback the user's server-side `migrationStatus` from admin — funds never moved, so the state is clean. This releases the lock on next launch and should restore app access.

**App fixes:**
- Give the gate an escape hatch when the server lock is set but no local checkpoint exists (e.g. offer the existing `use-rollback.ts` path, or a support-visible reset), so crash-mid-migration + reinstall is never a terminal loop.
- Guard the `useRoute().params` destructures and `stepContent[step]` lookups on the three phrase screens.
- Wrap the identified unhandled async paths.

**Follow-ups:**
- Symbolicate the pulled trace (`symbolicate-trace.sh`) and check Crashlytics for the device around the report timestamps to pin the exact frame.
- Reproduce on staging: lock a test account's migration status, fresh-install, log in, walk to phrase entry.
- If the trace lands in SDK `connect()`, reopen #3838 and file upstream with the repro.

grimen:
Closes #4070 — this is the lockout fix. The crash hardening on the phrase screens ships separately in #4088.

> Note for support/ops: merging this closes the issue, but the affected user still needs the server-side `migrationStatus` reset — see the comment on the issue.

## The problem

When a user starts the custodial → self-custodial migration, the server marks the account as mid-migration (`migrationStatus`) and the app locks itself into the migration flow until the transfer completes. The record of *where* the user got to, however, lives only on the device (the AsyncStorage checkpoint and the `migrationPendingAccounts` record).

If the app dies mid-flow and the user **reinstalls**, the local records are wiped while the server lock survives (Android Auto Backup is disabled, so nothing on-device outlives an uninstall). On the next login the gate sees "locked, nothing to resume" and restarts the flow from the explainer — **provisioning a brand-new wallet and recovery phrase on every cycle**, orphaning the previous one each time. The user in #4070 crashed at phrase entry, reinstalled, and went around this loop with no way out: the client has no mutation that can release the lock (the migration API surface is `migrationStart` / `migrationLnAddressTransfer` / `migrationCommit` only).

## The fix

Before resuming a locked migration, the gate now distinguishes two cases:

- **A wallet from an earlier run is still on the device** (pending record + wallet both present): restart exactly as before — `ensureAccount` reuses that wallet, so a crash or a 48 h checkpoint expiry *without* a reinstall behaves the way it does today. No orphans, no support ticket.
- **Nothing to resume and no wallet to reuse** (the reinstall signature): navigate once to the existing contact-support screen with a new greppable reason code `locked-without-checkpoint`, and record a non-fatal (dedup key `migration-locked-without-checkpoint`, `alwaysRecord`) so the cohort is visible in Crashlytics.

For this origin the support screen is **terminal** — header back hidden, hardware back swallowed (swipe-back was already off for the route). There is genuinely nothing behind it: `goBack()` would land on the gate's spent spinner, and the commit-path fallback would fabricate a commit screen for an account with no provisioned wallet. Every cold start replays the handover until the lock is released server-side, so the user cannot fall back into the loop.

## What this deliberately does not do

- **It cannot release the lock.** A backend cancel/rollback mutation would make this self-serve; until then support resets `migrationStatus` (safe for this cohort — the custodial balance never moved).
- The affected user from the report still needs that support-side reset.
- Support tooling should learn the new reason code `locked-without-checkpoint` (it appears verbatim in the ticket email, like the existing reason codes).

## Changes

- `app/screens/account-migration/hooks/use-reusable-pending-wallet.ts` (new): "is the pending wallet still on this device" — the same predicate `use-migration-account` already applies before reusing instead of provisioning, extracted so the gate can read it.
- `app/screens/account-migration/to-non-custodial/migration-gate.tsx`: the resume effect decides handover vs restart, and waits for the pending-wallet read the same way it already waited for the checkpoint.
- `app/screens/account-migration/to-non-custodial/contact-support-screen.tsx`: terminal back handling for the new `gate` origin; `resume` and commit origins unchanged.
- `app/types/migration.ts`: `MigrationSupportReason.LockedWithoutCheckpoint`, `MigrationSupportOrigin.Gate`.

No i18n changes: reason codes are deliberately untranslated (greppable in tickets), and the existing support-screen copy fits this cohort. The screen already tolerates the missing local wallet — the pubkey row just drops out, and the account id comes from the intact custodial session.

## Testing

All test-first (each watched fail before the fix):

- Gate: hands over only for locked + no checkpoint + no reusable wallet; still resumes when a reusable wallet exists; waits for the pending-wallet read; navigates once per mount; records the non-fatal exactly once with the dedup key.
- Support screen: `gate` origin swallows hardware back and renders no header back; `resume`/commit behavior unchanged.
- New hook: reuse matrix (record+wallet / record only / neither) and loading composition.

Full suite green on this branch (6,446 tests), `tsc`, ESLint and `check:translations` clean.

### DIFF
--- a/__tests__/screens/account-migration/hooks/use-migration-checkpoint.spec.ts
+++ b/__tests__/screens/account-migration/hooks/use-migration-checkpoint.spec.ts
@@ -155,6 +155,73 @@ describe("useMigrationCheckpoint", () => {
     await waitFor(() => expect(result.current.loading).toBe(false))
     expect(mockReportError).toHaveBeenCalledWith("Checkpoint load", expect.any(Error))
     expect(result.current.checkpoint).toBeNull()
+    /** Surfaced, not swallowed: an unreadable store must stay distinguishable from an
+     *  empty one, or the gate reads a transient failure as a wiped device. */
+    expect(result.current.hasError).toBe(true)
+  })
+
+  it("recovers through refetch after a failed load", async () => {
+    mockLoadCheckpoint.mockRejectedValueOnce(new Error("corrupt"))
+    mockLoadCheckpoint.mockResolvedValue({
+      step: MigrationCheckpoint.BackupAlerts,
+      savedAt: Date.now(),
+      accountId: "sc-account-2",
+    })
+
+    const { result } = renderHook(() => useMigrationCheckpoint())
+    await waitFor(() => expect(result.current.hasError).toBe(true))
+
+    await act(async () => {
+      await result.current.refetch()
+    })
+
+    expect(result.current.hasError).toBe(false)
+    expect(result.current.checkpoint).toBe(MigrationCheckpoint.BackupAlerts)
+    expect(result.current.hasResumableCheckpoint).toBe(true)
+  })
+
+  /** The error may only clear once the retry has SUCCEEDED: clearing it when the retry
+   *  starts would present the still-empty state as settled data for the length of the
+   *  read, and the gate would hand the user to support on it. */
+  it("keeps hasError raised while a refetch is still in flight", async () => {
+    mockLoadCheckpoint.mockRejectedValueOnce(new Error("corrupt"))
+
+    const { result } = renderHook(() => useMigrationCheckpoint())
+    await waitFor(() => expect(result.current.hasError).toBe(true))
+
+    let resolveReload: (stored: unknown) => void = () => {}
+    mockLoadCheckpoint.mockImplementation(
+      () =>
+        new Promise((resolve) => {
+          resolveReload = resolve
+        }),
+    )
+
+    let refetchDone: Promise<void> | undefined
+    act(() => {
+      refetchDone = result.current.refetch()
+    })
+    expect(result.current.hasError).toBe(true)
+
+    await act(async () => {
+      resolveReload(null)
+      await refetchDone
+    })
+    expect(result.current.hasError).toBe(false)
+  })
+
+  it("resolves instead of rejecting when the refetched load fails again", async () => {
+    mockLoadCheckpoint.mockRejectedValue(new Error("corrupt"))
+
+    const { result } = renderHook(() => useMigrationCheckpoint())
+    await waitFor(() => expect(result.current.hasError).toBe(true))
+
+    /** The gate's retry Promise.alls every refetch; a rejection there would double-report
+     *  a failure that already traveled through reportError and hasError. */
+    await act(async () => {
+      await expect(result.current.refetch()).resolves.toBeUndefined()
+    })
+    expect(result.current.hasError).toBe(true)
   })
 
   it("resolves true when the storage write succeeds", async () => {
--- a/__tests__/screens/account-migration/hooks/use-migration-support-email.spec.tsx
+++ b/__tests__/screens/account-migration/hooks/use-migration-support-email.spec.tsx
@@ -146,6 +146,35 @@ describe("useMigrationSupportEmail", () => {
     expect(Linking.openURL).toHaveBeenCalledWith(expectedMailto(expectedBody))
   })
 
+  /** The gate handover's reason is what tells support this is the locked-without-state
+   *  cohort (#4070) — only the code string in the ticket identifies it, so it has to land
+   *  verbatim in both the on-screen details and the email body. */
+  it("carries the gate handover's locked-without-checkpoint reason verbatim", async () => {
+    const { result } = renderHook(
+      () => useMigrationSupportEmail(MigrationSupportReason.LockedWithoutCheckpoint),
+      { wrapper },
+    )
+
+    expect(result.current.cardDetails[0]).toEqual({
+      label: LLSupport.reasonLabel(),
+      value: "locked-without-checkpoint",
+      isIdentifier: false,
+    })
+    expect(result.current.supportDetailsText).toContain(
+      `${LLSupport.reasonLabel()}: locked-without-checkpoint`,
+    )
+
+    await act(async () => {
+      result.current.sendSupportEmail()
+    })
+
+    expect(Linking.openURL).toHaveBeenCalledWith(
+      expect.stringContaining(
+        encodeURIComponent(`${LLSupport.reasonLabel()}: locked-without-checkpoint`),
+      ),
+    )
+  })
+
   /** The reason no longer travels with the account diagnostics, so it has to survive a
    *  device that can supply none of them. */
   it("carries the reason even when every account detail is missing", () => {
--- a/__tests__/screens/account-migration/hooks/use-pending-migration-accounts.spec.ts
+++ b/__tests__/screens/account-migration/hooks/use-pending-migration-accounts.spec.ts
@@ -170,6 +170,90 @@ describe("usePendingMigrationAccounts", () => {
       "Pending migration accounts load",
       expect.any(Error),
     )
+    /** Surfaced, not swallowed: an unreadable record must stay distinguishable from
+     *  "no pending wallet", or the gate reads a transient failure as a wiped device. */
+    expect(result.current.hasError).toBe(true)
+  })
+
+  it("recovers through refetch after a failed load", async () => {
+    mockLoadPendingProvisionedAccounts.mockRejectedValueOnce(new Error("read failed"))
+    mockLoadPendingProvisionedAccounts.mockResolvedValue({
+      "custodial-1": "sc-pending-1",
+    })
+
+    const { result } = renderHook(() => usePendingMigrationAccounts())
+    await waitFor(() => expect(result.current.hasError).toBe(true))
+
+    await act(async () => {
+      await result.current.refetch()
+    })
+
+    expect(result.current.hasError).toBe(false)
+    expect(result.current.pendingForActiveAccount).toBe("sc-pending-1")
+  })
+
+  /** The error may only clear once the retry has SUCCEEDED: clearing it when the retry
+   *  starts would present the still-empty map as settled data for the length of the
+   *  read, and the gate would hand the user to support on it. */
+  it("keeps hasError raised while a refetch is still in flight", async () => {
+    mockLoadPendingProvisionedAccounts.mockRejectedValueOnce(new Error("read failed"))
+
+    const { result } = renderHook(() => usePendingMigrationAccounts())
+    await waitFor(() => expect(result.current.hasError).toBe(true))
+
+    let resolveReload: (pending: Record<string, string>) => void = () => {}
+    mockLoadPendingProvisionedAccounts.mockImplementation(
+      () =>
+        new Promise((resolve) => {
+          resolveReload = resolve
+        }),
+    )
+
+    let refetchDone: Promise<void> | undefined
+    act(() => {
+      refetchDone = result.current.refetch()
+    })
+    expect(result.current.hasError).toBe(true)
+
+    await act(async () => {
+      resolveReload({})
+      await refetchDone
+    })
+    expect(result.current.hasError).toBe(false)
+  })
+
+  it("resolves instead of rejecting when the refetched load fails again", async () => {
+    mockLoadPendingProvisionedAccounts.mockRejectedValue(new Error("read failed"))
+
+    const { result } = renderHook(() => usePendingMigrationAccounts())
+    await waitFor(() => expect(result.current.hasError).toBe(true))
+
+    /** The gate's retry Promise.alls every refetch; a rejection there would double-report
+     *  a failure that already traveled through reportError and hasError. */
+    await act(async () => {
+      await expect(result.current.refetch()).resolves.toBeUndefined()
+    })
+    expect(result.current.hasError).toBe(true)
+  })
+
+  /** The record itself was read fine in the self-heal case, so it counts as a success
+   *  and clears a raised error rather than leaving the gate stuck on retry. */
+  it("clears hasError when a refetch lands on the self-heal path", async () => {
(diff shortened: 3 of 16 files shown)


grimen:
## On-device demo (iOS simulator)

Cold start of a **freshly installed app** (real reinstall: `simctl uninstall` + install, all local state gone) with the account locked mid-migration:

| before (main) | after (this PR) |
|---|---|
|  |  |
| The gate restarts the migration from the explainer — the loop from #4070: each cycle provisions a fresh orphaned wallet and re-walks the user into the crash. | The gate hands over once to the terminal contact-support screen with the greppable reason `locked-without-checkpoint`. No back control; the top-left tap at the end of the clip (where Back would be) does nothing. |

**What was simulated, honestly:** the server-side lock cannot be produced from the client (that is this PR's entire premise — no mutation can set or clear `migrationStatus`), so for both recordings `useMigrationLock` was TEMP-stubbed to return `isLocked: true`, and the gate's two data hooks (`useActiveApiKeys`, `useCustodialWalletBalances`) were TEMP-stubbed to their settled empty states to stand in for an authed session. Everything downstream — blocker takeover, gate decision, checkpoint/pending-wallet reads (both genuinely empty after the reinstall), navigation, support screen — is the real code. Same flow drove both sides.

**Bonus: the demo caught a real bug.** On the first take, the gate-origin support screen still showed a working native Back button that popped the user onto the gate's spent spinner — `headerLeft: () => null` does not remove native-stack's back button; `headerBackVisible: false` does. Fixed and regression-tested in ab45c5aa4.

<sub>Demos live on branch `assets/pr-4087-demo` — deleting it breaks them.</sub>

kngako:
Kimi K3 Review **(Non-blocking)**

---

# Review: `fix/migration-lock-escape-hatch-4070`

**Commits:** `c4780c44b` fix(migration): hand a locked migration with no resumable state to support (#4070) · `ab45c5aa4` fix(migration): suppress the native back button on the gate-origin support screen

**Scope:** `migration-gate.tsx`, `contact-support-screen.tsx`, new `use-reusable-pending-wallet.ts`, `types/migration.ts` (new `LockedWithoutCheckpoint` reason + `Gate` origin), three spec files, `ios/Podfile.lock`.

**Verified:** the three affected suites pass (67/67). The PR description's claims check out against the code: swipe-back is already disabled for the support route (`app/navigation/root-navigator.tsx:913`, `gestureEnabled: false`), the pubkey row drops out cleanly when no local wallet exists (`use-migration-support-email.ts:34` filters empty values), and the dedup key / once-per-mount replay behave as described.

## Findings

### 1. A failed storage read is indistinguishable from a wiped device — and triggers the support handover

`useMigrationCheckpointState` and `usePendingMigrationAccounts` both swallow load errors into empty defaults (`catch` → `loading=false`, empty state), so a transient AsyncStorage failure looks exactly like the post-reinstall signature this branch keys on. A locked user whose checkpoint and pending wallet actually exist on the device would be handed to support with `locked-without-checkpoint`, and because `hasResumedRef` claims the handover once per mount, a later successful read never reverses it — the user sits on a terminal support screen until they restart the app. This breaks the gate's own stated invariant that "a failed query never reads as an empty answer," which the lock, API-key, and balance reads all honor.

**Proposal:** propagate an `hasError` flag from the checkpoint and pending-account reads, mirroring the pattern `useMigrationLock` already uses (`hasError` traveling alongside `loading`). The gate then treats "locked + read error" like its other data errors: render the existing retry screen instead of navigating, and only hand over when both reads settled successfully *and* empty. If the current behavior is kept deliberately, a one-line comment at the handover site should acknowledge that a read failure intentionally errs toward support.

## Test Coverage Gaps

- **Load-failure path (ties to Finding 1):** no spec covers "locked + checkpoint/pending read error." It is currently untestable because no error state is exposed; once the flag exists, add a gate spec asserting the retry screen renders and no handover occurs.
- **Composed self-heal edge:** the gate spec mocks `useReusablePendingWallet` wholesale, and the hook spec mocks `usePendingMigrationAccounts` wholesale, so nothing exercises the gate's decision against the real pending-account logic — e.g. the self-heal case where the pending record points at the *active* account. One integration-style gate test with a real (storage-mocked) `usePendingMigrationAccounts` would cover the seam.
- **New reason end-to-end:** no spec renders the support screen/email with `MigrationSupportReason.LockedWithoutCheckpoint` to lock in that the reason code lands verbatim in the card details and email body. Low risk (reason codes pass through untranslated by design), but nothing guards it.

## Code Smells

- **Duplicated reuse predicate:** `use-reusable-pending-wallet.ts:20-24` is verbatim the same expression as `use-migration-account.ts:38-42` (the doc comment admits it). If `ensureAccount`'s reuse rule ever changes, the gate's resume-vs-handover decision silently diverges from what the restart would actually do — the same bug class this branch fixes. Extract a shared pure helper (e.g. `resolveReusablePendingAccount(pendingId, accounts)`) both hooks consume.
- **Unrelated `ios/Podfile.lock` churn:** checksum changes for `hermes-engine`, `ReactNativeDependencies`, and `RNPermissions` are unrelated to this fix and aren't listed under "Changes." Revert them or split into a housekeeping commit to keep the diff reviewable and bisect-clean.
- **"Gate origin is terminal" expressed twice:** `contact-support-screen.tsx` encodes it in both `handleBack`'s early return and the `headerLeft: null` branch; any future origin-behavior change must edit both. Minor, but a single derivation feeding both would prevent drift.
- **Telemetry volume (confirmed deliberate, noted for completeness):** `reportError(..., { alwaysRecord: true })` uses a process-lifetime `dedupKey` (`error-reporting.ts:70-79`), so every affected user logs one Crashlytics non-fatal per app launch until the server-side lock is reset. The PR description states this is intended visibility for the cohort; just confirm whoever triages Crashlytics expects the volume.

kngako (changes requested):
I'm happy with this change... but seems the Podfile.lock issue is in this PR as well.

kngako (dismissed):

grimen:
> I'm happy with this change... but seems the Podfile.lock issue is in this PR as well.

FTR https://github.com/blinkbitcoin/blink-mobile/issues/4005#issuecomment-5206733596

grimen:
Verified the review against the code; Finding 1 and the coverage gaps are real. Addressed in 580f48524 and 1de6efd39:

**Finding 1 — fixed (`580f48524`).** `useMigrationCheckpointState` and `usePendingMigrationAccounts` now expose `hasError` + `refetch` mirroring `useMigrationLock`; the gate joins them into the retry screen (scoped to `isMigrationLocked` — they feed no other decision, so an unlocked entry is never blocked by a storage error) and into the retry button's `Promise.all`. The handover effect guards on the errors **before** claiming `hasResumedRef`, so a retry that recovers still gets its resume-or-handover decision. One deliberate subtlety beyond the proposal: `hasError` clears only on a read that *succeeds*, never at the start of one — clearing on retry start would present the still-empty state as settled data for the length of the read and navigate to support through the retry button itself. Regression-tested (`keeps hasError raised while a refetch is still in flight`).

**Test coverage gaps — all three filled (`580f48524`).**
- Locked + read-error specs for both reads: retry screen renders, no navigation, no handover report, and the decision still lands after recovery.
- `migration-gate.integration.spec.tsx` runs the gate against the **real** `usePendingMigrationAccounts` → `useReusablePendingWallet` chain (storage mocked at the util seam): reuse-survives → resume, self-heal (record points at the active account) → handover + cleanup write, empty store → handover.
- `locked-without-checkpoint` locked in verbatim through `cardDetails`, `supportDetailsText`, the mailto body, and the rendered support screen.

**Duplicated reuse predicate — extracted (`1de6efd39`).** `resolveReusablePendingAccount` in `utils/migration-pending-account.ts`, consumed by both `ensureAccount` and `useReusablePendingWallet`.

**Not changed, deliberately:**
- *Podfile.lock* — already reverted in 970883e43 before this review posted; the underlying flip-flop is tracked in #4005 ([comment](https://github.com/blinkbitcoin/blink-mobile/issues/4005#issuecomment-5206733596)).
- *"Gate origin is terminal" expressed twice* — both branches read the single derived `isGateOrigin`, and they guard genuinely different surfaces (hardware/gesture back vs. header UI); unifying them would add a noop indirection for no drift protection beyond what the shared boolean already gives.
- *Telemetry volume* — one non-fatal per affected launch is the intended visibility for sizing the locked-out cohort until the lock reset ships; will flag it to whoever triages Crashlytics.

All 47 account-migration suites pass (549 tests); the PR-owned files hold 100% line/branch/function coverage.

kngako (approved):
Findings have been addressed

grimen:
## On-device demo (iOS simulator) — re-run on current head

Re-recorded after the review fixes (`580f48524`, `1de6efd39`), against staging on a fresh install with a custodial login — the local store holds no checkpoint and no pending wallet, the same empty state a reinstall leaves behind. Staging can't be put into a locked migration on demand, so the server-side lock was simulated by pinning the `migrationStatus` read to `InProgress`; everything downstream of that read — the gate decision, the local checkpoint/pending-wallet reads, navigation, and the support screen — is unmodified PR code.

| before (main) | after (this PR @ `1de6efd39`) |
|---|---|
|  |  |
| The locked gate finds nothing to resume and restarts the flow at the explainer on every cold start — the #4070 loop; each cycle through it provisions a fresh orphaned wallet. | The gate hands over once to the terminal contact-support screen: reason `locked-without-checkpoint`, account id from the intact custodial session, no pubkey row (no local wallet), no back control. A left-edge back-swipe does nothing, and each cold start replays the handover. |

<details>
<summary>Full-resolution stills</summary>

| before: explainer restart | after: terminal support screen |
|---|---|
|  |  |

</details>

<sub>Dev-build cold-start splash trimmed from the GIFs. Assets live on branch `assets/pr-4087-demo` — deleting it breaks them.</sub>

PR #4088, opened 2026-08-05
fix(self-custodial): harden phrase screens and async paths against crashes

### LINKED ISSUE #4070: bug(migration): server-side migration lock + crash at phrase entry locks user out permanently across reinstalls

grimen:
Part of #4070 — removes the crash candidates identified there. The migration-lock escape hatch ships separately. No behavior change on any happy path.

## Why

The lockout loop in #4070 was kept spinning by a crash on the recovery-phrase path. The nav tree sits inside the app-wide ErrorBoundary, so any render throw on these screens replaces the entire app with the fatal-error screen — which users report as "the app crashed". This PR closes every crash site the issue identified: unguarded route params, unhandled async rejections, and eager module-scope work on the phrase screen's render path.

## Route-param guards (render crashes)

The three phrase screens destructured `useRoute().params` bare. The route types promise the params, but deep links and navigation-state rehydration can deliver `undefined` or malformed values at runtime.

- **restore-phrase-screen**: invalid or missing `step` falls back to step 1 (where a restore starts anyway); `words` is accepted only if it is a real array of strings, so a malformed payload can't mis-seed the inputs. This also removes the second crash site on the same screen — `stepContent[step]` destructured `undefined` for any step outside {1, 2}.
- **backup-phrase-screen**: same step-1 fallback (it just re-shows the first six words).
- **backup-phrase-confirm-screen**: a confirm screen without its challenges is dead — nothing to type — so instead of a fallback it `replace`s itself with the first backup step, keeping the broken route out of the back stack. The single frame rendered before the redirect is safe (stable empty-challenges array; `useBackupConfirm` already guards the empty case).

Each guard reports a non-fatal once (`restore-phrase-params-missing` / `backup-phrase-params-missing` / `backup-confirm-params-missing`, all `alwaysRecord`) so bad entries stay visible instead of silently degrading. The param **types are unchanged** — no ripple through navigate call sites; the shared `isPhraseStep` guard lives next to `PhraseStep` in `stack-param-lists.ts`.

## Unhandled async rejections

- **`use-restore-phrase`**: the header Paste button's `Clipboard.getString()` was un-caught (clipboard access can reject, e.g. Android permission edge cases). Now caught → reported → toasts a new `RestoreScreen.pasteFailed` message. The key is translated into all 28 locales, with terminology matched to each locale's existing clipboard and backup-phrase strings.
- **`restore-method-screen`**: two spots. The exhaustiveness `throw` for a credential error no `CredentialError` models was reachable from an un-awaited `onPress` — an unhandled rejection if the native module ever returns a new error string. The compile-time `never` check stays; at runtime it reports (`alwaysRecord`) and toasts `restoreFailed`. And `read()` itself can reject, so the handler body is wrapped too.
- **`backup-state` provider**: the AsyncStorage read sat outside its try block and the mount-effect `load()` was fire-and-forget. This provider mounts *above* the ErrorBoundary, so a rejection here had no net at all. Read failures now report (`Backup state read`) and settle to the default state, mirroring the write path's existing convention.
- **`use-bip39-input`**: the 2048-word BIP-39 `Set` was built at module evaluation — with Metro inline requires, that is the phrase screen's first render. It is now built lazily on first use.

## Testing

All test-first (each watched fail before the fix):

- 12 param-guard cases across the three screens: `params: undefined`, out-of-range steps, malformed `words`/`challenges`, the confirm-screen redirect (`replace` to step 1), one-report-only assertions, and no-report-on-valid-params.
- 4 rejection cases: clipboard read rejects (toast + report, no unhandled rejection), credential `read()` rejects, unmodelled credential error (report instead of throw), storage read rejects (default state + report).

Full suite green on this branch (566 suites / 6,473 tests), `tsc`, ESLint and `check:translations` clean.


grimen:
## On-device proof (iOS simulator, debug build)

The restore-phrase screen entered with **no route params** — simulating deep-link / navigation-state-rehydration entry via a TEMP `initialRouteName` (no other changes):

| before (main): render error | before (main): what the user is left with | after (this PR) |
|---|---|---|
|  |  |  |
| `Cannot read property 'step' of undefined` at `restore-phrase-screen.tsx:32` — the exact crash site this PR guards. | Behind the dev overlay the app is dead on the splash: the throw killed the navigation tree before first paint. In a release build this lands in the app-wide ErrorBoundary's fatal-error screen. | Same entry, same missing params: the screen falls back to step 1 and is fully usable; the malformed entry is reported once to Crashlytics (`restore-phrase-params-missing`). |

The dev overlay in the first frame is the debug-build presentation of the throw; the code path is identical in release, minus the red box.

<sub>Screenshots live on branch `assets/pr-4088-screenshots` — deleting it breaks them.</sub>

kngako:
Kimi K3 Review

---

# Review: `fix/phrase-flow-crash-hardening-4070` @ `693a3003d`

**Scope:** 2 commits — `b61674e6a` (guard phrase-screen route params instead of crashing),
`693a3003d` (wrap unhandled async paths in restore/backup flows).

**Verified:** eslint clean on all touched files; the 6 affected specs pass (79 tests);
locales fully regenerated (spot-checked `de` — real translation, not an English fallback).

> **Note:** `yarn jest` also picks up stale spec copies under `.claude/worktrees/**`
> (228 failures from those, unrelated to this branch). Clean up the worktrees or add a
> `testPathIgnorePatterns` entry.

## Real strengths first

- Every changed failure path gets a regression test that would go red if the guard were
  reverted (`backup-phrase-confirm-screen.spec.tsx:131-196`,
  `restore-method-screen.spec.tsx:243-295`, `backup-state.spec.tsx:119-131`).
- Reporting goes through the single `reportError` boundary with dedup keys, matching the
  "observability on every fallback path" rule.
- The exhaustiveness-throw fix in `restore-method-screen.tsx:44-49` keeps the compile-time
  `never` check while removing the runtime throw — the right call.
- Comments on each guard explain the rehydration invariant; i18n mechanics (en source +
  all ~28 locales + types regenerated in the same PR) follow convention.

---

## Findings

### I1 (Important — blocking) — Restore screen: valid `step: 2` with missing/malformed `words` neither falls back nor reports

`app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx:44-55`

**Scenario:** navigation-state rehydration after an OS kill preserves `step: 2` but drops
or truncates the `words` array (non-serializable state is exactly what rehydration
mangles). The user lands on "Enter rest of your backup phrase" with empty inputs — asked
for words 7–12 of a phrase whose first 6 words were never entered — and nothing is
reported, so this variant of the #4070 telemetry hole stays invisible. The commit message
also claims the restore screen "falls back to step 1," which this path doesn't do.

**Proposal:** treat step 2 without a valid words array as invalid params — e.g.
`const hasValidParams = isPhraseStep(stepParam) && (stepParam === PhraseStep.First || isValidWords(wordsParam))`
— falling back to `PhraseStep.First` and reporting under the same
`restore-phrase-params-missing` dedup key. Update the "drops malformed words" spec, which
currently pins the stay-on-step-2 behavior, to expect the step-1 fallback and the report.

### I2 (Important — blocking) — Confirm-screen challenge validation is shape-only; semantically invalid challenges strand the user

`app/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.tsx:26-32`

**Scenario:** a crafted deep link or corrupted rehydrated state delivers
`challenges: [{ index: 99, word: "abandon" }]` — it passes `isValidChallenges`, so no
redirect fires, and the screen prompts for "word 100," which the user can never answer
correctly; the confirm button stays disabled with no in-screen escape. Duplicate indexes
would additionally collide on `key={challenge.index}` at line 116.

**Proposal:** validate semantics, not just shape — integer index within
`[0, words.length)`, non-empty word, unique indexes — and route failures through the same
report-and-redirect path. Better still, stop trusting `word` from params entirely: pass
only indexes and resolve words locally via `useWalletMnemonic()`, which also shrinks the
trusted surface to one integer array.

### I3 (Minor, pre-existing but adjacent — non-blocking) — Mnemonic words travel in navigation params, which rehydration persists

`app/navigation/stack-param-lists.ts:220`, `restore-phrase-screen.tsx` (`words` param),
`backup-phrase-confirm-screen.tsx` (`challenges[].word`)

**Scenario:** this PR makes navigation-state rehydration survivable, but rehydration is a
two-way door — the persisted nav state carrying all 12 `words` (restore step 2) or 3
phrase words (`challenges`) can be written to disk, the same secret-retention class
flagged on #3775 ("secrets must not linger in navigation params"). Not introduced here,
but this diff touches exactly this surface, so it's the right moment to note it.

**Proposal:** track as a follow-up if out of scope — carry challenge indexes only (per
I2) and re-derive step-2 words from a transient in-memory store rather than route params.
At minimum, add a "Notes" entry to the PR description stating whether navigation-state
persistence of phrase words was considered, since undocumented behavior around secrets is
itself a review finding here.

---

## Test Coverage Gaps

- **G1 — No test for semantically-invalid challenges.** The confirm-screen specs cover
  missing, empty, and shape-malformed params, but not out-of-range indexes, empty words,
  or duplicate indexes — the cases I2 is about. Add one test per case pinning the
  report-and-redirect behavior once the validation tightens.
- **G2 — The step-2-without-words test pins the wrong behavior.**
  `restore-phrase-screen.spec.tsx` ("drops malformed words") currently locks in staying
  on step 2 — a Tier-2 risk where the test passes while the dead-end state is alive. It
  should assert the step-1 fallback plus a `reportError` call, per I1.
- **G3 — Mocks deliver exactly what the guards expect.** Every new spec mocks `useRoute`
  directly, so the actual crash mechanism (real navigation-state rehydration producing
  mangled params) is never exercised. One integration test driving a real navigator with
  a rehydrated state object would prove the ErrorBoundary surface is closed rather than
  assumed closed.
- **G4 — Backup-state recovery after a failed read is untested.** `backup-state.spec.tsx`
  covers the rejecting read, but not that a subsequent `markBackupCompleted` still
  persists correctly after the provider settled to the default state — the recovery half
  of the fallback.

---

## Code Smells

- **S1 — Executable code between import blocks:** `isPhraseStep` is defined at
  `app/navigation/stack-param-lists.ts:6-9`, *above* the `lnurl-pay` and `@app/*` imports
  that resume at line 10. ESLint doesn't flag it today (no `import/first`), but any
  future import sorter — or reader — will trip on it. Move it below the import block.
- **S2 — Guard scaffolding is on its third copy:** validate-params +
  report-once-with-dedupKey effect + fallback now exists in near-identical form across
  the restore, backup, and confirm screens. Per the repo's DRY threshold this is the
  point to extract a small `useGuardedRouteParams(params, validate, { dedupKey, fallback })`
  hook before a fourth screen copies it and the variants silently diverge (as I1 shows
  they already have).
- **S3 — Producer and validator of the same shape live apart:** the `Challenge` type and
  `isValidChallenges` are declared in the confirm screen while `buildConfirmChallenges`
  lives in `../utils`. Co-locate the type and guard next to the builder so a future shape
  change can't desync the two ends of the contract.

kngako (dismissed):
The changes do as they say on the tin. 

Kimi K3 review is a little too excessive.

grimen:
Re: Kimi K3 review — addressed in 755924db2.

## Fixed

- **I1** — `restore-phrase-screen.tsx` now treats step 2 without usable words as invalid params: `hasValidParams = isPhraseStep(step) && (step === First || isValidStepTwoWords(words))`, falling back to step 1 and reporting under the existing `restore-phrase-params-missing` dedup key. "Usable" means all 12 slots present as strings with the first 6 non-blank — so truncated arrays and blank-leading rehydration artifacts also fall back, not just non-arrays. The in-code comment's "falls back to step 1" claim is now true.
- **I2** — `isValidChallenges` validates semantics: integer index in `[0, MNEMONIC_WORD_COUNT)`, non-blank word, unique indexes. Semantically invalid challenges take the existing report-and-redirect path.
- **G2** — the "drops malformed words" spec now pins the step-1 fallback plus the report; added cases for truncated and blank-leading arrays, and a valid-step-2 case pinning stay-and-seed.
- **G1** — confirm-screen specs for out-of-range, repeated, and non-integer indexes and blank words, each pinning report-and-redirect.
- **S3** — `Challenge`, both validators, and the shared `MNEMONIC_WORD_COUNT`/`WORDS_PER_STEP` constants now live in `onboarding/utils.ts` next to `buildConfirmChallenges`; the restore hook imports the constants from there. New `utils.spec.ts` covers every validator branch (and pins that `buildConfirmChallenges` output satisfies `isValidChallenges`).
- **S1** — `PhraseStep`/`isPhraseStep` moved below the import block.

Verified: 7 suites / 104 tests green, eslint and `tsc --noEmit` clean, coverage on touched files non-regressed with the new validation code fully covered (`utils.ts`, `stack-param-lists.ts`, `use-restore-phrase.ts` at 100%).

## Corrections to the findings' premises

- **I2's vector**: neither phrase screen appears in the linking config (`navigation-container-wrapper.tsx`), so a crafted deep link cannot reach them, and navigation-state persistence is not enabled (`onStateChange` is analytics-only, no `initialState`). The realistic source is an internal producer bug, which is why this landed as consistency hardening rather than a blocking exploit fix. Also, strictly, the challenge word rides in the param — a matching guess would enable confirm — but the user has no way to know it, so the dead-end conclusion stands.
- **I3's mechanism**: "persisted nav state … written to disk" assumes a persistence layer this app does not configure. The secret-retention concern is still fair as future-proofing (and params do linger in memory), so the indexes-only refactor stays on the follow-up list — but nothing writes phrase words to disk via navigation today.

## Follow-ups (not in this PR)

- **S2** — extract the shared `useGuardedRouteParams` scaffolding; this commit re-aligns the three variants, extraction is now mechanical.
- **I3** — carry challenge indexes only and resolve words via the wallet mnemonic.
- **G3** — one integration test driving a real navigator with a mangled rehydrated state.
- **G4** — backup-state recovery-write test after a failed read.
- The `.claude/worktrees/**` jest pickup is machine-local (session worktrees, not repo content); a `testPathIgnorePatterns` entry is reasonable regardless.

grimen:
Coverage follow-up in d7d537f20: every PR-owned source file now measures **100% statements / branches / functions / lines** under its own specs (`use-bip39-input.ts`, `stack-param-lists.ts`, `onboarding/utils.ts`, both phrase screens, the confirm screen, `restore-method-screen.tsx`, `use-restore-phrase.ts`, `backup-state.tsx`; 8 suites, 148 tests).

The gaps this closed were behavioral, not decorative: the bip39 suggestions memo and keyboard listeners had never executed (the old spec said as much in a comment), and the backup-state recovery path was untested — which also closes **G4** (persist-after-failed-read, persist rejection reporting, read resolving after unmount, and `markBackupCompletedFor` are all pinned now). Also newly pinned: both screens' focus/paste wiring and the deliberate swallow of a rejecting `restore()` (reported downstream by `useRestoreWallet`, so no double-report).

One annotation instead of a test: `load().catch` in `backup-state.tsx` is unreachable today because `readBackupState` catches everything internally; it is kept (and `istanbul ignore`-documented) as the net for future refactors, since the provider mounts above the app ErrorBoundary.

Remaining follow-ups from the review: S2 (shared guard hook), I3 (indexes-only challenges), G3 (real-navigator integration test).

kngako (dismissed):
Kimi blocking findings have been addressed.

grimen:
## On-device proof of the review fixes (iOS simulator, debug build)

Same method as the original demo: the screen is entered with the malformed params via a TEMP `initialRouteName` + `initialParams` in the root navigator (no other changes), simulating deep-link / rehydration-shaped entry. Before = `693a3003d` (the commit the review examined), after = `d7d537f20`, same device and same params on both sides.

### I1 — restore screen entered with `{ step: 2 }` and no `words`

| before: the silent dead end | after: step-1 fallback |
|---|---|
|  |  |
| "Enter rest of your backup phrase" over a phrase whose first 6 words were never entered — words 7-12 can never complete a valid mnemonic, the Restore button stays disabled forever, and nothing was reported. | Same entry, same params: the screen falls back to step 1, fully usable, and reports once under `restore-phrase-params-missing`. |

### I2 — confirm screen entered with `challenges: [{ index: 99, word: "abandon" }]`

| before: shape-valid, semantically dead | after: report-and-redirect |
|---|---|
|  |  |
| The old `typeof`-only validation accepts index 99 and asks for **"Enter word 100"** of a 12-word phrase — unanswerable, confirm disabled. | The semantic validation rejects it and the screen `replace`s itself with the first backup step, reported under `backup-confirm-params-missing`. The word list renders empty here only because this demo simulator has no self-custodial wallet — the redirect target, not the redirect, needs one. |

<sub>Screenshots live on branch `assets/pr-4088-screenshots` (appended to the existing demo images) — deleting it breaks them.</sub>

grimen:
## After-flows in motion (same entries as the stills, recorded end to end)

The earlier stills showed only the end states; these GIFs show the full recovery each fix produces. Same TEMP `initialRouteName` + `initialParams` entry, branch tip `d7d537f20`, cold launch included.

| I1 — `{ step: 2 }`, no `words`: lands on step 1, fully usable | I2 — `challenges: [{ index: 99, word: "abandon" }]`: report-and-redirect |
|---|---|
|  |  |
| Launch with the malformed step-2 params → splash → the screen falls back to **step 1** and typing works: "abandon", auto-advance, "ability", auto-advance — the restore can proceed normally. | Launch with the semantically invalid challenge → splash → the confirm screen rejects it and **`replace`s itself with the first backup step**; the countdown on "Save it now" shows the landing screen is live. Word list is empty only because this demo simulator has no wallet. |

<sub>GIFs live on branch `assets/pr-4088-demo` — deleting it breaks them. Stills from the earlier comment remain on `assets/pr-4088-screenshots`.</sub>

grimen:
## Input-row jank fix — spotted in the flow GIF above, fixed in `789b7dbbf`

The typing GIF exposed two pre-existing defects in the phrase input rows (they shipped with #3755; the stills never showed a filled-input state, so they went unseen until the flow was recorded):

1. **Horizontal jump on the first keystroke.** The index number was mounted only once the input had content — and the native `TextInput` paints the keystroke a frame ahead of the React commit, so the letter first rendered flush-left, then the number mounted and shoved it sideways. Repeats on every field via auto-advance.
2. **Vertical misalignment.** The number rendered in the themed `SourceSansPro-Regular` while the bare `TextInput` fell back to the system font, and `lineHeight` on an iOS `TextInput` anchors glyphs differently than on a `Text` — mismatched baselines inside box-centered rows.

Fix: the number slot is now always mounted at a fixed width and toggles opacity only (row layout can never depend on content), and both text elements share one typeface with no `lineHeight` override. Applied to `MnemonicWordInput` and the confirm screen's inline row; a new component spec keeps every PR-owned file at 100% coverage (9 suites, 153 tests).

| before (from the flow demo above) | after (`789b7dbbf`) |
|---|---|
|  |  |
| Watch word 1: "a" appears at the left edge, then jumps right as "1." pops in; the number floats above the word's baseline. | Same flow: the cursor starts in the reserved slot, the number fades in beside the text without anything moving, and "1." sits on the word's baseline. |

<sub>GIFs live on branch `assets/pr-4088-demo` — deleting it breaks them.</sub>

grimen:
## The remaining mid-word flash — root-caused and fixed in `045745784`

Sharp catch: the "typing-fix" GIF still showed a half-second artifact mid-word — text appearing horizontally centered and slightly lower, then snapping back. Frame-by-frame extraction identified it: the flashed text is never the field's content but always the **current autocomplete suggestion**. It was the `SuggestionBar` chip rendering on top of the word-2 row.

**Root cause:** `SuggestionBar` positioned itself `position: absolute` at `bottom: keyboardHeight` — but every screen that mounts it renders inside the `Screen` component's `KeyboardAvoidingView`, so its parent is *already* lifted above the keyboard. The bar's own offset stacked on top, floating the chips a full keyboard-height above the footer — directly onto the input rows. At normal typing speed the chip visibly squats there; fast input collapses the suggestions window to one frame, producing the flash. Pre-existing (not introduced by this PR), and latent on the confirm screen too.

**Fix:** the bar renders in flow above the bottom-anchored CTA with no keyboard tracking at all — the parent's keyboard avoidance is the single source of that offset. The spec now pins the no-absolute-positioning contract. Verified on device with a paused-typing flow so the chip stays visible:

| chips now render above the keyboard, rows untouched |
|---|
|  |
| Type "aba" → the "abandon" chip appears in the footer above the keyboard (previously it floated over the Word 2 row); frame-diff metrics over the input-row region confirm no transient above normal typing deltas. |

That makes three independent input-row defects surfaced by recording the flows (first-keystroke reflow, baseline mismatch, and this misplaced overlay) — all pre-dating this PR, all now fixed with pinned regression tests. Coverage on PR-owned files remains 100% (suggestion-bar.tsx included).

<sub>GIFs live on branch `assets/pr-4088-demo` — deleting it breaks them.</sub>

grimen:
**Note for reviewers:** the Paste header button on the restore screen renders off-center on iOS 26. That is not introduced by this PR — it's the Liquid Glass shared-background capsule misalignment tracked in #4072, which affects every custom header button app-wide (close buttons, settings bell, NFC, send scan, webview back, etc.). The universal fix is #3853, which opts custom header buttons out of the glass capsule via `hidesSharedBackground`, so it's intentionally out of scope here.

kngako (changes requested):
The `__tests__/components/mnemonic-word-input.spec.tsx:64:31` tests are failing.

kngako (approved):
👌🏿

grimen:
^ Re-triggers GitHub Action workflows since they were stuck still from yesterday outage.

PR #4096, opened 2026-08-06
fix(android): drop multitouch events hitting AOSP ScrollView pointerIndex crash

### LINKED ISSUE #4093: bug(android): IllegalArgumentException "invalid pointerIndex -1" in ScrollView.onTouchEvent on multitouch POINTER_UP
**Describe the bug**

Crashlytics reports a fatal Android crash:

```
Fatal Exception: java.lang.IllegalArgumentException: invalid pointerIndex -1 for MotionEvent { action=POINTER_UP(1), ..., pointerCount=2, ... }
       at android.view.MotionEvent.nativeGetAxisValue(MotionEvent.java)
       at android.view.MotionEvent.getY(MotionEvent.java:2902)
       at android.widget.ScrollView.onTouchEvent(ScrollView.java:969)
       at com.facebook.react.views.scroll.ReactScrollView.onTouchEvent(ReactScrollView.java:674)
       ...
       at com.swmansion.gesturehandler.react.RNGestureHandlerRootView.dispatchTouchEvent(RNGestureHandlerRootView.kt:40)
```

**Root cause**

This is a long-standing AOSP bug in `android.widget.ScrollView`, not app code. In the `ACTION_POINTER_UP` branch, the framework calls:

```java
mLastMotionY = (int) ev.getY(ev.findPointerIndex(mActivePointerId));
```

When a second finger lifts during a multitouch gesture that the ScrollView did not fully observe (part of the gesture stream was intercepted by react-native-gesture-handler / the JS responder system — `RNGestureHandlerRootView` is in the dispatch chain), the ScrollView's tracked `mActivePointerId` is stale, `findPointerIndex` returns `-1`, and `getY(-1)` throws. The crash event's `downTime` is ~92 s before `eventTime`, confirming a mangled/stale gesture stream rather than a normal scroll.

RN's `ReactScrollView.onTouchEvent` (line 674 in RN 0.85.3) just calls `return super.onTouchEvent(ev)` with no guard, and RN ships prebuilt Android artifacts, so patching `node_modules` source is not an option. Upgrading RN/RNGH does not help — the throw is in the Android framework and present in current Android releases.

**To Reproduce**

Not reliably reproducible on demand — requires a multitouch gesture on a scrollable screen where a second finger lifts after the gesture stream was partially intercepted (e.g. two fingers on a ScrollView while gesture-handler claims part of the stream). Seen in production via Crashlytics.

**Expected behavior**

The stale gesture is dropped silently; the app does not crash.

**Proposed fix**

Catch the framework exception at the activity boundary in `android/app/src/main/java/com/galoyapp/MainActivity.kt` — the standard production mitigation for this crash:

```kotlin
override fun dispatchTouchEvent(ev: MotionEvent): Boolean {
  return try {
    super.dispatchTouchEvent(ev)
  } catch (e: IllegalArgumentException) {
    // Android framework bug: ScrollView.onTouchEvent throws "invalid pointerIndex -1"
    // when a multitouch POINTER_UP arrives for a gesture it didn't fully observe
    // (gesture-handler interception). Safe to swallow; the gesture is already dead.
    false
  }
}
```

Optionally record the exception as a Crashlytics non-fatal inside the catch to keep visibility. The catch is deliberately narrow (`IllegalArgumentException` only, at the touch-dispatch boundary), so unrelated crashes are not masked. Dropping the event is safe — by the time it throws, the gesture is already inconsistent.

- Device: seen on touchscreen Android devices (from Crashlytics)
- OS: Android (framework bug present across current releases)
- Version: app on RN 0.85.3, react-native-gesture-handler 2.31.2

**Additional context**

Full thread dump available in Crashlytics. Main thread stack is entirely framework/RN dispatch — no app frames — consistent with the framework-level root cause.

grimen:
Closes #4093

## What

Fixes the Crashlytics fatal `IllegalArgumentException: invalid pointerIndex -1` thrown by the Android framework's `ScrollView.onTouchEvent` during multitouch (second finger lifting on a scrollable screen).

## Why this shape

The throw is an AOSP bug: on `ACTION_POINTER_UP`, `ScrollView` calls `ev.getY(ev.findPointerIndex(mActivePointerId))` with a stale pointer id when it didn't observe the full gesture stream — which happens when react-native-gesture-handler intercepts part of it (`RNGestureHandlerRootView` is in the crash's dispatch chain). It's not fixable in app or RN code: RN 0.85 ships prebuilt Android artifacts and `ReactScrollView.onTouchEvent` just calls `super`. Upgrading RN/RNGH doesn't help — the bug is in current Android releases.

The standard production mitigation is a narrow catch at the activity's touch-dispatch boundary:

- `SafeTouchDispatch` — a tiny pure-Kotlin helper (unit-testable without Robolectric) that swallows **only** `IllegalArgumentException`s whose message names a `pointerIndex` (both AOSP variants: `invalid pointerIndex -1`, `pointerIndex out of range`) and drops the event; anything else — other exceptions, other IAE messages, messageless IAEs — propagates, so genuine bugs still crash visibly. By the time the framework throws, the gesture is already inconsistent, so dropping is safe.
- The framework bug is reported as a **Crashlytics non-fatal at most once per process** — every occurrence carries the identical stack, so repeats add no signal and would flood the dashboard on an affected device.
- `MainActivity.dispatchTouchEvent` wires the helper around `super.dispatchTouchEvent`.
- `firebase-bom` + `firebase-crashlytics` added to the app module — RNFB declares its Firebase deps as `implementation`, so `FirebaseCrashlytics` wasn't on the app's compile classpath. The BoM version is read from `@react-native-firebase/app/package.json` (`sdkVersions.android.firebase`) — the same source RNFB's own gradle scripts resolve it from — so an RNFB upgrade moves both in lockstep instead of drifting against a hand-pinned literal.

## Also in this PR

- **`BitcoinPriceWidgetTest` was failing on `main`**: without an explicit `sdk`, Robolectric emulates targetSdk 36, which requires Java 21 — but the RN Gradle plugin pins the test toolchain to Java 17. Broken since the API 36 bump, unnoticed because no workflow ran these tests. Pinned to `sdk = [35]`.
- **New CI job** `android-unit-test` in `test.yml` running `./gradlew :app:testDebugUnitTest` on ubuntu-latest (the self-hosted small runners ship no Android SDK), with Gradle caching via `gradle/actions/setup-gradle`, so the Android unit tests actually gate PRs.

## Test coverage

`SafeTouchDispatchTest` (plain JUnit, 8 cases) pins every branch of the guard: passes through `true`/`false` from `superDispatch`; swallows and reports both AOSP message variants; rethrows unreported an IAE with an unrelated message, an IAE with no message, and non-IAE exceptions; and reports repeated framework throws exactly once while dropping each.

The 5-line `MainActivity` override is deliberately untested — a `ReactActivity` cannot be constructed on the JVM without faking the entire RN host, which would test the fakes rather than the wiring.

## Verification

```
$ cd android && ./gradlew :app:testDebugUnitTest
BUILD SUCCESSFUL
com.galoyapp.BitcoinPriceWidgetTest  tests="8" failures="0" errors="0"
com.galoyapp.SafeTouchDispatchTest   tests="8" failures="0" errors="0"
```

Resolved BoM confirmed unchanged after the package.json-derived version: `./gradlew :app:dependencies --configuration debugRuntimeClasspath` still resolves `firebase-bom:34.2.0`.

Manual repro of the framework bug is not reliably possible (requires a mangled multitouch stream); see the [before/after demo](https://github.com/blinkbitcoin/blink-mobile/pull/4096#issuecomment-5220544448) for a simulated repro of the crash-site behavior on both builds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,5 +45,8 @@ jobs:
           path: node_modules
           key: node-modules-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
       - run: yarn install --frozen-lockfile
-      - run: cd android && ./gradlew :app:testDebugUnitTest
+      # Caches ~/.gradle/caches and the wrapper between runs.
+      - uses: gradle/actions/setup-gradle@v4
+      - run: ./gradlew :app:testDebugUnitTest
+        working-directory: android
         timeout-minutes: 30
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,3 +29,21 @@ jobs:
         env:
           NODE_OPTIONS: --max-old-space-size=4096
         timeout-minutes: 20
+  android-unit-test:
+    name: Android unit tests
+    # GH-hosted: ships the Android SDK and JDK toolchains the Gradle build
+    # auto-detects, unlike the self-hosted small runners.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+      # Gradle needs node_modules for the RN gradle plugins and autolinking.
+      - uses: actions/cache@v4
+        with:
+          path: node_modules
+          key: node-modules-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
+      - run: yarn install --frozen-lockfile
+      - run: cd android && ./gradlew :app:testDebugUnitTest
+        timeout-minutes: 30
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -170,6 +170,11 @@ dependencies {
     androidTestImplementation('com.wix:detox:+')
     implementation 'androidx.appcompat:appcompat:1.1.0'
 
+    // Keep the BoM in sync with @react-native-firebase/app's pinned version
+    // (node_modules/@react-native-firebase/app/package.json -> sdkVersions.android.firebase)
+    implementation platform('com.google.firebase:firebase-bom:34.2.0')
+    implementation 'com.google.firebase:firebase-crashlytics'
+
     testImplementation("junit:junit:4.13.2")
     testImplementation("androidx.test:core:1.6.0")
     testImplementation("org.robolectric:robolectric:4.16")
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -170,9 +170,14 @@ dependencies {
     androidTestImplementation('com.wix:detox:+')
     implementation 'androidx.appcompat:appcompat:1.1.0'
 
-    // Keep the BoM in sync with @react-native-firebase/app's pinned version
-    // (node_modules/@react-native-firebase/app/package.json -> sdkVersions.android.firebase)
-    implementation platform('com.google.firebase:firebase-bom:34.2.0')
+    // BoM version read from the same source @react-native-firebase's own gradle
+    // scripts resolve it from, so an RNFB upgrade moves both in lockstep. The
+    // explicit crashlytics artifact is needed because RNFB declares its Firebase
+    // deps as `implementation`, keeping them off the app's compile classpath.
+    def rnfbSdkVersions = new groovy.json.JsonSlurper().parseText(
+        file("$rootDir/../node_modules/@react-native-firebase/app/package.json").text
+    ).sdkVersions
+    implementation platform("com.google.firebase:firebase-bom:${rnfbSdkVersions.android.firebase}")
     implementation 'com.google.firebase:firebase-crashlytics'
 
     testImplementation("junit:junit:4.13.2")
--- a/android/app/src/main/java/com/galoyapp/MainActivity.kt
+++ b/android/app/src/main/java/com/galoyapp/MainActivity.kt
@@ -6,6 +6,8 @@ import com.facebook.react.defaults.DefaultNewArchitectureEntryPoint.fabricEnable
 import com.facebook.react.defaults.DefaultReactActivityDelegate
 
 import android.os.Bundle
+import android.view.MotionEvent
+import com.google.firebase.crashlytics.FirebaseCrashlytics
 import com.zoontek.rnbootsplash.RNBootSplash
 
 class MainActivity : ReactActivity() {
@@ -26,4 +28,11 @@ class MainActivity : ReactActivity() {
     RNBootSplash.init(this, R.style.BootTheme) // ⬅️ initialize the splash screen
     super.onCreate(null)
   }
+
+  override fun dispatchTouchEvent(ev: MotionEvent): Boolean =
+      SafeTouchDispatch.dispatch(
+          onFrameworkBug = { FirebaseCrashlytics.getInstance().recordException(it) },
+      ) {
+        super.dispatchTouchEvent(ev)
+      }
 }
--- a/android/app/src/main/java/com/galoyapp/MainActivity.kt
+++ b/android/app/src/main/java/com/galoyapp/MainActivity.kt
@@ -29,8 +29,10 @@ class MainActivity : ReactActivity() {
     super.onCreate(null)
   }
 
+  private val safeTouchDispatch = SafeTouchDispatch()
+
   override fun dispatchTouchEvent(ev: MotionEvent): Boolean =
-      SafeTouchDispatch.dispatch(
+      safeTouchDispatch.dispatch(
           onFrameworkBug = { FirebaseCrashlytics.getInstance().recordException(it) },
       ) {
         super.dispatchTouchEvent(ev)
--- a/android/app/src/main/java/com/galoyapp/SafeTouchDispatch.kt
+++ b/android/app/src/main/java/com/galoyapp/SafeTouchDispatch.kt
@@ -1,12 +1,22 @@
 package com.galoyapp
 
-object SafeTouchDispatch {
+import java.util.concurrent.atomic.AtomicBoolean
+
+class SafeTouchDispatch {
+  private val reported = AtomicBoolean(false)
+
   /**
    * Guards against a long-standing AOSP bug: ScrollView.onTouchEvent throws
-   * IllegalArgumentException ("invalid pointerIndex -1") when a multitouch
+   * IllegalArgumentException from MotionEvent.getY ("invalid pointerIndex -1",
+   * or "pointerIndex out of range" on some releases) when a multitouch
    * POINTER_UP arrives for a gesture it didn't fully observe (e.g. after
    * react-native-gesture-handler intercepted part of the stream).
    * By then the gesture is already inconsistent, so dropping the event is safe.
+   *
+   * Only IllegalArgumentExceptions whose message names a pointerIndex are
+   * swallowed — any other exception is a real bug and propagates. The
+   * framework bug is reported at most once per process: every occurrence has
+   * the identical stack, so repeats add no signal and would flood Crashlytics.
    */
   fun dispatch(
     onFrameworkBug: (IllegalArgumentException) -> Unit,
@@ -15,7 +25,8 @@ object SafeTouchDispatch {
     try {
       superDispatch()
     } catch (e: IllegalArgumentException) {
-      onFrameworkBug(e)
+      if (e.message?.contains("pointerIndex") != true) throw e
+      if (reported.compareAndSet(false, true)) onFrameworkBug(e)
       false
     }
 }
--- a/android/app/src/main/java/com/galoyapp/SafeTouchDispatch.kt
+++ b/android/app/src/main/java/com/galoyapp/SafeTouchDispatch.kt
@@ -1,0 +1,21 @@
+package com.galoyapp
+
+object SafeTouchDispatch {
+  /**
+   * Guards against a long-standing AOSP bug: ScrollView.onTouchEvent throws
+   * IllegalArgumentException ("invalid pointerIndex -1") when a multitouch
+   * POINTER_UP arrives for a gesture it didn't fully observe (e.g. after
+   * react-native-gesture-handler intercepted part of the stream).
+   * By then the gesture is already inconsistent, so dropping the event is safe.
+   */
+  fun dispatch(
+    onFrameworkBug: (IllegalArgumentException) -> Unit,
+    superDispatch: () -> Boolean,
+  ): Boolean =
+    try {
+      superDispatch()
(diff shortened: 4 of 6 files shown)


kngako (changes requested):
Kimi K3 Findings

---

## Findings

### 1. The catch in `SafeTouchDispatch` is broader than the bug it targets

**Location:** `android/app/src/main/java/com/galoyapp/SafeTouchDispatch.kt:14`

Any `IllegalArgumentException` thrown anywhere in the view hierarchy during touch dispatch is swallowed, not just the AOSP ScrollView `pointerIndex -1` case. For example, if app or library code throws `IllegalArgumentException` from an `onTouchEvent` due to a genuine programming error (bad span index, invalid argument to a custom gesture handler), it will now be silently converted into a dropped touch plus a Crashlytics non-fatal instead of crashing visibly during QA — the user just experiences mysteriously dead touches. The commit message says "catch it narrowly," but the implementation catches by exception type only, and `IllegalArgumentException` is the most generic runtime exception in the SDK.

**Proposal:** Inside the catch, verify the exception actually matches the framework bug — e.g. `e.message?.contains("pointerIndex") == true` or the top stack frames reference `android.widget.ScrollView.onTouchEvent` — and rethrow anything that doesn't match. That keeps the workaround scoped to the unfixable framework case while letting real bugs still crash. The existing tests already make this a one-line behavioral change plus one new test case.

### 2. Unthrottled Crashlytics reporting on the main thread

**Location:** `android/app/src/main/java/com/galoyapp/MainActivity.kt:33`

Every matching multitouch event calls `recordException`, and an affected device can produce this on many gestures per session — all with an identical stack. In practice this floods the Crashlytics dashboard with thousands of duplicate non-fatals (burying other issues and counting against issue volume) and does Crashlytics I/O work inside `dispatchTouchEvent`, on the UI thread, precisely while the user is mid-gesture. For a user on a janky device who multitouches a scroll view repeatedly (the exact scenario of the bug), each pointer-up becomes a non-fatal report.

**Proposal:** Dedupe before reporting — e.g. an `AtomicBoolean` so the exception is recorded at most once per process start, since the stack trace carries no per-occurrence information anyway. Alternatively sample (record 1 in N) or report on a background executor. This keeps the diagnostic signal ("does it still happen, on which devices") without the noise.

### 3. Hand-pinned Firebase BoM duplicates RNFirebase's managed version and will drift

**Location:** `android/app/build.gradle:173-176`

`@react-native-firebase/crashlytics`'s own Gradle script already applies `platform("com.google.firebase:firebase-bom:...")` resolved from its `sdkVersions` (currently 34.2.0), and its version bumps with the JS package. The new literal `firebase-bom:34.2.0` will go stale on the next `@react-native-firebase` upgrade — worst case the app module's platform resolves a different BoM than the RNFirebase modules, and Crashlytics artifacts silently skew. The comment tells future maintainers to hand-sync two files, which is exactly the kind of instruction that gets missed.

**Proposal:** Keep the explicit `implementation 'com.google.firebase:firebase-crashlytics'` (it's genuinely needed since RNFB exposes it as `implementation`, not `api`), but resolve the BoM from the same source RNFirebase uses — `ReactNative.ext.getVersion("firebase", "bom")` — instead of hardcoding `34.2.0`. If that ext isn't reachable from the app module, a single `ext.firebaseBomVersion` in the root `build.gradle` referenced by both is still better than a magic literal in one file.

## Test Coverage Gaps

- **No integration test that `MainActivity` actually wires the guard.** `SafeTouchDispatchTest` is a pure JVM test of the helper, but nothing verifies `MainActivity.dispatchTouchEvent` routes through it and reports to Crashlytics — a future refactor of `MainActivity` could drop the override and every unit test stays green. A small Robolectric test building the activity and dispatching a synthetic event that throws would close this (the same suite infrastructure added for `BitcoinPriceWidgetTest` supports it).
- **No test for the narrowing behavior — because there is none yet.** If Finding 1 is addressed (message/stack matching), tests are needed for "IllegalArgumentException with unrelated message is rethrown" and "matching pointerIndex exception is swallowed." Currently the suite only asserts type-level swallowing, which codifies the overly broad behavior.
- **The SDK 35 pin has no guard against going stale.** `BitcoinPriceWidgetTest` now hardcodes `sdk = [35]`; when `targetSdk` next bumps, the pin silently lags and the Java 17/21 toolchain mismatch it fixed can recur unnoticed. There's no test or CI check comparing the pin to `targetSdkVersion` — at minimum the CI job added in commit `262c21017` should treat a Robolectric sandbox failure as the signal, which it now does, but a comment-only pin will still be missed. Also note `SafeTouchDispatchTest` itself doesn't need Robolectric at all, so it isn't covered by this pin discussion — fine, but worth keeping it Robolectric-free deliberately.

## Code Smells

- **Misleading "catch it narrowly" framing.** Both the commit message and the KDoc in `SafeTouchDispatch.kt` describe a narrow guard, but the code performs a catch-all of `IllegalArgumentException`. The docs should describe what the code does, or the code should match the docs (Finding 1 fixes this at the root).
- **Missing trailing newline in `MainActivity.kt`** (the diff's `\ No newline at end of file`), inconsistent with the rest of the repo and likely to produce churn in the next edit of the file.
- **CI job bypasses the repo's runner convention with only a comment as justification.** Commit `22e7dc222` recently moved tests to small self-hosted runners; `android-unit-test` goes back to `ubuntu-latest`. The reason (GH-hosted images ship the Android SDK/JDK) is legitimate and documented in the YAML comment, but if more jobs follow this pattern the self-hosted migration erodes — consider tracking "add Android SDK to self-hosted image" as follow-up work.
- **No Gradle build cache in the new CI job.** `node_modules` is cached but `~/.gradle` is not, so every PR run downloads the full Gradle/Android dependency graph within the 30-minute timeout. Using `gradle/actions/setup-gradle` (or caching `~/.gradle/caches` and `~/.gradle/wrapper`) would cut the job time substantially.
- **`cd android && ./gradlew ...` as a shell step** instead of `working-directory: android` — cosmetic, but the workflow's other steps use structured fields, and the structured form fails more readably.

kngako (approved):
**Approved because previous review is addressed:** all three findings (broad catch, unthrottled Crashlytics reporting, hand-pinned BoM) and several smells (trailing newline, Gradle cache, `working-directory`) are addressed. Verified locally: `./gradlew :app:testDebugUnitTest` — BUILD SUCCESSFUL, including the 7 new `SafeTouchDispatchTest` cases and the SDK-35-pinned `BitcoinPriceWidgetTest`.

## Minor Findings

### 1. "At most once per process" is actually "once per `MainActivity` instance" — **Minor**

**Location:** `android/app/src/main/java/com/galoyapp/SafeTouchDispatch.kt:6`, `android/app/src/main/java/com/galoyapp/MainActivity.kt:32`

The dedup `AtomicBoolean` is an instance field of `SafeTouchDispatch`, which is itself a field of `MainActivity` — but Android destroys and recreates the activity on configuration changes (rotation, dark-mode toggle, locale change), each time constructing a fresh `SafeTouchDispatch` with a reset flag. A user on an affected device who rotates the phone mid-session and multitouches a scroll view again will generate a fresh Crashlytics non-fatal per recreation, contradicting the KDoc's "reported at most once per process" claim. Impact is bounded (a handful of non-fatals per session, not a flood), so this is not blocking — but the code and its documentation currently disagree.

**Proposal:** Move `reported` into a companion object — `companion object { private val reported = AtomicBoolean(false) }` — so the flag is genuinely process-scoped regardless of the activity lifecycle; this is a two-line change matching the documented intent. Alternatively, if per-activity-instance reporting is considered acceptable, soften the KDoc to say "once per activity instance." The first option is preferred since the commit's stated goal was to prevent Crashlytics noise.

### 2. `JsonSlurper` on `node_modules` runs at every Gradle configuration, unguarded — **Minor**

**Location:** `android/app/build.gradle:178-180`

The BoM version is now parsed from `node_modules/@react-native-firebase/app/package.json` during configuration of the app module, so *any* Gradle invocation on a fresh clone or cleaned workspace (`gradlew clean`, `:app:assembleDebug`, IDE sync before `yarn install`) fails with a raw file-not-found stack trace. This comes up in practice for new contributors who open the project in Android Studio before running `yarn install`, or in CI jobs that invoke Gradle without JS deps installed. The current CI order (`yarn install` before Gradle) is unaffected, so this is purely a failure-UX concern.

**Proposal:** Guard the read with an existence check that throws an actionable error, e.g. `if (!rnfbPkg.exists()) throw new GradleException("node_modules missing: run 'yarn install' first")`, so the failure mode is a clear instruction instead of a Groovy I/O trace. Optionally parse lazily (only when the `dependencies` block is configured for a build variant that needs it) or cache the parsed value. No behavioral change for the happy path.

## Test Coverage Gaps

- **The dedup test can't catch the instance-scope bug from Finding 1.** `swallows repeated framework bugs but reports only the first` reuses a single `SafeTouchDispatch` instance, so it passes under both per-instance and per-process semantics. A test that constructs a second `SafeTouchDispatch` and asserts no re-report would lock in the intended scope — it would fail today, which is exactly the point.
- **Still no test of the `MainActivity` wiring.** The override, the Crashlytics callback, and the fact that the activity holds exactly one long-lived dispatcher remain untested; a Robolectric activity test (infrastructure already present) would cover it. Lower priority now that the helper logic itself is well covered, but the gap from the previous review is still open.
- **The narrowed matching has good coverage, with one edge unexercised:** case sensitivity. `contains("pointerIndex")` is case-sensitive, which is correct for the AOSP messages, but nothing documents or tests that a hypothetical `"PointerIndex"` variant would be rethrown. Minor — a comment near the matcher would suffice.

## Code Smells

- **KDoc overclaims the dedup scope** (see Finding 1) — "reported at most once per process" describes intent, not behavior. Docs and code diverged again right after being aligned.
- **Redundant `node_modules` restore + unconditional install remains in `test.yml`.** The `actions/cache` restore of `node_modules` followed by an unconditional `yarn install --frozen-lockfile` makes the cache little more than a warm-up; either skip the install on cache hit (with a `key`/`restore-keys` miss fallback) or drop the cache and rely on `setup-node`'s yarn cache. Pre-existing from the earlier commit, not new.
- **`gradle/actions/setup-gradle` placed after `yarn install`** — ordering works, but convention (and the action's docs) put it before build-affecting steps so its cache reporting covers the whole job; purely cosmetic.
- **Configuration-time file read not declared for configuration cache.** `file(...).text` in the `dependencies` block isn't registered as a build input, so if the project ever enables Gradle's configuration cache this becomes a cache-invalidation hazard. Not actionable today (RN projects rarely enable it), worth a one-line comment.

PR #4097, opened 2026-08-07
fix(developer): gate developer screen behind __DEV__

kngako:
With new tests on the navigation.

Fixes https://github.com/blinkbitcoin/blink-wip/issues/1055


grimen (changes requested):
## Findings

**1. The second secret-tap trigger still navigates to `developerScreen` ungated, so the PR closes only one of two entry points.**
`get-started-screen.tsx` lines 37–42 contain the same counter effect as `version.tsx`, but its `navigation.navigate("developerScreen")` was not wrapped in `__DEV__`. In a release build the route is no longer registered (per the new gate in `root-navigator.tsx`), so React Navigation silently swallows the action. This is also the path the legacy Appium specs use against the release APK (`e2e/01-phone-flow-and-resets.e2e.spec.ts:32`, `e2e/helpers.ts`), so anyone reviving those scripts hits an immediate timeout.

*Example:* on a production build, triple-tapping the logo on the Get Started screen dispatches a navigation to a nonexistent route — a silent no-op that looks like the gate works, while the call site the PR meant to remove is still live.

*Proposal:* one option is to apply the same `__DEV__` gate here; better, extract the shared trigger (see Code Smells) so the gate lives in one place.

**2. `DeveloperScreen` still ships in and evaluates at startup of the release bundle — the gate removes the route, not the code.**
`root-navigator.tsx` keeps the static top-level import, and Metro does not tree-shake static imports. `developer-screen.tsx`'s module body (module-scope `gql` `debugScreen` query, `GALOY_INSTANCES` with instance URLs, token save/replace UI, `DevSettings`) is bundled and executes in production.

*Proposal:* one option is a `__DEV__`-guarded inline `require` resolved inside the `__DEV__ &&` branch, which lets the minifier drop the module once `__DEV__` inlines to `false`; if shipping the code is considered acceptable, a comment noting the gate is registration-only would help.

## Test Coverage Gaps

- The new specs in `version.spec.tsx` pin only the `version.tsx` call-site guard. Reverting the `{__DEV__ && ...}` block in `root-navigator.tsx` leaves the whole suite green (`navigate` is a jest mock, so the assertion passes whether or not the target route exists) — the mechanism that actually gates release builds is unasserted. A test asserting `developerScreen` is absent from the rendered navigator when `__DEV__` is false would protect the fix that matters.
- The second trigger in `get-started-screen.tsx` has no coverage at all, which is where finding 1 hides.

## Code Smells

- The secret-menu counter (`useState` + `useEffect` + threshold + navigate) is copy-pasted in `get-started-screen.tsx` and `version.tsx`, and this PR made the copies diverge — gating one and missing the other. `app/hooks/` already hosts small guard hooks (`use-in-flight-guard`, `use-screen-security`); a single `useSecretMenuTrigger()` consumed by both `Pressable`s would put the `__DEV__` gate, tap threshold, and route name in exactly one place.

grimen (changes requested):
## Findings

**1. The no-static-import regression guard is porous (blocking).**
The spec's `/^import\s.*developer-screen.*$/m` check misses prettier-wrapped multi-line imports, `export ... from` re-exports, and an added unconditional module-scope `require()`, so the exact regression this spec exists to catch would pass every assertion.

*Example:* A refactor re-adds a multi-line `import {` … `} from "../screens/developer-screen"` — the specifier line starts with `}` so the regex never matches — and the developer screen ships in release bundles again while the spec stays green.

*Proposal:* One option could be a specifier-level check, e.g. asserting `developer-screen` occurs exactly once in `root-navigator.tsx`, inside the `__DEV__` ternary.

**2. Release-build taps still update state.**
The tap handler calls `setTapCount` unconditionally, so in release builds every tap on the version text or get-started logo re-renders the host screen for an intentionally inert feature, and `tapCount` grows unbounded because the only reset lives inside the dead `__DEV__` effect branch.

*Example:* A release-build user idly tapping the get-started logo forces a full screen re-render on every press with zero possible effect.

*Proposal:* One option could be a ref-based counter with an early `if (!__DEV__) return` and the navigate call inline in the handler — it preserves all four behaviors the specs pin without re-renders.

**3. Positive assertions pin exact source text.**
`toContain("{DeveloperScreen && (")` and the `__DEV__ ? require(…) : null` regex fail on behavior-preserving rewrites such as switching the ternary to `&&` or renaming the guard const, which trains future editors to mechanically loosen the one spec guarding the PR's core property.

*Example:* A cleanup PR rewrites the gate to the equivalent `const DeveloperScreen = __DEV__ && require(…)` form; both assertions fail despite identical release behavior, and the author relaxes whichever regex broke.

*Proposal:* Coarser intent-level checks at the altitude of the `view-backup-security-checks-screen.spec.tsx` precedent (identifier and route-name presence) could pin the property without the exact text.

**4. `developerScreen` stays unconditional in `RootStackParamList`.**
The PR makes the route dev-only, but the param list still declares it, so a future ungated `navigate("developerScreen")` type-checks cleanly, works in all dev testing, then silently no-ops in release builds.

*Example:* A settings row adds the navigate call without a `__DEV__` guard; in release, react-navigation drops the unhandled action and the tap does nothing.

*Proposal:* The entry can't be removed (the hook's typing needs it and TypeScript can't express build-conditional routes), so a warning comment on the param entry could be the remedy.

grimen (approved):

PR #4100, opened 2026-08-07
fix(migration): self-help copy when a restart can resolve the failure

### LINKED ISSUE #4098: bug(migration): Improve migration error screen self-help copy
## Summary
Ralphie reported that the migration error screen is too generic when migration hits a transient/start-refused state. Restarting the app resolved the problem, but the current copy pushed him to contact support for something he could self-resolve.

## Current behavior
- Migration error screen says a generic "something went wrong" message and makes "contact support" feel like the primary action.
- User opened a support ticket even though restarting the app fixed the issue.

## Expected behavior
- Explain that migration can take a few minutes.
- Offer a self-help action: close and reopen the app, then try again.
- Mention that the user will be notified when migration completes, if that notification path exists for this state.
- Keep contact support as a fallback, not the primary CTA.

## Context
- Source: Slack thread BD-00055 / customer UX feedback from Ralphie.
- Requested by @ihaveadifferentname on 2026-08-07.
- Please include @designsats in follow-up. Kgothatso was also requested for visibility from Slack (<@U0BJ058EX1D>), but I do not have a GitHub handle for him.

## Notes
- Treat this as UX copy/flow improvement for the migration error screen.
- This ticket intentionally follows the customer feedback framing, not the backend interpretation that `start-refused` means migration had already completed.

grimen:
Closes #4098

cc @designsats

## What happened

A user's migration hit `start-refused`; the support screen opened with the generic "Something went wrong … contact support" copy and a primary **Contact us** button, so he filed a ticket (BD-00055) — then restarted the app and the migration simply finished.

That is the sanctioned remedy for this state: the start rejection latches in memory only (`useEnsureMigrationStarted.retry()` deliberately no-ops once rejected, because the backend maps distinct causes onto one `MIGRATION_STATE_CONFLICT` code), and on the next cold start the gate resumes from the checkpoint — or `useResumeCompletedMigration` finishes a migration that had already completed server-side.

## What changed

`start-refused` now gets a self-help variant of the contact-support screen:

- Hero leads with **"Try restarting the app"** and instructions: funds are safe, migration can take a few minutes, close the app completely and reopen it, the migration picks up where it left off.
- **Contact support demotes to a secondary fallback** (no primary button — restarting cannot be a button, so the instruction is the primary affordance). It still opens the same pre-filled support email, and the address-copy control stays.
- The diagnostics card is unchanged and still shows `Reason: start-refused` verbatim, so a screenshot alone can still open a ticket if the restart does not help.

Every other reason (`locked-without-checkpoint`, `transfer-failed`, account-missing, …) keeps the support-first copy — those genuinely need support and cannot be restarted away. The self-help set is a one-line `ReadonlySet` to extend if more restart-resolvable reasons appear.

## Notes

- The issue asked to mention "you'll be notified when migration completes" *if that path exists* — it does not (no push/local notification fires on migration completion), so the copy says the truthful equivalent instead: the migration picks up where it left off when the app reopens. A completion notification would be a separate feature.
- New i18n keys land in `en/index.ts` only (locales fall back to English), per the usual convention; `i18n-types.ts` and `raw-i18n/source/en.json` are regenerated.
- No interaction with the terminal gate-origin behavior from #4087: `start-refused` only arrives with `origin: commit`.

## Tests

Three new cases in `contact-support-screen.spec.tsx`: self-help copy + intact diagnostics for `start-refused`, the demoted contact action still reaching `sendSupportEmail`, and a terminal reason keeping the support-first copy. `balances-overview-screen.spec.tsx` run as regression (start-refused → support handover unchanged).


kngako (changes requested):
## Findings

### I1 — "The migration will pick up where it left off" is only true when the server lock exists

**Anchors:** `app/i18n/en/index.ts` (selfHelp body), `migration-gate.tsx:199-204`, `use-ensure-migration-started.ts:36-40`

A refused start leaves the migration status at `NOT_STARTED` — no lock is armed — and the gate's resume path is explicitly gated on `isMigrationLocked`. So a user whose start was refused *before* any lock existed follows the copy, restarts, and lands on the ordinary migration entry/intro: nothing resumes automatically and they must re-tap through the flow themselves. The copy promise only holds for the subset of refusals where the account was already locked server-side (e.g. a state conflict from a prior armed lock).

**Proposal:** Reword the body so it doesn't promise automatic resumption — e.g. "open it again and return to the migration — it will continue from where it left off" only if the gate actually does that, otherwise "open it again and start the migration again from your account screen." The heavier alternative — letting the gate resume from a local checkpoint even without a lock — changes gate semantics for every entry path and is not worth it for a copy fix. Whichever wording is chosen, the claim should also be corrected in the SELF_HELP_REASONS comment (`contact-support-screen.tsx:26-27`) and the test doc block, which both assert the gate resumes on relaunch.

### 2 (blocking) — `start-refused` conflates transient and terminal refusals, so some users get restart advice that can never work

**Anchors:** `use-ensure-migration-started.ts:39-41,94-99,109`, `balances-overview-screen.tsx:119-121`, `contact-support-screen.tsx:28-30`

`isRejected` is set for *any* non-network failure, and the hook's own doc notes the backend maps the permanent dollar-balance rejection onto the same `MIGRATION_STATE_CONFLICT` code as a transient conflict — the client deliberately cannot tell them apart. A user with a terminal refusal (a dollar balance that arrived mid-flow, a cohort exclusion) sees "Close the app completely, then open it again," restarts, is deterministically refused again, and bounces back to the same screen. The copy's escape hatch ("Still seeing this screen after restarting? Contact support") bounds the damage to one wasted restart cycle, so this is acceptable as ticket deflection — but it should be a conscious, documented trade-off.

**Proposal:** Reword the `SELF_HELP_REASONS` comment, which currently asserts these are "[r]easons a user can clear themselves by restarting," to state that only a subset of `start-refused` occurrences are restart-resolvable and the variant accepts one futile restart for the rest. Longer term, ask the backend for a distinct error code for permanent refusals so terminal cases can keep the support-first copy. No code change is needed for this PR beyond the comment, since the copy already handles the still-stuck path.

### 3 (minor) — New self-help strings render in English in every non-`en` locale until translations land

**Anchors:** `app/i18n/raw-i18n/source/en.json`, `app/i18n/es/index.ts:5` (representative of all locales)

Only the English source was updated; every other locale merges the raw translations over `en`, so the new `selfHelp` block falls back to English. A Spanish- or French-speaking user who hits the refused-start state sees "Try restarting the app / Close the app completely…" in English inside an otherwise localized screen — noticeable precisely because this is the screen a confused user is reading most carefully.

**Proposal:** Stack A PR with the relevant translations from this PR.

---

## Test Coverage Gaps

- **Variant membership is pinned for only two of nine reasons.** The self-help branch is tested for `StartRefused` and the support-first branch for `LockedWithoutCheckpoint` plus the `unknown` fallback. The other six reasons (`PreviewUnavailable`, `BalancesUnavailable`, `SelfCustodialAccountMissing`, `SelfCustodialAccountNotOnDevice`, `TransferFailed`, `LnAddressTransferFailed`) have no assertion on which variant they get — a future edit adding one of them to `SELF_HELP_REASONS` would turn every test green. A table-driven test enumerating all `MigrationSupportReason` values with their expected variant would pin the set's membership, which is the actual contract of this feature.
- **The copy-details control is asserted present but never pressed in the self-help variant.** `contact-support-screen.spec.tsx` checks `getByText(LLSupport.copy())` under `start-refused`, but the `copyDetails` → `copyToClipboard(supportDetailsText)` path is only exercised in the support-first variant. Since the variant switch restructured the JSX around that control, a regression breaking its wiring only in self-help mode would not be caught.
- **`use-migration-diagnostics` has no dedicated spec.** Its behavior (dropping empty values, ordering identity fields) is only covered indirectly through `use-migration-support-email.spec.tsx`. This branch changed its public type (removed `isIdentifier`); a direct spec would pin the remaining contract so the next type change has something to break against.


---

## Code Smells

- **`SELF_HELP_REASONS` lives far from the reason taxonomy.** The set is defined in `contact-support-screen.tsx:28-30` while the reasons and their semantics are documented in `app/types/migration.ts:18-42`. Someone adding a new restart-resolvable reason to the enum gets no colocation or compile-time signal that the screen needs updating. Moving the set — or an `isRestartResolvable(reason)` helper — next to `MigrationSupportReason` keeps the knowledge in one place.
- **Single-member `ReadonlySet`.** `new Set([MigrationSupportReason.StartRefused])` buys nothing over `reason === MigrationSupportReason.StartRefused` until a second member exists. Mild, and defensible if growth is expected soon — but a plain comparison plus the same comment would read simpler today.
- **Two `as IconNamesType` casts where one annotation would do** (`contact-support-screen.tsx:72,78`). Typing the `copy` object once (`{ icon: IconNamesType; title: string; body: string; cta: string }`) removes both casts and restores excess-property checking on the literal.
- **Specs key the hero variant on `icon-refresh` / `icon-headset` testIDs**, coupling the tests to `GaloyIcon`'s internal `icon-<name>` testID convention rather than to user-visible output. The author clearly knew the text queries collide (the test comment explains the workaround), so this is the pragmatic choice — but a `within(hero)`-style query scoped to the `IconHero` container would assert the same thing without reaching into the icon component.

kngako (dismissed):
👌🏿

ihaveadifferentname (approved):

kngako (approved):
👌🏿

esaugomez31 (approved):

PR #4101, opened 2026-08-07
fix(migration): preserve currency, language and theme after migration

### LINKED ISSUE #4099: bug(migration): Preserve currency setting after migration
## Report

During the migration flow, the app appears to reset the user currency setting back to USD.

Customer/user report:

> It is also unfortunate that the currency setting is reset after the migration. When someone had EUR set, it defaults back to USD

## Expected behavior

If a user had EUR selected before migration, the app should preserve EUR after migration.

## Actual behavior

After migration, the currency setting defaults back to USD.

## Notes

- Related migration-flow feedback: https://github.com/blinkbitcoin/blink-mobile/issues/4098
- Requested by @openOMS from Slack #2-tech-bot on 2026-08-07.

grimen:
Fixes #4099

## Problem

Migrating a custodial account to self-custodial silently reset the user's display currency to USD (reported: EUR → USD). Language and theme reset the same way, to device default and system theme.

Settings are stored per account, in two different worlds: the custodial account keeps display currency and language **on the server** (`me.defaultAccount.displayCurrency`, `me.language`) and its theme under the custodial slot of the local theme map, while a self-custodial account reads all three from **per-account maps in persistent state**, keyed by account UUID. Migration provisions a brand-new `Crypto.randomUUID()` with no entry in any map, so after the account swap every read falls back to its hardcoded default. The server-side EUR still exists, but the custodial session that could read it is discarded on completion.

## Solution

Seed the three per-account maps at **provision time** — inside `ensureAccount` in `use-migration-account.ts`, right after the new account id is resolved — while the custodial session and its Apollo cache are guaranteed live. The completion path (and its resume variant after an app restart) runs after logout, when the server values are unreachable, so seeding there would be fragile; seeding at provision persists the values immediately and makes resume a non-issue.

- New pure `seedMigratedAccountSettings` (`app/store/persistent-state/migrated-account-settings.ts`) composes the three writes; unknown values (cold cache/offline) are skipped so the existing fallbacks still apply.
- New `useSeedMigratedAccountSettings` hook captures the custodial values via the existing cache-first `displayCurrency`/`language` queries, warming them while the user reads the migration explainer screens.
- The existing active-account setters delegate to new id-parameterized `...ForAccount` variants, since the seed targets a not-yet-active account (the active-account setters deliberately no-op while custodial is active).
- Seeding covers both the fresh-provision and the reused-pending-wallet paths, and is overwrite-idempotent.

Known limitation: users who already migrated cannot be backfilled — their custodial session and cache are gone; they re-pick their settings once in Settings.

## Testing

- New `migrated-account-settings.spec.ts` incl. the regression proof: seed with EUR/es/dark, activate the migrated id, read back EUR/es/dark; null snapshot leaves state identical and fallbacks intact.
- New `use-seed-migrated-account-settings.spec.ts`: seeding from query data, cold-cache no-op, cache-first/skip wiring.
- Extended `use-migration-account.spec.ts`: seeds on fresh provision and on pending-wallet reuse; does not seed on checkpoint short-circuit or guard-blocked runs.
- Extended the three persistent-state specs for the `...ForAccount` variants.
- 69 targeted tests green; `tsc` and `eslint` clean on all touched files.
- Manual iOS simulator run of the full migration flow still pending.

### DIFF
--- a/__tests__/screens/account-migration/hooks/use-complete-migration.spec.ts
+++ b/__tests__/screens/account-migration/hooks/use-complete-migration.spec.ts
@@ -44,6 +44,23 @@ jest.mock("@app/screens/account-migration/hooks/use-pending-migration-accounts",
   }),
 }))
 
+const mockReportError = jest.fn()
+
+jest.mock("@app/utils/error-logging", () => ({
+  reportError: (...args: unknown[]) => mockReportError(...args),
+}))
+
+const mockSeedMigratedSettings = jest.fn()
+
+jest.mock(
+  "@app/screens/account-migration/hooks/use-seed-migrated-account-settings",
+  () => ({
+    useSeedMigratedAccountSettings: () => ({
+      seedMigratedSettings: mockSeedMigratedSettings,
+    }),
+  }),
+)
+
 import { useCompleteMigration } from "@app/screens/account-migration/hooks/use-complete-migration"
 
 const complete = async (): Promise<boolean | undefined> => {
@@ -63,6 +80,7 @@ describe("useCompleteMigration", () => {
     mockRegistryLoading = false
     mockCheckpoint = "backupAlerts"
     mockDiscardCustodialSession.mockResolvedValue(undefined)
+    mockSeedMigratedSettings.mockResolvedValue(undefined)
   })
 
   it("does nothing and returns false when no account has been provisioned", async () => {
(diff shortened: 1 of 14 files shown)


kngako (dismissed):
I'm cool with everything else but this test coverage gap review point might be worth implementing before merging.

## Test Coverage Gaps

- **Theme/language overwrite on re-seed untested:** initial theme seeding is covered (`migrated-account-settings.spec.ts:32`), and the re-seed test ("overwrites a previous seed for the same account") covers the currency map only — nothing pre-seeds a theme or language entry for the target account and asserts a re-seed overwrites it.
  - **Example:** a user starts migration (custodial theme "light" and language "en" are seeded onto the pending account), abandons it, then changes theme to "dark" and language to "es" before retrying; the retry reuses the pending wallet and re-seeds. If a refactor broke the overwrite semantics (e.g. a "don't clobber existing entries" guard in `withThemePreferenceForAccount`), the stale values would ship and the suite would still pass.
  - **Proposal:** extend the existing overwrite test to pre-seed all three maps with stale values (`GBP` / `en` / `light`) and assert the re-seed replaces each (`EUR` / `es` / `dark`). One test covers all three since `seedMigratedAccountSettings` takes the full snapshot in a single call; function-level is the right place, no hook-level test needed.

ihaveadifferentname (approved):

kngako (approved):
👌🏿

PR #4103, opened 2026-08-07
fix(migration): wait for confirmed Spark receive before redirecting after migration

### LINKED ISSUE #4102: bug(migration): wait for confirmed Spark receive before redirecting after migration
Requested by: @ihaveadifferentname from Slack #2-tech-bot on 2026-08-07.

Tagging @grimen per request.

## Problem

Multiple users have reportedly landed in a newly migrated Spark account showing a zero balance, while investigation later finds that their funds are still stuck in transit.

The current migration/transfer flow appears to redirect the user after the send is confirmed. That can put the user into the new Spark account before the receive side has completed, creating a zero-balance state that looks like funds are missing.

## Expected behavior

Do not redirect the user into the new Spark account on confirmed send alone. Redirect only after the receive into the new Spark account is confirmed.

## Impact

Users may see a zero balance immediately after migration even though funds are still moving. This creates a high-trust support issue because it looks like funds disappeared.

## Notes

- First issue to resolve: gate the redirect on confirmed receive into the new Spark account.
- Reports mention funds being "stuck in transit" after users land in the zero-balance account.

grimen:
Fixes #4102

## Problem

The migration flow swapped the user into the new Spark account as soon as the backend reported `MigrationStatus === COMPLETED`. That status is the **sender's** word: the lightning payment into the new Spark wallet can still be in transit at that moment (Spark holds an incoming payment for an offline wallet and only claims it on the next sync — and the migration wallet's SDK is disconnected right then). Users landed in a zero-balance account that looked like lost funds.

There were two ungated swap paths: the transferring-funds screen, and the launch-time resume hook (`useResumeCompletedMigration`) that finishes a migration completed server-side but never swapped locally.

## What this does

- **New shared gate** `useMigrationReceiveConfirmation`, composed by **both** swap paths: after the server reports COMPLETED, it polls the provisioned wallet's synced balance until the receive is confirmed. The check runs through the existing serialized migration-SDK helper (`withMigrationSdk`), so it can never race the invoice-building connection, and the forced sync (`ensureSynced: true`) is itself what claims the pending payment — the first check typically claims and confirms in one shot. Any positive balance is the receive: the wallet is provisioned by this flow and never used before the swap.
- **Zero-receive migrations don't wait**: the commit-point preview's `receiveSats` is now persisted into the migration checkpoint (the only moment it is knowable — after the drain the preview reads an already-emptied balance). An expected receive of 0 opens the gate at once. Checkpoints saved by app versions before the field existed wait like funded ones; the delayed notice below is their escape hatch.
- **Delayed notice**: after a remote-configured window (`migrationReceiveDelayedNoticeMs`, default 1 minute — per product on the Slack thread, a minute after the confirmed send is more than enough) the transferring screen switches to "taking longer than usual — keep the app open" and offers a secondary Contact Support button (reason code `receive-delayed`). Polling never stops (while the escape hatch below is off), and the swap still fires by itself the moment the receive lands — including while the support screen sits on top, since the transfer screen stays mounted beneath it.
- **Delayed-redirect escape hatch**: a second remote-config flag, `migrationDelayedRedirectEnabled` (default **false**), releases the swap once the notice window elapses instead of holding it until the receive confirms — the pre-gate behavior, reachable without an app release should requirements change. While off (the product decision), nothing changes.
- The resume path simply doesn't swap while unconfirmed: the user stays on the working custodial session (strictly better than an empty new wallet) and the gate keeps checking, this session and the next.
- The migration-status poll now stops on the server's terminal phase rather than on the swap condition, so it doesn't hammer the server's resume routine during the receive wait.

## Notes

- `NoMnemonic` from the check passes through as confirmed on purpose: `completeMigration()` already reads the same absence and routes to support (`self-custodial-account-missing` / `not-on-device`); the gate must not invent a second handover path.
- Old checkpoints without `expectedReceiveSats` stay valid — the field is optional and validated like `accountId`, carried across step saves under the same owner rule.
- The `transferDelayed` strings are translated in all 28 locale files (AI-translated, register matched to each locale's existing migration strings), satisfying the locale-parity check on CI.
- The gate is fully inert while its caller skips it — a late failure that re-skips the gate cannot keep surfacing a stale delayed notice.
- Firebase: the `migrationReceiveDelayedNoticeMs` and `migrationDelayedRedirectEnabled` parameters can be created in the console later; the in-app defaults (1 min, off) cover until then.

## Tests

296 tests across the 10 affected suites, including a 23-case spec for the gate hook (polling cadence, no overlapping checks, the mount-skipped→armed transition, delayed-notice timing and withdrawal, escape-hatch release, legacy-checkpoint behavior, re-skip inertness, unmount races) and a new serialization case proving the balance check queues behind a transfer request on the same storage directory. The changed source files sit at 100% statement/branch coverage. `tsc`, `eslint`, and `check:translations` clean.

Before/after GIFs are in the PR comments (assets on branch `assets/pr-4103-demo`).

### DIFF
--- a/__tests__/config/feature-flags-default-blocks.spec.ts
+++ b/__tests__/config/feature-flags-default-blocks.spec.ts
@@ -108,4 +108,14 @@ describe("defaultRemoteConfig: compliance country lists", () => {
       "MM",
     ])
   })
+
+  /** The registered default is what runs before the first fetch lands and on every fetch
+   *  that fails; releasing the swap on the server's COMPLETED alone is the #4102 report. */
+  it("migrationDelayedRedirectEnabled defaults to holding the swap", () => {
+    expect(defaultRemoteConfig.migrationDelayedRedirectEnabled).toBe(false)
+  })
+
+  it("migrationReceiveDelayedNoticeMs defaults to the one minute product asked for", () => {
+    expect(defaultRemoteConfig.migrationReceiveDelayedNoticeMs).toBe(60_000)
+  })
 })
--- a/__tests__/screens/account-migration/hooks/use-complete-migration.spec.ts
+++ b/__tests__/screens/account-migration/hooks/use-complete-migration.spec.ts
@@ -6,11 +6,13 @@ const mockDiscardCustodialSession = jest.fn()
 
 let mockAccountId: string | undefined
 let mockCheckpoint: string | null
+let mockExpectedReceiveSats: number | null
 
 jest.mock("@app/screens/account-migration/hooks/use-migration-checkpoint-state", () => ({
   useMigrationCheckpointState: () => ({
     checkpoint: mockCheckpoint,
     accountId: mockAccountId,
+    expectedReceiveSats: mockExpectedReceiveSats,
     clearCheckpoint: mockClearCheckpoint,
   }),
 }))
@@ -62,6 +64,7 @@ describe("useCompleteMigration", () => {
     mockAccounts = [{ id: "sc-account-1" }]
     mockRegistryLoading = false
     mockCheckpoint = "backupAlerts"
+    mockExpectedReceiveSats = 21000
     mockDiscardCustodialSession.mockResolvedValue(undefined)
   })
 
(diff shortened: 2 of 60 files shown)


kngako (dismissed):
Kimi K3 review above with two interesting findings.

esaugomez31 (approved):

PR #4104, opened 2026-08-07
feat(migration): close the custodial account after the migration completes

### LINKED ISSUE #4134: fix(migration): high-severity review findings on the custodial account close
Follow-up to the review on [#4104](https://github.com/blinkbitcoin/blink-mobile/pull/4104) ([review](https://github.com/blinkbitcoin/blink-mobile/pull/4104#issuecomment-5228114740)), which will not be addressed in that PR. This ticket carries the high-severity items only; the four code smells (S1–S4) and G2 stay in the review thread.

All line references are pinned to [`83b1e00`](https://github.com/blinkbitcoin/blink-mobile/commit/83b1e00f8f777400356c15739d0abce6b94c2e37), the current head of the PR branch after the `main` merge — every item below was re-checked against it and still applies.

Common thread: the close is the one irreversible step in the migration, and each item below is a way the app can be wrong about whether it happened — either claiming a refusal that was really a success, or telling support about the wrong account.

---

## 1. A 401 on a user-initiated retry can be the echo of a close that landed, and is treated as a refusal (blocking)

[`use-close-custodial-account.ts:112-113`](https://github.com/blinkbitcoin/blink-mobile/blob/83b1e00f8f777400356c15739d0abce6b94c2e37/app/screens/account-migration/hooks/use-close-custodial-account.ts#L112-L113) claims a 401 is never the echo of an attempt that landed. That holds only for the automatic RetryLink resends #4104 disabled. The echo is still reachable by hand:

1. The first close succeeds server-side — the account is deleted and the Kratos identity is gone — but the response is lost to a network drop, so the hook returns `Retryable`.
2. The user presses **Try again** on the transfer screen, or the resume hook retries on foreground.
3. The second attempt authenticates with the now-dead token, gets a 401, and is classified `Refused`.

The user is reset to the close-refused support screen and support gets a ticket, plus a `Migration account close unauthenticated` non-fatal, for an account that is already deleted. The discard then fires a revocation against the dead identity with `isSessionAlive: true`.

The direction is right — never claim a deletion you cannot prove. What is wrong is the certainty of the refusal path and the comment justifying it.

- Correct the comment: the 401 branch also catches the retry-after-lost-response echo.
- Report it under a distinct label (e.g. "close unverifiable, account may already be deleted") so on-call verifies before acting.
- Word the close-refused support copy so support checks the account's state first.
- If the false handover turns out to be common, a cheap authenticated probe after the 401 could confirm the identity is gone and skip the ticket entirely.

## 2. A deferred close retries only on an OS foreground event, so the old account can stay open for a whole session

In [`use-resume-completed-migration.ts:57-73`](https://github.com/blinkbitcoin/blink-mobile/blob/83b1e00f8f777400356c15739d0abce6b94c2e37/app/screens/account-migration/hooks/use-resume-completed-migration.ts#L57-L73), an unsettled close during a resume sets `hasDeferredCloseRef` and waits for `AppState` to fire `active`. There is no timer, no in-app surface, and no bound. A user whose migration completes and who then keeps using the app for days without backgrounding it leaves the emptied custodial account open and unmentioned the entire time. The state only heals on the next process relaunch, since the checkpoint is kept and resume runs again.

The specs pin this as intended ("retries a deferred close when the app returns to the foreground"), so it is a deliberate design decision — but an invisible one, with no observability if it becomes the common path.

- Pair the `AppState` listener with a bounded backoff timer (retry once after a few minutes while the app stays foregrounded, still one attempt at a time). This keeps the "don't burn attempts against a dead network" property without depending on the OS event alone.
- Alternatively, surface the pending close through the existing wind-down/bulletin machinery so the state is at least visible.
- Either way, add a test that fails if the bound is removed. This subsumes **G2** from the review: no test currently encodes an upper bound, so "waits forever within a session" is locked in as correct.

## 3. A throw from the on-device finish is handed to support as `transfer-failed`

[`migration-transferring-funds-screen.tsx:194-195`](https://github.com/blinkbitcoin/blink-mobile/blob/83b1e00f8f777400356c15739d0abce6b94c2e37/app/screens/account-migration/to-non-custodial/migration-transferring-funds-screen.tsx#L194-L195) catches any throw from `completeMigration()` and navigates with `MigrationSupportReason.TransferFailed`.

The throws that can actually reach this catch are no longer transfer failures. `useCloseCustodialAccount` swallows its own errors into outcomes, so what remains are local failures in `finishOnDevice` — the session discard, the checkpoint clear, or the pending-account write — all of them after the funds have landed. Support receiving `transfer-failed` for a funded wallet with a failed local swap will go looking for a money-movement problem that never happened.

- Add a distinct `MigrationSupportReason` (e.g. `completion-failed`) for the catch in the completion effect, mirroring how `CustodialAccountCloseRefused` got its own reason and copy in #4104.
- The resume hook's catch should use the same reason.
- Pin it with a test that makes `clearCheckpoint` throw and asserts the new reason on the navigation params.

## 4. In the owner-mismatch handover, the ticket names the session's account, not the checkpoint owner's

[`use-complete-migration.ts:103-116`](https://github.com/blinkbitcoin/blink-mobile/blob/83b1e00f8f777400356c15739d0abce6b94c2e37/app/screens/account-migration/hooks/use-complete-migration.ts#L103-L116) handles a checkpoint whose owner id disagrees with the active session by finishing on-device and returning `CloseRefused`. The handover then names `custodialAccountId` = the **active session's** owner.

In the intended legacy case (`checkpointOwnerId === null`) that is the right account. In a true two-id mismatch the emptied account belongs to the checkpoint owner, while the ticket identifies the session's account — support would close or verify the wrong one. The Crashlytics report carries both ids, so this is a ticket-accuracy gap rather than a data-loss risk.

- When `checkpointOwnerId` is a non-null id that differs from the session owner, pass the checkpoint owner's id into the support handover: it is the account the migration actually emptied.
- `checkpointOwnerId` is already exposed by `useMigrationCheckpointState`, so this is a selection change in the mismatch branch plus a test asserting which id lands in the navigation params.

## 5. The retry-link wiring that protects `accountDelete` is untested (G1)

`shouldRetryUnauthorized` is well unit-tested in `__tests__/graphql/retry-policy.spec.ts`, but the composition that actually protects `accountDelete` — [`client.tsx:179-180`](https://github.com/blinkbitcoin/blink-mobile/blob/83b1e00f8f777400356c15739d0abce6b94c2e37/app/graphql/client.tsx#L179-L180) — has no test.

Dropping the `shouldRetryUnauthorized` call from the link leaves every existing test green while silently re-enabling 401 retries on `accountDelete`. That is the exact mechanism item 1 depends on being off, which is why this coverage gap is grouped with the high items rather than left as a nice-to-have.

- Add a link-level test: mock the fetcher, fire `accountDelete` at a 401, assert exactly one request.

---

### Not included

Left in the [review thread](https://github.com/blinkbitcoin/blink-mobile/pull/4104#issuecomment-5228114740) as lower severity: **S1** (`origin: Resume` passed from the commit path), **S2** (`isSessionAlive` restating the close outcome), **S3** (`CloseUnavailable` conflating three waiting states), **S4** (`statusCodeOf` returning `unknown`). **G2** is folded into item 2 above.

esaugomez31:
Closes blinkbitcoin/blink-wip#1057. Backend side is blinkbitcoin/blink#758.

Once a migration reports `COMPLETED`, the app now calls `accountDelete` to close the emptied custodial account and then discards the local session.

## Why the close goes first

`discardCustodialSession` removes the stored profile and clears the token, and that token is the only thing that can authenticate the deletion. Spend it and the account can never be closed, not on this launch and not on any later one. So the close runs before the discard, and a close that never settles leaves the session and the checkpoint exactly as they were, so a later attempt still has something to work with.

Both paths that finish a migration go through `useCompleteMigration` (the transfer screen, and the crash-resume on the next launch), so the call is wired in one place and covers both.

## Three outcomes

**Closed.** Discard, swap to the self-custodial account, success screen. The discard skips the `userLogout` revocation: `markAccountForDeletion` already deleted the Kratos identity and every session with it, so firing it anyway would eat a guaranteed two-second timeout and report a failure on every successful migration.

**Retryable.** Nothing settled server-side, so nothing is discarded. The transfer screen offers a retry and the next launch tries again. This covers a dropped network and the `MIGRATION_STATE_CONFLICT` the backend returns while a transfer is still draining.

**Refused.** The server said no for good. The migration still finishes, because the funds are already self-custodial and holding the user on a dead custodial session would cost them their wallet over an account only support can remove.

A 401 counts as Refused, not Closed. Only a successful payload is proof of a deletion: reporting one that never happened leaves a live custodial account behind an app that said it was gone, with nobody looking for it. `accountDelete` is now off both retry links, so the transport never resends it and a 401 is never the echo of an attempt that landed.

## Every outcome has to leave the user somewhere

The close is irreversible and runs exactly once, so a state the device can never prove its way out of cannot resolve to "try again": the transfer screen swallows hardware back, and an unwinnable retry is a user holding self-custodial funds behind a button that will never work.

The two unprovable states are told apart rather than sharing one answer:

- **The owner query has not answered** (a launch with no connection). Transient, so it stays Retryable, spends nothing and reports nothing. The resume hook re-arms it on the next foreground rather than on the next process start, since coming back to the app is the moment the connection is most likely back.
- **The checkpoint does not name its owner** (a record written before owners existed, which any active session would claim). No later attempt makes that provable, so it resolves as Refused: the migration finishes without ever firing the deletion, and the account goes to support. Reported once per process under a dedup key.

Two callers can also decide to finish at the same moment, since the resume hook stays mounted under the pushed migration stack. The second one now waits on the first and gets its answer, instead of being told "unavailable" and raising a retry footer over a migration that just finished.

## The support handover

Two things had to be sorted before the refusal path could actually work.

`BackupSuccessScreen` resets the whole root stack to Home about two seconds after its animation completes, and it does that from wherever it happens to be mounted. Stacking support on top of it means the handover gets wiped roughly five seconds after it appears, taking with it the only screen where the user can read and copy the ticket. The refusal resets to `[Primary, support]` instead: home underneath, no success screen, no timer. Back then leaves support for the funded wallet.

`useMigrationSupportDetails` skips its `me` query once the session is gone, which by then it is. Without help the ticket would carry the reason code, the self-custodial pubkey and the country, and nothing at all identifying the account the ticket exists to close. `useCompleteMigration` now surfaces the custodial account id, read before the discard clears it, and it reaches the support screen as a route param. The account-not-on-device handover had the same gap and picks this up too. The transfer screen reads that same id rather than opening a second owner query, so the close and the ticket that names it can never be stamped with different accounts.

## Reports

A refusal is the only trace of an account the migration left open, so its report is recorded unconditionally: the connectivity downgrade would otherwise turn `Migration account close failed: request timed out` into a breadcrumb and leave support with a ticket no non-fatal matches. The deletion cap is reported with its own message even when it arrives behind another error, so the label and the text always describe the same rejection.

## Two things worth a second opinion

The AC asks for the deletion cap to render as contact support, which it does. On top of that the refusal is reported with the custodial account id, so support can close the leftover account even if the user never writes in.

Separately: `bypassMaxDeletions` already exists and is passed on every deletion the user did not ask for (admin force-delete, admin phone change, admin email change). A migration has the same shape, since the user is not choosing to delete anything. Passing it on the migration path would make this refusal nearly unreachable and stop generating orphan accounts for support to clean up by hand.

## Testing

100% statement, branch, function and line coverage on every touched file. 84 suites and 926 tests green across `account-migration`, `graphql`, `settings-screen` and the send-confirmation screen that shares the idempotency helper.

`accountDelete`'s own screen had no spec at all, which is how the empty-payload crash survived: it now has one, covering the balance warning, the confirmation flow and all three deletion outcomes.

Not device-tested yet. blink#758 is not merged, and `accountDelete` on a `Migrated` account fails against current prod, so this must not ship before the backend is live.


kngako (dismissed):
Some test coverage gaps and 1 blocking finding to work on.

grimen (approved):

PR #4111, opened 2026-08-08
fix(migration): restart pre-commit checkpoints at the explainer on resume (regardless of checkpoint)

### LINKED ISSUE #4109: Fix migration flow checkpoint persistence and close behavior
## Problem
The migration flow is remembering/resuming too early in the flow: users are being brought back around ToS acceptance. That is not the intended checkpoint.

The checkpoint screen should only be the last screen before money is transferred.

Clarification from Andrej on 2026-08-08: Android arrow Back currently behaves correctly. The actionable issue is the checkpoint/resume behavior, not Android Back navigation.

The full close/exit action was intended to be served by the X button in the upper-right corner, which was part of the original specification.

After the user closes the flow, reopening migration should start from the beginning unless the user has reached the actual final checkpoint before money transfer.

## Implementation sequence
Because this is being handled last minute, implement in two steps to reduce risk:

1. Fix the migration checkpoint/resume point so the flow is not persisted at ToS acceptance or other early steps. The checkpoint should only apply at the last screen before money transfer.
2. Add the X button in the upper-right corner where relevant, and ensure closing/exiting does not persist an early flow step.

## Expected behavior
- Android arrow Back continues to behave correctly.
- The flow does not resume at ToS acceptance.
- The checkpoint screen appears only as the last screen before money is transferred.
- The X button closes/exits the migration flow where relevant.
- After closing/exiting before the real checkpoint, the next launch starts at the beginning of migration.
- Tests cover checkpoint persistence and restart-after-close behavior.

## Source
Requested by Lukas from Slack `#2-tech-bot` on 2026-08-08. Clarified by Andrej in the same Slack thread on 2026-08-08.

Requested stakeholders from Slack: <@U09TT98GTCZ> (Jonas), <@U0B81DVCNM6> (Esaú), <@U0BJ058EX1D> (Kgothatso).

Known GitHub stakeholders tagged: @grimen, @esaugomez31, @designsats.

kngako:
Only the BalancesOverview commit point resumes mid-flow; every earlier checkpoint (terms, backup method, cloud backup, backup alerts) now resolves to accountMigrationExplainer so the flow re-walks backup before offering the funds transfer again.

### DIFF
--- a/__tests__/screens/account-migration/hooks/use-migration-checkpoint.spec.ts
+++ b/__tests__/screens/account-migration/hooks/use-migration-checkpoint.spec.ts
@@ -359,24 +359,6 @@ describe("useMigrationCheckpoint", () => {
     expect(mockNavigate).toHaveBeenCalledWith("accountMigrationExplainer")
   })
 
-  it("restarts at the explainer when resuming from the terms checkpoint", async () => {
-    mockLoadCheckpoint.mockResolvedValue({
-      step: MigrationCheckpoint.TermsAndConditions,
-      savedAt: Date.now(),
-      accountId: "sc-account-1",
-    })
-
-    const { result } = renderHook(() => useMigrationCheckpoint())
-
-    await waitFor(() => expect(result.current.loading).toBe(false))
-
-    act(() => {
-      result.current.navigateToCheckpoint()
-    })
-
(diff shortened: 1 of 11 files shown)


esaugomez31 (approved):
LGTM!

grimen (approved):

PR #4119, opened 2026-08-09
fix(receive): rotate self-custodial on-chain address after a deposit

### LINKED ISSUE #4113: bug(self-custodial): Non-custodial on-chain receive address does not rotate after deposit
## Screenshot

Public X report about non-custodial on-chain receive address not rotating

## Summary
A migrated Android user reports that after using non-custodial mode and receiving BTC on-chain, the on-chain receiving address remains unchanged instead of rotating/presenting a fresh deposit address.

## User report
From public X screenshot:

> Is this a bug on my side?
>
> Im using non-custodial after I migrated in app(android).
>
> I recieved some btc on chain and the recieving addr does not change.
> I even label it, same deposit addr.

A reply from the Blink account says we present fresh addresses every time, so this appears to be an implementation bug.

## Expected behavior
After a non-custodial migrated user receives BTC on-chain, the app should present a fresh on-chain receiving address for the next deposit.

## Actual behavior
The user reports the same on-chain deposit address continues to be shown, even after labeling it.

## Known context
- Platform: Android
- Account state: migrated to non-custodial in-app
- Flow: receive BTC on-chain, then check the next on-chain receive/deposit address
- Reporter source: public X screenshot shared by @ihaveadifferentname in Slack #2-tech-bot on 2026-08-09

## Notes
Please verify whether address rotation is expected to occur immediately after an observed on-chain receive for migrated non-custodial wallets, and whether labels are accidentally pinning/reusing the same deposit address in the Android receive flow.

grimen:
Fixes #4113

## Problem

A migrated Android user received BTC on-chain into their non-custodial wallet and kept seeing the **same** deposit address afterwards. Reusing a paid address is a privacy leak: anyone who has ever paid it can watch the chain and link every later deposit to the same person.

The Spark SDK takes an explicit rotation flag, and its own typedef spells out the semantics:

> If true, rotate to a new deposit address. **Previous ones remain valid.** If false or absent, return the existing address (creating one if none exists yet).

Our only call site never set it, and the receive hook fetched the address once per SDK connection and parked it in `useState` — so the address was frozen for the session, and the next session got the same one back. This has been the behaviour since the file was first written; nothing was deliberately disabled.

Two things this was *not*: the user's labelling had no effect (there is no BIP21 `label=` parameter anywhere in the app), and rotation cannot strand old funds (`listUnclaimedDeposits` is wallet-wide and keyed by `txid:vout`, auto-claim included).

## Approach

The SDK never reports *which* address a deposit landed on — `PaymentDetails.Deposit` carries only a `txId`, and the deposit events carry only outpoints. So reuse is inferred rather than observed:

**If the wallet's on-chain receive history has grown since we handed out this address, the address was used — rotate.**

The id of the newest on-chain receipt is stored per account alongside the issued address. On the next visit (or the moment a receipt lands while the QR is on screen) the marker is recompared, and a mismatch triggers `newAddress: true`. This covers the reported case where the deposit arrives while the app is closed, and bounds address growth to roughly one per real deposit rather than one per screen visit.

## Changes

- `app/self-custodial/bridge/receive.ts` + `app/types/payment.ts` — the adapter can now request rotation; omitting the param preserves today's behaviour for every other caller.
- `app/self-custodial/storage/onchain-address.ts` *(new)* — the issued-address record and the `latestOnchainReceiptId` inference.
- `app/self-custodial/hooks/use-payment-request.ts` — marker-driven fetch/rotate effect. Requests are sequenced so a slow fetch cannot land after the one that replaced it.
- `app/screens/receive-bitcoin-screen/` — no UI change; the on-chain page keeps its two action buttons.

Custodial accounts are untouched — their address comes from the server's `onChainAddressCurrent`.

## Testing

`__tests__/self-custodial/bridge/receive.spec.ts` is new: there was no bridge receive spec at all, so the `newAddress` argument reaching the SDK was never asserted — which is precisely what let this ship. Also new: `storage/onchain-address.spec.ts` and `hooks/use-payment-request-onchain-rotation.spec.ts` (rotate on deposit, rotate on relaunch after a deposit, no rotation for a lightning receipt or a bare reconnect, manual rotation, failure keeps the current address).

163 tests pass across the 10 receive-related suites. Both fixes were mutation-checked: reverting the bridge flag and neutering the rotation decision each make the new specs fail.

## Address monitoring — resolved

The earlier draft flagged a gap limit as an open question. It is answered by Breez's own documentation for this SDK, which matches the doc comment in the bindings we ship (`@breeztech/breez-sdk-spark-react-native` 0.15.0):

> By default the existing address is returned; you can optionally request a new address to rotate to a fresh one for improved privacy. **All previously generated addresses remain monitored.**
> — [Receiving payments, Breez SDK Nodeless (Spark)](https://sdk-doc-spark.breez.technology/guide/receive_payment.html)

> If true, rotate to a new deposit address. **Previous ones remain valid.**
> — `ReceivePaymentMethod.BitcoinAddress`, generated bindings

So rotation cannot orphan funds sent to an older address, and no gap limit is documented. That removes the correctness risk; the rotate-on-use policy still stands on its own, since it keeps the number of issued addresses proportional to real deposits rather than to how often someone opens the Receive screen.

## Not done yet

- Manual verification against a real regtest deposit. The automatic path is covered by tests.

## Follow-ups (not in this PR)

Deliberately kept out so this diff stays about the bug:

- **#4121** / **#4122** — the receive request-state type is named `SelfCustodialPaymentRequestState` but types both account types, and two unrelated types share the name `InvoiceData`. Cleanup only, no behaviour change. #4122 is stacked on this branch.

---

## Update: the manual button is gone

A **New address** action shipped in the first draft of this PR and was removed in [`d38f37b6`](https://github.com/blinkbitcoin/blink-mobile/commit/d38f37b6613b1776b21fb89f86729c6e7dc61f7e) after review — @designsats and @pretyflaco both objected, and the reasoning holds: no user interaction should be required to avoid address reuse, and the action row has no space for a third button. Everything that existed only to serve it went with it, including the two `ReceiveScreen` i18n keys.

The automatic rotation described above is unchanged and is what fixes #4113.


ihaveadifferentname (approved):

esaugomez31 (approved):
LGTM

PR #4122, opened 2026-08-09
refactor(receive): give the shared receive request-state an honest name and home

### LINKED ISSUE #4121: refactor(receive): the shared receive request-state type is named self-custodial
## Summary

The receive screen's request-state contract is called `SelfCustodialPaymentRequestState` and lives in `app/self-custodial/hooks/types.ts`, but it describes the state of **both** account types. The custodial hook's return value is typed with it too.

## Where it shows

Every one of these is holding custodial state in a type named self-custodial:

- `app/screens/receive-bitcoin-screen/receive-screen.tsx` — `requestState: SelfCustodialPaymentRequestState`, assigned from `custodialRequest` when the account is custodial
- `app/screens/receive-bitcoin-screen/hooks/use-receive-flow.ts`
- `app/screens/receive-bitcoin-screen/hooks/use-receive-carousel.ts`
- `app/screens/receive-bitcoin-screen/hooks/use-display-payment-request.ts`
- `app/screens/receive-bitcoin-screen/hooks/use-onchain-resolver.ts`

## Why it is worth fixing

It is actively misleading in a codebase where the custodial/self-custodial split is load-bearing. A reader auditing "what is Spark-only?" gets the wrong answer from the type name, and anyone adding a field to the shared contract has to decide whether a self-custodial-named type is the right home — a question that shouldn't exist. It came up while reviewing #4113's fix, where a genuinely self-custodial-only field had to be added to this shared type.

The dependency also points the wrong way: the screen imports its own contract from `app/self-custodial/`, while `app/self-custodial/hooks/types.ts` imports back from `app/screens/receive-bitcoin-screen/payment/index.types`.

## Second, related wart

There are two different types named `InvoiceData`:

- `app/self-custodial/hooks/types.ts` — `{ invoiceType, paymentRequest?, address?, username?, getFullUriFn, getCopyableInvoiceFn }`
- `app/screens/receive-bitcoin-screen/payment/index.types.ts` — a union of `LightningInvoiceData | OnChainInvoiceData | PayCodeInvoiceData`

Same name, different shapes, both in the receive path. Whichever one a reader has in mind, half the code disagrees.

## Proposal

Rename the state type to something account-neutral and move it, with its `InvoiceData`, into a module owned by the receive screen — so the screen owns the contract and both hooks implement it. Mechanical rename plus import updates; no behaviour change, so the existing receive suites should stay green untouched.

Not urgent, and deliberately kept out of the #4113 fix to keep that diff about the bug.

grimen:
Fixes #4121.

## Problem

`SelfCustodialPaymentRequestState` types the request state of **both** account types. Five screen files were holding custodial state in a type whose name says otherwise:

- `receive-screen.tsx` assigns `custodialRequest` to a `SelfCustodialPaymentRequestState`
- `use-receive-flow.ts`, `use-receive-carousel.ts`, `use-display-payment-request.ts`, `use-onchain-resolver.ts` all alias it as their `RequestState`

In a codebase where the custodial / self-custodial split is load-bearing, that name gives the wrong answer to "what is Spark-only?". It also forced an awkward call in #4119: a genuinely self-custodial-only field (`rotateOnchainAddress`) had to be added to it, because that is where every other shared field already lived.

The dependency ran both ways too — the screen imported its contract from `app/self-custodial/`, while `app/self-custodial/hooks/types.ts` imported `InvoiceType` / `PaymentRequestStateType` back from the screen.

And there were two unrelated types both named `InvoiceData`: one in `self-custodial/hooks/types.ts`, one in `payment/index.types.ts` (a union of `Lightning | OnChain | PayCode` data). Same name, different shapes, same code path.

## Change

- `SelfCustodialPaymentRequestState` → **`ReceivePaymentRequestState`**
- the self-custodial `InvoiceData` → **`ReceiveInvoiceInfo`**, which frees the name it was shadowing
- both move from `app/self-custodial/hooks/types.ts` to **`app/screens/receive-bitcoin-screen/payment/request-state.types.ts`**

The screen now owns the contract and both hooks implement it, so the dependency runs one way: `self-custodial → screen`. A field only one side can populate stays optional and documents itself (`rotateOnchainAddress` already carries "absent for custodial accounts") rather than the whole type claiming an owner.

Git records the type file as a rename, not a rewrite — the body is unchanged apart from the two identifiers.

## Testing

Types only; there is no behaviour to test. The point is that **no test needed changing** — 423 tests across the 32 receive and self-custodial hook suites pass untouched, and `tsc` and `eslint` are clean. No test imported either type, so nothing was adjusted to make this pass.

## Note on the stack

#4119 has since merged, and this branch was rebased onto `main`, so it now stands alone: one commit, eight files. The rebase took a single conflict, in `use-payment-request.ts`, where `main` had gained the `resolveDepositMarker` helper on the same lines this renames the return type — the helper is kept whole and only the type name changed.


esaugomez31 (approved):
LGTM

PR #4126, opened 2026-08-09
fix(migration): keep checklist content clear of the sticky CTA in enlarged text mode

### LINKED ISSUE #4123: bug(migration): checkboxes collide with CTA button in grandma mode
## Screenshot

Migration checklist CTA collision in enlarged text mode

## Source

Requested from Slack `#2-tech-bot` by @designsats on 2026-08-09, with screenshot attached in the Slack thread. Follow-up E2E testing requirement added by @grimen in the same thread.

## Problem

In the non-custodial migration flow, when the user has enlarged system text / accessibility text scaling enabled ("grandma mode"), checklist content can collide visually with the primary CTA button.

Observed in the migration screen titled:

> What does it mean to move to non-custodial?

The bottom checklist item extends behind the sticky CTA (`I understand`), making both the checkbox/text and button hard to read and interact with.

## Expected behavior

- Checklist rows should remain readable at enlarged text sizes.
- The CTA should remain visually separated from content and fully tappable.
- Content near the bottom should scroll above the CTA or have enough bottom inset/padding.
- The default CTA styling should prevent text/content behind it from showing through.

## Suggested approach

Research whether this can be fixed at the shared component level. If possible, CTA buttons should have a solid black background by default so underlying content cannot bleed through or collide visually.

Also verify whether this screen needs additional bottom padding / safe-area-aware spacing when the CTA is sticky.

## Acceptance criteria

- Reproduce with enlarged system text / accessibility font scaling on simulator.
- Confirm no checkbox, text, or checklist row overlaps the primary CTA.
- Confirm the CTA remains readable and tappable.
- Validate on the migration checklist screen and any other migration screens using the same sticky CTA pattern.

grimen:
Fixes #4123

## Problem

With iOS accessibility text scaling turned up, the migration checklist screen ("What does it mean to move to non-custodial?") renders its last checkbox row on top of the **I understand** button, and the label reads straight *through* the button.

Two things combine to produce that one screenshot:

**1. The migration step content could not scroll.** Every migration step screen is built on `migration-step-layout.tsx`, which laid its content out in a plain `<View style={{flex: 1}}>` inside a `preset="fixed"` Screen. Enlarged text makes the hero plus five checkbox cards taller than that box, and React Native does not clip overflowing children — it draws them outside their parent, over the footer beneath. Nothing scrolled either, so rows 3–5 were unreachable and the checklist could never be satisfied.

**2. The footer band did not paint itself.** A disabled `GaloyPrimaryButton` is `opacity: 0.5` over primary, which is not a colour — it blends the button with whatever is behind it. The band under the sticky CTA relied on the screen showing through, so anything that reached those pixels was composited into the button.

## Changes

**`app/screens/account-migration/migration-step-layout.tsx`** — the content slot becomes a `ScrollView` with `contentContainerStyle: { flexGrow: 1 }`. `flexGrow` rather than `flex` means short content keeps exactly today's full-height box, so nothing moves at normal text sizes, while tall content is free to grow past the viewport and scroll. `contentStyle` now routes to `contentContainerStyle`, where the `{ gap: 20 }` that `migration-required-screen` passes belongs.

The footer stays a **flex sibling below** the scroll view rather than an overlay. That is the load-bearing part: the scroll viewport is always whatever height the buttons leave behind, so content cannot reach the footer's pixels — the overlap is impossible by construction rather than tuned away with padding, which would only hold until someone picks a larger text size. On top of that the band now paints `colors.white`, the same screen-surface token `Screen` uses, so the translucent CTA has an opaque surface of its own instead of borrowing one.

The scaffold also keeps the scroll indicator (it has no visible clipping edge, so the indicator is the only cue that content continues below the fold) and follows content that grows while the user is already at the bottom — `RevealedCheckboxList` reveals its next acknowledgement box only once the previous one is ticked, and at large text that box lands off-screen with nothing to scroll it into view.

This one edit covers all five screens on the scaffold: explainer, keep-receiving, download-history, api-service and migration-required. `balances-overview-screen` and `contact-support-screen` hand-rolled the same layout and already have their own ScrollViews.

**`app/components/atomic/galoy-primary-button/galoy-primary-button.tsx`** — one line: the loading spinner takes the title's colour rather than the library's hardcoded `'white'`, which was wrong against the black title the orange fill carries in dark theme. The button's resting appearance is untouched.

## Decisions

- **The button is not restyled.** An earlier revision of this PR gave every disabled primary button in the app an opaque `grey4` fill, a `grey3` outline and a darker title, so no CTA anywhere could composite content behind it. @designsats reviewed that and asked for the opposite split: the surface a sticky button sits on paints itself, the button stays exactly as designed. That is what is on the branch now — the app-wide restyle is withdrawn, along with the WCAG-contrast argument that only applied to the grey fill it introduced.
- **The band, not the button, owns the pixels.** `backgroundColor` on `buttonsContainer` rather than on the button means the invariant holds for whatever the footer contains — primary, secondary, or two stacked buttons — and costs nothing at normal text sizes.
- **Sibling footer, not an absolute overlay.** An overlay would need the scroll content padded by the measured footer height to keep the last row reachable; the sibling gives the same result (rows clip at the scroll edge, band solid below) with no measurement, and matches `contact-support-screen`, `balances-overview-screen` and `onboarding-screen-layout`, which all use this shape.

## Tests

- `__tests__/screens/account-migration/migration-step-layout.spec.tsx` (new, 13 tests) — asserts structure rather than pixel values: content renders inside a scroll container, the footer renders *outside* it, the band's paint comes from the footer rather than the screen behind it (the nearest painted ancestor must stop short of the scroll area), the content container can grow, `contentStyle` reaches the content container and not the scroll view, the indicator is not suppressed, and the five auto-scroll cases (grows at the bottom, first measurement, user scrolled up, user scrolled back down, content shrinks).
- `__tests__/components/atomic/galoy-primary-button.spec.tsx` (new, 3 tests) — the spinner tracks the title colour, callers can still override it, and the testID falls back correctly for a rendered title.

Reverting both source files to `main` fails 13 of the 16, so none are vacuous. Typecheck and lint clean; the 12 specs covering screens on this scaffold (236 tests) pass unchanged.

### DIFF
--- a/__tests__/components/atomic/galoy-primary-button.spec.tsx
+++ b/__tests__/components/atomic/galoy-primary-button.spec.tsx
@@ -2,6 +2,7 @@ import React from "react"
 import {
   ActivityIndicator,
   StyleSheet,
+  Text,
   View,
   type TextStyle,
   type ViewStyle,
@@ -164,5 +165,22 @@ describe("GaloyPrimaryButton", () => {
     )
 
     expect(getSpinnerColor()).toBe("#123456")
+  })
+
+  it("takes its testID from a string title, and skips it for a rendered one", () => {
+    renderButton()
+    expect(screen.getByTestId("Continue")).toBeTruthy()
+    screen.unmount()
+
+    // A node title has no string to name the button by, so the automatic
+    // testProps are skipped rather than stringified into a useless id.
+    render(
+      <ContextForScreen>
+        <GaloyPrimaryButton title={<Text>Continue</Text>} onPress={() => {}} />
+      </ContextForScreen>,
+    )
+
+    expect(screen.getByText("Continue")).toBeTruthy()
+    expect(screen.queryByTestId("Continue")).toBeNull()
   })
 })
--- a/__tests__/components/atomic/galoy-primary-button.spec.tsx
+++ b/__tests__/components/atomic/galoy-primary-button.spec.tsx
@@ -1,12 +1,5 @@
 import React from "react"
-import {
-  ActivityIndicator,
-  StyleSheet,
-  Text,
-  View,
-  type TextStyle,
-  type ViewStyle,
-} from "react-native"
+import { ActivityIndicator, StyleSheet, Text, type TextStyle } from "react-native"
 import { render, screen } from "@testing-library/react-native"
 
 import { GaloyPrimaryButton } from "@app/components/atomic/galoy-primary-button"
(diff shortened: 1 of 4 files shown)


designsats (approved):
Works as intended on android. Tested in emulator.

esaugomez31 (approved):

PR #4127, opened 2026-08-09
feat(send): show balances in send flow when hide balance is on

### LINKED ISSUE #4125: feat(hide-balance): show user balances in send flow balances overview when user taps on wallet selector
## Problem
In the Send flow, when the user taps the wallet selector / balances overview, balance amounts are not shown. The user has to exit the Send flow to check balances, then return to decide which balance to use.

## Expected behavior
When the user opens the wallet selector / balances overview from the Send flow, show enough balance information for them to decide which balance to send from.

## Reason
The user needs to be able to check their balances inside the Send flow to choose the correct balance without leaving the flow.

## Source
Requested by Andrej from Slack message: https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1786287999745639

Related GitHub PR comment: https://github.com/blinkbitcoin/blink-mobile/pull/3861#issuecomment-5227614560

## Notes
This is split out from the PR discussion so Send flow balance visibility can be tracked independently from hide-balance persistence.

grimen:
Fixes #4125

## Problem

With **Hide balance** enabled (Settings → Security), opening the Send flow and tapping the **From** wallet selector shows the masked placeholder (••••) on both rows instead of amounts. Nothing in the flow reveals them — the placeholder is not tappable and there is no eye toggle — so the user has to back out of Send, reveal balances on the home screen, and come back just to decide which wallet to pay from.

## Change

Hide balance exists to keep amounts off the screen at a glance. Once the user has deliberately entered the Send flow they are transacting and need the numbers, so the send screens no longer participate in it:

- **Details screen** — the wallet-picker rows and the collapsed "From" field always render the primary and secondary balance.
- **Confirmation screen** — the "From" block always renders the balance, and the insufficient-balance error now reads "Amount exceeds your balance of 0.00088413 BTC" instead of "Amount exceeds your balance of \*\*\*\*".
- Removed the dead `HIDDEN_AMOUNT_PLACEHOLDER` constant (its only consumer was that error string) and the now-unused `walletSelectorInfoContainerHidden` style.

Hide balance is **unchanged everywhere else**: home balance header, wallet overview, transaction list and transaction detail all still mask and still offer tap-to-reveal. Entering the Send flow does not flip the session flag.

## Decisions taken, so they can be challenged

Recording these explicitly — each one is a place where a reviewer may reasonably disagree.

**1. The Send flow opts out of hide-balance entirely, rather than gaining a reveal gesture.**
Considered and rejected: making the placeholder pressable in the Send flow (mirroring `balance-header.tsx`, which wraps the balance in a `TouchableOpacity onPress={switchMemoryHideAmount}`). Rejected because the wallet-picker row's press already has an owner — it selects the wallet — so a reveal gesture would either collide with selection or need a second target inside a row that is 40pt tall. Reveal-on-open would also be indistinguishable from just showing the amounts, at the cost of a stateful branch. *If a reviewer thinks hide-balance should be absolute, this is the decision to push back on.*

**2. Scope is the whole Send flow, not just the picker modal.**
The issue only names the picker. Extended to the collapsed "From" field and the confirmation screen because stopping at the modal produces a worse state than today: the user picks a wallet by its balance, the modal closes, and the balance they just used to decide is masked again on the very next screen. Confirmed with @grimen before implementing.

**3. The insufficient-balance error now names the real balance.**
`LL.SendBitcoinScreen.amountExceed({ balance })` previously interpolated `"****"` when hidden, producing "Amount exceeds your balance of \*\*\*\*". That is the one place where masking actively withheld the number the message exists to communicate. No i18n change — the `{balance}` placeholder is untouched, only what we pass into it.

**4. `HIDDEN_AMOUNT_PLACEHOLDER` deleted rather than left in place.**
`git grep` shows the confirmation screen was its only consumer; it predates `HiddenBalancePlaceholder`, which replaced the `****` text with dots everywhere else. Leaving it would be a dangling "this is how we mask" signal pointing at nothing. Trivially revertible if anyone wants it kept.

**5. The existing spec was inverted, not deleted.**
`send-details-hide-balance.spec.tsx` came from #3861 and asserted exactly the behaviour this PR removes (3 placeholders while hidden). Its scaffolding — the mocked `useSendWallets`, the placeholder stub, and `openWalletPicker()`, which calls `props.onPress()` directly because RNTL cannot drive `TouchableWithoutFeedback` — is the expensive part and was kept verbatim, so the diff is assertions only.

## Tests

- picker shows amounts when balances are visible (unchanged)
- picker still shows amounts while `hideAmount` is true (regression guard for #4125)
- the inline "From" field shows the balance while `hideAmount` is true

The existing hide-balance specs (balance header, wallet overview, transaction item, `HideAmountContainer`, placeholder component) are untouched and green — that is what proves the setting still works outside the Send flow.

## Not in scope

Two other points raised in the same thread are deliberately left out, since neither is a visibility bug:

- hide-balance **persistence** across app restarts (the `hideBalance` Apollo client field / `HideAmountContainer` re-sync)
- the **Android flicker** where the balance block resizes once loaded and the rows jump — removing the second container style may incidentally reduce it, but it is not a fix


kngako (dismissed):
Blocking on test coverage

designsats (approved):
should be good then

PR #4128, opened 2026-08-09
fix(home): show the pending onchain deposit as a row under the balance

### LINKED ISSUE #4120: fix(ui): Pending onchain deposit should persist on home screen
## Screenshot

Pending onchain deposit badge layout issue on home screen

Requested by: @designsats (Andrej) from Slack.

## Problem

The pending onchain deposit indicator on the home screen currently breaks on small screens when the user has enlarged system fonts enabled.

The current badge treatment also does not match the intended UI direction for this state.

## Expected behavior

Pending onchain deposits should persist visibly on the home screen without layout breakage across small devices and accessibility font sizes.

## Proposed solution

- Recreate the pending onchain deposit row using the existing outgoing transaction style.
- Add and use a `chain.svg` icon for the onchain pending state.
- Follow the Figma reference for spacing, typography, and icon placement.

## Design / context

- Figma: https://www.figma.com/design/9MQuQi8ZhXVvDibWSI3C4c/Blink?node-id=20023-103103&t=YhJoSqggisNoc4AR-1
- Builds on top of https://github.com/blinkbitcoin/blink-mobile/pull/4079

## Acceptance criteria

- Pending onchain deposit remains visible on the home screen.
- Layout holds on small screens.
- Layout holds with enlarged fonts / accessibility text sizes.
- Visual treatment matches the outgoing transaction style from the existing app UI.
- `chain.svg` is added and used where appropriate.

grimen:
Fixes #4120.

## What

The pending-deposit indicator was an amber pill beside the 32pt balance, so on a small screen with enlarged system fonts the two collided and pushed the balance off-centre. It now sits on its own line directly under the balance — the slot the unseen-transaction badge already uses — styled like an outgoing transaction badge (20pt, `grey2`) with the chain icon in front of the amount, per the Figma reference and @designsats' notes on the issue.

Scope is unchanged: the row shows the same amounts the pill did — custodial pending incoming transactions and self-custodial immature Spark deposits.

## How

- **Shared `AmountBadge`** (`app/components/amount-badge/`) carries the look, drop-in/out animation and font handling for that slot. `UnseenTxAmountBadge` and the new `PendingAmountBadge` are thin wrappers over it, so the two tenants of one slot cannot drift apart. The badge caps font scaling at 1.4x (same ceiling as the balance) and truncates on one line — protection the old badge never had, and the point of this issue.
- **`useBadgeSlotContent`** arbitrates the slot: the transient unseen badge takes it while it shows, plus 180ms so its exit animation finishes, and for at most one hold window per transaction — a stuck seen-state (an unfocused screen never schedules its dismissal) must not cost an unconfirmed deposit its slot forever. The pending row takes the slot back and keeps it until the deposit confirms.
- **`chain.svg`** is registered in the icon map and switched to `fill="currentColor"` so it follows the theme instead of being frozen at `#919BBA` — it now takes the exact colour token the amount text uses, in both themes.
- The `statusBadge`/ghost-pill machinery leaves `BalanceHeader`, and `StatusPill` (whose only consumer that was) is deleted in a separate commit, so keeping the atom is a one-line revert.
- Tapping the row with immature deposits still opens the unclaimed-deposits screen; custodial pending receives stay inert. `HomeScreen.pendingReceiveBadge` is kept as the row's accessibility label — the row itself only shows `+$78.31`, so the word "pending" has to reach screen readers somehow. No translation files change.

## Verification

`yarn tsc --noEmit` and lint clean; `home.spec`, `balance-header.spec`, `unseen-tx-amount-badge.spec` and the two new `amount-badge` specs pass (93 tests).

Screenshots below are from an isolated simulator with the pending amount forced to a fixed value (a real unconfirmed deposit is not reproducible locally); everything else is the real screen.

## Note for review

`badgeSlot` goes from `height: 35` to `minHeight: 35` — at 20pt x 1.4 the line box overruns a fixed 35, which is the same class of failure this issue is about. It nudges home-screen layout slightly at large text sizes for everyone, not only users with a pending deposit.


designsats (commented):
visually good, unable to test full interaction because this isnt implemented yet https://github.com/blinkbitcoin/blink-mobile/pull/4068 and its occupying same space.....

kngako (dismissed):
All looks good to me. Unsure how this would relate to the changes in #4068 as I haven't checked that out yet.

esaugomez31 (approved):
LGTM!

Here is what I addressed in this PR:

- **The pending row ignored the hide-amount setting.** The `!hideAmount` guard went away with `BalanceHeader.showBadge`, so a hidden balance still had the deposit amount spelled out underneath it — and announced by screen readers. Restored on `hasPendingAmount`, with the test that was deleted along with it.

- **The pending row could blink out for ~650ms.** `useBadgeSlotContent` reset `holdExpired` on the way *out* as well as the way in, so once the hold had handed the slot over, the unseen badge merely finishing took the slot back for one `durationOut` — unmounting the pending row and replaying its entry animation. The window is now re-armed only by a new transaction (`unseenKey`/`holdMs`), not by the previous one ending.

- **The font-cap test could not fail.** It asserted `<= 1.5` against a `1.4` constant, so loosening the cap back to the overrun this issue exists to fix would have stayed green. It now pins the exact value.

- **`mockIsFocused` leaked across suites.** It was set in a `beforeEach` but missing from `resetHomeScreenMocks`, so it silently changed the focus state of every describe block below it. Added to the reset.

- **Inline conditions in JSX props.** `disabled={!visible || !onPress}`, `accessibilityRole={onPress ? ... }`, `importantForAccessibility={...}`, `variant={isOutgoing ? ...}` and the pending row's `onPress` ternary are now named constants in the component bodies.

- **A mock of an API that does not exist.** `balance-header.spec` stubbed `switchMemoryHideAmount`, which appears nowhere in the codebase and which no assertion referenced. Removed, so the mock documents the real contract again.

All four commits are pushed to this branch. Typecheck, lint and the affected suites are green, and each fix ships with a test that fails without it.

Issue #3381, opened 2025-02-24
Hide balance should persist


designsats:
**Problem**
Tapping on balance on home screen hides balance, but this change doesnt persist when app is re-opened. User has to go to Settings, scroll down and change it there. Most dont do this. 
Furthermore, even changing this setting, resets every few moenths when new update gets pushed. 

**Solution**
Cache user choice server side. When balance is tapped on home, it should remain hidden. Similar to how other wallets do it.

designsats:
Resolved now

PR #4130, opened 2026-08-09
feat(hide-balance): persist the user's show/hide choice

### LINKED ISSUE #4124: feat(hide-balance): persist user choice when showing/hiding balance
## Problem
The hide/show balance setting does not persist after the user toggles it.

## Expected behavior
When a user chooses to show or hide their balances, that preference should persist across app sessions and relevant balance views until the user changes it again.

## Source
Requested by Andrej from Slack message: https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1786287999745639

Related GitHub PR comment: https://github.com/blinkbitcoin/blink-mobile/pull/3861#issuecomment-5227614560

## Notes
This is split out from the PR discussion so the persistence behavior can be tracked independently from send-flow balance visibility.

grimen:
Fixes #4124.

## What was broken

Two separate things, both reported as "hide balance doesn't persist":

1. **The tap toggle was session-only by design.** `hideAmount` was a plain `useState` in `HideAmountContainer`, thrown away when the app died. Tapping the eye hid your balance until you closed the app.
2. **The "Always hide balance" setting was not reliably persisted either.** It lived as an Apollo local-only cache field, and `app/graphql/client.tsx` skips `persistor.restore()` when there is no auth token — so self-custodial and logged-out users lost it on every launch, and everyone lost it on logout or a cache schema bump.

## Behavior now

| "Always hide balance" | At launch |
| --- | --- |
| **ON** | Always starts hidden. Tapping to reveal is a peek: it is not remembered. |
| **OFF** | Starts in whatever state you last left it in — hidden stays hidden, shown stays shown. |

Both values live in `PersistentState` (schema 16), which is independent of the auth token and of Apollo cache purges.

## Decisions, and what was rejected

Recording these so they can be argued with rather than reverse-engineered from the diff.

**1. Two-tier semantics rather than one boolean.**
The alternative was collapsing the eye toggle and the Security switch into a single persisted value. Rejected: "Always hide balance" would lose its meaning (it becomes just "balance is hidden"), and its title and description would need re-wording across EN + 28 locales. The two-tier model also satisfies the original request verbatim — #2460 item 3 asks for the toggle to be remembered *"even if default hide balance set to off"*, which presumes the default setting still exists. Sources: #2460 (item 3), #3381, and @designsats on #3861.

**2. Device-wide, not per-account.**
The alternative was keying by `resolveAccountKey` like `themeByAccountId` and the other preference modules. Rejected: hiding your balance is a choice about who can see your screen, not about an account — switching from custodial to self-custodial should not reveal it again.

**3. `PersistentState`, not the Apollo cache.**
This is the actual root cause of half the bug (see above), not just a tidier home. `PersistentState` loads before anything renders and is untouched by `clearStore`.

**4. Local `useState` stays the render source of truth.**
The persisted write is a mirror, fire-and-forget. A tap must never wait on AsyncStorage, and a failed write must never revert what is on screen.

**5. The legacy adoption bridge.**
`HideAmountContainer` reads the old Apollo `hideBalance` once at mount so users who set it on an older build keep it. It uses `client.readQuery` rather than `useHideBalanceQuery` — a one-shot read, not a live subscription that would fight the new state. The Apollo field stays declared, read-only, and can be deleted after a couple of releases; the comment in `client-only-query.ts` says so.

**6. Logout still clears the preference.** `use-logout.ts` calls `resetState()`, which wipes both new fields. This matches today's behavior (the Apollo cache is purged on `clearStore` too), so it is not a regression — but it means "device-wide" is really "device-wide until logout". Flagging it rather than leaving it to be discovered.

**7. Two pieces of adjacent cleanup, both inside code this PR rewrites:**
- `hiddenBalanceToolTip` — a local schema field with a writer, a typePolicy and zero readers. The only writer was the handler being rewritten here, so it goes.
- `switchMemoryHideAmount` → `toggleHideAmount`. The old name now actively lies.

## Out of scope

#4125 (balances in the send-flow wallet selector) and the Android flicker from the same Slack thread. This PR is the persistence half only.

## Verification

Unit: 13 new cases on the container state machine (seed, peek-vs-remember, setting transitions, adoption), plus new specs for the persistent-state helpers, the v15→v16 migration, and the Security screen switch — which also had no coverage at all before.

Manual, on an iOS simulator, force-quitting between every step:

1. Tap the eye to hide → relaunch → **still hidden**.
2. Tap to reveal → relaunch → **still shown**.
3. Turn "Always hide balance" ON → relaunch → **hidden**, even though the last remembered state was "shown".
4. Peek while it is ON → relaunch → **hidden again**, the peek was not remembered.


designsats (commented):
Works as intended.

designsats (approved):
works as intended, still

esaugomez31 (approved):

PR #4138, opened 2026-08-10
fix(android): useScreenSecurity to enforce screenshot protection

kngako:
<html><head></head><body><h1>Restore-phrase screen has no <code>useScreenSecurity()</code> — the funded mnemonic is screenshot- and recording-capturable</h1>
<p><strong>Component:</strong> <code>app/screens/self-custodial/onboarding/restore/</code>
<strong>Severity:</strong> Medium–High — impact is total, irreversible loss of a funded wallet; likelihood requires an already-present capture channel on the device.
<strong>Type:</strong> Missing security control / inconsistent application of an existing one</p>
<hr>
<h2>Summary</h2>
<p><code>useScreenSecurity()</code> is applied to the two screens that <strong>display</strong> a mnemonic and to the API-key reveal, but not to <code>RestorePhraseScreen</code> — the screen where a user <strong>types in an existing, already-funded</strong> 12-word phrase.</p>
<p>The asymmetry runs backwards. On the backup-creation screens the mnemonic is freshly generated and controls a wallet with a zero balance; a leak there is recoverable — generate a new wallet, back it up again. On the restore screen the phrase controls a wallet that already holds funds, and a BIP39 seed cannot be rotated. The protection is present where the loss is smallest and absent where it is largest.</p>
<p>The restore screen is also the <em>longest-lived</em> mnemonic surface in the app. It renders across two steps, six <code>TextInput</code>s each, with the user typing 12 words and a <code>SuggestionBar</code> rendering live BIP39 completions the whole time — minutes of exposure, versus a display screen the user glances at.</p>
<hr>
<h2>Current state of the control</h2>
<p><code>useScreenSecurity()</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-screen-security.ts#L7-L18"><code>app/hooks/use-screen-security.ts</code></a>) wraps <code>react-native-screenguard</code>:</p>
<pre><code class="language-ts">export const useScreenSecurity = (): void =&gt; {
  const { theme: { colors } } = useTheme()
  useEffect(() =&gt; {
    enableScreenSecurity(colors.black)
    return () =&gt; { disableScreenSecurity() }
  }, [colors.black])
}
</code></pre>
<p>Complete inventory of call sites at <code>8727998</code>:</p>
<pre><code class="language-console">$ grep -rn "useScreenSecurity()" app/
app/screens/settings-screen/api/api-key-secret-reveal.tsx:27
app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:34
app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx:55
</code></pre>

Screen | Sensitive material | Protected
-- | -- | --
manual-backup/backup-phrase-screen.tsx | Displays new mnemonic, 6 words/step | ✅ L55
manual-backup/view-backup-phrase-screen.tsx | Displays existing mnemonic (biometric-gated) | ✅ L34
settings-screen/api/api-key-secret-reveal.tsx | API key secret | ✅ L27
restore/restore-phrase-screen.tsx | User types a funded 12-word mnemonic | ❌ none
manual-backup/backup-phrase-confirm-screen.tsx | User types real mnemonic words from challenges route param | ❌ none
restore/cloud-restore-screen.tsx | Backup-decryption password (PasswordInput) | ❌ none
onboarding/cloud-backup-screen.tsx | Backup-encryption password | ❌ none

<p><code>RestorePhraseScreen</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx">full file</a>) imports 15 modules and calls neither <code>useScreenSecurity</code> nor anything equivalent. Words are rendered into plain <code>MnemonicWordInput</code> fields at <a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx#L196-L219">L196–L219</a>.</p>
<hr>
<h2>What the missing flag actually costs</h2>
<p><code>react-native-screenguard@2.0.0-beta5</code> is not a soft warning layer. Per its native sources:</p>
<ul>
<li><strong>Android</strong> — <code>ScreenGuardModule.java</code> calls <code>getWindow().addFlags(WindowManager.LayoutParams.FLAG_SECURE)</code>. Without it, on the restore screen:
<ul>
<li>The <strong>Recents/app-switcher thumbnail is captured and written to disk</strong> under <code>/data/system_ce/&lt;user&gt;/snapshots/</code>. Backgrounding the app mid-restore — an incoming call, a notification tap, checking the paper backup in another app — persists an image of the entered words to storage that survives the app being killed.</li>
<li><strong><code>MediaProjection</code> screen recorders capture the screen.</strong> Any installed app holding an active projection session — legitimate screen recorders, mirroring/casting tools, remote-support apps, or malware that social-engineered the projection consent dialog once — records the phrase as it is typed.</li>
<li>Third-party <strong>display mirroring</strong> to a non-secure external display shows the phrase.</li>
</ul>
</li>
<li><strong>iOS</strong> — <code>ScreenGuardImpl.mm</code> overlays a <code>UITextField</code> with <code>isSecureTextEntry</code> and registers a <code>UIApplicationUserDidTakeScreenshotNotification</code> observer, so both stills and screen recordings of the guarded layer come out blank. Without it the phrase lands in the camera roll on a screenshot, and in any active ReplayKit recording or AirPlay mirror.</li>
</ul>
<p>None of these require a compromised device or root. The recents-snapshot vector in particular fires on completely ordinary user behaviour.</p>
<hr>
<h2>Threat model</h2>
<p>The realistic chain is: a user is restoring a funded wallet onto a new or borrowed phone. During those two typing steps, any of the following captures the phrase —</p>
<ol>
<li><strong>Accidental self-capture.</strong> The user screenshots the screen to "save their place", or has a screen recorder running. Screenshots sync to Google Photos / iCloud Photos, at which point the seed is in a cloud account protected by a password, not by the seed's own security model.</li>
<li><strong>Backgrounding.</strong> Android writes the recents snapshot to disk. Anyone with brief physical access to the unlocked device, or a later forensic/ADB-backup extraction, recovers it.</li>
<li><strong>A capture-capable app already installed.</strong> Screen-recording, mirroring, and remote-support apps hold <code>MediaProjection</code>/ReplayKit permission indefinitely once granted. This is the single most common real-world path and is exactly the class <code>FLAG_SECURE</code> exists to close.</li>
<li><strong>Shoulder-surfing amplification.</strong> Not blocked by <code>FLAG_SECURE</code>, but the same long on-screen dwell time that makes recording effective makes this worse too.</li>
</ol>
<p>Outcome in every case: an attacker holding 12 words drains the wallet. There is no revocation, no rotation, no server-side control that helps.</p>
<hr>
<h2>Adjacent findings from the same review</h2>
<p>These came up while tracing the data flow. Two are real gaps; two are checks that came back <strong>clean</strong> and are recorded here so nobody re-investigates them.</p>
<h3>1. The confirm screen carries real mnemonic words in route params — and is also unprotected</h3>
<p><a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/navigation/stack-param-lists.ts#L224-L227"><code>stack-param-lists.ts</code> L224–L227</a>:</p>
<pre><code class="language-ts">selfCustodialBackupPhraseConfirm: {
  challenges: Array&lt;{ index: number; word: string }&gt;
  successMessage?: string
}
</code></pre>
<p><code>Challenge = { index, word }</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/utils.ts#L14"><code>utils.ts</code> L14</a>) holds the actual words, and <code>BackupPhraseConfirmScreen</code> has no <code>useScreenSecurity()</code>. Same for restore:</p>
<pre><code class="language-ts">selfCustodialRestorePhrase: { step: PhraseStep; words?: string[] }   // L244
</code></pre>
<p><code>useRestorePhrase.handleContinue</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/hooks/use-restore-phrase.ts#L72-L77">L72–L77</a>) puts the full 12-word array into navigation state on the step 1 → step 2 transition.</p>
<h3>2. Cleared: route params do <strong>not</strong> reach Crashlytics or Firebase Analytics</h3>
<p>Worth stating explicitly, because "mnemonic in route params" reads alarming.</p>
<ul>
<li><code>onStateChange</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/navigation/navigation-container-wrapper.tsx#L276-L288"><code>navigation-container-wrapper.tsx</code> L276–L288</a>) logs <code>screen_name</code> / <code>screen_class</code> only. No params.</li>
<li><code>reportError</code> → <code>recordAppError</code> → <code>crashlytics().recordError</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/utils/error-reporting.ts"><code>error-reporting.ts</code></a>) records the <code>Error</code> and a dedup key. The <code>restore-phrase-params-missing</code> report passes a synthetic <code>Error</code> with no param contents.</li>
<li><code>MainActivity.onCreate</code> calls <code>super.onCreate(null)</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/android/app/src/main/java/com/galoyapp/MainActivity.kt#L27-L30">L27–L30</a>), so Android does not persist and rehydrate the navigation state — the words never hit disk through that path.</li>
<li><code>selfCustodialRestorePhrase</code> is not present in the <code>linking.config.screens</code> map, so it is not reachable as a deep link with attacker-supplied <code>words</code>.</li>
</ul>
<p>So the route-param exposure is process-memory only. Still worth minimising, but it is not an exfiltration path today.</p>
<h3>3. Keyboard/IME hardening is right on Android, thinner on iOS</h3>
<p><code>MnemonicWordInput</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/components/mnemonic-word-input/mnemonic-word-input.tsx#L48-L60">L48–L60</a>) already sets <code>autoCorrect={false}</code> and <code>keyboardType="visible-password"</code>, which is the correct Android signal to keep the IME from learning the words. But <code>visible-password</code> is <strong>Android-only</strong> in React Native and is ignored on iOS, so an iOS third-party keyboard sees the phrase through the standard text-input path. Explicit <code>spellCheck={false}</code>, <code>textContentType="none"</code>, <code>autoComplete="off"</code>, and <code>importantForAutofill="no"</code> would be cheap defence in depth. (Out of scope for this issue; filing separately is fine.)</p>
<h3>4. Clipboard round-trip</h3>
<p>The backup screens expose a <strong>Copy</strong> header button and the restore screen a <strong>Paste</strong> one (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/hooks/use-restore-phrase.ts#L57-L65"><code>use-restore-phrase.ts</code> L57–L65</a>), so the mnemonic transits the system clipboard by design. <code>ApiKeySecretReveal</code> already uses <code>useClipboard(CLIPBOARD_CLEAR_MS)</code> to auto-clear; the mnemonic paths do not appear to. Separate issue, noted for completeness.</p>
<hr>
<h2>Reproduction</h2>
<p>Android, release or debug build:</p>
<ol>
<li>Navigate to <strong>Restore wallet → Recovery phrase</strong>.</li>
<li>Enter any six words on step 1 and continue.</li>
<li>Take a screenshot → succeeds, image saved to the gallery.</li>
<li>Press Recents → the phrase is legible in the app thumbnail. Kill the app; the snapshot file persists under <code>/data/system_ce/&lt;user&gt;/snapshots/</code>.</li>
<li>Start any screen recorder and repeat → the phrase is recorded.</li>
<li>Repeat steps 3–5 on <strong>Back up wallet → Manual → Recovery phrase</strong>: the screenshot is black, the recents thumbnail is black, the recording is black.</li>
</ol>
<p>Or assert the window flag directly while each screen is foregrounded:</p>
<pre><code class="language-bash">adb shell dumpsys window windows | grep -A3 'com.galoyapp' | grep -i 'FLAG_SECURE\|fl='
</code></pre>
<p>iOS: screenshot the restore screen (captures) vs. the backup phrase screen (blank).</p>
<hr>
<h2>Proposed fix</h2>
<h3>1. Add the hook to <code>RestorePhraseScreen</code></h3>
<pre><code class="language-diff">--- a/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx
+++ b/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx
@@
 import { headerRightNoGlass, noHeaderRight } from "@app/components/header-no-glass"
 import { WarningCard } from "@app/components/warning-card"
 import { SuggestionBar } from "@app/components/suggestion-bar"
+import { useScreenSecurity } from "@app/hooks/use-screen-security"
 import { useI18nContext } from "@app/i18n/i18n-react"
@@
   const navigation = useNavigation&lt;NativeStackNavigationProp&lt;RootStackParamList&gt;&gt;()
 
+  useScreenSecurity()
+
   /** Deep links and navigation-state rehydration can deliver missing or malformed params
</code></pre>
<p>Place it before the early returns so the <code>Restoring</code> and <code>Error</code> branches stay covered too.</p>
<h3>2. Extend to the other three mnemonic/password screens</h3>
<p><code>backup-phrase-confirm-screen.tsx</code>, <code>cloud-restore-screen.tsx</code>, <code>cloud-backup-screen.tsx</code>.</p>
<h3>3. ⚠️ Make the control ref-counted first — it is currently global, not per-screen</h3>
<p>This is a prerequisite, not a nice-to-have. <code>ScreenGuard.register()</code> / <code>unregister()</code> operate on the <strong>activity window</strong>, not on a component. The cleanup in <code>useScreenSecurity</code> unconditionally unregisters. <code>BackupPhraseScreen</code> pushes <code>BackupPhraseConfirmScreen</code> (<a href="https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/hooks/use-backup-phrase.ts#L56"><code>use-backup-phrase.ts</code> L56</a>) and stays mounted underneath it. If step 2 simply adds <code>useScreenSecurity()</code> to the confirm screen, then popping back to <code>BackupPhraseScreen</code> runs the confirm screen's cleanup and <strong>strips <code>FLAG_SECURE</code> from a screen that is still displaying the mnemonic</strong> — a regression introduced by the fix.</p>
<p>A module-level counter fixes it:</p>
<pre><code class="language-ts">// app/hooks/use-screen-security.ts
let activeGuards = 0

export const useScreenSecurity = (): void =&gt; {
  const { theme: { colors } } = useTheme()

  useEffect(() =&gt; {
    activeGuards += 1
    if (activeGuards === 1) enableScreenSecurity(colors.black)
    return () =&gt; {
      activeGuards -= 1
      if (activeGuards === 0) disableScreenSecurity()
    }
  }, [colors.black])
}
</code></pre>
<p>Two further details worth handling in the same change:</p>
<ul>
<li><code>enableScreenSecurity</code> is <code>async</code> and the hook does not await it, so there is a short unguarded window between mount and <code>FLAG_SECURE</code> landing. Setting the flag before the first sensitive frame paints (or rendering the inputs behind a <code>guardReady</code> gate) closes it.</li>
<li>Consider <code>useFocusEffect</code> instead of <code>useEffect</code>, so a screen left mounted lower in the stack does not hold the guard indefinitely. With ref-counting this is a refinement, not a correctness requirement.</li>
</ul>
<h3>4. Add a guard test so the next screen does not miss it</h3>
<pre><code class="language-ts">// __tests__/screen-security-coverage.test.ts
const MUST_GUARD = [
  "app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx",
  "app/screens/self-custodial/onboarding/restore/cloud-restore-screen.tsx",
  "app/screens/self-custodial/onboarding/cloud-backup-screen.tsx",
  "app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx",
  "app/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.tsx",
  "app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx",
  "app/screens/settings-screen/api/api-key-secret-reveal.tsx",
]

it.each(MUST_GUARD)("%s calls useScreenSecurity", (file) =&gt; {
  expect(readFileSync(file, "utf8")).toContain("useScreenSecurity()")
})
</code></pre>
<hr>
<h2>Acceptance criteria</h2>
<ul>
<li>[ ] <code>RestorePhraseScreen</code> calls <code>useScreenSecurity()</code>, covering the input, <code>Restoring</code>, and <code>Error</code> render branches.</li>
<li>[ ] <code>useScreenSecurity</code> is ref-counted; popping a guarded child back to a guarded parent leaves the parent guarded.</li>
<li>[ ] Android: screenshot, screen recording, and the Recents thumbnail are all blank on the restore screen (verified via <code>dumpsys window</code> showing <code>FLAG_SECURE</code>).</li>
<li>[ ] iOS: screenshot and ReplayKit recording of the restore screen come out blank.</li>
<li>[ ] <code>backup-phrase-confirm-screen</code>, <code>cloud-restore-screen</code>, and <code>cloud-backup-screen</code> are guarded.</li>
<li>[ ] A test fails if a screen on the list drops the hook.</li>
<li>[ ] Follow-up issues opened for iOS keyboard hardening (§3) and mnemonic clipboard auto-clear (§4).</li>
</ul>
<hr>
<h2>References</h2>
<ul>
<li><a href="https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_SECURE"><code>FLAG_SECURE</code> — <code>WindowManager.LayoutParams</code></a></li>
<li><a href="https://developer.android.com/reference/android/media/projection/MediaProjection"><code>MediaProjection</code> — Android Developers</a></li>
<li><a href="https://mas.owasp.org/MASVS/controls/MASVS-PLATFORM-3/">OWASP MASVS — MASVS-PLATFORM-3 (sensitive data in the UI)</a></li>
<li><a href="https://mas.owasp.org/MASTG/tests/">OWASP MASTG — Testing for Sensitive Data in Screenshots / Backgrounding</a></li>
<li><a href="https://github.com/gbumps/react-native-screenguard"><code>react-native-screenguard</code></a> — <code>android/.../ScreenGuardModule.java</code>, <code>ios/ScreenGuardImpl.mm</code></li>
<li><a href="https://reactnative.dev/docs/textinput#keyboardtype">React Native <code>TextInput.keyboardType</code></a> — <code>visible-password</code> is Android-only</li>
</ul></body></html># Restore-phrase screen has no `useScreenSecurity()` — the funded mnemonic is screenshot- and recording-capturable

**Component:** `app/screens/self-custodial/onboarding/restore/`
**Severity:** Medium–High — impact is total, irreversible loss of a funded wallet; likelihood requires an already-present capture channel on the device.
**Type:** Missing security control / inconsistent application of an existing one

---

## Summary

`useScreenSecurity()` is applied to the two screens that **display** a mnemonic and to the API-key reveal, but not to `RestorePhraseScreen` — the screen where a user **types in an existing, already-funded** 12-word phrase.

The asymmetry runs backwards. On the backup-creation screens the mnemonic is freshly generated and controls a wallet with a zero balance; a leak there is recoverable — generate a new wallet, back it up again. On the restore screen the phrase controls a wallet that already holds funds, and a BIP39 seed cannot be rotated. The protection is present where the loss is smallest and absent where it is largest.

The restore screen is also the *longest-lived* mnemonic surface in the app. It renders across two steps, six `TextInput`s each, with the user typing 12 words and a `SuggestionBar` rendering live BIP39 completions the whole time — minutes of exposure, versus a display screen the user glances at.

---

## Current state of the control

`useScreenSecurity()` ([`[app/hooks/use-screen-security.ts](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-screen-security.ts#L7-L18)`](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-screen-security.ts#L7-L18)) wraps `react-native-screenguard`:

```ts
export const useScreenSecurity = (): void => {
  const { theme: { colors } } = useTheme()
  useEffect(() => {
    enableScreenSecurity(colors.black)
    return () => { disableScreenSecurity() }
  }, [colors.black])
}
```

Complete inventory of call sites at `8727998`:

```console
$ grep -rn "useScreenSecurity()" app/
app/screens/settings-screen/api/api-key-secret-reveal.tsx:27
app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:34
app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx:55
```

| Screen | Sensitive material | Protected |
|---|---|---|
| `manual-backup/backup-phrase-screen.tsx` | Displays new mnemonic, 6 words/step | ✅ [[L55](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx#L55)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx#L55) |
| `manual-backup/view-backup-phrase-screen.tsx` | Displays existing mnemonic (biometric-gated) | ✅ [[L34](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx#L34)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx#L34) |
| `settings-screen/api/api-key-secret-reveal.tsx` | API key secret | ✅ [[L27](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/settings-screen/api/api-key-secret-reveal.tsx#L27)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/settings-screen/api/api-key-secret-reveal.tsx#L27) |
| **`restore/restore-phrase-screen.tsx`** | **User types a funded 12-word mnemonic** | ❌ **none** |
| `manual-backup/backup-phrase-confirm-screen.tsx` | User types real mnemonic words from `challenges` route param | ❌ none |
| `restore/cloud-restore-screen.tsx` | Backup-decryption password (`PasswordInput`) | ❌ none |
| `onboarding/cloud-backup-screen.tsx` | Backup-encryption password | ❌ none |

`RestorePhraseScreen` ([[full file](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx)) imports 15 modules and calls neither `useScreenSecurity` nor anything equivalent. Words are rendered into plain `MnemonicWordInput` fields at [[L196–L219](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx#L196-L219)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx#L196-L219).

---

## What the missing flag actually costs

`react-native-screenguard@2.0.0-beta5` is not a soft warning layer. Per its native sources:

- **Android** — `ScreenGuardModule.java` calls `getWindow().addFlags(WindowManager.LayoutParams.FLAG_SECURE)`. Without it, on the restore screen:
  - The **Recents/app-switcher thumbnail is captured and written to disk** under `/data/system_ce/<user>/snapshots/`. Backgrounding the app mid-restore — an incoming call, a notification tap, checking the paper backup in another app — persists an image of the entered words to storage that survives the app being killed.
  - **`MediaProjection` screen recorders capture the screen.** Any installed app holding an active projection session — legitimate screen recorders, mirroring/casting tools, remote-support apps, or malware that social-engineered the projection consent dialog once — records the phrase as it is typed.
  - Third-party **display mirroring** to a non-secure external display shows the phrase.
- **iOS** — `ScreenGuardImpl.mm` overlays a `UITextField` with `isSecureTextEntry` and registers a `UIApplicationUserDidTakeScreenshotNotification` observer, so both stills and screen recordings of the guarded layer come out blank. Without it the phrase lands in the camera roll on a screenshot, and in any active ReplayKit recording or AirPlay mirror.

None of these require a compromised device or root. The recents-snapshot vector in particular fires on completely ordinary user behaviour.

---

## Threat model

The realistic chain is: a user is restoring a funded wallet onto a new or borrowed phone. During those two typing steps, any of the following captures the phrase —

1. **Accidental self-capture.** The user screenshots the screen to "save their place", or has a screen recorder running. Screenshots sync to Google Photos / iCloud Photos, at which point the seed is in a cloud account protected by a password, not by the seed's own security model.
2. **Backgrounding.** Android writes the recents snapshot to disk. Anyone with brief physical access to the unlocked device, or a later forensic/ADB-backup extraction, recovers it.
3. **A capture-capable app already installed.** Screen-recording, mirroring, and remote-support apps hold `MediaProjection`/ReplayKit permission indefinitely once granted. This is the single most common real-world path and is exactly the class `FLAG_SECURE` exists to close.
4. **Shoulder-surfing amplification.** Not blocked by `FLAG_SECURE`, but the same long on-screen dwell time that makes recording effective makes this worse too.

Outcome in every case: an attacker holding 12 words drains the wallet. There is no revocation, no rotation, no server-side control that helps.

---

## Adjacent findings from the same review

These came up while tracing the data flow. Two are real gaps; two are checks that came back **clean** and are recorded here so nobody re-investigates them.

### 1. The confirm screen carries real mnemonic words in route params — and is also unprotected

[`stack-param-lists.ts` L224–L227](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/navigation/stack-param-lists.ts#L224-L227):

```ts
selfCustodialBackupPhraseConfirm: {
  challenges: Array<{ index: number; word: string }>
  successMessage?: string
}
```

`Challenge = { index, word }` ([`utils.ts` L14](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/utils.ts#L14)) holds the actual words, and `BackupPhraseConfirmScreen` has no `useScreenSecurity()`. Same for restore:

```ts
selfCustodialRestorePhrase: { step: PhraseStep; words?: string[] }   // L244
```

`useRestorePhrase.handleContinue` ([[L72–L77](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/hooks/use-restore-phrase.ts#L72-L77)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/hooks/use-restore-phrase.ts#L72-L77)) puts the full 12-word array into navigation state on the step 1 → step 2 transition.

### 2. Cleared: route params do **not** reach Crashlytics or Firebase Analytics

Worth stating explicitly, because "mnemonic in route params" reads alarming.

- `onStateChange` ([`navigation-container-wrapper.tsx` L276–L288](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/navigation/navigation-container-wrapper.tsx#L276-L288)) logs `screen_name` / `screen_class` only. No params.
- `reportError` → `recordAppError` → `crashlytics().recordError` ([`[error-reporting.ts](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/utils/error-reporting.ts)`](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/utils/error-reporting.ts)) records the `Error` and a dedup key. The `restore-phrase-params-missing` report passes a synthetic `Error` with no param contents.
- `MainActivity.onCreate` calls `super.onCreate(null)` ([[L27–L30](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/android/app/src/main/java/com/galoyapp/MainActivity.kt#L27-L30)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/android/app/src/main/java/com/galoyapp/MainActivity.kt#L27-L30)), so Android does not persist and rehydrate the navigation state — the words never hit disk through that path.
- `selfCustodialRestorePhrase` is not present in the `linking.config.screens` map, so it is not reachable as a deep link with attacker-supplied `words`.

So the route-param exposure is process-memory only. Still worth minimising, but it is not an exfiltration path today.

### 3. Keyboard/IME hardening is right on Android, thinner on iOS

`MnemonicWordInput` ([[L48–L60](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/components/mnemonic-word-input/mnemonic-word-input.tsx#L48-L60)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/components/mnemonic-word-input/mnemonic-word-input.tsx#L48-L60)) already sets `autoCorrect={false}` and `keyboardType="visible-password"`, which is the correct Android signal to keep the IME from learning the words. But `visible-password` is **Android-only** in React Native and is ignored on iOS, so an iOS third-party keyboard sees the phrase through the standard text-input path. Explicit `spellCheck={false}`, `textContentType="none"`, `autoComplete="off"`, and `importantForAutofill="no"` would be cheap defence in depth. (Out of scope for this issue; filing separately is fine.)

### 4. Clipboard round-trip

The backup screens expose a **Copy** header button and the restore screen a **Paste** one ([`use-restore-phrase.ts` L57–L65](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/restore/hooks/use-restore-phrase.ts#L57-L65)), so the mnemonic transits the system clipboard by design. `ApiKeySecretReveal` already uses `useClipboard(CLIPBOARD_CLEAR_MS)` to auto-clear; the mnemonic paths do not appear to. Separate issue, noted for completeness.

---

## Reproduction

Android, release or debug build:

1. Navigate to **Restore wallet → Recovery phrase**.
2. Enter any six words on step 1 and continue.
3. Take a screenshot → succeeds, image saved to the gallery.
4. Press Recents → the phrase is legible in the app thumbnail. Kill the app; the snapshot file persists under `/data/system_ce/<user>/snapshots/`.
5. Start any screen recorder and repeat → the phrase is recorded.
6. Repeat steps 3–5 on **Back up wallet → Manual → Recovery phrase**: the screenshot is black, the recents thumbnail is black, the recording is black.

Or assert the window flag directly while each screen is foregrounded:

```bash
adb shell dumpsys window windows | grep -A3 'com.galoyapp' | grep -i 'FLAG_SECURE\|fl='
```

iOS: screenshot the restore screen (captures) vs. the backup phrase screen (blank).

---

### 1. Add the hook to `RestorePhraseScreen`

```diff
--- a/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx
+++ b/app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx
@@
 import { headerRightNoGlass, noHeaderRight } from "@app/components/header-no-glass"
 import { WarningCard } from "@app/components/warning-card"
 import { SuggestionBar } from "@app/components/suggestion-bar"
+import { useScreenSecurity } from "@app/hooks/use-screen-security"
 import { useI18nContext } from "@app/i18n/i18n-react"
@@
   const navigation = useNavigation<NativeStackNavigationProp<RootStackParamList>>()
 
+  useScreenSecurity()
+
   /** Deep links and navigation-state rehydration can deliver missing or malformed params
```

Place it before the early returns so the `Restoring` and `Error` branches stay covered too.

### 2. Extend to the other three mnemonic/password screens

`backup-phrase-confirm-screen.tsx`, `cloud-restore-screen.tsx`, `cloud-backup-screen.tsx`.

### 3. ⚠️ Make the control ref-counted first — it is currently global, not per-screen

This is a prerequisite, not a nice-to-have. `ScreenGuard.register()` / `unregister()` operate on the **activity window**, not on a component. The cleanup in `useScreenSecurity` unconditionally unregisters. `BackupPhraseScreen` pushes `BackupPhraseConfirmScreen` ([`use-backup-phrase.ts` L56](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/self-custodial/onboarding/hooks/use-backup-phrase.ts#L56)) and stays mounted underneath it. If step 2 simply adds `useScreenSecurity()` to the confirm screen, then popping back to `BackupPhraseScreen` runs the confirm screen's cleanup and **strips `FLAG_SECURE` from a screen that is still displaying the mnemonic** — a regression introduced by the fix.

A module-level counter fixes it:

```ts
// app/hooks/use-screen-security.ts
let activeGuards = 0

export const useScreenSecurity = (): void => {
  const { theme: { colors } } = useTheme()

  useEffect(() => {
    activeGuards += 1
    if (activeGuards === 1) enableScreenSecurity(colors.black)
    return () => {
      activeGuards -= 1
      if (activeGuards === 0) disableScreenSecurity()
    }
  }, [colors.black])
}
```

Two further details worth handling in the same change:

- `enableScreenSecurity` is `async` and the hook does not await it, so there is a short unguarded window between mount and `FLAG_SECURE` landing. Setting the flag before the first sensitive frame paints (or rendering the inputs behind a `guardReady` gate) closes it.
- Consider `useFocusEffect` instead of `useEffect`, so a screen left mounted lower in the stack does not hold the guard indefinitely. With ref-counting this is a refinement, not a correctness requirement.

### 4. Add a guard test so the next screen does not miss it

```ts
// __tests__/screen-security-coverage.test.ts
const MUST_GUARD = [
  "app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx",
  "app/screens/self-custodial/onboarding/restore/cloud-restore-screen.tsx",
  "app/screens/self-custodial/onboarding/cloud-backup-screen.tsx",
  "app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx",
  "app/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.tsx",
  "app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx",
  "app/screens/settings-screen/api/api-key-secret-reveal.tsx",
]

it.each(MUST_GUARD)("%s calls useScreenSecurity", (file) => {
  expect(readFileSync(file, "utf8")).toContain("useScreenSecurity()")
})
```

---

## Acceptance criteria

- [ ] `RestorePhraseScreen` calls `useScreenSecurity()`, covering the input, `Restoring`, and `Error` render branches.
- [ ] `useScreenSecurity` is ref-counted; popping a guarded child back to a guarded parent leaves the parent guarded.
- [ ] Android: screenshot, screen recording, and the Recents thumbnail are all blank on the restore screen (verified via `dumpsys window` showing `FLAG_SECURE`).
- [ ] iOS: screenshot and ReplayKit recording of the restore screen come out blank.
- [ ] `backup-phrase-confirm-screen`, `cloud-restore-screen`, and `cloud-backup-screen` are guarded.
- [ ] A test fails if a screen on the list drops the hook.
- [ ] Follow-up issues opened for iOS keyboard hardening (§3) and mnemonic clipboard auto-clear (§4).

---

## References

- [`FLAG_SECURE` — `WindowManager.LayoutParams`](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_SECURE)
- [`MediaProjection` — Android Developers](https://developer.android.com/reference/android/media/projection/MediaProjection)
- [[OWASP MASVS — MASVS-PLATFORM-3 (sensitive data in the UI)](https://mas.owasp.org/MASVS/controls/MASVS-PLATFORM-3/)](https://mas.owasp.org/MASVS/controls/MASVS-PLATFORM-3/)
- [[OWASP MASTG — Testing for Sensitive Data in Screenshots / Backgrounding](https://mas.owasp.org/MASTG/tests/)](https://mas.owasp.org/MASTG/tests/)
- [`[react-native-screenguard](https://github.com/gbumps/react-native-screenguard)`](https://github.com/gbumps/react-native-screenguard) — `android/.../ScreenGuardModule.java`, `ios/ScreenGuardImpl.mm`
- [React Native `TextInput.keyboardType`](https://reactnative.dev/docs/textinput#keyboardtype) — `visible-password` is Android-only

### DIFF
--- a/__tests__/hooks/use-screen-security.spec.ts
+++ b/__tests__/hooks/use-screen-security.spec.ts
@@ -1,4 +1,4 @@
-import { renderHook } from "@testing-library/react-native"
+import { renderHook, waitFor } from "@testing-library/react-native"
 
 import { useScreenSecurity } from "@app/hooks/use-screen-security"
 
@@ -8,6 +8,11 @@ const mockDisableScreenSecurity = jest.fn()
 jest.mock("@app/utils/screen-security", () => ({
   enableScreenSecurity: (...args: string[]) => mockEnableScreenSecurity(...args),
   disableScreenSecurity: () => mockDisableScreenSecurity(),
+}))
+
+const mockReportError = jest.fn()
+jest.mock("@app/utils/error-logging", () => ({
+  reportError: (...args: readonly unknown[]) => mockReportError(...args),
 }))
 
 jest.mock("@rn-vui/themed", () => ({
@@ -35,5 +40,31 @@ describe("useScreenSecurity", () => {
     unmount()
 
     expect(mockDisableScreenSecurity).toHaveBeenCalled()
+  })
+
+  /** The enable is fire-and-forget; a native failure to install the guard must be
+   *  reported, not leaked as an unhandled rejection while the screen renders its seed
+   *  words unprotected. */
+  it("reports a rejected enable instead of leaking an unhandled rejection", async () => {
+    const failure = new Error("native failure")
+    mockEnableScreenSecurity.mockRejectedValueOnce(failure)
+
+    renderHook(() => useScreenSecurity())
+
+    await waitFor(() =>
+      expect(mockReportError).toHaveBeenCalledWith("Enable screen security", failure),
+    )
+  })
+
+  it("reports a rejected disable on unmount", async () => {
+    const failure = new Error("native failure")
+    mockDisableScreenSecurity.mockRejectedValueOnce(failure)
+
+    const { unmount } = renderHook(() => useScreenSecurity())
+    unmount()
+
+    await waitFor(() =>
+      expect(mockReportError).toHaveBeenCalledWith("Disable screen security", failure),
+    )
   })
 })
--- a/__tests__/screens/api-key-create.spec.tsx
+++ b/__tests__/screens/api-key-create.spec.tsx
@@ -69,8 +69,8 @@ jest.mock("@react-native-clipboard/clipboard", () => ({
 }))
 
 jest.mock("@app/utils/screen-security", () => ({
-  enableScreenSecurity: jest.fn(),
-  disableScreenSecurity: jest.fn(),
+  enableScreenSecurity: jest.fn(() => Promise.resolve()),
+  disableScreenSecurity: jest.fn(() => Promise.resolve()),
 }))
 
 loadLocale("en")
--- a/__tests__/screens/api-key-secret-reveal.spec.tsx
+++ b/__tests__/screens/api-key-secret-reveal.spec.tsx
@@ -42,8 +42,8 @@ jest.mock("@react-native-clipboard/clipboard", () => ({
 }))
 
 jest.mock("@app/utils/screen-security", () => ({
-  enableScreenSecurity: jest.fn(),
-  disableScreenSecurity: jest.fn(),
+  enableScreenSecurity: jest.fn(() => Promise.resolve()),
+  disableScreenSecurity: jest.fn(() => Promise.resolve()),
 }))
 
 loadLocale("en")
--- a/__tests__/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.spec.tsx
+++ b/__tests__/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.spec.tsx
@@ -117,6 +117,8 @@ describe("BackupPhraseConfirmScreen", () => {
   beforeEach(() => {
     jest.clearAllMocks()
     jest.useFakeTimers({ doNotFake: ["setImmediate"] })
+    mockEnableScreenSecurity.mockResolvedValue(undefined)
+    mockDisableScreenSecurity.mockResolvedValue(undefined)
     mockCheckpoint.mockReturnValue(null)
     mockMigrationAccountId.mockReturnValue("migration-uuid")
     mockCheckpointLoading.mockReturnValue(false)
--- a/__tests__/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.spec.tsx
+++ b/__tests__/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.spec.tsx
@@ -99,6 +99,15 @@ jest.mock("@app/utils/error-logging", () => ({
   reportError: (...args: readonly unknown[]) => mockReportError(...args),
 }))
 
+const mockEnableScreenSecurity = jest.fn()
+const mockDisableScreenSecurity = jest.fn()
+jest.mock("@app/utils/screen-security", () => ({
+  enableScreenSecurity: (...args: readonly unknown[]) =>
+    mockEnableScreenSecurity(...args),
+  disableScreenSecurity: (...args: readonly unknown[]) =>
+    mockDisableScreenSecurity(...args),
+}))
+
 loadLocale("en")
 const LL = i18nObject("en")
 
@@ -565,6 +574,34 @@ describe("BackupPhraseConfirmScreen", () => {
         route: "accountMigrationBalancesOverview",
         accountId: "migration-uuid",
       },
+    })
+  })
+
+  /** The screen shows individual mnemonic words with their positions, so it must carry
+   *  the same screenshot/screen-recording protection as the backup-phrase screen. */
+  describe("screen security", () => {
+    it("enables screenshot protection on mount", async () => {
+      render(
+        <ContextForScreen>
+          <BackupPhraseConfirmScreen />
+        </ContextForScreen>,
+      )
+      await flushEffects()
+
+      expect(mockEnableScreenSecurity).toHaveBeenCalledTimes(1)
+    })
+
+    it("disables screenshot protection on unmount", async () => {
+      const { unmount } = render(
+        <ContextForScreen>
+          <BackupPhraseConfirmScreen />
+        </ContextForScreen>,
+      )
+      await flushEffects()
+
+      unmount()
+
+      expect(mockDisableScreenSecurity).toHaveBeenCalledTimes(1)
     })
   })
 })
--- a/__tests__/screens/self-custodial/onboarding/restore/restore-phrase-screen.spec.tsx
+++ b/__tests__/screens/self-custodial/onboarding/restore/restore-phrase-screen.spec.tsx
@@ -32,6 +32,15 @@ jest.mock("@app/utils/error-logging", () => ({
   reportError: (...args: readonly unknown[]) => mockReportError(...args),
 }))
 
+const mockEnableScreenSecurity = jest.fn()
+const mockDisableScreenSecurity = jest.fn()
+jest.mock("@app/utils/screen-security", () => ({
+  enableScreenSecurity: (...args: readonly unknown[]) =>
+    mockEnableScreenSecurity(...args),
+  disableScreenSecurity: (...args: readonly unknown[]) =>
+    mockDisableScreenSecurity(...args),
+}))
+
 type MnemonicWordInputProps = {
   index: number
   value: string
@@ -533,6 +542,38 @@ describe("RestorePhraseScreen", () => {
       headerRight: undefined,
       // eslint-disable-next-line camelcase -- name dictated by @react-navigation/native-stack
       unstable_headerRightItems: undefined,
+    })
+  })
+
+  /** The screen shows a funded wallet's mnemonic, so it must carry the same
+   *  screenshot/screen-recording protection as the backup-creation screen. */
+  describe("screen security", () => {
+    it("enables screenshot protection on mount", async () => {
+      renderScreen()
+      await flushEffects()
+
+      expect(mockEnableScreenSecurity).toHaveBeenCalledTimes(1)
+    })
+
+    it("keeps screenshot protection active on the restoring and error states", async () => {
+      for (const status of ["restoring", "error"]) {
+        jest.clearAllMocks()
+        mockUseRestorePhrase.mockReturnValue({ ...defaultHookReturn, status })
+
+        renderScreen()
+        await flushEffects()
+
+        expect(mockEnableScreenSecurity).toHaveBeenCalledTimes(1)
+      }
+    })
+
+    it("disables screenshot protection on unmount", async () => {
+      const { unmount } = renderScreen()
+      await flushEffects()
+
+      unmount()
+
+      expect(mockDisableScreenSecurity).toHaveBeenCalledTimes(1)
     })
   })
 })
--- a/__tests__/screens/self-custodial/onboarding/restore/restore-phrase-screen.spec.tsx
+++ b/__tests__/screens/self-custodial/onboarding/restore/restore-phrase-screen.spec.tsx
@@ -130,6 +130,8 @@ const renderScreen = () =>
 describe("RestorePhraseScreen", () => {
   beforeEach(() => {
     jest.clearAllMocks()
+    mockEnableScreenSecurity.mockResolvedValue(undefined)
+    mockDisableScreenSecurity.mockResolvedValue(undefined)
     mockRouteParams = { step: 2, words: [...validStep2Words] }
     mockUseRestorePhrase.mockReturnValue(defaultHookReturn)
   })
--- a/__tests__/utils/screen-security.spec.ts
+++ b/__tests__/utils/screen-security.spec.ts
@@ -8,6 +8,11 @@ jest.mock("react-native-screenguard", () => ({
   unregister: (...args: readonly unknown[]) => mockUnregister(...args),
 }))
 
+const mockReportError = jest.fn()
+jest.mock("@app/utils/error-logging", () => ({
+  reportError: (...args: readonly unknown[]) => mockReportError(...args),
+}))
+
 /** The util keeps its reference count and call queue in module state, so each test
  *  loads a fresh copy. */
 const loadModule = (): typeof import("@app/utils/screen-security") => {
@@ -23,9 +28,17 @@ const loadModule = (): typeof import("@app/utils/screen-security") => {
 describe("screen-security", () => {
   beforeEach(() => {
     jest.clearAllMocks()
+    // Fake timers keep a retry scheduled by a rejected enable inert unless the test
+    // advances the clock — otherwise a stray timer from one isolated module could
+    // fire mid-test in another and pollute the call counts.
+    jest.useFakeTimers()
     mockInitSettings.mockResolvedValue(undefined)
     mockRegister.mockResolvedValue(undefined)
     mockUnregister.mockResolvedValue(undefined)
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
   })
 
   it("initializes and registers with the given background color on first enable", async () => {
@@ -160,5 +173,93 @@ describe("screen-security", () => {
 
     expect(mockRegister).toHaveBeenCalledTimes(2)
     expect(mockUnregister).toHaveBeenCalledTimes(1)
+  })
+
+  /** A rejected register is not the only way initSettings/register can fail: if
+   *  initialization itself rejects, no guard was installed and the next enable must
+   *  start over from initSettings. */
+  it("starts over from initSettings on the next enable when initialization fails", async () => {
+    const { enableScreenSecurity } = loadModule()
+    mockInitSettings.mockRejectedValueOnce(new Error("native failure"))
+
+    await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+    expect(mockRegister).not.toHaveBeenCalled()
+
+    await enableScreenSecurity("#000000")
+
+    expect(mockInitSettings).toHaveBeenCalledTimes(2)
+    expect(mockRegister).toHaveBeenCalledTimes(1)
+  })
+
+  /** A rejected unregister leaves the native state unknown, and the call-time
+   *  activeScreens gate means no later disable ever reaches the queue to retry it.
+   *  If `registered` stayed true, every later enable would skip the register and the
+   *  JS and native states would stay desynced for the rest of the process — so the
+   *  next enable must re-register. The rejection still propagates so the hook can
+   *  report it. */
+  it("re-registers on the next enable after a rejected unregister", async () => {
+    const { enableScreenSecurity, disableScreenSecurity } = loadModule()
+    mockUnregister.mockRejectedValueOnce(new Error("native failure"))
+
+    await enableScreenSecurity("#000000")
+    await expect(disableScreenSecurity()).rejects.toThrow("native failure")
+
+    await enableScreenSecurity("#000000")
+
+    expect(mockRegister).toHaveBeenCalledTimes(2)
+  })
+
+  // Must match ENABLE_RETRY_DELAY_MS in the util; a change there should fail here.
+  const RETRY_DELAY_MS = 10_000
+  // Must match ENABLE_RETRY_LIMIT in the util.
+  const RETRY_LIMIT = 3
+
+  describe("enable retry", () => {
+    it("retries a rejected registration while a protected screen is still mounted", async () => {
+      const { enableScreenSecurity } = loadModule()
+      mockRegister.mockRejectedValueOnce(new Error("native failure"))
+
+      await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+      expect(mockRegister).toHaveBeenCalledTimes(1)
+
+      await jest.advanceTimersByTimeAsync(RETRY_DELAY_MS)
+
+      expect(mockRegister).toHaveBeenCalledTimes(2)
+    })
+
+    it("stops retrying once the guard is registered", async () => {
+      const { enableScreenSecurity } = loadModule()
+      mockRegister.mockRejectedValueOnce(new Error("native failure"))
+
+      await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+      await jest.advanceTimersByTimeAsync(RETRY_DELAY_MS)
+      expect(mockRegister).toHaveBeenCalledTimes(2)
+
+      await jest.advanceTimersByTimeAsync(RETRY_DELAY_MS * (RETRY_LIMIT + 1))
+      expect(mockRegister).toHaveBeenCalledTimes(2)
+    })
+
+    it("reports each failed retry and gives up after a bounded number", async () => {
+      const { enableScreenSecurity } = loadModule()
+      mockRegister.mockRejectedValue(new Error("native failure"))
+
+      await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+      await jest.advanceTimersByTimeAsync(RETRY_DELAY_MS * (RETRY_LIMIT + 2))
+
+      expect(mockRegister).toHaveBeenCalledTimes(1 + RETRY_LIMIT)
+      expect(mockReportError).toHaveBeenCalledTimes(RETRY_LIMIT)
+    })
+
+    it("does not retry once the last protected screen has unmounted", async () => {
+      const { enableScreenSecurity, disableScreenSecurity } = loadModule()
+      mockRegister.mockRejectedValueOnce(new Error("native failure"))
+
+      await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+      await disableScreenSecurity()
+
+      await jest.advanceTimersByTimeAsync(RETRY_DELAY_MS * (RETRY_LIMIT + 2))
+
+      expect(mockRegister).toHaveBeenCalledTimes(1)
+    })
   })
 })
--- a/__tests__/utils/screen-security.spec.ts
+++ b/__tests__/utils/screen-security.spec.ts
@@ -20,11 +20,6 @@ const loadModule = (): typeof import("@app/utils/screen-security") => {
   return mod
 }
 
-const flushQueue = () =>
-  new Promise<void>((resolve) => {
-    setImmediate(resolve)
-  })
-
 describe("screen-security", () => {
   beforeEach(() => {
     jest.clearAllMocks()
@@ -102,14 +97,68 @@ describe("screen-security", () => {
     expect(mockUnregister).not.toHaveBeenCalled()
   })
 
-  it("keeps processing the queue after a native call rejects", async () => {
+  /** Registration state is tracked separately from the screen count: a rejected
+   *  register must not leave the module claiming protection that was never
+   *  installed. */
+  it("does not unregister after a rejected register, since no guard was installed", async () => {
     const { enableScreenSecurity, disableScreenSecurity } = loadModule()
     mockRegister.mockRejectedValueOnce(new Error("native failure"))
 
     await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
     await disableScreenSecurity()
-    await flushQueue()
 
+    expect(mockUnregister).not.toHaveBeenCalled()
+  })
+
+  it("retries registration on the next enable after a rejected register", async () => {
+    const { enableScreenSecurity, disableScreenSecurity } = loadModule()
+    mockRegister.mockRejectedValueOnce(new Error("native failure"))
+
+    await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+    await disableScreenSecurity()
+
+    await enableScreenSecurity("#000000")
+    expect(mockRegister).toHaveBeenCalledTimes(2)
+  })
+
+  /** Finding: the phrase screen's register rejects, the confirm screen is pushed on
+   *  top, and its enable must retry the registration rather than early-returning on a
+   *  count that claims protection which was never installed. */
+  it("registers for a screen pushed after another screen's registration failed", async () => {
+    const { enableScreenSecurity } = loadModule()
+    mockRegister.mockRejectedValueOnce(new Error("native failure"))
+
+    await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+    await enableScreenSecurity("#000000")
+
+    expect(mockRegister).toHaveBeenCalledTimes(2)
+  })
+
+  /** The replace path unmounts one protected screen as another mounts; the queued
+   *  disable sees the arriving screen and skips the unregister/register churn. */
+  it("does not churn the guard when one protected screen replaces another", async () => {
+    const { enableScreenSecurity, disableScreenSecurity } = loadModule()
+
+    const first = enableScreenSecurity("#000000")
+    const leaving = disableScreenSecurity()
+    const arriving = enableScreenSecurity("#000000")
+    await Promise.all([first, leaving, arriving])
+
+    expect(mockRegister).toHaveBeenCalledTimes(1)
+    expect(mockUnregister).not.toHaveBeenCalled()
+  })
+
+  it("keeps processing the queue after a native call rejects", async () => {
+    const { enableScreenSecurity, disableScreenSecurity } = loadModule()
+    mockRegister.mockRejectedValueOnce(new Error("native failure"))
+
+    await expect(enableScreenSecurity("#000000")).rejects.toThrow("native failure")
+
+    await enableScreenSecurity("#000000")
+    await disableScreenSecurity()
+    await disableScreenSecurity()
+
+    expect(mockRegister).toHaveBeenCalledTimes(2)
     expect(mockUnregister).toHaveBeenCalledTimes(1)
   })
 })
--- a/__tests__/utils/screen-security.spec.ts
+++ b/__tests__/utils/screen-security.spec.ts
@@ -1,0 +1,115 @@
+const mockInitSettings = jest.fn()
+const mockRegister = jest.fn()
+const mockUnregister = jest.fn()
+
+jest.mock("react-native-screenguard", () => ({
+  initSettings: (...args: readonly unknown[]) => mockInitSettings(...args),
+  register: (...args: readonly unknown[]) => mockRegister(...args),
+  unregister: (...args: readonly unknown[]) => mockUnregister(...args),
+}))
+
+/** The util keeps its reference count and call queue in module state, so each test
+ *  loads a fresh copy. */
+const loadModule = (): typeof import("@app/utils/screen-security") => {
+  let mod: typeof import("@app/utils/screen-security") | undefined
+  jest.isolateModules(() => {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    mod = require("@app/utils/screen-security")
+  })
+  if (!mod) throw new Error("failed to load screen-security module")
+  return mod
+}
+
+const flushQueue = () =>
+  new Promise<void>((resolve) => {
+    setImmediate(resolve)
+  })
+
+describe("screen-security", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockInitSettings.mockResolvedValue(undefined)
+    mockRegister.mockResolvedValue(undefined)
+    mockUnregister.mockResolvedValue(undefined)
+  })
+
+  it("initializes and registers with the given background color on first enable", async () => {
+    const { enableScreenSecurity } = loadModule()
+
+    await enableScreenSecurity("#000000")
+
+    expect(mockInitSettings).toHaveBeenCalledTimes(1)
+    expect(mockRegister).toHaveBeenCalledWith({ backgroundColor: "#000000" })
+  })
+
+  it("unregisters when the last protected screen disables", async () => {
+    const { enableScreenSecurity, disableScreenSecurity } = loadModule()
+
+    await enableScreenSecurity("#000000")
+    await disableScreenSecurity()
+
+    expect(mockUnregister).toHaveBeenCalledTimes(1)
+  })
+
+  /** Back-navigation scenario: the confirm screen is pushed on top of the phrase
+   *  screen; its unmount must not tear down the guard for the phrase screen still
+   *  showing the seed words underneath. */
+  it("keeps the guard registered when a stacked screen unmounts", async () => {
+    const { enableScreenSecurity, disableScreenSecurity } = loadModule()
+
+    await enableScreenSecurity("#000000") // phrase screen mounts
+    await enableScreenSecurity("#000000") // confirm screen pushed on top
+    await disableScreenSecurity() // confirm screen unmounts on back-navigation
+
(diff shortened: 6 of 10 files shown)


grimen (changes requested):
## Findings

**1. Adding the hook to the confirm screen removes protection from the phrase screen on back-navigation (blocking).**

[`useScreenSecurity`](https://github.com/blinkbitcoin/blink-mobile/blob/ac32d95cdc28017a99fb57d86f8a705db59d9924/app/hooks/use-screen-security.ts) is not reference-counted: its cleanup calls `ScreenGuard.unregister()` unconditionally, and the underlying calls are global rather than per-screen. The confirm screen is the only screen carrying this hook that gets pushed on top of another screen that also carries it, so its unmount tears down the guard for a screen that is still displaying the seed words.

*Example:* On step 2 of [`BackupPhraseScreen`](https://github.com/blinkbitcoin/blink-mobile/blob/ac32d95cdc28017a99fb57d86f8a705db59d9924/app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx#L54) the user taps continue, then the header back arrow on [`backup-phrase-confirm-screen.tsx#L64`](https://github.com/blinkbitcoin/blink-mobile/blob/ac32d95cdc28017a99fb57d86f8a705db59d9924/app/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.tsx#L64); the phrase screen becomes visible again with words 7–12 rendered and no screen guard, because its own effect has `[colors.black]` deps and never re-runs.

*Proposal:* One option could be to make the guard reference-counted in [`app/utils/screen-security.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/ac32d95cdc28017a99fb57d86f8a705db59d9924/app/utils/screen-security.ts) — a module-level counter that only unregisters at zero — rather than registering and unregistering per screen.

**2. The un-awaited async enable and disable can interleave and leave a handoff unprotected (blocking).**

`enableScreenSecurity` awaits `ScreenGuard.initSettings()` before `register()`, while `disableScreenSecurity` is a single `unregister()`, and the hook awaits neither. When one protected screen replaces another, both run in the same commit and the shorter disable can resolve after the longer enable.

*Example:* The invalid-params path calls `navigation.replace("selfCustodialBackupPhrase", …)` at [`backup-phrase-confirm-screen.tsx#L58`](https://github.com/blinkbitcoin/blink-mobile/blob/ac32d95cdc28017a99fb57d86f8a705db59d9924/app/screens/self-custodial/onboarding/manual-backup/backup-phrase-confirm-screen.tsx#L58), so `unregister()` (one tick) can land after `register()` (two ticks) and the freshly mounted phrase screen ends up unregistered even though its effect ran.

*Proposal:* A reference-counted wrapper that serializes the native calls would cover this together with finding 1.

## Test Coverage Gaps

- No test covers the back-navigation path from the confirm screen to the still-mounted phrase screen, which is where finding 1 shows up.

grimen (changes requested):
## Findings

**1. A failed registration leaves the counter claiming the guard is active (blocking).**
`activeScreens += 1` happens before the native work in `enableScreenSecurity`, and the rejection is never rolled back, so the count says a screen is protected when nothing was registered.
*Example:* `initSettings()`/`register()` rejects on the backup-phrase screen; the count stays at 1 with no guard installed, and pushing the confirm screen early-returns because the count is now 2 — so the confirm screen renders the seed words with screenshots fully allowed until the count drains back to 0.
*Proposal:* one option could be wrapping the enqueue in `try`/`catch` and decrementing on failure, or tracking "registered" separately from "requested" so the next mount retries.

**2. The rejected promise from `enableScreenSecurity` has no consumer (blocking).**
`useScreenSecurity` calls it fire-and-forget, so a native failure to install the screenshot guard is never reported anywhere.
*Example:* when `register` throws, the only trace is a "Possible Unhandled Promise Rejection" log — the new spec's `rejects.toThrow(...)` assertion confirms the rejection escapes — and the seed-phrase screen renders unprotected with nothing surfacing the failure.
*Proposal:* since the point of the PR is enforcing this control, a `.catch(reportError)` at the hook would make the failure visible.

## Test Coverage Gaps

- No test covers the state left behind after a rejected `register` — the spec asserts the rejection but not that a subsequent `enableScreenSecurity` still attempts registration.

grimen (changes requested):
## Findings

**1. Rejected unregister wedges the guard permanently (blocking).**
When `ScreenGuard.unregister()` rejects, `registered` stays `true` while `activeScreens` is 0, and nothing ever retries: every later enable early-returns on `if (registered)` and every later disable early-returns on `activeScreens === 0`. The rejected-register case has a retry and a spec; the rejected-unregister case has neither.
*Example:* the last protected screen unmounts and the library rejects (its `_logError` returns a rejected promise when `deactivateShield` throws), so either `FLAG_SECURE` stays on app-wide, or the next seed-phrase screen makes zero native calls and renders the mnemonic with screenshots enabled for the rest of the process.
*Proposal:* one option is to reset `registered` to `false` when unregister rejects, so the next enable re-registers, and add a spec for that path.

**2. A failed enable fails open silently (blocking).**
When the enable call rejects on mount, the error only goes to Crashlytics while the still-mounted screen keeps rendering the mnemonic unprotected; the util's retry only fires on the next screen's enable, and the screen has no status it could gate rendering on.
*Example:* `initSettings`/`register` rejects on mount of the backup-phrase screen and the user writes down all 12 words with screenshots and screen recording fully enabled.
*Proposal:* if report-and-continue is intentional a comment would help; otherwise the hook could expose the failure so the screen can block rendering or retry.

**3. Protection is still a per-screen opt-in.**
Each sensitive screen must remember to call `useScreenSecurity()`, and this PR is itself the second catch-up round for screens forgotten when their siblings got the hook.
*Example:* the emergency-recovery flow in PR #4108 adds new mnemonic-bearing screens; if they ship without the hook, seed words are capturable again until someone notices.
*Proposal:* a centrally checked set of secure route names could work — `NavigationContainerWrapper` already computes the active route name in `onStateChange`, and this PR's refcounted guard makes central enable/disable calls safe.

## Test Coverage Gaps

- No spec covers a rejected `unregister()` — the stuck-`registered` state in finding 1 is untested.

grimen (changes requested):
## Finding

**Screen security still fails open before and after a failed registration (blocking).**

The retry and error-reporting changes improve recovery, but useScreenSecurity still returns no readiness state. A protected route therefore paints its mnemonic or secret before the passive effect runs and while initSettings/register is pending. If the first attempt rejects, the same sensitive UI remains visible during the 10-second retry interval; after the bounded retries are exhausted it remains visible indefinitely. A later successful retry also cannot undo screenshots or recording frames captured before registration completed.

## Proposed implementation

Please make protection a fail-closed gate rather than a fire-and-forget side effect:

1. Have the screen-security manager expose an idempotent lease API. acquire(backgroundColor) should register one consumer and return a lease containing a ready promise plus an idempotent release method. Concurrent leases should share the serialized in-flight registration. The ready promise should resolve only after ScreenGuard.register succeeds, including any bounded retries, and reject only after retries are exhausted. Releasing the last lease should cancel pending retries and enqueue unregister.
2. Have useScreenSecurity own one lease for its mount lifetime and expose activating, active, or failed state. Ignore late promise settlement after unmount, report failures, and release exactly once so retrying cannot inflate the active-screen count.
3. Add a ScreenSecurityGate component that renders only a neutral placeholder while activating and a non-sensitive Retry/Back error state after failure. Mount the sensitive screen subtree only in the active state. This must include header actions such as Copy/Share, not only the visible mnemonic fields.
4. Wrap each protected route content with that gate. Moving the native call to useLayoutEffect alone is not sufficient because registration is asynchronous.

Regression coverage should use a deferred enable promise and assert that mnemonic/secret text and sensitive header actions are absent while registration is pending, appear only after resolution, and remain absent after rejection or exhausted retries. Keep the existing overlap, replacement, and rejected-unregister tests for the manager.

esaugomez31 (approved):
Approved — but this must be merged together with #4167, not on its own.

The last review here (the fail-closed gate: lease API, readiness state, `ScreenSecurityGate`, wrapped routes) was resolved in #4167, which is stacked on this branch. None of it is in this branch, so merging this alone leaves that review unaddressed: `useScreenSecurity()` still runs from a passive effect, so a protected screen paints its mnemonic before registration starts and stays visible while it is pending.

What this PR does deliver stands on its own and is correct:

- `useScreenSecurity()` on the restore and backup-confirm screens, which is the gap blinkbitcoin/blink-wip#1065 reports
- a reference-counted guard, so leaving the confirm screen no longer unregisters the guard for the phrase screen still mounted beneath it
- serialized native calls, so a short unregister cannot resolve after a longer register and tear down a guard that was just installed
- bounded retries on a failed registration, and the failure reported instead of swallowed
- recovery from a rejected unregister, which used to wedge the guard for the rest of the process

Verified locally on `26535ecb6`: tsc clean, 17 suites / 239 tests green. CI is green on Tests, Check Code and Android unit tests; the `Android` and `iOS` jobs fail in ~23s across every PR in the repo right now, so they are not from this branch.

PR #4139, opened 2026-08-10
fix(security): store the session token only in the secure key store

kngako:
The session auth token was persisted twice: in the secure key store (session profiles) and in plaintext AsyncStorage via persistentState.galoyAuthToken — the latter being the live credential sent with every GraphQL request.

- KeyStoreWrapper: dedicated activeAuthToken key with get/set/remove
- persistent-state: strip the token from the AsyncStorage blob on save, hydrate it from the key store on load; a legacy blob token is moved to the key store on first launch so existing sessions survive
- use-app-config: saveToken/saveTokenAndInstance mirror the token to the key store (empty token removes the key)
- use-logout: clear the active-token key on full logout and when removing the profile backing the active session

---

# Session JWT is persisted in plaintext AsyncStorage alongside the encrypted Keystore copy

**Component:** `app/store/persistent-state/`, `app/hooks/use-app-config.ts`, `app/utils/storage/`
**Severity:** Medium — full account takeover if the plaintext store is read, but reaching it needs root, a physical extraction, or a forensic toolchain. The quarantine sub-finding below raises it, because that copy survives logout.
**Type:** Sensitive data stored insecurely / duplicated source of truth

---

## Summary

The Galoy session JWT lives in two places at once:

- **Encrypted** — inside `sessionProfiles` in `react-native-secure-key-store`, which is AES-256-GCM `EncryptedSharedPreferences` on Android and a Keychain generic password on iOS.
- **Plaintext** — as `persistentState.galoyAuthToken`, `JSON.stringify`'d into the AsyncStorage key `persistentState`.

Every custodial login writes both. The second write buys nothing that the first does not already provide, and it undoes the protection the first one was chosen for.

**One correction to the framing:** the plaintext copy is not the redundant one — it is the **authoritative** one. `useAppConfig().token` reads `persistentState.galoyAuthToken` and that value is what Apollo attaches as the `Bearer` header on every request. The Keystore copy is a secondary list used for multi-account switching. So the fix is not "delete the AsyncStorage copy"; it is "invert which store is the source of truth," and the write ordering in `saveProfile` has to change with it (see [[Why this is not a one-line deletion](https://github.com/blinkbitcoin/blink-mobile/pull/4139#why-this-is-not-a-one-line-deletion)](#why-this-is-not-a-one-line-deletion)).

---

## Evidence

**The plaintext write path.** `saveToken` puts the JWT into React state ([`use-app-config.ts` L31–L43](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-app-config.ts#L31-L43)):

```ts
const saveToken = useCallback(async (token: string) => {
  updateState((state) => (state ? { ...state, galoyAuthToken: token } : undefined))
}, [updateState])
```

`PersistentStateProvider` flushes that state on every change ([`index.tsx` L45–L52, L74–L78](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/store/persistent-state/index.tsx#L45-L78)):

```ts
const savePersistentState = async (state: PersistentState): Promise<void> => {
  try { await saveJson(PERSISTENT_STATE_KEY, state) } catch (err) { ... }
}
```

and `saveJson` is a bare `AsyncStorage.setItem` ([`storage.ts` L37–L39](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/utils/storage/storage.ts#L37-L39)):

```ts
export const saveJson = async (key: string, value: unknown): Promise<void> => {
  await AsyncStorage.setItem(key, JSON.stringify(value))
}
```

`galoyAuthToken` is a first-class field of every schema version and of the default state ([`state-migrations.ts` L287–L291](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/store/persistent-state/state-migrations.ts#L287-L291)). It has been there since at least `8a3573a1` (2023-03-16).

**The encrypted write path.** The same login also stores the token inside a `ProfileProps` record ([`index.types.d.ts` L1–L14](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/screens/settings-screen/account/multi-account/index.types.d.ts#L1-L14) — the type carries `token: string`) via [`secureStorage.ts` L109–L119](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/utils/storage/secureStorage.ts#L109-L119):

```ts
public static async saveSessionProfiles(profiles: ProfileProps[]): Promise<boolean> {
  const serialized = JSON.stringify(profiles)
  await RNSecureKeyStore.set(KeyStoreWrapper.SESSION_PROFILES, serialized, {
    accessible: ACCESSIBLE.ALWAYS_THIS_DEVICE_ONLY,
  })
  ...
}
```

Both writes happen inside one function ([`use-save-session-profile.ts` L88–L127](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-save-session-profile.ts#L88-L127)) — `await saveToken(token)` at L92, `KeyStoreWrapper.saveSessionProfiles(...)` at L117. Every login route funnels through it: phone (`phone-login-validation.tsx:234,267`), email (`email-login-validate.tsx:63`), Telegram (`telegram-auth.ts:121`), device account (`use-create-device-account.ts:91`).

---

## What the two stores actually give you

**Android.** `react-native-secure-key-store@2.0.10` is a thin wrapper over AndroidX Security ([`[RNSecureKeyStoreModule.java](https://github.com/pradeep1991singh/react-native-secure-key-store/blob/master/android/src/main/java/com/reactlibrary/securekeystore/RNSecureKeyStoreModule.java)`](https://github.com/pradeep1991singh/react-native-secure-key-store/blob/master/android/src/main/java/com/reactlibrary/securekeystore/RNSecureKeyStoreModule.java)):

```java
return EncryptedSharedPreferences.create(
  "secret_shared_prefs",
  MasterKeys.getOrCreate(MasterKeys.AES256_GCM_SPEC),
  reactContext,
  EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
  EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
);
```

The master key lives in the AndroidKeyStore and is non-exportable — hardware-backed where a TEE/StrongBox is present. Lifting `secret_shared_prefs` off the device yields ciphertext.

AsyncStorage writes to an unencrypted SQLite database in the app's private `databases/` directory (`RKStorage` on the 2.x line). Lifting that file yields the JWT as a readable string.

So on Android the delta is total: one copy resists offline extraction, the other does not.

**iOS.** The Keystore copy is a `kSecClassGenericPassword` Keychain item; the AsyncStorage copy is a file in `Library/Application Support/RCTAsyncLocalStorage_V1/`. Both sit behind iOS Data Protection class keys, so the gap is narrower than on Android — but the Keychain item is still the one with an explicit accessibility class and per-item protection, and it is the store a reviewer or scanner expects a bearer token to be in.

### Threat model

The plaintext copy is what an attacker gets from: a rooted or jailbroken device; malware with root; a physical extraction or forensic toolchain (Cellebrite/GrayKey-class); a custom-recovery data pull; or any future bug that exposes the app's private directory. In every one of those the Keystore copy holds and the AsyncStorage copy does not.

A stolen JWT is full account access — balance, transaction history, and the ability to spend, for as long as the token is valid.

---

## Checks that came back clean

Recording these so nobody re-runs them.

- **Android backups are off.** `android:allowBackup="false"` ([`AndroidManifest.xml` L40](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/android/app/src/main/AndroidManifest.xml#L40)), so the token is not in `adb backup` or Auto Backup / Google Drive backup.
- **iOS backups are excluded by default.** `RNCAsyncStorage.mm` sets `NSURLIsExcludedFromBackupKey` on its storage directory unless the `RCTAsyncStorageExcludeFromBackup` Info.plist key says otherwise, and the default is `@YES`. `ios/GaloyApp/Info.plist` does not override it. The token is not in iCloud/iTunes backups.
- **The token is not logged or reported.** No `console.log` of the token or of `persistentState`; `recordAppError` records `Error` objects and dedup keys only; `logScreenView` sends route names only.
- **The developer screen is dev-only.** Its "Copy access token" / "Share access token" buttons are behind `__DEV__` at [`root-navigator.tsx` L184–L186](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/navigation/root-navigator.tsx#L184-L186) and are not in release builds.

---

## Sub-finding: quarantined state copies the token to keys nothing ever deletes

This one is worse than the main finding and should probably be fixed first.

When a persisted-state migration fails, the raw blob is written to a fresh, timestamped AsyncStorage key ([`index.tsx` L16–L27](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/store/persistent-state/index.tsx#L16-L27)):

```ts
const PERSISTENT_STATE_QUARANTINE_PREFIX = "persistentStateQuarantine"

const quarantineRawState = async (rawData: unknown): Promise<void> => {
  const key = `${PERSISTENT_STATE_QUARANTINE_PREFIX}.${Date.now()}`
  const ok = await saveString(key, JSON.stringify(rawData))
  ...
}
```

`rawData` is the whole prior state, `galoyAuthToken` included, and it is stringified with no redaction. The existing test asserts exactly that round-trip (`__tests__/store/persistent-state/index.spec.tsx:280` — `expect(JSON.parse(payload)).toEqual(corruptedState3)`).

Then:

```console
$ grep -rn "persistentStateQuarantine\|QUARANTINE" app/ __tests__/
app/store/persistent-state/index.tsx:17:  const PERSISTENT_STATE_QUARANTINE_PREFIX = "persistentStateQuarantine"
app/store/persistent-state/index.tsx:20:  const key = `${PERSISTENT_STATE_QUARANTINE_PREFIX}.${Date.now()}`
__tests__/store/persistent-state/index.spec.tsx:280: expect(key).toMatch(/^persistentStateQuarantine\.\d+$/)
```

Nothing reads them. Nothing deletes them. `resetState()` rewrites only the `persistentState` key ([[L95–L98](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/store/persistent-state/index.tsx#L95-L98)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/store/persistent-state/index.tsx#L95-L98)), and `useLogout` clears `SCHEMA_VERSION_KEY` plus Keystore entries ([`use-logout.ts` L44–L53](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-logout.ts#L44-L53)) — neither touches the quarantine prefix.

Result: one failed migration leaves a plaintext JWT in AsyncStorage that **survives logout, survives account switching, and accumulates one new key per failure, forever.** Whatever bound the live token's lifetime does not apply to these.

## Sub-finding: overwriting is not erasing

`savePersistentState` overwrites the JSON value in place. SQLite does not zero freed pages, and AsyncStorage never issues a `VACUUM`, so the previous value — including a logged-out token — can remain recoverable in the database file until the pages are reused. Not a headline on its own; it does mean "logout clears it" is weaker than it looks.

## Note: the `accessible` option is a no-op on Android

`saveSessionProfiles` passes `accessible: ACCESSIBLE.ALWAYS_THIS_DEVICE_ONLY`, but `RNSecureKeyStoreModule.set()` never reads its `options` argument — the flag only takes effect on iOS. There, `ALWAYS_THIS_DEVICE_ONLY` maps to `kSecAttrAccessibleAlwaysThisDeviceOnly`, which Apple deprecated; `AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY` is the current equivalent and keeps the item out of backups just the same. Worth correcting while this code is open.

Separately, `react-native-secure-key-store` has not shipped a release in years. The repo already depends on `react-native-keychain@^10` (used for device-account credentials and the iOS mnemonic backup), so consolidating onto it is a natural follow-up.

---

## Reproduction

Android, release build, logged in:

```bash
# 1. Plaintext copy — readable string
adb shell "run-as com.galoyapp ls databases/"          # debuggable builds; on release use a rooted device
adb shell "su -c sqlite3 /data/data/com.galoyapp/databases/RKStorage \
  \"select value from catalystLocalStorage where key='persistentState';\""
# → {"schemaVersion":15,"galoyInstance":{"id":"Main"},"galoyAuthToken":"eyJhbGciOi..."}

# 2. Encrypted copy — ciphertext
adb shell "su -c cat /data/data/com.galoyapp/shared_prefs/secret_shared_prefs.xml"
# → base64 AES-256-GCM blobs; no readable token

# 3. Confirm the token works
curl -H "Authorization: Bearer <token from step 1>" https://api.blink.sv/graphql \
  -H 'content-type: application/json' -d '{"query":"{ me { id username defaultAccount { id } } }"}'
```

Quarantine persistence:

```bash
# Corrupt the stored state so migration fails, relaunch, then log out and check:
adb shell "su -c sqlite3 /data/data/com.galoyapp/databases/RKStorage \
  \"select key from catalystLocalStorage where key like 'persistentStateQuarantine%';\""
# → keys remain, each containing a full plaintext token
```

---

### 1. Purge and redact quarantine records (ship first, independently)

```ts
const REDACTED_KEYS = ["galoyAuthToken"] as const

const quarantineRawState = async (rawData: unknown): Promise<void> => {
  const safe =
    rawData && typeof rawData === "object"
      ? { ...(rawData as Record<string, unknown>),
          ...Object.fromEntries(REDACTED_KEYS.map((k) => [k, "[redacted]"])) }
      : rawData
  const key = `${PERSISTENT_STATE_QUARANTINE_PREFIX}.${Date.now()}`
  ...
}
```

Migration diagnostics need the *shape* of the broken state, not its secrets. Then add a sweep — on boot and in `useLogout` — that `AsyncStorage.getAllKeys()`, filters the prefix, and `multiRemove`s anything older than a short retention window (or all of them on logout).

### 2. Make the Keystore the single source of truth for the token

Drop `galoyAuthToken` from `PersistentState` and resolve the active token from `KeyStoreWrapper.getSessionProfiles()` (the record already carries `selected: boolean`), holding it in React context / memory for the app's lifetime. `useAppConfig().token` and `useEffectiveAuthToken()` keep their signatures; only the backing store changes.

Add schema version 16 whose migration drops the field and, in the same pass, removes any lingering quarantine keys:

```ts
const migrate15ToCurrent = async (state: PersistentState_15): Promise<PersistentState> => {
  const { galoyAuthToken, ...rest } = state
  if (galoyAuthToken) await ensureTokenInKeystore(galoyAuthToken)   // don't strand logged-in users
  return { ...rest, schemaVersion: 16 }
}
```

### 3. Why this is not a one-line deletion

`saveProfile` writes the plaintext copy **before** it knows the Keystore copy will land ([[L88–L117](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-save-session-profile.ts#L88-L117)](https://github.com/GaloyMoney/blink-mobile/blob/8727998d16c334581f2d0f3d7d82a19ec2703ecf/app/hooks/use-save-session-profile.ts#L88-L117)):

```ts
await saveToken(token)                       // L92  → AsyncStorage
...
const profile = await tryFetchUserProps({ token, fetchUsername })
if (!profile) return                         // L106 → bails; Keystore never written
...
await KeyStoreWrapper.saveSessionProfiles([{ ...profile, ...}, ...cleaned])   // L117
```

`tryFetchUserProps` swallows its errors and returns `undefined` on any network failure, so a login on a flaky connection today produces a session that exists **only** in AsyncStorage. Delete that write without reordering and those users are logged out on the spot.

The write order has to become: persist to Keystore first with whatever identity fields are available, then enrich the profile once `getUsernames` succeeds. The existing "heal a profile saved without `accountId`" logic at L100–L103 already anticipates this shape — it just needs the token write moved inside the same guarantee. `saveTokenAndInstance` (developer screen) and the direct `saveToken` call sites in `use-switch-to-next-profile.ts`, `use-resolve-transaction-account.ts:359`, `multi-account/profile.tsx`, and `use-discard-custodial-session.ts` all need the same treatment.

### 4. Follow-ups

- Migrate `KeyStoreWrapper` from the unmaintained `react-native-secure-key-store` to `react-native-keychain@^10`, already in the dependency tree.
- Switch iOS accessibility from `ALWAYS_THIS_DEVICE_ONLY` to `AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`.

---

## Acceptance criteria

- [ ] `select value from catalystLocalStorage where key='persistentState'` on a logged-in device contains no JWT.
- [ ] No `persistentStateQuarantine.*` key contains a token; quarantine records are redacted at write time.
- [ ] Quarantine keys are swept on boot and cleared on logout.
- [ ] A user logged in on a build with schema 15 stays logged in after upgrading to schema 16 — the migration moves the token into the Keystore rather than dropping it.
- [ ] A login whose `getUsernames` call fails still produces a session recoverable from the Keystore alone (regression test for §3).
- [ ] Logout leaves no readable token in either store.
- [ ] Follow-up issues opened for the `react-native-keychain` migration and the iOS accessibility class.

---

## References

- [[OWASP MASVS — MASVS-STORAGE-1: sensitive data in system credential storage](https://mas.owasp.org/MASVS/controls/MASVS-STORAGE-1/)](https://mas.owasp.org/MASVS/controls/MASVS-STORAGE-1/)
- [[MASTG — Testing Local Storage for Sensitive Data (Android)](https://mas.owasp.org/MASTG/tests/android/MASVS-STORAGE/MASTG-TEST-0001/)](https://mas.owasp.org/MASTG/tests/android/MASVS-STORAGE/MASTG-TEST-0001/)
- [AndroidX `EncryptedSharedPreferences`](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences)
- [[Android Keystore system](https://developer.android.com/privacy-and-security/keystore)](https://developer.android.com/privacy-and-security/keystore)
- [`[@react-native-async-storage/async-storage](https://github.com/react-native-async-storage/async-storage)`](https://github.com/react-native-async-storage/async-storage) — `RNCAsyncStorage.mm`, backup-exclusion default
- [`[react-native-secure-key-store](https://github.com/pradeep1991singh/react-native-secure-key-store)`](https://github.com/pradeep1991singh/react-native-secure-key-store) — `RNSecureKeyStoreModule.java`
- [`[kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlockthisdeviceonly)`](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlockthisdeviceonly)

### DIFF
--- a/__tests__/hooks/use-app-config.spec.ts
+++ b/__tests__/hooks/use-app-config.spec.ts
@@ -1,0 +1,112 @@
+import { renderHook, act } from "@testing-library/react-native"
+
+import { resolveGaloyInstanceOrDefault } from "@app/config"
+import { useAppConfig } from "@app/hooks/use-app-config"
+
+const mockReportError = jest.fn()
+const mockUpdateState = jest.fn()
+
+const mockSecureStore = new Map<string, string>()
+const failSetFor = new Set<string>()
+jest.mock("react-native-secure-key-store", () => ({
+  __esModule: true,
+  default: {
+    get: (key: string) =>
+      mockSecureStore.has(key)
+        ? Promise.resolve(mockSecureStore.get(key))
+        : Promise.reject(new Error(`key not found: ${key}`)),
+    set: (key: string, value: string) => {
+      if (failSetFor.has(key)) return Promise.reject(new Error("keystore locked"))
+      mockSecureStore.set(key, value)
+      return Promise.resolve(true)
+    },
+    remove: (key: string) => {
+      mockSecureStore.delete(key)
+      return Promise.resolve(true)
+    },
+  },
+  ACCESSIBLE: {
+    ALWAYS_THIS_DEVICE_ONLY: "ALWAYS_THIS_DEVICE_ONLY",
+    WHEN_UNLOCKED_THIS_DEVICE_ONLY: "WHEN_UNLOCKED_THIS_DEVICE_ONLY",
+  },
+}))
+
+jest.mock("@app/store/persistent-state", () => ({
+  usePersistentStateContext: () => ({
+    persistentState: { galoyAuthToken: "", galoyInstance: { id: "Main" } },
+    updateState: mockUpdateState,
+  }),
+}))
+
+jest.mock("@app/utils/error-logging", () => ({
+  reportError: (...args: unknown[]) => mockReportError(...args),
+}))
+
+// Applies the updater the way PersistentStateProvider would, so tests can see
+// what the in-memory state became.
+const applyLastUpdate = (state: Record<string, unknown>) =>
+  mockUpdateState.mock.calls.map(([update]) => update(state)).at(-1)
+
+describe("useAppConfig token persistence", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockSecureStore.clear()
+    failSetFor.clear()
+  })
+
+  it("saveToken writes the token to the key store and into in-memory state", async () => {
+    const { result } = renderHook(() => useAppConfig())
+
+    await act(async () => {
+      await result.current.saveToken("session-token")
+    })
+
+    expect(mockSecureStore.get("activeAuthToken")).toBe("session-token")
+    expect(applyLastUpdate({ galoyAuthToken: "" })).toEqual({
+      galoyAuthToken: "session-token",
+    })
+  })
+
+  it('saveToken("") removes the key-store token', async () => {
+    mockSecureStore.set("activeAuthToken", "session-token")
+    const { result } = renderHook(() => useAppConfig())
+
+    await act(async () => {
+      await result.current.saveToken("")
+    })
+
+    expect(mockSecureStore.has("activeAuthToken")).toBe(false)
+  })
+
+  it("saveTokenAndInstance persists the token alongside the instance switch", async () => {
+    const { result } = renderHook(() => useAppConfig())
+
+    await act(async () => {
+      await result.current.saveTokenAndInstance({
+        token: "session-token",
+        instance: resolveGaloyInstanceOrDefault({ id: "Main" }),
+      })
+    })
+
+    expect(mockSecureStore.get("activeAuthToken")).toBe("session-token")
+  })
+
+  it("reports a rejected key-store write instead of resolving silently", async () => {
+    failSetFor.add("activeAuthToken")
+    const { result } = renderHook(() => useAppConfig())
+
+    await act(async () => {
+      await result.current.saveToken("session-token")
+    })
+
+    // The in-memory login proceeds for this session, but the failure to
+    // persist is surfaced — otherwise the session would evaporate at next
+    // launch with nothing recorded anywhere.
+    expect(mockSecureStore.has("activeAuthToken")).toBe(false)
+    expect(mockReportError).toHaveBeenCalledTimes(1)
+    expect(mockReportError.mock.calls[0][0]).toBe("persist auth token")
+    expect(applyLastUpdate({ galoyAuthToken: "" })).toEqual({
+      galoyAuthToken: "session-token",
+    })
+  })
+})
--- a/__tests__/hooks/use-logout.spec.ts
+++ b/__tests__/hooks/use-logout.spec.ts
@@ -1,0 +1,246 @@
+import { renderHook, act } from "@testing-library/react-native"
+
+import { SCHEMA_VERSION_KEY } from "@app/config"
+import useLogout from "@app/hooks/use-logout"
+
+const mockResetState = jest.fn()
+const mockUserLogoutMutation = jest.fn()
+const mockMultiRemove = jest.fn()
+const mockGetDeviceToken = jest.fn()
+const mockReportError = jest.fn()
+
+// In-memory stand-in for the secure key store: the assertions below run
+// against the real KeyStoreWrapper, so the profile/active-token interplay is
+// exercised end to end rather than through mocked wrapper methods.
+const mockSecureStore = new Map<string, string>()
+const failRemoveFor = new Set<string>()
+jest.mock("react-native-secure-key-store", () => ({
+  __esModule: true,
+  default: {
+    get: (key: string) =>
+      mockSecureStore.has(key)
+        ? Promise.resolve(mockSecureStore.get(key))
+        : Promise.reject(new Error(`key not found: ${key}`)),
+    set: (key: string, value: string) => {
+      mockSecureStore.set(key, value)
+      return Promise.resolve(true)
+    },
+    remove: (key: string) => {
+      if (failRemoveFor.has(key)) return Promise.reject(new Error("keystore locked"))
+      mockSecureStore.delete(key)
+      return Promise.resolve(true)
+    },
+  },
+  ACCESSIBLE: {
+    ALWAYS_THIS_DEVICE_ONLY: "ALWAYS_THIS_DEVICE_ONLY",
+    WHEN_UNLOCKED_THIS_DEVICE_ONLY: "WHEN_UNLOCKED_THIS_DEVICE_ONLY",
+  },
+}))
+
+jest.mock("@react-native-async-storage/async-storage", () => ({
+  multiRemove: (...args: unknown[]) => mockMultiRemove(...args),
+}))
+
+jest.mock("@react-native-firebase/messaging", () => () => ({
+  getToken: () => mockGetDeviceToken(),
+}))
+
+jest.mock("@app/store/persistent-state", () => ({
+  usePersistentStateContext: () => ({ resetState: mockResetState }),
+}))
+
+jest.mock("@app/graphql/generated", () => ({
+  ...jest.requireActual("@app/graphql/generated"),
+  useUserLogoutMutation: () => [mockUserLogoutMutation],
+}))
+
+jest.mock("@app/utils/analytics", () => ({ logLogout: jest.fn() }))
+jest.mock("@app/utils/error-logging", () => ({
+  reportError: (...args: unknown[]) => mockReportError(...args),
+}))
+
+const profileA = { token: "tok-a", username: "alice", accountId: "acct-a" }
+const profileB = { token: "tok-b", username: "bob", accountId: "acct-b" }
+
+const seedProfiles = (profiles: { token: string }[]) =>
+  mockSecureStore.set("sessionProfiles", JSON.stringify(profiles))
+
+const storedProfiles = (): { token: string }[] =>
+  JSON.parse(mockSecureStore.get("sessionProfiles") ?? "[]")
+
+describe("useLogout", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockSecureStore.clear()
+    failRemoveFor.clear()
+    mockUserLogoutMutation.mockResolvedValue({ data: { userLogout: { success: true } } })
+    mockMultiRemove.mockResolvedValue(undefined)
+    mockGetDeviceToken.mockResolvedValue("device-token")
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  it("full logout clears the active token and all profiles from the key store", async () => {
+    mockSecureStore.set("activeAuthToken", "tok-a")
+    seedProfiles([profileA, profileB])
+
+    const { result } = renderHook(() => useLogout())
+    await act(async () => {
+      await result.current.logout()
+    })
+
+    expect(mockSecureStore.has("activeAuthToken")).toBe(false)
+    expect(mockSecureStore.has("sessionProfiles")).toBe(false)
+    expect(mockResetState).toHaveBeenCalled()
+  })
+
+  it("full logout also wipes PIN, biometrics and the schema version key, without calling the server mutation", async () => {
+    mockSecureStore.set("PIN", "1234")
+    mockSecureStore.set("pinAttempts", "3")
+    mockSecureStore.set("isBiometricsEnabled", "1")
+    mockSecureStore.set("activeAuthToken", "tok-a")
+    seedProfiles([profileA])
+
+    const { result } = renderHook(() => useLogout())
+    await act(async () => {
+      await result.current.logout()
+    })
+
+    expect(mockSecureStore.has("PIN")).toBe(false)
+    expect(mockSecureStore.has("pinAttempts")).toBe(false)
+    expect(mockSecureStore.has("isBiometricsEnabled")).toBe(false)
+    expect(mockMultiRemove).toHaveBeenCalledWith([SCHEMA_VERSION_KEY])
+    // No token context means there is nothing to revoke server-side.
+    expect(mockUserLogoutMutation).not.toHaveBeenCalled()
+  })
+
+  it("full logout with stateToDefault: false keeps the in-memory state", async () => {
+    const { result } = renderHook(() => useLogout())
+    await act(async () => {
+      await result.current.logout({ stateToDefault: false })
+    })
+
+    expect(mockResetState).not.toHaveBeenCalled()
+  })
+
+  it("removing the profile backing the active session clears the active token", async () => {
+    mockSecureStore.set("activeAuthToken", "tok-a")
+    seedProfiles([profileA, profileB])
+
+    const { result } = renderHook(() => useLogout())
+    await act(async () => {
+      await result.current.logout({ stateToDefault: false, token: "tok-a" })
+    })
+
+    // Regression: the deleted profile's token must not linger as the active
+    // token — it would otherwise be hydrated back into memory on next launch.
+    expect(mockSecureStore.has("activeAuthToken")).toBe(false)
+    expect(storedProfiles().map((p) => p.token)).toEqual(["tok-b"])
+    expect(mockResetState).not.toHaveBeenCalled()
+    expect(mockUserLogoutMutation).toHaveBeenCalledWith(
+      expect.objectContaining({
+        context: { headers: { authorization: "Bearer tok-a" } },
+      }),
+    )
+  })
+
+  it("removing an inactive profile keeps the active token untouched", async () => {
+    mockSecureStore.set("activeAuthToken", "tok-b")
+    seedProfiles([profileA, profileB])
+
+    const { result } = renderHook(() => useLogout())
+    await act(async () => {
+      await result.current.logout({ stateToDefault: false, token: "tok-a" })
+    })
+
+    expect(mockSecureStore.get("activeAuthToken")).toBe("tok-b")
+    expect(storedProfiles().map((p) => p.token)).toEqual(["tok-b"])
+  })
+
+  it("isValidToken: false skips the server revocation but still cleans up locally", async () => {
+    mockSecureStore.set("activeAuthToken", "tok-a")
+    seedProfiles([profileA, profileB])
+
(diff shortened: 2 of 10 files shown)


grimen (changes requested):
## Findings

**1. A failed push-token fetch skips the whole key-store cleanup (blocking).**

In [`use-logout.ts:42`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/hooks/use-logout.ts#L42) the `await messaging().getToken()` call runs at the top of the `try`, before any token removal. If it throws, every removal below is skipped, the `catch` only reports the error, and `finally` still calls `resetState()`. Before this PR that reset was a real logout because the token lived in the persisted state; now it survives in `activeAuthToken` and `hydrateAuthToken` restores it on the next launch.

*Example:* a user on a device without Play Services (or simply offline) taps log out, sees the app return to the welcome screen, and is signed back into the same account after restarting.

*Proposal:* one option could be to wrap the device-token fetch in its own `try`/`catch`, or to clear the key store before any network call.

**2. Hydrating the token on `MigrationStatus.NoData` re-authenticates fresh installs (blocking).**

[`persistent-state/index.tsx:55`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/store/persistent-state/index.tsx#L55) reads the key store even when there is no persisted state at all. On iOS the keychain outlives app deletion, so "no data" does not mean "no session".

*Example:* a user deletes the app and reinstalls it; the app opens already logged into the previous install's account while every other setting is at its default.

*Proposal:* hydrating only on the `Ok` path would keep a fresh install logged out.

**3. Quarantining the raw blob writes the plaintext token back to disk (blocking).**

[`index.tsx:58`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/store/persistent-state/index.tsx#L58) calls `quarantineRawState(result.rawData)`, and for legacy users that raw data still contains the `galoyAuthToken` this PR is removing from disk. The quarantine key is never cleaned up, so the copy is permanent.

*Example:* a legacy user with a corrupted state blob ends up with their session token stored in cleartext under `persistentStateQuarantine.<ts>` for the lifetime of the install.

*Proposal:* stripping `galoyAuthToken` before serializing the quarantine payload would keep the PR's guarantee intact.

**4. The `Failed` path pairs the live token with the default Galoy instance.**

[`index.tsx:59`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/store/persistent-state/index.tsx#L59) re-attaches the hydrated token to `defaultPersistentState`, whose `galoyInstance` is `{ id: "Main" }`, so the deliberate quarantine-and-reset now keeps the user authenticated against a possibly different backend.

*Example:* someone running a Custom or staging instance via the developer screen hits a corrupted state blob and starts sending that instance's bearer token to Main.

*Proposal:* leaving the `Failed` path fully logged out would preserve the clean-slate intent, and comes for free with the change proposed in findings 2 and 3.

**5. The legacy migration ignores whether the key-store write succeeded.**

[`index.tsx:42`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/store/persistent-state/index.tsx#L42) discards the boolean returned by `setActiveAuthToken`. If the keychain write fails the token exists only in memory, and the next `savePersistentState` overwrites the disk copy with `galoyAuthToken: ""`.

*Example:* a keychain write fails once during migration; the user keeps working that session, then finds themselves logged out at next launch with nothing recorded anywhere.

*Proposal:* checking the result and either keeping the blob copy or calling `recordAppError` would make the failure visible.

**6. `persistToken` reports success even when the key store rejected the write.**

[`use-app-config.ts:11`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/hooks/use-app-config.ts#L11) updates in-memory state and resolves regardless of the return value, and the token no longer goes to disk as a fallback.

*Example:* a login on a device with a locked or full keychain looks completely successful, then evaporates on the next app start.

*Proposal:* propagating or reporting the `false` would let the caller surface a real failure.

## Test Coverage Gaps

- The spec "resets state even when fetching the push device token fails" exercises the finding 1 path but only asserts `resetState`; it does not assert that the key-store token is gone, which is what makes the bug invisible.
- No spec covers `MigrationStatus.NoData` with a token present in the key store (finding 2), so the reinstall-stays-logged-in behavior is untested in either direction.
- No spec asserts what `quarantineRawState` actually writes (finding 3), so the plaintext token in the quarantine payload passes silently.
- No spec covers `setActiveAuthToken` returning `false` during the legacy migration (finding 5) or during `persistToken` (finding 6); both failure branches are unexercised.

## Code Smells

- `persistToken` in [`use-app-config.ts:7`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/hooks/use-app-config.ts#L7) is typed `Promise<void>` even though both key-store calls it wraps return a boolean, so the signature itself is what discards the failure signal described in finding 6.
- [`hydrateAuthToken`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/store/persistent-state/index.tsx#L36) does two unrelated things behind one name: a read-through hydrate, and a one-off legacy write that moves an old on-disk token into the key store. Splitting the migration out would make the `NoData` and `Failed` call sites obviously wrong rather than subtly wrong.
- The same `hydrateAuthToken(...)` call is repeated on all three switch branches in [`loadPersistentState`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/store/persistent-state/index.tsx#L46); resolving the state in the switch and hydrating once after it would leave a single place to reason about.
- `PersistentState` still declares `galoyAuthToken` as a normal persisted field while [`savePersistentState`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/store/persistent-state/index.tsx#L64) always blanks it on the way out, so the type no longer describes the on-disk schema and future migrations reading `state.galoyAuthToken` will silently see `""`.
- [`use-logout.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/b9a29bb018ed6a2310f29c70fbb22d980b50c5b7/app/hooks/use-logout.ts#L43) now removes the active token two different ways depending on the branch — a compare-then-remove when a token is passed, an unconditional remove otherwise — which is what makes the read failure in the compare path easy to miss.

PR #4140, opened 2026-08-10
fix(lnurl): require https for decoded lnurl urls and service callbacks

kngako:
js-lnurl and lnurl-pay fetch the URL embedded in a bare bech32 LNURL verbatim, with no enforcement that the scheme is https, and the app added no scheme check of its own before using the decoded URL or the service-returned callback. A crafted QR/NFC tag could therefore trigger cleartext http fetches of the LNURL service and of the withdraw/pay callbacks (which carry the invoice).

Decode bare lnurl1 strings and vet the scheme in resolveLnurlDestination before any network call, reject non-https withdraw and pay callbacks returned by the service, and re-check the callback at the two direct fetch sites (redeem screen, NFC receive) as defense in depth.

### DIFF
--- a/__tests__/payment-destination/lnurl.spec.ts
+++ b/__tests__/payment-destination/lnurl.spec.ts
@@ -1,3 +1,4 @@
+import { bech32 } from "bech32"
 import { LNURLResponse, LNURLWithdrawParams, getParams } from "js-lnurl"
 import { requestPayServiceParams, LnUrlPayServiceResponse, Satoshis } from "lnurl-pay"
 
@@ -5,7 +6,10 @@ import {
   createLnurlPaymentDestination,
   resolveLnurlDestination,
 } from "@app/screens/send-bitcoin-screen/payment-destination"
-import { DestinationDirection } from "@app/screens/send-bitcoin-screen/payment-destination/index.types"
+import {
+  DestinationDirection,
+  InvalidDestinationReason,
+} from "@app/screens/send-bitcoin-screen/payment-destination/index.types"
 import { createLnurlPaymentDetails } from "@app/screens/send-bitcoin-screen/payment-details"
 import { ZeroBtcMoneyAmount } from "@app/types/amounts"
 import { PaymentType } from "@blinkbitcoin/blink-client"
@@ -40,7 +44,7 @@ const throwError = () => {
 const manualMockLnUrlPayServiceResponse = (
   identifier: string,
 ): LnUrlPayServiceResponse => ({
-  callback: "mocked_callback",
+  callback: "https://example.com/callback",
   fixed: true,
   min: 0 as Satoshis,
   max: 2000 as Satoshis,
@@ -68,7 +72,7 @@ const manualMockLNURLWithdrawParams = (): LNURLWithdrawParams => ({
   // Example structure. Adjust according to your actual LNURLWithdrawParams type
   tag: "withdrawRequest",
   k1: "some_random_string",
-  callback: "http://example.com/callback",
+  callback: "https://example.com/callback",
   domain: "example.com",
   maxWithdrawable: 2000,
   minWithdrawable: 0,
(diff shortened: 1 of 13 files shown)


grimen (changes requested):
## Findings

**1. Bare-bech32 gate misses LUD-17 URIs, which is where the `.onion` cleartext downgrade actually bites (blocking).**

The new scheme check at [`lnurl.ts:44`](https://github.com/blinkbitcoin/blink-mobile/blob/252eefadc7e74b95e47a524c51b2dee2b34806f0/app/screens/send-bitcoin-screen/payment-destination/lnurl.ts#L44) only fires when `decodeBareLnurl` returns a URL, and that function returns `null` for anything not starting with `lnurl1`. `@blinkbitcoin/blink-client` also classifies `lnurlw://` and `lnurlp://` destinations as `PaymentType.Lnurl` and passes them through raw, so those skip the gate entirely and go straight to `getParams`. Both underlying libraries then drop to plain http whenever `.onion` appears anywhere in the string — js-lnurl matches `/\.onion($|\W)/`, lnurl-pay does `parsedUrl.includes(".onion")`.

*Example:* a tag reading `lnurlw://attacker.com/w/.onion/x` is fetched over `http://attacker.com/w/.onion/x`, which is the exact cleartext service fetch this PR sets out to prevent.

*Proposal:* one option could be to normalise LUD-17 schemes to `https`/`http` before the check and reject the `http` case, so the gate covers both destination forms rather than only the bech32 one.

**2. Both new rejections render the generic destination error instead of the LNURL one (blocking).**

[`destination-information.tsx:95`](https://github.com/blinkbitcoin/blink-mobile/blob/252eefadc7e74b95e47a524c51b2dee2b34806f0/app/screens/send-bitcoin-screen/destination-information.tsx#L95) reads `case InvalidDestinationReason.LnurlError || InvalidDestinationReason.LnurlUnsupported:`, which evaluates to `LnurlError` alone, so `LnurlUnsupported` has never had a branch. This PR routes both new callback rejections ([`lnurl.ts:62`](https://github.com/blinkbitcoin/blink-mobile/blob/252eefadc7e74b95e47a524c51b2dee2b34806f0/app/screens/send-bitcoin-screen/payment-destination/lnurl.ts#L62), [`lnurl.ts:89`](https://github.com/blinkbitcoin/blink-mobile/blob/252eefadc7e74b95e47a524c51b2dee2b34806f0/app/screens/send-bitcoin-screen/payment-destination/lnurl.ts#L89)) to `LnurlUnsupported`, so they fall through to `default`.

*Example:* scanning a code whose service returns an `http://` callback shows "enter a valid destination" rather than the LNURL error with its advice tooltip.

*Proposal:* the pre-existing bug is out of scope, but reusing `LnurlError` for the two new rejections — as the bare-URL check at line 44 already does — would keep them on the intended message.

**3. `decodeBareLnurl` fails open on any decode error.**

[`app/utils/lnurl.ts:8`](https://github.com/blinkbitcoin/blink-mobile/blob/252eefadc7e74b95e47a524c51b2dee2b34806f0/app/utils/lnurl.ts#L8) returns `null` both for "this isn't a bare LNURL" and for "this is a bare LNURL I couldn't decode", and the caller reads `null` as permission to skip the scheme check. The app resolves `bech32@2` while js-lnurl bundles `bech32@1.1.4`, so the two decoders need not agree on every input.

*Example:* a `lnurl1…` string that this repo's bech32 rejects on checksum or charset but js-lnurl accepts reaches `getParams` with no scheme check at all.

*Proposal:* for the implementer to decide — if the two decoders are assumed to agree, a comment saying so would settle it; otherwise returning `null` only for a non-`lnurl1` prefix and signalling decode failure separately would let the caller reject rather than pass.

## Test Coverage Gaps

- Neither hook guard is exercised: [`use-lnurl-withdraw-redemption.spec.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/252eefadc7e74b95e47a524c51b2dee2b34806f0/__tests__/screens/redeem-lnurl-withdrawal-screen/hooks/use-lnurl-withdraw-redemption.spec.tsx) and [`use-lnurl-withdraw.spec.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/252eefadc7e74b95e47a524c51b2dee2b34806f0/__tests__/receive-bitcoin/hooks/use-lnurl-withdraw.spec.ts) are untouched by this PR and contain no `http://` callback case, so both new rejections could be deleted without a test failing.
- `app/utils/lnurl.ts` has no unit spec — neither `decodeBareLnurl`'s prefix/failure behavior nor `isHttpsUrl`'s handling of malformed input is pinned.
- No test covers a `lnurlw://` or `lnurlp://` destination, which is why finding 1 slipped through.

grimen (approved):

PR #4144, opened 2026-08-11
fix(region): resolve the creation region when an option is submitted, not on mount

### LINKED ISSUE #3924: feat(compliance): Region determination (aka Geo-blocking) — three-mode user journey (Custodial / NC Enhanced / NC Anon)

esaugomez31:
## What

Region rules for account creation now run when the user submits a choice, not when a screen mounts. Nothing about the user is read until there is a choice to hold against a rule.

Previously the country decided which account types to offer, so the lookup had to happen before anything was on screen. It now decides only whether the option the user picked may proceed.

## Changes

**Both account types are always offered.** The list no longer depends on where the user is. A refused option is answered on continue, which is where the regional rule already lived.

**The refusal says which rule refused it.** A closed region, a first Blink account that cannot be opened there, and a location that could not be read are three different outcomes, and the unsupported-region screen now words each on its own terms. The first-signup case points at self-custodial only where this build actually offers it.

**An unreadable location is no longer reported as a regional block.** It says so, and asks the user to check their connection.

**Incognito reads nothing, the connection included.** Self-custodial's region rule moved to after the mode screen, and it never runs for Incognito. That screen already promised this in its own copy: Enhanced says we use your region to unlock valuable services, Incognito promises maximum privacy. It now holds. The trade is deliberate: `selfCustodialCreationBlockedCountries` no longer applies to a user who picks Anon.

**Get Started asks nothing where a choice is still ahead.** With both types on offer it only navigates. A build offering a single type has no selection screen, so pressing create is itself that choice and is answered there.

**The country comes from the connection alone.** An account being created has no phone of its own, and borrowing another account's would make the verdict depend on which one happens to be open. This matches `regionCheck`, which resolves from the IP.

**No local fallback.** The country-code cache answers `SV` for an unset value, so reading it turned "could not resolve" into a country nobody read, and let a blocked user through. An unresolved location now stays unresolved and is answered as such.

**A submit button waits only on what its own check reads.** The device's account count gates the custodial first-signup rule, so the screens that submit custodial wait for it and the ones that submit self-custodial do not.

**The selection cannot change mid-check.** A card pressed while an answer is in flight would otherwise leave that answer describing the option the user moved away from.

**Restore never consults the region.** It creates no account, so it has nothing to check and nobody to locate.

## Screenshots

Staging, El Salvador. Walked with a probe on every point that can read a country, counting real lookups rather than reading the screen: zero through Get Started, the selection screen, the mode screen and both restore paths; exactly one on submitting custodial or Enhanced; zero on Anon, including from a country forced into both blocklists.

| Scenario | What it shows | Screenshot |
|---|---|---|
| Get Started | Opening it resolves no location at all | 01-get-started |
| Account type selection | Both types offered immediately, nothing resolved yet | 02-account-type-selection |
| Custodial picked | Continue enabled | 03-custodial-selected |
| Blocked country, one account already held | Allowed: the rule refuses only a user's very first Blink account | 04-allowed-terms |
| Blocked country, no accounts | Refused, pointing at the alternative that is available | 05-refused-first-signup |
| Same refusal, self-custodial turned off | The offer is withheld rather than sending the user to a dead end | 06-refused-no-alternative |

## Note

This is the client side only. The custodial half of these rules is due to move to the backend once `regionCheck` and `custodialRestrictions` land; the shape here keeps that swap to one hook.


grimen (changes requested):
## Findings

**1. Restore mode waits on a rule it never evaluates (blocking).**
`isContinueDisabled` on `account-type-selection.tsx:81` includes `!isFirstSignupRuleReady` in every mode, but the restore branch and the self-custodial create branch never call `checkBlockReason` and never read the account count. The deleted code was explicit about this with `loading: isRestore ? false : loading`.

*Example:* On a cold start the user picks "Log in or restore", reaches the selection screen, and Continue stays inert until both `listSelfCustodialAccounts()` and `KeyStoreWrapper.getSessionProfiles()` settle; because `reloadSelfCustodialAccounts` re-sets `selfCustodialHydrating` whenever `persistentState.activeAccountId` changes, the button can also re-disable mid-session and swallow a press.

*Proposal:* One option could be to apply the readiness term only when it is actually consulted, e.g. `isCreateMode && selected === AccountOption.Custodial`.

**2. Incognito creation is not checked against the blocked-country list (blocking).**
On `choose-experience-screen.tsx:138` the Incognito path creates a self-custodial wallet without evaluating `selfCustodialCreationBlockedCountries`, while the Enhanced path on the same screen is refused. The PR body notes this is deliberate, but the two remote-config lists are named as peers and only one of them enforces anything.

*Example:* A user in a country on `selfCustodialCreationBlockedCountries` taps "Enhanced" and is refused, then taps "Incognito" and completes creation.

*Proposal:* For the implementer to confirm — if this is intended, recording the product or compliance sign-off in the PR body (and a comment at the call site) would make the asymmetry legible; otherwise the Incognito path could run the same check.

grimen (approved):

PR #4145, opened 2026-08-12
feat(region): shape the client's region rules after the queries that will serve them

### LINKED ISSUE #3924: feat(compliance): Region determination (aka Geo-blocking) — three-mode user journey (Custodial / NC Enhanced / NC Anon)

esaugomez31:
## What

Moves the client's custodial region rules onto the two queries that now serve them, and carries the self-custodial account mode to the LNURL server that acts on it. [blink#756](https://github.com/blinkbitcoin/blink/pull/756) is merged and live on staging, so this no longer only shapes the client after those queries — it reads them.

Backend:
- [blink#756](https://github.com/blinkbitcoin/blink/pull/756) — `regionCheck` and `custodialRestrictions`. **Merged and integrated here.**
- [blink-lnurl-server#36](https://github.com/blinkbitcoin/blink-lnurl-server/pull/36) — account mode persisted server-side. **Merged and integrated here.**

Ticket: [blink-wip#1035](https://github.com/blinkbitcoin/blink-wip/issues/1035).

## Where each rule is answered

| Rule | Before | Now |
|---|---|---|
| Custodial creation, session sanctions | `custodialCreationBlockedCountries` + IP | `regionCheck` |
| Custodial dollar balance, transfers | `custodial*BlockedCountries` + phone or IP | `custodialRestrictions` |
| First custodial signup | Device account count | Unchanged. The count is on the device and the server cannot see it |
| Every self-custodial rule | Firebase lists | Unchanged. A self-custodial wallet has no Blink account behind it |

**The server picks the country it answers for.** `custodialRestrictions` selects by account level — verified phone for level 1-2, request IP for level 0 — so the client never chooses between sources. That removes the phone-versus-IP waterfall from the custodial path entirely: an authenticated custodial session now performs **zero** client-side IP lookups.

**The three custodial lists are deleted**, along with their compiled-in defaults: `custodialCreationBlockedCountries`, `custodialDollarBalanceBlockedCountries`, `custodialTransferBlockedCountries`. The four that survive are the ones no server verdict covers.

**Verdicts are re-asked for at each session start.** Both queries are `no-cache` and refetch on foreground, and the shared IP lookup — memoised for the whole JS process — is dropped there too, through a generation counter the lookup hooks subscribe to. Clearing the promise alone only served consumers mounting afterwards, so a screen already on display kept the country it resolved a session ago. That was the latch this work exists to remove.

**Anon is excluded explicitly**, not left to the query's own `skip`. Apollo parks a skipped query in `standby` and would not refetch it, but the promise that an Anon account is never located should not rest on that.

## Policy on an unanswered verdict

`custodialRestrictions` **fails closed**: no verdict, no gated feature, matching `UnknownRegionPolicy = FAIL_CLOSED` in the PRD's P0 values.

`regionCheck` **fails open**: an unreachable server does not raise the sanctions blocker. Deliberate — `regionCheck` shares an API with every other custodial feature, so if it cannot be reached neither can sending, receiving or converting, and blocking the whole session on an outage helps nobody. Flagged here because the two policies differ on the same failure.

## The account mode on the LNURL server

The mode is the LNURL server's business too: it decides whether the server mints invoices for a self-custodial account's Lightning Address. Enhanced lets it answer and records the country it saw; Anon stops both.

**The address is never registered or deleted from here.** Only the mode is written. Deleting the address would free the username, and a later re-registration could lose it to someone else — so a switch can never cost the user their name.

**The push is driven off disagreement, not off the moment the user picks.** The choice is made on screens with no connected SDK to sign with, so the chosen mode and the server-confirmed one are stored separately (schema 20) and the gap between them is the work owed. What lands is recorded, so a launch with nothing to say stays silent: each Enhanced push costs the server a paid country lookup against a per-IP budget.

**An account holding no mode is settled from the server before anything is assumed.** Accounts predating the modes, and abandoned migrations, reach a connected wallet with nothing stored. The same wallet may already be Anon on another device, so the server is asked first; only a server that answers "none" gets the Enhanced default pushed at it.

**A restored wallet takes its mode from the server, not from a fresh question.** It was chosen on a device this one may never have been. The picker now appears only when the server could not be asked at all — assuming a default there would push it over whatever the account really holds.

Requests are signed with the wallet's own key and DER-encoded: `mode:{mode}:{pubkey}-{timestamp}` to write, the bare `{pubkey}-{timestamp}` to read. They go to the host the account's address is spelled with — `blink.sv` or `staging.blink.sv` — which is what serves the authenticated `/lnurlpay/{pubkey}` routes. Not the custodial `lnAddressHostname`: `pay.*` fronts the payment app and 404s them.

## Verification

Typecheck, lint and `graphql-check` (against the live staging endpoint) all clean. **100% statements, branches, functions and lines** on every file this PR adds or touches.

Walked on an Android emulator against staging, on a custodial account:

| Case | Result |
|---|---|
| Backend answers | `{"dollarBalance":false,"transfer":false}` — the real query |
| Unrestricted | Dollar balance visible, transfer and send enabled, nothing broken |
| Restricted verdict | Blocker presents, and the USD→BTC divest stays available — the remediation path the backend contract requires |
| `dollarBalance` alone | Dollar row greyed with its amount still shown, transfer disabled, receive/send/scan untouched |
| Client-side IP reads, authed custodial session | **0** |

Region reads across account creation, traced on device:

| Step | Reads |
|---|---|
| Get Started, account type, mode selection | 0 |
| Submit Anon | **0** |
| Submit Enhanced | 1 |

Anon resolves nothing at any point, and even Enhanced resolves only on submit — never on mount.

The mode, walked on the same emulator against staging with a registered address (`esaumodetest01@staging.blink.sv`):

| Case | Result |
|---|---|
| Custodial session | **0** requests to the LNURL server |
| Self-custodial holding no mode | Nothing pushed until a mode is known |
| Enhanced chosen | `200`; the signature verifies as ECDSA over `sha256(message)` against the wallet pubkey |
| Switched to Anon | `200`, with a timestamp strictly newer than the previous one |
| Relaunch | **0** repeat requests — the stored confirmation holds |
| Server answers `500` | Nothing confirmed; the next launch retries and confirms |
| Invoice while Enhanced | bolt11 minted |
| Invoice while Incognito | Refused: `recipient not accepting payments` |
| Invoice back on Enhanced | minted again, with no re-registration |

### DIFF
--- a/__tests__/components/restricted-region/restricted-region.spec.tsx
+++ b/__tests__/components/restricted-region/restricted-region.spec.tsx
@@ -13,16 +13,31 @@ jest.mock("@app/utils/ip-country-lookup")
 
 let mockIpCountry: string | undefined
 let mockIpSettled = true
-const mockUseIpCountryLookup = jest.fn<
-  { countryCode: string | undefined; isSettled: boolean },
-  [boolean]
->(() => ({ countryCode: mockIpCountry, isSettled: mockIpSettled }))
-jest.mock("@app/hooks/use-device-location", () => ({
-  ...jest.requireActual("@app/hooks/use-device-location"),
-  useIpCountryLookup: (enabled: boolean) => mockUseIpCountryLookup(enabled),
-  /** The wrapper resolves through the module-internal binding, so it must be mocked
-   *  alongside the lookup or a consumer of it would hit the real hook. */
-  useIpCountryCode: (enabled: boolean) => mockUseIpCountryLookup(enabled).countryCode,
+/** The custodial verdict is the server's now, so the region is driven through the query
+ *  the provider reads. `restricted` follows the same list the old lookup was matched
+ *  against, which is what these cases are written in terms of. */
+const CUSTODIAL_BLOCKED_COUNTRIES = ["CU", "IR"]
+const mockUseRegionCheckQuery = jest.fn((_options?: unknown) => ({
+  data: {
+    regionCheck: {
+      countryCode: mockIpCountry,
+      custodialCreationAllowed: !CUSTODIAL_BLOCKED_COUNTRIES.includes(
+        (mockIpCountry ?? "").toUpperCase(),
+      ),
+      restricted: CUSTODIAL_BLOCKED_COUNTRIES.includes(
+        (mockIpCountry ?? "").toUpperCase(),
+      ),
+    },
+  },
+  loading: !mockIpSettled,
+}))
+jest.mock("@app/graphql/generated", () => ({
+  ...jest.requireActual("@app/graphql/generated"),
+  useRegionCheckQuery: (options: unknown) => mockUseRegionCheckQuery(options),
+}))
+
+jest.mock("@app/self-custodial/hooks/use-self-custodial-account-mode", () => ({
+  useSelfCustodialAccountMode: () => ({ isAnonMode: false }),
 }))
 
 let mockActiveAccountType: AccountType | undefined = AccountType.SelfCustodial
@@ -37,7 +52,6 @@ jest.mock("@app/hooks/use-account-registry", () => ({
 let mockRemoteConfigReady = true
 jest.mock("@app/config/feature-flags-context", () => ({
   useRemoteConfig: () => ({
-    custodialCreationBlockedCountries: ["CU", "IR"],
     selfCustodialCreationBlockedCountries: ["KP"],
   }),
   useFeatureFlags: () => ({ remoteConfigReady: mockRemoteConfigReady }),
@@ -161,7 +175,9 @@ describe("RestrictedRegionProvider", () => {
     const { getByTestId } = renderWithProvider()
 
     expect(getByTestId("restricted-value").props.children).toBe("false")
-    expect(mockUseIpCountryLookup).toHaveBeenCalledWith(false)
+    expect(mockUseRegionCheckQuery).toHaveBeenCalledWith(
+      expect.objectContaining({ skip: true }),
+    )
     expect(mockGateRelease).toHaveBeenCalled()
     expect(mockGateHold).not.toHaveBeenCalled()
   })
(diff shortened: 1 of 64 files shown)


grimen (changes requested):
## Findings

**1. The account that signs the mode push and the account it is recorded under can differ (blocking).**

`sdk` comes from `useSelfCustodialWallet()` while `activeAccountId` and `accountMode` come from persistent state, and `useSdkLifecycle` only clears `sdk` in its effect cleanup ([use-sdk-lifecycle.ts:299](https://github.com/blinkbitcoin/blink-mobile/blob/91126171e62f27342833aa18a2d2051f85194556/app/self-custodial/hooks/use-sdk-lifecycle.ts#L299)). Because `AccountModeSyncMount` is a descendant of the provider, its effects run before that cleanup, so on the commit where the user switches accounts the hook sees the new account id alongside the previous account's connected SDK.

*Example:* Account A is Enhanced and confirmed with a live Lightning Address, account B is Anon and unconfirmed. On the switch, `isPushDue` is true, so [use-account-mode-sync.ts:59](https://github.com/blinkbitcoin/blink-mobile/blob/91126171e62f27342833aa18a2d2051f85194556/app/self-custodial/hooks/use-account-mode-sync.ts#L59) signs with A's `identityPubkey`, posts `anon` to `/lnurlpay/{A}/mode` — silencing A's Lightning Address — and records the confirmation under B, which is then never pushed. The resolve effect at [:47](https://github.com/blinkbitcoin/blink-mobile/blob/91126171e62f27342833aa18a2d2051f85194556/app/self-custodial/hooks/use-account-mode-sync.ts#L47) has the same defect in reverse, adopting A's server mode as B's local mode.

*Proposal:* One option could be to gate both effects on `connectedAccountId === activeAccountId` — `useSdkLifecycle` already returns `connectedAccountId`, it just isn't placed on the wallet context.

**2. A dropped request is treated as a served "unknown region" verdict and sticks for the session.**

[use-account-restrictions.ts:175](https://github.com/blinkbitcoin/blink-mobile/blob/91126171e62f27342833aa18a2d2051f85194556/app/hooks/use-account-restrictions.ts#L175) falls back to `RESTRICTED_UNKNOWN_REGION` whenever `data` is absent, and the query's `error` is never read. With `fetchPolicy: "no-cache"`, the only thing that re-issues the query is the foreground refetch in [app-state.tsx:112](https://github.com/blinkbitcoin/blink-mobile/blob/91126171e62f27342833aa18a2d2051f85194556/app/navigation/app-state.tsx#L112).

*Example:* A user foregrounds the app during a brief connectivity gap; the refetch fails, and their Dollar balance stays hidden and the transfer button disabled for the rest of the session even after the network recovers — they have to background and foreground again.

*Proposal:* One option could be to read `error` separately from a served no-verdict answer, so a transport failure retries rather than latching; fail-closed on a server-delivered unknown region can stay as-is.

**3. After a lookup reset, a failed re-lookup reports the previous session's country as settled.**

`useIpCountryLookup` now re-runs on `lookupGeneration` and clears `hasLookupFinished`, but the resolve callback writes only on success while [use-device-location.ts:218](https://github.com/blinkbitcoin/blink-mobile/blob/91126171e62f27342833aa18a2d2051f85194556/app/hooks/use-device-location.ts#L218) marks the lookup finished unconditionally, leaving the stale `ipCountryCode` in place.

*Example:* The migration balances preview ([use-migration-balances-preview.ts:75](https://github.com/blinkbitcoin/blink-mobile/blob/91126171e62f27342833aa18a2d2051f85194556/app/screens/account-migration/hooks/use-migration-balances-preview.ts#L75)) is on screen, the user switches networks and foregrounds, `resetIpCountryLookup()` fires, the fresh lookup fails, and the hook reports `isSettled: true` with the country resolved under the old connection — the cross-session latch this PR sets out to remove.

*Proposal:* One option could be to clear `ipCountryCode` when the generation changes, so a failed re-lookup reports unresolved rather than stale.

## Test Coverage Gaps

- `__tests__/self-custodial/hooks/use-account-mode-sync.spec.ts` never varies the account id while a live `sdk` is present, so the mismatch in finding 1 passes through both effects untested.
- No test covers a failing `custodialRestrictions` query separately from a served no-verdict answer, so the latch in finding 2 is invisible to the suite.
- No test re-runs `useIpCountryLookup` after a generation bump with a failing lookup, which is the path in finding 3.

grimen (approved):

PR #4146, opened 2026-08-12
fix(scan): stop the gallery picker from crashing the app

### LINKED ISSUE #4095: [Android] Crash: image-picker double callback invoke on QR-from-gallery flow (SIGABRT, "callback arg cannot be called more than once")
### Summary

Production Android crash (SIGABRT): React Native aborts with `callback arg cannot be called more than once` (`JavaTurboModule.cpp:125`). The offending module is `react-native-image-picker` (we're on 7.1.2), and the app call site is the **"pick QR image from gallery"** flow on the scan screen (`app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx` → `launchImageLibrary`).

Symbolized stack (top frames):

```
#04 google::LogMessageFatal::~LogMessageFatal()
#05 facebook::react::createJavaCallback(...) lambda [JavaTurboModule.cpp:125]
#06 facebook::react::JCxxCallbackImpl::invoke
#10 com.facebook.react.bridge.CxxCallbackImpl.invoke
#12 com.imagepicker.e.b        (R8-obfuscated onAssetsObtained executor lambda)
#16 com.imagepicker.d.run
#19 java.util.concurrent.ThreadPoolExecutor.runWorker
```

### Root cause

Race in the library's `ImagePickerModuleImpl.onAssetsObtained`: the JS callback field is only cleared *inside* the async executor task, so a re-entrant/duplicate `onActivityResult` delivered before the task runs schedules a second task, and both invoke the same callback — the second invoke trips RN's fatal single-invocation guard. The catch block in the same lambda can also double-invoke. Full analysis in the upstream report.

**Upstream issue (filed by us, with suggested fix):** react-native-image-picker/react-native-image-picker#2414
Related iOS-flavored report: react-native-image-picker/react-native-image-picker#2390

### Why upgrading doesn't help

The race is still present on upstream `main` (latest release 8.2.1) — verified against the current source. Bumping the dependency alone won't fix the crash.

### Proposed fix

Add a `patch-package` patch to `ImagePickerModuleImpl.java` that takes the callback atomically before submitting the executor task:

```java
void onAssetsObtained(List<Uri> fileUris) {
    final Callback cb;
    synchronized (this) {
        cb = this.callback;
        this.callback = null;
    }
    if (cb == null) return;
    Executors.newSingleThreadExecutor().submit(() -> {
        try {
            cb.invoke(getResponseMap(fileUris, options, reactContext));
        } catch (RuntimeException exception) {
            cb.invoke(getErrorMap(errOthers, exception.getMessage()));
        }
    });
}
```

The patch can be dropped once upstream ships a fix for #2414.

### Environment

- react-native-image-picker 7.1.2
- react-native 0.85.3 (new architecture)
- Android arm64, production/R8 build

esaugomez31:
## Summary

Picking an image from the gallery on the scan screen could kill the app, in two unrelated ways. This fixes both, because they share a single entry point and a single test path: the gallery button on the scan screen.

The first is the crash the ticket describes. The second was found while trying to reproduce the first, and turned out to be the one that reproduces on demand.

Closes blinkbitcoin/blink-mobile#4095

## The reported crash: a callback invoked twice

`ImagePickerModuleImpl.onAssetsObtained` clears its callback field inside the executor task rather than before submitting it. A duplicate `onActivityResult` delivered before that task runs therefore passes the null check in `onActivityResult` and schedules a second task holding the same callback. React Native aborts the process the moment a callback is invoked more than once, which is the `SIGABRT` with `callback arg cannot be called more than once` in the report.

The patch claims the callback under a lock before the method returns, so the hand-off is single-shot. This also closes two smaller holes in the same method: the old `finally { callback = null }` wiped the callback belonging to a *second* pick, and the `resultCode != RESULT_OK` fall-through reached the switch with a null callback.

The cancel path had the same defect and is fixed the same way. `onActivityResult` invoked the callback inside a `try` and then invoked *the same callback again* from its own `catch`, so the handler for a failed hand-off was itself the abort it was handling. Its `return` also sat inside the `try`, leaving the failure path to fall through to the switch below. It now claims the callback under the same lock, settles the response before a single `invoke`, and returns on every route out.

Upgrading does not help. The same code is on upstream `main` at 8.2.1, verified against the source, so a patch is the only route until the upstream fix lands.

## The crash found while testing: the decoder runs out of memory

`RNQRGenerator.detect` was handed the picked image at whatever resolution the gallery returned. The decoder holds the decoded bitmap and allocates one integer per pixel on top of it to build its luminance source, so its peak is eight bytes per pixel of whatever it receives. A 12000x12000 photo asks for 576MB against a heap that caps out near 200MB, so the app dies before it can read anything:

```
java.lang.OutOfMemoryError: Failed to allocate a 576000016 byte allocation
  at RNQrGeneratorModule.createLuminanceSource(RNQrGeneratorModule.java:394)
  at RNQrGeneratorModule.scanQRImage(RNQrGeneratorModule.java:301)
```

File size is irrelevant; pixel count is what matters. The image used to reproduce this is 3.8MB on disk. Any phone shooting at 100MP or above is in range.

Four changes were needed, and none of them works without the others:

- The picker is asked for at most 2048px on the longest side, which puts the decoder's peak at 34MB and keeps it inside the 96MB heap of a budget device. This bound is below what a modern camera produces, so unlike a looser bound it does re-encode an ordinary photo rather than passing it through untouched. That is the deliberate tradeoff: a code filling the frame still lands at eleven pixels per module even on a version-40 QR, the densest there is, while a code occupying a small corner of a very large photo is the case given up. A 4096 bound would preserve those, but at a 134MB peak it is too close to the ceiling it exists to stay under.
- `Utils.resizeImage` decoded the full-resolution bitmap before scaling it down, so bounding the request alone would only have moved the same allocation from the decoder into the picker, where it is worse: `OutOfMemoryError` is an `Error`, so neither `catch (Exception)` there nor `catch (RuntimeException)` in `onAssetsObtained` catches it, the executor buries it in an unread `Future`, and the promise never settles. The gallery button would have gone dead instead of crashing. The patch subsamples during the decode, so the full bitmap is never materialised. `onAssetsObtained` now also catches `Throwable` and settles its response before the single `invoke`, so an `Error` from that decode still reaches JS instead of hanging the promise, and the per-pick executor is shut down rather than leaking a thread on every gallery open.
- The same hole existed on iOS, in the opposite direction: passing `maxWidth`/`maxHeight` at all turns on `ImagePickerUtils resizeImage`, whose `[image drawInRect:]` materialises the source at full resolution. Bounding the request would have introduced a full-resolution decode where the previous `maxWidth: 0` returned early and passed the original `NSData` through undecoded. The patch adds `ImagePickerUtils boundedImage`, which downsamples during the decode with `CGImageSourceCreateThumbnailAtIndex` before `resizeImage` runs.
- `getOrientation` returns null for an image with no EXIF orientation tag, and both callers dereference it. The resulting `NullPointerException` is swallowed by `resizeImage`, which returns the original uri as if no resize had been requested. Screenshots have no orientation tag, and a screenshot is a common way to keep a QR, so without this the fix would have skipped its most likely case. It now falls back to `ORIENTATION_UNDEFINED`, the neutral value both callers already treat as "leave it alone".

## Also fixed

Only `errorCode === "permission"` was surfaced. Every other code, including the `"others"` the native side returns when `getResponseMap` throws, arrived with no assets and fell through in silence: the button read as dead and nothing reached error reporting. Those now report and tell the user. The message stays generic rather than reusing "we could not find a QR code in the image", which would be a claim about a photo the app never got to look at.

The screen's catch narrowed on `err instanceof Error`, so a native module rejecting with a plain object or string, or an `Error` crossing a realm boundary, produced no alert and nothing reported. It now wraps whatever it caught with `toError`. The detail goes to error reporting rather than into the dialog, which would otherwise show the user a serialized native payload.

The gallery button did not check `pending`, so a QR decoded from a picked image was dropped in silence by `processInvoice`'s guard whenever a scan was already in flight: no alert, nothing reported, the same dead-button reading the rest of this change removes. It is now disabled and dimmed while a scan is in flight, which says the same thing the guard does but before the user spends a trip through the gallery on it.

## Test plan

- `yarn jest __tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx`: 20 passed. Nine are new and cover a flow that had none: the image is bounded before the decoder sees it, a QR found in the picked image resolves, an image with no QR tells the user, an empty pick resolves nothing, the picker stays shut while a scan is in flight, a picker rejection is surfaced, a rejection that is not an `Error` is surfaced, a refused permission is surfaced, and any other picker failure is reported rather than swallowed
- The in-flight test was checked against its own absence: with the guard removed it fails `Expected 0 calls, received 1`, so it pins the behaviour rather than following it
- `yarn jest __tests__/screens/send-bitcoin-screen/`: 90 passed across 10 suites, so nothing else on this screen moved
- Device (Android emulator), against a purpose-built 12000x12000 image: the crash reproduced three times before the change, with the same 576MB allocation each time. After the change the app survives, the decoder runs to completion and reports `NotFoundException` because the test image genuinely holds no QR. `javap` on the built `Utils.class` confirms the patched method is in the installed APK rather than a cached one

- Device (Android emulator), cancel path: `cancelCallback` confirmed present in the installed dex rather than a cached class, then 24 gallery cycles (12 cancel, 6 pick, 6 rapid double-back). Logcat shows 24 `PICK_IMAGES` intents for 24 cycles, so no cycle was a no-op. PID unchanged throughout, zero `SIGABRT`, zero "called more than once", zero ANRs, and the screen ends the run with camera and gallery both live

## Not verified

The iOS half is not device-tested; there is no macOS here. It is covered only by the compile in `e2e:build ios.sim.debug`. The reasoning above is from the upstream source, not from a run.

Neither double-invoke fix has a reproduction. Triggering either needs `invoke` itself to throw, which cannot be forced by hand, and the race did not reproduce across three rounds on a physical device with "Don't keep activities" enabled. Both rest on the code path being plainly wrong and on the upstream analysis. The 24-cycle run above proves the cancel path did not regress, which was the actual risk in touching it, not that the double-invoke is closed.

## Closed findings

Raised in review, checked, and not applying here. Recorded so they are not re-opened:

- *The new error alert leaves `pending` stuck if dismissed with the back button.* It cannot: `Alert.alert` hardcodes `cancelable: false` in its Android config and overrides it only when the caller passes `options.cancelable`, which this screen does not. The dialog is not dismissible, so the button's `onPress` is the only way out. See `react-native/Libraries/Alert/Alert.js:90`.

## Follow-ups, deliberately not in this PR

Real, but in code this PR does not touch. Keeping them out keeps the diff reviewable:

- `Utils.resizeImage` catches `Exception` and returns the original uri, so a failure there hands JS an unbounded image.
- `collectUrisFromData(data)` dereferences a null `Intent` under `RESULT_OK`, which some OEM galleries return.
- GIFs skip the bound on iOS; upstream excludes them on purpose to avoid breaking the animation.

## Note for whoever bumps this dependency

`postinstall` runs `patch-package --error-on-fail` against a `^7.1.2` range, so a future 7.x release will fail installs until the patch is regenerated. Same tradeoff as the existing camera-kit and modal patches. The double-invoke half can be dropped once upstream ships its fix; the subsampling and orientation halves are not filed upstream yet.

### DIFF
--- a/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
+++ b/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
@@ -104,6 +104,12 @@ jest.mock("@app/utils/toast", () => ({
   toastShow: (...args: unknown[]) => mockToastShow(...args),
 }))
 
+const mockReportError = jest.fn()
+jest.mock("@app/utils/error-logging", () => ({
+  ...jest.requireActual("@app/utils/error-logging"),
+  reportError: (...args: readonly unknown[]) => mockReportError(...args),
+}))
+
 const alertSpy = jest.spyOn(Alert, "alert").mockImplementation(() => {})
 
 beforeAll(() => {
@@ -355,7 +361,7 @@ describe("ScanningQRCodeScreen", () => {
     }
 
     /**
-     * The decoder allocates one integer per pixel of whatever it is handed, so an
+     * The decoder holds the decoded bitmap and one integer per pixel on top of it, so an
      * unbounded photo is an out-of-memory crash rather than a slow read.
      */
     it("bounds the picked image before the decoder ever sees it", async () => {
@@ -364,8 +370,8 @@ describe("ScanningQRCodeScreen", () => {
 
       expect(mockLaunchImageLibrary).toHaveBeenCalledWith({
         mediaType: "photo",
-        maxWidth: 4096,
-        maxHeight: 4096,
+        maxWidth: 2048,
+        maxHeight: 2048,
       })
     })
 
@@ -417,7 +423,33 @@ describe("ScanningQRCodeScreen", () => {
       const screen = await renderScreen()
       await openGallery(screen)
 
-      await waitFor(() => expect(alertSpy).toHaveBeenCalled())
+      await waitFor(() =>
+        expect(alertSpy).toHaveBeenCalledWith("Unexpected error occurred"),
+      )
+      expect(mockReportError).toHaveBeenCalledWith(
+        "scanning-qrcode",
+        new Error("picker exploded"),
+      )
+    })
+
+    /**
+     * A native module can reject with something that is not an Error, and narrowing the
+     * catch on instanceof would drop those rejections with no alert and nothing reported.
+     * What it rejected with belongs in error reporting, not in a dialog.
+     */
+    it("surfaces a picker rejection that is not an Error", async () => {
+      mockLaunchImageLibrary.mockRejectedValue({ code: "E_PICKER_UNAVAILABLE" })
+
+      const screen = await renderScreen()
+      await openGallery(screen)
+
+      await waitFor(() =>
+        expect(alertSpy).toHaveBeenCalledWith("Unexpected error occurred"),
+      )
+      expect(mockReportError).toHaveBeenCalledWith(
+        "scanning-qrcode",
+        new Error('{"code":"E_PICKER_UNAVAILABLE"}'),
+      )
     })
 
     it("tells the user when the image library permission was refused", async () => {
@@ -440,6 +472,10 @@ describe("ScanningQRCodeScreen", () => {
       await openGallery(screen)
 
       await waitFor(() => expect(alertSpy).toHaveBeenCalled())
+      expect(mockReportError).toHaveBeenCalledWith(
+        "scanning-qrcode",
+        new Error("Image library failed: others Unsupported file type"),
+      )
       expect(mockToastShow).not.toHaveBeenCalled()
       expect(mockDetect).not.toHaveBeenCalled()
     })
--- a/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
+++ b/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
@@ -455,6 +455,23 @@ describe("ScanningQRCodeScreen", () => {
       expect(mockResolveDestination).not.toHaveBeenCalled()
     })
 
+    /**
+     * processInvoice drops whatever it is handed while a scan is in flight, so a picker
+     * opened now would cost the user a trip through the gallery and end in silence.
+     */
+    it("stays shut while a scan is already in flight", async () => {
+      mockResolveDestination.mockReturnValue(new Promise(() => {}))
+
+      const screen = await renderScreen()
+      await act(async () => {
+        lastReadCode?.({ nativeEvent: { codeStringValue: "lnbc1inflight" } })
+      })
+
+      await openGallery(screen)
+
+      expect(mockLaunchImageLibrary).not.toHaveBeenCalled()
+    })
+
     it("surfaces a picker failure instead of dying on it", async () => {
       mockLaunchImageLibrary.mockRejectedValue(new Error("picker exploded"))
 
--- a/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
+++ b/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
@@ -353,6 +353,44 @@ describe("ScanningQRCodeScreen", () => {
     expect(mockResolveDestination).toHaveBeenCalledTimes(1)
   })
 
+  /**
+   * pending gates every scan and is only cleared from this alert, so a rejection that
+   * skipped the alert left the camera and the gallery button dead for the whole screen.
+   */
+  it("keeps scanning alive when resolving rejects with something that is not an Error", async () => {
+    mockResolveDestination.mockRejectedValue({ code: "E_RESOLVE" })
+
+    await renderScreen()
+    await fireScan("lnbc1first")
+
+    await waitFor(() =>
+      expect(alertSpy).toHaveBeenCalledWith(
+        "Unexpected error occurred",
+        "",
+        expect.anything(),
+      ),
+    )
+    expect(mockReportError).toHaveBeenCalledWith(
+      "scanning-qrcode",
+      new Error('{"code":"E_RESOLVE"}'),
+    )
+
+    const [, , buttons] = alertSpy.mock.calls[0]
+    await act(async () => {
+      buttons?.[0].onPress?.()
+    })
+
+    mockResolveDestination.mockResolvedValue({
+      valid: true,
+      destinationDirection: DestinationDirection.Send,
+      validDestination: { paymentType: "Lightning" },
+      createPaymentDetail: jest.fn(),
+    })
+    await fireScan("lnbc1second")
+
+    expect(mockResolveDestination).toHaveBeenCalledTimes(2)
+  })
+
   describe("picking a QR from the gallery", () => {
     const openGallery = async (screen: RenderAPI) => {
       await act(async () => {
--- a/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
+++ b/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx
@@ -1,7 +1,13 @@
 import React from "react"
 import { Alert } from "react-native"
 import { Network as mockSparkNetwork } from "@breeztech/breez-sdk-spark-react-native"
-import { act, render, waitFor } from "@testing-library/react-native"
+import {
+  act,
+  fireEvent,
+  render,
+  waitFor,
+  type RenderAPI,
+} from "@testing-library/react-native"
 
 import { loadLocale } from "@app/i18n/i18n-util.sync"
 import { Network } from "@app/graphql/generated"
@@ -81,13 +87,21 @@ jest.mock("@react-native-clipboard/clipboard", () => ({
   default: { getString: jest.fn().mockResolvedValue("") },
 }))
 
+const mockDetect = jest.fn()
 jest.mock("rn-qr-generator", () => ({
   __esModule: true,
-  default: { detect: jest.fn().mockResolvedValue({ values: [] }) },
+  default: { detect: (...args: unknown[]) => mockDetect(...args) },
 }))
 
+const mockLaunchImageLibrary = jest.fn()
 jest.mock("react-native-image-picker", () => ({
-  launchImageLibrary: jest.fn().mockResolvedValue({ assets: [] }),
+  launchImageLibrary: (...args: unknown[]) => mockLaunchImageLibrary(...args),
+}))
+
+const mockToastShow = jest.fn()
+jest.mock("@app/utils/toast", () => ({
+  ...jest.requireActual("@app/utils/toast"),
+  toastShow: (...args: unknown[]) => mockToastShow(...args),
 }))
 
 const alertSpy = jest.spyOn(Alert, "alert").mockImplementation(() => {})
@@ -129,6 +143,8 @@ describe("ScanningQRCodeScreen", () => {
     lastReadCode = null
     mockScanContext.mockReturnValue(custodialScanContext)
     mockSelfCustodialWallet.mockReturnValue({ sdk: null })
+    mockLaunchImageLibrary.mockResolvedValue({ assets: [] })
+    mockDetect.mockResolvedValue({ values: [] })
   })
 
   it("calls resolveDestination with sdk=null when active wallet is custodial", async () => {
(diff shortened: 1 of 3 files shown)


grimen (changes requested):
## Findings

**1. The size bound still lets an out-of-memory error escape unhandled (blocking).**
The executor lambda in `patches/react-native-image-picker+7.1.2.patch:31` catches only `RuntimeException`, so the very allocation this patch bounds can still die as an `Error` and leave the JS promise unsettled. `launchImageLibrary` resolves only from the native callback (`src/platforms/native.ts`), so if `cb.invoke` never runs, nothing settles.
*Example:* a 12000×12000 pick gives `inSampleSize` 2 (the largest factor keeping both axes ≥ 4096), so `decodeStream` allocates 144MB and `createScaledBitmap` another 67MB; on the ~200MB heap the PR body cites, the resulting `OutOfMemoryError` bypasses the catch, dies inside an unread `Future`, and the gallery button goes permanently dead with no alert and nothing in error reporting.
*Proposal:* one option could be catching `Throwable` and falling back to `getErrorMap(errOthers, ...)` so the hand-off always settles.

**2. The 4096 bound leaves the decoder's peak close to the ceiling it is meant to stay under (blocking).**
At the bound, `rn-qr-generator` allocates a 4096² bitmap (67MB) plus an `int[width*height]` in `createLuminanceSource` (another 67MB), and `android/app/src/main/AndroidManifest.xml` sets no `largeHeap`. The library's own guard does not help: `canBitmapFitInMemory` compares `getBitmapSize` in megabytes against `availableMemory()` in bytes, so it is always true and `getRightScale` always returns 1 — which is why the reported trace shows an unsampled 576MB allocation.
*Example:* on a budget device with a 96–128MB heap, a picked 4096² image, or a 4000×3000 photo that `shouldResizeImage` passes through untouched, reproduces the same crash this PR fixes.
*Proposal:* a 2048 bound still leaves roughly 8×8 pixels per module on a version-40 QR and cuts the peak fourfold, if that tradeoff is acceptable.

**3. The iOS half of the size bound is unpatched and newly forces a full-resolution decode (blocking).**
Passing `maxWidth`/`maxHeight` turns on `ImagePickerUtils resizeImage` in `mapImageToAsset`, which runs `UIGraphicsBeginImageContext(newSize)` plus `[image drawInRect:]` and materialises the source pixel buffer at full resolution. With the previous `maxWidth: 0`, that function returned early and the original `NSData` was passed through undecoded.
*Example:* the same 144MP pick that Android now subsamples costs roughly 576MB on iOS, and the test plan in the PR body is Android-only.
*Proposal:* one option could be bounding iOS in the patch via a downsampled `CGImageSourceCreateThumbnailAtIndex`; otherwise stating explicitly that iOS is untested here would set expectations.

**4. One silent path survives the sweep against the dead gallery button.**
The catch at `app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:364` reports and alerts only when `err instanceof Error`, so any other rejection value falls through to nothing at all.
*Example:* a native module rejecting with a plain object or string, or a cross-realm `Error`, produces no alert and no `reportError` — the exact failure mode the rest of this change removes.
*Proposal:* an `else` branch wrapping the value in an `Error` before reporting would close it.

## Test Coverage Gaps

- The reporting half of the new `if (result.errorCode)` branch is unverified: `__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx:133` asserts only `alertSpy`, and `reportError` is never mocked or asserted anywhere in the spec — removing the `reportError` call keeps all 17 tests green.

kngako (approved):
No blocking findings. Here's what Claude verified and what followed by possible minor changes.

## Verification done

- **Patch integrity** — `react-native-image-picker+7.1.2.patch` reverse-applies cleanly from the installed tree and forward-applies cleanly to pristine 7.1.2 (checked in a scratch copy). `postinstall` runs `patch-package --error-on-fail`, and every CI workflow does `yarn install --frozen-lockfile`, so the patch is exercised in CI.
- **Java compiles** — `ReadableMap` is imported; `getResponseMap`/`getErrorMap`/`getCancelMap` all return `ReadableMap`; `cb` is effectively final for the lambda; `response` is definitely assigned on both branches.
- **The Android bound actually reaches the decoder** — `getResponseMap` → `resizeImage` is on the library path, and `shouldResizeImage` was previously returning `false` (maxWidth 0, quality 100), so passing 2048 is what turns the resize on. `deleteFile` inside `resizeImage` removes the app-private copy from `getAppSpecificStorageUri`, not the user's photo.
- **`calculateInSampleSize` is correct for rotated images** — `origDimens`/`newDimens` are both in display orientation, and the AND is symmetric across the swap, so the subsequent `createScaledBitmap` only ever shrinks in both branches.
- **The iOS bound is effective** — the PHPicker path builds the image via `[[UIImage alloc] initWithData:]`, which is lazy, so `boundedImage` genuinely preempts the full decode. HEIC picks fall through `getFileType`'s `default: @"jpg"`, so they get re-encoded from the bounded image rather than passed through at full size. Orientation survives (`…WithTransform:@YES` → upright CGImage → `drawInRect:` keeps it upright).
- **User cancel does not hit the new alert** — `getCancelMap()` returns `didCancel` with no `errorCode`, so it falls past both new guards and stays silent.
- **Freeze is genuinely closed** — RN sets `cancelable: false` on Android alerts, so the OK button is the only way out and `setPending(false)` always runs.
- 20/20 tests in the spec, 90/90 across `__tests__/screens/send-bitcoin-screen`, `tsc --noEmit` clean, eslint clean.

## Worth changing

1. **`reportError("scanning-qrcode", toError(err))` throws away the attribution** — [[scanning-qrcode-screen.tsx:284](https://claude.ai/epitaxy/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:284)](app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:284), [[:326](https://claude.ai/epitaxy/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:326)](app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:326), [[:377](https://claude.ai/epitaxy/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:377)](app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:377). `reportError` only uses its first argument when the second isn't an `Error`; pre-wrapping with `toError` guarantees it never is, so the operation string is inert. These are the only 3 of 118 call sites in the app that pre-wrap — every other one passes the raw `unknown`. Net effect on the case the branch just started reporting: Crashlytics gets `{"code":"E_PICKER_UNAVAILABLE"}` with a stack rooted in `error-reporting.ts` and nothing naming the screen. Keeping the payload *and* the tag is a one-liner, e.g. `reportError("scanning-qrcode", new Error(\`scanning-qrcode: ${toError(err).message}\`))` — the two tests pinning the current message would need the same prefix.

2. **`showImagePicker` has no re-entrancy guard of its own** — [[:331](https://claude.ai/epitaxy/app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:331)](app/screens/send-bitcoin-screen/scanning-qrcode-screen.tsx:331). `disabled={pending}` covers a scan in flight, not a picker already open, and `pending` is false for the whole time the gallery is up. A double-tap calls `launchImageLibrary` twice; on Android the second `this.callback = callback` orphans the first, so that promise never settles. Harmless now (the patch turned what used to be the double-invoke abort into a silent orphan), but an `isPickerOpen` ref would close it.

3. **Retrying the *same* code after an error is still a no-op** — `scannedCache` is only cleared on blur, so once a payload has been seen, dismissing the alert and pointing at the same QR does nothing until the user leaves and re-enters. The new test at [[scanning-qrcode-screen.spec.tsx:360](https://claude.ai/epitaxy/__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx:360)](__tests__/screens/send-bitcoin-screen/scanning-qrcode-screen.spec.tsx:360) works around this with two different payloads. It's plausibly deliberate (the camera re-reads at frame rate, so without the cache an error would become an alert storm), but "scanning stays alive" only holds for a different code, which is worth stating somewhere.

4. **Android degrades back to unbounded on resize failure** — if `BitmapFactory.decodeStream` returns null or the mime is unsupported, `resizeImage`'s `catch (Exception)` returns the original URI and `rn-qr-generator` gets the full-size image, i.e. the OOM the change targets. Low likelihood, no worse than today.

5. **iOS still never nils `self.callback`** — the double-invoke class the Android half closes is untouched in `ImagePickerManager.mm`. I traced every `self.callback(...)` site and each one returns, so I found no reachable double-invoke; noting it only because the branch's framing is platform-neutral.

6. **Comment density** — the three multi-sentence rationale blocks in the TSX are much heavier than the rest of the file, which is essentially comment-free. Your call, but the *why* in them is good enough that I'd keep it and just tighten.

One operational note: this needs a fresh native build on both platforms to take effect — a JS-only reload picks up none of it.

grimen (approved):

PR #4147, opened 2026-08-13
fix(phone-input): restore the country flags the picker stopped rendering

### LINKED ISSUE #4132: bug(phone-picker): Fix flag display in phone number multi-field
## Problem
The phone number country-code multi-field is missing the country flag in at least two places after the 3.0 update:

- Onboarding phone number entry
- Send flow "send to SMS" phone number entry

Current state shows only the country code, for example `+503`, without the flag.

## Expected behavior
The phone picker should show the country flag alongside the country code, matching the pre-3.0 / desired design.

## Screenshots
Current state:

Current phone picker without flag

Desired state:

Desired phone picker with flag

## Scope
Research and fix the phone picker / phone number multi-field component so the flag renders wherever this component is used.

## Priority
Low priority visual bug.

## Source
Requested by Andrej from Slack message: https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1786299354817179

esaugomez31:
## Summary

The country button in every phone field lost its flag, and so did the country list behind it. Both come from the same cause, and neither surface raised an error while it happened.

`react-native-country-picker-modal` declares its flag props through `Component.defaultProps`. React 19 no longer applies that to function components, so each one arrives `undefined` at the app boundary: `Flag` returns `null` without `withFlagButton`, the country list drops its flag column without `withFlag`, and an unset `withEmoji` sends the picker to fetch flag images from a remote host instead of reading the emoji data it already ships. That is the "after the 3.0 update" in the report: the app moved to React 19.2.3.

Upgrading is not an option. The library's last release is 2.0.0 from June 2022, which is what we are on.

Closes blinkbitcoin/blink-mobile#4132

## Changes

**One place that owns the flag configuration**

- `app/components/phone-input/country-code-picker.tsx` (new): the country button and its picker, with the props the library stopped defaulting declared once. Three call sites had the same twenty-line `renderFlagButton` inline, so the bug had to be fixed three times and a fourth caller would have been born flagless the same way. They now pass only what genuinely differs between them: the selection handlers and the button style.
- `phone-input.tsx`, `phone-login-input.tsx` and `phone-registration-input.tsx` delegate to it.

**The country list was also frozen at mount**

The library reads its country list in an effect keyed on neither the list nor the codes, and an empty list means every country. Mounting before `useSupportedCountriesQuery` answered therefore left the modal offering all of them for good, which is why Antarctica, a country with no calling code, was on the list. The list now decides the identity of the subtree that reads it, so the supported set replaces the default one as soon as it arrives.

## Test plan

- `yarn jest __tests__/components/phone-input`: 21 passed. **100% on the new component**, and `phone-input.tsx` went from no spec at all to 96.5%
- The two assertions that pin the restored props fail when those props alone are removed, so they cannot pass by coincidence. Same for the refocus assertions
- The library is stubbed in the new spec on purpose: under Jest the dead `defaultProps` still resolve, because `React.createElement` kept that legacy path, so rendering the real library would show a flag either way and prove nothing. What is pinned is the props contract, which is the thing that broke
- `yarn tsc --noEmit` and `eslint` on the changed files: clean, and no console output in the new specs
- Consumer specs still green: `send-destination.spec.tsx` and `account-type-selection`
- Device (Android emulator): flag renders on the send flow button, flags render in the country list, and Antarctica is gone from it

## Screenshots

| Before | After |
| :---: | :---: |
| antes | despues |

The country list behind the button, which had lost its flags for the same reason:

| Before | After |
| :---: | :---: |
| modal-antes | modal-despues |

## Note for whoever touches this library next

Nine other components declare defaults the same way. Only the three used from app-level JSX were losing them, because elements the library builds internally still go through `React.createElement`, which kept resolving `defaultProps` in React 19. Anything new rendered from our side should pass what it needs explicitly rather than trust a default.

### DIFF
--- a/__tests__/components/phone-input/country-code-picker.spec.tsx
+++ b/__tests__/components/phone-input/country-code-picker.spec.tsx
@@ -5,6 +5,7 @@ import { fireEvent, render } from "@testing-library/react-native"
 import { CountryCodePicker } from "@app/components/phone-input/country-code-picker"
 
 const mockPicker = jest.fn()
+const mockPickerMount = jest.fn()
 const mockFlag = jest.fn()
 const mockOnOpen = jest.fn()
 
@@ -18,19 +19,27 @@ const mockOnOpen = jest.fn()
  */
 jest.mock("react-native-country-picker-modal", () => {
   const ReactActual = jest.requireActual("react")
+
+  const CountryPickerMock = (props: Record<string, unknown>) => {
+    mockPicker(props)
+    /** Counts mounts rather than renders: an effect with no dependencies runs again only
+     *  when this subtree is a new instance, which is exactly what the key has to force. */
+    ReactActual.useEffect(() => {
+      mockPickerMount()
+    }, [])
+    const renderFlagButton = props.renderFlagButton as (args: {
+      countryCode?: string
+      onOpen: () => void
+    }) => React.ReactNode
+    return renderFlagButton({
+      countryCode: props.countryCode as string | undefined,
+      onOpen: mockOnOpen,
+    })
+  }
+
   return {
     __esModule: true,
-    default: (props: Record<string, unknown>) => {
-      mockPicker(props)
-      const renderFlagButton = props.renderFlagButton as (args: {
-        countryCode?: string
-        onOpen: () => void
-      }) => React.ReactNode
-      return renderFlagButton({
-        countryCode: props.countryCode as string | undefined,
-        onOpen: mockOnOpen,
-      })
-    },
+    default: CountryPickerMock,
     Flag: (props: Record<string, unknown>) => {
       mockFlag(props)
       return ReactActual.createElement("Flag")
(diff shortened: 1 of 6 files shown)


grimen (changes requested):
## Findings

**1. The remount test passes whether or not the fix is present (blocking).**

The [`rebuilds the picker when the supported countries finally arrive`](https://github.com/blinkbitcoin/blink-mobile/blob/d86d7c37a2b0a1c5c67e9046f2c2e04af0e33f5e/__tests__/components/phone-input/country-code-picker.spec.tsx#L141-L155) test compares `mockPicker.mock.calls.length` before and after the rerender, but the mocked picker is a plain function component that React re-invokes on every re-render, remount or not, so the count always grows.

*Example:* Removing `key={countryListKey}` from `country-code-picker.tsx` and re-running the spec still passes, so the regression this PR fixes could return without any test failing.

*Proposal:* One option could be to pin mount identity instead of call count — for example having the mock increment a counter inside `useEffect(..., [])`, which only runs again on an actual remount.

grimen (approved):

PR #4152, opened 2026-08-13
fix(migration): stop the armed gate from restarting an unlock already in progress

### LINKED ISSUE #4150: bug: Android migration PIN loop after locked-without-checkpoint
## Summary

A Brazilian Android user is stuck in a PIN loop after trying to switch to non-custodial. Support payload reports `locked-without-checkpoint`, which suggests migration entered a locked state before a checkpoint was saved.

## User/account context

- User: Fabio Lourenco (`dr4gonshiryu`)
- Account ID: `20ff7132-c798-4ae8-8537-0066ffe59711`
- Region/language: Brazil, Portuguese
- Platform: Android
- App version: `3.0.25.50000870`
- Wallet state noted by support: L2 `ACTIVE`, 15 sats, USD 0.00
- Migration status noted by support: still `ACTIVE`, not `MIGRATED`/`CLOSED`

## Reported behavior

Original Portuguese report:

> Quando vou digitar o Pin aparece outra tela igual pra digitar o pin, toda vez que vou abrir a carteira acontece isso. Tudo pq fui mudar pra não custodíada.

English summary:

> When he goes to enter the PIN, another identical PIN screen appears. Every time he opens the wallet this happens. It started because he tried to switch to non-custodial.

## Expected behavior

After entering the PIN, the app should either continue the migration flow, resume from a saved checkpoint, or show a recoverable error state instead of returning to another identical PIN screen.

## Actual behavior

The app repeatedly shows the PIN screen and does not progress. The user is effectively locked out.

## Evidence

- Slack case source: https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1786631033775449?thread_ts=1786631033.775449&cid=C0AEYAALSCQ
- Screenshot/file from Slack: https://blinkbtc.slack.com/files/U0AE0J46RU5/F0BPZ2A4GNA/image.png

## Requester

Requested by Jonas (`@grimen`) in Slack.

esaugomez31:
Closes #4150.

## The bug

A Brazilian Android user reported that the PIN screen reappears empty while he is typing, every time he opens the app, and that it started when he tried to switch to non-custodial.

## Cause

`useMigrationBlocker` turns visible on a server answer (`migrationStatus`, read `no-cache` on every launch, with its loading state deliberately ignored). That answer lands a beat into the launch, by which time the user is already on the PIN screen. The armed-gate effect then reset the root stack, which tore that screen down and mounted a fresh, empty one.

Only an account with a server-side migration lock (or an armed wind-down gate) reaches that code path, which is why it started when he began the migration and why nobody else saw it.

## Video

Both recorded on an emulator against staging, with the network throttled to mobile latency. Two digits are typed as soon as the pad appears, and nothing is touched after that.

**Before** — the digits vanish on their own and the pad starts over

https://github.com/user-attachments/assets/742478e2-e101-42fe-b4d6-4e03c769dbc7

**After** — the digits survive, one entry gets through, and the flow lands on Contact support

https://github.com/user-attachments/assets/800b9aac-128a-45d0-8094-7eedee8c86f6

## The fix

The reset now looks at what is actually in front of it instead of guessing from the lock flag.

**It never tears down an unlock in progress.** The focused route is read off the state's own index, and judged by its `screenPurpose` rather than its name: Settings reuses the very same `pin` and `authentication` routes to SET a PIN or turn biometrics on while the app is already unlocked, and reading the name alone would mistake that for an unlock. The new `app/navigation/unlock-routes.ts` owns that answer, and the unlock screens now type their own options against the same list, so a fourth unlock step cannot be added to one and forgotten in the other.

**A reset an unlock held back retries on every navigation until the way is clear.** Releasing the lock is not the same event as leaving the unlock — the biometric prompt falls back to the PIN pad and steps back onto itself with the lock already down — so keying the retry on the lock would spend it too early. The two triggers do not carry the same authority: arming owns both of the reset's jobs, while a retry exists only to finish the pop, and must not bounce a session elsewhere on the strength of a lock its own flow never lowered (the three-strikes logout resets the stack without unlocking).

**A reset only happens when it would achieve something.** Two jobs qualify: popping whatever is stacked on top of the blocker, and routing a still-locked session through the unlock entry, which is the only thing that lowers the lock and so the only thing that drains the queued deeplink. A stack the blocker is not part of at all — signing in, unlocking, the landing screen a logout returns to — is left alone.

**A retry never pops the migration entry the armed gate lets through**, since that is the one screen that can legitimately appear above the blocker after the gate armed.

`blockerEntryRoute` is kept as it is on `main`: an earlier revision of this branch removed it, which stranded the lock raised and froze the queued migration deeplink for a session that signs in without ever unlocking. `isAppLockedRef` is now written by the two setters instead of by an effect, because the retry runs during a navigation dispatch, before that commit lands.

## Verification

Reproduced and re-verified on an Android emulator against staging, with the network throttled to mobile latency. The status query keeps its real timing; only its value is forced, so no server-side lock is armed on the test account.

Before:

```
PinScreen MOUNTED
migrationStatus resolved                  (+2.9s)
resetToBlocker -> authenticationCheck, isAppLocked: true
PinScreen UNMOUNTED
PinScreen MOUNTED                         (a second, empty screen)
```

After:

```
PinScreen MOUNTED
migrationStatus resolved                  (+4.2s)
                                          (no reset is even warranted: the cold-start
                                           stack has no blocker route to pop)
PinScreen UNMOUNTED                       (the user's own unlock)
```

Digits typed during the window survive, one PIN entry gets through, and the flow lands on the Contact support screen with `locked-without-checkpoint`.

93 tests pass across `__tests__/navigation` and `__tests__/screens/authentication-screen`; `unlock-routes.ts` is at 100% coverage and every branch added to the wrapper is covered. Each guard was mutation-tested by breaking it one at a time: 17 of the 18 mutations fail at least one spec. The one that does not is clearing the deferral when the kill-switch hides the blocker, which is defence in depth rather than behaviour — the re-arming path already recomputes it.

## Scope

This removes the PIN loop for everyone affected. It does not release a server-side lock, so an account that also lost its local checkpoint still lands on the support screen and needs backend to clear it.

Rebased over #4151, so the generated GraphQL types are no longer part of this diff.

### Known limitations

- Arming still pops routes the gate itself is allowed to open (the dollar transfer, the support screen) if the blocker's visibility flips while one of them is open. Only the retry path makes the allowance, since those screens cannot predate an arming. Behaviour that predates this change; a full allowlist of gate-sanctioned routes would be its own ticket.
- A resume unlock that falls back to the PIN pad steps back onto the biometric prompt with the app already unlocked and back blocked, so the user has to tap Unlock once more. That dead end predates this branch; the reset correctly keeps waiting while it is on screen.
- The arming decision is one-shot per visibility flip. A session that arms while the blocker is not in the stack, and only later reaches it while still locked, is not routed through the unlock. The flip is a once-per-launch event, so this is latent rather than reachable today.

### DIFF
--- a/__tests__/navigation/migration-flow-registration.spec.ts
+++ b/__tests__/navigation/migration-flow-registration.spec.ts
@@ -1,5 +1,7 @@
 import { readFileSync } from "fs"
 
+import { isMigrationRoute } from "@app/navigation/migration-routes"
+
 // The flow's screens navigate to each other by route name, and every screen spec mocks
 // navigation, so a route that was never registered still passes everywhere else and only
 // fails on device. Pinned with source-level assertions for the same reason the developer
@@ -19,6 +21,21 @@ describe("migration flow registration in root-navigator", () => {
   it("registers every step of the flow", () => {
     FLOW_ROUTES.forEach((route) => {
       expect(navigatorSource).toContain(`name="${route}"`)
+    })
+  })
+
+  it("keeps every registered migration screen inside the armed gate's allowance", () => {
+    /** The armed-gate reset pops whatever it does not recognise as the gate's own, and an
+     *  enumerated allowance is exactly what went wrong there: it named the deeplink entry
+     *  and missed the screens the gate opens itself. Read off the navigator so a screen
+     *  registered later is covered without anyone remembering to add it. */
+    const registeredRoutes = [
+      ...navigatorSource.matchAll(/name="(accountMigration\w+)"/g),
+    ].map(([, routeName]) => routeName)
+
+    expect(registeredRoutes.length).toBeGreaterThan(FLOW_ROUTES.length)
+    registeredRoutes.forEach((routeName) => {
+      expect(isMigrationRoute(routeName)).toBe(true)
     })
   })
 
--- a/__tests__/navigation/migration-routes.spec.ts
+++ b/__tests__/navigation/migration-routes.spec.ts
@@ -1,0 +1,27 @@
+import { isMigrationRoute } from "@app/navigation/migration-routes"
+
+describe("isMigrationRoute", () => {
+  it("recognises the deeplink entry the armed gate lets through", () => {
+    expect(isMigrationRoute("accountMigrationEntry")).toBe(true)
+  })
+
+  it("recognises the screens the armed gate opens itself, which the entry alone missed", () => {
+    /** Where a gate resuming a locked migration actually sends the user: a checkpoint step,
+     *  or contact support when there is nothing to resume. Coverage of every registered
+     *  screen is anchored on the navigator source, in migration-flow-registration.spec. */
+    expect(isMigrationRoute("accountMigrationContactSupport")).toBe(true)
+    expect(isMigrationRoute("accountMigrationTransferringFunds")).toBe(true)
+  })
+
+  it("rejects the screens the armed-gate reset exists to pop", () => {
+    expect(isMigrationRoute("Primary")).toBe(false)
+    expect(isMigrationRoute("scanningQRCode")).toBe(false)
+    expect(isMigrationRoute("conversionDetails")).toBe(false)
+  })
+
+  it("rejects the unlock routes, which must never be preserved above the blocker", () => {
+    expect(isMigrationRoute("authenticationCheck")).toBe(false)
+    expect(isMigrationRoute("authentication")).toBe(false)
+    expect(isMigrationRoute("pin")).toBe(false)
+  })
+})
--- a/__tests__/navigation/navigation-container-wrapper.spec.tsx
+++ b/__tests__/navigation/navigation-container-wrapper.spec.tsx
@@ -36,9 +36,9 @@ let mockOnStateChange: ((state: unknown) => void) | undefined
 jest.mock("@react-navigation/native", () => ({
   ...jest.requireActual("@react-navigation/native"),
   createNavigationContainerRef: () => ({
-    reset: (state: { routes: { name: string }[] }) => {
+    reset: (state: { index: number; routes: { name: string; params?: object }[] }) => {
       mockReset(state)
-      mockRootState = { index: 0, routes: state.routes }
+      mockRootState = { index: state.index, routes: state.routes }
     },
     isReady: () => mockIsReady,
     getRootState: () => mockRootState,
@@ -92,6 +92,7 @@ import {
   processLinkForAction,
   useAuthenticationContext,
 } from "@app/navigation/navigation-container-wrapper"
+import { MigrationSupportOrigin, MigrationSupportReason } from "@app/types/migration"
 import { AuthenticationScreenPurpose, PinScreenPurpose } from "@app/utils/enum"
 
 describe("processLinkForAction", () => {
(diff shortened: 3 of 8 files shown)


grimen (changes requested):
## Findings

**1. The deferred retry can reset past a still-locked app (blocking).**

[`mustRouteThroughUnlock = isAppLockedRef.current && isArming`](https://github.com/blinkbitcoin/blink-mobile/blob/5c739dc56b4e6f13f50d965ba5b146563a236aaa/app/navigation/navigation-container-wrapper.tsx#L186) evaluates to `false` for every `"unlock-cleared"` retry, because `isArming` is false on that path. A retry that fires while the app is still locked therefore resets to `Primary` instead of `authenticationCheck`, and the unlock screen is destroyed before the user has entered anything.

*Example:* On a resume relock the stack is `[Primary, X, authenticationCheck(isResume)]`; `migrationStatus` resolves and arming defers because an unlock sits in front; [`MigrationGate`](https://github.com/blinkbitcoin/blink-mobile/blob/5c739dc56b4e6f13f50d965ba5b146563a236aaa/app/screens/account-migration/to-non-custodial/migration-gate.tsx#L236) — mounted under the unlock screen inside `PrimaryNavigator` — settles and navigates to `accountMigrationContactSupport` from its once-per-mount effect, which fires `onStateChange`, runs the pending retry, and lands the reset on `Primary` with `isAppLocked` still `true`. The user reaches the app without PIN or biometric, and the queued migration deeplink stays frozen because it waits on `!isAppLocked`.

*Proposal:* One option could be to have the deferral carry the original trigger's authority, or at minimum re-consult `isAppLockedRef` when choosing the target rather than deriving it from `isArming`.

**2. The sanctioned-route allowance does not match where the armed gate navigates (blocking).**

The [whitelist](https://github.com/blinkbitcoin/blink-mobile/blob/5c739dc56b4e6f13f50d965ba5b146563a236aaa/app/navigation/navigation-container-wrapper.tsx#L180) names only `accountMigrationEntry`, but the armed gate never navigates there — it navigates to `accountMigrationContactSupport` and to the checkpoint step screens. The retry consequently resets away screens the gate itself opened. In the opposite case, when the whitelist does match, clearing `isResetDeferredRef` early abandons the deferred pop altogether.

*Example:* In the scenario above the retry resets the contact-support screen back to the gate root, which re-runs its resume effect on the fresh mount and navigates again; conversely, on a whitelist hit any pre-arming screen left between `Primary` and the migration entry survives and stays reachable by back-navigation over a closed account.

*Proposal:* If the intent is "leave anything the armed gate opened alone", the allowance could cover the gate's whole route family rather than the entry route alone, and the deferred pop could stay pending instead of being cleared on a match.

## Test Coverage Gaps

- No test exercises an `"unlock-cleared"` retry that fires while `isAppLocked` is still `true` — the 93 passing navigation and authentication-screen specs all cover paths where the retry runs after the lock is gone.
- No test covers a retry landing while the focused route is a gate-opened screen other than `accountMigrationEntry` (`accountMigrationContactSupport` or a checkpoint step).

grimen (approved):

PR #4154, opened 2026-08-13
perf(transaction-history): stop opening a transaction re-rendering every mounted row

### LINKED ISSUE #4137: perf(transaction-history): opening a transaction re-renders every mounted row and is slow
## Problem

Tapping a row in the transaction history is visibly laggy, and the lag scales with
how much history the account has. The delay is not in the details screen itself —
it is the list behind it. Opening a transaction triggers a re-render of *every*
row currently mounted in the `SectionList`, and each of those re-renders is far
more expensive than it needs to be.

On an account with a long history this is hundreds of rows re-rendering, each
rebuilding its stylesheets and constructing fresh `Intl.NumberFormat` instances,
before the details screen gets a frame. It is worst on Android, where Hermes'
`Intl` implementation is markedly slower than JSC's.

## Root causes

Five separate issues compound here. Each one alone is survivable; together they
turn a navigation push into a full-list re-render storm.

1. **Every row subscribes to navigation focus.**
   `app/components/transaction-item/transaction-item.tsx` calls `useIsFocused()`
   unconditionally. That hook re-renders *every* subscriber on each focus change,
   so the moment you open a transaction, all mounted rows re-render. The
   subscription only exists to drive the bounce-in animation, which applies to at
   most one row (`highlight`).

2. **A per-row arrow function defeats `React.memo`.**
   `app/screens/transaction-history/transaction-history-screen.tsx` passes
   `onPress={() => handleItemPress(item.id)}` from `renderItem`. A new closure
   identity on every render means `MemoizedTransactionItem` never bails out, so
   the memo wrapper is doing nothing.

3. **Inline style props rebuild every StyleSheet, every render.**
   The row calls `useStyles({ isFirst, isLast, isOnHomeScreen, highlight })` with
   an object literal. `makeStyles` memoizes on the props object's *identity*, so a
   fresh literal invalidates the cache and rebuilds all stylesheets in the file on
   each render of each row.

4. **`Intl.NumberFormat` is constructed per amount, per render.**
   `formatCurrencyHelper` in `app/hooks/use-display-currency.ts` builds a new
   formatter on every call. Constructing one is the single most expensive
   operation in rendering a row, and the options only ever vary by fraction-digit
   count — a handful of distinct values across the whole app.

5. **`windowSize` is left at the default.**
   The `SectionList` in the transaction history screen does not set `windowSize`,
   so RN's default of 21 keeps roughly ten screens of rows mounted on either side
   of the viewport. That default is what makes every list-wide re-render as
   costly as it is.

## Steps to reproduce

1. Sign in to an account with a long transaction history (several hundred
   transactions — the effect is proportional to history length).
2. Open **Transaction history**.
3. Scroll partway down so a healthy number of rows are mounted.
4. Tap any transaction.

**Expected:** the details screen pushes immediately.
**Actual:** a visible delay before the push; the deeper the scroll and the longer
the history, the worse it gets. Most pronounced on Android.

## Expected behaviour

Opening a transaction should be independent of how many rows are mounted and how
long the account's history is. Only the row that actually animates should
subscribe to navigation focus; unchanged rows should not re-render at all. Details should open/close seemingly instantly.

## Acceptance requirements

- [ ] Opening a transaction does not re-render rows whose props are unchanged.
- [ ] Only rows with `highlight` set subscribe to navigation focus; non-highlighted
      rows never call `useIsFocused`.
- [ ] `MemoizedTransactionItem` actually bails out on re-render with unchanged
      props — the list passes one stable `onPress` for all rows, and the row
      receives the `txid` as an argument rather than closing over it.
- [ ] Stylesheets are not rebuilt on every render of every row.
- [ ] At most one `Intl.NumberFormat` is constructed per distinct fraction-digit
      count, not one per formatted amount.
- [ ] `windowSize` is tuned down from the default while still leaving enough
      headroom that fast scrolling does not show blank cells.
- [ ] The bounce-in animation on the highlighted row still plays, unchanged, when
      arriving from a notification or deep link.
- [ ] No visual or behavioural change to the rows themselves — amounts, hidden-
      balance placeholders, dates, descriptions and icons all render as before.

## Notes

Seems to be affecting Android harder.

grimen:
Fixes #4137.

Opening a transaction pushed a re-render through every mounted row of the `SectionList` before the details screen got a frame, and each of those re-renders did more work than it needed to. Five changes across four files: cut the re-render at its source, make `React.memo` actually bail out, and make the remaining per-row work cheaper. No visual or behavioural change to the rows.

## 1. Focus subscription moved into a subcomponent only ever-highlighted rows mount

`TransactionItem` called `useIsFocused()` unconditionally. That hook re-renders *every* subscriber on each navigation focus change, so the instant you opened a transaction, every mounted row re-rendered. The subscription existed only to drive the bounce-in animation, which applies to at most one row.

Extracted a `BouncingRow` wrapper that owns `useIsFocused`, `useSharedValue`, `useBounceInAnimation` and `useAnimatedStyle`. The row body is built as a plain element and wrapped only once it has had reason to animate:

```tsx
// Latched: mounted the first time the row is highlighted, kept for its lifetime.
const everHighlighted = React.useRef(highlight)
if (highlight) everHighlighted.current = true

return everHighlighted.current ? <BouncingRow visible={highlight}>{row}</BouncingRow> : row
```

**Decision:** the wrapper is latched rather than conditioned directly on `highlight`. A plain `highlight ? <BouncingRow>… : row` changes the element *type* at that position, and React does not diff across a type change — so the moment a highlight cleared (which happens on every tap of an unseen transaction, via `seenTxIds`), the entire row subtree unmounted and rebuilt instead of re-rendering, discarding any bounce still in flight. Latching keeps the tree stable across that flip. The obvious alternative — always render `BouncingRow` and pass `highlight` in — was rejected because `useIsFocused` lives inside it and hooks cannot be conditional, so it would put a focus subscription, a shared value and an extra `Animated.View` back into *every* row, which is the exact cost this PR exists to remove.

**Decision:** `visible={highlight}` is passed rather than hardcoded `true`. Besides being what the latch requires, it restores the pre-extraction semantics: `useBounceInAnimation` resets `scale` and clears its pending timer when `visible` goes false, so a highlight clearing mid-bounce now unwinds cleanly.

**Decision:** rows that are never highlighted — the overwhelming majority — lose their wrapping `Animated.View`. That view only ever carried an identity transform, so layout is unchanged, but it does remove one node from the native view tree. A row that *has* been highlighted keeps its wrapper even after the highlight clears; that is the latch above, and it is one view on at most a couple of rows. Called out here because it's the one part of this PR that touches what actually gets mounted.

This is the single biggest win of the five.

## 2. `onPress` receives the txid, so the list can share one callback

The screen passed `onPress={() => handleItemPress(item.id)}` from `renderItem`. A fresh closure identity on every render meant `MemoizedTransactionItem` never bailed out — the memo wrapper was inert. The prop signature became `(txid: string) => void`, the row binds its own id, and the screen passes its stable `useCallback` directly.

**Decision:** the `onPress ? handlePress : undefined` guard is load-bearing. `handlePress` is always defined, so passing it unconditionally would make every row pressable — including the contact-list rows, which pass no `onPress` and must stay inert.

## 3. Memoized the object passed to `useStyles`

`useStyles({ isFirst, isLast, isOnHomeScreen, highlight })` was called with an object literal. `makeStyles` memoizes on the props object's *identity*, so a fresh literal invalidated the cache and rebuilt every `StyleSheet` in the file on every render of every row.

## 4. `Intl.NumberFormat` cached by fraction-digit count

`formatCurrencyHelper` constructed a new formatter on every call — once per amount, per row, per render. Construction is the most expensive operation in rendering a row, and Hermes' implementation is markedly slower than JSC's, which is why this hit Android hardest. A module-level `Map` keyed on fraction digits covers it, since that is the only option that ever varies (the locale is hardcoded `en-US`).

The cache is unbounded in principle but bounded in practice: the key space is the set of fraction-digit counts across supported display currencies, a handful of values.

The existing FIXME about using `.format` rather than `.formatNumber` still applies and moved to the call site — it previously sat inside the options literal, commenting on an option that doesn't exist.

## 5. `windowSize` reduced, render props hoisted

Both transaction `SectionList`s ran on RN's default `windowSize` of 21, which keeps roughly ten screens of rows mounted either side of the viewport — that default is what made every list-wide re-render as expensive as it was.

**Decision:** `7`, which still leaves three screens of headroom on each side. If fast scrolling ever shows blank cells, the fix is to raise this to 9, not to revert it.

**Stated plainly, because it is the one change here with a user-visible downside:** this is the only part of the PR not backed by a measurement. The benchmark measures mount and focus-change render counts, not scrolling, so the blank-cell risk on a fast fling on low-end Android is argued rather than observed — hence the device pass called for at the bottom. `getItemLayout` would remove the risk outright by letting the list skip measurement, but it was rejected here: both lists are `SectionList`s, where a correct `getItemLayout` has to account for section headers as well as rows, and the rows are not a fixed height (the `subtitle` variant is taller). Getting that subtly wrong produces worse scroll artefacts than the ones it would prevent. Note also that this reduction now compounds with §1–§4 rather than substituting for them: each retained cell is far cheaper than it was, so most of the win is already banked before `windowSize` is touched at all.

Alongside it, `renderSectionHeader` and `keyExtractor` (and `renderItem` on the contacts list) were inline literals recreated on every screen render; all are now `useCallback`s.

## Scope

The contacts transaction list (`contact-transactions.tsx`) had the same list-level pattern, so it got the same treatment. Its rows stay non-pressable via the guard in §2. Its new `useCallback`s sit above the component's early returns to keep hook order stable.

## Tests

**`__tests__/screens/transaction-history-screen.spec.tsx`** — the `MemoizedTransactionItem` stub now records the `onPress` identity it is handed. One test renders the screen, forces a re-render via the wallet filter dropdown, and asserts every row across both renders got the *same* function object. Verified this test fails when the per-row arrow is restored.

**`__tests__/components/transaction-item.test.tsx`** — the `@react-navigation/native` mock changed from `useIsFocused: () => true` to `jest.fn(() => true)` so the subscription is assertable, and the `IconTransaction` stub now counts its own mounts so a remount of the row subtree is assertable too. Five tests added: a non-highlighted row never calls `useIsFocused`, a highlighted row does, pressing a row calls `onPress` with its own `txid`, flipping `highlight` true→false leaves the mount count at 1, and that same flip hands `useBounceInAnimation` `visible: false` rather than dropping the wrapper. Verified the subscription test fails when all rows wrap in `BouncingRow`, and both latch tests fail against the plain ternary (mount count 2).

**`__tests__/hooks/use-display-currency.spec.tsx`** — one test spies on `Intl.NumberFormat`, formats 25 amounts at a fraction-digit count no other test in the file uses (7, so the module-level cache is cold on the first run and warm on any later one), and asserts at most one construction plus unchanged output. The spy is restored in an `afterEach` rather than at the end of the test body: jest is configured with neither `restoreMocks` nor `resetMocks` here, so a failing assertion would otherwise leave `Intl.NumberFormat` spied for the rest of the file and bury the real failure under every later formatting test. Confirmed by breaking the assertion deliberately — 1 failed, 39 passed.

**Rejected:** a component-level "memo bails out on unchanged props" test. It passed against the unfixed code too — `React.memo` was always there and the test's own props were stable — so it guarded nothing. The defect was at the call site, which is where the assertion now lives.

## Measured before/after

`__tests__/perf/transaction-list.bench.tsx` renders N **real** rows inside a **real** `NavigationContainer` and then navigates to a second screen — which is exactly what pushing the detail screen does — and counts what that costs the list behind it. Nothing on the measured paths is mocked: the real `useIsFocused` drives the subscription, styles go through the real `makeStyles`, amounts through the real `useDisplayCurrency`. Only `IconTransaction` is stubbed, purely so it can count row renders.

**Rows re-rendered when a transaction is opened** — the defect in the issue:

| History length | Before | After |
|---:|---:|---:|
| 50 rows | 50 | **0** |
| 200 rows | 201 | **0** |
| 400 rows | 401 | **0** |

Before scales linearly with how much history is mounted, which is exactly the reported symptom. After is zero at every size — not merely fewer, none. The highlighted row's wrapper still re-renders on focus, but its content element is unchanged so React bails before reaching the row body.

**Work done during the initial mount**, same harness:

| History length | `StyleSheet.create` before → after | `Intl.NumberFormat` before → after |
|---:|---:|---:|
| 50 rows | 100 → **50** | 200 → **2** |
| 200 rows | 401 → **200** | 802 → **2** |
| 400 rows | 801 → **400** | 1602 → **2** |

Stylesheet builds halve to exactly one per mounted row. Formatter construction collapses to **2 for the entire process lifetime**, not per size — one for each distinct fraction-digit count in play (0 for sats, 2 for USD). Those 2 are constructed on the first cold render and never again; every subsequent run in the same process measures 0. The harness reports that figure as `intlNumberFormatsFirstRun` alongside the full per-run array, rather than a median: the cache is module-level and survives the repeat loop, so a median across repeats would describe `BENCH_REPEATS` rather than a mount, and would not be comparable against the "before" side where every run constructs.

Every figure above was identical across all three repeats at every size.

**Wall-clock is deliberately not reported.** In this environment it is dominated by warm-up noise — 400 rows repeatedly timed *faster* than 200 — so it would be a misleading number. The render counts are exact and environment-independent, and they are the thing the issue is actually about.

**One methodology note, because it changes how the numbers should be read:** mount is asynchronous here (currency list, i18n, Apollo fragments all settle over several ticks). An early version of the harness measured before the tree was quiescent, which attributed late mount renders to the navigation and made the first run report 200 re-renders on the fixed code. The committed harness waits for the row-render count to stop moving before measuring, and the fixed code then reports 0 on every run including the cold one.

The harness is committed as `.bench.tsx`, which jest's `testRegex` (`test|spec`) excludes — it does not run in CI. That is deliberate (it is a manual A/B harness, not a test), with the consequence that nothing will report when it rots; its header now says so, so a future reader does not assume CI is keeping it honest. To reproduce:

```
BENCH_ROWS=200 BENCH_REPEATS=3 npx jest --testRegex "__tests__/perf/.*\.bench\.tsx$"
```

Switch between sides with `git checkout origin/main -- app/` and `git checkout HEAD -- app/`.

## Verification

```
yarn jest __tests__/components/transaction-item.test.tsx \
          __tests__/hooks/use-display-currency.spec.tsx \
          __tests__/screens/transaction-history-screen.spec.tsx \
          __tests__/screens/transaction-history-self-custodial.spec.tsx
```

70 passed, 4 suites. `tsc --noEmit` and `eslint` clean on all touched files.

Still worth a device pass before merge, particularly on Android: fast-scroll a long history to confirm `windowSize={7}` shows no blank cells, and arrive from a notification to confirm the bounce-in still plays and still settles when the highlight clears.

### DIFF
--- a/__tests__/components/transaction-item.test.tsx
+++ b/__tests__/components/transaction-item.test.tsx
@@ -76,8 +76,17 @@ jest.mock(
 )
 
 // --- child components that would otherwise require native modules ---
+// The mount counter is what makes a remount of the row subtree assertable:
+// re-rendering leaves it alone, unmounting and rebuilding bumps it.
+let mockIconMounts = 0
 jest.mock("@app/components/icon-transactions", () => ({
-  IconTransaction: () => null,
+  IconTransaction: () => {
+    const react = jest.requireActual("react")
+    react.useEffect(() => {
+      mockIconMounts += 1
+    }, [])
+    return null
+  },
 }))
 
 jest.mock("@app/components/transaction-date", () => ({
@@ -151,8 +160,10 @@ import { useIsFocused } from "@react-navigation/native"
 import { useHideAmount } from "@app/graphql/hide-amount-context"
 import { useAppConfig } from "@app/hooks"
 import { useDisplayCurrency } from "@app/hooks/use-display-currency"
+import { useBounceInAnimation } from "@app/components/animations"
 
 const mockUseFragment = useFragment as jest.Mock
+const mockUseBounceInAnimation = useBounceInAnimation as jest.Mock
 const mockUseIsFocused = useIsFocused as jest.Mock
 const mockUseHideAmount = useHideAmount as jest.Mock
 const mockUseAppConfig = useAppConfig as jest.Mock
@@ -179,6 +190,7 @@ const makeTx = (overrides = {}) => ({
 
 beforeEach(() => {
   jest.clearAllMocks()
+  mockIconMounts = 0
 
   mockUseFragment.mockReturnValue({ data: makeTx() })
 
@@ -296,6 +308,31 @@ describe("MemoizedTransactionItem", () => {
       render(<MemoizedTransactionItem txid="tx-1" highlight />)
 
       expect(mockUseIsFocused).toHaveBeenCalled()
+    })
+
+    it("does not remount the row when the highlight clears", () => {
+      // Tapping an unseen transaction marks it seen, which flips highlight
+      // true -> false. Swapping the element type at that position would throw
+      // the whole row subtree away and rebuild it.
+      const { rerender } = render(<MemoizedTransactionItem txid="tx-1" highlight />)
+      expect(mockIconMounts).toBe(1)
+
+      rerender(<MemoizedTransactionItem txid="tx-1" highlight={false} />)
+
+      expect(mockIconMounts).toBe(1)
+    })
+
+    it("stops the bounce when the highlight clears, without dropping the wrapper", () => {
+      const { rerender } = render(<MemoizedTransactionItem txid="tx-1" highlight />)
+      expect(mockUseBounceInAnimation).toHaveBeenLastCalledWith(
+        expect.objectContaining({ visible: true }),
+      )
+
+      rerender(<MemoizedTransactionItem txid="tx-1" highlight={false} />)
+
+      expect(mockUseBounceInAnimation).toHaveBeenLastCalledWith(
+        expect.objectContaining({ visible: false }),
+      )
     })
   })
 
--- a/__tests__/components/transaction-item.test.tsx
+++ b/__tests__/components/transaction-item.test.tsx
@@ -51,8 +51,10 @@ jest.mock("@app/i18n/i18n-react", () => ({
 }))
 
 // --- @react-navigation/native ---
+// jest.fn so the focus subscription itself is assertable: only highlighted rows
+// should subscribe, otherwise opening a transaction re-renders every mounted row.
 jest.mock("@react-navigation/native", () => ({
-  useIsFocused: () => true,
+  useIsFocused: jest.fn(() => true),
 }))
 
 // --- @app/components/animations ---
@@ -145,11 +147,13 @@ jest.mock("@app/types/amounts", () => ({
 
 // ─── imports needed for mocking ───────────────────────────────────────────────
 import { useFragment } from "@apollo/client"
+import { useIsFocused } from "@react-navigation/native"
 import { useHideAmount } from "@app/graphql/hide-amount-context"
 import { useAppConfig } from "@app/hooks"
 import { useDisplayCurrency } from "@app/hooks/use-display-currency"
 
 const mockUseFragment = useFragment as jest.Mock
+const mockUseIsFocused = useIsFocused as jest.Mock
 const mockUseHideAmount = useHideAmount as jest.Mock
 const mockUseAppConfig = useAppConfig as jest.Mock
 const mockUseDisplayCurrency = useDisplayCurrency as jest.Mock
@@ -254,6 +258,17 @@ describe("MemoizedTransactionItem", () => {
       expect(mockOnPress).toHaveBeenCalledTimes(1)
     })
 
+    it("passes its own txid to onPress, so the list can share one callback", () => {
+      const mockOnPress = jest.fn()
+
+      const { getByTestId } = render(
+        <MemoizedTransactionItem txid="tx-1" onPress={mockOnPress} />,
+      )
+      fireEvent.press(getByTestId("transaction-item"))
+
+      expect(mockOnPress).toHaveBeenCalledWith("tx-1")
+    })
+
     it("pressing the hidden placeholder does not toggle hide state", () => {
       const mockSwitch = jest.fn()
       mockUseHideAmount.mockReturnValue({
@@ -267,6 +282,20 @@ describe("MemoizedTransactionItem", () => {
       fireEvent.press(getByTestId("hidden-balance-placeholder"))
 
       expect(mockSwitch).not.toHaveBeenCalled()
+    })
+  })
+
+  describe("re-render cost", () => {
+    it("does not subscribe to navigation focus when not highlighted", () => {
+      render(<MemoizedTransactionItem txid="tx-1" />)
+
+      expect(mockUseIsFocused).not.toHaveBeenCalled()
+    })
+
+    it("subscribes to navigation focus when highlighted", () => {
+      render(<MemoizedTransactionItem txid="tx-1" highlight />)
+
+      expect(mockUseIsFocused).toHaveBeenCalled()
     })
   })
 
--- a/__tests__/components/transaction-item.test.tsx
+++ b/__tests__/components/transaction-item.test.tsx
@@ -194,6 +194,10 @@ beforeEach(() => {
 
   mockUseFragment.mockReturnValue({ data: makeTx() })
 
+  // clearAllMocks keeps return values, so pin the default here rather than let
+  // a test that needs an unfocused screen leak into the ones after it.
+  mockUseIsFocused.mockReturnValue(true)
+
   mockUseHideAmount.mockReturnValue({
     hideAmount: false,
     switchMemoryHideAmount: jest.fn(),
@@ -281,6 +285,15 @@ describe("MemoizedTransactionItem", () => {
       expect(mockOnPress).toHaveBeenCalledWith("tx-1")
     })
 
+    it("stays inert when no onPress is given", () => {
+      // The contact-transactions list renders rows without a handler, and they
+      // must not look tappable: handlePress is always defined internally, so
+      // only the guard at the call site keeps those rows from being pressable.
+      const { getByTestId } = render(<MemoizedTransactionItem txid="tx-1" />)
+
+      expect(getByTestId("transaction-item").props.onPress).toBeUndefined()
+    })
+
     it("pressing the hidden placeholder does not toggle hide state", () => {
       const mockSwitch = jest.fn()
       mockUseHideAmount.mockReturnValue({
@@ -322,7 +335,22 @@ describe("MemoizedTransactionItem", () => {
       expect(mockIconMounts).toBe(1)
     })
 
-    it("stops the bounce when the highlight clears, without dropping the wrapper", () => {
+    it("does not remount the row when the highlight turns on", () => {
+      // The screen's highlight baseline and last-seen ids both arrive after the
+      // rows have mounted, so the row the user came to see mounts unhighlighted
+      // and flips false -> true. That is the common direction, and it must not
+      // rebuild the subtree either.
+      const { rerender } = render(
+        <MemoizedTransactionItem txid="tx-1" highlight={false} />,
+      )
+      expect(mockIconMounts).toBe(1)
+
+      rerender(<MemoizedTransactionItem txid="tx-1" highlight />)
+
+      expect(mockIconMounts).toBe(1)
+    })
+
+    it("stops the bounce when the highlight clears, without dropping the subscriber", () => {
       const { rerender } = render(<MemoizedTransactionItem txid="tx-1" highlight />)
       expect(mockUseBounceInAnimation).toHaveBeenLastCalledWith(
         expect.objectContaining({ visible: true }),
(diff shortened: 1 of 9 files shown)


kngako (dismissed):
Kimi K3 review matches @grimen review. Test coverage gaps are of interest. 

---

## Findings

### F1 (Low) — A highlight flip now unmounts and remounts the whole row

`app/components/transaction-item/transaction-item.tsx:228` —
`return highlight ? <BouncingRow>{row}</BouncingRow> : row`

When a transaction's highlight state flips, the element type at that position
changes (`BouncingRow` vs `ListItem`), so React unmounts and remounts the entire
row subtree instead of just stopping the animation — previously every row was
always wrapped in the same `Animated.View`, so the row stayed mounted through the
same transition. Concretely: a user taps a highlighted transaction, the detail
screen opens, `setSeenTxIds` flips `highlight` to `false`, and on return the row
they just tapped is torn down and rebuilt — re-running `useFragment`, resetting
any `ListItem` internal state, and risking a one-frame flicker at exactly the
moment the user is looking at it.

Proposal: keep the wrapper type stable across highlight flips so the `ListItem`
never remounts. Hoist the `useSharedValue`/`useAnimatedStyle` into an
always-mounted shell and render the focus-subscribing piece
(`useIsFocused` + `useBounceInAnimation`) as a nullable child only when
`highlight` is set, passing the shared value down. That preserves the perf win —
only highlighted rows subscribe to focus — while the row's element identity stops
depending on the highlight flag. If the team judges the flicker too rare to
matter, at minimum note the trade-off in the commit/PR so it's a deliberate
choice.

### F2 (Low) — The formatter-cache test is coupled to everything else that constructs an `Intl.NumberFormat`

`__tests__/hooks/use-display-currency.spec.tsx:333`

`jest.spyOn(Intl, "NumberFormat")` counts *every* construction during the
`renderHook` lifetime, not just the 7-fraction-digit one under test, and the
module-level cache means the count also depends on which tests ran earlier in the
file — the inline comment ("7 fraction digits is used by no other test in this
file") is the only thing keeping it deterministic. This breaks the day
`useDisplayCurrency` formats anything else during render (say a default
`formatMoneyAmount` call with 2 fraction digits on a cold cache): the spy sees 2
constructions and the test fails even though per-digit caching works perfectly.

Proposal: make the assertion observe only the behavior under test. Filter the spy
calls to the 7-digit signature
(`numberFormatSpy.mock.calls.filter(([, opts]) => opts?.maximumFractionDigits === 7)`)
and assert that count is `1` — with a cold cache that's exact, and unrelated
formatters can't pollute it. If the cold-cache assumption itself should be
removed, wrap the hook import in `jest.isolateModules`/`jest.resetModules` so the
test owns its module state instead of relying on file-wide ordering.

### F3 (Informational) — `windowSize: 7` trades scroll-blanking risk for memory, and nothing verifies the trade

`app/screens/transaction-history/transaction-history-screen.tsx:71` and
`app/screens/people-screen/contacts/contact-transactions.tsx:33`

Dropping the SectionList window from RN's default 21 to 7 shrinks the mounted
row set (part of the win), but it also narrows the buffer that keeps fast
scrolling from outrunning the renderer. The scenario: a user with years of
history flings the list hard on a low-end Android; rows beyond the ~3-viewport
headroom unmount and the user sees blank gaps until the next batch renders. No
test in the branch (or possible in jsdom) would catch that — it's a device-behavior
change riding inside a commit message that promises "no visual or behavioural
change".

Proposal: this doesn't need code, it needs a verification note. Add a line to the
PR description asking QA to fling-scroll a long history on a low-end device, or
record a before/after scroll capture if the team has the tooling. If blanking
shows up, the escape hatch is bumping to `windowSize: 9` (still less than half
the old default) rather than reverting the whole change.

---

## Test Coverage Gaps

- **The bounce wiring itself is untested after the move.** The new tests assert
  *that* `useIsFocused` is subscribed only for highlighted rows, but nothing
  asserts a highlighted row still drives `useBounceInAnimation` with
  `visible: true` and the 300ms/120ms timing constants. A refactor that renders
  `BouncingRow` but drops the animation call would be fully green. Mock
  `@app/components/animations` and assert the hook receives focus transitions for
  a highlighted row.
- **The non-pressable branch has no test.** The `onPress ? handlePress : undefined`
  guard (`transaction-item.tsx:180`) exists so rows without a handler (e.g. the
  contact-transactions list) don't become pressable — a regression here silently
  makes non-navigating rows look tappable. Add a test that renders without
  `onPress` and asserts the rendered `ListItem` received no press handler.
- **`contact-transactions.tsx` has zero spec coverage.** The screen got the same
  stable-callback and `windowSize` treatment as the history screen, but only the
  history screen gained a callback-identity test. If the renderItem stability
  matters there (it does — same memoization argument), it deserves the same pin.
- **No key-collision test for the formatter cache.** Nothing formats two
  currencies with different fraction-digit counts interleaved and asserts both
  stay correct — the test that would catch a future change to the cache key
  (e.g. someone adds locale or currency to the options but not the key).

---

## Code Smells

- **Duplicated `WINDOW_SIZE` constant and verbatim comment** in
  `transaction-history-screen.tsx:71` and `contact-transactions.tsx:33`. Two
  copies of both the value and its rationale will drift; hoist a shared
  `TRANSACTION_LIST_WINDOW_SIZE` next to `MemoizedTransactionItem` or in a small
  list-config module.
- **Side effect during render in the bench harness**: `ListScreen` assigns
  `benchGlobal.__navigate` in the render body
  (`__tests__/perf/transaction-list.bench.tsx:86`). It's idempotent so it's
  harmless today, but it's exactly the pattern StrictMode double-rendering
  punishes — move it into a `useEffect`.
- **Magic numbers in `settle()`** (`transaction-list.bench.tsx:167`): 100
  attempts × 15ms is a tuned heuristic with no named constants, and
  `jest.setTimeout(300000)` sits mid-`describe`. Naming them
  (`SETTLE_POLL_MS`, `SETTLE_MAX_ATTEMPTS`) would let the next person adjust
  without reverse-engineering.
- **Ceremonial memoization**: `keyExtractor` wrapped in `useCallback` in both
  screens buys stable identity the hard way — a module-level
  `const keyExtractor = (item) => item.id` is stable for free and reads simpler.
  (`renderSectionHeader` legitimately needs the hook since it closes over
  `styles`.)
- **`BouncingRow` under-describes its job**: the name says "animation", but the
  component exists to *own the focus subscription* — which is why the file needs
  a comment explaining it. A name like `FocusBounceRow` would carry that intent
  in the code.

kngako (approved):
👌🏿

PR #4155, opened 2026-08-13
chore(spark): upgrade Breez SDK Spark to 0.22.0

### LINKED ISSUE #4148: chore(spark): upgrade Breez SDK Spark to 0.22.0
## Summary

Upgrade Blink Mobile to Breez SDK Spark `0.22.0`.

## Context

- Requested by Jonas from Slack.
- Source: https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1786624645213009
- Release: https://github.com/breez/spark-sdk/releases/tag/0.22.0
- Verified latest release at creation time: `0.22.0`, published `2026-08-13T12:07:47Z`.

## Acceptance Criteria

- Blink Mobile uses Breez SDK Spark `0.22.0`.
- iOS and Android builds pass with the upgraded SDK.
- Any migration notes or API changes from the release are handled.
- Spark-related smoke/regression testing is completed before release.

grimen:
Fixes #4148.

Upgrades `@breeztech/breez-sdk-spark-react-native` 0.15.0 → 0.22.0 ([release notes](https://github.com/breez/spark-sdk/releases/tag/0.22.0)). Seven minors of a pre-1.0 SDK; the notable upstream fixes we pick up: doubled balance from a leaf counted in two maps ([breez/spark-sdk#996](https://github.com/breez/spark-sdk/pull/996)), coalesced concurrent transfer claims ([#991](https://github.com/breez/spark-sdk/pull/991)), static-deposit refund retry on `MISSING_ENTITY` ([#998](https://github.com/breez/spark-sdk/pull/998)), operator-auth retry with session-expiry margin ([#1035](https://github.com/breez/spark-sdk/pull/1035)), and mempool.space as the default mainnet chain service ([#1042](https://github.com/breez/spark-sdk/pull/1042)).

## What actually broke (verified by diffing the generated bindings of both versions)

Five compile-level breaks reached our code; everything else on the 0.22 surface we call is unchanged.

| Break | Fix |
|---|---|
| `defaultExternalSigner` → `defaultExternalSigners`, returns `{ breezSigner, sparkSigner }`, and the identity-pubkey read became **async** | [`bridge/wallet.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/5190c860dfd9a1a4e9f14060ed7a28f221d50a6e/app/self-custodial/bridge/wallet.ts): `deriveWalletIdentityPubkey` is now async and frees **both** signers in `finally` |
| …async ripple into the three call sites | `use-backup-methods` awaits inside its existing async callback; `use-migration-support-details` chains inside its existing mounted-guarded effect; `useWalletIdentity` becomes state+effect returning `{ pubkey, loading }` |
| `PrepareSendPaymentRequest.paymentRequest`: `string` → tagged `PaymentRequest` enum | Bridge wraps with `new PaymentRequest.Input({ input })` in `send.ts`/`convert.ts`; callers keep passing plain strings — the enum never leaks past the bridge |
| `ConversionDetails.from/.to` → `conversions: Conversion[]` legs | Auto-convert matcher reads `conversions[0]?.from` with a comment pinning the single-leg AMM assumption |
| `UpdateUserSettingsRequest` gained required `sparkMasterIdentityPublicKey` | Both stable-balance calls pass `undefined` (leave the setting untouched) |

Plus one break `tsc` caught that the field-level diff missed: four new `SdkError_Tags` variants broke our exhaustive error-classification map. `InsufficientCpfpFunds` → insufficient funds; the optimization/exit state conflicts (`OptimizationAlreadyRunning`, `OptimizationCancelled`, `FundingUtxoConflict`) → generic, since the app never starts those flows.

## Decisions (and rejected alternatives)

- **Mechanical upgrade only.** 0.22's new capabilities are deliberately not adopted: `Config.crossChainConfig` stays unset (defaults to disabled — it starts background websockets and a whole new payment path), and the SDK-native unilateral exit (`prepareUnilateralExit`/`unilateralExit`, [breez/spark-sdk#992](https://github.com/breez/spark-sdk/pull/992)) is raised as a design question on blink-wip#807 instead of wired in here, because it overlaps the in-flight app-side approach in #4108.
- **`useWalletIdentity` returns `{ pubkey, loading }` instead of keeping the bare string.** The alternative (empty string while pending) makes the cloud-backup screen's "missing pubkey" failure toast fire during a normal race. The consumer folds `loading` into its existing disabled state and `handleBackup` early-returns silently mid-derivation, so `uploadFailed` now only fires for a *settled* empty pubkey.
- **Derivation failures resolve to `""` rather than throwing** at the three call sites, routing into each screen's existing local-failure toast instead of an unhandled rejection (the old sync code could throw with no catch above it).
- **Test-mock fidelity over minimal edits.** The shared jest mock's `SdkEvent_Tags`/`SdkError_Tags` were aligned to the real 0.22 enums (`Optimization` → `AutoOptimization`, `NewDeposits` added, four new error tags), and a stale `conversionId` fixture was moved to the real `conversions: []` shape, so the mocks can't mask a drifted assumption.
- **A completed conversion with no legs counts as already converted.** 0.22 persists a conversion's `status` but rebuilds `conversions` on retrieval, so a settled conversion can surface with an empty legs array while its child payment is still syncing. Reading that as "not converted" (the first draft of this branch, spec included) lets the app fire a second real conversion of the user's sats. The spend-side gate now matches and reports it — a false match skips one auto-convert, a false miss spends twice. `findRecentConversionId` stays strict, because pairing a receive needs a real source amount.
- **The cloud CTA is gated on the keychain read as well as the derivation.** `identityLoading` alone left the earlier window open: `useWalletMnemonic` reports an empty string while reading, so the hook reported not-loading with an empty pubkey and a tap flashed `uploadFailed`. `useWalletMnemonicState` exposes the read state; `useWalletMnemonic` keeps its signature so the other call sites are untouched. A *settled* empty phrase still raises the toast — that is a real local failure.
- **The patch-package patch carries only the Kotlin fix.** It was first generated from a tree that still held a local Gradle build, so it shipped 34 `android/build/` entries — two of them binary, captured without `--binary`, which planted a 0-byte `R.jar` and a 0-byte Kotlin `build-history.bin` into every install. Regenerated from a freshly extracted 0.22.0 with `android/build` removed.
- **Unrelated pod checksum churn is reverted rather than committed as generated.** `hermes-engine`, `ReactNativeDependencies` and `RNPermissions` changed checksum with no version change, and the branch put `RNPermissions` back to the value #3781 wrote which #4011 had already replaced on main. A `pod install` here reproduces the flap, so it is machine-local noise: the lockfile now carries only the two `BreezSdkSparkReactNative` lines this upgrade actually changed.

## Coverage added

Both signers freed on success *and* on rejection, plus a throwing destructor still freeing the other signer and preserving the result; `useWalletIdentity` loading transitions, rejection settling, re-derivation on mnemonic change, and no state write after unmount; `useWalletMnemonicState` loading transitions across stored / not-stored / no-account / rejected reads; cloud-backup silent-while-reading and silent-while-deriving vs toast-on-settled-failure; the legless conversion matching the auto-convert gate while still being skipped for pairing, and still converting when it predates the record or is already claimed; `NewDeposits` actually triggering a refresh; the derivation-rejection paths at both async-ripple call sites; wrapped `PaymentRequest.Input` shapes; the new required settings key; all four new error-tag classifications.

Every one of the new tests was confirmed to fail against the unfixed code, so none of them are vacuous.

## Review findings addressed

Six findings were raised against `ee8f1eb7`. Each fix is its own commit, and each was verified against the *unfixed* code first — the listed test fails without its fix, so none of them are vacuous.

| # | Finding | Fix | How it was proven |
|---|---|---|---|
| 1 | Patch shipped 34 `android/build/` entries; patch-package planted 0-byte `R.jar` + `build-history.bin` on every install | [`4af3e73`](https://github.com/blinkbitcoin/blink-mobile/commit/4af3e7398) | Deleted the package from `node_modules`, reinstalled, re-applied — evidence below |
| 2 | Completed conversion with empty legs read as "not converted" → second real conversion of the user's sats | [`7f230d8`](https://github.com/blinkbitcoin/blink-mobile/commit/7f230d814) | Spec inverted; gate still declines to match when the conversion predates the record or is already claimed |
| 3 | Backup CTA live during the keychain read → spurious `uploadFailed` toast | [`6eccb21`](https://github.com/blinkbitcoin/blink-mobile/commit/6eccb2183) | New spec asserts `loading` true and no toast mid-read, and that a *settled* empty phrase still raises it |
| 4 | Throwing `uniffiDestroy` stranded the other signer and replaced the result | [`253ceec`](https://github.com/blinkbitcoin/blink-mobile/commit/253ceecf3) | New specs assert the second destroy runs and that neither the pubkey nor the original rejection is masked |
| 5 | Two inline `SdkEvent_Tags` mocks missing `NewDeposits`, so `REFRESH_EVENTS` held `undefined` | [`af635df`](https://github.com/blinkbitcoin/blink-mobile/commit/af635dfe1) | Added a refresh assertion, then removed the tag again to confirm it fails |
| 6 | Three unrelated pod checksums changed | [`36f90b6`](https://github.com/blinkbitcoin/blink-mobile/commit/36f90b687) | Lockfile diff vs merge-base is now exactly the two `BreezSdkSparkReactNative` lines |

A coverage audit of all ten app files this upgrade touched found two further gaps — `deriveWalletIdentityPubkey` became async and both call sites gained a `.catch(() => "")` that nothing exercised. Closed in [`e8617fb`](https://github.com/blinkbitcoin/blink-mobile/commit/e8617fb35). `parse.ts` and `send.ts` already fall through safely on 0.22's new `CrossChainAddress` tag, so no change was needed there.

## Verification

- `yarn tsc --noEmit` clean; lint clean
- Targeted jest: 154 suites / 2253 tests green across `__tests__/self-custodial`, `__tests__/screens/self-custodial` and `__tests__/screens/account-migration`
- **Patch applies to a pristine tree** — the decisive check for finding 1. `node_modules/@breeztech/breez-sdk-spark-react-native` was deleted, `yarn install --check-files` re-extracted 0.22.0, and patch-package re-applied:

```
$ grep -c '^diff --git' patches/@breeztech+breez-sdk-spark-react-native+0.22.0.patch
1                                     # was 36
$ wc -l  patches/@breeztech+breez-sdk-spark-react-native+0.22.0.patch
31                                    # was 700

patch-package 8.0.0
Applying patches...
@breeztech/breez-sdk-spark-react-native@0.22.0 ✔

$ find node_modules/@breeztech/breez-sdk-spark-react-native -size 0 -type f | wc -l
0                                     # the 0-byte R.jar / build-history.bin are gone
$ ls node_modules/@breeztech/breez-sdk-spark-react-native/android/
CMakeLists.txt  build.gradle  cpp-adapter.cpp  generated  gradle.properties
proguard-rules.pro  src               # no build/ directory at all
$ grep -c 'reactContext.currentActivity' .../BreezSdkSparkPasskeyModule.kt
3                                     # the intended fix still lands on all three sites
```
- **iOS**: full native build against the 0.22 xcframework, installs and boots on a simulator, non-custodial account creation flow reaches the home screen
- **Android**: the upstream RN 0.85 incompatibility in the SDK's new passkey module needed a patch — `currentActivity` property access was removed from `ReactContextBaseJavaModule` Kotlin in RN 0.80, so the patch routes through the public `reactContext.currentActivity` getter. Worth reporting upstream to breez/spark-sdk. `assembleDebug` was green for all ABIs on the original branch, and the module was **rebuilt after the patch was regenerated** to confirm the slimmed patch still compiles — the Kotlin task recompiled rather than replaying a cached result, which is the point:

```
> Task :breeztech_breez-sdk-spark-react-native:compileDebugKotlin
> Task :breeztech_breez-sdk-spark-react-native:compileDebugJavaWithJavac
> Task :breeztech_breez-sdk-spark-react-native:bundleDebugAar
> Task :breeztech_breez-sdk-spark-react-native:assembleDebug

BUILD SUCCESSFUL in 3m 49s
47 actionable tasks: 43 executed, 4 up-to-date
```
- **Keyed local build — the SDK really connects on 0.22.** An earlier revision of this description claimed wallet-level smoke was impossible locally. That was half right: a local build needs `BREEZ_API_KEY` *and* `SPARK_TOKEN_IDENTIFIER` (`createSdkConfig` calls `requireSparkTokenIdentifier()` while building `stableBalanceConfig`, so it throws before `connect` even with a valid key). The second value is the mainnet USDB token id published across `breez/spark-sdk` `docs/breez-sdk/snippets/*`. One more trap: `react-native-config` defaults to `.env`, so a populated `.env.local` is ignored unless the build runs with `ENVFILE=.env.local` — otherwise the key silently compiles in empty.

  With both supplied that way (gitignored, never committed), a debug build on an isolated simulator reached:

  - non-custodial account creation completing through to the home screen, balances rendering `0 SAT` / `$0.00` on a fresh wallet — so `connect` and the initial sync ran against 0.22
  - **receive**: a real mainnet BOLT11 invoice generated (`lnbc1p487s…hjcpdj9qkl`), with the on-chain card rendered alongside it
  - **cloud backup**: tapping the CTA produced *"iCloud Drive is not available"* — the `startSession` branch — and **not** `uploadFailed`. `handleBackup` only reaches `startSession` after clearing both the pending guard and the empty-pubkey check, so this is end-to-end proof that the async `defaultExternalSigners` → `getIdentityPublicKey` derivation the upgrade introduced works against the real SDK, that the keychain read settled first, and that the local-failure toast correctly stayed silent. The simulator has no iCloud account, so the upload itself cannot complete.

### Smoke checklist (throwaway account — the SDK migrates its store on first connect, no downgrade path)

Items marked done were covered by the keyed local build above. The rest move real funds, so they still need a funded throwaway account.

- [x] **new** wallet connects and syncs, balance correct (`0 SAT` on a fresh account)
- [ ] a **0.15-written** wallet connects, syncs, balance correct (not merely non-zero — #996 fixed a doubling bug). *Not covered: the local run created a fresh 0.22 wallet, so the store migration path is untested.*
- [ ] history loads and pages *(empty on a fresh wallet — nothing to page)*
- [x] receive: LN invoice generated; on-chain card rendered
- [ ] receive: invoice actually gets paid
- [ ] send: BOLT11 + Spark address (`PaymentRequest.Input` on both rails)
- [ ] convert BTC↔USDB both directions (new `conversions[]` shape + auto-convert matcher)
- [ ] stable balance toggle on/off (`updateUserSettings`)
- [x] cloud backup: async pubkey derivation succeeds, no `uploadFailed` flash (the upload itself needs a real iCloud account)
- [ ] password-manager backup
- [ ] migration support details shows the pubkey *(needs a migrating account, not a fresh one)*


kngako (dismissed):
We need to track the react-native version mismatch that resulted in the patch as stated in finding 1 (making it a soft-block as it can only cause us pain on the next bump of the breez-sdk). 

---

## Findings

### Finding 1 — Breez passkey patch has no upstream-issue link or exit criterion (BLOCKING)

`patches/@breeztech+breez-sdk-spark-react-native+0.22.0.patch` (all 31 lines) patches `currentActivity` → `reactContext.currentActivity` in three methods of `BreezSdkSparkPasskeyModule.kt` with no header or inline comment. The bug is unfixed upstream (`main` of `breez/breez-sdk-spark-react-native` still uses the deprecated getter at all three sites) and no tracking issue exists.

When SDK 0.23+ ships, the next upgrader will see this patch either fail to apply or silently keep applying, with no way to know whether it is safe to drop — exactly the "silently dropped patch re-shipping a production crash under a green pipeline" scenario the repo's `--error-on-fail` + upstream-link convention was created to prevent (docs/pr-review.md §2.8, #3825). The convention already exists in-repo: `patches/react-native-credentials-manager+0.8.1.patch` carries an `Upstream: <url>` comment.

**Proposal:** File an issue at `breez/breez-sdk-spark-react-native` (RN 0.80+ removed the deprecated `getCurrentActivity()` Kotlin property access used by the passkey module), then add a patch header comment in the style of the credentials-manager patch: `Upstream: <link> (remove this patch once a release includes the fix)`. One comment block; no code change.

### Finding 2 — Residual one-frame window where the cloud-backup CTA is tappable with an empty pubkey (non-blocking)

`app/screens/self-custodial/onboarding/hooks/use-wallet-mnemonic.ts:73-94` — `useWalletIdentity` initializes `loading` with `useState(false)` and only flips it to `true` inside the passive effect. When the keychain read settles, `setMnemonic`/`setLoading(false)` batch into one commit; on that render `mnemonicLoading=false`, `identityLoading=false`, `pubkey=""`, so `identityPending` in `use-cloud-backup.ts:67` is false and the button is enabled for exactly one commit until the identity effect flushes.

A tap in that frame hits the `if (!identityPubkey)` backstop (`use-cloud-backup.ts:84`) and flashes the "Upload failed" toast — the very toast commit `6eccb2183` set out to remove, now compressed from the full read window to a single frame. It fails safe (no cloud call), so this is UX residue, not a hole; it is also structurally invisible to the current tests because `renderHook`/`act()` flush effects before assertions.

**Proposal:** Derive `loading` synchronously instead of effect-setting it — e.g. store `{ for: mnemonic, pubkey }` and compute `loading = mnemonic !== derived.for`, which is true in the same render the mnemonic changes. This also fixes the first-commit lie from `useState(false)` for any future direct consumer of `useWalletIdentity`. No new state, no extra effect.

### Finding 3 — New silent `.catch(() => "")` fallbacks carry no telemetry (non-blocking)

Four new catch-and-empty-string fallbacks: `use-wallet-mnemonic.ts:48` (keychain read), `use-wallet-mnemonic.ts:85` (pubkey derivation), `use-backup-methods.ts:66`, `use-migration-support-details.ts:57`. The project standard is "no silent fallbacks without `logError`/crashlytics" (docs/pr-review.md §2.3).

In the backup flows the failure does surface as a user toast, but in `useMigrationSupportDetails` a signer failure settles to an empty `pubKey` on the screen whose entire purpose is giving support diagnostics — a support agent asks the user for the pubkey, the field is blank, and there is zero telemetry explaining why.

**Proposal:** Add a crashlytics `reportError` in at least the two derivation catches (never logging the mnemonic itself). The keychain-read catch at line 48 is practically unreachable (`KeyStoreWrapper.getMnemonicForAccount` already swallows to `null`, `app/utils/storage/secureStorage.ts:148-154`) and can be left as-is with a comment saying so.

### Finding 4 — Legless-conversion match has no amount or direction guard (non-blocking)

`app/self-custodial/auto-convert/executor.ts:148` — `if (!details.conversions?.length) { … return true }` treats *any* completed legless conversion newer than `recordCreatedAtMs` as a match, regardless of amount or direction. If the SDK persistently (not transiently) returns empty legs for an unrelated manual conversion the user did right after a receive, that record's auto-convert is skipped forever.

The blast radius is bounded (one skipped auto-convert per affected record, and the commit's fail-closed rationale in the comment at lines 144-147 is sound — a false miss double-spends sats, a false match skips once). But the skip is silent except for one `reportError`.

**Proposal:** No structural change required. Optionally include the payment id in the crashlytics breadcrumb (already done) and add a debug-level log when the skip fires so support can correlate a user report to the suppression.

### Finding 5 — Fail-open asymmetry beside the new fail-closed code (non-blocking, pre-existing)

`app/self-custodial/auto-convert/executor.ts:158-160` — `catch { return false }` on the `listPayments` call in `hasAlreadyConverted` reports "not converted" with **no** `reportError` on an SDK/network hiccup, exposing exactly the double-conversion the new legless guard works to prevent. This predates the branch (identical at merge-base), but the diff makes the asymmetry more visible.

Scenario: the dedup check runs during a transient SDK reconnect; the hiccup returns `false`, and the executor fires a second conversion for a deposit that was already converted.

**Proposal:** Add `reportError("hasAlreadyConverted", err)` in the catch. Consider failing closed (skip auto-convert this trigger) since auto-convert is re-triggered on later receives anyway — that matches the fail-closed philosophy the branch just applied two branches up.

### Finding 6 — `InsufficientCpfpFunds` mapped to user-facing "insufficient funds" (non-blocking)

`app/self-custodial/sdk-error.ts:27` maps the new 0.22 `InsufficientCpfpFunds` SDK error to the same user-facing code as an empty wallet. A user with an ample balance who tries to claim a deposit while the SDK lacks CPFP fee-bump funds sees `sc_insufficient_funds`, which reads as "your wallet is empty" — misleading at exactly the moment support tickets get filed.

It matches the `MaxDepositClaimFeeExceeded` precedent, so it is defensible, but a distinct code (or `Generic`) would misdirect less.

**Proposal:** Map to `Generic`, or add a dedicated code if the UI can phrase it ("network fee reserve too low"). One-line change plus one assertion in `__tests__/self-custodial/sdk-error.spec.ts`.

### Finding 7 — Unrelated yarn.lock churn (non-blocking)

`yarn.lock` splits and re-duplicates the `react-is`, `string-width`, `strip-ansi`, and `wrap-ansi` stanzas with identical resolved versions/integrity hashes — yarn re-resolution noise from the version bump, not a dependency change. It bloats this diff and every future lockfile-diff baseline. The breez pin itself is correct (`"0.22.0"` exact, `package.json:66`).

**Proposal:** Run `yarn dedupe` or regenerate the lockfile from clean main + the bump to confirm the noise is minimal; otherwise accept and move on.

---

## Test Coverage Gaps

The branch's own behavioral changes are well pinned — the legless-conversion test asserts `already-converted` *and* `getConversionQuote` not called, the `destroySigner` tests assert both destroys fire through a throw, and reverting `useCloudBackup` to `useWalletMnemonic` genuinely turns the new mid-read test red. Remaining gaps:

1. **Multi-leg conversion arrays untested** — `executor.ts:105` deliberately reads `conversions[0]?.from` with a comment that a multi-hop route "would need to pick its entry leg here deliberately," but no test has `conversions` with >1 leg, so the first-leg choice is unpinned.
2. **Account-switch while mounted untested** — `useWalletMnemonicState` re-entering `loading:true` with the *previous* account's phrase still in state (stale-phrase-in-flight) has no test.
3. **No screen-level cloud-backup test** — the wiring of `useCloudBackup.loading` to the disabled/spinner state of `GaloyPrimaryButton` (`cloud-backup-screen.tsx:112-118`) is covered only at hook level.
4. **Finding 2's one-frame window is untestable as written** — a regression test only becomes possible once `loading` is derived rather than effect-set; landing the proposed fix enables pinning it.

---

## Code Smells

1. **`SdkEvent_Tags` defined three times** — the shared `__mocks__` file plus inline copies in `wallet.spec.tsx:56` and `use-sdk-lifecycle.spec.ts:17` (per `af635dfe1`). Pre-existing pattern, but every SDK upgrade now requires keeping three copies in sync; importing the shared mock removes the drift risk.
2. **Two near-identical async-settle effect blocks** in `use-wallet-mnemonic.ts` (`useWalletMnemonicState` and `useWalletIdentity`: mounted-guard + catch + settle). Tolerable at two copies; a third should extract a shared `useAsyncState` primitive.
3. **`useWalletIdentity(mnemonic: string)` invites phrase residency** — taking the raw phrase as an argument means every consumer must hold the phrase in memory to get a pubkey, the opposite of the file's own documented laziness (`useLoadWalletMnemonic`, lines 10-14). `useMigrationSupportDetails` shows the better pattern (transient load, derive, discard). Not worth churning now.
4. **Dead-ish public surface (pre-existing, unrelated):** `app/hooks/use-payments.ts:39,59` exposes a `receiveOnchain` adapter no production code consumes — worth a follow-up cleanup ticket, not this PR.
5. **`new PaymentRequest.Input(...)` wrap duplicated** in `convert.ts:112` and `send.ts:70` — two call sites with a clarifying comment; acceptable per the project's "three similar lines" rule.

esaugomez31 (approved):
I tested everything related to self-custodial: all the transaction types and all the conversions. Everything works well, and it feels a bit faster than the current version: the Stable Balance shows up a little sooner than before. I went through the points from the review below, and none of them blocked the testing.

Here is what I resolved:

---

## Resolved

### Finding 2 — Residual one-frame window where the cloud-backup CTA is tappable with an empty pubkey (fixed in `fe2c5a669`)

`useWalletIdentity` no longer stores `loading` in state. The pubkey is now kept next to the phrase it was derived from (`{ forMnemonic, pubkey }`), and both `pubkey` and `loading` are computed from the current phrase, so the hook reports loading in the same render the phrase changes instead of one render later.

This also closes a second case with the same root cause: when the phrase changes from wallet A to wallet B, the hook no longer hands back wallet A's pubkey next to wallet B's phrase.

New test: "never pairs a phrase with the pubkey of the previous one", which only became possible once `loading` was derived (your test gap 4).

### Finding 3 — New silent `.catch(() => "")` fallbacks carry no telemetry (fixed in `fe2c5a669`, `0bc9eeb74`, `f1b9a01f0`)

The three derivation catches now call `reportError("deriveWalletIdentityPubkey", err)`, in `use-wallet-mnemonic.ts`, `use-backup-methods.ts` and `use-migration-support-details.ts`. The phrase itself is never passed to the report.

The keychain read at `use-wallet-mnemonic.ts:48` was left as it is with a comment saying why, exactly as you proposed: `KeyStoreWrapper.getMnemonicForAccount` already swallows to `null`, so that promise has no reachable rejection.

Each of the three has a test asserting the report.

### Finding 6 — `InsufficientCpfpFunds` mapped to user-facing "insufficient funds" (fixed in `b650f2cae`)

Moved to `Generic`, next to the other unilateral-exit tags, with a comment explaining that a CPFP fee reserve is not the same problem as an empty wallet. The assertion in `sdk-error.spec.ts` was flipped to `Generic`.

### Finding 7 — Unrelated yarn.lock churn (fixed in `12747f7e0`)

The four regrouped stanzas (`react-is`, `string-width`, `strip-ansi`, `wrap-ansi`) are back to the grouping `main` has. `yarn.lock` is now 4 changed lines against `main`, all of them inside the breez block.

I did not run `yarn dedupe` for this: the stanzas are not duplicate resolutions, only a different way of grouping the alias descriptors, so deduping would have touched more than these four. I checked the result with yarn's own parser instead: it parses, and every descriptor involved still resolves to the same version as before.

### Test coverage gap 1 — Multi-leg conversion arrays untested (fixed in `47375329a`)

`findRecentConversionId` now has a test with two legs that pins the entry leg as the one being compared, plus a test for a payload where the legs array is missing entirely rather than empty.

### Test coverage gap 3 — No screen-level cloud-backup test (fixed in `529206143`)

`cloud-backup-screen.spec.tsx` now asserts that the continue button renders its loading state while `useCloudBackup` reports loading.

PR #4157, opened 2026-08-13
fix(backup): let users dismiss the secure-wallet modal

### LINKED ISSUE #4156: fix: allow dismissing secure wallet notice when backup flow blocks app use
## Summary
A user report says the self-custodial backup notice ("Secure your funds" / "Secure wallet") can become impossible to dismiss. The user says force-closing and reopening the app does not clear the popup, leaving them unable to use the wallet until they complete the secure-wallet flow.

This has security/privacy impact: the reporter says they needed to make a transaction and ended up writing their seed phrase in public because the app was effectively blocking normal wallet use.

## User report
From the attached screenshot:

> Hey found a large annoying bug for self custodial
> It doesn't let you close this pop up at all even if you force close the app
> Had to write me seed phrase in public cuz I had to do a tx
> Mom said is happened to her aswell

Screenshot evidence: [`IMG_3469.png`](https://slack-files.com/T097VQ50E84-F0BPZN5TTV4-8fad5c8233) from the source Slack thread.

## Affected screen
- Home wallet screen with backup prompt overlay
- Prompt title: "Secure your funds"
- Prompt copy: "We highly recommend you backup your wallet to prevent a complete loss of funds in case you lose this device."
- Primary CTA: "Secure wallet"
- Close icon visible in the top-right of the prompt, but report says it cannot be dismissed

## Expected behavior
Users should be able to dismiss or defer the backup notice and continue using the wallet, especially for urgent transactions. If the backup warning is intentionally persistent, it still must not trap users in a state where the wallet is unusable or where they feel forced to expose seed material.

## Actual behavior
The popup reportedly remains stuck and cannot be closed, including after force-closing and reopening the app.

## Impact
- Blocks wallet use / transaction flow
- Reproduced by at least two users according to the report
- Creates dangerous pressure to handle seed phrase material in unsafe environments

## Notes for investigation
- Check whether dismissal state is persisted correctly for self-custodial wallets.
- Check whether the close button handler is disabled, covered by another view, or immediately reopens the modal on app resume.
- Check whether testnet/basictesting state differs from production behavior; the screenshot shows `basictesting` but the user report describes real-world impact.

## Source
Requested by Lukas (`ihaveadifferentname`) in Slack: https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1786634402371659

grimen:
Fixes #4156

## The bug

The self-custodial backup prompt ("Secure your funds" / "Secure wallet") could not be dismissed. A user reported writing their seed phrase down in public because the prompt made the wallet unusable and they needed to transact; a second user hit the same thing.

It is not a rendering glitch — [`shouldShowModal`](https://github.com/blinkbitcoin/blink-mobile/blob/9b889f63b/app/hooks/use-backup-nudge-state.ts#L82-L87) never consulted `isDismissedRecently`, which only the banner read:

```ts
const shouldShowModal =
  !isBackedUp && isSelfCustodial && isWalletReady && loaded &&
  satsBalance >= backupNudgeModalThreshold      // isDismissedRecently never checked
```

So pressing the X called `dismissBanner`, wrote `backupNudgeDismissedAt:<accountId>`, and then recomputed the modal as visible on the same render. `CustomModal` is `dismissable`, so the backdrop tap and the Android back button routed to the same dead handler. Force-closing did nothing either: the stored timestamp was re-read on launch, but no code path fed it to the modal. The only escapes were completing the backup or letting the balance fall below `backupNudgeModalThreshold` (21 000 sats).

A second, quieter bug lived in the same lines: the modal and the banner shared one dismissal key, so closing the modal also silenced the banner for 24 h.

## The fix

Split the dismissal in two. The modal persists and honours its own `backupNudgeModalDismissedAt:<accountId>` key; the banner keeps the existing key with unchanged semantics.

Once the modal is dismissed, the home-screen banner ("Your funds are at risk") takes over — the modal threshold (21 000 sats) sits well above the banner threshold (2 100), so the banner condition is satisfied the moment the modal steps aside, unless the banner is inside its own cooldown (see the caveat under Decisions).

## Decisions

**Why the banner takes over instead of everything going quiet.** The alternative — reuse the single key so one dismissal silences both surfaces — is a smaller diff, but it leaves a self-custodial user with >21 000 unbacked sats with no warning at all for a day. Handing over to the non-blocking card keeps the security signal on screen while unblocking the wallet, which is what the issue asks for.

**Why the existing banner key is left as-is.** Reusing it for the modal would have made every user who had dismissed the banner also start with a pre-dismissed modal. Keeping the keys separate means users currently trapped get the modal once more and can then close it for good — the right behaviour for a security nudge.

**Why the hook moved namespace.** `useBackupNudgeState` is self-custodial logic — every surface it returns is gated on account type, its dismissal keys are scoped to a self-custodial account id, and it reads the self-custodial backup provider. It has no meaning for a custodial account, so it now lives in `app/self-custodial/hooks/` (spec mirrored, exported through the namespace index) rather than in the shared `app/hooks/`. The shared hooks it consumes — `useAccountRegistry`, `useActiveWallet`, `useIsSelfCustodialAccount` — are genuinely account-type-agnostic and stay put. No behaviour change, and neither unmerged recovery branch touches this file, so it does not conflict with them.

**Why the modal cooldown is remote-configurable and the banner's is not.** The modal is the blocking surface, so its nag interval is the knob ops is most likely to want to retune after this ships. `backupNudgeModalCooldownMs` defaults to 24 h, matching the banner constant. `asNumber()` returns 0 for a malformed remote value and a zero cooldown would resurrect this exact bug, so the read falls back to the shipped default when the value is not positive.

**Why `multiGet`, and why its result is read by key.** Two keys now load on mount; one round trip keeps the existing "don't flash the nudge before storage is read" guarantee behind the single `loaded` flag. The reply is looked up in a `Map` rather than destructured positionally, because **`multiGet` does not echo the requested key order on Android**: the legacy module we ship (`AsyncStorage_useNextStorage` is unset) runs `WHERE key IN (…)` with no `ORDER BY` and appends the keys it did not find as `[key, null]` at the end. Reading it positionally swapped the two timestamps for any user who had dismissed one surface but not the other, which reinstated this very bug on Android — see [`AsyncStorageModule.java:130-152`](https://github.com/react-native-async-storage/async-storage/blob/main/packages/default-storage/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java).

**Why the read fails open.** Android surfaces SQLite disk-full and `CursorWindow` errors as a rejected promise. Without a `catch`, `loaded` never flipped and *both* nudges stayed hidden for the whole session with nothing reported. Unreadable storage now reports and shows the nudge: a security prompt should fail loud, not quiet.

**Why the nudge ignores the dollar-balance restriction.** `useTotalBalance` zeroes the USD wallet for users in a dollar-restricted region, and the nudge used that number as its threshold. That is a display rule: the balance is hidden, not gone. A self-custodial user in such a region holding most of their value in stable tokens could therefore sit under both the 21 000-sat modal threshold and the 2 100-sat banner threshold and never be prompted, while the funds sat unbacked on the device. The zeroing is now opt-out (`applyDollarRestriction`, default true), so the home-screen display is unchanged and only the nudge opts out. This is a pre-existing bug rather than a regression in this branch, fixed here because it defeats the very prompt this PR is repairing.

**Why the "Secure wallet" CTA no longer dismisses.** It used to call `onClose` — i.e. `dismissModal` — on its way to the backup screen, so starting the flow and abandoning it bought a full cooldown with nothing backed up. It now only navigates. `isVisible` is already gated on `isFocused`, so the modal hides while the backup flow is open and returns if the user backs out, and finishing the backup flips `isBackedUp` and retires the nudge anyway. The X, the backdrop and the Android back button still dismiss, so nobody is trapped.

**Why the load effect ignores stale replies.** Switching self-custodial accounts re-runs it, and nothing orders the previous account's read before the current one's. Both paths are guarded, not just the successful one: fail-open clears the dismissal state, so a late *rejection* from the old account would have wiped the new account's freshly loaded timestamps and re-shown a modal it had already dismissed.

**One caveat on the takeover.** The banner takes over unless it is itself inside its own 24 h cooldown — dismiss the banner in the morning and the modal in the afternoon and both home-screen surfaces stay quiet for the rest of the day. That is deliberate: two explicit dismissals are two explicit dismissals, and it is not a silent state, because `shouldShowSettingsBanner` carries no threshold and no dismissal check, so the Settings warning stays up for as long as the account is unbacked.

## Tests

`__tests__/self-custodial/hooks/use-backup-nudge-state.spec.ts` — the `multiGet` mock now mimics the Android contract (found rows first, missing keys appended as `[key, null]`) instead of echoing the requested order. That change alone reddens two of the cases below against the positional read, which is the regression proof for the Android bug. New and updated cases, each verified to fail against the unfixed hook:

- dismissing the modal hides it and writes `backupNudgeModalDismissedAt:<accountId>`
- the banner takes over once the modal is dismissed
- a stored modal timestamp within the cooldown keeps it hidden on a fresh mount (the force-close half of the report)
- a non-default `backupNudgeModalCooldownMs` is respected
- the two dismissals stay independent in both directions
- **only the modal key present, returned in Android order → the modal stays dismissed and the banner shows**
- **a rejected read still resolves `loaded`, leaves the nudge visible and reports the error**
- **a failed dismissal write is reported**
- **a custodial account clears dismissal state, finishes loading and never persists a key**
- **the balance is measured with the dollar-display restriction opted out**
- **a stale read from a previously active account cannot overwrite the current one — on both the resolve and the reject path**
- **with both home-screen surfaces dismissed, the settings banner still shows**

`__tests__/components/backup-nudge-modal.spec.tsx` — the CTA navigates *without* calling `onClose`; the close X still dismisses and does not navigate.

`app/self-custodial/hooks/use-backup-nudge-state.ts` and `app/components/backup-nudge-modal/backup-nudge-modal.tsx` are both at 100% statements, branches, functions and lines. Every assertion listed here was run against a deliberately broken version of its own fix and fails without it.

`__tests__/config/backup-nudge-modal-cooldown.spec.tsx` — new. `backupNudgeModalCooldownMs`' fallback was previously untested even though it is what stops a malformed remote value (`asNumber()` returns 0) from making the modal undismissable again. It drives the real provider rather than a copy of the derivation. Worth knowing if you touch it: the provider's initial state *is* `defaultRemoteConfig`, so asserting "equals the default" right after render passes before the async fetch commits — and would still pass with the fallback deleted. The spec waits for a sentinel remote value to arrive first, so the assertion means what it says.

`__tests__/components/balance-header/use-total-balance.spec.ts` — the restriction is still applied by default, and skipped when the caller opts out.

`__tests__/screens/home.spec.tsx` — asserts the modal's `onClose` is wired to `dismissModal`, not `dismissBanner`.

## Ops follow-up

`backupNudgeModalCooldownMs` needs creating in the Firebase Remote Config console (default `86400000`). Until it exists the app uses the in-code default, so this does not block the merge.

## Demo

Same tap on both sides — the close X on the modal. Left: the X does nothing. Right: the modal closes and the "Your funds are at risk" banner takes over.

| before | after |
|---|---|
|  |  |

<sub>Recorded on an iOS simulator with the hook's server-owned inputs stubbed (unbacked self-custodial account, 22 000 sats) so the prompt is reachable without an account — the dismissal logic in the recording is the real thing. Demos live on branch `assets/pr-4157-demo`; deleting it breaks them.</sub>

### DIFF
--- a/__tests__/components/backup-nudge-modal.spec.tsx
+++ b/__tests__/components/backup-nudge-modal.spec.tsx
@@ -1,4 +1,5 @@
 import React from "react"
+import { TouchableOpacity } from "react-native"
 import { render, fireEvent } from "@testing-library/react-native"
 
 import { BackupNudgeModal } from "@app/components/backup-nudge-modal"
@@ -127,7 +128,7 @@ describe("BackupNudgeModal", () => {
     expect(getByText("Secure wallet")).toBeTruthy()
   })
 
-  it("navigates to backup method screen and closes on button press", () => {
+  it("navigates to the backup method screen without dismissing on button press", () => {
     const onClose = jest.fn()
     const { getByTestId } = render(
       <BackupNudgeModal isVisible={true} onClose={onClose} />,
@@ -135,8 +136,23 @@ describe("BackupNudgeModal", () => {
 
     fireEvent.press(getByTestId("secure-button"))
 
-    expect(onClose).toHaveBeenCalledTimes(1)
     expect(mockNavigate).toHaveBeenCalledWith("selfCustodialBackupMethod")
+    // Starting the backup flow is not finishing it: a user who backs out of the
+    // backup screen must be nudged again rather than get the cooldown.
+    expect(onClose).not.toHaveBeenCalled()
+  })
+
+  // The close affordance is CustomModal's own TouchableOpacity around a "close"
+  // GaloyIcon, which this suite mocks away - hence matching by type. This is the
+  // path the original bug report could not escape, so it stays covered here.
+  it("dismisses on close button press", () => {
+    const onClose = jest.fn()
+    const view = render(<BackupNudgeModal isVisible={true} onClose={onClose} />)
+
+    fireEvent.press(view.UNSAFE_getAllByType(TouchableOpacity)[0])
+
+    expect(onClose).toHaveBeenCalledTimes(1)
+    expect(mockNavigate).not.toHaveBeenCalled()
   })
 
   it("uses primary color for warning icon", () => {
--- a/__tests__/components/backup-nudge-modal.spec.tsx
+++ b/__tests__/components/backup-nudge-modal.spec.tsx
@@ -1,5 +1,4 @@
 import React from "react"
-import { TouchableOpacity } from "react-native"
 import { render, fireEvent } from "@testing-library/react-native"
 
 import { BackupNudgeModal } from "@app/components/backup-nudge-modal"
@@ -128,7 +127,7 @@ describe("BackupNudgeModal", () => {
     expect(getByText("Secure wallet")).toBeTruthy()
   })
 
-  it("navigates to the backup method screen without dismissing on button press", () => {
+  it("navigates to backup method screen and closes on button press", () => {
     const onClose = jest.fn()
     const { getByTestId } = render(
       <BackupNudgeModal isVisible={true} onClose={onClose} />,
@@ -136,23 +135,8 @@ describe("BackupNudgeModal", () => {
 
     fireEvent.press(getByTestId("secure-button"))
 
-    expect(mockNavigate).toHaveBeenCalledWith("selfCustodialBackupMethod")
-    // Starting the backup flow is not finishing it: a user who backs out of the
-    // backup screen must be nudged again rather than get the cooldown.
-    expect(onClose).not.toHaveBeenCalled()
-  })
-
-  // The close affordance is CustomModal's own TouchableOpacity around a "close"
-  // GaloyIcon, which this suite mocks away - hence matching by type. This is the
-  // path the original bug report could not escape, so it stays covered here.
-  it("dismisses on close button press", () => {
-    const onClose = jest.fn()
-    const view = render(<BackupNudgeModal isVisible={true} onClose={onClose} />)
-
-    fireEvent.press(view.UNSAFE_getAllByType(TouchableOpacity)[0])
-
     expect(onClose).toHaveBeenCalledTimes(1)
-    expect(mockNavigate).not.toHaveBeenCalled()
+    expect(mockNavigate).toHaveBeenCalledWith("selfCustodialBackupMethod")
   })
 
   it("uses primary color for warning icon", () => {
--- a/__tests__/components/balance-header/use-total-balance.spec.ts
+++ b/__tests__/components/balance-header/use-total-balance.spec.ts
@@ -92,5 +92,22 @@ describe("useTotalBalance", () => {
         expect.objectContaining({ amount: 50_000, currencyCode: "USD" }),
       )
     })
+
+    // Hiding a restricted dollar balance is a display rule. Callers that measure
+    // how much is at risk on the device must see the funds that are really there.
+    it("keeps the USD balance when the caller opts out of the display restriction", () => {
+      mockUseDollarBalanceRestricted.mockReturnValue(true)
+      const convert = buildConvertSpy()
+      mockConvertMoneyAmount.mockReturnValue(convert)
+
+      renderHook(() => useTotalBalance(wallets, { applyDollarRestriction: false }))
+
+      const usdCalls = convert.mock.calls.filter(
+        (args) => (args[0] as unknown as { currencyCode: string }).currencyCode === "USD",
+      )
+      expect(usdCalls[0][0]).toEqual(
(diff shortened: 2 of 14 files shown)


kngako (approved):
image

After getting the backup prompt I was able to close it. 

image

There's still sufficient warnings on the app about backing up the wallet. 

image

PR #4158, opened 2026-08-13
fix(security): use a CSPRNG for the payment idempotency key

grimen:
Closes blinkbitcoin/blink-wip#1142 — item 4 of the red-team hardening epic [blink-wip#1092](https://github.com/blinkbitcoin/blink-wip/issues/1092).

## Problem

`useGetUuid` in `use-send-payment.ts` built the `X-Idempotency-Key` UUID for every payment mutation (intraledger, LN, on-chain, USD variants) from `Math.random()`, which is not a CSPRNG — the key guarding a whole send-screen session was predictable.

## Fix

Generate the key with `Crypto.randomUUID()` from `react-native-quick-crypto` (already a pinned dependency), mirroring the existing self-custodial idempotency key in `app/self-custodial/payment-details/lnurl.ts`. The `useMemo(..., [])` stability is intentional — one idempotency key per screen session — and is preserved (covered by a new test).

## Decisions

- **Not** consolidating the app's three coexisting UUID strategies (`Crypto.randomUUID()`, `uuidv4 + react-native-securerandom` in `use-create-device-account.ts`, and this one) into a shared util. That's a cleanup, not a security fix, and would widen the review surface of a security PR. Can be a follow-up.
- The `uuid` dependency stays in package.json — `use-create-device-account.ts` still uses it.

## Tests

New `__tests__/screens/send-bitcoin-screen/use-send-payment.spec.tsx`: asserts all nine payment mutations receive the CSPRNG-generated `X-Idempotency-Key` in their context headers, and that the key is stable across re-renders.

## Review round 1 (2026-08-14)

A /code-review pass confirmed the CSPRNG switch is sound but caught the
one-key-per-session invariant being neither enforced nor tested:

- **Vacuous stability test** — the global quick-crypto mock returns a fixed
  UUID, so the Set-size assertion could not fail (proved by deleting the
  useMemo: tests stayed green). Fixed in https://github.com/blinkbitcoin/blink-mobile/commit/43066dff76fced5a1ae7dc939d9c5094cf245582: distinct primed values per
  test, literal header assertions, and a pinned mint count. Re-verified by
  the same mutation — it now fails the spec.
- **useMemo → useState** — React may discard a memo cache and re-mint the
  key on a mounted component, silently defeating backend dedupe on a
  same-session retry. Fixed in https://github.com/blinkbitcoin/blink-mobile/commit/87d57b0274cf0148e93f8add33bf2e20490dbdc9 with a useState lazy initializer.
  Rejected alternative: keeping useMemo and only strengthening the test —
  the test runs under a renderer that never drops memo caches, so it cannot
  police production runtime modes; the mechanism itself has to carry the
  guarantee. A useRef would work identically; useState was chosen as the
  more idiomatic single-init pattern in this codebase.
- Dropped a stray assertion on pre-existing unrelated behavior from the
  stability test.

### Note on same-mount retry semantics (unchanged by this PR)

`hasAttemptedSend` resets on `PaymentSendResult.Failure`, so a retry within the
same screen session deliberately reuses the same key — that is the invariant
above working as intended. If the backend caches failure responses by
idempotency key, such a retry returns the cached failure rather than
re-attempting. This behaviour is identical before and after this PR (the old
`useMemo` had the same semantics); flagging it so it is not mistaken for a
regression introduced here.

## Proof

Nothing here changes a pixel, so the evidence is behavioural rather than visual.

### The test can now fail — same mutation, before and after

The claim above is that the original spec could not detect key churn. Both runs
below use the **identical** mutation of the production hook, minting per render
instead of once per mount:

```diff
 const useGetUuid = () => {
-  const [randomUuid] = useState(() => Crypto.randomUUID())
-  return randomUuid
+  return Crypto.randomUUID()
 }
```

**Before** (spec as it was at `7e9e79a9`) — the broken hook passes:

```
PASS __tests__/screens/send-bitcoin-screen/use-send-payment.spec.tsx
  useSendPayment idempotency key
    ✓ uses a CSPRNG-generated UUID as the X-Idempotency-Key on every mutation (11 ms)
    ✓ keeps the same key across re-renders (one key per screen session) (1 ms)

Tests:       2 passed, 2 total
```

**After** (spec at `43066df`) — the same broken hook is caught:

```
FAIL __tests__/screens/send-bitcoin-screen/use-send-payment.spec.tsx
  useSendPayment idempotency key
    ✓ uses a CSPRNG-generated UUID as the X-Idempotency-Key on every mutation (8 ms)
    ✕ keeps the same key across re-renders (one key per screen session) (1 ms)

  ● useSendPayment idempotency key › keeps the same key across re-renders (one key per screen session)

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 1
    Received number of calls: 2

      66 |     rerender({})
      67 |
    > 68 |     expect(randomUUIDMock).toHaveBeenCalledTimes(1)
         |                            ^

Tests:       1 failed, 1 passed, 2 total
```

Mutation reverted afterwards; the branch as pushed is green.

### What the key is a function of

The old key was fully determined by `Math.random`'s internal state — replay that
state and the same key comes back. The new one draws from the platform CSPRNG,
so the same replay yields nothing:

```
BEFORE — Math.random-seeded UUID, same PRNG state twice:
  run 1: 1c954672-4cb2-49e9-b21c-c7ca5555de88
  run 2: 1c954672-4cb2-49e9-b21c-c7ca5555de88
  identical: true

AFTER — CSPRNG UUID, same PRNG state twice:
  run 1: da75b0bf-0d7a-44ca-8d44-5707c900ecaa
  run 2: e8c55fdc-b992-4c46-9b55-db585118afca
  identical: false
```

The seeded PRNG stands in for an attacker who knows `Math.random`'s state; V8's
xorshift128+ state is recoverable from observed outputs, which is what made the
old construction predictable in principle. The demo used node's `randomUUID` for
the "after" line — on device it is quick-crypto's, and both draw from the
platform CSPRNG rather than `Math.random`.

---

## Review round 2 (2026-08-18)

Two reviews agreed on the same thing: the CSPRNG swap was correct but hardened the
wrong property. The key was minted once per **mount** of the confirmation screen, and
every retry that matters happens on a fresh mount — back out, re-enter, and a brand-new
key sailed past the backend's dedupe. Scope has been widened to fix that properly rather
than defer it.

### The reframing

The key identifies a **payment intent** — destination, settlement amount, source wallet
— not a render tree. It now rides on the `PaymentDetail` as a mutable holder
(`IdempotencyKeyRef`), mirroring what the self-custodial rail already does in
[`app/self-custodial/payment-details/lnurl.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/45aea9a47/app/self-custodial/payment-details/lnurl.ts#L142-L162).
The detail is owned by the details screen, which stays mounted beneath the confirmation
screen, so it already has the lifetime the key needs.

**Which rebuilds keep the key:**

| setter | | why |
|---|---|---|
| `setConvertMoneyAmount` | keep | display currency only; fires from an effect, possibly mid-flight |
| `setInvoice` (lnurl) | keep | **load-bearing** — every "Next" fetches a fresh bolt11 for the same merchant and amount; without this one intent carries two keys and the merchant can be paid twice |
| `setSuccessAction` | keep | post-settlement metadata, never sent |
| `setMemo` | keep | the one deliberate trade — see below |
| `setAmount` | re-mint | amount is the payload; reusing the key would 409 a corrected payment, or on a store-and-replay backend hand back the *original* result |
| `setSendingWalletDescriptor` | re-mint | `walletId` is in the payload |

**The `setMemo` trade, called out explicitly:** a memo *is* in the wire payload, so a
deduped retry discards an edited note. It changes neither who is paid nor how much, and
the alternative is paying twice. Money safety wins, but it is a real trade and worth
disagreeing with if you see it differently.

### Findings addressed

- **Key scoped to the mount** (both reviews) — fixed in 27bc039e9. A remount now reuses
  the intent's key; a different intent gets a fresh one. Both directions are tested.
- **Minting during render is uncatchable** — fixed in d99ca9eda. Minted lazily inside the
  send callback, so a quick-crypto/Nitro failure rejects the promise the screen already
  catches instead of whitescreening the tree through `ErrorBoundary`. **No fallback**: if
  the CSPRNG is unavailable nothing is sent.
- **A thrown mutation dead-ends the screen** — fixed in d99ca9eda. `hasAttemptedSend` was
  answering two questions: it gated the slider *and* told the screen to stop validating
  against a balance the backend may already have debited. Simply resetting it would have
  re-armed the balance check on exactly the path where money most likely *did* move,
  disabling the slider again with the balance message hidden behind the payment error
  (`errorMessage = paymentError || invalidAmountErrorMessage`). So it is split:
  `hasAttemptedSend` stays sticky, a new `sendLocked` gates the slider and is released on
  a `Failure` and on a throw. The screen now reads `disabled={... || !sendPayment || ...}`
  — one source of truth.
- **Where the key is stamped** — instead of per-call `context` at the 11 mutate call sites
  in the factories, the nine mutate functions are wrapped at the single point where they
  are handed to the payment detail. A mutation added later is covered automatically, which
  is the guarantee the hook-level options used to give.
- **Test findings** — `jest.requireActual` spread (so `PaymentSendResult.AlreadyPaid` is
  the real value, not a hand-stubbed subset), `refetchQueries` asserted with `toEqual`,
  `jest.clearAllMocks()` dropped, `useGetUuid` deleted outright, `/** */` comment blocks.
- **Coverage gaps** — added: fresh key per intent, failure→retry under the same key,
  thrown→retry under the same key, terminal statuses keep the send withheld, CSPRNG
  failure sends nothing, and the per-setter preserve/re-mint rule across all five
  factories.
- **Header on the wire** (gap 4) — 45aea9a47 runs a real `ApolloClient` over a capturing
  terminal link with the generated module unmocked. This mattered more than it looks:
  moving from hook-level to per-call context means *Apollo* decides whether the context
  reaches the operation, so without this test a merge behaving differently would have
  dropped the header in production with every other spec still green. It also asserts the
  `HomeAuthed` refetch does **not** inherit the key.

### Two things I did not change

- **`verifyPaymentSettled` not running for onchain/intraledger** is structural, not an
  oversight. It resolves settlement by extracting a payment hash from a **bolt11
  invoice** ([`use-verify-payment-settled.ts:130-147`](https://github.com/blinkbitcoin/blink-mobile/blob/45aea9a47/app/screens/send-bitcoin-screen/hooks/use-verify-payment-settled.ts#L130-L147)),
  which those rails do not have. The `paymentRequest` ternary is a correct guard. The
  other half of that finding — the dead retry button — is fixed.
- **`useState` vs `useMemo`** — correct that React 19.2.3 does not discard memo caches, so
  the original commit message overstated its case. Moot now: the key no longer lives in
  the mount at all.

### On retry and `AGENTS.md`

"Payment mutations must NOT have retry logic" targets *automatic* transport resends —
[`retry-policy.ts:7-15`](https://github.com/blinkbitcoin/blink-mobile/blob/45aea9a47/app/graphql/retry-policy.ts#L7-L15)
says recovery is "owned by the callers, which surface a user-driven retry". Re-enabling
the slider is that. It is only safe *because* the key is now stable: before this change a
resend after a remount was a genuine second payment. The two halves ship together.

A 409 re-enables the slider uniformly, with no error classification — the hook cannot
tell a 409 from a socket reset, and pretending otherwise is where this class of bug comes
from. For lightning/lnurl a resend re-runs the verification poll; for onchain/intraledger
it 409s again harmlessly, because a per-intent key makes a duplicate impossible by
construction.

### Verification

`tsc --noEmit` and ESLint clean. 139 tests green across the send suite, plus 103 in
`__tests__/graphql`. **Every new test was verified by mutation** — the production code it
protects was broken and the test confirmed red before being kept. Twelve mutations in
total (key ignored, key constant, throw no longer reopens, `Failure` no longer reopens,
header dropped, header renamed, wrapper neutered, `Math.random` fallback, lock removed,
`refetchQueries` dropped, render-time mint restored, and the lnurl `setInvoice` re-mint
that would let a merchant be paid twice).

### CI status

`Tests` and `Android unit tests` green at 66330e4ab; merged `main` in to pick up its
`generated.ts`.

`Check Code` is red on **repo-wide codegen drift, not on this branch**. `tsc` and ESLint
both pass; the failing step is the `generated.ts` guard. `codegen.yml` reads the schema
from the live `https://api.staging.blink.sv/graphql`, so the check reflects whatever
staging is serving at the moment it runs. Reproduced by regenerating against **main's own**
`generated.ts`: it gains 29 lines of `AccountLevelLimits` / `accountLimitsByLevel`, i.e.
the account-limits work now on staging and unrelated to anything here. This branch's
`generated.ts` is byte-identical to main's, and any PR run right now fails the same way.
Deliberately not regenerating it here — that schema churn does not belong in a security
PR, and it would go stale again on the next staging deploy.

`iOS` and `Android` builds were already failing on the previous head (87d57b027) before
these commits. `E2E` is the unmaintained suite.

### Coverage audit (f83113127)

Audited the diff for untested paths rather than assuming the suite was complete. Two real
gaps, both now closed:

**Unasserted decisions.** 6 of the 24 preserve/re-mint decisions had no assertion, and
`createAmountOnchainPaymentDetails` had no test at all. All 24 are now covered across all
six factories.

**Vacuous re-mint assertions — the more serious one.** Every re-mint assertion started
from a freshly created detail, whose `params` never carried a ref. "The setter dropped the
key" and "the params never held one" are indistinguishable in that setup, so the assertion
passed whether or not the factory dropped anything. Caught by mutation: deleting
`idempotencyKeyRef: undefined` from a re-mint setter left the suite green. The explicit
drop is only load-bearing on a *chained* detail, where a preserve setter has already put
the ref into `params` — exactly the real sequence (edit a note, then change the amount).
Re-mint assertions now start from such a chained detail and the mutation is red. This also
strengthened the `setAmount` assertions, which were passing for the same hollow reason.

Also added: the screen mapping the CSPRNG-failure sentinel to a generic message instead of
showing the user `idempotency-key-unavailable`. That needed `requireActual` spread into the
`use-send-payment` mock — wholesale mocking made `IDEMPOTENCY_KEY_UNAVAILABLE` `undefined`
and silently disabled the comparison, the same trap flagged for the generated module.

Final ratio: 725 lines of test against ~204 lines of production change. 257 tests green
across the send suite; 17 mutations verified red in total.


kngako (changes requested):
## Findings

### Finding 1 — BLOCKING: the ambiguous-failure retry path still rotates the key

`setHasAttemptedSend(true)` in `use-send-payment.ts:234` is only reset when the mutation *returns* `PaymentSendResult.Failure` (`use-send-payment.ts:252-254`). If the mutation **throws** instead — network drop, timeout, 5xx, i.e. precisely the cases where the client cannot know whether the payment landed — the error propagates to `handleSendPayment`'s catch, `hasAttemptedSend` stays `true`, and `sendPayment` becomes `undefined` for the rest of the mount.

**When it comes up:** a user swipes to pay an invoice on flaky hotel Wi-Fi; the request reaches the server and executes, but the response is lost. The user sees an error, the swipe button is now dead (`sendPayment` is `undefined`), so they go back and re-confirm — which remounts the hook and mints a **new** idempotency key. The retry sails straight past the backend's 409 dedupe and the user pays twice. The `useState` fix guarantees key stability only within a mount, but the retries that matter most all happen on a fresh mount.

**Proposal:** catch the throw inside the hook, reset `hasAttemptedSend`, and re-throw so the user can retry on the same mount with the same key; a true duplicate then hits the backend 409 and the existing `isIdempotencyConflict` → `verifyPaymentSettled` flow in `send-bitcoin-confirmation-screen.tsx:319-351` resolves it safely. Keep this strictly user-initiated — per project rules, payment mutations must not get automatic retry logic (the `RetryLink` exclusion already covers that). For defense in depth, consider keying the idempotency key by payment destination (e.g. per `paymentRequest`) rather than per mount, so even a remounted retry of the same invoice dedupes server-side.

### Finding 2 — non-blocking: the spec hand-mocks an auto-generated module

The new spec stubs `@app/graphql/generated` with a hand-written factory listing all 9 payment mutation hooks plus enum values (`use-send-payment.spec.tsx:7-24`), while `generated.ts` is auto-generated and churns with every schema update (see the recent codegen commit `8d5a68086` on this very branch's base).

**When it comes up:** a schema regeneration renames or adds a payment mutation hook, or `useSendPayment` starts consuming one more export (e.g. `Transaction` as a value instead of a type). The spec then fails with an unhelpful `TypeError: (0 , generated_1.useXxxMutation) is not a function` — or worse, someone "fixes" it by trimming the hook's export list to match the stale mock instead of updating the mock.

**Proposal:** build the mock with `jest.requireActual("@app/graphql/generated")` and override only the nine mutation hooks, so every other export tracks codegen automatically. If the manual list is kept deliberately as a change-detector, add a one-line comment saying so and a test name that makes the intent obvious when it breaks. Either way, keep the `toHaveLength(9)` pin — that part is a useful tripwire.

---

## Test Coverage Gaps

1. **No test that a remount mints a fresh key.** The safety property that prevents the *opposite* bug — two different payments sharing a key and the second being silently deduped against the first — is untested. With the global quick-crypto mock returning a fixed UUID by default, a regression that hoists the key to module scope (one key for the whole app lifetime) would stay green. Render two separate hook instances, prime distinct UUIDs, and assert the keys differ.

2. **No test of the failure → retry path the `useState` change exists to protect.** The commit message for `87d57b027` cites same-mount retry after `PaymentSendResult.Failure` as the live path, but the spec never executes `sendPayment` — the mutation functions are inert `jest.fn()`s. A test that makes the mocked mutation return `Failure`, asserts `hasAttemptedSend` resets, re-invokes `sendPayment`, and checks the second call carries the same `X-Idempotency-Key` would lock in the actual invariant.

3. **No test for the thrown-error path** (see Finding 1). Whichever way Finding 1 is resolved, the behavior on a rejected mutation promise — whether `hasAttemptedSend` resets and which key the retry uses — deserves an explicit spec, since it is the double-payment risk boundary.

4. **Header never observed on the wire.** The spec asserts the `options` object passed to the Apollo hooks at render time, but nothing executes a mutation through a link/Apollo mock to prove the header survives into the actual HTTP request. A light integration-style test with a mocked Apollo link (the repo already mocks Apollo elsewhere) would close the loop.

## Code Smells

- **Redundant mock clearing** (`use-send-payment.spec.tsx:41-46`): `jest.clearAllMocks()` immediately followed by `randomUUIDMock.mockReset()` — reset is a superset of clear for this mock, and the comment only justifies the reset. Drop the `clearAllMocks()` or the comment stops making sense to the next reader.

- **`useGetUuid` no longer describes what it does** (`use-send-payment.ts:164`): it is not "getting" a UUID, it is minting a once-per-mount idempotency key via a `useState` initializer. Renaming it to `useIdempotencyKey` would make the call site self-documenting (`const idempotencyKey = useIdempotencyKey()`) and put the invariant in the name rather than only in the comment.

- **Key minted eagerly even when nothing can be sent**: `useGetUuid()` runs at hook mount even when `sendPaymentMutation` is `null`/`undefined` (the hook is called unconditionally at `send-bitcoin-confirmation-screen.tsx:156`). Harmless — one native CSPRNG call — but lazy minting on first actual send would tie the key's lifetime to the payment intent rather than the screen's render lifecycle.

- **Commit message path is wrong**: `7e9e79a93` references `payment-details/lnurl.ts`; the actual file is `app/self-custodial/payment-details/lnurl.ts`. Trivial, but it will mislead anyone grepping for the referenced pattern later.

kngako (approved):

PR #4159, opened 2026-08-13
feat(security): gate the WebView WebLN bridge to allowlisted origins

grimen:
Closes blinkbitcoin/blink-wip#1139 — item 1 (the Medium lead item) of the red-team hardening epic [blink-wip#1092](https://github.com/blinkbitcoin/blink-wip/issues/1092).

> **One deliberate deviation from the issue text.** #1139's remediation summary lists `setSupportMultipleWindows={false}`. This PR ships `true`. That line of the remediation is wrong: `false` makes `onOpenWindow` dead code on Android, so `window.open` navigates the trusted WebView in place — defeating the very requirement it sits next to. The intent of the item is met; the letter is not. Reasoning and sources under "Review follow-ups" below.

## Problem

`webview.tsx` — the app's only WebView, used for KYC/Onfido and the fiat buy/sell pages — had no origin controls at all:

- the `url` route param rendered unvalidated, with the default `originWhitelist` (http+https);
- the WebLN shim was re-injected on **every** navigation, including cross-origin redirects;
- the hole is actually wider than the audit stated: `onMessage` was wired unconditionally, so a page didn't even need the injected shim — anything posting a correctly-shaped message to `window.ReactNativeWebView.postMessage` reached the live `sendPayment` handler, which navigates into the send flow with an attacker-chosen bolt11 (pre-fill/phishing vector; payment still needs user confirmation downstream).

## Fix — three layers

1. **Entry gate**: only origins of the active galoy instance's `kycUrl`/`fiatUrl` may load; a refused URL shows the generic error alert and never mounts the WebView.
2. **Injection gate**: the WebLN shim is injected per load, decided against `e.nativeEvent.url` of the document actually loading — a cross-origin redirect never carries the bridge, and redirecting back re-injects correctly.
3. **Message gate** (the load-bearing one): `onMessage` drops any event whose origin is not a bridge origin.

Plus cheap native-layer hardening: `originWhitelist` https-only (widened to http only when the entry URL itself is http — Local instance / dev screen), `setSupportMultipleWindows={true}` (**deliberately not `false`** — see "Review follow-ups" below; `false` makes the popup handler dead code on Android), and `onOpenWindow` routing https popups to the system browser and dropping everything else.

The new `app/utils/webview-origin.ts` helpers are fail-closed (`URL.origin` exact matching — no substring logic; unparseable/`about:blank`/`javascript:`/`data:`/`file:` URLs match nothing, and a misconfigured Custom-instance URL shrinks the allowlist rather than opening it).

## Decisions & rejected alternatives

- **Bridge origins = fiat only.** KYC/Onfido pages have no use for `window.webln`; the only live handler exists for the buy/sell flow. Under `allowArbitraryUrl` the entry origin doubles as the bridge origin so WebLN stays testable from the developer screen.
- **Allowlist derived from the active instance, not hardcoded** — Main/Staging/Local/Custom keep working with zero config.
- **Off-origin https navigation stays allowed.** Onfido KYC redirects through its own domains mid-flow; hard-blocking navigation would break KYC. The boundary is "no bridge off-origin", not "no navigation off-origin".
- **No `onShouldStartLoadWithRequest` policy** — it misses iOS server-side redirects and is platform-asymmetric for iframes; building the boundary there would be fragile. Scheme filtering is delegated to the natively-enforced `originWhitelist`.
- **Dev escape hatch is a route param** (`allowArbitraryUrl`, set only by the dev-gated developer screen). The `webView` route is not in the deep-link config, so the param cannot be forged from outside the app.

## Accepted residual risks

- A hostile iframe *inside* an allowlisted Blink-owned page is within the page owner's trust domain.
- The page-title-as-header behavior is left as is — entry is now origin-restricted, and changing header behavior would touch KYC UX.
- The audit's KYC-token-in-query-string observation is a server-side URL design issue, out of scope here.

## Tests

- New `__tests__/utils/webview-origin.spec.ts` (22 cases): normalization, fail-closed set, substring attacks (`kyc.blink.sv.evil.com`, `evil.com/kyc.blink.sv`), scheme/port confusion, dedupe/drop.
- `webview-screen.spec.tsx` extended (24 cases): bridge injected on fiat origin / not after off-origin load / never on KYC; off-origin messages dropped; refused entry never mounts; dev escape hatch works; hardening props asserted. Existing fixtures moved from `example.com` hosts to the Main-instance origins the test context pins.

---

## Review follow-ups (a0b49cd)

[a0b49cd](https://github.com/blinkbitcoin/blink-mobile/commit/a0b49cd1c30f3d787cd20210349a7c7c90271389) addresses findings from review of e4c0b50.

**`setSupportMultipleWindows` is now `true`, and that is load-bearing.** Setting it `false` made `onOpenWindow` dead code on Android: the open-window event is dispatched only from [`RNCWebChromeClient.onCreateWindow`](https://github.com/react-native-webview/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebChromeClient.java#L89), which the platform calls only while multiple windows are supported ([the library's own default is `true`](https://github.com/react-native-webview/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt#L79)). So `window.open` / `target=_blank` navigated *this* trusted WebView in place — the opposite of the intent, and the one case the hardening was meant to cover. With the handler registered, react-native-webview drops the popup itself, so nothing opens in-app either way. The old prop was asserted by a test, which is why the contradiction survived; the test now asserts `true` and says why.

**Blank Custom-instance URLs no longer lock the user out.** The developer screen seeds every URL input with `""` when the current instance is not Custom and [saves them verbatim](https://github.com/blinkbitcoin/blink-mobile/blob/a0b49cd1c30f3d787cd20210349a7c7c90271389/app/screens/developer-screen/developer-screen.tsx#L92-L98), and `resolveGaloyInstanceOrDefault` only backfilled `undefined`. The empty strings therefore overrode the Main defaults, `entryOrigins` came out empty, and every KYC and buy/sell entry failed the check behind the generic error — with nothing in the logs to distinguish it from a genuine block. Fixed [in the resolver rather than at the developer screen](https://github.com/blinkbitcoin/blink-mobile/blob/a0b49cd1c30f3d787cd20210349a7c7c90271389/app/config/galoy-instances.ts#L80-L92): blank now counts as absent for *every* field, so `posUrl`, `lnAddressHostname` and the rest get the same protection instead of one patched call site.

**The injection latch is a ref, and latches before injecting.** Two native progress callbacks can arrive before React re-renders; both passed the old state guard, double-injecting the WebLN shim. `react-native-webln`'s `inject.ts` accumulates rather than replaces, so duplicate listeners and pollers registered and one `sendPayment` could navigate twice. This one predates the PR but sits in a function it rewrites, and the PR made the origin decision depend on the same timing.

**Three silent-failure paths now leave a breadcrumb** (all `expected: true`, so they are Crashlytics breadcrumbs, not non-fatals): a refused entry URL records which origin was refused and what was allowed; a fiat flow that hops off-origin records that the bridge was withheld; a dropped `window.open` target records the scheme. The bridge breadcrumb is gated on the entry itself being a bridge origin, so KYC — which never has a bridge — stays quiet.

**The `allowArbitraryUrl` invariant is now tested.** Its safety rested on "the `webView` route is not deep-linkable", asserted only in a doc comment, with nothing to catch a later PR adding a deep link for it. The screens map moved to `app/navigation/deep-link-screens.ts` and [a test locks it](https://github.com/blinkbitcoin/blink-mobile/blob/a0b49cd1c30f3d787cd20210349a7c7c90271389/__tests__/navigation/deep-link-screens.spec.ts), top level and nested, plus a positive assertion so the traversal cannot go vacuously green.

### Reviewed and deliberately not changed

- **`window.opener` is severed for https popups.** Routing them to the system browser does break the opener/`postMessage` channel that popup-based 3DS and bank SSO use. Kept anyway: pre-PR those popups landed in a `WebView` that `onCreateWindow` creates but never attaches to a view hierarchy, so the user could not see or interact with them either. If a partner flow turns out to need this, the fix is an in-app popup screen, not re-opening in-place navigation.
- **`onOpenWindow` forwards any https target without a host allowlist.** An external browser open exposes no app credentials or session, and scoping it to instance origins would break ordinary help/ToS links. The scheme check stays the boundary.
- **`originWhitelist` widening to `http://*` for an http entry** was raised as a host-level wildcard. It is — but `extractOrigin` in the library matches scheme+host+port, the pre-PR code passed no `originWhitelist` at all, and this only applies to a Local/dev entry. Unchanged prior default, not a regression from this PR.
- **The message gate reads the top-frame URL,** so a third-party iframe embedded by a bridge origin passes it. Already listed as an accepted residual risk above; the code comment now says so explicitly instead of implying frame-level identity.

### Tests

134 tests green across the webview, config, navigation and authentication suites. Each new test was verified non-vacuous by mutation: reverting the popup prop, the ref latch, and the blank-string filter each fails exactly its own test and nothing else.

- `webview-screen.spec.tsx`: multiple-window support asserted on with the Android reasoning; two progress events in one load inject each payload once; sub-threshold progress injects nothing; off-origin load now asserts theme JS *is* re-injected (previously the bridge count alone could not distinguish a working origin gate from a stuck latch); breadcrumbs for refused entry, withheld bridge and dropped popup targets, and the KYC quiet case; `http:` popup target dropped alongside `javascript:`.
- `galoy-instances.spec.ts`: blank and whitespace-only Custom URLs fall back to Main defaults while non-blank persisted values survive.
- `deep-link-screens.spec.ts` (new): `webView` absent top level and nested, with a positive control.

### DIFF
--- a/__tests__/config/galoy-instances.spec.ts
+++ b/__tests__/config/galoy-instances.spec.ts
@@ -74,6 +74,36 @@ it("backfills fields missing from a persisted Custom instance with Main defaults
   expect(res.blockExplorer).toBe("https://mempool.space/tx/")
 })
 
+it("backfills fields persisted as blank strings on a Custom instance", () => {
+  // The developer screen seeds every URL input with "" when the current instance
+  // is not Custom and saves them verbatim. Keeping those blanks would override
+  // the Main defaults: an empty kycUrl/fiatUrl empties the WebView entry
+  // allowlist and locks KYC and buy/sell out behind a generic error.
+  const customInstanceWithBlanks = {
+    id: "Custom",
+    name: "Custom",
+    graphqlUri: "https://api.custom.com/graphql",
+    graphqlWsUri: "ws://ws.custom.com/graphql",
+    authUrl: "https://api.custom.com",
+    posUrl: "",
+    kycUrl: "",
+    fiatUrl: "   ",
+    lnAddressHostname: "custom.com",
+    blockExplorer: "https://mempool.space/tx/",
+    sparkExplorer: "https://sparkscan.io/tx/",
+  } as never
+
+  const res = resolveGaloyInstanceOrDefault(customInstanceWithBlanks)
+
+  expect(res.kycUrl).toBe(GALOY_INSTANCES[0].kycUrl)
+  expect(res.fiatUrl).toBe(GALOY_INSTANCES[0].fiatUrl)
+  expect(res.posUrl).toBe(GALOY_INSTANCES[0].posUrl)
+  // non-blank persisted values are still kept
+  expect(res.id).toBe("Custom")
+  expect(res.graphqlUri).toBe("https://api.custom.com/graphql")
+  expect(res.lnAddressHostname).toBe("custom.com")
+})
+
 it("backfills fields persisted as undefined on a Custom instance", () => {
   const customInstanceWithUndefined = {
     id: "Custom",
--- a/__tests__/navigation/deep-link-screens.spec.ts
+++ b/__tests__/navigation/deep-link-screens.spec.ts
@@ -1,0 +1,41 @@
+import { DEEP_LINK_SCREENS } from "@app/navigation/deep-link-screens"
+
+/**
+ * The webView route's `allowArbitraryUrl` param bypasses the WebView entry-origin
+ * allowlist (see stack-param-lists.ts). That is only safe while no deep link can
+ * reach the route: a crafted `blink://…` URL carrying the param would otherwise
+ * load an attacker origin in the trusted WebView, WebLN bridge included.
+ *
+ * These tests fail the moment someone makes the route deep-linkable, which is the
+ * signal to re-derive the bypass rather than to delete the assertion.
+ */
+
+type ScreenConfig = Record<string, unknown>
+
+const collectRouteNames = (screens: ScreenConfig): string[] =>
+  Object.entries(screens).flatMap(([name, value]) => {
+    const nested =
+      value && typeof value === "object" && "screens" in value
+        ? collectRouteNames((value as { screens: ScreenConfig }).screens)
+        : []
+    return [name, ...nested]
+  })
+
+describe("deep-linkable routes", () => {
+  it("does not expose the webView route at the top level", () => {
+    expect(Object.keys(DEEP_LINK_SCREENS)).not.toContain("webView")
+  })
+
+  it("does not expose the webView route nested under any navigator", () => {
+    expect(collectRouteNames(DEEP_LINK_SCREENS as ScreenConfig)).not.toContain("webView")
+  })
+
+  it("still exposes the routes deep links depend on", () => {
+    // Guards the traversal itself: a helper that silently returned nothing would
+    // make the assertions above vacuous.
+    const routeNames = collectRouteNames(DEEP_LINK_SCREENS as ScreenConfig)
+    expect(routeNames).toContain("Home")
+    expect(routeNames).toContain("circlesDashboard")
+    expect(routeNames).toContain("sendBitcoinDestination")
+  })
+})
--- a/__tests__/screens/webview-screen.spec.tsx
+++ b/__tests__/screens/webview-screen.spec.tsx
@@ -1,19 +1,36 @@
 import React from "react"
 import { it } from "@jest/globals"
+import { Alert } from "react-native"
+import type { ReactTestInstance } from "react-test-renderer"
 import { render, waitFor } from "@testing-library/react-native"
 
+import { injectJs } from "react-native-webln"
+
 import { WebViewScreen } from "@app/screens/webview/webview"
+import { RootStackParamList } from "@app/navigation/stack-param-lists"
 import { ContextForScreen } from "./helper"
 
+type TestRoute = {
+  key: string
+  name: "webView"
+  params: RootStackParamList["webView"]
+}
+
+const mockInjectJavaScript = jest.fn()
+
 jest.mock("react-native-webview", () => {
+  const ReactActual = jest.requireActual<typeof React>("react")
   const { View } = jest.requireActual("react-native")
 
-  const MockWebView = React.forwardRef<
-    React.ComponentRef<typeof View>,
-    React.ComponentProps<typeof View>
-  >((props, ref) => {
-    return <View ref={ref} {...props} testID="webview" />
-  })
+  const MockWebView = ReactActual.forwardRef<unknown, React.ComponentProps<typeof View>>(
+    (props, ref) => {
+      ReactActual.useImperativeHandle(ref, () => ({
+        injectJavaScript: mockInjectJavaScript,
+        goBack: jest.fn(),
+      }))
+      return <View {...props} testID="webview" />
+    },
+  )
   MockWebView.displayName = "WebView"
 
   return {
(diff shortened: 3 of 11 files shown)


kngako (approved):
**No blocking findings.**

---

## Findings

### Finding 1 — non-blocking: fixed `dedupKey` values collapse distinct origins into one report per session

All three new breadcrumbs use constant dedup keys (`webview-entry-refused`, `webview-webln-bridge-withheld`, `webview-window-open-dropped` in `webview.tsx`), and `recordAppError` dedups on the key alone via a session-scoped `Set` (`app/utils/error-reporting.ts:76-79`). Only the first occurrence of each is ever recorded to crashlytics.

**When it comes up:** a user on a misconfigured Custom instance hits the KYC lockout, support asks them to retry, and the second refusal — now with a corrected but still wrong `kycUrl` pointing at a different origin — is silently swallowed. The breadcrumb comments in `webview.tsx` explicitly promise "record which origin was refused" so that a misconfigured instance and a genuine block can be told apart, but that only holds for the first distinct origin per session; every subsequent distinct origin is indistinguishable from a duplicate.

**Proposal:** include the offending origin in the dedup key, e.g. `dedupKey: \`webview-entry-refused:${originOf(url) ?? "unparseable"}\``, and likewise for the other two call sites. That keeps the spam protection the dedup was for (a retry loop on the same origin still collapses) while preserving the per-origin diagnostic the comments advertise. No test change is strictly needed, but asserting two different refused origins produce two `recordAppError` calls would pin the behavior.

### Finding 2 — non-blocking: the iframe caveat in the `onMessage` gate misdescribes per-platform behavior

The comment above the message gate (`webview.tsx`) states `event.nativeEvent.url` "is the TOP-FRAME url: a third-party iframe embedded by a bridge origin passes the check." That is only true for the legacy Android `addJavascriptInterface` fallback, which reports `mWebView.getUrl()` (`RNCWebView.java:449`). On iOS the value is the posting *frame's* `frameInfo.request.URL` (`RNCWebViewImpl.m:788`) — and the messaging script there is `forMainFrameOnly:YES`, so iframes cannot post at all — and on modern Android the `WebMessageListener` path reports the posting frame's `sourceOrigin` (`RNCWebView.java:256`). So on current iOS and Android the gate is actually *stronger* than documented: iframe messages arrive with the iframe's own origin and are dropped.

**When it comes up:** the residual exposure is real but narrow — a device whose WebView lacks `WEB_MESSAGE_LISTENER` falls back to `addJavascriptInterface`, which is visible to all frames, and there any third-party iframe on the fiat page can invoke `sendPayment` and push an attacker-chosen invoice into the send flow (the user still has to confirm in-app). A future reader trusting the comment could also "fix" the wrong layer — e.g. add top-frame reporting to match the comment — and accidentally weaken modern platforms to the legacy behavior.

**Proposal:** rewrite the comment to state the actual matrix: frame-level origin on iOS and modern Android, top-frame URL only on the legacy Android fallback, which is the one case where an embedded iframe passes. Keep the nonce-handshake note but scope it to the fallback path. If the legacy Android population still matters, a cheap mitigation is to drop messages whose reported URL carries no path (the `WebMessageListener` source-origin form) when the top-frame title/URL state disagrees — but correcting the comment is the minimum.

### Finding 3 — non-blocking: the refusal alert can be dismissed on Android without navigating back

The `entryAllowed === false` path shows `Alert.alert` with a single OK button whose `onPress` is the only `navigation.goBack()` (`webview.tsx`). On Android, `Alert` is cancelable by default, so the hardware back button dismisses the dialog without firing `onPress`, leaving the user on the blank `<Screen />` the early return renders.

**When it comes up:** a user on a broken Custom instance opens Buy/Sell, gets the generic error alert, taps the hardware back button out of habit, and lands on a featureless empty screen that looks like an app hang. It is recoverable — the header back control is still set via `useLayoutEffect` — but the one affordance the flow was designed around (OK → goBack) has been bypassed, and the only breadcrumb already fired once per session (Finding 1), so repeat visits surface nothing new.

**Proposal:** pass `{ cancelable: false }` to the alert so the OK button is the only way out on Android, or render the refusal inline (an error view with a back button) instead of an alert over an empty screen. The inline option also removes the effect-driven side effect entirely, which makes the refusal path testable without spying on `Alert`.

---

## Test Coverage Gaps

1. **The alert's OK → `goBack` hand-off is not asserted.** The refusal tests check `typeof buttons?.[0]?.onPress === "function"` but never invoke it, so a regression that drops the `navigation.goBack()` call (or the cancelable-dismiss path in Finding 3) stays green. Invoking the captured `onPress` and asserting on the mocked navigation would close it.

2. **Nothing pins the wrapper to `DEEP_LINK_SCREENS`.** The deep-link spec asserts against the extracted module, but no test verifies `navigation-container-wrapper.tsx` actually feeds that module into the linking config. A future refactor that re-inlines the screens config in the wrapper would keep every security assertion green while the invariant they protect (no deep link to `webView`) silently stops being enforced against the live config. A lightweight test reading the wrapper's linking object, or a lint-style import assertion, would tie the two together.

3. **The message gate's fail-closed path is only covered at the util level.** `isAllowedOrigin(undefined, …)` is tested in `webview-origin.spec.ts`, but no screen-level test sends an `onMessage` event with an empty or missing `nativeEvent.url` — which is exactly what the legacy Android bridge produces when `getUrl()` returns null mid-load. The behavior is correct (drop), but the contract is unpinned where it matters.

4. **The blank-string backfill is only tested for URL-ish display fields.** `galoy-instances.spec.ts` covers blank `kycUrl`/`fiatUrl`/`posUrl`, but not a blank `graphqlUri` — the case where a Custom instance silently falls back to the Main API while keeping a custom `authUrl`, producing a mixed-config client that talks to two backends. The current behavior may be the intended one, but a test should say so explicitly.

5. **No e2e coverage of the popup and external-browser path.** `onOpenWindow` → `openExternalUrl` is unit-tested with a mocked WebView, but the Android behavior the second commit fixes (`setSupportMultipleWindows` gating `onCreateWindow`) is precisely the kind of platform interaction mocks cannot see. A detox step that triggers a `target=_blank` link in a webview flow would catch a regression of this exact fix; if that is impractical, a comment in the e2e docs noting the gap is the fallback.

## Code Smells

- **Stale test-location reference** (`stack-param-lists.ts`, `allowArbitraryUrl` docstring): the invariant test is described as "over DEEP_LINK_SCREENS in navigation-container-wrapper," but it lives in `__tests__/navigation/deep-link-screens.spec.ts` and the module in `app/navigation/deep-link-screens.ts`. The pointer will send the next reader to the wrong file.

- **`Alert.alert("Error", "Webview not ready")` survives the refactor un-internationalized** (`webview.tsx`, `onLoadProgress` guard). AGENTS.md requires all user-facing strings via typesafe-i18n, and this PR rewrote the surrounding block while introducing LL-based alerts elsewhere in the same file — a missed opportunity, and now the file mixes both patterns.

- **`weblnHandler` is rebuilt on every render** (`webview.tsx`): `onMessageHandler(webview, {…})` runs in the component body, allocating the full handler object each render for a value only ever consumed inside the `onMessage` closure. A `useMemo` (or module-level factory taking the ref) would match the care taken with the other memos in this diff. Behaviorally harmless today.

- **The moved WebLN handler still returns fabricated success payloads** — `sendPayment` resolves `{ preimage: "preimage" }` to the page regardless of whether the user completes the payment in the send flow. Pre-existing stub, but this PR relocated the handler verbatim, and a fake preimage telling the fiat partner "paid" is the kind of placeholder that stops looking like one once it survives a security refactor untouched.

- **Blank-string backfill also strips `name: ""`** (`galoy-instances.ts`): a persisted Custom instance with a blank name now inherits Main's display name ("Blink"), so the settings UI would show the wrong instance name. Cosmetic, and blank names should not occur, but the filter is broader than the URL-motivated comment above it describes — either scope the filter to URL fields or note the side effect.

PR #4160, opened 2026-08-13
fix(security): store the active session token in the Keychain instead of AsyncStorage

grimen:
Fixes [blinkbitcoin/blink-wip#1093](https://github.com/blinkbitcoin/blink-wip/issues/1093) (close manually on merge — cross-repo `Fixes` doesn't auto-close).

## Problem

The active GraphQL session bearer token (`galoyAuthToken`) was stored in plaintext: it lived inside the `persistentState` JSON blob in AsyncStorage (unencrypted SQLite on Android, unprotected file on iOS), and the whole blob — token included — was rewritten on every unrelated state change. Meanwhile the multi-account session profiles, which contain the same tokens, already live in the Keychain via `KeyStoreWrapper`. Two extra leak surfaces beyond the issue text: failed-migration quarantine copies (`persistentStateQuarantine.<ts>`) kept raw tokens forever, and the blob's plaintext copy was continuously refreshed.

## What changed

- New Keychain slot in [`secureStorage.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/15fa186a518c96ccaf158a532a314febdf1747e7/app/utils/storage/secureStorage.ts#L142-L172): `getActiveToken` / `setActiveToken` / `removeActiveToken` under the key `galoyAuthToken`, `ACCESSIBLE.ALWAYS_THIS_DEVICE_ONLY`.
- [`persistent-state/index.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/15fa186a518c96ccaf158a532a314febdf1747e7/app/store/persistent-state/index.tsx#L73-L144): the persisted blob no longer carries the token. Load hydrates it from the Keychain and merges it into the in-memory state; a legacy blob token is adopted into the Keychain and the blob is immediately re-saved scrubbed. Save splits the write: blob-minus-token → AsyncStorage, token → Keychain only when it actually changed (a ref tracks the last-persisted token, so theme/quiz/currency churn never touches the Keychain).
- Quarantine hygiene: new quarantine writes redact the token to `"[REDACTED]"`; pre-existing quarantine keys are scrubbed by a fire-and-forget pass at boot.
- Fresh install (`MigrationStatus.NoData`) removes any leftover Keychain token — the iOS Keychain survives uninstall, and without this a reinstall would silently resurrect the previous session (a behavior change from today).
- Full logout explicitly removes the Keychain token alongside the other Keychain removals in [`use-logout.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/15fa186a518c96ccaf158a532a314febdf1747e7/app/hooks/use-logout.ts#L55) (defense in depth; `resetState()` also zeroes it through the save path).

The in-memory `PersistentState` type still has `galoyAuthToken`, so every consumer (`useAppConfig`, Apollo client build, `use-has-custodial-account`, …) is untouched.

1. **Persistence-layer split, not a schema-17 migration.** The migration chain requires the field in every historical version and the in-memory type is consumed app-wide; a field-removal bump would churn every consumer and ~30 test fixtures, and schema 16 just landed. Rejected: formal schema bump. `migratePersistentState` is 100% untouched (its specs pass unmodified).
2. **`ALWAYS_THIS_DEVICE_ONLY`, not `WHEN_UNLOCKED_THIS_DEVICE_ONLY`.** `Info.plist` declares `UIBackgroundModes: remote-notification`, so iOS can cold-start the app while locked; a `WHEN_UNLOCKED` read would silently return `""` (KeyStoreWrapper swallows keystore errors) and the process would run logged-out. The same tokens also already sit in `sessionProfiles` at `ALWAYS_THIS_DEVICE_ONLY`, so a stronger class here buys nothing until both move together — deliberately not in this PR.
3. **Keychain wins when both stores hold a token** (`keychainToken || blobToken`). A blob token can only be the legacy copy, a failed-scrub leftover, or an Android auto-backup restore without the keystore — always equal to or older. The blob is consulted only while the Keychain slot is empty, and scrubbed either way.
4. **No scrub when adoption fails.** Destroying the only copy of the credential is worse than one more boot with the plaintext blob; the failure is reported via crashlytics and adoption retries next boot.
5. **Rejected: pointer-into-profiles design** (store only an account id, resolve the token from the Keychain `sessionProfiles` blob). Tokens can exist before a profile does (profile save happens after `saveToken`, and can fail on the `/me` fetch), so the pointer could dangle and log users out.
6. **Known cost:** downgrading to an older build after the scrub logs the user out (old builds only read the blob). Unavoidable with any scrub design.

## Verification

- 244 tests across 20 suites: `secure-storage.spec.ts` (new active-token block asserts the exact key + accessibility class), `store/persistent-state/` (hydration from Keychain, legacy adoption + immediate scrub, adoption-failure keeps the blob, Keychain-precedence, no Keychain writes on unrelated changes, failed-write retry, reinstall cleanup, quarantine redaction old + new), new `use-logout.spec.ts` (full logout clears the slot, profile logout doesn't), plus unchanged `persistent-storage.spec.ts`, `state-migrations.spec.ts`, `use-save-session-profile.spec.ts`, `use-effective-auth-token.spec.ts`, `use-has-custodial-account.spec.ts`.
- `tsc -p .` clean; ESLint clean on all touched files.

Out of scope (issue suggests them; backend or separate hardening tracks): server-side session lifetime cap, cert pinning, root/jailbreak detection, Play Integrity/App Attest, tightening `sessionProfiles` accessibility.

## Review round 2 — decisions

A review pass at 15fa186a found six blocking issues in how the boot-time migration handler reconciles the blob and the Keychain. All six are fixed, with tests locking each path. Judgement calls:

1. **A failed migration now recovers the session from the Keychain** (defaults + Keychain token) instead of booting logged out. Rejected: clearing the token — a corrupt settings blob shouldn't cost an intact credential, and the previous middle ground (logged out now, token orphaned in the Keychain, session silently resurrected one boot later) was the worst of both.
2. **An unrecognized `schemaVersion` is now `Failed`, not `NoData`.** A downgrade from a future build must not be classified as a reinstall and have its credential wiped. Landed in `state-migrations.ts` because that's where the classification is made; patching around it in the caller would leave the same trap for future callers.
3. **The load path now returns `persistedToken` alongside the state** — what the Keychain durably holds — and the provider seeds its dirty-check ref from that. A failed legacy-token adoption therefore retries on the first save instead of silently skipping the Keychain write and destroying the only copy.
4. **The reinstall wipe now also clears `sessionProfiles`** (reachable through the account switcher, same uninstall-surviving Keychain), and retries + reports instead of failing silently. Not covered: per-account mnemonic entries — the account index lives in AsyncStorage and doesn't survive uninstall, so they can't be enumerated at wipe time. Known limitation for a follow-up.
5. **The quarantine scrub isolates failures per entry** (one corrupt entry no longer aborts the sweep) and writes a done-marker after one clean pass — post-move quarantine writes are redacted at write time, so a clean sweep is permanent and the boot-time key scan stops running forever.
6. **The token-free blob is now a compile-time guarantee**: the only blob writer takes `Omit<PersistentState, "galoyAuthToken"> & { galoyAuthToken?: never }`, and saves are serialized through a queue so the last-persisted-token ref is provably single-writer.

## Before vs after — proof

**Executable.** The round-2 spec files, run unchanged against the pre-fix head [`15fa186a`](https://github.com/blinkbitcoin/blink-mobile/commit/15fa186a518c96ccaf158a532a314febdf1747e7): **9 failed, 49 passed** — each failure reproduces one review finding (failed-adoption retry, session-profile wipe, keychain recovery on failed migration, unrecognized-schema classification ×2, wipe retry+report, corrupt-quarantine isolation, sweep done-marker ×2). Same specs at the current head: all pass, and the full CI suite is green ([Tests run](https://github.com/blinkbitcoin/blink-mobile/actions/runs/31791123740/job/94738036712), 7332 tests). The 49 passing in both runs are the pre-existing suite — the fixes changed no behavior the PR had already locked.

**Visual.** Same simulator, same binary (the six fix commits are pure TS; only the Metro-served JS differs), same persisted state: a logged-in staging session whose `persistentState` blob was then set to `schemaVersion: 999` — the downgrade-from-a-future-build scenario. Old code classifies that blob as a fresh install and deletes the keychain token; new code routes it to Failed and recovers the session from the keychain.

| Logged in (baseline) | Corrupt blob → old code `15fa186a` | Corrupt blob → new code |
|---|---|---|
|  |  |  |
| Real staging session (token obtained via the `userLogin` API and seeded into the keychain by a TEMP boot line, removed before any shot) | Session destroyed: token wiped, welcome screen | Session recovered: home screen. Balances shimmer because recovery resets settings — including the instance — to defaults (Main), while this demo account lives on staging; a production user would reload normally |

<sub>Images live on branch `assets/pr-4160-demo` (embedded by commit SHA) — deleting that branch breaks them.</sub>

**Round-2 follow-up:** the reinstall wipe list now lives in `KeyStoreWrapper.clearUninstallSurvivingCredentials` (secure-storage layer) instead of inline in the state loader — adding a future uninstall-surviving keychain slot and adding it to the wipe are now the same edit in the same file. The loader keeps only the trigger (fresh install) and the crashlytics reporting. Behavior unchanged; retry/report semantics fully spec-locked at the storage layer (5 specs), loader wiring locked at the provider (3 specs), key paths mutation-checked. Mnemonics remain deliberately excluded (documented in the method).

## Grafted from PR #4139

[#4139](https://github.com/blinkbitcoin/blink-mobile/pull/4139) (kngako) solves the same issue with a different architecture (hook-level Keychain mirroring). Two of its hardenings are independent of that architecture and are grafted here, adapted to this PR's API, in [`55bbb39a`](https://github.com/blinkbitcoin/blink-mobile/commit/55bbb39a2aa2e2f8238e70ac3ea33084cc88439e):

1. **The logout device-token fetch is isolated.** Previously a failed Firebase `getToken()` threw ahead of every removal, skipping the entire local key-store cleanup (this PR's save-path mirroring would still clear the active token via `resetState`, but PIN, biometrics flag and `sessionProfiles` stayed behind). Now the fetch failure is reported and only the server-side revocation is skipped — forced, since `UserLogoutInput.deviceToken` is non-null.
2. **Removing the profile that backs the active session also drops the Keychain token.** The account-switch flow saves the next token immediately afterwards, so in the happy path this changes nothing; it closes the crash window in between, where the Keychain would otherwise hold a token whose profile is gone and resurrect that half-dead session on next launch.

Also adapted from #4139: a spec for `use-switch-to-next-profile` locking the deactivate-before-save ordering. Its `use-app-config` spec was NOT adapted — it tests hook-level mirroring, which doesn't exist in this design.

## Review round 6

Answering [kngako's review](https://github.com/blinkbitcoin/blink-mobile/pull/4160#pullrequestreview-4941589002) at `55bbb39a2`; full item-by-item response in [this comment](https://github.com/blinkbitcoin/blink-mobile/pull/4160#issuecomment-5316011831). All six findings reproduced against the code before acting; fixes in [`d0146250b`](https://github.com/blinkbitcoin/blink-mobile/commit/d0146250befdfb2c6b0fa92526e3daf645a2c7db). Judgement calls recorded here:

1. **An unreadable blob no longer takes the fresh-install branch.** `loadJson` reported "key absent" and "bytes unparseable" identically, so a truncated blob ran the reinstall wipe — deleting the active token *and* `sessionProfiles`, which also empties the account registry derived from them. That was strictly worse than the pre-branch behaviour it replaced. The loader now reads text and parses it itself; `loadJson` is left alone so the migration-checkpoint callers are unaffected.
2. **An unparseable blob is quarantined as a description, not as bytes.** `redactToken` can only redact a parsed object; a damaged string may be cut mid-token, so no pass can prove the credential is gone. Quarantine therefore stores `unparseable` / `byteLength` / `parseError` — enough to tell truncation from garbage, nothing to leak. Rejected the alternative (regex-redact the raw string) because it cannot promise what this branch exists to promise.
3. **An empty-string blob counts as damage, not as a fresh install.** A zero-length value is not how an absent key reads, so it is a write that went wrong — and must not be answered by deleting credentials.
4. **Known residual, stated not hidden:** `loadString` still swallows a throwing `getItem` into `null`, so a failed read is still indistinguishable from an absent key. Fixing that means changing `loadString`'s contract for every caller; documented at the call site instead.
5. **`getAllKeys` returns `null` on failure.** Swallowing to `[]` let the sweep mark itself done having read nothing. The done-marker is kept (boot should not re-list every key each launch) and simply withheld unless the listing succeeded. Cheap because this PR added `getAllKeys` and has its only call site.
6. **Miss-vs-error discrimination lives in `secureStorage`.** `RNSecureKeyStore` *rejects* a missing key, so "a successful read that comes back empty" does not exist and the reviewer's proposed guard was unimplementable. Both native modules reserve code `404` for a missing key, so `readActiveToken` keys off that; adoption is skipped on any other rejection rather than overwriting a possibly-newer keychain token with the older blob copy. `getActiveToken` stays as the collapsing convenience for callers where an empty result leads to doing nothing.
7. **Active-token removal goes through the provider.** `clearToken()` on the context removes durably through the save queue, retries once, reports on persistent failure, and updates the dirty-check ref. `use-logout` no longer touches that keychain slot at all. Kept it awaited rather than routing through `updateState`: React would batch the clear away against the caller's next `saveToken`, reopening the crash window the round-5 graft closed.
8. **The blob-then-keychain crash window is documented, not engineered away.** Reordering only moves the mismatch and the two stores cannot be written atomically. The suggested reconciliation (validate the token against the profile for `activeAccountId`) does not fit — `activeAccountId` is a self-custodial account id / the `Custodial` sentinel, not a profile key. A stale token beside newer settings fails the next authenticated call and takes the existing 401 path; a test pins the order so the choice is explicit.
9. **Declined smells, with reasons:** comment density (the rationale is security reasoning needed at the point of edit), the duplicated `galoyAuthToken` literal (the type pin makes a mismatch a `tsc` error — that is the point), the scrub's `.catch(() => {})` (verified not lint-mandated; kept as insurance against `recordAppError` throwing on a floating promise), and quarantine retention (accurate, pre-existing, out of scope).

Coverage: all six named gaps closed, plus `readActiveToken` and `getAllKeys` unit tests. Four mutations verified caught — corrupt-blob routing, listing-failure marker, adoption read guard, removal retry.


kngako (changes requested):
## Findings

### Finding 1 — BLOCKING: an unreadable blob takes the NoData branch and wipes the keychain session

`loadJson` (`app/utils/storage/storage.ts:57-64`) returns `null` in three different situations — key absent, `AsyncStorage.getItem` throws, and `JSON.parse` throws on a corrupt/truncated blob. `loadPersistentState` routes all three to `MigrationStatus.NoData` (`state-migrations.ts:343-344`), which now triggers `KeyStoreWrapper.clearUninstallSurvivingCredentials` — destroying the active token and every session profile (`index.tsx:163-174`).

**When it comes up:** a user's `persistentState` blob is truncated by an interrupted write, or AsyncStorage has a transient read failure at boot. Before this branch the token lived in the blob, so a corrupt blob meant a lost session anyway; now the token deliberately lives in the keychain to survive exactly this — the Failed branch even comments "losing settings must not cost the session" — yet corrupt JSON never reaches Failed. One storage hiccup at boot permanently logs the user out and deletes all session profiles, which is worse than the pre-branch behavior the change set out to improve.

**Proposal:** make `loadJson` distinguish "absent" from "unreadable" — e.g. return `null` only when `getItem` returns null, and a sentinel (or throw) when `getItem`/`JSON.parse` fails. Route the unreadable case to `MigrationStatus.Failed` with the raw string as `rawData`, so it gets quarantined and the keychain is preserved. Only a confirmed-absent key should ever trigger the reinstall wipe.

### Finding 2 — BLOCKING: the quarantine scrub marks itself permanently done when `getAllKeys` fails

`getAllKeys` swallows errors and returns `[]` (`storage.ts:35-41`). Inside `scrubQuarantinedTokens` (`index.tsx:56-104`), an empty key list means zero quarantine entries, `allClean` stays `true`, and the `persistentStateQuarantineScrubDone` marker is written — so the sweep never runs again.

**When it comes up:** a device upgraded from a version that wrote raw-token quarantine entries boots once with a flaky AsyncStorage (low storage, OS-level I/O error). The listing fails, returns `[]`, the marker is persisted, and the legacy quarantine copies keep holding plaintext session tokens forever — the exact artifacts commits `6883dfc1f`/`13528d6d0` exist to redact, silently skipped for the life of the install.

**Proposal:** have `getAllKeys` surface failure (return `null` or throw) and write the done-marker only when the key listing genuinely succeeded and every entry was verified clean. Alternatively drop the marker entirely: redact-at-write already covers new entries, and an idempotent re-sweep each boot is cheap relative to app startup. If the marker stays, a test for the listing-failure path is the minimum bar.

### Finding 3 — non-blocking: a transient `getActiveToken` failure can overwrite a newer keychain token with a stale blob token

`getActiveToken` returns `""` both when the slot is empty and when the keystore read throws (`secureStorage.ts:150-156`), and `loadPersistentState` treats `""` as "adopt the blob token" (`index.tsx:139-143`). Adoption is unconditional once the read comes back empty.

**When it comes up:** the keychain holds a fresh token T1, the blob still carries legacy token T0 because an earlier boot-time scrub write failed, and the keychain read throws transiently at the next boot. `setActiveToken(T0)` then overwrites T1, the scrub erases T0 from the blob, and the user is left on a stale — possibly already server-side revoked — session with no trace of T1 anywhere.

**Proposal:** distinguish "confirmed empty" from "read failed" before adopting — for example, retry the read once (mirroring the `removeWithRetry` pattern already in this diff) and skip adoption unless a read succeeds and comes back empty. If the keystore API can differentiate miss-vs-error, plumb that through instead of collapsing both to `""`. The failure case should take the Failed-branch posture: report, keep the blob copy untouched, and let the next boot retry.

### Finding 4 — non-blocking: the profile-logout keychain removal failure is swallowed

`use-logout.ts:57-60` awaits `KeyStoreWrapper.removeActiveToken()` but ignores its `false` return and reports nothing. The surrounding comment states the purpose: preventing a crash from resurrecting a session whose profile is gone.

**When it comes up:** a user deactivates the profile backing the active session and the keystore removal fails (the same transient keystore errors the reinstall wipe already retries against). The profile is gone but the keychain token survives, so the next boot restores a session with no backing profile — precisely the resurrection this commit fixes — with no error reported to indicate it happened.

**Proposal:** check the boolean result, retry once (reuse the `removeWithRetry` shape from `clearUninstallSurvivingCredentials`), and `recordAppError` on persistent failure. This mirrors the failure handling the same diff already applies to the reinstall wipe, so it is a consistency fix as much as a robustness one.

### Finding 5 — non-blocking: blob-then-keychain write order opens a crash window pairing fresh settings with a stale token

`savePersistentState` (`index.tsx:190-215`) writes the AsyncStorage blob first and the keychain token second. Before this branch, token and settings were a single atomic blob write; the split introduces a window where the two stores disagree.

**When it comes up:** a user switches accounts — one `updateState` sets `galoyAuthToken: B` and `activeAccountId: B` together — and the app is killed between the blob write and the keychain write. The next boot loads account B's settings with account A's still-persisted keychain token, pairing one account's UI state with another account's session credential.

**Proposal:** the window cannot be closed by reordering alone (token-first just moves the mismatch), so reconcile at load: validate the keychain token against the session profile for `activeAccountId` — profiles live in the keychain and are authoritative — or store `activeAccountId` alongside the token as a single keychain value so the pair updates atomically. At minimum, document the ordering choice and its failure mode in the code, since the queue comment currently claims serialization is the only invariant that matters.

### Finding 6 — non-blocking: `use-logout` mutates the keychain behind the provider's dirty-check ref

The provider's `lastPersistedTokenRef` is the single source of truth for "what the keychain durably holds" (`index.tsx:117-123`), but `use-logout` writes and removes the active token directly (`use-logout.ts:57-60,68`) without updating the ref. When `switchToNextProfile` finds no next profile, the keychain token is removed while state and ref still hold it.

**When it comes up:** the last remaining profile is deactivated, no new token is saved, and the user keeps the app open — later `updateState` calls produce saves that see `galoyAuthToken === lastPersistedTokenRef.current` and skip the keychain write, so the keychain stays empty while the in-memory session looks alive. The next boot finds neither keychain nor blob token and silently logs the user out, despite every save in between reporting success.

**Proposal:** route active-token removal through the provider (e.g. an explicit `clearToken` on the context, or `updateState` clearing `galoyAuthToken`) so the ref and save queue remain the only writers. If a direct keychain write is unavoidable in logout, at least invalidate the ref alongside it — the invariant "ref equals durable keychain content" is load-bearing for the retry logic and should be maintained by construction, not by convention.

---

## Test Coverage Gaps

1. **No test that a corrupt/unparseable blob spares the keychain.** The specs cover absent (`null`) blobs and unrecognized-schema blobs, but nothing makes `JSON.parse` throw or `getItem` reject — the exact inputs that Finding 1 shows are misrouted to the reinstall wipe. A test asserting "unreadable blob → no `clearUninstallSurvivingCredentials` call, keychain token survives" would catch the regression.

2. **No test for `getAllKeys` failure during the scrub.** The quarantine-hygiene describe block covers corrupt entries, failed redaction writes, and the done-marker skip — but not a failed key listing, which Finding 2 shows writes the marker and permanently disables the sweep. Mock `getAllKeys` to reject and assert the marker is withheld.

3. **No test for a transient `getActiveToken` failure during adoption** (Finding 3). The adoption describe block covers failed *adoption writes* and stale-blob preference, but never a failed *read* with a legacy blob token present — the combination that overwrites a newer keychain token.

4. **No test for `removeActiveToken` returning `false` in logout** (Finding 4). The logout spec asserts the happy-path removal calls but never exercises a rejected keystore removal, so the swallowed-failure behavior is locked in rather than guarded.

5. **No test pinning the blob/keychain write order or its crash semantics** (Finding 5). The serialization test proves saves are queued, but nothing documents that the blob lands before the keychain write within a single save, or what a boot should do when they disagree. An order-asserting test would at least make the invariant explicit for whoever touches it next.

6. **No integration-style test that logout-induced keychain writes stay consistent with the provider ref** (Finding 6). The logout and switch-to-next-profile specs mock `KeyStoreWrapper` and the persistent-state context separately, so the ref/keychain divergence in the no-next-profile path is invisible to the suite.

## Code Smells

- **Comment essays dominate `index.tsx`**: nearly every statement carries a multi-line rationale comment (~a third of the file is prose). The security reasoning is genuinely valuable, but at this density the code is hard to scan; the design narrative (adoption ordering, reinstall contract, quarantine lifecycle) belongs in `docs/` with terse pointers in the code.

- **The `galoyAuthToken` literal is duplicated across two modules** (`secureStorage.ts:5`, `index.tsx:31`) with a type-level pin whose comment admits it exists only because "specs mock this module wholesale." Test-mock strategy is shaping production code structure; unmocking the constant (or exporting it from a tiny unmocked module) would remove the workaround and the comment that explains it.

- **`removeWithRetry` is an opaque one-liner** (`secureStorage.ts:405-410`): `(await remove()) || (await remove())` retries immediately with no backoff, which is unlikely to help the transient keystore failures it targets, and reads as a trick rather than a policy. A small loop with a named constant (and ideally a micro-delay) would state the intent.

- **Redundant `.catch(() => {})` on `scrubQuarantinedTokens`** (`index.tsx:127`): the function already wraps its entire body in try/catch and never rejects, so the caller-side swallow is a second defensive layer that obscures the actual error contract — either the function guarantees no-throw (then drop the catch) or it doesn't (then the internal catch is misplaced).

- **Quarantine entries accumulate forever**: `quarantineRawState` writes a new `persistentStateQuarantine.<timestamp>` key on every failed boot and nothing ever evicts them (pre-existing, but the new sweep is the natural place to add retention). On a device with a persistently failing migration this grows unboundedly.

- **`let context` in `use-logout.ts:51` is declared untyped** and assigned only in one branch; giving it an explicit `Context` type (or building it at the mutation call site) would remove the implicit-any-adjacent pattern in a security-sensitive path.

- **`loadPersistentState` carries four responsibilities** (scrub trigger, migration dispatch, keychain adoption, reinstall wipe) in one switch; the per-status branches are already self-contained blocks with their own comments, so extracting `handleOk`/`handleNoData`/`handleFailed` helpers is a low-risk readability win.

kngako (approved):

PR #4161, opened 2026-08-13
fix(security): fall back to the app PIN in the local auth gates instead of failing open

grimen:
Fixes blinkbitcoin/blink-wip#1064

## What

`useBiometricGate` had two paths that set `authenticated = true` without ever prompting, both live on the screen that reveals a funded wallet's 12-word seed:

1. **PIN never consulted** — with the app's biometrics toggle off, the gate opened immediately, even when the user had deliberately set an app PIN. A PIN-only user got *no challenge at all* before the phrase.
2. **Sensor-gone fail-open** — `required` defaulted to `false`, so an unavailable (or erroring) sensor also opened the gate, including for users who *had* biometrics enabled but lost enrolment.

This PR replaces it with `useLocalAuthGate`, which mirrors the PIN-OR-biometrics disjunction the app lock already uses ([`app-state.tsx#L62-L65`](https://github.com/blinkbitcoin/blink-mobile/blob/9b889f63b62c46951d4ba46280da9700a5a94484/app/navigation/app-state.tsx#L62-L65)) and fails closed everywhere a configured factor can't be satisfied.

## How — four commits, reviewable in order

1. **`feat(pin-screen)`** — a new `PinScreenPurpose.ChallengePin` verifies the app PIN for a calling screen and reports back through optional route-param callbacks (the `selectionScreen.onSelect` precedent). A challenge never touches the app lock: success resolves to the caller instead of `completeUnlock`, dismissal (swipe/back) counts as a decline exactly once via a `beforeRemove` listener, and the third wrong guess keeps the app lock's logout-and-reset answer. Inert on its own.
2. **`feat(hooks)`** — `useLocalAuthGate({ description, onFailure, required = true })`: biometrics when available; prompt failure, cancel, or a missing sensor falls through to the PIN challenge (or fails, never opens); `required` now governs **only** the no-factor-configured case. Inert on its own.
3. **`fix(security)`** — migrates all three call sites and deletes the old hook. The backup-phrase screen also defers the mnemonic read: `useViewBackupPhrase()` used to run before authentication (only rendering was gated); the words and the header Copy handler now exist only after the gate passes.
4. **`fix(biometrics)`** — stops passing `fallbackEnabled: true`. The library's iOS policy is hardcoded biometric-only, so that flag only rendered a fallback button whose tap could only ever reject with `UserFallback` — an affordance that promised a fallback and delivered a dismissal.

## Behaviour matrix

| Screen | No factor configured | Biometric fails / sensor gone |
|---|---|---|
| View recovery phrase | opens, as today (explicit `required: false`) | PIN challenge if set, else back |
| Card details | toast + back (**was: opened silently**) | PIN challenge if set, else toast + back |
| Card change PIN | toast + back (as before) | PIN challenge if set (**was: dead end**), else toast + back |

## Judgement calls (and rejected alternatives)

- **Lockout at 3 wrong guesses logs out and resets, same as the app lock.** The keystore `pinAttempts` counter is shared with the app lock. The rejected alternative — fail the challenge and go back — would hand out one fresh guess per re-entry into any gated screen: an unbounded brute-force channel against the PIN that protects the whole app.
- **Callbacks in route params** rather than a context/event bus: it is the repo's one existing result-to-caller pattern (`selectionScreen.onSelect`), keeps the challenge stateless, and the caller stays mounted beneath the pin screen so the closures stay live. Caveat: non-serializable params are dropped on state rehydration after process death; both callbacks are optional and every call is optional-chained, so a restored challenge degrades to fail-closed (re-entering the PIN pops back to the caller's spinner and the user backs out).
- **Backup phrase stays reachable with no lock configured** (explicit `required: false` + comment). Product decision from the issue discussion: the recovery phrase is the user's escape hatch to their own funds and must not be blocked behind lock setup. Any factor the user *did* configure is now enforced — which is the entire fix.
- **Card details fails closed for no-lock users** (a behaviour change): the screen is a convenience view of remote card data and is built to grow PAN/CVV display, so the recovery-phrase argument does not apply. The failure toast explains and points at Security settings.
- **`required` default flipped to `true`** and narrowed: it can no longer waive the sensor-unavailable path (that *was* fail-open #2). Only opt-outs are written explicitly, with a comment.

## Tests

130 tests across six suites (measured 100% statement/branch/function/line coverage on the PR-authored files: the gate hook, pin-screen, card-change-pin, view-backup-phrase); the matrix maps to the issue's acceptance criteria:

- `use-local-auth-gate.spec.ts` (new): {pin on/off} × {bio on/off} × sensor {ok/unavailable/throws} × prompt {success/failure} × challenge {success/decline} × `required` both ways, plus late-callback-after-unmount and keystore-error cases.
- `pin-screen.spec.tsx`: challenge success never calls `setAppUnlocked`/`reset`; decline fires exactly once (double `beforeRemove` still once); wrong PIN persists the shared counter; third wrong guess logs out; existing `AuthenticatePin` cases untouched as regression proof.
- `view-backup-phrase-screen.spec.tsx`: PIN-only user is challenged, with words, header Copy button, and even the mnemonic *read* all absent until the challenge passes; sensor-gone with biometrics on never opens.
- `card-details-screen.spec.tsx`: the old default arrangement (sensor unavailable → screen renders) *was the bug* and is now the fail-closed case.
- `biometricAuthentication.spec.ts`: asserts `fallbackEnabled: false`.

i18n: two new keys (`PinScreen.enterPin`, `AuthenticationScreen.authenticationRequired` — the latter moved out of `CardFlow.*` during review, see below) shipped with translations in all 28 locales; generated types committed.

## Out of scope (follow-up issues on blink-wip)

- Gating API-key creation (the issue's item 4) — a different screen and threat model.
- Migrating off `react-native-fingerprint-scanner` for a real OS device-credential fallback (item 3) — native work on both platforms; the PIN fallback here removes the urgency.

## Review fixes (2026-08-14)

Commits addressing the code-review findings and the coverage audits, with the judgement calls behind them:

1. [`453a467a`](https://github.com/blinkbitcoin/blink-mobile/commit/453a467a51745367f55dfd82fb3ff03c0e10fee6) — **keypad input lock + keystore-authoritative attempts counter.** The lockout awaited `logout()` + a 1s sleep with the keypad still live, so a correct PIN typed in that window resolved the challenge against a dying session. A ref locked at completion-dispatch closes the window for every purpose; only try-again paths re-arm it. The local `pinAttempts` mount snapshot is gone — each attempt reads the keystore at decision time. *Rejected alternative:* re-syncing the snapshot on focus — still leaves the pre-load race and a second concurrently-mounted pin screen stale. The lockout's reset also moved into a `finally` so a rejected `logout()` cannot strand the caller (the challenge marks `challengeResolvedRef` before the lockout runs, so no failure callback would ever fire).
2. [`b1401104`](https://github.com/blinkbitcoin/blink-mobile/commit/b1401104e4a67d3c60d623959ba5db1cdf18e2c7) — **`push` instead of `navigate` + mounted guards on gate initiation.** v7's `navigate` onto a focused same-name route replaces its params, so a deferred challenge could rewrite the resume app-lock's screen (dropping `isResume` and its dismissal guards); `push` cannot coalesce. The `mounted` flag now also gates initiation: a caller popped while the keystore reads or sensor probe were in flight no longer presents a phantom challenge/OS prompt (whose wrong guesses burned the shared lockout budget). The hook doc also stops claiming parity with the app unlock: the gate is *deliberately stricter*. *Rejected alternative:* making the unlock fail closed too — that would brick a biometrics-only user with a dead sensor out of the entire app; behind an in-app gate the cost is one screen, at the unlock it's everything. A shared factor-reading helper across the three policy sites was likewise skipped: they sequence their reads differently, and touching the app-lock path here buys ~4 lines of dedup.
3. [`2b65d50d`](https://github.com/blinkbitcoin/blink-mobile/commit/2b65d50db1c63c9abdf0dbc6bfe67e3b7d4bae94) — **backup-phrase failure explains itself + copy moves to `AuthenticationScreen.*`.** The backup screen was the one gated caller with a silent `goBack()` on failure — arguably the highest-stakes bounce of the three. It now shows the same authentication-required toast. The string moved out of `CardFlow.*` so the self-custodial screen doesn't import card-namespaced copy (*rejected:* reusing the `CardFlow` key — minimal diff, but it spreads the cross-domain smell this PR otherwise removes). Every locale's existing translation moved with it, unchanged. All three callers now pass the toast message as a function so `toastShow` logs `isTranslated` correctly.

4. [`1c2e4682`](https://github.com/blinkbitcoin/blink-mobile/commit/1c2e46829ddbfb9a09844bb6b6c5d77aa2839bb9) — **full coverage for the review fixes themselves.** A coverage audit of commits 1–3 found six untested paths in the new behavior; all are now specced: keypad re-arm after a wrong non-lockout guess (unlock + challenge), the SetPin two-entry flow (its handoff runs through the same input lock — a broken re-arm would have wedged PIN creation with a green suite), backspace between attempts and during the lockout window, the reset landing when `logout()` rejects (the error is now swallowed explicitly rather than escaping as an unhandled rejection), the gate's catch handler after unmount, and the backup screen's toast on the biometric-failure path.

5. [`2ad12269`](https://github.com/blinkbitcoin/blink-mobile/commit/2ad1226914e21fcb7b5ac7c266ceb01141eec59f) — **decline semantics + one failure handler.** The challenge's `beforeRemove` treated *any* removal as the user declining; it now checks the removing action's type — pop-family (POP/POP_TO_TOP/GO_BACK, verified against the installed native-stack's dismissal dispatches) is a decline, while a RESET/REPLACE the challenge doesn't own resolves silently instead of toasting into a screen the reset just unmounted. `onFailure` now carries the reason (`noFactor` | `unavailable` | `declined`), and a shared `useAuthGateFailureHandler` replaces the three verbatim caller blocks: **deliberate declines bounce silently** (*rejected alternatives:* a short neutral toast — feedback nobody needs for their own cancel, plus 28 new locale strings; keeping the single copy — it told users who HAD factors configured to go set them up). The settings-advice toast stays for `noFactor`/`unavailable`. Also deletes the backup screen's `noHeaderRight` effect — dead since the authenticated-only component split, with a comment crediting the wrong mechanism.
6. [`adf4b10e`](https://github.com/blinkbitcoin/blink-mobile/commit/adf4b10e6751d95b86d4e120d66ed54c2922cfad) — **orphaned key removal.** `CardFlow.PinScreens.ChangeFlow.biometricRequired` had zero call sites left; dropped from en, all 28 locales, and the generated types.

7. [`cf0c8f57`](https://github.com/blinkbitcoin/blink-mobile/commit/cf0c8f57de27d067f953604767cd10f7ff06ae17) — **coverage to 100% on PR-authored files.** A scoped coverage run found three changed-code lines without a test (the handler's toast message closure, the one-attempt-remaining branch, the setPin-store-failure path through the re-armed keypad); all covered. Remaining uncovered lines in card-details/biometricAuthentication are pre-existing code this PR does not touch.

8. [`0292ebf7`](https://github.com/blinkbitcoin/blink-mobile/commit/0292ebf70b0442908d2a7cd5c9040824ec73312d) — **decline no longer strands the caller (found by demoing).** Recording the review-fix demos surfaced a live bug: declining the challenge left the gated screen on its pending spinner forever. The `beforeRemove` listener runs inside the removing pop's dispatch, so the failure handler's `goBack` coalesced with that pop and was swallowed — the challenge popped, the caller never did. The decline callback now defers one tick so the stack settles before the handler's `goBack`; a spec pins the deferral. Demo GIFs of both failure behaviors are in the PR comments.

**Deferred by decision:** the lockout's `logout()` is custodial-shaped (server `userLogout` + `SESSION_PROFILES` wipe) even when a self-custodial user exhausts the challenge on their own backup phrase. The right per-mode consequence is a product call — split out to blinkbitcoin/blink-wip#1145 rather than guessed at here.

## Review fixes (2026-09-02)

Seven of the eight findings are addressed in `df29d964`…`80b6a5f6`, one commit per area; see the PR comment for the per-point detail.

**Point 8, stated here as asked: the fail-closed behaviour for device-biometrics-only users is intended.** The gate enforces what the user configured *in the app*. The Security screen shows biometrics as off for that cohort, and the keystore cannot tell "never enabled" apart from "deliberately turned off" — prompting the OS anyway would contradict both. The failure toast points at Security settings, so the screen is one setting away rather than unreachable.


kngako (changes requested):
Inconsistency with the video and with what I get when running on an emulator. Some commits might still need to be pushed?

esaugomez31 (dismissed):
Rebased onto `main` and the review feedback applied.

**Rebase.** The branch was 47 commits behind and conflicting: #4162 had replaced the PIN attempt counter with an escalating time lockout, and #4210/#4211 moved the PIN and biometrics slots to react-native-keychain. Resolved into 5 commits — five of the original 13 only fixed a `pin-screen.tsx` that no longer exists.

Three resolutions worth knowing:

- `ChallengePin` no longer counts attempts for itself; it runs through `usePinLockout`, so it shares the one budget and lockout with the app unlock.
- `useLocalAuthGate` moved to `readIsPinEnabled` / `readIsBiometricsEnabled`. The collapsing `getIs*` variants scored a failed keystore read as "no factor configured", which on the backup screen turned straight back into the fail-open this PR exists to close.
- The gate left the `app/hooks` barrel. Re-exporting it pulled `react-native-fingerprint-scanner` into every consumer, which is what broke CI: no test failed, 32 suites failed to load.

**Review feedback** (@kngako):

- Security score showed a biometrics error instead of opening the PIN screen. Real bug, pre-existing on `main` — the `appLock` signal accepts either factor but its action only attempted biometrics. It now falls through to set-PIN. The biometrics switch keeps its error, since there the user asked for biometrics by name.
- The PIN screen has a visible dismiss control on set-PIN and challenge; the app lock keeps none. Back arrow rather than a close X, per the reply above.

**Verified:** tsc and eslint clean, 108 suites / 1379 tests green, `security-screen.tsx` at 100% across the board. Set-PIN checked on the simulator.

kngako (approved):
This is a big PR. Reviewing functionality, as I think there has been enough rounds, it works well.

Find below test notes.

---


Changes function as expected.

image

image


---

Once Pin is set the app asks for the PIN when the users account gets opened.

image


image


---

The PIN gets applied on all the users accounts. Which helps avoid confusion.

---

When a user selects opens Blink via an intent (pay an invoice from a website) the PIN lock intercepts the intent and once the user enters the correct PIN it opens the intended intent.

image

image

---

Incorrect PIN attempts are persisted between session. And reset once successfully input.

image

PR #4162, opened 2026-08-13
feat(security): escalating time lockout for PIN attempts

grimen:
Closes blinkbitcoin/blink-wip#1140 — item 2 of the red-team hardening epic [blink-wip#1092](https://github.com/blinkbitcoin/blink-wip/issues/1092).

## Problem

The PIN screen allowed 3 instant guesses (third → logout), and the attempt counter was persisted fire-and-forget — killing the app mid-write lost the increment. A brief-device-access attacker got 3 tries in ~3 seconds.

## Fix

Escalating wall-clock lockout, persisted in secure storage next to the existing counter: failure 1 → **30s**, failure 2 → **60s**, failure 3 → **logout** (unchanged — dropping the session is the strongest client-side response available, so it stays as the backstop).

- The lock survives kill/relaunch: loaded on mount and clamped via `clampLockedUntil` to *now + longest scheduled lockout*, so a wall clock rolled backward after the write (or a corrupt value) can never lock the user out indefinitely.
- While locked the keypad is disabled (guard in `addDigit` + `disabled` on the buttons) and the helper text shows a 1s-tick countdown (`PinScreen.tryAgainIn`, new key shipped AI-translated in all 28 locales).
- Both writes happen **before** the result is shown (awaited `Promise.all`), fixing the lost-increment window. The two writes aren't atomic, but either partial outcome is strictly better than today's total loss.
- The lock is cleared everywhere the attempt counter already was: successful PIN unlock, biometric success (`authentication-screen.tsx`), PIN disable (`security-screen.tsx`), logout (`use-logout.ts`), and PIN (re)creation.
- The `SetPin` flow never locks; the existing resume-lock guard already prevents backing out of a locked screen.

## Honest limits (by design, stated for the record)

- On a rooted device all client-side state is resettable. This raises the bar for the casual brief-access attacker; it is **not** hardware-backed attempt counting (that would be native Keystore work far out of proportion for a Low-severity finding).
- `Date.now()`-based: clock manipulation can shorten the lockout — never extend it, thanks to the clamp.

## Tests

- New pure spec `pin-lockout.spec.ts`: schedule, cap, negative input, remaining-time floor, and the clamp (including the it-must-still-expire case).
- `pin-screen.spec.tsx` extended (fake timers with real `setImmediate` so `flushEffects` settles): persists attempt+lock before showing the result; keypad inert while locked even for the correct PIN; re-enables after 30s and clears both keys on success; starts locked from a persisted future lock; absurd persisted lock still expires on schedule; third failure still logs out; set-pin flow never locks. The suite now calls `loadLocale("en")` — without it every i18n string renders as `""` and text assertions are vacuous (pre-existing gap in this spec).

## Review notes

- First iteration clamped the *remaining* time instead of the *stored timestamp* — that kept an absurd future lock pinned at 60s remaining forever. The spec that caught it ("clamps an absurd persisted lockout…") is in the suite; the clamp now happens once at load time.

---

## Round 2 — review fixes (supersedes parts of the above)

A review of the first iteration found ten issues, two of them full bypasses of the lockout this PR exists to add. Head is now [`6175d57`](https://github.com/blinkbitcoin/blink-mobile/commit/6175d57e8fc94c5836aef1f14e48082cc9ad26ed).

### The two bypasses, and the one cause behind them

The lockout was decided from React state. State starts empty, hydrates late, and is captured stale by callbacks.

1. **Relaunch.** `pinAttempts` and `lockedUntil` were read in an async mount effect. A guess entered before it resolved skipped an active lock *and* recorded itself as the first failure, writing `attempts=1` over a stored `2`. Kill, relaunch, guess, repeat, and the third-failure logout was never reached.
2. **Backspace.** The failure branch awaited two keystore writes before bumping the count or clearing the digits, and backspace was gated only by a `disabled` prop. Backspace plus a digit re-entered the handler on the same stale count, so two wrong guesses were recorded as one.

**Fix:** [`verifyPin()`](https://github.com/blinkbitcoin/blink-mobile/blob/6175d57e8fc94c5836aef1f14e48082cc9ad26ed/app/screens/authentication-screen/pin-verification.ts) re-reads the attempt count and lock from the keystore at submit time and returns a verdict. Hydration can no longer open a window, because nothing about the decision comes from render state. The existing [`useInFlightGuard`](https://github.com/blinkbitcoin/blink-mobile/blob/6175d57e8fc94c5836aef1f14e48082cc9ad26ed/app/hooks/use-in-flight-guard.ts) hook (already used by seven features) drops a concurrent second submission, which the re-read alone cannot do.

The screen is now three layers: pure schedule maths, an async decision module with no React in it, and a hook holding the guard and the countdown.

### Judgement calls, and what was rejected

- **An unrecordable failure logs the user out.** If the keystore write fails, the lockout exists only in memory, and memory dies with the process — so a disabled keypad would be bypassed by a force-quit. Ending the session is the only refusal that survives a relaunch. *Rejected:* keeping the user logged in with a dead keypad (bypassable), and reporting to Crashlytics only (leaves the bypass open). The success path is deliberately the opposite: a failed *clear* after a correct PIN still lets the user in, since they proved the PIN and a stale lock is bounded by the clamp anyway.
- **The two lockout keys are now private behind a paired API.** `getPinFailureState` / `setPinFailureState` / `clearPinFailureState`, with the single-key accessors `private`. The invariant "clear one, clear both" was hand-maintained at five call sites, and this PR's first round had to go and add the second line to every one of them — good evidence it does not hold by convention. A future single-key caller is now a compile error. `resetPinAttempts` (wrote `"0"`) and `removePinAttempts` (deleted the key) were merged, since the only reader maps both to `0`.
- **`secureStorage.ts` collapsed onto four private helpers.** 21 of 25 methods were the same try/op/catch block. `accessible` stays a **required** argument with no default, so no future method can silently downgrade a key's protection class — the same "forgot the second half" failure the paired API exists to kill. The 539-line existing spec asserts the exact key and accessibility mode per method and passed unchanged outside the PIN block, which is what makes touching the mnemonic methods safe. *Rejected:* leaving the boilerplate (the PIN family was being rewritten anyway, and the shared reader is what fixes the NaN bug below in one place).
- **`useCountdown` was not reused.** The repo has one, and the review suggested it, but it floors where this screen ceils. Driving the keypad from it would lift the lock up to a second early, and the label would read "try again in 0s" for a full second while the pad was still dead. Changing its rounding would alter invoice-expiry display for two unrelated consumers with ten tests pinning current behaviour. A 20-line `useLockoutCountdown` keeps millisecond precision for the decision and ticks every 250ms, so the pad re-enables just *after* the true expiry.
- **Copy shows both lines.** `tryAgainIn` drops its "Too many failed attempts." sentence and renders under the attempts line, so a single typo no longer reads as "too many" and the remaining count stays visible. One reworded key plus one new key (`lockoutUnavailable`), translated across all 28 locales. `sr.json`'s `tryAgainIn` was Latin script while the rest of its `PinScreen` block is Cyrillic; corrected.

### Also fixed

- The "1 attempt remaining" warning lived in component state and was lost across a relaunch, so a user could be logged out — PIN and session profiles wiped — with no warning. It is derived from the stored count now.
- Biometric success cleared the lock fire-and-forget while the PIN path awaited the same two writes; now awaited, matching its own documented rationale.
- `clampLockedUntil` applied to only one of the two readers and was never written back, so a far-future stored value re-imposed the full lockout on **every** launch. `remainingLockoutMs` caps too, and the repaired value is persisted once.
- `getPinAttemptsOrZero` returned `NaN` for a corrupt entry, and `NaN < MAX_PIN_ATTEMPTS - 1` is `false` — so a corrupt counter logged the user out on their *first* wrong PIN. The shared numeric reader guards it.
- `MAX_PIN_ATTEMPTS` is derived from the schedule length, so adding a tier cannot silently widen the attempt budget.
- The keypad stayed visibly enabled while a verification ran, silently dropping presses; it is disabled for that window too.

### Tests

New specs for the decision module, both hooks, and the storage pairing; `pin-screen.spec.tsx` now drives a real in-memory keystore so a test can unmount and re-render to model a force-quit. 160 tests pass across the affected suites.

Both bypass fixes were checked by mutation, not just by being green: reverting the verify-time re-read turns 10 tests red, and removing the keypad guard turns the in-flight test red. That last check mattered — the first two in-flight tests I wrote passed *with the guard removed*, because the hook's own guard already dropped the second submit. They were replaced with assertions on what is actually observable.


kngako (changes requested):
## Findings

### Finding 1 (blocking): Unreadable stored PIN silently burns the attempt budget

`verifyPin` (`pin-verification.ts:69-73`) calls `getPinOrEmptyString()`, whose `catch` swallows every keystore error and returns `""` (`secureStorage.ts` `read`). In the authenticate flow a PIN is always set, so `""` means "read failed" — yet the entry is scored as a wrong PIN, persisted, and escalated. A legitimate user whose Android Keystore throws transiently on three unlock attempts (e.g. keystore temporarily locked after a reboot race) will be force-logged-out and have their PIN wiped by `logout()`, without typing a single wrong digit — and unlike the write path, this path reports nothing, so support would see zero diagnostics.

**Proposal:** Make the read tri-state at the wrapper (e.g. `getPin(): Promise<string | null>` where `null` is unreadable, distinct from "no PIN") and add a `verifyPin` outcome such as `{ outcome: "unreadable" }` that neither unlocks nor consumes budget and calls `recordAppError` with `alwaysRecord: true`. The screen can fail closed (keep the keypad disabled, show a retry message) without logging the user out. Add a test where the PIN read rejects and the stored attempt count is asserted unchanged.

### Finding 2 (blocking): `setPinFailureState` is two independent writes; a partial failure contradicts the fail-closed guarantee

`setPinFailureState` (`secureStorage.ts:165-176`) writes `pinAttempts` and `pinLockedUntil` via `Promise.all` and only ANDs the results. If the attempts write lands but the lock write fails, the user is logged out with "Couldn't record the failed attempt" even though the attempt *was* recorded — the message and the premise of the `unrecorded` path are false. The worse half: if the lock write lands but the attempts write fails, the failure is silently lost, so after the lock expires the stored count is unchanged and a force-quitting attacker gets a free attempt cycle — precisely the bypass this branch exists to close, triggered by a single-key fault.

**Proposal:** Store the pair as one serialized value under a single key (e.g. `pinFailureState = '{"attempts":2,"lockedUntil":…}'`) so a write is atomic and the pairing problem disappears — the file's own comment already calls them "one logical value split across two keys". If two keys must stay (migration cost), write `attempts` first and reconcile on read: a lock with a lower-than-expected count should be treated as the count that produced it. Either way, add a `verifyPin`-level test that simulates one key failing and pins the resulting stored state, not just the boolean.

### Finding 3 (non-blocking): A forward clock jump lifts the lockout instantly

The backward-clock case is carefully clamped, but `remainingLockoutMs(lockedUntil, now)` trusts `Date.now()` in the other direction: a user who sets the device clock forward past `lockedUntil` ends the lock immediately, compressing the escalating delays to nothing. The 3-attempt budget still caps total guesses before the forced logout, so the impact is limited to pacing, not the budget — but the headlining "escalating time lockout" is defeated on any device where the attacker can reach clock settings.

**Proposal:** At minimum, document the accepted limitation in `pin-lockout.ts` (the file already documents the backward-clock reasoning; the asymmetry should be explicit and reviewable). If pacing matters to the threat model, persist the *remaining* duration alongside `lockedUntil` and treat `lockedUntil - now > remainingAtWrite` as tampering that keeps the lock, or back the countdown with a monotonic clock where the platform offers one.

### Finding 4 (non-blocking): `clearPinFailureState` swallows failed erases, leaving sticky destructive state

`clearPinFailureState` (`secureStorage.ts:180-185`) discards the boolean results of both `erase` calls; the comment "a key that was already absent is the cleared state" conflates *absent* with *failed to remove*. If the `pinAttempts` erase fails while the key is present, `attempts >= MAX_PIN_ATTEMPTS` persists: every later wrong entry is instantly `exhausted` (forced logout + PIN wipe) until a correct PIN is entered, and hydration can render "-1 attempts remaining" via `attemptsLeftAfter` (`use-pin-lockout.ts:38`) returning a negative count.

**Proposal:** Have `clearPinFailureState` report erase failures (`recordAppError`, non-blocking) instead of dropping them, and clamp defensively at read time: `attemptsLeftAfter` should floor at a sensible display value (or `readPinLockState` should normalize `attempts >= MAX_PIN_ATTEMPTS` with no active lock). A small test with a rejecting `remove` on one key pins the reporting.

## Test Coverage Gaps

- **`use-logout.ts` change is unpinned.** The swap from `removePinAttempts()` to `clearPinFailureState()` (`use-logout.ts:51`) has no test — there is no `use-logout` spec at all. If the clear were dropped, a stale lockout would survive logout; add an assertion that logout clears both lockout keys. Same for `security-screen.tsx:107` (`removePin` path).
- **Finding 1 and Finding 2 paths have no tests** (by construction — the current seams make the states unreachable). They should land with the fixes: PIN-read rejection leaves attempts unchanged; one-of-two write failure leaves a consistent stored state.
- **Biometric-unlock clearing is only order-tested.** `authentication-screen.spec.tsx` pins that `clearPinFailureState` precedes `setAppUnlocked`, but nothing pins what happens if the clear rejects (the await means a keystore fault now delays biometric unlock — untested, and `handleAuthenticationSuccess`'s caller doesn't catch).
- The new `verifyPin` / wrapper split is otherwise well covered: the schedule, relaunch bypass, re-entrancy, and in-flight dimming all have tests that would fail if the core line were reverted. No E2E/Detox coverage of the lockout exists; acceptable for this layer, but worth a note in the PR test plan.

## Code Smells

- **One logical value, two keys** (`secureStorage.ts:113-117`): the paired API with private single-key accessors is a workaround for non-atomicity; a single serialized key removes the class entirely (see Finding 2).
- **Inconsistent `readonly`** on the `PinVerification` union (`pin-verification.ts:16-19`): the `wrong` variant's fields are mutable while every sibling field is `readonly`.
- **Magic PIN length**: `4` is hardcoded in three places in `pin-screen.tsx` (`addDigit` guard, completion check, circle rendering) — pre-existing, but this PR rewrote `addDigit` and kept the literal.
- **`attemptsLeftAfter` can go negative** (`use-pin-lockout.ts:38`): fine for well-formed state, but it's the kind of derived-value assumption (stored count never exceeds the budget) that Finding 4's sticky state violates — clamp or comment the invariant.
- **Comment claims vs. behavior**: the `clearPinFailureState` doc comment ("there is no failure here a caller could act on") is only true for absent keys, not failed erases — a load-bearing comment that overstates the guarantee (§2.6 treats these as reviewable artifacts).

designsats (approved):
Works as intended

kngako (approved):
All previous findings fully addressed.

PR #4163, opened 2026-08-13
feat(backup): make the cloud-backup password mandatory

grimen:
Closes blinkbitcoin/blink-wip#1141 — item 3 of the red-team hardening epic [blink-wip#1092](https://github.com/blinkbitcoin/blink-wip/issues/1092).

## Problem

The cloud-backup screen offered encryption as an **opt-in** checkbox, default unchecked, and the validator passed with the box unchecked — one tap uploaded `{"encrypted": false, "mnemonic": "..."}` (the BIP-39 phrase in cleartext) to iCloud / Google Drive. A compromised cloud account then meant compromised funds.

## Fix

The password is now always required; the checkbox and every `isEncrypted` code path are gone. The crypto itself is unchanged (min 12 chars, PBKDF2-SHA256 600k → AES-128-GCM — the audit judged it sound).

- `buildBackupPayload` requires `password` at the type level **and** throws on an empty one (plain `Error`, not a new `BackupPayloadErrorReason` — the reason enum maps restore-side failures to user copy; an empty password at build time is a programmer error the validation layer already prevents).
- **Restore compatibility is the hard constraint and is untouched**: `PlainBackupPayload`, `isEncryptedBackup`, `parseBackupPayload`, `parseEncryptedBackupPayload`, and the whole restore flow keep accepting legacy plaintext backups already in users' clouds. Since the app can no longer produce that shape, the parse-side tests now pin it with hand-written plaintext fixtures instead of the builder.
- Users with an existing plaintext backup get an encrypted replacement the next time they run backup, through the existing overwrite-confirmation dialog — a strict upgrade, no special handling.

## Decisions & product notes

- **Product decision (confirmed):** require a password rather than the audit's alternative of a device-held key — a device-held key would make the backup useless in the lost-device case, which is the whole point of a cloud backup.
- The screen is a migration checkpoint (`MigrationCheckpoint.CloudBackup`) and stays completable; users unwilling to set a password still have the manual-backup path.
- The security-score signal (`use-security-signals.ts`) keys `cloudBackup` on completion method, not encryption — no change needed, and it stays truthful since every new cloud backup is encrypted by construction.

## i18n

`encryptCheckbox` removed from EN and all 28 locale JSONs (stale-key check); `description` reworded to say the backup is always password-protected, with updated translations in all 28 locales; regenerated `i18n-types.ts`/`source/en.json` committed.

## Tests

439 tests across the onboarding + payload suites pass. Highlights: "throws when built without a password" (`@ts-expect-error` + empty-string), "parses a legacy plaintext payload (restore compatibility)", metadata/`isEncryptedBackup`/`parseEncryptedBackupPayload` legacy cases on hand-written fixtures, upload assertions flipped to `\"encrypted\":true`, and the form/screen/validation specs reworked for the always-required password (net −190 lines).

## Review follow-ups (3bbc6bf9)

A code review of b655aec0 raised five points. Three are fixed in [3bbc6bf9](https://github.com/blinkbitcoin/blink-mobile/commit/3bbc6bf9fa72c10d11d5dc2df5d17c77d406bf27); two are deliberately not changed, with the reasoning below so a reviewer can disagree with it.

**Fixed — silent failure on encryption (the one that mattered).** `buildBackupPayload` was called with no `try`/`catch`, and `handleBackup` is wired straight into `GaloyPrimaryButton onPress`, so its promise is never awaited. Before this PR the default path built a plaintext payload and ran no crypto; now every user runs PBKDF2-600k + AES-GCM at that line. Any throw — a native crypto failure, or an empty password slipping past the disabled-button gate — became an unhandled rejection: no toast, no telemetry, and the user stranded after they had already signed in to Drive/iCloud and confirmed the overwrite dialog. It now reports through the `error-logging` boundary and shows the existing `uploadFailed` toast, matching every other failure branch in the hook.

**Fixed — the upload assertions did not test the security property.** They checked `expect.stringContaining('"encrypted":true')`, which a payload carrying both that flag and a stray plaintext `mnemonic` field would pass. The spec now runs the **real** `buildBackupPayload` against mocked crypto primitives instead of a hand-rolled stub, so the assertions see the shape that actually reaches the cloud, and asserts the uploaded payload contains neither the phrase nor a `"mnemonic"` field — on the fresh path and on the path that overwrites a legacy plaintext backup (where the phrase is present in the downloaded file and must not be carried forward). Both throw sources are covered too: encryption failure, and an empty password.

**Fixed — the disabled Continue gate was untested.** `mockIsValid` was reset to `true` in `beforeEach` and never flipped, so `disabled={!isValid}` had no coverage. With the checkbox gone that gate is the only thing keeping an invalid password out of the builder, so it is now tested directly (and is non-vacuous: the sibling test presses the same element and does reach `handleBackup`).

**Not changed — no error text on an untouched empty password field.** `shouldShowPasswordError` requires `hasPassword`, so an empty field shows no error, and `useCloudBackupForm` additionally resets `passwordTouched` whenever the password is empty, making the touched-and-empty state unreachable by design. Surfacing an error there would mean tearing out that reset. It is also unnecessary: the empty field is not unexplained — its placeholder reads *"(12 characters minimum)"* and the screen description states the backup is password-protected. Yelling "Minimum 12 characters" in red at someone who has not typed anything yet would be worse than the current state.

**Not changed — `preset="fixed"` on a screen that now always renders the password fields.** The concern is keyboard clipping on a small device. But pre-PR, with the checkbox ticked — the encrypted path this PR makes mandatory — the screen rendered the checkbox **plus** both password fields **plus** the warning banner. The tallest-content case therefore got one row *shorter* here, so this is not a regression introduced by this PR. `preset="fixed"` is also the convention across the whole self-custodial onboarding flow, and `presets.scroll.inner` has no `flexGrow: 1`, so switching would collapse the `justifyContent: "space-between"` layout unless a `flexGrow` override is passed too. Changing one screen's preset against the flow's convention is not worth doing without a device check; worth a separate look if the flow is ever revisited as a whole.

Tests: 443 pass across the 32 onboarding + payload suites (was 439). ESLint and `tsc --noEmit` clean.


kngako (changes requested):
## Findings

### Finding 1 — BLOCKING: `handleBackup` has no in-flight guard, and the button stays live through the entire pre-upload phase

`loading` returned by both provider hooks is only `true` inside `upload`/`downloadById` (`use-google-drive-backup.ts:201,225`; `use-icloud-backup.ts:81,97`). Everything before that — `startSession` sign-in, the existing-backup download prompt gap, the `confirmDialog` wait, and the now-mandatory synchronous `pbkdf2Sync` 600k-iteration derivation (`crypto.ts:27`, invoked at `use-cloud-backup.ts:127`) — runs with the Continue button enabled and showing no spinner. `handleBackup` itself has no concurrency guard.

**When it comes up:** a user taps Continue, the Google sign-in sheet completes, and while the overwrite-confirmation dialog is open (or during the brief frozen-UI crypto window on a low-end device) they tap Continue again. Two `handleBackup` runs interleave: two overwrite dialogs stack, and two uploads race on the same Drive/iCloud file. The content is identical and encrypted, so funds are never at risk — but the user sees duplicate dialogs and the branch's own premise ("don't strand or confuse the user in this handler") is only half delivered. The dialog window predates this branch, but commit `3bbc6bf9f` explicitly claims to harden this handler, and the mandatory crypto step widens the unguarded window.

**Proposal:** guard the whole handler with the existing `useInFlightGuard` pattern (the fix `docs/pr-review.md` prescribes for unguarded double-invocation), or add a local `busy` state in `useCloudBackup` set before `startSession` and cleared in a `finally`, and disable the button on `busy || loading`. Pin it with a test that a second `handleBackup()` invocation while the first is in flight is a no-op. Optionally also flip `loading` on around the `buildBackupPayload` call so the button spins during key derivation instead of the UI appearing frozen.

### Finding 2 — non-blocking: `reportError` drops the operation context for `Error` instances

`reportError("Cloud backup encryption", err)` (`use-cloud-backup.ts:134`) routes through `error-logging.ts:8-10`, which only uses the `operation` string when `err` is *not* an `Error`. Both throw sources named in the commit message — the builder's `new Error(...)` and any native crypto failure — are `Error` instances, so Crashlytics receives a bare `"buildBackupPayload requires a password…"` or vendor crypto message with no indication it came from the cloud-backup flow.

**When it comes up:** after release, a spike of native crypto failures hits the crash dashboard as an anonymous `Error: <native message>`; nobody can tell it is the new mandatory-encryption path without manually symbolicating and cross-referencing timestamps, and any alerting keyed on the operation name never fires. The observability the commit sets out to add is silently weaker than intended.

**Proposal:** in `error-logging.ts`, always wrap — `new Error(`${operation} failed: ${err.message}`, { cause: err })` — or pass the operation through `RecordAppErrorOptions` context instead of the message. Whichever shape is chosen, extend the hook spec's encryption-failure test to assert the reported error carries the "Cloud backup encryption" context, so the tripwire survives a refactor of the wrapper.

### Finding 3 — non-blocking: two divergent definitions of a valid backup password

The form gate requires ≥12 characters via `validatePassword` (`cloud-backup-validation.ts:24`), but the builder's last line of defense only rejects the empty string (`backup-payload.ts:83`). The type-level requirement enforces *presence*, not *policy*: any future caller — or a regression in the validation hook — that passes `"short"` gets a weakly-encrypted backup uploaded with no complaint.

**When it comes up:** a second backup surface is added later (e.g. an automated re-backup after restore, or a settings-screen "back up again" entry point) and its author reasonably assumes `buildBackupPayload` enforces the password policy, wiring it to a field with weaker or no validation. A 1-character password then protects a real mnemonic in someone's cloud, and nothing fails loudly.

**Proposal:** have `buildBackupPayload` delegate to the shared `validatePassword` (already imported by the validation module) and throw on `!validatePassword(password).valid`, so "valid backup password" is defined in exactly one place. Update the existing throw test to also cover a too-short password, which then pins the policy at the builder level rather than only at the screen. Restore is unaffected — the parse side never re-validates the password.

---

## Test Coverage Gaps

1. **No concurrency test for `handleBackup`** (ties to Finding 1). Nothing fails if a second invocation is allowed to interleave — the double-dialog/double-upload behavior could ship or regress unnoticed. A test that fires `handleBackup()` twice before the first resolves and asserts one session/one upload would lock it in.

2. **No test that the busy state covers the crypto window** (ties to Finding 1). If the fix wraps `buildBackupPayload` in the loading state, a test asserting the hook reports busy between confirmation and upload would prevent a future refactor from narrowing the window again.

3. **The restore-hook spec hand-mocks `parseBackupPayload`** (`use-cloud-restore.spec.ts:48`), so the PR's stated invariant — "restore compatibility is untouched" — is pinned only at the `backup-payload.spec` level. Per the house rule that mocks must not neutralize the assumption being relied on, one restore-level test feeding a legacy plaintext fixture through the *real* parser would close the loop between "the parser accepts it" and "a user can actually restore from it".

4. **No E2E coverage of the now-only backup path.** The Detox specs never touch the cloud-backup screen, so the mandatory-password flow ships without end-to-end confirmation on either platform. At minimum the PR's test plan should record manual QA on both iOS (iCloud) and Android (Drive), since the screen changed shape (checkbox removed, fields always visible).

## Code Smells

- **`uploadFailed` copy is misleading on the encryption path** (`use-cloud-backup.ts:135`): the toast says "Upload failed" when no upload was ever attempted. Users reporting the error — and support reading screenshots — will be pointed at their cloud account when the real failure was local. A distinct generic backup-failure message would cost one i18n key.

- **Duplicated mnemonic literal in the hook spec** (`use-cloud-backup.spec.ts`): `MNEMONIC = "youth indicate void"` restates the string hardcoded inside the `useWalletMnemonic` mock a few lines below. If one drifts, the `not.toContain(MNEMONIC)` assertions silently weaken. Reference one constant for both.

- **`handleBackup` is a 12-dependency orchestrator** (`use-cloud-backup.ts:154-167`): session, download, dialog copy, crypto, upload, analytics, and completion all live in one callback. Per the repo's "pure core, thin hook" vocabulary, the payload-build-and-toast decision is a candidate for a pure function unit-testable without `renderHook` — worth doing if this handler grows one more branch, not before.

- **The new block comment documents only half the hazard** (`use-cloud-backup.ts:122-124`): it explains the unawaited-`onPress` throw risk well, but the same unawaited handler is also what makes Finding 1's double-invocation possible. When Finding 1 is fixed, extend the comment so the next reader knows the guard and the try/catch protect the same boundary.

---

## Verdict

Finding 1 is blocking: the branch's stated goal is to make this handler failure-safe, and it still permits stacked overwrite dialogs and racing uploads from a plain double-tap — a one-pattern fix (`useInFlightGuard`) with a clear test shape. Findings 2 and 3 are non-blocking but cheap, and both already have natural homes for their tests. Everything else — the mandatory-password change itself, legacy restore compatibility, i18n regeneration, and the new failure-path tests — is in good shape and matches the commit prose.

esaugomez31 (approved):
@grimen @kngako

I applied and tested the review points. Everything works as expected.

## Findings

1. **BLOCKING: no in-flight guard on `handleBackup`** — Fixed. Wrapped in `useInFlightGuard` plus a busy state covering the whole handler, so the CTA is no longer live during sign-in, the overwrite dialog and the PBKDF2 window.
2. **`reportError` drops the operation context** — Fixed, in its own commit (`13fd2e574`) since `error-logging.ts` is shared by 136 call sites. Used `recordError(error, jsErrorName)` rather than re-wrapping, so the message and the original stack stay intact.
3. **Two divergent definitions of a valid password** — Fixed. `buildBackupPayload` delegates to `validatePassword`, so the policy lives in one place.

## Test Coverage Gaps

1. **Concurrency test** — Added.
2. **Busy state over the crypto window** — Added.
3. **Restore spec hand-mocks `parseBackupPayload`** — Removed the hand-mocks; the real parser runs now, plus a legacy plaintext restore test.
4. **E2E** — Not added.

## Code Smells

- **`uploadFailed` copy on the encryption path** — Fixed. New `backupFailed` key across all i18n files.
- **Duplicated mnemonic literal in the hook spec** — Fixed.
- **12-dependency orchestrator** — Left as is, per your own scoping ("not before").
- **Comment documents only half the hazard** — Extended.

Rebased onto `main`, which resolves the conflict. Tested on iOS / iCloud and the flow works end to end, with the uploaded backup encrypted and carrying no plaintext phrase. Android / Drive still needs a device pass.

esaugomez31 (dismissed):

esaugomez31 (approved):

PR #4164, opened 2026-08-14
feat(map): source merchants from BTC Map

### LINKED ISSUE #4071: feat: Source merchants from BTCMap

kngako:
Based on https://github.com/blinkbitcoin/blink-mobile/pull/4064 without any of the "add to map" functionality.

The map showed only Blink businesses from the `businessMapMarkers` query. It now shows the ~29k merchants on BTC Map (https://btcmap.org), the community-maintained OpenStreetMap overlay the wider bitcoin ecosystem already uses. Tapping a pin opens a bottom sheet with the same details btcmap.org's own drawer shows. We read that data; nothing here writes back to it.

The place list is held offline in a chunked AsyncStorage cache: a cold start seeds from the CDN snapshot (one gzipped ~550 KB response instead of six uncompressed API pages), and later starts ask the API for what changed, at most once an hour. The cached map is drawn the moment it is read rather than behind a spinner. Details are fetched per place on tap, since the API has no batch endpoint and the snapshot deliberately carries only coordinates and an icon.

Three API behaviours drove the sync design, and each is commented where it bites: `updated_since` is exclusive while the API orders by `(updated_at, id)` and pages on `updated_at` alone, so a page boundary inside a group of rows sharing a timestamp would drop the rest of that group — 11% of rows share theirs; the CDN's Last-Modified is the snapshot's generation time, minutes ahead of the newest record it contains, so it is backdated before being used as a cursor; and payment methods have no first-class field, arriving only through the `osm:` tag passthrough.

Pins reuse BTC Map's own teardrop and cluster discs so the map reads as the same map, with clustering by supercluster (pure JS, no native module). Marker glyphs are the Material Icons names BTC Map tags places with, which needs the font registered on both platforms — hence the Info.plist and build.gradle lines, and a native rebuild.

Two deliberate departures from btcmap.org:

- The open/closed badge is only drawn for places near the user. btcmap.org evaluates hours against the *viewer's* clock despite a source comment claiming otherwise, so a New Yorker sees a Tokyo cafe as closed while it is open. The raw hours string is always shown. The parser covers the syntax that actually occurs in the data and returns "unknown" for the rest rather than guess — sending someone to a shut shop is the one failure this feature can cause.
- Boost purchase, verify-location and comments are out of scope; those are BTC Map account actions.

The pay-this-business callout goes with `businessMapMarkers` — BTC Map places carry no Blink username, so there is nothing to route to.

Adds an ODbL attribution line: the data is OpenStreetMap-derived and the licence requires visible credit.

image

image


grimen (changes requested):
Reviewed head `0ae98d901a547dc79053cc42839b4a8ddeb90502`.

## Real strengths first

The new `app/btcmap/` boundary is clear, wire data is mapped into small domain types, and the pure helpers for opening hours, geo, verification, icons, storage, and detail fetching are thoughtfully documented and well tested. The targeted suite is green: 11 suites / 89 tests, with the BTC Map data layer at 96.71% statement and 91.52% branch coverage. TypeScript and targeted ESLint also pass.

## Request changes

The following findings are blocking because they include an Android rendering regression risk, no operational rollback path, permanent data loss at a pagination boundary, stale data for long-lived map mounts, and bypass of the repository's shared web opener.

### I1 — Let cluster markers rasterize before freezing — `app/components/map-component/cluster-marker.tsx:45`

On Android, `tracksViewChanges={false}` from the first frame can snapshot the custom SVG/Text before it paints, producing blank discs or missing counts. `PlaceMarker` already documents and handles this react-native-maps failure mode. Apply the same settle-then-freeze behavior here, re-enable it when the drawn count/tier changes, and add a fake-timer component test that pins the transition.

### I2 — Gate the BTC Map rollout behind Remote Config — `app/btcmap/config.ts:5-6`

These hard-coded third-party endpoints replace the map's complete data source, but there is no way to disable the feed if its shape or merchant/payment data becomes unsafe after release. Register a `btcMapPlacesEnabled` key with the repository's safe default, keep the hook inert when disabled, and add a test for the off state.

### I3 — Preserve rows in an overfull timestamp group — `app/btcmap/api.ts:211`

If more than `BTCMAP_PAGE_SIZE` rows share one `updated_at`, each rewound request returns the same first 5,000 IDs; this branch stops and persists the old cursor, so later rows at that timestamp are never applied. The current test only asserts that the loop stops, which pins the data-loss behavior rather than losslessness. Trigger a full snapshot reseed (or another provably lossless fallback) and test a timestamp group larger than one page.

### I4 — Resync when the mounted map cache expires — `app/btcmap/use-places.ts:119`

This effect reruns only after manual retry. Because the tab stays mounted across tab switches and backgrounding, returning days later never repeats the one-hour age check. Trigger loading on screen focus/app resume or schedule the next stale check, with fake-timer/AppState coverage showing that a fresh mount eventually becomes stale and synchronizes.

### I5 — Use the shared opener for web destinations — `app/components/map-component/place-sheet.tsx:111-112`

Merchant websites, social profiles, required-app web links, and the BTC Map profile currently go straight to the system browser. Route HTTP(S) through `openExternalUrl` for the established in-app-browser-with-fallback behavior, while retaining `Linking` for `tel:`, maps, mail, and payment schemes. Update the tests: they currently explicitly assert the undesirable raw `Linking.openURL` behavior.

## Test coverage

Coverage is substantial but not complete: targeted runtime coverage is **78.34% statements / 74.46% branches / 68.46% functions / 77.69% lines**. The new cluster marker, place marker, and main Map component currently measure 0%, and the suite passes while all of the timestamp, focus/resume, shared-opener, and marker-rasterization defects above remain.

Suggested landing order: lossless sync fallback and Remote Config kill switch first, then mounted refresh and Android marker settling, then shared link handling. Each fix should have a regression test that fails if its core line is reverted.

designsats (changes requested):
Android testing: 
https://github.com/user-attachments/assets/cb8c9ea9-d521-44a7-bcf9-b22ee664a4a9

**Review of the BTC Map implementation:**

1. The map doesn't respect the app theme (dark/light).
2. The "Navigate" CTA label should use the `background` color — currently it's white in dark mode, which is wrong.
3. We don't use shadows under buttons. The "center on user location" button incorrectly has one.
4. The "Select all" text button and the links in the merchant details should be bold.
5. Links in the merchant details should also be underlined.
6. The partially open merchant sheet should fully reveal the "Verified" info — it needs 20px more on my device (we need a universal offset so it behaves this way on every screen size).
7. The companion-app info isn't using our standard info box component. Don't mimic it — take it from Storybooks.
8. Verify that the merchant details content uses the `paragraph3` style (14px); it looks too small in testing.
9. Remove the merchant icon, name only.
10. Show socials like any other link, last in the merchant info list.
11. Icons should match the link color (`primary`).
12. Labels collide in high-density places "Berlín, SV" 
[Screenshot](https://github.com/user-attachments/assets/5182d68f-7c62-4602-9a62-c362bdce8702) 

General note: Make sure components are never reinvented, always taken as thy are in existing flows. 

## Screenshots

| Before | After |
| ------ | ----- |
|    Image    |    map-moneybadger   |




Appending Claude review of code: 

## Design-system pass

Checked the UI in this PR against the Blink design system (tokens in `app/rne-theme/colors.ts`, the component specs, and what's already in `app/components/atomic/`). This is a design review, not a code review — **no changes requested yet**, and nothing here blocks the data work, which I'm not commenting on.

First, credit where it's due: the accessibility fundamentals are right. 44px hit targets on the contact rows, chips and the location button; `hitSlop={12}` on the 22px share icon; 48px on Close. Skeletons rather than spinners, and the sheet's peek height measured rather than guessed. The sheet shell itself — `borderTopRadius: 20`, `colors.white` fill, 1px `grey4` top border, 20px horizontal padding — matches our Figma `bottom-sheet` component exactly, including the radius. And the open/closed badge correctly uses `_green` / `error`.

### Two things are approved as-is — please don't change them

I ruled on both today and logged them as documented exceptions, so they don't get re-raised in future reviews:

- **`PIN_COLOR_DARK = "#8D9EDD"`** stays. It's an invented colour and a fifth blue in an app that already has four, but `primary` lightens to amber in dark mode and collides with the boosted pin. No existing token solves that.
- **Material Icons alongside Phosphor** stays, scoped to the map screen. BTC Map's glyph vocabulary is server-driven and open-ended across ~29k places; `GaloyIcon` exposes ~155 Phosphor glyphs, so a mapping would collapse a large share of pins to a generic marker and cost the map its at-a-glance legibility. Worth noting the knock-on: this makes vendoring Phosphor to drop the npm dep harder, since we now carry a second icon font.

### The one that matters: the Navigate button's label

`place-sheet.tsx` sets `navigateText: { color: colors._white }`. `_white` is `#FFFFFF` in **both** themes — it opts out of the inversion `white` gives you:

| Theme | Fill | Label | Ratio | AA @ 17px/700 (needs 4.5:1) |
|---|---|---|---|---|
| light | `#fc5805` | `#FFFFFF` | 3.2:1 | fail |
| dark | `#ffad0d` | `#FFFFFF` | **1.9:1** | fail badly |

Dark mode is the urgent half. Using `colors.white` instead (as `GaloyPrimaryButton` does) makes it black-on-amber at 11.2:1. Light mode stays at 3.2:1 — but that failure is app-wide and pre-existing on every Blink CTA, caused by the inverted token, and fixing it belongs in `galoy-primary-button.tsx` rather than here. So the ask is just: don't be worse than the baseline.

### Two more colour swaps

- **`colors.success` in `cluster-marker.tsx`** → should be `colors._green`. We never define `success`, so it falls through to the @rn-vui library default (`#52c41a` light / `#439946` dark). Every cluster disc on the map is currently painted a library green rather than Blink's `#00A700`. (Two existing files already do this, so it's not new — but this puts it on the map.)
- **`PIN_GLYPH_COLOR = "#FFFFFF"`** → `colors._white`. Same intent, and it's the token that exists for exactly this.

### The change that makes the diff smaller

Every control in the sheet is hand-rolled, and each has a shipped equivalent:

| In the PR | Use |
|---|---|
| `navigate` pill | `GaloyPrimaryButton` |
| `close` text button | `GaloySecondaryButton` |
| "requires app" card | `GaloyInfo` — its Figma intent is literally *"not active info box"* |

That's roughly **−45 lines net**: the PR gets smaller and more on-system at the same time, and it resolves the dark-mode contrast issue, most of the scale drift below, and the `grey5`/`grey7` point for free. If you only take one thing from this comment, take this one.

### Smaller stuff, fine as follow-up

- **Scale drift.** Our radii are `8 / 16 / 999` and spacing is `3 5 8 10 14 20 30`. The PR uses radii `28` (navigate), `12` (infoCard), `10` (avatar, badge), `16`+border (chip), `20` (statusPill), and gaps of `12`, `6`, `4`, `16`, `24`. Type sizes `17`, `13`, `11` aren't on the product scale (`24 20 18 16 14 12`). The pull tab is `40×3` where Figma specs `26×3`.
- **`grey5` vs `grey7`.** `grey5` means "this responds to touch"; `grey7` is static. The status badges and the info card use `grey5` but aren't tappable.
- **Type scale.** Every `Text` sets a raw `fontSize`/`fontWeight`. The convention is the theme's `type="p1"|"p2"|…` with `bold` (236 uses across `app/`); since the theme pins `SourceSansPro-Regular` and varies weight only through those props, hand-set `fontWeight: "600"` gets synthetically bolded on Android instead of loading the real face.

### One question

Is **`PIN_COLOR_BOOSTED = "#F7931A"`** deliberate — bitcoin orange as a cross-app signal, matching btcmap.org — or did it come along with the pin work? If it's deliberate I'll log it as an exception like the two above. If not, `_primary1` `#FFBE0B` is the palette value.

grimen (approved):

PR #4167, opened 2026-08-14
refactor(security): gate seed-phrase screens on the screen guard for fail closed

kngako:
refactor related to fixes applied in https://github.com/blinkbitcoin/blink-wip/issues/1065

Screen security was fire-and-forget: useScreenSecurity kicked off registration in a passive effect and every protected screen painted its mnemonic or API secret immediately. Before initSettings/register resolved, during the 10-second retry interval after a rejected attempt, and forever after the bounded retries were exhausted, the sensitive UI stayed visible with screenshots and screen recording fully enabled — and a later successful retry cannot undo frames captured in that window. This makes protection a fail-closed gate instead of a side effect.

Manager (app/utils/screen-security.ts): the enable/disable pair becomes an idempotent lease API. acquire(backgroundColor) registers one consumer and returns a lease: a ready promise plus a release method. ready resolves only once ScreenGuard.register has succeeded, including the bounded retries (which now run inside the registration cycle rather than as detached timers), and rejects only once those retries are exhausted; the retries report each failure, the lease holder reports the final one. Concurrent leases share the one in-flight cycle through a registering guard plus a shared waiter list, so a screen pushed while another is registering joins the same attempt and sees the same outcome. release() is idempotent, so a retried effect cleanup cannot deflate the count; releasing the last lease wakes and cancels a pending retry — a retry nobody waits for must not register a guard nobody will unregister — and enqueues the unregister.

Two subtleties are handled where they bite, both pinned by specs:

- Register/unregister decisions are made inside the serialized queue at run time, not at call time. A lease acquired while a teardown is in flight (the native unregister already awaited, `registered` still reading true) re-registers behind the teardown instead of trusting a flag that is about to go stale; a lease that arrives earlier, before the queued teardown runs, simply cancels it — the guard stays on and no churn occurs either way.
- A lease released before its cycle ran makes no native calls at all: the cycle checks the lease count at run time and abandons quietly, and its waiter is removed at release so the never-settling ready promise cannot leak a late settlement into an unmounted screen.

The rejected-unregister recovery from the previous commit is kept: the teardown resets `registered` in a finally, so a rejected unregister cannot wedge the guard, and the rejection still propagates for reporting.

Hook (app/hooks/use-screen-security.ts): owns one lease for the mount lifetime and exposes "activating" | "active" | "failed" instead of returning nothing. Late settlement after unmount is ignored, failures are reported, and the lease is released exactly once.

Gate (app/components/screen-security-gate/): ScreenSecurityGate mounts its children only in the active state. While activating it renders a blank themed Screen — deliberately neutral, since nothing on a protected route is safe to show yet; on failure it renders a non-sensitive error state (errors.generic + Try Again/Back, all existing i18n keys, so no locale regeneration). Try Again remounts the hook subtree, which releases the old lease and acquires a fresh one; Back leaves via navigation.goBack(). Because the whole subtree is unmounted until active, the screens' setOptions effects that install the Copy/Share/Paste header actions never run while gated — header actions are covered, not just the visible fields.

Routes: all five protected screens are wrapped — backup-phrase, backup-phrase-confirm, view-backup-phrase, restore-phrase, and the API key secret reveal. One intended behavior change: on the view-backup screen the biometric prompt now fires only after the guard is active (protect first, then prompt), instead of racing registration.

Coverage: the manager spec is rewritten around leases — deferred-register sharing between concurrent leases, ready pending before and resolving after a retry succeeds, rejection only after exhaustion with per-retry reporting, retry cancellation on last release, the mid-teardown re-registration race, the cancelled-queued-teardown no-churn window, release idempotency, and the kept overlap/replacement/rejected-unregister cases; the suite uses resetAllMocks after an unconsumed mockReturnValueOnce was observed leaking into a later test and hanging it. The hook spec covers the state transitions, late-settlement-after-unmount, and exactly-once release. The gate spec covers the placeholder and error states and asserts a sensitive subtree's setOptions effect does not run while gated. Screen-level regression tests drive a deferred lease and assert the seed words / API secret and their header actions are absent while registration is pending, appear only once it resolves, and remain absent after rejection — and the Try Again/Back path is exercised on the failure state.

Verified: full jest suite green (596 suites, 6932 tests), tsc:check and eslint clean.


grimen (changes requested):
## Findings

**1. The one-time-secret dismissal guard no longer runs at mount (blocking).**
`ApiKeySecretRevealContent` installs the `beforeRemove` listener and sets `headerBackVisible: false` / `gestureEnabled: false`, but the gate no longer mounts that content until protection is active. While registration is pending or failed, nothing prevents dismissal.

*Example:* The user creates an API key, the reveal screen mounts, registration is still in flight, and a back swipe pops the screen — `ApiKeyCreateScreen` unmounts with its `secret` state and the server-once secret is gone without warning.

*Proposal:* One option could be hoisting the dismissal guard above the gate, or having the gate itself block removal whenever the gated content holds unrecoverable state.

**2. The failure state's Back button can silently discard the secret (blocking).**
`app/components/screen-security-gate/index.tsx:63` calls `navigation.goBack()` unconditionally from the error view, regardless of what the gated screen was holding.

*Example:* Registration fails on the API-key reveal screen; the user taps Back on the error view, the create screen pops, and the freshly created secret is discarded with no chance to copy it.

*Proposal:* For screens whose content is unrecoverable, one option could be suppressing or confirming Back and leaving only Try Again.

**3. The activating state paints an empty screen for up to ~30 seconds.**
The `activating` branch renders `{null}`, while the acquire cycle runs one initial attempt plus three retries at 10 s each.

*Example:* On a build where the native module didn't link, `ScreenGuard.register` rejects immediately, so tapping "View backup phrase" shows a blank screen with no spinner or text for roughly 30 seconds before the error appears.

*Proposal:* Rendering an `ActivityIndicator` the way `ViewBackupPhraseContent` already does for its own pending state would cover this — a spinner isn't sensitive content.

**4. Fail-closed also blocks wallet restore.**
The fail-closed policy is applied uniformly, including to `RestorePhraseScreen`, which takes a phrase the user already holds rather than displaying one.

*Example:* A user reinstalling on a device where registration never succeeds cannot reach the seed-word inputs at all, since Try Again re-runs the same failing cycle — leaving no path back to the funds.

*Proposal:* This is for the implementer to decide — if hard-blocking restore is intended, a note in the PR body would help; otherwise degrading to a warning on input-only screens could be an option.

## Test Coverage Gaps

- `__tests__/components/screen-security-gate.spec.tsx` mocks `useScreenSecurity` and flips a module-level `securityState`, so "re-acquires protection when retry is pressed" only asserts that the mock's new value renders — the release/acquire pair the `key` remount is supposed to trigger is never exercised.
- No test covers dismissal (back swipe / hardware back) while the gate is in `activating` or `failed` on the API-key reveal screen.
- No test covers the failure-state Back button popping a screen that holds unrecoverable state.

grimen (changes requested):
## Findings

**1. A same-tick lease replace silently drops the new background color (blocking).**
`startRegistration` only calls `registerWithRetries` when `!registered`, and a released-then-immediately-reacquired lease (the "replace" no-churn path) never resets `registered` to false.

*Example:* A protected screen registers with the light-mode background, then the user flips to dark mode — `useScreenSecurity`'s effect releases the old lease and acquires a new one with the dark color in the same tick. The queued teardown sees the arriving lease and cancels itself, so `registered` stays `true` and the new registration task skips `registerWithRetries` entirely — the native guard never receives the new color. The only "no churn" test (`__tests__/utils/screen-security.spec.ts`) passes `"#000000"` for both the leaving and arriving lease, so it never catches this.

*Proposal:* One option could be tracking the color the guard was actually registered with and re-registering when it changes, even on the no-churn path.

**2. The malformed-params redirect no longer runs when the guard keeps failing (blocking).**
The missing/invalid-`challenges` auto-redirect effect now lives inside `BackupPhraseConfirmContent`, which `ScreenSecurityGate` only mounts once the guard is active — and this screen passes no `failOpen`.

*Example:* A deep link or rehydrated navigation state delivers this screen with no valid challenges on a device where the guard keeps failing to register (e.g. native module issue). Previously the redirect ran unconditionally and self-healed to backup step 1; now the user is stuck on the generic Retry/Back error view, and Back exits the flow entirely with no way back to the redirect.

*Proposal:* If this is intentional, a comment would help; otherwise, one option could be running the malformed-params check ahead of the gate, before the sensitive content ever needs to mount.

**3. Registration failures are now reported up to ~30s late (blocking).**
`reportError("Enable screen security", ...)` fires only from `lease.ready`'s rejection handler, which per `registerWithRetries` only rejects after `ENABLE_RETRY_LIMIT` (3) retries at `ENABLE_RETRY_DELAY_MS` (10s) each are exhausted — versus reporting on the very first native rejection as before.

*Example:* A device starts failing screen-guard registration; monitoring gets no signal for up to 30 seconds while the user already sits on the fail-closed spinner, delaying incident detection.

*Proposal:* One option could be reporting on the first failed attempt too (in addition to the final one), so monitoring sees the failure onset rather than only the exhaustion.

**4. The two new i18n keys are only in English.**
`discardSecretTitle`/`discardSecretBody` were added to `en/index.ts`, `i18n-types.ts`, and `raw-i18n/source/en.json` only — no other locale file is touched in this diff.

*Example:* A non-English user who hits a registration failure on the API-key-secret-reveal screen and presses Back sees this confirmation Alert render in raw English.

*Proposal:* These should go through the usual AI-translation pass across all locales before merge.

esaugomez31 (dismissed):
Approved.

Reviewing the gate turned up two defects that shared one root cause, and both are fixed on this branch.

**The root cause.** `useScreenSecurity` re-runs on `colors.black`, which is `#000000` in light and `#FFFFFF` in dark, and it returns to `"activating"` while the guard re-registers with the new mask colour. The gate answers `"activating"` by unmounting its subtree — so an ordinary theme flip, including an automatic sunset switch, tears the protected content down mid-use.

**1. Header actions outlived the unmount that was meant to hide them** — fixed in 5b902a357.

`navigation.setOptions` has no owner: the route keeps the last value it was handed, and unmounting the component that set it does not revert it. This branch already depended on that behaviour, since the restore screen clears its own Paste button by hand on the step 1 to step 2 transition.

So when the guard dropped and the gate swapped the phrase for its spinner, Copy stayed in the header, still closed over `handleCopy`, one tap from putting all twelve words on the clipboard while the guard was off — and it stayed live on the terminal failure state too, alongside the fail-closed error view. Each of the three effects now reverts to `noHeaderRight` on cleanup, so an action's lifetime matches the content it belongs to.

**2. The unmount destroyed everything the user had typed** — fixed in f06a9ae9a.

Both screens where the user types kept their answers inside the gated subtree, so the unmount wiped every word already entered: a user restoring a funded wallet lost all of it to a theme change, with no warning and nothing to recover.

The state moves above the gate and is handed down — `restore-phrase` hoists `useRestorePhrase` into the screen, and `backup-phrase-confirm` gains a `GatedConfirm` layer between the redirect guard and the gate. The gate can now hide the words without destroying them.

This is safe above the gate because neither hook touches anything on mount: `useRestorePhrase` has no effects at all (its clipboard read and its restore both hang off presses), and `useBackupConfirm`'s only effect is the all-correct auto-advance, which cannot fire while the inputs are empty. Both were checked before moving them, and `useMigrationCheckpoint` and `useCompleteBackup`, which came up with the confirm layer, are `useCallback` only.

Passing the validated challenges down also leaves one place that judges the route params, so the two copies can no longer drift, and the `EMPTY_CHALLENGES` fallback that existed for the second copy is gone.

**Verification.** `tsc` and `eslint` clean; 33 suites / 518 tests green across the onboarding screens, 92 suites / 1375 with the screen-security, API-key and migration suites included. The four screens touched measure 100% statements, branches, functions and lines. Each regression test was confirmed to fail against the unfixed screen before the fix was applied, and each commit typechecks and passes on its own.

esaugomez31 (changes requested):
### Blockers

1. `app/utils/screen-security.ts:136` - the color-change branch wraps `unregister()` in a `try/finally` with no `catch`, so a rejection propagates past the `registerWithRetries` call on line 145. `unregister()` does reject in practice (the library's `_logError` returns `Promise.reject`), so a theme flip can end with the guard torn down, `registered = false`, no re-registration attempted, and every gated screen on the failure view.

2. `app/utils/screen-security.ts:116` - each acquire enqueues its own retry cycle, and `settleWaiters` drains a shared array. If cycle 1 exhausts its attempts it rejects every waiter, including leases that arrived after it started; cycle 2 then runs a fresh cycle and resolves into an empty array. Net result: the guard is on but both screens sit on the failure view. Try Again has a related problem, since `release()` decrements the count and wakes the sleeping cycle, but the remount re-acquires before the woken cycle resumes, so it continues with its partially spent attempt budget and rejects the new lease early.

3. `app/screens/settings-screen/api/api-key-secret-reveal.tsx:148` - the gate is used without `failOpen`. Combined with `headerBackVisible: false`, `gestureEnabled: false` and the `beforeRemove` preventDefault, a device where registration keeps failing leaves only Try Again or discard. The secret never mounts, so there is no copy or share path, and the key already exists server side by then.

4. `app/screens/self-custodial/onboarding/manual-backup/view-backup-phrase-screen.tsx:40` - `useLocalAuthGate` runs inside the gated subtree. `useScreenSecurity` re-runs on `colors.black`, so a theme change unmounts the content and re-prompts for biometrics or PIN. If a PIN challenge is already pending when that happens, the cleanup sets `mounted = false` and `onChallengeSuccess` is swallowed: the correct PIN does nothing and a second prompt gets pushed.

### Non-blockers

5. `app/components/screen-security-gate/index.tsx:64` - nothing bounds the `activating` state, and `failOpen` is only checked after that branch returns, so it cannot apply while activating. I could not find a concrete way to make the native call hang, so this is hardening rather than a live bug, but a timeout seems worth having.

6. `app/utils/screen-security.ts:122` - `desiredColor` is never reset, while `registeredColor` is cleared in both teardown paths, so `if (color === undefined) return` is unreachable. It is also the only path that returns without settling waiters, so clearing `desiredColor` later would leave every lease hanging. Either drop the branch or make it settle its waiters.

7. `app/hooks/use-screen-security.ts:34` - `registerWithRetries` already reports each failed attempt, the first one under the same "Enable screen security" string used here. One exhausted registration with two protected screens mounted emits six reports for a single event.

8. `app/components/screen-security-gate/index.tsx:86` - the failure view shows `errors.generic()`, which names no cause and implies retrying later will help, though on an unsupported device it never will. The PR already adds `ApiScreen.discardSecretTitle` and `discardSecretBody` across all 28 locale files, so a dedicated string would not add locale work.

9. `app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx:56` - `useBackupPhrase` holds `expiresAt` in a `useRef` allocated on mount, so a gate re-activation restarts the 10 second countdown and greys out Continue again for a user who already waited it out.

10. `app/screens/self-custodial/onboarding/manual-backup/backup-phrase-screen.tsx:46` - the step param is derived in the wrapper and again in the content. The comment explains why the report lives above the gate and that reasoning is fine, but passing the validated step down as a prop, the way `backup-phrase-confirm-screen.tsx:234` does, would keep one source of truth.

11. `app/components/screen-security-gate/index.tsx:95` - `onPress={onBack ?? (() => navigation.goBack())}` builds the handler inline in the prop and allocates a new arrow on every render. Worth pulling out into a named const.

12. `app/screens/self-custodial/onboarding/restore/restore-phrase-screen.tsx:254` - the comment states that `useRestorePhrase` has no mount effects, but it calls `useBip39Input`, which registers keyboard listeners in a `useEffect`. The conclusion still holds, only the stated reason is wrong.

Renny8698 (approved):

grimen (changes requested):
## Findings

**1. A theme flip during registration resolves the new lease on the stale colour, then drops the guard under mounted content (blocking).**
`settleWaiters` in `app/utils/screen-security.ts:151` resolves every waiting lease when any registration task succeeds, regardless of which colour that task registered with. If a theme flip releases a light lease and acquires a dark one while the light registration is still in flight or sleeping between retries, the light cycle's success resolves the dark lease's `ready` and the gate mounts the content. The dark lease's own queued task then sees `registeredColor !== color` and calls `ScreenGuard.unregister()` under that mounted content. If the re-register then exhausts its retries, `waiters` is already empty, so nothing is rejected and `useScreenSecurity` stays `"active"` with the guard off.

*Example:* The first registration on "View backup phrase" fails, and automatic dark mode kicks in during the 10 s retry sleep. The retry lands with the old colour, the words render, and the guard is then unregistered to apply the new one. If that re-register keeps failing, the words stay on screen unprotected. A spec against `b33c73bf9` shows the dark lease resolving, then `unregister`, then four rejected `register` calls, with the lease never rejected.

*Proposal:* One option could be settling only the waiters whose colour matches the one just registered, leaving the rest pending for their own task.

**2. A rejected unregister in the colour-change branch leaves the guard down with no re-registration (blocking).**
The `try/finally` around `ScreenGuard.unregister()` at `app/utils/screen-security.ts:135` has no `catch`, so a rejection skips `registerWithRetries` on line 145 and falls through to rejecting every waiter. The library's `unregister` rejects through `_logError`, so this is reachable.

*Example:* A theme flip on a protected screen hits a rejected `unregister()`. The guard ends up torn down with `registered = false`, no register call is ever made, and every gated screen lands on the failure view.

*Proposal:* One option could be catching the unregister rejection, reporting it, and proceeding to `registerWithRetries`, the same way the teardown path treats `false` as the recoverable state.

**3. An exhausted cycle rejects leases whose own cycle later succeeds (blocking).**
Each acquire enqueues its own task, but `settleWaiters` drains the shared `waiters` array. When the first cycle exhausts its retries it rejects every waiter, including a lease that joined mid-cycle. That lease's own queued task then registers successfully and resolves into an empty array.

*Example:* The backup phrase screen's registration is failing, and the user pushes the confirm screen before the retries run out. Both screens get rejected, the second task then lands the guard, and both still sit on the failure view with the guard on.

*Proposal:* One option could be having a waiter settled only by the task its acquire enqueued, or re-queuing the waiters that arrived after the failing cycle started.

## Test Coverage Gaps

- `"applies a color change that arrives while a registration is in flight"` awaits `arriving.ready` before the new-colour register has run and only asserts the final register call, so it passes with finding 1 in place.
- `__tests__/hooks/use-screen-security.spec.ts` and `__tests__/components/screen-security-gate.spec.tsx` mock `acquireScreenSecurity`, so no spec drives a theme flip through the real manager.
- No spec covers a rejected `unregister()` in the colour-change branch, a theme flip during a retry sleep, or a lease joining a cycle that then exhausts.

PR #4173, opened 2026-08-15
fix(self-custody): patch uniffi-bindgen-react-native RustBuffer for 32-bit Android

### LINKED ISSUE #4172: bug: Self custodial mode causes 32bit Android devices to crash.
**Describe the bug**
The app crashes with a Rust panic null RustBuffer had non-zero capacity inside the Breez Spark SDK's UniFFI bridge — but only in 32-bit processes. The C++/JSI bridge layer and the Rust  
   library disagree on the binary layout of the RustBuffer FFI struct on 32-bit ABIs (x86, armeabi-v7a). It's not really about Android 10 — it's that Android 10 devices like Tecno's run a  
   32-bit userspace (and your repro used a 32-bit x86 Android 10 emulator). On 64-bit devices the bug is latent because the layouts accidentally match. 

**To Reproduce**

Use an Android 32 bit device or emulator to load a new self custodial wallet.

**Expected behavior**

It should work as expected

Tecno Pop 4 (Android 10, 32 Bit)

**Additional context**

kngako:
Fixes https://github.com/blinkbitcoin/blink-mobile/issues/4172, a deterministic SIGABRT crash when opening the self-custody wallet (Breez Spark SDK) on 32-bit Android processes (armeabi-v7a, x86), seen in the wild on Android 10 devices that ship a 32-bit userspace (e.g. Tecno Spark/Pop/Camon lines) and reproduced on a 32-bit x86 Android 10 emulator. Crash evidence: docs/tecno-crash/tombstone_00..05 (all six identical).

Symptom
-------
signal 6 (SIGABRT) on the React Native JS thread (mqt_v_js) with abort message "null RustBuffer had non-zero capacity" — a Rust panic with panic=abort inside libbreez_sdk_spark_bindings.so (shipped by @breeztech/breez-sdk-spark-react-native@0.22.0). The symbolized backtrace shows the panic firing in std::panicking with the oldest application frame inside bip39::Mnemonic::to_seed_normalized, i.e. it detonates on the very first self-custody call (app/self-custodial/bridge/wallet.ts:31, defaultExternalSigners deriving key material from the wallet mnemonic).

Root cause
----------
The panic is an assertion in uniffi-core 0.29's
RustBuffer::destroy_into_vec():

    if self.data.is_null() {
        assert!(self.capacity == 0, "null RustBuffer had non-zero capacity");

It fires because the C++/JSI bridge and the Rust library disagree on the binary layout of the RustBuffer FFI struct on 32-bit ABIs:

- Rust side (uniffi 0.29, confirmed by disassembling RustBuffer::from_vec at 0xb4d6ec in the shipped x86 .so): #[repr(C)] struct RustBuffer { capacity: u64, len: u64, data: *mut u8 } -> 20 bytes on 32-bit, fields at offsets 0 / 8 / 16
- C++ side (uniffi-bindgen-react-native@0.29.3-1, cpp/includes/RustBuffer.h): struct RustBuffer { size_t capacity; size_t len; uint8_t *data; } -> 12 bytes on 32-bit (size_t is 4 bytes), fields at offsets 0 / 4 / 8

On 64-bit targets size_t == uint64_t so the layouts accidentally match and the bug is latent — which is why only 32-bit devices crash. On 32-bit, a by-value RustBuffer handed to Rust is misread: Rust's u64 capacity spans C++'s cap+len (non-zero garbage) while Rust's data pointer is read from memory C++ never wrote (often NULL) -> data == NULL + capacity != 0 -> assertion -> abort. The by-value return path is equally broken, so every SDK call moving a byte buffer across the boundary is poisoned.

Same class of bug as mozilla/uniffi-rs#2624. Upstream fixed the header in uniffi-bindgen-react-native >= 0.30.0-1 (fields are uint64_t there), but @breeztech/breez-sdk-spark-react-native@0.22.0 (latest published) still depends on ^0.29.3-1.

Fix
---
patch-package patch changing the C++ struct fields from size_t to uint64_t so both sides of the FFI agree on the layout. The prebuilt Rust .so already uses the u64 layout and the bridge C++ is compiled during the app build, so this aligns the two sides without touching the prebuilt library. Applied automatically by the existing postinstall (yarn patch-package --error-on-fail); requires a rebuild of the Android app to take effect.

Screenshot of the fix running in an emulator.

image

Follow-ups
----------
- Report upstream to Breez so the SDK is rebuilt against uniffi-bindgen-react-native >= 0.30 (this breaks every 32-bit user of their React Native SDK); drop this patch once the dependency is updated.
- Consider dropping armeabi-v7a from abiFilters (android/app/build.gradle) or gating the self-custody feature behind a 64-bit ABI check until the upstream fix lands.

### DIFF
--- a/patches/uniffi-bindgen-react-native+0.29.3-1.patch
+++ b/patches/uniffi-bindgen-react-native+0.29.3-1.patch
@@ -1,0 +1,18 @@
+diff --git a/node_modules/uniffi-bindgen-react-native/cpp/includes/RustBuffer.h b/node_modules/uniffi-bindgen-react-native/cpp/includes/RustBuffer.h
+index 68fb1d8..c433065 100644
+--- a/node_modules/uniffi-bindgen-react-native/cpp/includes/RustBuffer.h
++++ b/node_modules/uniffi-bindgen-react-native/cpp/includes/RustBuffer.h
+@@ -10,8 +10,11 @@
+ #include "UniffiCallInvoker.h"
+ #include <jsi/jsi.h>
+ 
++// Patched (patch-package): match uniffi-core 0.29's #[repr(C)] RustBuffer
++// layout (u64 capacity, u64 len, ptr data) on 32-bit targets, where size_t
++// is 4 bytes and would otherwise misalign the FFI struct.
+ struct RustBuffer {
+-  size_t capacity;
+-  size_t len;
++  uint64_t capacity;
++  uint64_t len;
+   uint8_t *data;
+ };


esaugomez31 (dismissed):
## High

### `uniffi-bindgen-react-native` is not pinned in `resolutions`

`package.json:267`

The patch filename hardcodes `0.29.3-1`, but the only thing holding that version in place is the transitive `^0.29.3-1` range declared by `@breeztech/breez-sdk-spark-react-native`, which permits any `0.29.x`. `yarn.lock` pins `0.29.3-1` today, but any `yarn upgrade` or `yarn add` that re-resolves the tree can move it.

Both outcomes are bad. If a newer `0.29.x` touched `cpp/includes/RustBuffer.h`, even just reordering the includes in the hunk's context at lines 10-11, `patch-package --error-on-fail` fails in `postinstall` and every `yarn install` breaks for every dev and every CI job. If it did not touch the file, patch-package applies it with nothing but a version-mismatch warning and we silently carry a patch authored against a different release.

This repo already uses `resolutions` for exactly this kind of transitive pin (`types-ramda: 0.29.4`, `ejs: 3.1.9`, `@react-native/codegen: 0.85.3`). Add `"uniffi-bindgen-react-native": "0.29.3-1"`.

### Nothing fails at build time if the patch goes away

`patches/uniffi-bindgen-react-native+0.29.3-1.patch:12`

The whole fix is three tokens in a `node_modules` header, and nothing in the tree asserts that they are there. If the patch is dropped during a dependency bump, or a future `0.29.x` reshapes the file enough that patch-package lands a fuzzy or partial result, the build still succeeds and the regression comes back as `signal 6 (SIGABRT)` on 32-bit users' phones, which is the exact failure this PR exists to remove.

A `static_assert` turns that into a compile error:

```cpp
static_assert(offsetof(RustBuffer, data) == 2 * sizeof(uint64_t),
              "RustBuffer must match uniffi-core's u64 capacity/len layout");
```

This is not dead weight on 64-bit CI. `android/gradle.properties:31` sets `reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64`, so `armeabi-v7a` is compiled on every Android build, and the assert fires there (offset 8 unpatched versus 16 patched).

## Medium

### A second source of the same panic survives this patch, and the patch widens it

`node_modules/@breeztech/breez-sdk-spark-react-native/cpp/generated/breez_sdk_spark_bindings.cpp:359`

`Bridging<RustCallStatus>::fromJs` declares `RustCallStatus status;` with no initializer, then assigns `error_buf` only when the JS object has an `errorBuf` property and `code` only when it has a `code` property. This is live code, reached from 13+ foreign-future trampolines in `breez_sdk_spark.cpp` (for instance `Bridging<UniffiForeignFutureStructU8>::fromJs` at line 2867, which likewise leaves `rsObject` default-initialized), and it is handed straight to a Rust `UniffiForeignFutureComplete*` pointer. If a JS-side callback ever returns a status with `code != 0` and no `errorBuf`, Rust reads a `MaybeUninit<RustBuffer>` full of stack garbage and `destroy_into_vec()` aborts with the same string this PR is chasing.

Today Rust only reads `error_buf` when `code != Success`, so the trigger is narrow and this is latent rather than the reported crash. It is worth flagging because the description treats the ABI layout as the sole source of that panic, and because the direction of travel is wrong: on `armeabi-v7a` this patch grows `sizeof(RustCallStatus)` from 16 to 32 bytes (measured), so the uninitialized region doubles from 12 to 24 bytes. `RustCallStatus status = {UNIFFI_CALL_STATUS_OK};` closes it, matching what line 263 already does. That means a second patch file, so it is a scope call, but if we skip it we should say so.

### No CI job can validate this fix or catch its regression

`.github/workflows/test.yml:49`

`android-unit-test` runs `./gradlew :app:testDebugUnitTest`, a JVM-only task that never invokes `externalNativeBuild`, so it never compiles `RustBuffer.h` at all. `.github/workflows/e2e.yml` boots an x86_64 emulator, and as measured above the pre- and post-patch structs are byte-identical on `x86_64` (24 bytes, offsets 0/8/16), so that run exercises code this change does not affect and would pass identically with the patch deleted.

Net: the only evidence this works is the manual emulator screenshot in the description, and a future regression reaches production undetected. Since `armeabi-v7a` already compiles on every Android build, the cheap coverage is the `static_assert` above. A real behavioral check would need a 32-bit ABI in the e2e matrix.

### The description points at an SDK version this repo does not ship

`package.json:66`

The description states twice that the bad `.so` ships in `@breeztech/breez-sdk-spark-react-native@0.22.0`, and that `0.22.0 (latest published) still depends on ^0.29.3-1`. This branch pins `"@breeztech/breez-sdk-spark-react-native": "0.15.0"`, and that is what is installed.

Beyond sending a reviewer to the wrong package version, it anchors the removal criterion ("drop this patch once the dependency is updated") to a bump we are not going to perform. I confirmed 0.15.0 does depend on `^0.29.3-1` and does ship the affected `.so` for `armeabi-v7a`, so the claim just needs restating against 0.15.0.

## Low

### The in-patch comment does not carry the removal condition, and misnames the defect

`patches/uniffi-bindgen-react-native+0.29.3-1.patch:9`

The description has the two facts that matter for retiring this patch (upstream fixed the header in `uniffi-bindgen-react-native >= 0.30.0-1`, and the patch should go once Breez bumps that dependency), but PR descriptions are not what someone reads six months from now. They read the comment, which currently says only `// Patched (patch-package): match uniffi-core 0.29's ...`, with no signal that the patch may already be obsolete, and no explanation waiting for whoever hits a hard `--error-on-fail` install break.

`patches/react-native-credentials-manager+0.8.1.patch` is the precedent here: a `/** ... */` block, a `Patched (blink-mobile):` prefix rather than `Patched (patch-package):`, and a closing `Upstream: <link>` line. Worth matching, including the block form instead of a run of `//` lines.

Separately, "would otherwise misalign the FFI struct" is not accurate. I measured alignment as correct on both sides in both variants. The defect is a struct size and field-offset mismatch (12 bytes at 0/4/8 versus 20-24 bytes at 0/8/16). Anyone chasing the word "misalign" goes looking for an `alignas` problem that does not exist.

### The crash evidence is not in the repo

The description opens with "Crash evidence: `docs/tecno-crash/tombstone_00..05` (all six identical)" and builds the whole symbolized-backtrace argument on those files. `git ls-tree -r` on this branch matches neither `tecno` nor `tombstone`, and the diff contains exactly one file. Nothing in `.gitignore` excludes them, so they were simply never committed.

For a fix whose justification is entirely forensic, that leaves the backtrace, the abort message, and the "all six identical" claim unverifiable. Either commit them under `docs/` or attach them here.

### Record why the upstream-bump route was rejected

Follow-ups list reporting this to Breez and dropping `armeabi-v7a` from `abiFilters`, but not the option the description itself raises: a `resolutions` entry forcing `uniffi-bindgen-react-native >= 0.30.0-1`, where this is already fixed. That may well be the wrong call, since Breez's generated C++ was produced by the 0.29.3-1 bindgen and could fail to compile against 0.30 headers, but that reasoning is what should be written down so nobody re-litigates it on the next bump.

It also interacts with the first finding: if the 0.30 route is off the table, the `resolutions` pin at 0.29.3-1 stops being advisable and becomes required. And this is the sixth active patch in `patches/`, so it is worth a line on when we expect to be rid of it.

grimen (approved):

PR #4177, opened 2026-08-17
fix(navigation): stop screens that replace the header back button from crashing

### LINKED ISSUE #4176: bug(kyc): app loading error before KYC flow from hook order mismatch
## Summary

Two KYC support cases are blocked before reaching the KYC flow because the mobile app renders the generic application-data loading error screen.

The error detail shown by one customer is:

```
Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

This points to a React rendering bug in the app rather than a network/device issue or a KYC provider shutdown. Tomas confirmed KYC is active and another Kenya KYC was processed today.

## Affected support cases

- BD-00138: Kenya user, L1 ACTIVE, 0 sats
- BD-00140: Bangladesh user, L1 ACTIVE, 705 sats

## Expected behavior

- Users can open the KYC entry path without the app crashing/loading the generic error screen.
- Hook order remains stable across all account states in the KYC-related render path.

## Notes

- Reported from Slack by Esaú on 2026-08-17.
- Source thread: https://blinkbitcoin.slack.com/archives/C0AEYAALSCQ/p1786987047049509?thread_ts=1786987047.049509&cid=C0AEYAALSCQ
- Screenshots in the Slack thread show the generic error screen and the hook-order error detail.

## Screenshots

Image

esaugomez31:
Closes #4176

## What breaks

`headerBackControl()` handed native-stack the button **component** instead of a render function. native-stack does not render `headerLeft` as a component: it invokes it inline while computing the header config (`useHeaderConfigProps`, called from `NativeStackView`'s own body), so the button's three hooks (`useNavigation`, `useStyles`, `useTheme`) ran on that fiber instead of on its own.

Any screen that then replaces `headerLeft` with a render function that calls no hooks leaves the same fiber rendering three fewer hooks than the previous pass. React throws `Rendered fewer hooks than expected`, and the app-wide error boundary swaps the whole navigation tree for the "issues loading the application data" screen.

Two screens do exactly that:

- the KYC webview, which installs its own back control through `headerLeftNoGlass`, so identity verification takes the app down the moment it opens
- the account-delete flow, which sets `headerLeft: () => null` while the deletion runs

## Reproduced, then verified after the fix

Reproduced on an Android emulator against staging with an account shaped like the reported ones (level 1, no username, zero balance, phone from the same region): Settings, Account, Identity Verification crashed with the exact error the two customers reported. The error boundary's component stack pointed at the fiber that computes the header config, which is what identified the header as the source rather than any KYC code.

After the fix, that same path opens the Sumsub webview, and the header back returns to the onboarding screen. Also exercised on device: Settings, Transaction limits, Fee rates, Account and every back out of them; the four tabs; Receive; and the nested phone-login stack, whose back now pops within its own stack. Zero hook errors and zero JS errors across all of it.

## Scope

Shipping since 3.0.22 (7 Aug), so 3.0.22 through 3.0.26 are affected. It is not region-specific and not an account state, so the reported accounts need no KYC reset: any user who opens the in-app KYC flow hits it.

## Tests

New spec for `headerBackControl` covering the contract (what `headerLeft` receives must be a render function that calls no hooks), the press that dispatches `goBack`, the invisible placeholder staying out of the accessibility tree, and the platform inset. 100% coverage on the changed file across statements, branches, functions and lines. `__tests__/navigation` plus `__tests__/components` stay green: 111 suites, 1182 tests.

## Commits

The cleanup lands first on purpose: it moves the platform check into the stylesheet, which is what makes the inset assertions observable, so the fix commit arrives with its tests already passing and each commit is green on its own.

### DIFF
--- a/__tests__/components/header-back-control/header-back-control.spec.tsx
+++ b/__tests__/components/header-back-control/header-back-control.spec.tsx
@@ -1,0 +1,108 @@
+import React from "react"
+import { Platform, StyleProp, StyleSheet, ViewStyle } from "react-native"
+import { fireEvent, render, waitFor } from "@testing-library/react-native"
+import { ThemeProvider } from "@rn-vui/themed"
+
+import theme from "@app/rne-theme/theme"
+import {
+  headerBackControl,
+  InvisibleBackButton,
+} from "@app/components/header-back-control/header-back-control"
+
+import {
+  DefaultTheme as navigationDefaultTheme,
+  ThemeProvider as NavigationThemeProvider,
+} from "@react-navigation/native"
+import { NativeStackHeaderBackProps } from "@react-navigation/native-stack"
+
+/* Guards the contract documented on headerBackControl: what `headerLeft` receives must
+ * be a render function that calls no hooks of its own (#4176). */
+
+const mockGoBack = jest.fn()
+
+jest.mock("@react-navigation/native", () => ({
+  ...jest.requireActual("@react-navigation/native"),
+  useNavigation: () => ({ goBack: mockGoBack }),
+}))
+
+const BACK_BUTTON_LABEL = "Go back"
+
+const originalPlatformOS = Platform.OS
+
+const headerProps: NativeStackHeaderBackProps = { canGoBack: true }
+
+const backButtonStyle = (button: { props: { style?: StyleProp<ViewStyle> } }) =>
+  StyleSheet.flatten(button.props.style)
+
+/* HeaderBackButton reads react-navigation's own theme, which normally comes from the
+ * NavigationContainer the header lives in. */
+const renderHeaderLeft = (element: React.ReactNode) =>
+  render(
+    <NavigationThemeProvider value={navigationDefaultTheme}>
+      <ThemeProvider theme={theme}>{element}</ThemeProvider>
+    </NavigationThemeProvider>,
+  )
+
+describe("headerBackControl", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  afterEach(() => {
+    Object.defineProperty(Platform, "OS", {
+      configurable: true,
+      value: originalPlatformOS,
+    })
+  })
+
+  it("returns an element instead of invoking the button component", () => {
+    const element = headerBackControl()(headerProps)
+
+    expect(React.isValidElement(element)).toBe(true)
+  })
+
+  it("goes back when the themed button is pressed", async () => {
+    const { getByLabelText } = renderHeaderLeft(headerBackControl()(headerProps))
+
+    fireEvent.press(getByLabelText(BACK_BUTTON_LABEL))
+
+    // HeaderBackButton defers its onPress to the next frame.
+    await waitFor(() => expect(mockGoBack).toHaveBeenCalledTimes(1))
+  })
+
+  it("renders the invisible placeholder when going back is not allowed", () => {
+    const element = headerBackControl({ canGoBack: false })(headerProps)
+
+    expect((element as React.ReactElement).type).toBe(InvisibleBackButton)
+  })
+
+  it("keeps the placeholder out of the accessibility tree", () => {
+    const { queryByLabelText } = renderHeaderLeft(
+      headerBackControl({ canGoBack: false })(headerProps),
+    )
+
+    // It only holds the header slot, so a screen reader must never announce it.
+    expect(queryByLabelText(BACK_BUTTON_LABEL)).toBeNull()
+  })
+
+  /* Both cases assert the correction this component owns, not the final on-screen
+   * position: HeaderBackButton's own platform styles are frozen when its module loads,
+   * so they stay on the iOS branch whatever this test does to Platform. */
+  it("adds no inset correction on iOS", () => {
+    const { getByLabelText } = renderHeaderLeft(headerBackControl()(headerProps))
+
+    expect(backButtonStyle(getByLabelText(BACK_BUTTON_LABEL))).toMatchObject({
+      marginLeft: 0,
+    })
+  })
+
+  it("pulls the button left on Android, where the native header adds its own inset", () => {
+    Object.defineProperty(Platform, "OS", { configurable: true, value: "android" })
+
+    const { getByLabelText } = renderHeaderLeft(headerBackControl()(headerProps))
+
+    expect(backButtonStyle(getByLabelText(BACK_BUTTON_LABEL))).toMatchObject({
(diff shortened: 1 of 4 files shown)


kngako (dismissed):
The findings are non-blocking. Would be good to address some of the test coverage gaps though. 


---

## Findings

### Finding 1 — non-blocking: `headerBackControl` mints a fresh component identity on every call, including inside navigator render bodies

Each invocation of `headerBackControl` creates a new `HeaderBack` function (`header-back-control.tsx:62-68`), and three navigators call it inline in their `screenOptions` on every render (`app/navigation/root-navigator.tsx:252`, `:1026`, `:1084`). This is harmless today only because native-stack 7.16 *invokes* `headerLeft` as a function rather than rendering it as an element, so the child tree (`HeaderBackButtonWithTheme`) keeps a stable type and never remounts.

**When it comes up:** a react-navigation upgrade (or a copy of this pattern into the JS `stack` navigator, whose header renders `headerLeft` via `createElement`) makes the identity churn visible: every `RootNavigator` re-render — theme change, language change, any state bump in that component — would remount the back button on every screen, resetting its internal state and re-running its mount effects. It would manifest as flicker or lost press animation state across the whole app, with no single screen to blame.

**Proposal:** hoist the parameterless default to module scope — e.g. `const defaultHeaderBack = headerBackControl()` next to the navigators and reference it from all three `screenOptions` — so the identity is created once. Keep the parameterized calls in the options callbacks (`root-navigator.tsx:978`, `:993`, `:1001`), since those already re-evaluate only when route params change. Add one line to the `headerBackControl` docstring stating the returned function must stay hook-free *and* that callers should treat it as a render function, not a component, so the constraint survives the next refactor.

### Finding 2 — non-blocking: production code restructured to serve a test seam

The refactor moved the `Platform.OS` check from a module-level constant into the `makeStyles` factory (`header-back-control.tsx:70-82`), so it is re-evaluated on every `useStyles()` call. In production `Platform.OS` never changes, so the only behavioral difference is that the test file can now mutate `Platform.OS` per-test (`header-back-control.spec.tsx:27`, `:93`) and observe the result without re-importing the module.

**When it comes up:** the pattern sets a precedent — the next time a module-level platform constant is "in the way" of a test, it gets demoted to call time too, and the codebase slowly loses the (admittedly small) compile-time-ish guarantees of module constants. More concretely, if `@rn-vui/themed`'s `makeStyles` ever memoizes the styles object per theme across hook calls, the Android test would silently validate a cached style and pass while the app behavior it claims to pin has drifted.

**Proposal:** either revert to the module-level constant and write the platform assertions with `jest.isolateModules`/`jest.resetModules` plus a `Platform.OS` mock applied *before* importing the component, or keep the current shape and add a short comment in `useStyles` saying the check lives inside the factory so tests can vary the platform. The first option keeps the production code honest; the second at least makes the trade-off explicit for the next reader. Both are cheap; leaving it unexplained is the only bad option.

---

## Test Coverage Gaps

1. **The actual #4176 failure mode is not exercised.** The six tests pin the contract indirectly (`React.isValidElement`, press behavior), but nothing reproduces the real crash sequence: a screen mounting with the default `headerBackControl()` header and then calling `navigation.setOptions({ headerLeft: <hookless render> })`, as `webview.tsx:60-68` and `delete.tsx:126-129` do. A regression that reintroduces a hook into the returned render function *would* fail the current spec (invalid hook call when invoked in the test body), but only by accident of how the test invokes it. A detox e2e step that opens the KYC webview and navigates back — or at minimum a comment in the spec naming the two screens that depend on this contract — would close the loop.

2. **The options-callback `canGoBack` path is untested.** `root-navigator.tsx:993` and `:1001` pass `route.params?.canGoBack`, relying on the `= true` default when params are absent. No test covers `headerBackControl({ canGoBack: undefined })` or the param-toggle transition (`canGoBack: true` → `false` via `setParams`), which is exactly where a future "simplify the defaults" refactor would break the onboarding screens.

3. **Header-prop passthrough is untested.** `HeaderBackButtonWithTheme` spreads `{...props}` before overriding `style`/`onPress` (`header-back-control.tsx:42-48`); nothing asserts that header-supplied props like `tintColor` or `label` survive. Dropping the spread would keep all six tests green while silently un-theming the back chevron.

4. **The Android inset test depends on global mutation + non-memoized styles.** It redefines `Platform.OS` (`spec:93`) and relies on `useStyles` re-evaluating its factory — see Finding 2. There is no test asserting the iOS and Android corrections in a single suite run with module isolation, so a memoization change in the theming library could turn the Android assertion into a false positive.

## Code Smells

- **Docstring citation is slightly off** (`header-back-control.tsx:52-61`): `useHeaderConfigProps` is not "called from NativeStackView's own body" — it runs inside a per-screen child component of `NativeStackView` (`NativeStackView.native.js:181`). The mechanism described (hooks landing on the header-config fiber, not the button's) is correct, but the loose citation will send the next debugger to the wrong function body.

- **`HeaderBack.displayName = "HeaderBack"` never surfaces** (`header-back-control.tsx:65`): native-stack invokes `headerLeft` as a function, so `HeaderBack` never becomes a fiber and its `displayName` appears nowhere in devtools — every screen's back button just shows as `HeaderBackButtonWithTheme`. Either give per-instance names (e.g. include `canGoBack`) or drop the line.

- **`HeaderBack` shadows native-stack's real `canGoBack` prop** (`header-back-control.tsx:63-64`): the closure value wins over the `canGoBack` in `NativeStackHeaderBackProps`, so a root-of-stack screen with the default `headerBackControl()` renders a live back button whose press no-ops. Pre-existing behavior, but the prop/closure collision is now easier to miss because both are named `canGoBack`.

- **`contact-support-screen.tsx` change is comment-only but rides along in the fix commit.** The rewritten comment is accurate and worth having; it would just be cleaner as part of the refactor commit `57e2732c0` so the fix commit stays a single behavioral change.

- **Spec mocks `useNavigation` module-wide** (`header-back-control.spec.tsx:20-23`): fine here, but the file also renders real navigation theme providers via `requireActual` — a future test in this file that needs the real `useNavigation` will have to untangle the partial mock. A `jest.spyOn`-style scoped mock would age better.

---

## Verdict

**Approve — mergeable as-is, no blocking findings.** The fix is small, correctly diagnosed, verified against the installed navigation libraries, and comes with a real test suite that would have caught the original bug. Findings 1 and 2 are follow-up hygiene (component-identity discipline and test-seam honesty), and the coverage gaps are worth a short follow-up ticket rather than holding this crash fix.

grimen (approved):

PR #4178, opened 2026-08-18
feat(map): search the merchants and filter them by category

### LINKED ISSUE #4071: feat: Source merchants from BTCMap

kngako:
The ~29k BTC Map merchants could only be reached by panning the map until one came into view. This puts a search field and a category filter over the top of it, so "where can I spend bitcoin near here, of the kind I actually want" can be asked rather than hunted for.

## Search

The field is a button, not an input: tapping it opens a full-screen list, which is the only place a keyboard and its results both fit. Before anything is typed the list is what is nearby; typing narrows it. Picking a result flies the map to that place at the zoom where pins stop clustering and opens the place sheet on it, so dismissing the sheet leaves the user looking at the place they chose rather than wherever they were beforehand.

BTC Map's `places/search` takes no text query, only a circle, so typing cannot be a request. One call fetches the named places around the view when the search opens, and every keystroke filters that list on the device. Results appear as fast as the letters do, and a search keeps working on the connection someone standing in an unfamiliar town actually has.

That endpoint also ignores `limit` and `fields` and returns everything in the circle, so the radius is the only thing bounding the response — ~150 KB over a city at 10 km, ~470 KB at 200. It is capped at 25 km, far enough to be worth travelling and small enough to ask for on mobile data, with a 5 km floor so searching while zoomed into one street still reaches the next one. Nothing is fetched until the search is opened: the map tab stays mounted for the life of the process, so an unconditional fetch would spend a few hundred KB on every app start for a feature most sessions never touch.

The centre is snapped to the same ~1.1 km privacy grid the map labels already use, so opening a search does not tell BTC Map where the user is any more precisely than looking at the map already did. That snapping moves out of `use-place-names.ts` into `geo.ts` so its two callers cannot drift apart, and `fetchPlaceNamesNear` becomes a thin reduction of the new `fetchPlacesNear` rather than a second copy of the same request.

Matching is substring over an accent- and case-folded name, not fuzzy: a wrong-but-plausible match sends someone walking to the wrong shop, and "did you mean" guesses are not worth that. Only combining marks are stripped, never the letters carrying them, so names in scripts that do not decompose — Cyrillic, Thai, CJK — are left exactly as they were typed rather than folded into something no keystroke can match.

Distances are measured from the phone's own position, which is also what orders the list; the number on a row and the position of that row have to come from the same point or the list reads as unsorted. Without location permission there is no honest distance to print, so a row falls back to the place's address — about 45% of named places carry one in the area sampled — and the rest are left as just a name, which is still the thing being searched for.

## Category filter

BTC Map publishes no category. The only classification that reaches us is the Material icon name it tags each pin with, and that is all the offline snapshot carries, so `btcmap/categories.ts` buckets those ~157 icon names into fifteen categories. Filtering then runs against the cached list with no network and no per-place detail fetch; the alternative is ~29k requests. It also means two places drawn with the same pin share a bucket, which is what anyone reading the map already assumes.

Measured against the current feed the buckets place 98.9% of the ~29k places. The remainder — community groups, churches, "we don't know what this is" — land in `other` rather than disappearing from a filtered map, which is also where a new icon goes when BTC Map adds one without us.

Nothing selected means no filter, not no places. A map that went blank as the last toggle was switched off would read as broken and would leave no way back, so an empty selection is what makes "clear all" a way back to the whole map rather than a way to empty it. In that case the place list is returned by identity rather than copied, because the clusterer rebuilds its index over all ~29k points whenever that array changes.

The filter applies to the search list too: someone who has just narrowed the map to groceries and then searches is still looking for groceries, and a list that contradicted the map behind it would be worse than a shorter one.

## Notes

- 24 new `MapScreen` strings, translated into all 28 locales. `common.search` is reused rather than duplicated. `metersAway`/`kilometersAway` carry the number as a `{distance: string}` placeholder, and a spec asserts every locale still contains it — losing it in translation would drop the distance silently while the row still rendered.
- The status pill that reports a failed place sync moves down to clear the search bar, which now owns the top edge of the map.
- Depends on the react-native-maps patch in the previous commit; without it the filter appears to do nothing, because removed markers stay on the map.
- No new dependencies and no new permissions. Location was already requested for centring and the opening-hours badge, and the search degrades to addresses without it.
- Covered by specs for the icon buckets (including that every bucket is reachable and no icon is filed twice), the fold-and-rank helpers, the fetch hook's caps and staleness guard, the search modal's states, the filter sheet's toggles, and the map's wiring of both. `yarn tsc:check`, eslint over `app` and `__tests__`, and the full jest suite are green.
- Verified on an Android emulator against the live BTC Map feed over San Salvador: nearby list with distances, typed narrowing, the empty state, selecting a result, and each filter showing only its own category.

The search and filter bar is at the top.

image

Search defaults to showing shops in the view (or closest to the users location)

image

Typing filters on these stores.

image

image

Categories filters on what gets shown on the map

image


designsats (approved):
- Search works fast and smoothly on my android

Minor non-blocking fix: 
- empty state text should have same text size as regular search result - paragraph2 16px, seems smaller in test build 
[Figma](https://www.figma.com/design/9MQuQi8ZhXVvDibWSI3C4c/Blink?m=auto&node-id=20943-118543&t=YnvKuiYaUHpUszCj-1)

Image

grimen (changes requested):
## Findings

**1. Search result does not open the place sheet on iOS (blocking).**
`handleSearchSelect` in `app/components/map-component/index.tsx:221` closes the search modal and opens the place sheet in the same batched update. Both are React Native `Modal`s, and iOS silently drops a presentation that begins while another modal is still dismissing.

*Example:* On iOS, tapping a search result flies the map to the place but the sheet never appears, so the user has to find and tap the pin manually; the PR notes verification on an Android emulator only, where this does not reproduce.

*Proposal:* One option could be to defer `setSelectedPlace` until the search modal has actually dismissed, via `onDismiss` on iOS or a `requestAnimationFrame`/`InteractionManager` hop.

**2. Previous area's places stay listed during a new fetch (blocking).**
`app/btcmap/use-place-search.ts:80` deliberately keeps the old `places` while a fetch is in flight, but never checks that the new fetch covers the same area, and `enabled` sits in the dependency list so every reopen refetches.

*Example:* Search in San Salvador, close the modal, pan to another city and reopen search — for the length of the request the old city's places render with distances and remain pressable, and tapping one flies the map back to the old city.

*Proposal:* Keeping the old list only when the snapped `latitude`, `longitude` and `radiusKm` are unchanged, and clearing it otherwise, would cover this; the comment at line 63 also claims reopening does not re-ask, which no longer holds now that `enabled` flips false→true on every open.

## Test Coverage Gaps

- No spec covers reopening the search after the map has moved to a different area, which is the stale-list path in finding 2.

grimen (approved):

PR #4185, opened 2026-08-20
fix(btcmap): These are a number of UI/UX fixes on the btcmap feature.

### LINKED ISSUE #4071: feat: Source merchants from BTCMap

kngako:
Fixes these from @designsats 

---

1. The map doesn't respect the app theme (dark/light).
2. The "Navigate" CTA label should use the `background` color — currently it's white in dark mode, which is wrong.
3. We don't use shadows under buttons. The "center on user location" button incorrectly has one.
4. The "Select all" text button and the links in the merchant details should be bold.
5. Links in the merchant details should also be underlined.
6. The partially open merchant sheet should fully reveal the "Verified" info — it needs 20px more on my device (we need a universal offset so it behaves this way on every screen size).
7. The companion-app info isn't using our standard info box component. Don't mimic it — take it from Storybooks.
8. Verify that the merchant details content uses the `paragraph3` style (14px); it looks too small in testing.
9. Remove the merchant icon, name only.
10. Show socials like any other link, last in the merchant info list.
11. Icons should match the link color (`primary`).
12. Labels collide in high-density places "Berlín, SV" 
[Screenshot](https://github.com/user-attachments/assets/5182d68f-7c62-4602-9a62-c362bdce8702)


grimen (changes requested):
## Findings

**1. Lowering the clustering cut-off to zoom 15 makes label collisions worse (blocking).**
`CLUSTERING_DISABLED_ZOOM` drops from 17 to 15 in [`use-place-clusters.ts:15`](https://github.com/blinkbitcoin/blink-mobile/blob/163349d1581423589a71c90449fab045e3f7e736/app/components/map-component/use-place-clusters.ts#L15), but `LABEL_MIN_ZOOM` is also 15 and [`index.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/163349d1581423589a71c90449fab045e3f7e736/app/components/map-component/index.tsx) renders one `PlaceLabelMarker` per individual place with no decluttering pass anywhere. Zoom 15 and 16 therefore go from a handful of cluster discs to one pin plus one label per place, and the `MAX_RENDERED` cap now silently drops pins that used to be visible inside a cluster.
*Example:* A zoom-15 viewport over a dense city centre is roughly 1.8 × 3.8 km, so it renders hundreds of pins and the same number of overlapping names — the exact "Berlín, SV" overlap the designer flagged.
*Proposal:* One option could be raising `LABEL_MIN_ZOOM` to 17 instead, so pins and labels still arrive together but only once there is room for them.

## Test Coverage Gaps

- The zoom-15 clustering change has no test — [`use-place-clusters.spec.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/163349d1581423589a71c90449fab045e3f7e736/app/components/map-component/use-place-clusters.spec.ts) only has a comment updated, so nothing verifies that individual places are returned at zoom 15 or how the render cap behaves there.
- The new resting-offset math in [`place-sheet.tsx:165`](https://github.com/blinkbitcoin/blink-mobile/blob/163349d1581423589a71c90449fab045e3f7e736/app/components/map-component/place-sheet.tsx#L165) (`peekBottom = y + height` and the `- insets.bottom` subtraction) has no test — `place-sheet.spec.tsx` never asserts the snap point.

designsats (changes requested):
Reviewed at head `7560cc603`, against the 12-item ticket in [#4164 (review)](https://github.com/blinkbitcoin/blink-mobile/pull/4164#pullrequestreview-4964297607).

Verified locally on this branch: `tsc --noEmit` clean, `eslint` clean, **103/103 map-component tests pass** (up from 101 — two new). Built and ran both variants on device: release on a Pixel 7a, debug on an API 36 emulator.

**Ticket score: 10 of 12 done.** Items 2–11 all check out. Items 1 and 12 are not addressed, and 4 of the 5 follow-on asks in the same review were dropped without comment.

---

## Real strengths first

**S1 — Item 6 is the best work in the PR.** The ticket said "it needs 20px more on my device (we need a universal offset)". You correctly refused the magic number and solved the general case: the snap point is now the peek's measured *bottom edge* (`y + height`, so the handle and padding above it count), plus `insets.bottom` so the home indicator can't swallow the verification row.

```ts
// place-sheet.tsx:165
const restingOffset = peekBottom
  ? Math.max(0, sheetHeight - peekBottom - insets.bottom)
  : sheetHeight
```

The `Math.max(0, …)` clamp means a peek measured taller than the sheet stops at fully-open instead of pushing the top edge off screen — and there's a test for exactly that. This is the right shape of fix.

**S2 — The new tests actually pin the behaviour.** `place-sheet.spec.tsx` gets two cases that would fail on a revert: one asserting the sheet stays at `translateY: sheetHeight` until a layout event lands (no guessing then hopping), one asserting `translateY: sheetHeight - (24 + 200) - 34` with a mocked 34px inset. The lazy `mockBottomInset` so a single test can raise the inset before render is a tidy touch.

**S3 — An unlisted bug fix worth calling out.** `isRenderablePlace` was made generic in `btcmap/api.ts:108`, so the guard stops narrowing `name`/`address` off the caller's type:

```ts
const isRenderablePlace = <T>(value: T): value is T & RenderablePlaceWire =>
```

That's a real defect the ticket never mentioned — search rows were losing their fields to the filter. Please add it to the PR description; per `docs/pr-review.md` every behaviour change needs a note.

**S4 — Item 2 was fixed in both places.** `colors._white` → `colors.white` on `navigateText`, and the same latent bug on the filter icon in `map-search-bar.tsx:71`, which the ticket didn't ask for. Since `white` inverts `#FFFFFF`→`#000000`, the Navigate label is now black-on-amber in dark mode (~11.2:1) instead of white-on-amber (1.9:1).

**S5 — No orphans left behind.** `materialIconName` and `usePinColor` are dropped from `place-sheet.tsx` but still correctly used by `place-marker.tsx`; `boosted` survives because the badge at line 433 still needs it. Clean removal.

---

## Important

### I1 — Item 12 was implemented, then reverted with no explanation

`163349d15 feat(map): stop clustering where the labels start` did the work — and its own commit message says "eslint clean, all 101 map-component tests pass". It was then backed out by `46fc7672d`, whose entire body is:

```
Revert "feat(map): stop clustering where the labels start"

This reverts commit 163349d1581423589a71c90449fab045e3f7e736.
```

No reason given. This is the item the ticket attached a screenshot for ("Berlín, SV"), and `marker-layout.ts` still has no collision handling. Either restore it or say in-thread why raising the cluster threshold was the wrong mechanism — a documented, reasoned pushback is a perfectly good resolution here, but silence isn't. 
Berlín remains an unclustered mess.

### I2 — Item 1 (map theme) has no corresponding change

Nothing in the diff touches map theming. The wiring at `index.tsx:255` predates both this PR and #4164:

```tsx
customMapStyle={themeMode === "dark" ? MapStyles.dark : MapStyles.light}
```

`git log` shows `map-styles.json` last changed in #4164 itself, and only to strip POI label colours. So either this was fixed inside #4164 before merge and should be struck from this PR's list, or it's still open. Worth flagging that `customMapStyle` is Google-provider-only — it's a no-op against Apple Maps on iOS, so if the map is ever checked there it will not be dark regardless. Please confirm which case this is.

### I3 — The component-swap ask was 1-for-3, and it was the flagged one

The review said, of replacing the hand-rolled controls: *"If you only take one thing from this comment, take this one."* `GaloyInfo` landed. The other two didn't:

```tsx
// place-sheet.tsx:387 — still hand-rolled
<Pressable style={styles.navigate} onPress={navigate} accessibilityRole="button">
  <Text style={styles.navigateText}>{LL.MapScreen.navigate()}</Text>
</Pressable>

// place-sheet.tsx:530 — still hand-rolled
<Pressable testID="close-place-sheet" style={styles.close} onPress={onClose} …>
```

These should be `GaloyPrimaryButton` and `GaloySecondaryButton`. This is also the root cause of D1 below — swapping them deletes the off-scale `borderRadius: 28` / `minHeight: 56` / `fontSize: 17` block rather than requiring a separate cleanup. The ticket's general note was explicit: *"Make sure components are never reinvented, always taken as they are in existing flows."*

### I4 — Two colour-token swaps from the same review are untouched

- `cluster-marker.tsx:41,64,71` still uses `colors.success`, which we never define — it falls through to the @rn-vui default (`#52c41a` / `#439946`). Every cluster disc on the map is a library green rather than Blink's `#00A700`. Should be `colors._green`.
- `pin-shape.tsx:28` still has `PIN_GLYPH_COLOR = "#FFFFFF"` where `colors._white` is the token that exists for it.

Both are one-line changes.

---

## Non-gating notes

**D1 — Scale drift persists.** Radii `28` (navigate), `10` (badge), and type sizes `17` (navigate, close), `13` (error, retry) are off the product scale (`8 / 16 / 999` and `24 20 18 16 14 12`). Mostly resolved for free by I3.

**D2 — `grey5` on a non-tappable badge.** `place-sheet.tsx:632` — `grey5` signals "responds to touch"; the status badge doesn't. `grey7` is the static equivalent. The `infoCard` instance of this is gone now that `GaloyInfo` owns it, so this is the last one. It should be active or grey7. 

**D3 — Item 8 is satisfied by size but not by token.** `address`, `rowText`, `rowLink` and `description` are all `fontSize: 14`, which matches `p3`. But they set it raw rather than using `type="p3"`, so they miss `lineHeight: 18` and will drift from `p3` if the scale ever moves. `requiresAppLink` does use `type="p3"` — worth making the rest consistent.

**D4 — The `GaloyInfo` adoption uses a newline join.** The message and link are one `Text` tree separated by a literal `{"\n"}`:

```tsx
<GaloyInfo>
  {LL.MapScreen.requiresApp()}
  {"\n"}
  <Text type="p3" style={styles.requiresAppLink} onPress={…}>
```

It renders correctly and `colors.blue5` is restated deliberately (it matches `GaloyInfo`'s own tint and inverts properly, so this is right). But it forced the test to loosen to `{ exact: false }` because no node carries the message alone. Acceptable as-is; noting it because a `GaloyInfo` that accepted a link slot would be the cleaner fix if this pattern recurs.

---

## Suggested landing order

1. **I4** — four lines total, and it's the only finding currently visible on every screenful of map.
2. **I3** — the biggest net-negative diff; takes D1 and most of D2 with it.
3. **I1 / I2** — these need an answer more than they need code. If item 12 was reverted for a good reason and item 1 was fixed upstream in #4164, say so and the ticket closes at 12/12.
4. **D3 / D4** — follow-up, not gating.

grimen (changes requested):
## Findings

**1. The resting sheet slices the row below the peek (blocking).**

`restingOffset` subtracts `insets.bottom` from the peek's measured bottom edge, so the sheet rests that much higher than the peek actually needs. The peek only pads itself by 14, so on a device with a home indicator the extra lift uncovers roughly 20pt of the content that follows it, and what shows is the top of the next row rather than empty space.

*Example:* on a place with a phone number, at rest on an iPhone with a 34pt bottom inset, `+503 7515 6510` is cut through the middle of its glyphs by the screen edge, underline included — see the after recording in https://github.com/blinkbitcoin/blink-mobile/pull/4185#issuecomment-5356783622.

*Proposal:* one option could be to keep the lift in the layout rather than in the offset — pad the peek by `14 + bottomInset` and rest on its measured bottom edge unchanged — so the strip above the home indicator is the peek's own padding and no row is half-shown. The new snap-point test would need its expectation moved along with it.

grimen (approved):

designsats (approved):
Visually correct from my end. Tested on Android device

PR #4186, opened 2026-08-20
feat(map): let level 2 custodial accounts add a place to BTC Map

### LINKED ISSUE #4184: feat(map): Add Merchants - custodial Users can submit store locations to BTCMap
Blink now uses BTCMap to show users stores. Users may want to add new locations that are not in BTCMap

Add UI to allow users to add new locations to BTCMap.

Alternative would require users to go to BTCMap.org to add new stores.

**Additional context**

This will make use of the authentication proxy.

kngako:
The map has been a window onto BTC Map that could only be read. Someone standing in a shop that takes bitcoin and is not on the map had nowhere to say so from. This adds the flow that lets them: drop a pin, name what is there, and submit it for BTC Map's community to survey.

Referenced against d4rp4t's `feat--btcmap-integration`, which prototypes the same feature. The two trees have diverged — that branch refactored the map into `app/components/map-components/`, while this one builds on `app/btcmap/` plus `app/components/map-component/` — so what is carried over is the shape of the feature (a pill on the map, a pin to place, name/address/category, a TODO where the submission goes) rather than the code.

## Who gets it

Level 2 Custodial accounts only: 

The submission is to leave through our own backend, which needs a Blink session behind it, so a self-custodial account has nothing to send it with. The button is absent rather than present and failing at the end of a filled-in form.

`useIsSelfCustodialAccount` rather than `useActiveWallet().isSelfCustodial` deliberately. The latter encodes SDK availability as well as account type and reports `false` while the self-custodial SDK is still loading, which would flash the button into view for exactly the users it exists to exclude. The hook answers the narrower question — which kind of account is signed in — and keeps answering it during those renders.

## Placing the pin

Two steps rather than one. d4rp4t draws the pin above a bottom sheet and calls `measure()` on it, converting its `pageY` against the screen's midpoint into a latitude offset. This draws the pin at the centre of the map view instead, with its tip on the centre point and nothing else on screen but a hint and a bar, so where it points *is* `region.latitude/longitude`: no measuring, no conversion, and no way for the two to drift apart. Pan until the tip is over the door, then confirm.

`PlaceLocator` is `pointerEvents="box-none"` above its bar, so the map keeps every pan and pinch it had before the overlay opened. The coordinates are read from `regionRef.current` rather than `region` state, so confirming lands on the region the user is actually looking at rather than one render behind it.

While the pin is down, the search bar, the category filter and the add button all leave: they are about reading the map, and none of them belongs over one that is being used to point at something. The existing pins go quiet too — a tap is aiming at a coordinate, not asking about what is already there.

Two things share the bottom of the map and had to be worked around rather than covered. The ODbL credit is an attribution we are obliged to keep legible, so the bar sits 34px up to clear it. The location button is genuinely useful here — centring on yourself and then nudging the pin onto your own shop is the common way to place one — so `LocationButtonCopy` takes an optional `bottom` and is lifted above the bar instead of being hidden.

## Describing the place

Full screen rather than a sheet over the map. By this point the location question has been answered, so there is nothing left to see behind it, and three fields plus a keyboard do not fit under a sheet. It also sidesteps the keyboard-avoidance problem a bottom sheet with inputs creates on Android.

The category is a wrapping grid of chips, not a picker behind a row. There are fifteen of them, so they all fit on screen and are one tap away instead of three, and the choice being made stays visible while it is made. They are the buckets `btcmap/categories.ts` already defines for the filter, so this adds no category strings and no second taxonomy to keep in sync.

A name and a category are both required; the address is not. The name is the only thing identifying the place to whoever surveys it, and the category is what the map draws — BTC Map classifies a place by the icon on its pin, so a place with no category has no pin. Submit stays disabled rather than explaining itself afterwards, since which of the two is missing is visible on the form, and `accessibilityState` carries that so the explanation is not only in the translucency.

What has been typed lives exactly as long as one attempt. The map re-keys the form per attempt, so "change location" is a correction that keeps everything typed, while cancelling out of the locator afterwards discards it. Without the key that abandoned draft reappeared in the next attempt — found while wiring it, and both directions are now pinned by tests.

## Sending it

Not wired, as intended: the backend proxy does not exist yet.

`btcmap/submission.ts` holds everything that does not depend on it — the payload type, trimming, the required-field and coordinate-range rules, and the six-decimal coordinate formatting OpenStreetMap itself uses — so the endpoint landing is a one-line change at the TODO in `map-component/index.tsx`.

Submit does not claim success. d4rp4t answers with "Your suggestion has been sent!", which would not be true; this shows a warning that sending places is not switched on yet. Thanking someone for a submission that went nowhere is worse than telling them it did not go. Swap `submissionNotReady` for a success path when the proxy lands. The payload is logged under `__DEV__` only — what someone typed about their shop has no reason to be in a release build's log.

## Notes

- 13 new `MapScreen` strings, translated into all 28 locales. `common.cancel`, `common.next`, `common.submit` and `common.close` are reused rather than duplicated. None of the new strings interpolates anything, so there is no placeholder for a translation to lose.
- `app/btcmap/config.ts` and both docs asserted the app was a read-only consumer of BTC Map. That is no longer quite true, so the claim is now the narrower one that holds: everything drawn is read from there, the one thing sent back is a proposal for a place that is missing, and nothing in the app edits or verifies a place that already exists.
- No new dependencies, no new permissions, no new remote-config flag. Location was already requested for centring and the opening-hours badge, and placing a pin does not need it.
- 22 new specs: `submission.spec.ts` over the validation and formatting rules, `add-place-modal.spec.tsx` over the form's required fields and chip behaviour, and additions to `map-component.spec.tsx` over the custodial gate (offered, withheld from self-custodial, withheld when signed out), the handover of the map to the pin, the coordinates the form is handed, the round trip back to the pin, and the draft lifetime.
- The two draft-lifetime tests were mutation-checked: dropping the per-attempt key bump kills "does not carry an abandoned place into the next one", keying on the step instead kills "keeps what has been typed while the pin is moved", and the restored component passes both.
- `yarn tsc:check`, eslint over `app` and `__tests__`, and the full jest suite (658 suites, 7999 tests) are green. `yarn check:translations` is a post-commit guard — it greps `git diff` — so it was verified by re-running `yarn update-translations` and confirming the generated files do not move.
- Not yet verified on a device. The layout numbers — the bar clearing the ODbL credit, the location button lifting above the bar, the pin's tip landing on the centre point — are reasoned rather than seen, and are worth a look on a real screen before this merges.


esaugomez31 (changes requested):
@kngako

## Blockers

### 1. `check-code` is red, and this PR is what broke it

[Check Code job](https://github.com/blinkbitcoin/blink-mobile/actions/runs/33093629275/job/98592721127):

```
Detected 1 invalid document:
 - Unknown type BtcMapPlaceSubmitInput. Did you mean FeedbackSubmitInput?
 - Cannot query field btcMapPlaceSubmit on type Mutation.
make: *** [Makefile:2: check-code] Error 1
```

Exactly one invalid document, and it is the one this branch adds, so the failure is introduced here rather than pre-existing. It reproduces locally with `yarn graphql-check`: same exit code, same two lines. `graphql-check` validates `app/graphql/generated.gql` against the live `schema:` URL in `codegen.yml`, so the `local-schema.gql` stub satisfies codegen but never reaches the validator. It is also the last step of `check-code` (`package.json:14`), so every branch cut from this one inherits the red check.

Android and iOS are failing too, but for an unrelated reason: both die in `upload-cloud-storage` with `ENOENT: no such file or directory, lstat '.../screenRecord.mp4'`.

### 2. `uuidv4()` has no RNG under React Native, so the Map tab crashes for every user

`app/components/map-component/index.tsx:118` and `:300`

uuid@9.0.1 resolves to its browser RNG under Metro, because `resolverMainFields` includes `"browser"` (`metro.config.js:40`), and that RNG throws `crypto.getRandomValues() not supported` when there is no global crypto. React Native 0.85.3 ships none (no `getRandomValues` anywhere in `Libraries` or `src`), `react-native-get-random-values` is not a dependency, `react-native-quick-crypto`'s `install()` is never called, and `index.js` installs no polyfill. The only other uuid call site in the repo works around exactly this: `app/screens/get-started-screen/use-create-device-account.ts:23` passes `{ random }` from `react-native-securerandom`.

The call sits in a `useRef` argument, so it is evaluated on MapComponent's first render, before `canAddPlace` is ever consulted. Signed-out, self-custodial and level-0 accounts hit it as well, not only the level-two custodial accounts the feature targets. The unit tests stay green because Node exposes `globalThis.crypto`.

Suggested fix: reuse the `generateSecureRandom(16)` pattern from `use-create-device-account.ts`, or add `react-native-get-random-values` and import it from `index.js`.

### 3. The mutation is not served by any deployed API, and a failed submission is retried five times

`app/graphql/local-schema.gql:5-37`

Introspecting both endpoints returns `null` for `BtcMapPlaceSubmitInput`, on `api.staging.blink.sv` and on `api.blink.sv`. At runtime the resulting 400 reaches `shouldRetryOperation`; `btcMapPlaceSubmit` is not in `noRetryOperations`, and `hasIdempotencyKey` inspects request headers while the submissionId is a GraphQL variable, so RetryLink resends up to five times (`app/graphql/client.tsx:163`) before the catch returns `false`. The user waits through five doomed round trips and is then told to check their connection. Nothing holds the feature back in the meantime: `canAddPlace` (`app/components/map-component/index.tsx:95`) has no remote-config term.

### 4. The address the form collects is never sent

`app/btcmap/use-place-submission.ts:46-52`

The modal renders a labelled address field bounded by `PLACE_ADDRESS_MAX_LENGTH` (`app/components/map-component/add-place-modal.tsx:151-165`), and `buildPlaceSubmission` trims it and attaches it to the payload (`app/btcmap/submission.ts:51,60`). The mutation input is then built as `{ submissionId, name, category, latitude, longitude }`. It could not be sent in any case, because `BtcMapPlaceSubmitInput` has no `address` field, so adding one would not type-check. A user types their street address, sees the success toast, and BTC Map receives a name and a pin. Both new specs bracket the gap without catching it: the modal spec asserts `onSubmit` receives the address, and the hook spec asserts the mutation variables omit it.

Either drop the field until the backend accepts one, or land `address: String` on the input first.

## Non-blockers

### 5. Every backend failure is collapsed into a single boolean

`app/btcmap/use-place-submission.ts:56`

The document selects `errors { message }` and then discards it through `Boolean(payload && payload.errors.length === 0 && payload.place)`, and the bare `catch` on the next line swallows ApolloErrors the same way. A permanent rejection (rate limit, level too low, duplicate place) is indistinguishable from a dropped request, so both show "check your connection". Since the failure path deliberately keeps the form open on the same submissionId, the natural response is to retry the identical payload indefinitely.

### 6. No in-flight guard on submit

`app/components/map-component/add-place-modal.tsx:74-76`

`submit()` calls `onSubmit` and returns immediately, while the handler it is given (`index.tsx:313`) is an async round trip. The button's only gate is `disabled={!submission}` at line 207, which stays false for the whole request, and there is no spinner. Repeated taps fire concurrent mutations and stack toasts, and a response that lands after the modal is closed still toasts over the map.

### 7. The kill switch covers only half the feature

`btcMapPlacesEnabled` appears only in `app/btcmap/use-places.ts`. Turning it off during an incident empties the pins while the Add-place button stays on screen and submissions keep flowing to the backend, which is the outcome the flag exists to avoid.

### 8. The docs contradict the code added in the same PR

`docs/architecture.md:163` still says "Not wired. `btcmap/submission.ts` shapes and validates the payload; the call that would forward it to BTC Map is a TODO in `map-component/index.tsx`". There is no TODO, and `handlePlaceSubmit` fires a live mutation. Line 162 also records the gate as `useIsAuthed() && !useIsSelfCustodialAccount()`, omitting the `isAtLeastLevelTwo` term the code actually applies. `docs/source-tree-analysis.md:200` lists `submission.ts` but not `use-place-submission.ts`.

### 9. Composite conditions inlined into JSX props

`!submission` is written three times in adjacent props at `add-place-modal.tsx:205-209` (`style`, `disabled` and `accessibilityState`), so the visual, touch and screen-reader states can drift apart. In `index.tsx`, `isVisible={addStep === "describing"}` at line 521 sits next to `isVisible={isSearchOpen}` at line 500 and `isVisible={isFilterOpen}` at line 511, and line 492 puts both a comparison and a ternary in a prop value. A named constant per condition (`isSubmitDisabled`, `isDescribingPlace`, `isLocatingPlace`) covers all of them.

### 10. Dead translation key

`chooseCategory` is added to `app/i18n/en/index.ts:2145`, all 28 translation JSONs and two generated type blocks, and nothing renders it. The form labels the category field with `placeCategory`, and the chip row has no placeholder. There are no references outside the i18n files themselves.

### 11. `other` is submittable, and the taxonomy is ours rather than OSM's

`other` is the last entry of `PLACE_CATEGORIES`, so it renders as a selectable chip, and `buildPlaceSubmission` only checks the field for truthiness, which means a place can be submitted with no usable category. The other fourteen values are this app's own bucketing of Material icon names, invented for the client-side filter, and they go on the wire with no mapping to an OSM tag. The accepted values are worth agreeing with the backend and pinning as an enum.

### 12. Three hand-rolled copies of the design-system buttons

`GaloyPrimaryButton`'s `buttonStyle` is `minHeight: 50` plus `colors.primary` and its `disabledStyle` is `opacity: 0.5`, which is exactly what `submit`/`submitDisabled` restate in `add-place-modal.tsx`, down to the comment about translucency. `place-locator.tsx:117-142` repeats the same pair twice more. The copies lose `testProps`, `TouchableHighlight` and the `loading` prop that item 6 needs, and they hardcode `borderRadius: 25` where the theme's Button uses 50. The text inputs are duplicated from `place-search-modal.tsx:243-249` and the chips from `PercentageSelector` in the same way.

### 13. `level-context` is mocked wholesale in the spec

`__tests__/components/map-component/map-component.spec.tsx:122-124` replaces the module without `jest.requireActual`, so `AccountLevel` and `LevelContextProvider` become undefined for everything in the render tree, and `useLevel` returns one field where the real hook returns five. The `is-authed-context` mock three lines above uses the partial pattern correctly.

### 14. One Portuguese string is in the wrong variant

`app/i18n/raw-i18n/translations/pt.json:2147` reads "Verifica a tua ligação e tenta novamente", which is European Portuguese in the tu form, in a file that renders the same English sentence as "Verifique sua conexão e tente novamente" at lines 2429, 2432 and 2671.

### 15. Exports and selected fields with no consumer

`SubmitPlace` is exported and re-exported through the `app/btcmap` barrel but is used only inside its own module. The mutation selects `origin` and `externalId` on `place`, and neither is read anywhere in the app; only the presence of `place` is checked.

designsats (dismissed):
got this error Image

grimen (changes requested):
## Findings

**1. Submission failures are shown behind the native modal (blocking).**

AddPlaceModal uses React Native's native `Modal`, and when submission fails the form stays open with no error shown — the only toast is mounted outside the modal, so it renders behind it.

*Example:* A user submits with a flaky connection, the request fails, and nothing on screen tells them — the form just sits there looking unchanged.

*Proposal:* Render the error toast/message from inside AddPlaceModal instead of relying on the app-level toast mount.

**2. The pre-submit and success copy says every place is reviewed before it appears (blocking).**

`placeReviewNote` and `placeSubmitted` both promise review happens before publication, but trusted-source Blink submissions can appear on BTC Map immediately and get reviewed afterward.

*Example:* A user reads "BTC Map's community surveys every new place before it appears on the map," submits, and the place is already live before any review happened.

*Proposal:* Update both strings (and translations) to describe the actual publication timing, for the implementer to confirm which submissions this affects.

**3. A late response from an abandoned add-place attempt can affect a newer one.**

`handlePlaceSubmit`'s post-request guard only checks `addStep === "describing"`, not whether it's still the same attempt.

*Example:* A user submits Place A, closes the form before the response lands, reopens and reaches the describing step for Place B — Place A's late response then closes Place B's form or shows Place A's error over it.

*Proposal:* Capture the current `addSession`/`submissionId` at send time and check it still matches after the async boundary before applying the result.

**4. Cluster markers stay interactive while positioning the pin.**

`handlePlacePress` is gated on `addStep`, but `handleClusterPress` isn't, so tapping a cluster while aiming the add-place pin still moves and zooms the camera.

*Example:* A user is nudging the pin onto their shop, taps near a cluster, and the camera jumps away from where they were aiming.

*Proposal:* Gate cluster presses with the same `addStep` check used for place presses.

**5. Confirming the pin location can use a stale region.**

`regionRef` only updates in `onRegionChangeComplete`, and the confirm button has no disabled state during camera animation.

*Example:* A user taps "center on me," then immediately taps "Next" before the animation settles — the confirmed coordinates are the previous region, not the one on screen.

*Proposal:* Update the ref during movement or disable confirmation until the camera settles, for the implementer to decide which fits better.

**6. Backend refusal messages bypass i18n.**

`submitPlace` returns the backend's raw English `errors[0].message`, which is shown directly in the toast without going through `LL`.

*Example:* A non-English user gets a rate-limit or duplicate-place refusal in English.

*Proposal:* Map known refusal reasons to localized strings, with a localized generic fallback for unrecognized ones.

## Test Coverage Gaps

- Client-side name validation (`buildPlaceSubmission`) accepts 1–2 char names and up to 120 chars — no test pins these bounds against the backend's actual limits, so a drift like this ships unnoticed.

grimen (dismissed):
## Findings

**1. The pin can still be re-aimed while its send is in flight, so a slow success lands the place at the wrong spot (blocking).**
Change and close stay enabled during `isSubmitting` ([`add-place-modal.tsx:160`](https://github.com/blinkbitcoin/blink-mobile/blob/e679effe51616c5e7c71f2190d446b9abcce49b0/app/components/map-component/add-place-modal.tsx#L160)), and the send carries the coordinates from when Next was tapped; the late-success close at [`index.tsx:415`](https://github.com/blinkbitcoin/blink-mobile/blob/e679effe51616c5e7c71f2190d446b9abcce49b0/app/components/map-component/index.tsx#L415) then ends the attempt while the user is mid-correction.
*Example:* on a slow network the user submits with the pin at A, taps Change and pans to B; when A's send resolves, the toast announces success and the locator closes, but BTC Map has the place at A and nothing says the correction was dropped.
*Proposal:* disabling Change and close while a send is in flight would keep the pin where the send actually goes; if leaving them enabled is intended, the success path could say the place went out at its original spot.

**2. `canAddPlace` makes the whole map re-render on wallet activity.**
The gate at [`index.tsx:100`](https://github.com/blinkbitcoin/blink-mobile/blob/e679effe51616c5e7c71f2190d446b9abcce49b0/app/components/map-component/index.tsx#L100) uses `useIsSelfCustodialAccount`, which goes through `useActiveWallet` and subscribes `MapComponent` to both wallet contexts.
*Example:* a payment arriving while the map tab is open re-runs clustering, marker lists, and label-collision checks for 29k places just to re-decide one button's visibility.
*Proposal:* `useAccountRegistry().activeAccount?.type === AccountType.SelfCustodial` answers the same question without the wallet subscriptions (and still avoids the flash the comment guards against), or the gate plus button could move into a small child component.

esaugomez31 (changes requested):
One blocker from my pass, the rest is minor.

**`app/components/map-component/add-place-modal.tsx:132` — the modal can be closed mid-submit.**

The close button and `onRequestClose` stay live while `isSubmitting`, unlike the change-location control. If the modal is dismissed while the mutation is in flight, the `addStepRef.current !== "describing"` guard in `handlePlaceSubmit` swallows the result: no error surfaces and the typed place is lost. The next attempt mints a fresh `submissionId`, so if that first send did reach the backend we create a duplicate place, which is exactly what the idempotency key is there to prevent.

**Proposed fix:** gate the close button and `onRequestClose` on `isSubmitting`, the same way the change-location control already is.

esaugomez31 (approved):

PR #4188, opened 2026-08-20
fix(transaction-history): stop the unseen-highlight spec racing the render chain

grimen:
## The flake

`TransactionHistoryScreen › highlights none when lastSeen ids are missing` fails intermittently on CI:

```
Expected substring: "no-highlight"
Received string:    "507f1f77bcf86cd799439012:highlight"
```

Seen on [run 32070613689](https://github.com/blinkbitcoin/blink-mobile/actions/runs/32070613689), where all three `jest.retryTimes` attempts failed the same way — the tell that this is a lost race, not a random mock leak.

## Root cause

With no lastSeen ids the screen renders the same rows three different ways in quick succession. Instrumenting the screen locally prints the sequence:

| render | baseline | lastSeen in cache | rows |
|---|---|---|---|
| … | `null` | `""` / `""` | no rows yet |
| 4 | `{btcId:"", usdId:""}` | `""` / `""` | **highlighted** — the mapper's "never seen ⇒ highlight the newest" rule ([`transaction-highlight.ts:26-29`](https://github.com/blinkbitcoin/blink-mobile/blob/8242efe5c/app/custodial/mappers/transaction-highlight.ts#L26-L29)) |
| 5 | `{btcId:"", usdId:""}` | `…011` / `…012` | not highlighted — `markTxSeen` wrote the ids into the cache |
| 6 | `{btcId:"…011", usdId:"…012"}` | `…011` / `…012` | settled |

The spec waited only for the row to *exist* and then asserted **outside** `waitFor`, so it read render 4, 5 or 6 depending on how many effect flushes `waitFor`'s real-timer polling had performed — i.e. on runner load. Locally React batches renders 4-6 inside `render()`'s own `act`, which is why it never reproduces on a laptop.

## The fix

`__tests__/helpers/wait-for-stable-render.ts` — flush effects until the rendered tree stops changing, then assert.

Judgement calls, and what was rejected:

- **`waitFor` on the expected value instead** — rejected. It fixes the common case but not this one: when an intermediate frame carries the same value the test expects (A → B → A), `waitFor` returns on the first frame and never sees the flip. Same for asserting absence: every frame before the element arrives satisfies it. The helper's own spec demonstrates this — `waitFor` returns with the chain provably unfinished.
- **Replacing `waitFor` with the helper** — rejected after it broke the filter tests. "Nothing changed over the last few event-loop turns" cannot distinguish a settled tree from one still waiting on a slow query. The two compose: wait for what must be there, settle, then assert. That rule is in the docblock.
- **Snapshotting the tree with `JSON.stringify` / `pretty-format`** — rejected, both blow up on rendered props (circular context refs; `RangeError: Invalid string length`). The snapshot covers element types, text and *primitive* props only. Callback identities are skipped on purpose: a fresh closure per render is not a change a test can see, and counting it would make every tree "unstable" forever. Both decisions are pinned by tests.
- **Changing the screen or the mapper** — rejected. The transient highlight is existing behaviour, the settled state is already `no-highlight`, and the mapper's never-seen rule is pinned by its own unit spec. This PR only stops the test reading a frame the chain is still mid-way through.
- **Leaning on `jest.retryTimes`** — it stays as a safety net, but it never rescued this one.

The helper is applied to the rest of the spec's transient-sensitive assertions too, including "highlight doesn't flip off after `markTxSeen`", which was expressing the same intent with a duplicated trailing `waitFor`.

## Second commit

`reloadSelfCustodialAccounts` is exposed through the account-registry context, so a consumer can start a second read while the first is still awaiting AsyncStorage; whichever resolved last won, so a stale answer could overwrite fresher entries. Each read now carries a request number and only the newest is applied. Removing the guard makes the new test fail.

I also tried guarding the same read against resolving after unmount, and dropped it: React 19 ignores those updates silently, so the test for it passed with and without the guard. Worth being explicit that this does **not** silence the `act(...)` warnings in the CI logs — those come from updates while the provider is still mounted after a test body returns, which is what #4008's `skipHydration` addresses.

## Verification

- `__tests__/screens`, `__tests__/hooks`, `__tests__/custodial`, `__tests__/self-custodial/mappers`: **340 suites / 4084 tests green**, no retries fired.
- The target spec 10× with `CI=1` under saturating CPU load: 10/10 green, no retries.
- Each new test checked against its own falsification: reverting the guard fails the stale-read test; the `waitFor`-returns-early case is asserted directly.
- `tsc --noEmit` clean (the one `supercluster` error is pre-existing and unrelated), eslint clean.

### DIFF
--- a/__tests__/helpers/wait-for-stable-render.spec.tsx
+++ b/__tests__/helpers/wait-for-stable-render.spec.tsx
@@ -1,0 +1,128 @@
+import React from "react"
+import { Text, View } from "react-native"
+import { render, waitFor } from "@testing-library/react-native"
+
+import { waitForStableRender } from "./wait-for-stable-render"
+
+/**
+ * Renders "A", flips to "B" on the next timer turn, then back to "A" — the
+ * A -> B -> A shape that makes `waitFor` on the final value return too early.
+ */
+const Flipper = ({ flips }: { flips: number }) => {
+  const [value, setValue] = React.useState("A")
+  const [done, setDone] = React.useState(0)
+
+  React.useEffect(() => {
+    if (done >= flips) return undefined
+
+    const timer = setTimeout(() => {
+      setValue((current) => (current === "A" ? "B" : "A"))
+      setDone((current) => current + 1)
+    }, 0)
+
+    return () => clearTimeout(timer)
+  }, [done, flips])
+
+  return (
+    <View>
+      <Text testID="value">{value}</Text>
+      <Text testID="flips-done">{String(done)}</Text>
+    </View>
+  )
+}
+
+/**
+ * Never stops re-rendering, but its structure and text never change — only the
+ * callback identity handed to the child does. That is the churn the snapshot
+ * deliberately ignores; counting it would make this tree "unstable" forever.
+ */
+const ChurningCallback = () => {
+  const [renders, setRenders] = React.useState(0)
+
+  React.useEffect(() => {
+    const timer = setTimeout(() => setRenders((current) => current + 1), 0)
+    return () => clearTimeout(timer)
+  }, [renders])
+
+  return (
+    <View>
+      <Text testID="value" onPress={() => undefined}>
+        A
+      </Text>
+    </View>
+  )
+}
+
+/** Changes nothing but a primitive prop, which the snapshot does track. */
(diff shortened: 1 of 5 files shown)


esaugomez31 (approved):
LGTM!

PR #4196, opened 2026-08-22
fix(settings): add missing custodial account subtitle to custodial banner

designsats:
Closes blinkbitcoin/blink-wip#998

## Problem

The **Custodial account** banner on the settings screen shows the lightning address as a single line, while the equivalent self-custodial banner renders the username plus a `Non-custodial account` subtitle. The custodial banner was missing its `Custodial account` subtitle (see issue screenshot).

## Root cause

`CustodialAccountBanner` in `app/screens/settings-screen/account/banner.tsx` rendered only the username `<Text>` with no subtitle row — there wasn't even a `custodialAccount` i18n key to render.

## Fix

- Add a `SettingsScreen.custodialAccount` key to `app/i18n/en/index.ts` / `raw-i18n/source/en.json` and regenerate all locales via `yarn update-translations`.
- Wrap the username in a `textContainer` view and add a conditional `type="p3"` subtitle (`LL.SettingsScreen.custodialAccount()`) shown when the user is logged in — mirroring the self-custodial banner's layout.
- Username line gets `numberOfLines={1}` + `ellipsizeMode="middle"` so long lightning addresses don't wrap; logged-out state remains the single-line CTA.

### Translation policy

Locales keep an established local term where one exists (e.g. tr "Saklamalı hesap", hu "Letétkezelt fiók", hr "Račun sa skrbništvom", ja カストディアルアカウント). Where crypto media predominantly uses the English loanword or no reference could be found, the string keeps the loanword but translates "account" (e.g. cs "Custodial účet", de "Custodial-Konto", fr "Compte custodial").

Note: this PR includes mechanical re-indentation of the locale JSON files (consistent 4-space indent from the regeneration output); the only semantic change per locale is the new key.

## Test plan

- [x] New spec `__tests__/screens/settings-screen/settings/account-banner.spec.tsx` (32 tests): subtitle shown for logged-in user, hidden when logged out, self-custodial banner unaffected, every locale provides the key
- [x] Revert check: spec fails against pre-fix component
- [x] `locale-parity` + banner specs green (60/60)
- [x] `yarn eslint:check` clean on touched files
- [x] `yarn check:translations` passes
- [x] Verified on Android emulator (Medium_Phone_API_36.1): subtitle renders under the lightning address

Pre-existing failures unrelated to this PR: tsc/jest errors in `app/self-custodial/*` (breez-sdk type drift) also fail on `main`.

## Screenshots

| Before | After |
|---|---|
| image | image |

(Emulator verification; happy to attach device screenshots if useful.)


kngako (dismissed):
No blocking findings.

@designsats will decide if he wants to address test coverage gaps in this PR or in a future/stacked PR.

---

## Findings

### Finding 1 — non-blocking: "Custodial" ships as an English loanword in 12 locales whose sibling label uses a native word

The new `SettingsScreen.custodialAccount` and the existing `SettingsScreen.nonCustodialAccount` render in the same slot of the same banner (`banner.tsx:86` and `banner.tsx:148`) — a user switching accounts sees one, then the other. `2de99c353` deliberately kept "Custodial" untranslated ("keep custodial loanword"), but the already-shipped sibling in these locales does not use the loanword, so the pair no longer reads as opposites:

| locale | existing `nonCustodialAccount` | new `custodialAccount` |
|---|---|---|
| af | Nie-**bewaar**rekening | **Custodial** rekening |
| ca | Compte no **custodi** | Compte **custodial** |
| cs | Ne**svěřenský** účet | **Custodial** účet |
| de | Nicht-**verwahrtes** Konto | **Custodial**-Konto |
| el | Λογαριασμός μη **φύλαξης** | **Custodial** λογαριασμός |
| hy | Ոչ **պահպանող** հաշիվ | **Custodial** հաշիվ |
| lg | Akawunti eta**kuumibwa** | **Custodial** akawunti |
| nl | Niet-**bewaarde** rekening | **Custodial** rekening |
| qu | Mana **waqaychasqa** kuenta | **Custodial** kuenta |
| sk | Ne**svereneský** účet | **Custodial** účet |
| sw | Akaunti isiyo ya **udhamini** | Akaunti ya **custodial** |
| xh | Iakhawunti enga**gcinwa**nga | Iakhawunti ye-**Custodial** |

The other 16 locales are consistent either way (loanword on both sides in es/fr/it/pt/ro/id/ms; native on both sides in ar/da/hr/hu/ja/sr/th/tr/vi).

**When it comes up:** a Dutch user with a custodial and a self-custodial account switches between them on the settings screen. The subtitle flips between "Niet-bewaarde rekening" and "Custodial rekening" — two unrelated words for one concept, so the label that was added specifically so the two account types read as a pair reads as a mistranslation instead. The same happens in German, Czech, Greek and the rest of the table.

**Proposal:** derive each string from its sibling's stem so the pair is the same word with and without the negation — e.g. af `Bewaarrekening`, nl `Bewaarde rekening`, de `Verwahrtes Konto`, cs `Svěřenský účet`, el `Λογαριασμός φύλαξης`, hy `Պահպանող հաշիվ`, ca `Compte custodi`, sw `Akaunti ya udhamini`, xh `Iakhawunti egcinwayo`, qu `Waqaychasqa kuenta`, lg `Akawunti ekuumibwa`. These are mechanical derivations from copy already in the app, not new translations, but a native-speaker glance is still warranted (the Slovak sibling `Nesvereneský` itself looks like a typo for `Nezverenecký`, so fix both together). Edit the lines in place — reserialising the JSONs reflows 22 of them, which is exactly what `163578603` did and the merge had to undo.

### Finding 2 — non-blocking: the one-line cap and middle ellipsis now apply to the logged-out call-to-action

`banner.tsx:81` sets `numberOfLines={1} ellipsizeMode="middle"` on the title `Text` unconditionally. That is right for a lightning address (keep the start and the `@domain`), but the same element renders `logInOrCreateAccount` when logged out — a sentence, not an address — and that branch has no subtitle, so the 70dp row had room for the two lines it could previously wrap into (2 × 24dp line height).

**When it comes up:** the title's horizontal budget is ~285dp on a 360dp phone (12dp screen padding, 4dp row padding, 31dp icon, 12dp gap). The Hungarian CTA "Bejelentkezés vagy fiók létrehozása" is ~255dp at the default 16px, so it fits — but at the OS "Large" text setting (1.15×) or on a 320dp device it overflows and is cut in the *middle*: "Bejelentkezés va…létrehozása". Slovak, Greek and Catalan (32 chars) are one step behind it. Before this branch the sentence simply wrapped.

**Proposal:** apply the cap only to the branch that shows an address — e.g. `numberOfLines={isUserLoggedIn ? 1 : undefined}` and `ellipsizeMode={isUserLoggedIn ? "middle" : undefined}`, or split the two states into two `Text` elements so each carries the props it needs. Alternatively drop the cap entirely for the logged-out state and let it wrap as before; the subtitle only exists when logged in, so the second line is free. Whichever you pick, the spec's Text mock preserves `numberOfLines`, so it can be asserted (see Test Coverage Gap 3).

---

## Test Coverage Gaps

1. **Tap behaviour is untested — `mockNavigationReset` is declared (`account-banner.spec.tsx:65`) and never asserted.** The banner's only interaction is `onPress`: logged-out resets the stack to `getStarted`; logged-in is a no-op. Add a positive test (`fireEvent.press` on the CTA → `reset` called with `{ index: 0, routes: [{ name: "getStarted" }] }`) and a negative one for the logged-in state. The negative assertion is meaningful only because the positive one proves the wiring — keep them adjacent so a future mock change can't silently make both pass.

2. **The "logged-out" test is not the logged-out configuration.** `useIsAuthed` is pinned to `true` for the whole file (`:46`) and only `currentLevel` is flipped to `NonAuth` (`:139`). Real logged-out is token absent → `skip: !isAuthed` → query skipped; the test instead exercises "token present, level not yet in cache", which is a different (transient) state. Make `useIsAuthed` a `mock*` variable and set it `false` in the logged-out test so `skip` is exercised — and add an assertion that `useSettingsScreenQuery` was called with `skip: true`, which the current `() => mockSettingsScreenQuery()` wrapper discards (have the wrapper forward its argument).

3. **"keeps the logged-out state single-line" (`:138`) asserts nothing about lines.** The `mockThemedText` helper renders `Text` as a host element keeping its props precisely so `numberOfLines`/`ellipsizeMode` can be asserted; the test only checks text presence. Either assert the props (which also pins the outcome of Finding 2) or rename the test to what it checks ("hides the custody label when logged out").

4. **The `loading` → `Skeleton` path is untested.** `Skeleton` is mocked to `() => null` (`:72`), so the state is unobservable. Render it as a host element with a testID and add a test that `loading: true` shows the skeleton and neither the title nor the subtitle.

5. **Only level `ONE` is covered for the logged-in custodial state.** A trial account (`ZERO`) is the state where the custody label is arguably most informative. One `it.each` over `["ZERO", "ONE", "TWO", "THREE"]` closes this cheaply (remember `it.each` needs `@jest/globals` in this repo).

---

## Code Smells

1. **The 28-test `custodialAccount i18n coverage` block (`:168-182`) duplicates `__tests__/i18n/locale-parity.spec.ts`.** That suite already asserts every English leaf key exists in every locale (`locale-parity.spec.ts:149`), so these 28 tests cannot fail without it also failing. They also make a component spec read the translation directory from disk. Delete the block; the parity suite is the guard.

2. **`AccountLevel` is hand-copied into the mock (`:54-61`).** The five values mirror `app/graphql/level-context.ts:3-9` by hand. Mock only `useLevel` and spread `jest.requireActual("@app/graphql/level-context")` for the rest, so the test compares against the real enum.

3. **Dead test fixtures.** `mockCopyToClipboard` (`:38`) is never asserted; `selfCustodialEntries: []` (`:153`) is not read by the component; the custodial `beforeEach` sets `mockUseSelfCustodialWallet` (`:118`) even though the custodial banner never calls that hook. Each is harmless but signals the spec was adapted from another rather than written against this component — trim them so the setup documents what the component actually consumes.

4. **Branch history.** Twelve commits for a three-commit change: eight upstream merges, one commit (`f103fe684`) whose content is already on `main`, and `163578603` carrying an 87k-line locale reflow that the merge then reverts. If this lands as a merge commit the repository history keeps the reflow and its reversal. Squash-merge (or rebase onto `main` and drop the reflow before pushing) so the history holds the one-line-per-locale change that actually ships.

5. **The banner body is now duplicated between the two banners.** `iconContainer` + `textContainer` + one-line title + `p3` subtitle appear verbatim in `CustodialAccountBanner` (`:77-89`) and `SelfCustodialAccountBanner` (`:140-150`). Two copies is tolerable; a third would not be. Optional: extract a local `BannerBody({ title, subtitle })` in the same file. This is plain de-duplication, not a replacement for a Galoy atomic component.

designsats:
Addressed Finding 1 (custodial / non-custodial label pairing) from [the review](https://github.com/blinkbitcoin/blink-mobile/pull/4196#pullrequestreview-5202229656) in b0f19308d9 and 6a941f5a7f. Lines were edited in place, no JSON reserialisation.

`custodialAccount` now uses the same stem as its existing `nonCustodialAccount` sibling in 11 locales:

| locale | `nonCustodialAccount` | `custodialAccount` |
|---|---|---|
| af | Nie-bewaarrekening | Bewaarrekening |
| ca | Compte no custodi | Compte custodi |
| cs | Nesvěřenský účet | Svěřenský účet |
| de | Nicht-verwahrtes Konto | Verwahrtes Konto |
| el | Λογαριασμός μη φύλαξης | Λογαριασμός φύλαξης |
| hy | Ոչ պահպանող հաշիվ | Պահպանող հաշիվ |
| lg | Akawunti etakuumibwa | Akawunti ekuumibwa |
| nl | Niet-bewaarde rekening | Bewaarde rekening |
| qu | Mana waqaychasqa kuenta | Waqaychasqa kuenta |
| sw | Akaunti isiyo ya udhamini | Akaunti ya udhamini |
| xh | Iakhawunti engagcinwanga | Iakhawunti egcinwayo |

**Slovak** goes the other way: both labels use the loanword, matching the "non-custodial účet" wording the Slovak migration copy already uses.

| locale | `nonCustodialAccount` | `custodialAccount` |
|---|---|---|
| sk | ~~Nesvereneský účet~~ → Non-custodial účet | Custodial účet |

The misspelt `NonCustodialInfoBulletin.title` was changed to "Toto je non-custodial účet" too.

These are derived from copy already in the app rather than fresh translations, so a native-speaker check is still welcome.

PR #4197, opened 2026-08-23
fix(migration): restore the home entry into the guided migration once receiving is disabled

### LINKED ISSUE #4195: bug(migration): persistent Migrate bulletin disappears once the region is RECEIVE_DISABLED, leaving no visible entry on the home dashboard
## Summary

On 3.0.25 (current store release) a custodial user whose region has reached `RECEIVE_DISABLED` no longer sees the persistent **"Important — Migrate"** bulletin on the home dashboard. The only home-screen entry into the migration becomes the session-dismissible **"Migrate now"** modal; once the user closes it, nothing on the dashboard says "Migrate" anymore. Users are reporting "the banner with the Migrate button is gone" and not finding their way into the flow.

This is app logic, not a backend change: the backend returns `windDown.status = RECEIVE_DISABLED`, and the app deliberately hides the bulletin for that status.

## Where it happens

- `app/screens/account-migration/hooks/use-migration-reminder-bulletin.ts` — `isVisible: windDown?.status === WindDownStatus.PreCutoff` (docstring: *"once receiving is disabled the migrate-now modal takes over"*).
- `app/screens/account-migration/hooks/use-migrate-now-prompt.ts` — modal `isVisible = RECEIVE_DISABLED && !killSwitch && !isDismissedForSession`; dismissal is per-session state, and the modal also waits while any other home modal (upgrade / restriction / stablesats / convert) is up.
- `app/screens/home-screen/home-screen.tsx` — renders `<MigrationReminderBulletin>` only from the hook above (`:893`), `<MigrateNowModal>` at `:782`.

Unchanged in tags 3.0.26–3.0.30, so the next release behaves the same.

## Why it matters now

All wind-down regions except `wave2` (TR PK DZ NP) are armed on prod (`default` EU/EEA since 2026-08-21, `us` since 2026-08-19, `cn-mm` since 2026-08-10), so this is the state the large majority of affected users are in for the rest of the migration window (final deadline 2026-08-31). The integration contract also describes `RECEIVE_DISABLED` as *"Receiving-off banner, Receive greyed out, Send/Scan/Migrate available"* — the shipped app only has the modal in that phase.

## Current entry points in `RECEIVE_DISABLED` (for reference)

1. "Migrate now" modal on each cold start (dismissible with X).
2. Tapping the greyed-out Receive button reopens the modal (not discoverable).
3. Settings → "Move to non-custodial".
4. Deep link `account-migration`.

## Expected

The persistent home bulletin keeps showing in `RECEIVE_DISABLED` (in addition to the modal), with phase-appropriate copy, e.g. *"Receiving has stopped. Move to a non-custodial account via guided migration, or withdraw, before {date}."* and the **Migrate** CTA, so there is always a visible path into the flow from the dashboard. Non-dismissible, same as today's pre-cutoff bulletin.

## Suggested change

- `use-migration-reminder-bulletin.ts`: `isVisible` = `PRE_CUTOFF || RECEIVE_DISABLED`; expose the phase so the component can pick the copy variant.
- `migration-reminder-bulletin.tsx` + `AccountMigration.reminderBulletin` i18n: add the receive-disabled body string (the current body says "Receiving stops {receiveStopsDate}", which reads wrong once it already has).
- Keep the existing `GATED_CLOSED` behaviour (blocker takes the whole home) and the offboard-only replacement rule in `use-wind-down-home-nudges.ts` as is.

## Acceptance

- [ ] Custodial account with `windDown.status = RECEIVE_DISABLED`: home shows the non-dismissible bulletin with Migrate CTA after the "Migrate now" modal is closed; CTA routes to `accountMigrationEntry`.
- [ ] `PRE_CUTOFF` bulletin unchanged; `GATED_CLOSED` unchanged; `windDown == null` shows no wind-down UI.
- [ ] Offboard-only bulletin still replaces (not stacks with) the reminder.
- [ ] Kill-switch (`nonCustodialEnabled` off) behaviour unchanged for the bulletin (routes through the entry dispatcher, which handles the disabled case).

esaugomez31:
Closes #4195

Once a region reaches `RECEIVE_DISABLED`, the home dashboard loses every visible way into the guided migration. This restores both of them: the persistent bulletin, and the two greyed surfaces users actually tap when looking for an explanation.

## Part 1 — the bulletin disappears in `RECEIVE_DISABLED`

`useMigrationReminderBulletin` gated visibility on `PRE_CUTOFF` only, so the moment the backend flips a region's `receiveDisabled` flag the card is gone. The only remaining home entry is the "Migrate now" modal, which is dismissible for the session: once the user closes it with the X, nothing on the dashboard mentions the migration until the next cold start. In sessions where another modal or the pending region check outranks it, the modal never shows at all and the dashboard is silent from the first frame.

The bulletin now spans both phases where the migration is still reachable from the dashboard. `GATED_CLOSED` is untouched: the blocker replaces the whole navigator there, so there is no dashboard left to hold a card.

The copy needed a second variant. The existing body ends with *"Receiving stops {date}"*, which past the cutoff announces a past event in the future tense. The hook now reports its phase and the component picks between:

- `PRE_CUTOFF` — unchanged.
- `RECEIVE_DISABLED` — *"Receiving has stopped. Move to a non-custodial account via guided migration, or withdraw, before {date}."*

New key `AccountMigration.reminderBulletin.bodyReceiveDisabled`, translated into all 28 languages.

## Part 2 — the greyed surfaces open the wrong modal

Reported in [this comment](https://github.com/blinkbitcoin/blink-mobile/issues/4195#issuecomment-5381422512). The disabled **Transfer** button and the greyed dollar row ("not available in your region") share one handler, `onGatedDollarTap`, which opened the Dollar Balance restriction modal: a title, a Close button, and nothing else. For an account whose custodial service is ending, that is a dead end on the two surfaces a user hunting for the migration will try first.

The handler now reopens the migrate-now nudge before falling back to that modal. Priority is `Anon > sanctions > wind-down nudge > region explanation`. The new branch is gated on `canReopen`, which is exactly "the migrate-now nudge can surface", so:

- A region-gated account with **no** wind-down keeps the existing restriction modal.
- An account whose self-custodial stack is switched off keeps it too.
- `PRE_CUTOFF` keeps it as well: there is no migration modal to show in that phase, and the bulletin with its Migrate CTA is already on screen.

## Before / after

Both recordings walk the same path: cold start, close the "Migrate now" modal, look for a way back into the migration on the dashboard, then tap the disabled Transfer button and the greyed dollar row.

### Before

https://github.com/user-attachments/assets/ed67b504-ef45-4f66-9a74-a2ecdad66c2b

### After

https://github.com/user-attachments/assets/c7399b95-5c20-451f-af3b-82e09a1041fd

## Testing

- `tsc --noEmit` and eslint clean.
- 57 suites / 943 tests green, including `home.spec.tsx`, the reminder-bulletin hook and component specs, `wallet-overview.spec.tsx` and `use-account-restrictions.spec.ts`.
- Each commit verified green on its own, not just the branch tip.
- Device-tested on Android against a custodial account with the wind-down state and the region verdict stubbed locally (stubs are not part of this branch).

New coverage: bulletin visible and reporting its phase in `RECEIVE_DISABLED`; the receive-disabled body replacing the pre-cutoff one; home forwarding the phase; the gated tap reopening the nudge; the fallback to the restriction modal when no nudge can surface; sanctions still outranking the nudge.

## Out of scope

In an offboard-only region during `RECEIVE_DISABLED`, the gated tap now opens a modal whose CTA leads to a flow those accounts cannot complete. This is not introduced here: the greyed Receive button already reopens the same prompt with no offboard check, and the modal auto-shows each session regardless. Worth a product decision on whether the offboard-only rule should extend to the modal as it does to the bulletin.

The wind-down phase comes from backend flags, not from the dates it serves, so `receiveDisabledAt` can name a day that has already passed while the phase is still `PRE_CUTOFF`. Nothing the app can fix, but it is what makes the old copy read wrong in the first place.


kngako (dismissed):
Kimi K3 review. Not if the gated dollar tap bypass is intentional or what but if it is then it is the only blocking change.
 
---

## Findings

### Finding 1 — Gated dollar tap bypasses the offboard-only suppression (BLOCKING)

`useWindDownHomeNudges` deliberately suppresses the reminder bulletin for offboard-only
regions because "offboard-only regions cannot open any account, so the reminder's Migrate
CTA is a dead end" (`use-wind-down-home-nudges.ts:45-47`). The new `onGatedDollarTap`
branch (`home-screen.tsx:586-589`) reuses `migrateNowPrompt.canReopen`, which is only
`isReceiveDisabled && !isSelfCustodialDisabled` (`use-migrate-now-prompt.ts:30`) — it
never consults the offboard-only state. The dead end the masking exists to prevent is
therefore reintroduced through the new surface.

**When it comes up:** an offboard-only-region custodial account (phone country in the
Remote Config `offboardOnlyCountries` list) is in the `ReceiveDisabled` wind-down phase
with the self-custodial stack enabled, and its dollar balance is also region-gated. The
user taps the greyed dollar row or the disabled transfer button hunting for a way out,
and gets the migrate-now prompt whose CTA routes into a guided migration their region
cannot complete — while the bulletin on the same dashboard was hidden precisely to avoid
that contradictory instruction.

**Proposal:** mask the reopen with the offboard state in the same module that already
owns the masking: in `useWindDownHomeNudges`, expose the prompt's `canReopen` already
ANDed with `!offboardBulletin.isVisible` (mirroring how `reminderBulletin.isVisible` is
masked), so `home-screen.tsx` keeps reading a single nudge bundle. Add a home-level test:
offboard bulletin visible + `canReopen` true + gated dollar tap → compliance modal shows,
`reopen` not called. Note the migrate-now modal also auto-fires for offboard-only
accounts each session (`use-migrate-now-prompt.ts:34`) — pre-existing and out of this
PR's scope, but the same mask would fix both; worth a follow-up ticket if not done here.

### Finding 2 — Bulletin CTA dead-ends in ReceiveDisabled when the self-custodial stack is off (non-blocking, needs a decision)

The reminder bulletin intentionally skips the self-custodial kill-switch
(`use-migration-reminder-bulletin.ts:21-25`), routing the CTA through the migration entry
dispatcher, which shows the gate's "temporarily unavailable" screen when the stack is
disabled. That rationale was written for `PreCutoff`, where the account works normally
and the bulletin is informational; in `ReceiveDisabled` the same flag-off account gets no
migrate-now modal (`canReopen` false), a toast on the Receive tap, and now a permanent
non-dismissible bulletin whose only button leads to "temporarily unavailable" — three
different failure answers on one screen.

**When it comes up:** ops flips the self-custodial kill-switch off (SDK incident) after
August 1. A custodial user in a wind-down region opens the app: receiving is already off,
the modal can't surface, and the one prominent action on their dashboard — the bulletin's
Migrate button — walks them into a "temporarily unavailable" screen every time they tap
it, with no way to dismiss the card.

**Proposal:** either accept and pin it (a hook-spec test: `ReceiveDisabled` +
kill-switch off → bulletin still visible, with a comment stating this is deliberate), or
hide the bulletin when `useSelfCustodialDisabled()` is true in the `ReceiveDisabled`
phase only, keeping the documented `PreCutoff` behavior. If it stays visible, consider
copy that acknowledges the temporary unavailability rather than promising a migration the
tap cannot start. Cheapest correct step is the pinning test — the behavior is currently
undocumented-by-test either way.

---

## Test Coverage Gaps

- **Gap 1 — Offboard × gated-tap interplay (ties to Finding 1).** No test combines
  `offboardBulletin.isVisible` with a gated dollar tap. `use-wind-down-home-nudges.spec.ts`
  covers the bulletin masking (`:59`) but nothing pins that the reopen path respects the
  same suppression. Reverting the proposed mask would leave every test green.
- **Gap 2 — Kill-switch × ReceiveDisabled bulletin (ties to Finding 2).** The hook spec
  mocks wind-down status but never the self-custodial flag, so the bulletin's visibility
  under `ReceiveDisabled` + disabled stack is unpinned in both directions.
- **Gap 3 — Hidden-state `phase` fallback.** The hook documents "reports the pre-cutoff
  phase while hidden" (`use-migration-reminder-bulletin.ts:32`), but no spec asserts
  `phase === PreCutoff` when `isVisible` is false — a decided-but-implicit behavior left
  accidental.
- **Gap 4 — Both gated surfaces.** The new home tests exercise the disabled transfer
  button and note the dollar row "shares this exact callback (`onGatedTap`), whose wiring
  wallet-overview.spec covers" (`home.spec.tsx:2196-2199`) — verified, that spec does pin
  the wiring (`wallet-overview.spec.tsx:194-205`). No gap here; recorded for completeness.

## Code Smells

- **Smell 1 — `phase` is meaningful only when visible, enforced by comment.** The hook
  returns `phase: PreCutoff` as a sentinel while hidden (`use-migration-reminder-bulletin.ts:32-35`).
  A discriminated union (`{ isVisible: false } | { isVisible: true; phase; ... }`) would
  make the invariant structural instead of a comment a future consumer can miss. Low
  priority — the repo's documented-invariant convention tolerates this.
- **Smell 2 — `receiveDisabledTimestamp` is dead data in one of the two phases.** The
  component still requires the prop and computes `receiveStopsDate` unconditionally
  (`migration-reminder-bulletin.tsx:44-47`), but the `ReceiveDisabled` body never reads
  it. Computing it inside the pre-cutoff branch (or splitting props per phase) would keep
  the render honest.
- **Smell 3 — Phase semantics encoded in three modules.** `isReceiveDisabled` /
  `isPreCutoff` derivations from `windDown.status` now live in `use-migrate-now-prompt.ts`,
  `use-migration-reminder-bulletin.ts`, and `use-wind-down-receive-blocked.ts`. Fine at
  three copies; if a fourth appears, extract a shared phase helper per the repo's
  DRY-with-a-threshold rule.
- **Smell 4 — Pointless async wrapper.** `const handleMigrate = async () => onMigrate()`
  (`migration-reminder-bulletin.tsx:51`) wraps a sync callback in a promise for no
  reason. Pre-existing, one-line cleanup if the file is touched again.

grimen (approved):
## Decisions

**1. Offboard-only migrate-now dead end — not fixed here.**
Kimi's review flagged that an offboard-only-region account could reach the migrate-now prompt via the gated dollar tap, landing on a CTA into a migration that account can't complete. Tracing the actual migration flow (`migration-entry-screen.tsx`, `migration-gate.tsx`) found no code that gates on offboard-only status anywhere in it, so the dead end isn't confirmed to exist.

*Alternative considered:* AND the offboard signal into `canReopen` in `use-wind-down-home-nudges.ts`, mirroring the bulletin's existing masking.

*Why rejected:* would suppress the one remaining entry point into migration for offboard-only accounts without evidence the flow actually fails for them — the opposite of what we want. Discussed and confirmed with the CEO as out of scope for this PR.

**2. Bulletin CTA routes to "temporarily unavailable" when the kill-switch is on — pre-existing, not introduced here.**
When the self-custodial stack is disabled, the reminder bulletin's CTA still navigates into `accountMigrationEntry` and lands on `TemporarilyUnavailableScreen`, while the greyed Receive/dollar taps short-circuit straight to a toast instead. This wiring predates the PR — it was already true for the `PreCutoff` bulletin — this PR only extends the bulletin's visibility into `ReceiveDisabled`, so it surfaces more often.

*Follow-up:* worth aligning the bulletin CTA with the toast pattern used elsewhere, as a fast follow rather than in this PR.

PR #4199, opened 2026-08-24
revert: escalating time lockout for PIN attempts (#4162)

grimen:
Reverts #4162.

## Why

The "Works as intended" approval on #4162 was hasty and wrong. On a device with more than one saved account, the third-strike "logout" this PR adds doesn't actually lock the attacker out — it runs the generic `logout()` → `resetState()` path, which lands on whichever *other* saved session is next. If that session is unprotected (e.g. the app's local anonymous non-custodial wallet, or another account that never had a PIN set), the attacker is inside a real, funded wallet within the lockout's own 3-strike + wait budget (~70s in the repro), having never entered a correct PIN.

Screen recording of the repro: three wrong PINs on a PIN-protected account, waiting out the 30s and 60s lockouts as designed, the third strike triggers "Too many failed attempts. Logging out.", and two seconds later the app is sitting on the Home screen of an "Anon user" wallet with a live balance and full Transfer/Receive/Send/Scan access.

This isn't a flaw in the lockout schedule (the 30s/60s escalation, persistence-before-feedback, and clock-rollback clamp all behaved correctly) — it's that the PIN gate was only ever designed to protect the currently-active account's screen, not to guarantee that the app's normal logout/account-switch machinery can't surface a different, unguarded session. That's a structurally different problem than what #4162 set out to fix, so scoping a fix on top of the current PR risks leaving other paths into an unprotected session (deep links, background/foreground resume, biometric fallback) with the same hole.

## What this does

Straight `git revert` of the squash-merged commit (8e01f5597). No files touched by #4162 were modified by later commits on `main`, so the revert applies cleanly. `tsc --noEmit` shows only pre-existing, unrelated errors (self-custodial bridge/breez-sdk types, missing `supercluster` types); the `authentication-screen`/`security-screen` suites (29 tests) pass after the revert with no dangling references to the removed modules.

## Follow-up

Re-opens blinkbitcoin/blink-wip#1140, moved back to Review. A new draft PR will scope the actual fix: logout from a locked PIN screen must be provably terminal — it must not resolve onto any session that is itself unguarded — built test-first against the multi-account bypass this PR closes.


esaugomez31:
Force-pushed both branches in this stack; the original commits were replaced. Rationale and context in blinkbitcoin/blink-wip#1140.

PR #4204, opened 2026-08-24
fix(migration): surface a recoverable error when the commit screen prerequisites never settle

### LINKED ISSUE #4189: bug(migration): post-reset recurrence leaves Confirm disabled and resume state missing
## Summary

Post-reset recurrence of the custodial → non-custodial migration failure documented in #4070.

Support case `BD-00025` received a backend migration-state reset on 2026-08-19. The customer retried on 2026-08-20 using Android app `3.0.25.50000870`, but again reached the move-funds/balance-review screen with **Confirm disabled**. The app then closed/reset out of the flow.

This follow-up intentionally omits customer identifiers and balances. Support can provide the account UUID privately to the assigned engineers.

## Environment

- Device: Samsung Galaxy M11
- OS: Android 12
- App: `3.0.25.50000870`
- Account state after the retry: ACTIVE custodial; not migrated
- Related original report: #4070
- Related app fixes: #4087 and #4088

## Observed behavior

1. Engineering completed the backend migration reset.
2. Customer fully closed and reopened Blink, then restarted migration.
3. Customer progressed through the flow to the move-funds/balance-review screen.
4. The Confirm button was visible but remained disabled.
5. The app then closed/reset out of the flow.
6. A subsequent backend check showed no completed migration.

## Telemetry from the retry

In the 12-hour investigation window:

- 6 `startMigrationFlow` calls
- 4 successful migration-preview reads
- 18 `resumeMigrationFlow` calls
- 5 recent `app.migrationflow.resumeMigrationFlow` failures with:

```text
CouldNotFindMigrationFlowStateError
```

No `migrationCommit` or completed migration was observed. No migration Lightning-address-transfer execution was observed for the retry.

Two `LnPaymentPendingError` events in the same account window were verified as unrelated ordinary `lnInvoicePaymentSend` attempts and should be excluded from the migration diagnosis.

## Relevant client condition

`balances-overview-screen.tsx` keeps Approve/Confirm disabled until all of the following are true:

- preview is ready;
- server migration start is confirmed;
- Lightning-address transfer succeeds.

The retry reached preview/start activity, but the flow did not reach a stable commit-ready state. The repeated `CouldNotFindMigrationFlowStateError` is the strongest observed failure signal.

## Expected behavior

After a backend reset, restarting migration should produce a stable resumable flow and either:

- enable Confirm once all prerequisites complete; or
- display a specific recoverable error with a safe retry/support path.

## Actual behavior

The flow appears to lose or fail to recover its server-side state after restart. Confirm remains disabled without a customer-actionable explanation, and the app exits/resets from the migration flow.

## Requested investigation

- Determine why `resumeMigrationFlow` cannot find state after the backend reset and successful start/preview activity.
- Confirm whether the reset leaves stale client/server migration identifiers or creates a new flow that later resume calls cannot resolve.
- Ensure the balance-review screen surfaces a recoverable error instead of leaving Confirm indefinitely disabled.
- Define whether another retry is safe before support asks the customer to proceed again.
- Add regression coverage for: backend reset → app restart → migration preview → Lightning-address transfer → commit.

## Current support handling

Support has told the customer not to retry, reinstall, or log out while this recurrence is investigated. The account remains custodial and the migration did not complete.

esaugomez31:
Relates to #4189.

## Problem

The migration commit screen (`balances-overview-screen.tsx`) enables Approve only once four prerequisites are satisfied. Two of them could stay unsatisfied forever without reporting anything:

- The lightning-address re-point builds its proof through the SDK, which connects and signs under a per-storage-dir lock and can stall without ever throwing. An attempt that never answers sets no flag at all.
- The re-point returns early and silently when either account id is null, and the custodial owner id is null whenever its query answers without one.

In both cases the user is left on a screen showing their balances with a permanently disabled Approve, no message, no retry, and the hardware back swallowed. That is the state reported in #4189, where telemetry shows no lightning-address-transfer execution for the retry.

Reproduced on an Android emulator by forcing the owner id to never arrive: balances rendered, Approve grey, taps did nothing.

## Changes

**Bound the re-point** (`use-migration-ln-address-transfer.ts`). An attempt that does not answer within 45s settles as a connection issue, which the screen already renders as a message plus a retry. A throw raised before the mutation (the keychain read sits outside `run`'s own try) is reported and settled apart from a stall, so support is never told an attempt stalled when it threw. An attempt whose ids flicker null mid-flight no longer has its answer dropped: the effect re-runs but cannot fire again against the claimed attempt number, so discarding the answer already in the air left every flag false for good.

**Report why the owner is unknown** (`use-custodial-owner-id.ts`). The hook now distinguishes a query that never ran (`isSkipped`) from one that failed (`hasError`) from one that answered, and exposes a `refetch` that stays inert while skipped. A failure is deliberately not split into network and server kinds: an authenticated custodial session always has an owner, so every error is a request that failed rather than an answer.

**Give every outcome a destination** (`balances-overview-screen.tsx`, `types/migration.ts`). An owner query that answered with no owner hands over to support under a new `custodial-owner-missing` reason. Everything else offers the retry, which now also refetches the owner and the checkpoint.

The rule the screen follows: hand over only on an answer, retry on anything else. Contact support stays on screen throughout, so no path is a dead end.

## Deliberately not a handover

A missing provisioned account offers the retry rather than support. The checkpoint keys its id by an owner it resolves through its own instance of an uncached query, which can fail while this screen's instance is healthy, so a null read there is not proof that no account was provisioned. A retry costs a user who really has none a single press; a handover would cost one whose lookup merely dropped their migration, at the point of no return.

The id sources also drop out of `isRetryable` once the re-point has completed. Past that point they hold nothing back, and the retry replaces Approve on this screen, so offering it would hide a commit the user is ready to make.

## Testing

- 805 tests across 54 suites, typecheck and lint clean.
- 16 new tests covering the stall, the throw, the flickering ids, the skipped and failed owner query, the unreadable checkpoint, and the retry wiring.
- The three new guards were mutation-tested: removing each one by hand fails tests.
- Device: the handover was verified on an emulator, reaching the support screen with the reason, account id, username and phone filled in. The 45s path is covered by tests only.

## Not covered here

The backend half of #4189 stays open: why `resumeMigrationFlow` cannot find its state after a reset, and whether another retry is safe. One client-side answer the ticket asks for: the app holds no migration flow identifier. `migrationStart` takes no arguments, the re-point and the commit send only a proof of possession, and the checkpoint stores the step, the provisioned account and the expected receive figure. Stale client identifiers can be ruled out.

Full regression coverage for reset, restart, preview, transfer, commit needs a backend harness that does not exist on the app side.

### DIFF
--- a/__tests__/screens/account-migration/hooks/use-custodial-owner-id.spec.ts
+++ b/__tests__/screens/account-migration/hooks/use-custodial-owner-id.spec.ts
@@ -5,7 +5,8 @@ import { AccountType } from "@app/types/wallet"
 
 let mockIsAuthed = true
 let mockActiveAccount: { type: string } | undefined
-let mockQueryResult: { data: unknown; loading: boolean }
+let mockQueryResult: { data: unknown; loading: boolean; error?: unknown }
+const mockRefetch = jest.fn()
 const mockUseMigrationOwnerQuery = jest.fn()
 
 jest.mock("@app/graphql/is-authed-context", () => ({
@@ -26,7 +27,10 @@ describe("useCustodialOwnerId", () => {
     mockIsAuthed = true
     mockActiveAccount = { type: AccountType.Custodial }
     mockQueryResult = { data: undefined, loading: false }
-    mockUseMigrationOwnerQuery.mockImplementation(() => mockQueryResult)
+    mockUseMigrationOwnerQuery.mockImplementation(() => ({
+      ...mockQueryResult,
+      refetch: mockRefetch,
+    }))
   })
 
   it("returns the Galoy account id for a custodial session", () => {
@@ -42,6 +46,7 @@ describe("useCustodialOwnerId", () => {
     expect(mockUseMigrationOwnerQuery).toHaveBeenCalledWith({
       skip: false,
       fetchPolicy: "no-cache",
+      notifyOnNetworkStatusChange: true,
     })
   })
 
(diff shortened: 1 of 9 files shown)


grimen (changes requested):
## Findings

**1. Retry can never fix the failure it's shown for (blocking).**

When the checkpoint's own internal owner query fails (`useMigrationCheckpointState`'s private `useCustodialOwnerId()` call) while this screen's separate `useCustodialOwnerId()` instance succeeds, `selfCustodialAccountId` settles to null and `isProvisionedAccountMissing` shows the retry. `handleRetry` calls `refetchOwnerId()` and `refetchCheckpoint()`, but neither reaches the checkpoint's internal owner query — `refetchCheckpoint` only reloads AsyncStorage, and the internal query's `refetch` is never exposed by `useMigrationCheckpointState`.

*Example:*
A transient failure hits only the checkpoint's uncached owner query; the user taps retry indefinitely and `selfCustodialAccountId` stays null forever, since nothing re-runs that query short of remounting the screen.

*Proposal:*
One option could be exposing the internal owner query's `refetch`/`hasError` from `useMigrationCheckpointState` so this screen's retry can target it directly, or sharing one owner-query result between the two hooks instead of running it twice.

## Test Coverage Gaps

- No spec exists for `balances-overview-screen.tsx`, so the retry path added here (`isProvisionedAccountMissing`, `isIdSourceRetryable`, `handleRetry`) has no coverage at all, including this scenario where the checkpoint's internal owner query fails independently of the screen's own.

grimen (approved):

PR #1161, opened 2023-02-16
chore: fonts cleanup

nicolasburtey:
before:
-rw-r--r--@ 1 n  staff  41977671 Feb 16 15:15 app-arm64-v8a-release.apk
-rw-r--r--@ 1 n  staff  35179554 Feb 16 15:15 app-armeabi-v7a-release.apk
-rw-r--r--@ 1 n  staff  99688871 Feb 16 15:15 app-universal-release.apk
-rw-r--r--@ 1 n  staff  43504625 Feb 16 15:15 app-x86-release.apk
-rw-r--r--@ 1 n  staff  43940437 Feb 16 15:15 app-x86_64-release.apk
-rw-r--r--@ 1 n  staff      1484 Feb 16 15:15 output-metadata.json

after (first commit)
-rw-r--r--@ 1 n  staff  38779778 Feb 16 15:17 app-arm64-v8a-release.apk
-rw-r--r--@ 1 n  staff  31981661 Feb 16 15:17 app-armeabi-v7a-release.apk
-rw-r--r--@ 1 n  staff  96490911 Feb 16 15:17 app-universal-release.apk
-rw-r--r--@ 1 n  staff  40306730 Feb 16 15:17 app-x86-release.apk
-rw-r--r--@ 1 n  staff  40742526 Feb 16 15:17 app-x86_64-release.apk
-rw-r--r--@ 1 n  staff      1484 Feb 16 15:17 output-metadata.json

3rd commits:
-rw-r--r--@ 1 n  staff  36956227 Feb 16 18:38 app-arm64-v8a-release.apk
-rw-r--r--@ 1 n  staff  30158105 Feb 16 18:38 app-armeabi-v7a-release.apk
-rw-r--r--@ 1 n  staff  94667311 Feb 16 18:38 app-universal-release.apk
-rw-r--r--@ 1 n  staff  38483186 Feb 16 18:38 app-x86-release.apk
-rw-r--r--@ 1 n  staff  38918986 Feb 16 18:38 app-x86_64-release.apk
-rw-r--r--@ 1 n  staff      1484 Feb 16 18:38 output-metadata.json

PR #4210, opened 2026-08-25
feat(security): add a keychain adapter and read-through migration helper

esaugomez31:
Fixes [blinkbitcoin/blink-wip#1160](https://github.com/blinkbitcoin/blink-wip/issues/1160) (close manually on merge — cross-repo `Fixes` doesn't auto-close).

Step 1 of 4 of [blinkbitcoin/blink-wip#1143](https://github.com/blinkbitcoin/blink-wip/issues/1143), itself item 5 of the [#1092](https://github.com/blinkbitcoin/blink-wip/issues/1092) red-team report.

## Problem

`react-native-secure-key-store` (`^2.0.9`, unmaintained since ~2019) holds every credential the app owns: the PIN, its lockout state, the biometrics flag, `sessionProfiles`, `galoyAuthToken` and the per-account BIP-39 mnemonics. Nothing is broken today. The risk is the ordinary one — an unmaintained native dependency holding funds-bearing key material eventually breaks on some OS release, and it gets fixed under pressure with users locked out.

Moving those slots is a funds-at-risk data migration, so this lands the storage adapter and the read-through helper **without moving a single slot**. `secureStorage.ts` is untouched and [`__tests__/utils/secure-storage.spec.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/main/__tests__/utils/secure-storage.spec.ts) needs zero changes; that is the acceptance criterion for "no behaviour change", and both hold.

One constraint shapes everything else. `clearSecureKeyStore` (`ios/RNSecureKeyStore.m` l.118) issues an **unscoped** `SecItemDelete` over `kSecClassGenericPassword` and `kSecClassKey` — every such item the app owns, not just the library's own. It runs from `handleAppUninstallation`, the first statement of `get:` and `set:`, and fires whenever `NSUserDefaults` key `RnSksIsAppInstalled` is false, i.e. the first call after any reinstall, since `NSUserDefaults` is wiped by an uninstall and the Keychain is not.

Harmless today, because the app owns no generic passwords. **Lethal after the migration** if the new store used them: on a reinstall boot, any slot that misses in the new store falls through to the legacy library, which then deletes every migrated mnemonic. The read-through design is itself what pulls the trigger. `kSecClassInternetPassword` is not in that class list, which is why the app's existing Keychain usage is safe today ([`use-credential-backup.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/screens/self-custodial/onboarding/hooks/use-credential-backup.ts), [`use-create-device-account.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/screens/get-started-screen/use-create-device-account.ts)) and why the new store stores its items the same way.

## What changed

Three new modules under `app/utils/storage/`, plus the test scaffolding they need.

**[`secure-store.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/feat--keychain-adapter-1160/app/utils/storage/secure-store.ts)** — the Keychain adapter. `serverFor`, `secureRead`, `secureExists`, `secureWrite`, `secureRemove`, over **internet credentials with a distinct `server` per slot** under `secure-store.blink.local/<slot>`. The secret goes in `password`; `username` carries the slot name so items are self-describing in a Keychain dump. `false ⇒ absent, throw ⇒ failed` holds on both platforms: `getInternetCredentials` resolves `false` only for `errSecItemNotFound` and rejects on every other `OSStatus`, and Android resolves `false` only when the prefs entry is missing, rejecting `E_CRYPTO_FAILED` / `E_KEYSTORE_ACCESS_ERROR`. `secureExists` goes through `hasInternetCredentials`, which resolves `YES` even on `errSecInteractionNotAllowed`, so a probe answers **before first unlock**. The write options object carries `accessible` and nothing else.

**[`legacy-key-store.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/feat--keychain-adapter-1160/app/utils/storage/legacy-key-store.ts)** — the legacy read and erase primitives, owning the `404`-vs-other-code classification and the `setResetOnAppUninstallTo(false)` guard. A `no-restricted-imports` rule in `.eslintrc.json` keeps new call paths from reaching the library directly; `secureStorage.ts` holds the only other exception, and it goes away when its slots move in [#1161](https://github.com/blinkbitcoin/blink-wip/issues/1161). The module deliberately has no write primitive — migrated values are only ever written to the new store.

**[`secure-store-migration.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/feat--keychain-adapter-1160/app/utils/storage/secure-store-migration.ts)** — the lazy per-slot migration.

- `readThrough` — new store first. `found` returns immediately and never touches the legacy library, which is what keeps a migrated user off it entirely. `failed` returns `failed` without falling back. Only when both stores confirm a miss does `absent` come back. A legacy hit is written to the new store, then the legacy copy is erased; a failed write leaves the legacy copy intact and retries on the next read.
- `removeThrough` — deletes from both stores, legacy copy first. A new-store-only remove is not a remove: the next read would miss, fall through and write the old value straight back.
- `existsThrough` — the probe first, since it is the only one that answers before first unlock, falling through to a full read when it comes back `no`, so an unmigrated slot is never reported as unset.
- A per-slot queue serializes everything that can move a value between the stores, and a 30s deadline keeps one hung native call from holding a slot's queue open for the rest of the process.
- A legacy-hit counter, `legacy_key_store_hit`, tagged by key class only. `mnemonic:<accountId>` is cut at the first separator, so an account id cannot reach telemetry, Crashlytics or an error message.

**Test scaffolding.** `__mocks__/react-native-keychain.js` grew `hasInternetCredentials`. A new root `__mocks__/react-native-secure-key-store.js` rejects `get` with the missing-key code: without it the legacy library loads for real in every consumer spec, `NativeModules.RNSecureKeyStore` is undefined and every legacy read throws, flipping the default world of ~26 specs from "absent" to "failed". Behavioural mocking stays inline per spec.

1. **Internet credentials, not generic passwords.** Rejected `setResetOnAppUninstallTo(false)` as the load-bearing defense: `resetOnAppUninstall` is instance state re-initialised to `YES` on every module construction and the setter is a fire-and-forget void bridge method, so it is defended by call ordering rather than by structure. It is called anyway as defense in depth — it also covers `kSecClassKey` items and any future dependency.

2. **The guard is re-armed before every legacy read, not once at module load.** The wipe is only ever skipped, never disarmed: `handleAppUninstallation` sets its `RnSksIsAppInstalled` flag *inside* the branch it guards, so a skipped wipe leaves the trigger armed for the next call. It is iOS-only — the Android module exports `get` / `set` / `remove` and nothing else — and it is left off the erase, the one entry point that never reaches `handleAppUninstallation`.

3. **One `server` per slot, and never one another feature owns.** iOS matches internet credentials by server with `kSecMatchLimitOne` and no account predicate, so several items sharing a server collapse to one OS-picked entry — the trap documented in `use-credential-backup.ts`. One item per server sidesteps it, and `setInternetCredentialsForServer` deletes by server before every insert, so a duplicate cannot be created even by changing the username. The same by-server delete is why a test pins the namespace apart from `BLINK_DOMAIN`, whose items are the iCloud mnemonic backup.

4. **A failed new-store read returns `failed` and does not fall back.** The legacy copy may be stale, and falling back would re-enter the legacy library on exactly the path where its wipe is most likely to be armed.

5. **A failed legacy read returns `failed`, not `absent`.** "Flaky keystore" and "nothing there" are indistinguishable, and scoring the first as absent is what deletes credentials. `absent` has exactly one producer: both stores confirming a miss, which is what keeps a genuine fresh install honest.

6. **The new store wins over legacy unconditionally, without comparing.** Neither store carries a timestamp, so any other rule is a guess. Returning on the first hit enforces it for free and makes an orphaned legacy copy unreachable.

7. **Migration bookkeeping never costs availability.** A failed migrating write still returns the value, a failed erase is reported and tolerated, and the telemetry call is isolated on both halves — the synchronous one by a `catch`, the promise it returns by its own handler.

8. **`removeThrough` stops when the legacy copy is not provably gone.** While the new store still holds the value it answers first, so the survivor stays unreachable; emptying the new store anyway would turn "two copies, one deleted" into "one copy, and it is the stale one". Rejected the alternative — carry on when the legacy store cannot say what is left, on the grounds that a copy nobody can read is a copy nobody can resurrect — because it trades a completed delete on a broken store for a silently restored credential the moment that store recovers.

9. **A per-slot queue, not a global one.** A read racing a remove on one slot is not hypothetical: a logout removes the auth token while other callers still read it. Unserialized, the read resolves its legacy value before the remove runs and writes it back after. Unrelated slots never wait on each other, and the entry is dropped once nothing follows it.

10. **The helpers never reject.** Every failure arrives as `failed` or `false`, so a consumer that today swallows a keystore error into `null` keeps working unchanged when it moves over.

11. **The legacy-hit counter is not deduplicated per process.** Once a slot has migrated the new store answers first and it never fires again, so a repeat is real signal that the migrating write keeps failing. Removal in [#1163](https://github.com/blinkbitcoin/blink-wip/issues/1163) is gated on it going quiet, so it has to exist from the first migrating build. It lives in the migration module rather than `app/utils/analytics.ts` because that module's import graph reaches screens, and `secureStorage.ts` starts depending on this helper in #1161.

12. **`accessible` is a required parameter, never defaulted.** `Keychain.ACCESSIBLE` has no `ALWAYS_THIS_DEVICE_ONLY`, which is what every non-mnemonic slot uses today, so each slot's protection class is a decision to be made as it moves in #1161 rather than one to inherit silently here.

13. **`secureWrite` refuses an empty value.** `secureRead` reports one as absent, so storing it would leave a slot that reads as unset and sends the helper back to the legacy store. Clearing a slot is `secureRemove`'s job.

## Verification

- 146 tests across 5 suites: the four new specs plus `secure-storage.spec.ts`, which passes **unmodified** under the new root mock.
- 100% statements, branches, functions and lines on all three new modules.
- **33 mutations**, each applied alone and reverted after: every one turns the suite red. They cover all nine read-through branches (new-hit with legacy never called, new-miss + legacy-hit with write-before-erase ordering asserted through a shared call-order array, both-miss, legacy-read-fails, new-write-fails, legacy-erase-fails, both-present-and-differ, concurrent reads, the write path), the remove and exists policies, the queue, the deadline, the telemetry redaction and the four adapter guards. Three of them do not merely fail the suite — dropping the telemetry promise handler, or queueing the unsettled promise, aborts the Node process on an unhandled rejection.
- An API-stability spec pins `Object.getOwnPropertyNames(KeyStoreWrapper).sort()` to a literal list. 23 files under `app/` call `KeyStoreWrapper` and ~26 specs mock it as a partial object literal, none of them naming the underlying library, so as long as that surface only grows the blast radius of the whole backend swap is one file.
- `tsc -p .`, ESLint and Prettier clean on all ten files.
- The release JS bundle built from this branch contains **no** string from the new modules — `secure-store.blink.local`, `legacy_key_store_hit`, `storage-legacy-erase-failed` and `setResetOnAppUninstallTo` are all absent from 34.7 MB, while `galoyAuthToken`, `sessionProfiles`, `pinFailureState` and `RNSecureKeyStore` are present. Nothing imports the new modules, so Metro never pulls them in and the shipped bundle is unchanged.

## Device test

iOS simulator (iPhone 16 Pro, iOS 18.6), Metro serving this branch, self-custodial account with a saved session.

| Step | Result |
|---|---|
| Terminate the app, cold launch | Splash renders, no red screen and no Metro resolution error |
| Wait for the bundle to load | Home renders on the active account |
| Session across the kill | Restored — the keystore read that backs it still works |
| Balances and action row | Bitcoin, Dollar, Transfer / Receive / Send / Scan all intact |
| Non-custodial banner and tab bar | Unchanged |

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -124,6 +124,17 @@
     "no-plusplus": "error",
     "no-promise-executor-return": "error",
     "no-proto": "error",
+    "no-restricted-imports": [
+      "error",
+      {
+        "paths": [
+          {
+            "name": "react-native-secure-key-store",
+            "message": "Import the legacy key store only from app/utils/storage/legacy-key-store.ts, which re-arms the uninstall guard before every call. See blinkbitcoin/blink-wip#1160."
+          }
+        ]
+      }
+    ],
     "no-return-assign": "error",
     "no-return-await": "error",
     "no-script-url": "error",
@@ -164,6 +175,17 @@
     "jest/no-async-promise-executor": "off"
 
   },
+  "overrides": [
+    {
+      "files": [
+        "app/utils/storage/legacy-key-store.ts",
+        "app/utils/storage/secureStorage.ts"
+      ],
+      "rules": {
+        "no-restricted-imports": "off"
+      }
+    }
+  ],
   "settings": {
     "react": {
       "version": "detect"
--- a/__mocks__/react-native-keychain.js
+++ b/__mocks__/react-native-keychain.js
@@ -4,6 +4,7 @@ const resetGenericPassword = jest.fn(() => Promise.resolve(true))
 
 const setInternetCredentials = jest.fn(() => Promise.resolve({ service: "mock-service" }))
 const getInternetCredentials = jest.fn(() => Promise.resolve(false))
+const hasInternetCredentials = jest.fn(() => Promise.resolve(false))
 const resetInternetCredentials = jest.fn(() => Promise.resolve(undefined))
 
 const ACCESSIBLE = {
@@ -21,6 +22,7 @@ export {
   resetGenericPassword,
   setInternetCredentials,
   getInternetCredentials,
+  hasInternetCredentials,
   resetInternetCredentials,
   ACCESSIBLE,
 }
@@ -31,6 +33,7 @@ export default {
   resetGenericPassword,
   setInternetCredentials,
   getInternetCredentials,
+  hasInternetCredentials,
   resetInternetCredentials,
   ACCESSIBLE,
 }
--- a/__mocks__/react-native-secure-key-store.js
+++ b/__mocks__/react-native-secure-key-store.js
@@ -1,0 +1,38 @@
+// Root manual mock, applied automatically to every spec.
+//
+// Without it the legacy library loads for real, `NativeModules.RNSecureKeyStore`
+// is undefined, and every legacy read throws a TypeError. That reads as "the
+// keystore failed", not as "nothing is stored" — so the read-through helper
+// would report `failed` in the default world of every consumer spec instead of
+// `absent`. Rejecting with the missing-key code keeps the default state
+// "absent everywhere", which is what those specs assume.
+//
+// Both native modules reject a missing key with code "404"
+// (ios/RNSecureKeyStore.m `get`, RNSecureKeyStoreModule#get). React Native turns
+// a native rejection into an Error carrying that code, so the mock does too.
+const keyNotFoundError = () =>
+  Object.assign(new Error("key does not present"), { code: "404" })
+
+const get = jest.fn(() => Promise.reject(keyNotFoundError()))
+const set = jest.fn(() => Promise.resolve("key stored successfully"))
+const remove = jest.fn(() => Promise.resolve("cleared alias"))
+const setResetOnAppUninstallTo = jest.fn(() => true)
+
+const ACCESSIBLE = {
+  WHEN_UNLOCKED: "AccessibleWhenUnlocked",
+  AFTER_FIRST_UNLOCK: "AccessibleAfterFirstUnlock",
+  ALWAYS: "AccessibleAlways",
+  WHEN_PASSCODE_SET_THIS_DEVICE_ONLY: "AccessibleWhenPasscodeSetThisDeviceOnly",
+  WHEN_UNLOCKED_THIS_DEVICE_ONLY: "AccessibleWhenUnlockedThisDeviceOnly",
+  AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY: "AccessibleAfterFirstUnlockThisDeviceOnly",
+  ALWAYS_THIS_DEVICE_ONLY: "AccessibleAlwaysThisDeviceOnly",
+}
+
+export { ACCESSIBLE }
+
+export default {
+  get,
+  set,
+  remove,
+  setResetOnAppUninstallTo,
+}
--- a/__tests__/utils/key-store-api-stability.spec.ts
+++ b/__tests__/utils/key-store-api-stability.spec.ts
@@ -42,6 +42,7 @@ const PINNED_SURFACE = [
   "prototype",
   "read",
   "readActiveToken",
+  "readSessionProfiles",
   "readWithStatus",
   "removeActiveToken",
   "removeIsBiometricsEnabled",
--- a/__tests__/utils/key-store-api-stability.spec.ts
+++ b/__tests__/utils/key-store-api-stability.spec.ts
@@ -1,0 +1,90 @@
+import KeyStoreWrapper from "@app/utils/storage/secureStorage"
+
+/**
+ * 23 files under `app/` call `KeyStoreWrapper`, and ~26 specs mock it as a
+ * partial object literal holding only the methods their subject calls. None of
+ * them name the underlying library. So as long as this surface only ever grows,
+ * swapping the storage backend under it (blinkbitcoin/blink-wip#1143) touches
+ * one file and no consumer at all.
+ *
+ * This spec is what makes that a checked property rather than an intention: a
+ * rename, a removal or a signature move shows up here first, in a failure whose
+ * message names the member. Adding a method is expected and safe, and updating
+ * the list below is the deliberate act that records it.
+ */
+const PINNED_SURFACE = [
+  "ACTIVE_TOKEN",
+  "IS_BIOMETRICS_ENABLED",
+  "LEGACY_PIN_ATTEMPTS",
+  "MNEMONIC",
+  "MNEMONIC_NETWORK",
+  "PIN",
+  "PIN_FAILURE_STATE",
+  "SESSION_PROFILES",
+  "clearPinFailureState",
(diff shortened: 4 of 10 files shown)


grimen (changes requested):
## Findings

**1. Timeout doesn't cancel the underlying task, allowing a stale write after removal (blocking).**

The per-slot queue in `onSlot` exists to serialize reads and removes on the same storage slot, but its 30s timeout only races a wrapper promise — it never cancels the real task chain. When a hung native call outlives the timeout, `onSlot` marks the operation failed and frees the slot for the next queued operation.

*Example:* A `readThrough({slot: "authToken"})` call hangs past 30s inside `secureRead`/`legacyRead`. `onSlot` reports failure and releases the slot, so a concurrent `removeThrough` (e.g. triggered by logout) runs and clears both stores. The original hung call then finally resolves, and its abandoned continuation calls `secureWrite` with the stale legacy value — silently restoring the credential that was just removed. The test at lines 492–510 only exercises a promise that never resolves, so this path isn't covered.

*Proposal:* One option is to have the timeout also mark the slot's generation/epoch as stale so a late-resolving task's write is dropped if the slot has moved on, or to propagate actual cancellation into the underlying task chain.

**2. Incognito mode selection is persisted but never enforced (blocking).**

Selecting Incognito on `choose-experience-screen.tsx` only writes `selfCustodialAccountModeByAccountId`; no runtime code reads that value. `useDeviceLocation` therefore continues performing IP-based location lookups, and region-gated features behave exactly as they do in Enhanced mode, so the advertised privacy choice has no behavioral effect.

*Example:* A user selects Incognito expecting to avoid location detection, but the next location-dependent flow still invokes the IP lookup and applies the same feature gating as Enhanced mode.

*Proposal:* Consult the active account's saved mode before device-location detection and every region-gated feature path, skipping IP lookup and disabling the corresponding features when the account is Incognito.

grimen (approved):

grimen (approved):

blink-admin (approved):

grimen (approved):

Issue #1162, opened 2023-02-16
make it possible to force an update for the user


nicolasburtey:
ie: in case of gql deprecation

nicolasburtey:
this has been done on the mobile side. we just need to integrate this into the CI/CD pipeline now.

sandipndev:
Can you explain what is to be done here?
I would like to accelerate this work.

nicolasburtey:
a first iteration on this issue has been done. closing this issue.

PR #4211, opened 2026-08-26
feat(security): move the session, PIN and biometrics slots to react-native-keychain

esaugomez31:
Fixes [blinkbitcoin/blink-wip#1161](https://github.com/blinkbitcoin/blink-wip/issues/1161) (close manually on merge — cross-repo `Fixes` doesn't auto-close).

Step 2 of [blinkbitcoin/blink-wip#1143](https://github.com/blinkbitcoin/blink-wip/issues/1143). Stacked on [#4210](https://github.com/blinkbitcoin/blink-mobile/pull/4210) (step 1, the adapter and the read-through helper) and on [#4209](https://github.com/blinkbitcoin/blink-mobile/pull/4209), which touches the same file and merges first. Both are in the base of this branch, so the diff below is the three commits this PR owns.

## Problem

`react-native-secure-key-store` has been unmaintained since ~2019 and holds six of the app's secrets: the session token, the multi-account session profiles, the PIN, its lockout state, the pre-lockout attempt count and the biometrics flag. Nothing is broken today; the risk is the ordinary one, an unmaintained native dependency holding credentials that will eventually break on some OS release, fixed under pressure with users locked out.

Its read contract is also the weaker half of a problem [#4209](https://github.com/blinkbitcoin/blink-mobile/pull/4209) fixed one level up. The library rejects a missing key rather than resolving empty, so "nothing stored" and "the read failed" arrive the same way, and on iOS they are genuinely indistinguishable.

## What changed

**The six slots read and write through the Keychain-backed store** ([`secureStorage.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/0e6f2db4c/app/utils/storage/secureStorage.ts)). Reads go through the read-through helper from step 1, so a slot still living in the legacy store migrates the first time it is touched and its legacy copy is erased; writes land in the new store only. The mnemonic slots deliberately stay where they are — they move in [blinkbitcoin/blink-wip#1162](https://github.com/blinkbitcoin/blink-wip/issues/1162), so the file carries both sets of primitives for one step.

**The protection class changes for all six**, from `ALWAYS_THIS_DEVICE_ONLY` to `AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`. `react-native-keychain` has no equivalent of the old one — Apple deprecated `kSecAttrAccessibleAlwaysThisDeviceOnly` in iOS 12 — and the closest class is a strict improvement, since the old one was readable while the device was locked. What it costs is one window: `Info.plist` declares `UIBackgroundModes: remote-notification`, so a silent push can launch the app after a reboot but before the first passcode entry, where a read returns `errSecInteractionNotAllowed` and the adapter reports `failed`.

**That window is why the gates now fail closed.** `readIsPinEnabled` and `readIsBiometricsEnabled` answer `yes` / `no` / `failed`, and the two places that decide whether the app is locked — [`app-state.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/0e6f2db4c/app/navigation/app-state.tsx) on resume and [`authentication-check-screen.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/0e6f2db4c/app/screens/authentication-screen/authentication-check-screen.tsx) on launch — treat anything but a definite `no` as a lock. Before this, a read that could not answer returned `false` and the lock screen simply did not appear.

**The recovery-phrase gate joins them.** [`use-biometric-gate.ts`](https://github.com/blinkbitcoin/blink-mobile/blob/0e6f2db4c/app/screens/card-screen/hooks/use-biometric-gate.ts) collapsed a failed read to "biometrics off" and authenticated without prompting. Its consumer is the view-backup-phrase screen, so the seed phrase could be shown with no biometric prompt on exactly the read that this PR makes reachable.

**The reinstall wipe covers the new slots.** Keychain internet credentials outlive an uninstall the same way the legacy items did, so `clearUninstallSurvivingCredentials` now removes the PIN, the lockout state and the biometrics flag alongside the token and the profiles.

[`persistent-state/index.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/0e6f2db4c/app/store/persistent-state/index.tsx) has zero diff, which is the signal the issue asks for that the boot path from [#4160](https://github.com/blinkbitcoin/blink-mobile/pull/4160) survived untouched.

1. **An uncertain lock routes to the authentication screen, not the PIN keypad.** Failing closed on the enablement flag can send a user who never set a PIN to a lock screen, and `verifyPin` would answer `unreadable` on every entry — no attempt spent, so the exhausted-logout path never fires, and the keypad has neither a back affordance nor a logout. The authentication screen carries the logout escape and still routes to the PIN, so a lock we inferred rather than read goes there.

2. **`preserve` is one group, not per-slot.** The reinstall wipe takes the PIN with the session because a real reinstall otherwise boots the next owner into a PIN nobody on the device chose, which is unrecoverable. The cost is a known false positive: the caller reaches the wipe through `MigrationStatus.NoData`, which a throwing AsyncStorage read can produce without a reinstall, so the app lock now goes with the session on that path. Taken deliberately — it costs a lock the user can set again, on a boot that already signed them out loudly.

3. **Mnemonics start surviving an uninstall on iOS.** The legacy library's unscoped sweep was the only thing deleting `mnemonic:<accountId>` items on reinstall, and the read-through helper disarms that sweep before every legacy touch. This is the product decision recorded as finding 2 of [blinkbitcoin/blink-wip#1143](https://github.com/blinkbitcoin/blink-wip/issues/1143), which blocks [#1162](https://github.com/blinkbitcoin/blink-wip/issues/1162); it arrives ahead of that answer and nothing in the storage layer can defer it. Recorded in the code at the wipe.

4. **`readPin` is not added, despite the issue's table.** `getPin()` already returns `null` for both an absent and a failed read, and `verifyPin` treats that as unreadable: no attempt spent, no forced logout, a retry offered. The contract the sibling exists to provide already holds, so it would ship with no caller.

5. **The ESLint `no-restricted-imports` exception for `secureStorage.ts` stays.** The issue expects it to go with this step, but the mnemonic slots still import the legacy library directly. It goes with [#1162](https://github.com/blinkbitcoin/blink-wip/issues/1162).

6. **Three commits, not one per slot.** The issue asks for one commit per slot so that a storage move can never land without its read contract. All six move together and their contracts land in the same commit, so the failure that rule prevents cannot occur here; the split instead follows the three separable concerns — the storage move, the app-lock gates, and the recovery-phrase gate.

## Verification

- 1223 tests across 93 suites: the storage spec plus every transitive consumer — `persistent-state`, the authentication screens, `app-state`, `use-logout`, the multi-account screens, `use-account-registry`, the backup-phrase screen and the step-1 helper specs.
- `secureStorage.ts` at **100%** statements, branches, functions and lines. `app-state.tsx` and `use-biometric-gate.ts` likewise. `authentication-check-screen.tsx` is at 100% lines, with one pre-existing branch uncovered (the light/dark logo).
- Mutation-verified across files: collapsing `readActiveToken`'s failed read to absent turns two storage tests red, and [`index.spec.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/0e6f2db4c/__tests__/store/persistent-state/index.spec.tsx) already locks the other half — a failed read must not scrub the blob nor write the keychain. Flipping either lock gate from fail-closed to fail-open turns exactly one test red each.
- Dead code removed: with both stores reporting an empty value as absent, the empty-payload guard in `readSessionProfiles` became unreachable.
- `tsc -p .`, ESLint and Prettier clean on all ten files.

## Before / after — the lock, skipped and restored

Same emulator, same account, a PIN set, and the same forced fault: the Keychain
existence probe made to fail, which is what the pre-first-unlock window does on
its own. Only the two gate files differ between the shots.

| before (`main`) | after (this PR) |
|---|---|
|  |  |
| The probe answers `false`, so the app decides no lock is set and opens straight to the home screen — balances and all — for a user who does have a PIN. | The probe answers `failed`, the gate refuses to read that as "no lock", and the app stops at the unlock screen. It is the authentication screen rather than the bare keypad, so **Use PIN** and **Log Out** are both there — decision 1 above, on screen. |

<sub>Images live on branch `assets/pr-4211-demo`, embedded by commit SHA — deleting that branch breaks them.</sub>

## Device test

Android emulator, staging. The build was installed **over** the previous one, never uninstalled, so the existing install's values were in the legacy store and the boot exercised the real migration.

| Step | Result |
|---|---|
| Launch after the in-place upgrade | Session intact, balances rendered |
| Account switcher | Both custodial profiles listed, the active one checked |
| Switch accounts and back | Both survive, the check follows |
| Force-stop and relaunch | Both survive, the active account is correct |
| Security and Privacy | Both toggles read correctly as off |
| Set a PIN | Toggle on, security score 1/4, `Biometrics/PIN: Enabled` |
| Force-stop and relaunch | The lock screen appears |
| Enter the PIN | Unlocks to home |
| Turn the PIN off | Toggle off, score back to 0/4 |

On disk after the upgrade: the Keychain store (`shared_prefs/secret_shared_prefs.xml`) holds its encrypted entries, and no file under the app container carries `galoyAuthToken`, `sessionProfiles` or `RNSecure` any more — the legacy copies are gone.

`logcat` clean throughout. Four of the six slots are covered on device this way; the biometrics flag has no sensor on the emulator and is covered by tests only.


grimen (changes requested):
## Findings

**1. Biometric gate can wave the user through with no authentication (blocking).**
`useBiometricGate` only short-circuits when the settings read definitively says biometrics are off (`no`). When the read comes back `failed` instead, the code falls through to the sensor-availability check, and if the sensor is unavailable while `required` is left at its default `false`, it calls `setAuthenticated(true)` without ever prompting the user.

*Example:*
`view-backup-phrase-screen.tsx` calls `useBiometricGate({ onlyIfBiometricsEnabled: true })` without `required: true`; on a device with no biometric hardware, hitting the pre-first-unlock Keychain window this PR introduces (`AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`) makes `readIsBiometricsEnabled()` return `failed`, and the recovery phrase screen unlocks with zero authentication.

*Proposal:*
Treating `failed` the same as `yes` (fail closed) for this check, rather than only `no`, would close the gap — the function's own new comment already states this intent.

**2. Migrated writes can overwrite or resurrect credentials (blocking).**
`migratedWrite` calls `secureWrite` directly, while reads, existence probes, and removals are serialized through `onSlot`. A first read-through migration can therefore overwrite a newer value written concurrently, and an in-flight write can complete after logout removes the slot, bringing the credential back.

*Example:*
A read observes the new active-token slot as absent and obtains the legacy token. `setActiveToken()` then writes a freshly rotated token, but the read-through migration subsequently writes its stale legacy token over it. In the inverse race, a delayed `setActiveToken()` can finish after `removeActiveToken()` reports success and restore the session.

*Proposal:*
Route migrated writes through the same per-slot queue as reads and removals so operations preserve invocation order. Add regression tests for both read-through versus save and delayed write versus removal; reverting the serialization should make them fail.

grimen (dismissed):

grimen (approved):

grimen (dismissed):

blink-admin (dismissed):

grimen (dismissed):

blink-admin (approved):

PR #4214, opened 2026-08-26
fix(scan): stop reporting a merchant lnurl service failure as an invalid QR code

esaugomez31:
Related to blinkbitcoin/blink-wip#1175

This does not close the ticket. It fixes the message the ticket reported; that a live bill completes the payment flow still needs a fresh code to confirm.

## The report was a misdiagnosis, and our error message caused it

The ticket says Blink rejects a MoneyBadger/Wigroup bill QR as an invalid QR code, and asks whether `za.wigroup.co/bill/...` is missing from the scanner allowlist.

It is not missing. `@blinkbitcoin/blink-client` 1.1.0 already matches the host through its Yoyo merchant rule and turns the scanned URL into a Money Badger lnurl:

```
https://za.wigroup.co/bill/172366037
  parses to  https%3A%2F%2Fza.wigroup.co%2Fbill%2F172366037@cryptoqr.net
```

What actually happens next, measured against the live services:

| Request | Response |
|---|---|
| `GET https://cryptoqr.net/.well-known/lnurlp/https%3A%2F%2Fza.wigroup.co%2Fbill%2F172366037` | `HTTP 500` `{"status":"ERROR","reason":"[GET /scan][400] getQrInfoBadRequest {\"code\":\"UNKNOWN_CODE\",\"message\":\"unknown scanned code data\"}"}` |
| `GET https://za.wigroup.co/bill/172366037` | `HTTP 404` |

The receipt on the ticket rules out an expired code: it is stamped 15:57 and the phone clock in the screenshot reads 3:57, so it was scanned as it came off the printer. The code was live and Money Badger still could not resolve it.

Two things on that receipt are worth a look from someone who knows the Money Badger integration. The QR sits under "DOWNLOAD OUR LOYALTY APP, COFFEE STAMPS VOUCHERS & MORE", and the apps the shop lists as accepted are Bootlegger Loyalty, SnapScan and Zapper, with no Money Badger. The merchant rule in `blink-client` matches any URL containing `wigroup.co`, so a loyalty code matches it just as a bill would. Whether that is the cause is for Money Badger to answer; either way it is not something the app can fix.

The app bug is what we did with that failure. `resolveLnurlDestination` wrapped the whole pay-request block in one `catch {}` and returned `LnurlUnsupported` for anything that threw, which the scanner renders as **"This is not a valid Bitcoin address or Lightning invoice"**. A shopper standing at a till was told their code was not a payment code at all, and that is the message that sent this ticket looking for a parser gap.

> [!NOTE]
> The scanner never failed to read the format. `parsePaymentDestination` returns `valid: true` for the reported URL, as the merchant Yoyo. The scan failed because the bill no longer exists, and the app's only defect was calling that "not a valid Bitcoin address or Lightning invoice".

## The fix

A new `InvalidDestinationReason.LnurlServiceError`, raised only when the failure is provably on the service side:

- an HTTP `5xx`, the service failing out loud,
- a body of `{"status":"ERROR"}`, the service saying so in LUD-06's own words, or
- a rejection carrying the request but no response, which is how a refused connection, DNS failure, TLS error or timeout arrives.

Everything else keeps the behaviour it had. A `404` is the service answering that this destination does not exist, so a mistyped lightning address still gets its old message, and an lnurl-auth or channel-request QR is still reported as unsupported rather than told to try again later.

The `ERROR` body needs the response while it is still whole, because `lnurl-pay` flattens it into a plain `Error` carrying the service's reason text, indistinguishable from its own input validation. So the pay request now goes through a fetcher this file owns. The URL it receives is already vetted: `lnurl-pay` checks it is a URL and refuses onion hosts before calling any fetcher, and the app vets the lnurl it was derived from.

The pay request is now guarded on its own, so a failure of the account lookup that follows it is no longer scored as a fact about the lnurl.

Both surfaces that show an invalid destination were taught the new reason: the scanner alert, and the send destination screen for a pasted code.

## Screenshots

| Before | After |
| :---: | :---: |
| 1175-before | 1175-after |

Scanning the reported QR on an Android emulator against the live `cryptoqr.net`. Before: "Invalid QR Code / This is not a valid Bitcoin address or Lightning invoice". After: "Code Not Available / We could not process this code. It may have expired, or the service may be temporarily unavailable. Ask for a new code or try again later."

## Tests

15 new tests, all affected suites green (496 tests across 36 suites, locale parity included), `tsc` and `eslint --max-warnings 0` clean.

- The reported URL is driven end to end through the real `blink-client` parser, asserting it reaches the merchant lnurl service rather than being rejected as unparseable, and that it resolves to a payable destination when the service can answer.
- Each classification branch is pinned: `500`, `503`, a `{"status":"ERROR"}` body and a request that never got a response as service failures; `404`, `Invalid lnUrlOrAddress`, `Invalid pay service params` and a failed account lookup as unsupported.
- New spec for `DestinationInformation`, which had no coverage.

Also verified on device: a QR holding an unparseable string still shows "Invalid QR Code", and a lightning address that 404s keeps its previous message, so the new copy is scoped rather than blanket.

## Found by the device test: this is not only a MoneyBadger problem

Scanning a swap merchant code on the emulator landed on "Enter a valid destination", because `swap.blink.sv` reports its own outage as HTTP 200 with `{"status":"ERROR","reason":"swap provider unavailable"}`. That is one of blink's own services telling a user that a code it just handed them is not a valid destination. The `ERROR`-body branch above is what closes it, and the same scan now reads "We could not process this code. It may have expired, or the service may be temporarily unavailable".

## Not covered

That a **working** till code completes a payment is covered by a test, not by the device. No Money Badger code that the service can resolve was available, including the reported one. Confirming that end to end needs a code from a shop that is actually enabled with Money Badger.

## Left for its own change

`destination-information.tsx` has `case InvalidDestinationReason.LnurlError || InvalidDestinationReason.LnurlUnsupported:`, which evaluates to `"LnurlError"`, so `LnurlUnsupported` never matches and falls through to the generic default. It is a real bug, it predates this branch, and a comment in `lnurl.ts` documents relying on it, so fixing it here would change messaging for paths this ticket is not about.


kngako (dismissed):
Approve the change while highlightning test coverage gaps that should probably be addressed in a future PR or something.


--- 

## Test Coverage Gaps

1. **The custom fetcher's success path is never exercised.** Every spec mocks
   `requestPayServiceParams` wholesale, so nothing pins that `fetchLnurlPayJson` hands a
   valid pay body through to `lnurl-pay`'s parser. Per the repo's own review norm, mocks
   should not neutralize the assumption being fixed — add one test using the real
   `requestPayServiceParams` with only `axios.get` mocked, returning valid pay params, and
   assert a `valid: true` destination.
2. **No test for the real outage sequence.** In production a down service makes `getParams`
   return `{status: "ERROR"}` (js-lnurl never rejects) *and* `requestPayServiceParams`
   reject; the specs always give `getParams` a benign response. One test with
   `getParams` returning the ERROR shape plus a 500 rejection would pin the actual #1175
   sequence end-to-end.
3. **No tests for the 4xx boundary.** Nothing locks 404-is-unsupported vs. other-4xx
   behavior (see Finding 1).
4. **`LnurlUnsupported` rendering in `destination-information` is untested.** The new spec
   covers `LnurlServiceError`, `LnurlError`, and `UnknownDestination` only — which is why
   the pre-existing fallthrough in Code Smell 1 has survived unnoticed.
5. **No test for a malformed LUD-06 error body.** A `{status: "ERROR"}` response with a
   missing or non-string `reason` exercises the `String(data.reason ?? "")` path
   (lnurl.ts:60); worth one cheap case.

grimen (approved):

PR #4215, opened 2026-08-26
fix(android): keep content clear of the system bars under Android 15 edge to edge

esaugomez31:
Closes blinkbitcoin/blink-wip#887

## What the audit found

From Android 15 a target of 35 or above draws edge to edge, and this app targets 36, so production is already edge to edge. Only three screens call `useSafeAreaInsets`, which reads as thin coverage, but it is not where the work happens: [`app/components/screen/screen.tsx`](https://github.com/blinkbitcoin/blink-mobile/blob/main/app/components/screen/screen.tsx) wraps every screen in a `SafeAreaView` and derives its edges from whether a navigation header is shown, and the bottom tab bar adds `insets.bottom` to its own height. Most screens are covered by that alone.

Two classes are not, and I walked all 21 header-less routes to find them.

**Screens that opt out with `unsafe` and then place controls at a fixed offset from the window edge.** One real case: the QR scanner.

**Screens on a route the navigator marks `headerShown: false` that never pass that on to `Screen`.** `Screen` only adds the top edge when it is told the header is hidden, so these draw their first row under the status bar. Six of them.

## The scanner

`scanning-qrcode-screen.tsx` renders `<Screen unsafe>` and placed its controls with two hardcoded numbers:

```ts
close:       { marginTop: 40 }                       // stood in for the status bar
openGallery: { top: screenHeight - 96, height: 128 } // ends 32px past the window
```

`screenHeight` comes from `Dimensions.get("window")`, which under edge to edge includes the area the system bars are drawn over. The gallery and clipboard buttons ran off the bottom of the window and the navigation bar cut them in half. Both offsets now come from `useSafeAreaInsets()`.

## Screenshots

This is the change with the visible impact, so it is the one shown here. Captured on an Android 15 emulator against both navigation modes.

Three-button navigation, where the navigation bar cuts the controls:

| Before | After |
| :---: | :---: |
| 01-scanner-3button-BEFORE | 02-scanner-3button-AFTER |

Gesture navigation:

| Before | After |
| :---: | :---: |
| 03-scanner-gesture-BEFORE | 04-scanner-gesture-AFTER |

The six header-less screens are centered layouts, so their fix moves content by the height of the status bar without a difference a screenshot can show. They are covered by tests instead, below.

## The six header-less screens

`conversionSuccess`, `selfCustodialBackupSuccess`, `selfCustodialModeSwitchSuccess`, `accountMigrationStart`, `accountMigrationTransferringFunds` and `temporarilyUnavailable` now pass `headerShown={false}`, which is what `sendBitcoinCompleted`, `contactDetail` and `peopleHome` were already doing correctly.

## Edge-to-edge setup, as the ticket asks it be documented

`app.tsx` had no `SafeAreaProvider`. React Navigation ships `SafeAreaProviderCompat` inside each navigator, which is why insets worked at all, but only inside a navigator and only from the first measured frame. The tree is now wrapped in `SafeAreaProvider` with `initialWindowMetrics`, so insets are correct on the first frame and available to anything rendered outside a navigator. React Navigation's compat degrades to a plain `View` when it finds insets already in context, so values inside navigators are unchanged.

**On `BootTheme`:** `styles.xml` comments that `BootTheme` should inherit from `Theme.BootSplash` or `Theme.BootSplash.EdgeToEdge`. I tried the `EdgeToEdge` parent and backed it out. `BootTheme` is `MainActivity`'s manifest theme, so that parent makes the whole activity window edge to edge on every release from Android 8, not only on 15, and `postBootSplashTheme` does not reset those window attributes afterwards. The window is already edge to edge on 15 through the target SDK, so it buys nothing there and puts the older releases at risk. `styles.xml` is untouched.

## Tests

14 new tests.

- `__tests__/navigation/header-less-screen-insets.spec.ts` walks the navigators and asserts that every route declaring `headerShown: false` passes it on to `Screen`. It derives the route list from the source rather than a hand-kept list, so a new header-less screen cannot be added without either handling its insets or being recorded in the exception set, and a second test fails if an exception entry goes stale.
- The scanner spec mocks `useSafeAreaInsets` and asserts the bottom controls are offset by `insets.bottom` and the close button by `insets.top`.

519 tests across 39 suites pass. `tsc` and `eslint --max-warnings 0` are clean.

## Verification

Android 15 (API 35) emulator in both gesture and three-button navigation, switching modes live with `cmd overlay enable`, and an Android 14 (API 34) emulator for the no-regression check.

## Not folded in

blink-wip#888 (deprecated `setStatusBarColor` / `setNavigationBarColor`) and blink-wip#889 (portrait lock) are their own tickets and stay out of this one.

### DIFF
--- a/__tests__/navigation/header-less-screen-insets.spec.ts
+++ b/__tests__/navigation/header-less-screen-insets.spec.ts
@@ -16,6 +16,8 @@ const NAVIGATION_DIR = path.join(APP_DIR, "navigation")
 
 type Route = { name: string; component: string; file: string }
 
+type HeaderLessScreen = { label: string; component: string }
+
 const readSource = (file: string): string => fs.readFileSync(file, "utf8")
 
 const navigationSources = (): string[] =>
@@ -43,6 +45,16 @@ const headerLessRoutes = (): Route[] => {
   return routes
 }
 
+const BAILOUT_RENDER = /return\s*\(?\s*<([A-Z]\w*)(?![\w.])[^<>]*\/>/g
+
+/**
+ * A component the file hands the whole screen over to before any header is in play: the
+ * navigator bails out to it instead of mounting its tree, or a header-less route
+ * delegates to it. It draws with no toolbar above it, so the same rule reaches it.
+ */
+const bailoutRenders = (source: string): string[] =>
+  [...source.matchAll(BAILOUT_RENDER)].map((match) => match[1])
+
 const sourceFiles = (dir: string): string[] =>
   fs.readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
     const full = path.join(dir, entry.name)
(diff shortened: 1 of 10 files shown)


grimen (changes requested):
## Findings

**1. Regression guard only checks that `headerShown=` is present, not its value (blocking).**
`declaresTopInset` in `header-less-screen-insets.spec.ts` treats any `<Screen headerShown=...>` as compliant, regardless of what the value actually is. This means the test can't distinguish `headerShown={false}` from `headerShown={true}`, so it can't actually detect the regression it's meant to guard against.

*Example:* A future header-less route whose screen renders `<Screen headerShown={true}>` by mistake still passes the guard, and the Android-15 status-bar overlap this PR fixes comes back with a green suite.

*Proposal:* Check the tag for `headerShown={false}` (or the negation) rather than the mere presence of the prop.

## Test Coverage Gaps

- `TemporarilyUnavailableScreen` is invisible to the new guard: it's only ever direct-rendered (from `root-navigator.tsx` and `migration-gate.tsx`), never registered as an `<X.Screen>` route, and it's not listed in `HANDLES_ITS_OWN_INSETS` either — reverting this PR's `headerShown={false}` change on `temporarily-unavailable-screen.tsx` leaves the full suite green.

grimen (approved):

PR #4224, opened 2026-08-28
feat(security): move the per-account mnemonics to react-native-keychain

esaugomez31:
Step 3 of blinkbitcoin/blink-wip#1143, tracked in blinkbitcoin/blink-wip#1162: the two mnemonic slots move to `react-native-keychain`, the last ones still living in `react-native-secure-key-store`.

## The product question, answered

The ticket was blocked on one decision: today's behaviour is that a reinstall clears the mnemonics on iOS, and moving stores would make them survive for the first time. **The answer is to keep today's behaviour** — a wallet's recovery path is its backup phrase and its cloud backup, not a keychain item outliving the app that wrote it.

That is now something the code decides rather than something it inherits. It used to happen because the legacy library's own reinstall sweep deleted its items on the first call after a reinstall, and only if that call came before anything read a mnemonic — boot ordering nobody controls. The mnemonics are cleared by `clearUninstallSurvivingCredentials` instead.

## What is in here

**The migration itself.** Both slots read through the helper from blinkbitcoin/blink-mobile#4210, with two deliberate differences from the six session slots:

- **The protection class does not change.** They are already `WHEN_UNLOCKED_THIS_DEVICE_ONLY` and stay there, so the silent-push window accepted for the session slots never applies to key material.
- **The legacy copy is kept** (`deleteLegacyOnMigrate: false`). A mnemonic is immutable per account, so the two stores cannot diverge and the stale-value argument behind erasing the others does not hold. What it buys is a downgrade or rollback that still finds the value. The copies go in the step 4 purge, never on the read path.

**An account list stored beside the values.** The wipe has to reach keys whose names carry an account id, the app's own index lives in AsyncStorage and a reinstall clears it, and enumerating the keychain is the one refactor this area must not have: on iOS `getAllGenericPasswordServices` returns the legacy library's service, and a reset on it deletes every legacy item at once. A list we maintain ourselves leaves the wipe with a fixed slot list and no enumeration.

**A sweep after boot.** Read-through alone migrates a value when something reads it, so an account the user never opens would still be in the old store when it is dropped, and its mnemonic would become unreachable. The sweep enumerates the index and reads each one, off the critical path via `InteractionManager` and never awaited.

**Three contract fixes**, without which the migration is not safe to ship, because read-through widens the window where a read can fail and each of these scored that failure as a fact about the user's wallet:

- `use-sdk-lifecycle.ts` mapped a failed read to `Unavailable`, which reads as "you have no wallet". Now `Error`, which the retry and backoff wiring already listens on.
- `findSelfCustodialAccountByMnemonic` could not tell a failed read from "not this account", so restoring a wallet already on the device reported no match and the caller created a second account for the same seed.
- The create/restore rollback discarded the delete result, so a failed rollback left key material under an account id the index never learned about.

**A closed invariant.** With the mnemonics behind the helper, `secureStorage.ts` no longer imports the legacy library at all, so its `no-restricted-imports` exception is gone and `legacy-key-store.ts` is the only door left. That file's comment said the invariant was incomplete and named this step as what would finish it.

## Verified on device

Two seeded self-custodial accounts, upgrading over the existing install:

- `ttest1` opened with its own wallet (26,056 sat and its pending deposit), `ttest2` with a different one (65,112 sat). Two accounts is the point: a mnemonic migrated under the wrong key produces a plausible-looking wallet, and only the switch catches it.
- `files/datastore/RN_KEYCHAIN.preferences_pb` went from 0 bytes to holding both accounts' `mnemonic` and `mnemonic_network` slots plus the account list — so the sweep migrated the account that was never opened, not just the active one.
- `shared_prefs/secret_shared_prefs.xml` kept all 6 entries: nothing was erased, as intended.
- AsyncStorage: 19 keys before and after, none lost, `persistentState` and `selfCustodialAccountIndex` byte-identical.

The device run is also what caught a real defect: the account list was only written when a wallet was created or restored, never when an existing one migrated — so for every upgrading install, which is the whole point of this change, the list would have been empty and the wipe would have had nothing to clear. Fixed, and confirmed on the device.

`tsc`, `eslint` and `prettier` clean. 3118 tests across the touched areas and 5040 across the rest, all passing; `secureStorage.ts` at 100% statements and branches.

## Accepted limitations

- The account list is read-modify-written without a lock. Two accounts created at exactly the same moment could leave one unrecorded, and its mnemonic would then survive a reinstall. Same outcome as today's behaviour, and the flows that write it are sequential.
- A failed read and an absent mnemonic remain indistinguishable to callers other than the three fixed here.
- Keystore key invalidation on Android (changing or removing the device lock) can make the new store unreadable. OEM-variable and not reliably reproducible on an emulator.
- The migration is not atomic across slots: a kill mid-migration leaves some moved and some not. Per-slot read-through is idempotent, so the next read finishes the job.


grimen (changes requested):
## Findings

**1. `loadString` conflates "no data" with "read failed," letting a transient storage error trigger destructive mnemonic deletion (blocking).**

`loadString(PERSISTENT_STATE_KEY)` returns `null` both when the key genuinely doesn't exist and when `AsyncStorage.getItem()` throws. Both cases collapse into `MigrationStatus.NoData`, which is the same signal `clearUninstallSurvivingCredentials()` treats as "this is a fresh install, wipe every tracked mnemonic."

*Example:* A transient `AsyncStorage` read failure (disk pressure, OS-level storage hiccup) on an existing install produces the exact same `null` as an actual fresh install. The app deletes every mnemonic it was supposed to protect, even though the user never reinstalled and has funds tied to that key material.

*Proposal:* Distinguish "key absent" from "read threw" at the `loadString` layer (e.g. propagate the error or return a distinct status) so `clearUninstallSurvivingCredentials()` only wipes on an unambiguous fresh-install signal, not on any `null`.

**2. Reinstalls from before `MNEMONIC_ACCOUNTS` existed have no tracking list, so legacy mnemonics are silently left behind (blocking).**

An absent `MNEMONIC_ACCOUNTS` list is treated the same as an empty list. That's correct for a genuine fresh install, but a reinstall of a release that predates this tracking mechanism also has an absent list — and the old global sweep that used to catch this case has been disabled in this PR.

*Example:* A user who installed a pre-`MNEMONIC_ACCOUNTS` build, then reinstalls after upgrading, has real mnemonics in the keychain but no tracking list. `clearUninstallSurvivingCredentials()` sees an empty list, does nothing, and those legacy mnemonics permanently survive the uninstall/reinstall — the opposite failure mode from finding 1, but the same root cause: no way to tell "empty because fresh" from "empty because untracked."

*Proposal:* Add an explicit strategy for credentials created before the tracking index existed, for example falling back to the old blanket sweep whenever the tracking list is absent (as opposed to present-but-empty).

**3. `trackMnemonicAccount`/`untrackMnemonicAccount` read-modify-write `MNEMONIC_ACCOUNTS` outside the slot's serialization queue (blocking).**

Every other mutation of a keychain slot goes through `onSlot()` to serialize concurrent access, but these two functions call `secureRead`/`secureWrite`/`secureRemove` directly, so the read and the write are not atomic with respect to other callers.

*Example:* Boot-time `sweepMnemonicMigration` tracking one account can race a concurrent `setMnemonicForAccount` call from a user creating or restoring a wallet. Both read the same stale list and both write their own version back; whichever write lands second overwrites the first, silently dropping an account ID from the list. Since `clearUninstallSurvivingCredentials()` trusts this list to decide what to wipe, the dropped account's mnemonic can survive an uninstall it should have been deleted by, or be deleted incorrectly depending on which side loses the race.

*Proposal:* Wrap the whole read-modify-write transaction in a single `onSlot(MNEMONIC_ACCOUNTS, ...)` call rather than serializing only the individual read and write.

**4. `readMnemonicAccounts()` treats malformed data as an empty list instead of failing safely (blocking).**

Valid JSON that isn't an array, and arrays containing non-string entries, both fall through to `[]`. The function's own invariant says unsafe/unexpected shapes should return `null` so callers can detect corruption, which is exactly what the sibling unparseable-JSON branch already does correctly.

*Example:* If the stored value is ever a non-array or contains a non-string entry (from corruption or a future bug), `readMnemonicAccounts()` reports `[]`. `clearUninstallSurvivingCredentials()` only escalates via `onFailure('mnemonic account list')` on `null`, so with `[]` it proceeds as if there were zero tracked accounts, skipping the mnemonic wipe with no error reported.

*Proposal:* Return `null` for any shape other than a well-formed string array, matching the existing unparseable-JSON path, so cleanup fails safely and the corruption gets reported instead of silently ignored.

kngako (changes requested):
## Findings

### Finding 1 — A mnemonic whose tracking write fails is invisible to the reinstall wipe, forever [BLOCKING]

**When it comes up:** A user creates a self-custodial wallet. `setMnemonicForAccount` ([secureStorage.ts:697](app/utils/storage/secureStorage.ts:697)) writes the seed to the **new** store only — writes never read through, and `mnemonicSlotFor` never writes to the legacy store — then calls `trackMnemonicAccount`, which swallows every failure it can hit: `onSlot`'s timeout rejection is caught, and `secureWrite`'s `false` return is discarded ([secureStorage.ts:635-648](app/utils/storage/secureStorage.ts:635)). One keychain hiccup on the `mnemonicAccounts` slot and the id is never recorded. The user later uninstalls and reinstalls (or sells the phone). The wipe reads the list, the id is not in it, no `onFailure` fires because the list itself read fine, and `eraseEntireLegacyStore` cannot reach the value — it deletes generic passwords by service, and this seed is an *internet* credential. The seed phrase stays in the keychain of a device that has changed hands, while the app that could have deleted it has lost the AsyncStorage index that named the account. The comment at [secureStorage.ts:625](app/utils/storage/secureStorage.ts:625) argues this "is the behaviour before this list existed" — it is not: on `main` that mnemonic was in the legacy store, where `clearSecureKeyStore` swept it.

**Proposal:** Invert the order — track the id first, then write the value. The list is then a superset of what is stored, which is the safe direction: a spurious id costs the wipe one no-op `deleteMnemonicForAccount` (which already untracks ids whose value is gone), whereas a missing id costs an unwipeable seed. Concretely, in `setMnemonicForAccount` await `trackMnemonicAccount(accountId)` before `writeThrough`, and stop discarding `secureWrite`'s result inside `trackMnemonicAccount` — return a boolean and let `setMnemonicForAccount` report a failed track through `recordAppError` even when the value write succeeds, so the sweep has a reason to re-record it on the next boot. The spec at [secure-storage.spec.ts:313](__tests__/utils/secure-storage.spec.ts:313) currently pins the unsafe ordering and will need inverting with it.

### Finding 2 — The wipe that now destroys seed phrases is still gated on a heuristic this file documents as false-positive-prone, with no corroboration [BLOCKING]

**When it comes up:** `handleFreshInstall` fires whenever `migratePersistentState` answers `NoData` ([index.tsx:271](app/store/persistent-state/index.tsx:271)), and the comment above `clearUninstallSurvivingCredentials` already names the problem: "a throwing AsyncStorage read can produce [NoData] without a reinstall". Commit `c9cd78c60` closes the loudest route by adding `readString`, but the branch also changes what that branch *destroys*: on `main` the wipe stopped at the biometrics flag; here it deletes every tracked mnemonic **and** `resetGenericPassword`s the whole legacy store, so both copies of every seed go at once. Two routes into `NoData` remain open. `loadPersistentState` merges "key absent" with "blob parsed to a falsy value" into the same `data = null` ([index.tsx:265-277](app/store/persistent-state/index.tsx:265)) — `migratePersistentState`'s guard is `if (!data)` ([state-migrations.ts:518](app/store/persistent-state/state-migrations.ts:518)), so a stored `"null"`, `"0"` or `"false"` is scored as a fresh install and spends the wipe. And nothing at all corroborates the verdict before the mnemonic half runs: a user whose AsyncStorage is damaged rather than absent loses their funds, irreversibly, on a boot that gave them no chance to intervene.

**Proposal:** Two changes, both cheap. First, route "found but falsy" to `handleUnusableBlob` rather than `NoData` — keep `data` as a `{ present: boolean; value: unknown }` through the call, or have `loadPersistentState` return early on `read.status === "found" && !parsed`, so the fresh-install signal can only come from an absent key. Second, require positive evidence before the mnemonic half specifically: `listSelfCustodialAccounts()` already reads the AsyncStorage index, which a genuine reinstall clears, so a successful read returning a **non-empty** list is proof this is not a fresh install — skip the mnemonic loop and `eraseEntireLegacyStore`, report it, and let the session-credential half proceed as it does today. An index read that *fails* should also skip that half and retry next boot, since it proves nothing either way.

### Finding 3 — The new `Error` status for an unreadable mnemonic is a dead end, behind a retry button that does nothing [BLOCKING]

**When it comes up:** Commit `133c23c37` changes `initialize()` so a `failed` mnemonic read sets `ActiveWalletStatus.Error` instead of `Unavailable` ([use-sdk-lifecycle.ts:274](app/self-custodial/hooks/use-sdk-lifecycle.ts:274)), on the stated grounds that "Error is the status the retry and backoff wiring already listens on". It is not. `scheduleBackoffRetry` is reached from exactly one place — `refreshWallets`'s catch ([use-sdk-lifecycle.ts:175](app/self-custodial/hooks/use-sdk-lifecycle.ts:175)) — and its callback is `if (sdkRef.current) refreshWallets()`, which is null here because nothing ever connected. Meanwhile `Error` is in `SELF_CUSTODIAL_BLOCKED_STATUSES` ([offline-gate.tsx:12](app/self-custodial/components/offline-gate.tsx:12)), so the user is shown `PaymentOfflineNotice` — whose retry button calls `refreshWallets()` ([payment-offline-notice.tsx:23](app/self-custodial/components/payment-offline-notice.tsx:23)), which returns on its first line for the same reason. The provider's `retry` (`setRetryCount`, the one thing that would re-run `initialize()`) is consumed only by the create- and restore-wallet onboarding hooks. So a user whose keychain refused one read taps a retry button that provably cannot succeed, for the rest of the process; the old `Unavailable` was equally terminal but at least did not offer the button.

**Proposal:** Make `PaymentOfflineNotice`'s retry re-run the lifecycle rather than the refresh — take `retry` alongside `refreshWallets` from `useSelfCustodialWallet()` and call `retry()` when there is no connected SDK (`sdk === null`), falling back to `refreshWallets()` otherwise. Optionally add one automatic attempt for this specific branch by scheduling a `setRetryCount` bump through the existing `useBackoffRetry`, since the pre-first-unlock window the protection-class note describes resolves on its own within seconds. Either way, correct the comment: the backoff wiring listens on a live SDK, not on the status.

### Finding 4 — The sweep counts an unmigrated mnemonic as migrated, and its verdict is thrown away [NON-BLOCKING]

**When it comes up:** `sweepMnemonicMigration` scores `readMnemonicWithStatus(...) === "found"` as a migration ([account-index.ts:194](app/self-custodial/storage/account-index.ts:194)). But `runRead` returns the **legacy** value whenever it found one, regardless of whether the migrating `secureWrite` succeeded ([secure-store-migration.ts:240-248](app/utils/storage/secure-store-migration.ts:240)) — that is deliberate there ("migration bookkeeping must never cost availability"), but it means a device whose keychain refuses every write reports `{ status: "ok", migrated: 12 }` while nothing moved. The caller then discards the result entirely (`sweepMnemonicMigration().catch(() => {})`, [index.tsx:411](app/store/persistent-state/index.tsx:411)), so even an honest `incomplete` never reaches telemetry. Since #1163 gates dropping the legacy dependency on this migration being finished, the one signal that decision needs is both over-optimistic and unread.

**Proposal:** Have `readThrough` report whether the value is now in the new store — either a `migrated: boolean` alongside `found`, or a separate `status: "found-legacy"` — and count only that in the sweep, treating a legacy hit with a failed write as a failure rather than a success. Then consume the result at the call site: log a `sweep_incomplete` analytics event (or a `recordAppError`) on `{ status: "incomplete" }` next to the existing `legacy_key_store_hit` counter, so "has the sweep finished on real devices" is answerable from the same dashboard.

### Finding 5 — One `dedupKey` for every account makes the sweep's failure count invisible in telemetry [NON-BLOCKING]

**When it comes up:** Every per-account sweep failure is recorded under the constant `dedupKey: "storage-mnemonic-sweep-failed"` ([account-index.ts:189](app/self-custodial/storage/account-index.ts:189)). A user with five self-custodial accounts whose keychain is locked reports one error, not five, and — because the key is identical across boots — reports nothing at all on the second and later launches. The `failures` counter the function computes correctly is then dropped by the caller (Finding 4), so the only surviving trace understates a total migration failure as a single blip.

**Proposal:** Keep the dedup (it exists so a broken device does not flood Crashlytics) but move the count into the payload rather than the key: report once per sweep, after the loop, with `failures` and the index size as custom keys — `recordAppError(new Error(\`Mnemonic sweep incomplete: ${failures}/${result.entries.length}\`), { dedupKey: "storage-mnemonic-sweep-failed" })`. Never put the account id in either the key or the message, for the reason `keyClassOf` exists.

### Finding 6 — A malformed tracked list is never repaired, and the next write forgets every id inside it [NON-BLOCKING]

**When it comes up:** If `mnemonicAccounts` ever holds something that is not a list of strings, `untrackMnemonicAccount` deliberately leaves it alone ([secureStorage.ts:661](app/utils/storage/secureStorage.ts:661)) — correct, since rewriting it would forget accounts whose seeds are still stored. But `trackMnemonicAccount` does the opposite: on `malformed` it starts from `[]` ([secureStorage.ts:640](app/utils/storage/secureStorage.ts:640)), so the next wallet the user creates replaces the damaged list with a single-entry one and every id it held is gone for good. The wipe then reports a clean sweep over mnemonics it never looked at, which is precisely the outcome `MnemonicAccountsRead`'s three-way split was introduced to prevent. Nothing else writes this slot, so the state is hard to reach — but it is also unrecoverable once reached.

**Proposal:** Make the sweep the repair path. It already enumerates every account in the index and calls `rememberMnemonicAccount` for each; have `readMnemonicAccounts` returning `malformed` trigger a rebuild from that enumeration on the next boot rather than a silent reset on the next write — e.g. a `rebuildMnemonicAccounts(ids: readonly string[])` that replaces the slot wholesale, called by `sweepMnemonicMigration` only when the list reads `malformed`. Until then, `trackMnemonicAccount` should at minimum `recordAppError` before it discards a malformed list.

### Finding 7 — A transient AsyncStorage failure boots on defaults, and the first setting change overwrites the intact blob [NON-BLOCKING]

**When it comes up:** `handleUnreadableStore` is a real improvement — it no longer spends the reinstall wipe on a read that merely failed — but it returns `defaultPersistentState` ([index.tsx:246](app/store/persistent-state/index.tsx:246)), and nothing marks the session as untrustworthy afterwards. `defaultPersistentState` is three fields ([state-migrations.ts:493](app/store/persistent-state/state-migrations.ts:493)): no `activeAccountId`, no display currency, no account modes, no quiz or transaction-seen state. The user boots still signed in (the keychain token is kept) but with no account selected; they tap their wallet in the picker, `setActiveAccountId` calls `updateState`, `hasModified` flips, and `savePersistentStateBlob` writes defaults-plus-one-field over a blob that was perfectly intact and would have read fine on the next boot. `handleUnusableBlob` at least quarantines a copy first; this path has nothing to quarantine and so leaves none. The comment "This boot runs on defaults and the next one reads the blob again" is true only if nothing writes in between, which the very next tap does.

**Proposal:** Carry the failure into the provider and suppress persistence for the session — return a `loadFailed: true` alongside `state`/`persistedToken`, hold it in a ref, and have the `persistentState` save effect skip `savePersistentStateBlob` (but not the keychain token write, which is independent) while it is set. In-memory changes still work for the session; the blob is simply not rewritten from a state that was never loaded. A one-line spec that asserts no `saveJson` after an `updateState` on a failed-read boot pins it.

---

## Test Coverage Gaps

1. **The sweep's wiring into boot is untested.** Deleting the whole `InteractionManager.runAfterInteractions(() => sweepMnemonicMigration()...)` block from `PersistentStateProvider` ([index.tsx:410](app/store/persistent-state/index.tsx:410)) leaves `persistent-state/index.spec.tsx` at 44/44 green (mutation-verified). `mockSweepMnemonicMigration` is declared and given a default resolution but never asserted on. Since the sweep is the *only* thing that records mnemonic accounts for upgrading installs, and the reinstall wipe reaches nothing without those records, this is the highest-value missing assertion on the branch: one `await waitFor(() => expect(mockSweepMnemonicMigration).toHaveBeenCalled())` after the provider mounts.

2. **Nothing covers a mnemonic write that succeeds with a tracking write that fails** — Finding 1's exact hole. `secure-storage.spec.ts:313` covers the opposite (value write fails ⇒ no track), and `:281` covers the happy path, but no case makes the `mnemonicAccounts` write reject while the `mnemonic:alice` write succeeds and then asserts what `clearUninstallSurvivingCredentials` does about it. That case is what would have caught the ordering.

3. **The sweep spec mocks `KeyStoreWrapper` wholesale, so "read from legacy, migrating write failed" cannot be distinguished** (Finding 4). Every sweep case feeds `readMnemonicWithStatus` a canned `{ status: "found" }`, so the `migrated` count is only ever tested against a mock that cannot fail halfway. An integration case in `secure-storage.spec.ts` — legacy hit, `setInternetCredentials` rejecting, assert the value still comes back and the legacy copy is untouched — would pin the real contract the sweep is counting.

4. **No test distinguishes `NoData` from an absent key and `NoData` from a falsy blob** (Finding 2). `persistent-state/index.spec.tsx` covers absent-key ⇒ wipe and failed-read ⇒ no wipe, both well, but nothing stores `"null"` or `"0"` and asserts that the wipe does *not* fire. Adding those two cases is what would stop the falsy route from reopening later.

5. **Nothing asserts the `Error` status is recoverable** (Finding 3). `use-sdk-lifecycle.spec.ts` asserts it "falls to Error, not Unavailable, when the keystore cannot answer" — a real assertion about the value, but it stops there. No spec drives the retry affordance the user actually sees, so a status with no working retry passes. A case that sets the status, renders `OfflineGate`, presses `payment-offline-retry` and expects `readMnemonicWithStatus` to be called a second time would fail today and pass after the fix.

6. **No sweep case for an index entry whose mnemonic is `absent`.** The `else` branch in `sweepMnemonicMigration` handles it (no count, no `rememberMnemonicAccount`, but the network marker is still read), and nothing pins that — an account listed in the index but never given a seed is a real state, reachable through a failed create that rolled back after `addSelfCustodialAccountId`.

7. **Two `eraseEntireLegacyStore` cases test states iOS cannot produce.** `resetGenericPasswordForOptions` resolves `@(YES)` or rejects ([RNKeychainManager.m:522-535](node_modules/react-native-keychain/ios/RNKeychainManager/RNKeychainManager.m:522)), so "passes a false resolution through as a failed erase" guards against a return value the native module never emits. Harmless as defence, but it is not coverage of a reachable path and should not be read as such.

---

## Code Smells

1. **`InteractionManager` is deprecated in this React Native version and warns on every boot.** The test run prints `InteractionManager has been deprecated and will be removed in a future release … use 'requestIdleCallback' instead`, pointing straight at [index.tsx:410](app/store/persistent-state/index.tsx:410). The intent ("after the interactions this boot has queued") is right; `requestIdleCallback` expresses it without adding a new call site to an API on its way out, and without a console warning on the boot path of every session.

2. **Two comments now assert things the code does not do.** [secureStorage.ts:625](app/utils/storage/secureStorage.ts:625) says an untracked id "is the behaviour before this list existed" — on `main` that mnemonic was in the legacy store and the module's own sweep cleared it (Finding 1). [use-sdk-lifecycle.ts:271](app/self-custodial/hooks/use-sdk-lifecycle.ts:271) says "Error is the status the retry and backoff wiring already listens on" — the backoff is armed from `refreshWallets` and gated on a live SDK (Finding 3). The commentary on this branch is unusually good and load-bearing, which is exactly why a claim that no longer holds is expensive: the next reader will trust it.

3. **`SweepResult` is a public discriminated union nothing consumes.** Exported from [account-index.ts:155](app/self-custodial/storage/account-index.ts:155), returned faithfully, asserted in the spec, and dropped by its only caller. Either wire it up (Finding 4) or stop exporting the type, so it does not read as a contract someone is honouring.

4. **`clearPinFailureState` can *write* during the reinstall wipe.** [secureStorage.ts:334-356](app/utils/storage/secureStorage.ts:334) falls back to writing a cleared value when both erases fail, which is right for its normal caller — but `clearUninstallSurvivingCredentials` invokes it through `removeWithRetry`, so a wipe can finish by creating a fresh `pinFailureState` internet credential that then outlives the *next* uninstall. Passing a flag (or calling `migratedErase` directly from the wipe) keeps "erase everything" from ending in a write.

5. **The `.eslintrc.json` `plugins` reformat is unrelated churn.** Splitting the one-line array across seven lines and dropping a blank line ([`.eslintrc.json`](.eslintrc.json)) is a Prettier pass riding in a security branch; the substantive change — removing `secureStorage.ts` from the `no-restricted-imports` override, now that it no longer imports the legacy library — is correct and is two lines. Separating them would make the security diff reviewable on its own.

6. **`secureStorage.ts` is 747 lines and now owns three unrelated concerns.** Session slots (token, profiles, PIN, lockout, biometrics), the reinstall wipe, and the per-account mnemonic store with its own tracked-account transaction log all live in one class. The mnemonic block — `mnemonicKeyFor` through `setMnemonicNetworkForAccount`, plus `readMnemonicAccounts`/`trackMnemonicAccount`/`untrackMnemonicAccount` — is self-contained and has a different protection class, a different legacy policy and a different consumer; lifting it into `mnemonic-store.ts` beside `secure-store.ts` would leave the wipe importing it explicitly, which is the dependency the wipe's own comment says it wants to make visible.

grimen (changes requested):
## Findings

**1. The delete flow still treats a failed mnemonic read as "no mnemonic", so it skips the has-funds warning and then erases the seed.**
`probeSelfCustodialAccountWallets` (`probe-account-wallets.ts:43`) reads through `getMnemonicForAccount`, which still collapses `failed` into `null`, so the probe returns `NoMnemonic`. `profile-row.tsx:128` answers `NoMnemonic` with the plain confirm dialog (`setConfirmVisible(true)`) rather than the has-funds warning, and `handleConfirm` then runs `deleteWallet` → `deleteMnemonicForAccount`, which now clears both stores. A delete does not decrypt, so it can succeed where the read just failed. The PR fixes three consumers because read-through widens the window in which a read can fail. The body lists the rest as an accepted limitation, and this one is the most destructive of them.

*Example:* A non-active account holds funds. Its mnemonic read fails once, from the 30 s slot timeout or an Android `E_KEYSTORE_ACCESS_ERROR`. The user taps Remove and gets the plain confirm instead of the funds warning, confirms, and both copies of the seed are gone.

*Proposal:* one option is to probe through `readMnemonicWithStatus` and map `failed` to `ProbeFailed`, which `surfaceProbeFailure` already handles. `withMigrationSdk` (`migration-transfer-request.ts:101`) has the same collapse, less destructively (it routes to support as `SelfCustodialAccountMissing`), and the same change fits there.

**2. One unreadable account makes every restore on the device fail, including restores of unrelated seeds.**
`findSelfCustodialAccountByMnemonic` returns `ReadFailed` at the first entry whose read fails, before it compares the later entries, and `use-restore-wallet.ts:89` throws on `ReadFailed`, which shows `restoreFailed`. A slot whose read keeps failing therefore blocks every restore until that account is deleted. The Android Keystore invalidation listed under accepted limitations is one route to that, and it makes every entry fail at once. The PR body names the backup phrase as a wallet's recovery path, and this blocks it in exactly the case where the user needs it.

*Example:* An Android user changes the device lock and every mnemonic read starts failing. They enter their written-down phrase on the restore screen, and restore fails every time with a generic error.

*Proposal:* one option is to finish the scan so a readable match still wins, and when nothing matched but some entry failed, let the user choose to continue. A duplicate account for the same seed can be removed later; a restore that cannot run cannot be worked around. If blocking the restore is deliberate, to rule out duplicates, a comment noting the trade-off would help.

**3. The boot sweep decrypts every account's seed into the JS heap on every launch, and on iOS each launch while the device is locked reports every account as a failed read.**
`sweepMnemonicMigration` calls `readMnemonicWithStatus` for every indexed account on every boot, with no completion marker. For a slot that has already migrated, `runRead` returns the decrypted value from the new store. The sweep only looks at `status`, but the phrase is now a JS string, which cannot be zeroed, even for accounts the user never opens. Before this PR a seed entered JS only when its account was used. The same read fails on iOS while the device is locked, because `WHEN_UNLOCKED_THIS_DEVICE_ONLY` gives `errSecInteractionNotAllowed` (`secure-store.ts:52`). So every locked launch counts every account as a failure and raises `Mnemonic sweep incomplete: N/N`.

*Example:* A user with four self-custodial accounts opens the app normally, and all four seeds are decrypted into memory although one is active. A background launch while the phone is locked records a `4/4` non-fatal.

*Proposal:* one option is to probe through `existsThrough` instead. `runExists` answers `yes` from `hasInternetCredentials` without decrypting, and before first unlock too, and falls through to `runRead` (which migrates) only when the new store is empty. The sweep would keep re-recording every indexed account each boot, which is what repairs a failed or reset tracking write, while no longer decrypting anything that has already migrated. The network marker can go the same way.

## Test Coverage Gaps

- Nothing covers the delete flow with a failed read: `probe-account-wallets.spec.ts:76` pins `NoMnemonic` only for an absent value, and no `profile-row` case asserts that the funds warning still shows when the read fails (finding 1).
- `findSelfCustodialAccountByMnemonic` has "a mnemonic read fails → read-failed" (`account-index.spec.ts:398`), but no case where a failing entry comes before a matching one, which is where the early return costs a match (finding 2).
- No sweep case covers a locked device, where every read fails and the whole index is reported as incomplete (finding 3).

esaugomez31:
Both rounds of feedback have been applied:

## Findings

1. **A mnemonic whose tracking write fails is invisible to the wipe** — Fixed. The id is recorded before the value is written, so the list is a superset of what is stored; `trackMnemonicAccount` returns a boolean and a mnemonic stored without being tracked is reported.
2. **The wipe is gated on a heuristic with no corroboration** — Fixed. A blob that parses to an empty value goes to the damaged-blob path instead of the fresh-install branch, and the key-material half is split out and waits for the account index to be provably absent. Absence, not emptiness: an unrecognisable index degrades to zero entries.
3. **The `Error` status is a dead end behind a retry that does nothing** — Fixed. The offline notice takes `retry` alongside `refreshWallets` and re-runs the lifecycle when no SDK ever connected, falling back to a refresh otherwise.
4. **The sweep counts an unmigrated mnemonic as migrated, and its verdict is thrown away** — Fixed by finding 10's change: "migrated" now means present in the new store, checked without the migrating fallback. An incomplete sweep is reported with the share left behind.
5. **One `dedupKey` for every account hides the failure count** — Fixed. One non-fatal per sweep carrying `failures/total`; each account's cause survives as a breadcrumb.
6. **A malformed tracked list is never repaired** — Fixed. The sweep rebuilds it from the index, in one slot-queue turn so a concurrent write is not overwritten. An empty index rebuilds nothing rather than clearing the slot.
7. **A transient read boots on defaults and the first change overwrites the blob** — Fixed. The read is retried once; if it still fails the session holds back blob writes. The keychain token keeps saving, since it failed independently.
8. **The delete flow treats a failed read as "no mnemonic" and erases the seed** — Fixed. The probe reads with status and maps `failed` to `ProbeFailed`; the same change applies to the migration request. The error handed to the caller carries no account id.
9. **One unreadable account makes every restore fail** — Fixed. The scan finishes, so a readable match still wins; only when nothing matched and entries were unreadable does it answer no match, which is reported. The trade-off is a possible duplicate account, taken deliberately: a duplicate is removable, a blocked restore is not.
10. **The sweep decrypts every seed on every launch, and reports every account as failed while locked** — Fixed. It asks whether each mnemonic exists instead of reading it, which answers from the new store without decrypting and before first unlock, falling through to the migrating read only when the new store is empty. The network marker went the same way.

## Test Coverage Gaps

All covered, except where noted:

- The sweep's wiring into boot, plus a rejecting sweep not breaking the boot.
- A mnemonic write that succeeds with a tracking write that fails.
- A legacy hit whose migrating write was refused, pinning that "found" is not "moved".
- `NoData` from an absent key distinguished from `NoData` from a blob that parses to nothing.
- The retry affordance on the `Error` status.
- An index entry whose mnemonic is absent.
- The delete flow with a failed read, and a failing entry ordered before a matching one.
- A device where every probe fails, which is the locked-launch case.
- The two `eraseEntireLegacyStore` cases that iOS cannot produce are kept but annotated as defence against a library change, not coverage of a reachable path.

## Code Smells

1. **`InteractionManager` is deprecated and warns on every boot** — Fixed. Scheduled on the first idle window instead, bounded by a timeout so a busy boot still sweeps.
2. **Two comments assert things the code does not do** — Fixed. Both rewritten; the claim that the backoff wiring listens on the status was false, and the claim that an untracked id matched previous behaviour was too.
3. **`SweepResult` is a public union nothing consumes** — Fixed. No longer exported.
4. **`clearPinFailureState` can write during the wipe** — Fixed. The wipe erases both keys directly, so it cannot end by creating an entry that outlives the next uninstall.
5. **The `.eslintrc.json` reformat is unrelated churn** — Fixed. Reverted; only the override change remains.
6. **`secureStorage.ts` owns three unrelated concerns** — Agreed, but deferred to a PR of its own. The file already mixed session slots and mnemonics before this branch; the split is a refactor of structure that predates it, and moving it here would triple the diff under review.

## Also changed

Two things found while applying the above, outside both rounds:

- The wipe erased the legacy store after the per-account deletes, but those deletes refuse to touch the new store until the legacy copy is provably gone — so a legacy store that would not answer failed every one of them, with the erase that would have unblocked them arriving too late. It now runs first.
- Three writers of the tracked list wrote after an `await` without re-checking whether their slot turn had been abandoned, so a timed-out task could land a stale snapshot over a concurrent write and drop an id the wipe is the only reader of.

## Known limitation

If the index cannot answer on a genuine reinstall, the erase is deferred rather than performed. That boot holds back its blob write so the next one reaches the same branch and asks again, which is the retry rather than a durable marker — a marker would have to live either where an uninstall clears it, making it useless, or where it survives one, making it dangerous.

kngako (changes requested):
Review of the 22 commits on claude/branch-code-review-580be3 — 6 findings (2 blocking), 8 test coverage gaps, 8 code smells.

---

## Findings

### 1. `holdBlobWrites` turns a failed wipe into a sign-out on every launch — [BLOCKING]

`handleFreshInstall` ([index.tsx:213](app/store/persistent-state/index.tsx:213)) runs `clearUninstallSurvivingCredentials` unconditionally, then holds the blob back whenever the key-material half did not finish ([:254](app/store/persistent-state/index.tsx:254), [:277](app/store/persistent-state/index.tsx:277)). The hold is deliberate and its reasoning is sound in isolation — the erase stays owed only while the blob is absent. What it does not account for is that the same absent blob re-triggers the *session* half. Before this branch the loop self-terminated: the first `updateState` wrote the blob and boot 2 was ordinary. Now, on the held paths, it cannot terminate.

**When it comes up:** a user reinstalls on a device with two self-custodial accounts, and one `deleteMnemonicForAccount` fails twice — one stubborn keychain item is enough. `onFailure("mnemonic")` fires, `erased` goes false, the blob is held. They onboard and sign in. Next launch: blob still absent, so the token, session profiles, PIN and biometrics flag are wiped again and they are back at the login screen. Probe A shows exactly this over three boots (`[true, true, true]`, blob never written), and shows the baseline terminating after one boot when the erase succeeds. The `presence === Unknown` path is worse: it holds the blob without ever attempting the erase, so the retry it is holding for never happens. Each boot also files a fresh `alwaysRecord` Crashlytics report, so the device emits one per launch indefinitely.

**Proposal:** stop making the retry depend on the blob staying absent. Carry the owed work *inside* the blob instead — a `pendingReinstallKeyMaterialWipe` flag on `PersistentState`. A genuine reinstall clears the blob and re-triggers everything as it does today; a held boot writes the blob, which ends the session-wipe loop, and the next boot reads the flag and retries only the key-material half. That is strictly more information than the blob's absence carries, and it removes the objection recorded at [:272](app/store/persistent-state/index.tsx:272) — a marker that an uninstall clears is precisely what is wanted here, because a reinstall should re-run the wipe. If the flag is too large a change for this branch, bound the hold instead (three boots, then write the blob and report loudly); an unbounded hold is the part that cannot ship.

### 2. A tracked list that will not parse makes the mnemonic wipe a permanent no-op — [BLOCKING]

`clearUninstallSurvivingKeyMaterial` reaches migrated mnemonics only through `readMnemonicAccounts` ([secureStorage.ts:626](app/utils/storage/secureStorage.ts:626)); on anything but `ok` it calls `onFailure("mnemonic account list")` and skips the loop entirely ([:586](app/utils/storage/secureStorage.ts:586)). `failed` and `malformed` are handled identically, but they are opposites: a `failed` read retries fine on the next boot, while `malformed` is permanent, because the only repair — `rebuildMnemonicAccountsIfMalformed` — is driven by `sweepMnemonicMigration` and returns `"failed"` when the index is empty ([secureStorage.ts:681](app/utils/storage/secureStorage.ts:681)). A reinstall has just cleared that index.

**When it comes up:** a device whose `mnemonicAccounts` value is truncated reinstalls. The legacy copies go (`eraseEntireLegacyStore` still runs first), but every mnemonic the sweep had already migrated stays readable in the new store — Probe C shows `alice`'s phrase still returning from `getMnemonicForAccount` after the wipe, with nothing reported but a generic `"mnemonic account list"` label. The device then enters Finding 1's boot loop. When the user finally onboards, the sweep rebuilds the list from the *new* account ids, the list stops being malformed, the hold lifts — and the previous owner's seed phrases are orphaned in the keychain for good, with no code path left that can name them.

**Proposal:** remove the all-or-nothing parse. Store the list as newline-separated ids rather than JSON, so a damaged value costs at most one line and the `malformed` state stops existing — which also retires `trackMnemonicAccount`'s "rewrite from scratch and forget every id" branch ([:730](app/utils/storage/secureStorage.ts:730)), the other place this value's fragility is load-bearing. If JSON has to stay, then at minimum split the two cases: keep the previous good value in a second slot and fall back to it on `malformed`, and have the wipe quarantine the raw damaged bytes so the ids remain recoverable by hand. Either way `onFailure` should distinguish "could not read the list" from "read it and it was garbage", because only one of them will fix itself.

### 3. The session half has no legacy-store unblock, so a reinstall can leave the previous owner signed in — [NON-BLOCKING, pre-existing]

`clearUninstallSurvivingKeyMaterial` runs `eraseEntireLegacyStore` first and says why: `removeThrough` refuses to empty the new store until the legacy copy is provably gone, so a legacy store that cannot answer would fail every delete ([secureStorage.ts:566](app/utils/storage/secureStorage.ts:566)). Every slot in `clearUninstallSurvivingCredentials` goes through the same `migratedErase` → `removeThrough` path and gets no such unblock, and it runs first.

**When it comes up:** a reinstall on an iOS device where the legacy library rejects reads with anything other than 404 — an entitlement fault, a keychain the OS is refusing to query. Probe B: all six session slots report failure, no new-store slot is emptied, and `getActiveToken()` still returns the previous owner's token afterwards. The new owner boots into someone else's session, which is the precise outcome the wipe exists to prevent. This is inherited from `main` (the session slots moved in #1161), but this branch is where the mechanism was diagnosed and solved, and it now holds a fixed instance of the same bug twenty lines from an unfixed one.

**Proposal:** move the presence check ahead of both halves in `handleFreshInstall`, and on `Absent` run `eraseEntireLegacyStore` once before either. That keeps the legacy mnemonic copies intact on the `Present` path — the rollback guarantee `mnemonicSlotFor` documents — while unblocking the session half on the only path where it matters. If the ordering is too invasive, give the wipe a `removeThrough` variant that accepts an unprovable legacy copy for the session slots specifically: they are `AFTER_FIRST_UNLOCK` and they mutate, so a resurrected stale token is a revoked token, which is the trade `deleteLegacyOnMigrate: true` already makes for them.

### 4. The migration transfer path reports the raw keychain error and still routes the user to support — [NON-BLOCKING]

`withMigrationSdk` now returns `toSdkFailure(stored.err)` on a failed read ([migration-transfer-request.ts:105](app/self-custodial/migration-transfer-request.ts:105)). Two things follow that the comment above it does not claim. The error reaches `reportError("Migration receive check", err)` ([use-migration-receive-confirmation.ts:157](app/screens/account-migration/hooks/use-migration-receive-confirmation.ts:157)) with its cause intact — while the two sibling call sites changed in the same commit both drop it, on the stated grounds that a keychain error can carry the server string, which ends in the account id ([probe-account-wallets.ts:50](app/self-custodial/probe-account-wallets.ts:50), [validate-network.ts:20](app/self-custodial/providers/validate-network.ts:20)). And `classifySdkError` returns `Generic` for anything that is not an `SdkError`, so a keychain rejection lands on `MigrationSdkStatus.Failed` — the settled status that hands the user to support, per `toSdkFailure`'s own docblock.

**When it comes up:** a user runs the custodial-to-self-custodial migration while the keychain is momentarily refusing reads. The comment says this is "a retryable failure, not a missing key", but only `ConnectionError` pauses on the shared retry; `Failed` ends the flow at `MigrationSupportReason.TransferFailed`. The support reason is now honest, which is a real improvement over `SelfCustodialAccountMissing`, but nothing retries, and the account id may travel to Crashlytics on the way.

**Proposal:** return `toSdkFailure(new Error("Mnemonic read failed"))` so the cause is dropped the same way the two siblings drop it — `classifySdkError` already scores the real error as `Generic`, so no classification is lost. Then decide explicitly whether a keychain read failure should be retryable: if it should, return `ConnectionError` for it rather than routing through `toSdkFailure` at all, and say so in the comment. Add the `probe-account-wallets.spec.ts` sibling assertion that the account id stays out of the error the caller is handed.

### 5. The new `Error` status has no automatic recovery, and its manual one is behind a gated screen — [NON-BLOCKING]

A failed mnemonic read now sets `ActiveWalletStatus.Error` instead of `Unavailable` ([use-sdk-lifecycle.ts:276](app/self-custodial/hooks/use-sdk-lifecycle.ts:276)), which is right — `Unavailable` reads as "there is no wallet here". The comment is candid that recovery is not automatic. But the one automatic trigger that exists, the `AppState → active` handler, calls `refreshWallets`, which returns on its first line without an SDK ([:119](app/self-custodial/hooks/use-sdk-lifecycle.ts:119)); the backoff retry is gated the same way. `dfff7e4d4` adds the only working recovery, and it lives on `PaymentOfflineNotice`, which `OfflineGate` renders on Scan, Send and Merchant Selection.

**When it comes up:** a keystore hiccup at launch leaves a self-custodial user on Home with a wallet that reports an error. Backgrounding and reopening the app — the first thing anyone tries — does nothing, because `refreshWallets` has no SDK to refresh. They have to navigate into Send or Scan to find a Retry button, or kill and relaunch the app.

**Proposal:** apply the same branch `PaymentOfflineNotice` now makes to the AppState handler — when the app becomes active with no SDK and a status in the blocked set, call `retry()` instead of `refreshWallets()`. It is the same two-line condition, it reuses the `retryCount` wiring already in `wallet.tsx`, and it turns the most common user reflex into the recovery the branch built.

### 6. The reinstall wipe's docblock still describes the pre-split design — [NON-BLOCKING]

`clearUninstallSurvivingCredentials`'s docblock says "the mnemonics are removed here instead" and "Mnemonics are included, and reached through the account list stored beside them" ([secureStorage.ts:476](app/utils/storage/secureStorage.ts:476), [:494](app/utils/storage/secureStorage.ts:494)). `3ceb88829` moved all of that into `clearUninstallSurvivingKeyMaterial`; the method it annotates now touches no mnemonic at all. `legacy-key-store.ts:19` carries the same error, attributing the deliberate mnemonic clear to `clearUninstallSurvivingCredentials`.

**When it comes up:** the split is the whole safety property of this design — one half is unconditional and costs a re-login, the other waits for a second witness because it costs someone's money. The next person to touch this reads the docblock on the unconditional half, believes it already handles mnemonics, and either adds a mnemonic slot to the wrong list or removes the presence gate as redundant. Both are one-line edits with unrecoverable consequences.

**Proposal:** rewrite the two paragraphs to describe what the method now does — session credentials only, unconditional, costs a re-login — and give the split itself one statement somewhere it cannot drift, most naturally in the `clearUninstallSurvivingKeyMaterial` docblock which already argues the asymmetry. Fix the `legacy-key-store.ts` attribution in the same pass, since both were written before the split and neither was revisited after it.

---

## Test Coverage Gaps

1. **No spec crosses a boot boundary.** The persistent-state spec pins the hold *within* one boot — "the blob stays absent, so the next boot is a fresh install again" is asserted as a comment, not as a test. Probe A is three renders in a loop with the keychain modelled as a variable; it takes about thirty lines and it is the only thing that shows Finding 1. Without it, the suite's strongest statement about the held paths is that nothing was written, which is also what a correct implementation looks like.

2. **`clearUninstallSurvivingKeyMaterial`'s failure cases assert the report, never the survivor.** "reports a list it cannot parse rather than reading it as no accounts" and its two siblings check `onFailure` and stop there. Probe C shows the fact that matters — the mnemonic is still readable afterwards — and no existing case would fail if the skip silently became a delete of the wrong slot, or if the loop ran over an empty list and reported success.

3. **Nothing pins the session half against a legacy store that cannot answer at all.** `legacyRefusesToDrop` exists in the spec and is used one slot at a time, which is the right shape for "this slot is stubborn". The scenario in Finding 3 is different — every slot fails together and the previous owner's token survives — and it has no case. This is the gap that let the asymmetry with the key-material half go unnoticed.

4. **`migration-transfer-request.spec.ts` has the failed-read case but not its privacy sibling.** `probe-account-wallets.spec.ts` gained "keeps the account id out of the error it hands the caller" in the same commit; the migration spec gained only "reports a failed read as a retryable failure". The missing assertion is exactly the one that would have caught Finding 4, and the test name asserts a retryability the code does not deliver.

5. **`rebuildMnemonicAccountsIfMalformed` is never tested from a reinstall-shaped state.** It has good coverage for "rebuilds from the ids it is given", "never clears the slot when there is nothing to rebuild from" and so on — but nothing combines an empty index with a malformed list, which is the state a reinstall produces and the one where the repair silently cannot help. A case asserting that combination returns `"failed"` *and* saying in its name why that is a dead end would have surfaced Finding 2 from the repair side.

6. **The Android shape of the key-material wipe is untested.** `legacy-key-store.spec.ts` pins `eraseEntireLegacyStore`'s Android no-op, but the `clearUninstallSurvivingKeyMaterial` describe only ever runs the iOS path. Nothing states that an Android reinstall still clears the tracked mnemonics from the new store even though the legacy erase did nothing — which is the platform where the whole legacy-store argument is a no-op and the tracked list is the only mechanism.

7. **`PaymentOfflineNotice`'s branch is tested against a mock that decides the answer.** Both directions are covered, but `useSelfCustodialWallet` is mocked wholesale, so nothing pins the premise the fix rests on: that `sdk` really is null in the statuses `OfflineGate` shows for a wallet that never started. If a future change left a stale SDK on the ref through an `Error` transition, the fix would quietly become a no-op and every test here would still pass.

8. **The sweep's steady-state cost is unasserted.** "gives the same answer on a second run" covers the result; nothing covers the work. A case pinning that a fully migrated index performs no writes on the second run would turn the intended no-op into a property, and would catch a regression where `rememberMnemonicAccount` started rewriting the list on every boot.

---

## Code Smells

1. **`findSelfCustodialAccountByMnemonic`'s new comment describes a behaviour that only existed inside this branch.** "One entry that will not answer used to end the scan, so a single damaged slot made every restore on the device fail" ([account-index.ts:179](app/self-custodial/storage/account-index.ts:179)) is true of `0fc35fd4b`, which `35490bb96` then reverted. On `main` the scan already carried on past an unreadable entry. Anyone reading the merged history will look for a bug that never shipped; squashing the pair, or rewriting the comment to state the trade-off rather than the history, removes the archaeology.

2. **`removeWithRetry` exists twice, byte-identical, in the two wipe halves.** The comment explaining the single retry lives on the first copy only, so the second reads as an unexplained duplicate. One module-level helper taking `(remove, what, onFailure)` would carry the explanation once and make the two halves visibly the same policy.

3. **The per-account wipe failure reports the same label for every account.** `onFailure("mnemonic")` in a loop produces N identical `alwaysRecord` Crashlytics entries that say nothing about how many accounts were involved — and the sweep, two files away, deliberately reports `failures/total` exactly once for the same reason. The wipe should report the share too, not the repetition.

4. **The sweep re-probes every indexed account on every boot, forever.** Four or five keychain round-trips per account per launch, with no completion marker and nothing in the code saying when it retires. #1163 is the natural end, but a reader today cannot tell whether the permanence is intended or overlooked; a sentence in the docblock would settle it.

5. **`mnemonicIsMigrated` is the one read on its slot that bypasses the queue.** Everything else on `mnemonic:<id>` goes through `onSlot`; this calls `secureExists` directly ([secureStorage.ts:913](app/utils/storage/secureStorage.ts:913)). Only a count depends on it, so the impact is nil today, but it is an unserialised read on a slot whose serialisation is argued at length elsewhere, and the docblock does not mention the exception.

6. **The docblocks are long enough that they drift.** Several run 15–25 lines above a 5-line method, and the prose restates what the code says rather than what the code cannot — which is how `clearUninstallSurvivingCredentials`'s went stale (Finding 6) without anyone noticing for three commits. The genuinely load-bearing parts are the native-behaviour citations and the rejected alternatives; the narration of the control flow is what ages.

7. **`SWEEP_IDLE_TIMEOUT_MS` lives in `persistent-state/index.tsx`.** That module otherwise knows nothing about mnemonics, and the constant is used once, at the call site of a function imported from `account-index`. It belongs beside the sweep, exported, so the timeout and the thing it bounds are read together.

8. **The idle handle is discarded and the load effect has no cleanup.** `requestIdleCallback(...)` returns a handle that is dropped and `cancelIdleCallback` is never called. Harmless in practice — the sweep is idempotent and `onSlot` serialises the writes — but a fast unmount/remount, or StrictMode's double invocation in dev, schedules two sweeps with nothing expressing that this is known to be safe.

PR #4225, opened 2026-08-29
fix(onboarding): stop the level 2 upgrade screen starting KYC without user input

kngako:
The name/surname step of the level 2 upgrade would skip itself: the form painted, then the screen jumped straight into the KYC webflow without waiting for any input, often ending on an "internal server error" dialog. It looked intermittent because it never happened on a brand new account.

Mechanism
---------
FullOnboardingFlowScreen ran an effect that called startKyc() whenever the onboardingStatus query came back AWAITING_INPUT, bypassing both the name fields and the "Is the spelling of your name correct?" confirmation. It was the only unattended caller of startKyc in the app.

It fired in two distinct ways, and both had to go:

1. After the query resolved. On mount `loading` is true and onboardingStatus is undefined, so the guard is false; the effect fires on the first render *after* the network-only query resolves. By then the name form has already painted -- hence "the screen appears, then leaves on its own". Because the component state is still empty at that point, the mutation went out as { firstName: "", lastName: "" }. Nothing in the Apollo link chain touches variables, so the empty strings reached the server verbatim.

2. On every keystroke. AWAITING_INPUT is not in the terminal-status early return, so the form stayed live underneath. startKyc is a useCallback whose deps include firstName/lastName, and the effect depended on startKyc, so a single character typed into the form fired a second mutation and navigated again -- this time with a partial name such as { firstName: "K", lastName: "" }. A fired-once ref, or dropping startKyc from the dep array, would have left this half of the bug alive.

Why it seemed to work the first time: a fresh account is NOT_STARTED, so the effect no-ops and the form behaves correctly. Completing that first start moves the account to AWAITING_INPUT, and from then on every entry to the screen auto-fires. The error dialog's Ok handler calls goBack(), popping the screen, and all three entry points re-push a fresh instance whose network-only query returns AWAITING_INPUT again -- so the user could never get back to the form by re-entering.

The effect is not a regression from a recent change. git blame puts it in the screen's first commit, 25eecfc18 (Oct 2023), alongside the name inputs and a mutation that already took { firstName, lastName }; the two have contradicted each other from the start. There was one window where it was harmless -- 1744025d5 (#3553) deleted the name fields and made kycFlowStart argument-less, so auto-fire and button-press were the same call -- and b7cd3cd73 (#3572) restored the names 23 days later without revisiting the effect.

Fix
---
Delete the effect. AWAITING_INPUT already falls through both early returns to the name form, so it now behaves like NOT_STARTED: the form waits, and Next is disabled until both fields are filled, which makes an empty-name start structurally impossible rather than merely guarded. This also breaks the re-entry trap, since every entry now reaches the form.

The Next button's disabled check now uses .trim(), so a whitespace-only name can no longer reach the provider.

Tests
-----
The spec had pinned the bug in place: it mocked kycFlowStart with exactly { firstName: "", lastName: "" } and asserted the auto-navigation.

It now mocks the payload the real user path produces and adds a deliberate *trap* mock for the empty-name payload, returning a distinguishable token. The trap is load-bearing. Without it, an unrequested start misses the mock link, gets swallowed by the hook's catch, and never navigates -- so the "does not start on its own" test passes against the unfixed screen. Verified by mutation test: re-inserting the effect turns both new tests red, with token=test-token-web-trap in the assertion diff.

The negative assertion also avoids the `waitFor(() => expect(x).not.toHave BeenCalled(), { timeout: 500 })` idiom used elsewhere in the file, which resolves on its first non-throwing evaluation -- before the query has even resolved -- and so passes against code that definitely navigates. It anchors on the form rendering, then asserts synchronously.

Also in this change:
- assert workflow_run_id in the navigation URL; it has been unasserted since f1e68d95e (#3573) added it for Onfido support.
- move the theme-parameter test to NOT_STARTED. The theme param is orthogonal to the status, and on AWAITING_INPUT that test could not detect the bug.
- correct the stale KycFlowType mock in the useKycFlow spec, which declared { Full, Basic }. Neither member exists in the real enum, so KycFlowType.Card evaluated to undefined on both sides of its assertion and the test proved nothing about `type`.

Deliberately out of scope
-------------------------
- type: KycFlowType.UpgradeLevelTwo. This screen is the app's only kycFlowStart caller that sends no discriminator, and the enum member is used nowhere in the codebase, so the server has some default for the absent case. It is the one available change that could introduce a *new* server-side failure; it should ship alone, once backend confirms the default, so any regression stays attributable.
- An explicit "Continue KYC verification" resume branch. Better product behaviour for a returning user, but it would send {"input":{}} -- a wire shape no path in this app has ever sent -- and a nameless resume gives the affected users no way to correct an applicant that was created with blank names by this very bug.
- useKycFlow's error dialog (inverted title/body, raw err.message leaked into user-facing text, goBack() on Ok) and the retry amplification (kycFlowStart is absent from noRetryOperations and has no idempotency key, so RetryLink replays it up to 5x on a 5xx; there is also no in-flight latch, and the card flow never passes `loading` to its button). All real, all in code shared with the card flow, all tracked separately.

Note that this only removes one of three candidate causes for the reported internal-server-error. Two staging calls settle which it is: kycFlowStart with empty names and no type, and a second kycFlowStart against an account already in AWAITING_INPUT. If it is the latter, the error will still appear -- but only once, and only because the user asked for it.

To reproduce against this fix you need an account already in AWAITING_INPUT; a fresh one is NOT_STARTED and never showed the bug.

---

Here's a video of the behaviour with the fix.

https://github.com/user-attachments/assets/d825e344-20a2-4e41-9e19-8a072d7bea42

### DIFF
--- a/__tests__/hooks/use-kyc-flow.spec.ts
+++ b/__tests__/hooks/use-kyc-flow.spec.ts
@@ -25,8 +25,8 @@ const mockKycFlowStart = jest.fn()
 jest.mock("@app/graphql/generated", () => ({
   useKycFlowStartMutation: () => [mockKycFlowStart],
   KycFlowType: {
-    Full: "FULL",
-    Basic: "BASIC",
+    Card: "CARD",
+    UpgradeLevelTwo: "UPGRADE_LEVEL_TWO",
   },
 }))
 
--- a/__tests__/hooks/use-kyc-flow.spec.ts
+++ b/__tests__/hooks/use-kyc-flow.spec.ts
@@ -119,6 +119,81 @@ describe("useKycFlow", () => {
     })
   })
 
+  it("trims leading and trailing whitespace from firstName and lastName", async () => {
+    mockKycFlowStart.mockResolvedValue({
+      data: {
+        kycFlowStart: {
+          tokenWeb: "t",
+          workflowRunId: "w",
+        },
+      },
+    })
+
+    const { result } = renderHook(() =>
+      useKycFlow({ firstName: "  John ", lastName: "\tDoe\n" }),
+    )
+
+    await act(async () => {
+      await result.current.startKyc()
+    })
+
+    expect(mockKycFlowStart).toHaveBeenCalledWith({
+      variables: {
+        input: { firstName: "John", lastName: "Doe", type: undefined },
+      },
+    })
+  })
+
+  it("preserves internal whitespace in names", async () => {
+    mockKycFlowStart.mockResolvedValue({
+      data: {
+        kycFlowStart: {
+          tokenWeb: "t",
+          workflowRunId: "w",
+        },
+      },
+    })
+
+    const { result } = renderHook(() =>
+      useKycFlow({ firstName: "Mary Jane", lastName: "van der Berg" }),
+    )
+
+    await act(async () => {
+      await result.current.startKyc()
+    })
+
+    expect(mockKycFlowStart).toHaveBeenCalledWith({
+      variables: {
+        input: { firstName: "Mary Jane", lastName: "van der Berg", type: undefined },
+      },
+    })
+  })
+
+  it("sends empty string for whitespace-only names", async () => {
+    mockKycFlowStart.mockResolvedValue({
+      data: {
+        kycFlowStart: {
+          tokenWeb: "t",
+          workflowRunId: "w",
+        },
+      },
+    })
+
+    const { result } = renderHook(() =>
+      useKycFlow({ firstName: "   ", lastName: " " }),
+    )
+
+    await act(async () => {
+      await result.current.startKyc()
+    })
+
+    expect(mockKycFlowStart).toHaveBeenCalledWith({
+      variables: {
+        input: { firstName: "", lastName: "", type: undefined },
+      },
+    })
+  })
+
   it("builds URL with locale and theme mode", async () => {
     mockKycFlowStart.mockResolvedValue({
       data: {
--- a/__tests__/hooks/use-kyc-flow.spec.ts
+++ b/__tests__/hooks/use-kyc-flow.spec.ts
@@ -179,9 +179,7 @@ describe("useKycFlow", () => {
       },
     })
 
-    const { result } = renderHook(() =>
-      useKycFlow({ firstName: "   ", lastName: " " }),
-    )
+    const { result } = renderHook(() => useKycFlow({ firstName: "   ", lastName: " " }))
 
     await act(async () => {
       await result.current.startKyc()
--- a/__tests__/screens/full-onboarding-flow-screen.spec.tsx
+++ b/__tests__/screens/full-onboarding-flow-screen.spec.tsx
@@ -1,7 +1,8 @@
 import React from "react"
+import { Alert } from "react-native"
 import { it } from "@jest/globals"
 import { MockedResponse } from "@apollo/client/testing"
-import { render, waitFor } from "@testing-library/react-native"
+import { act, fireEvent, render, waitFor } from "@testing-library/react-native"
 
 import { FullOnboardingFlowScreen } from "@app/screens/full-onboarding-flow/full-onboarding-flow"
 import {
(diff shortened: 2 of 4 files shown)


grimen (dismissed):
## Findings

**1. The trim guard gates the button but not the submitted payload (blocking).**
The new `!firstName.trim()` check only disables the Next button; the confirmation alert and `useKycFlow({ firstName, lastName })` still use the raw state, so padded names reach the KYC provider verbatim.
*Example:* keyboard autocomplete leaves `firstName = "John "`; Next enables and `kycFlowStart` receives the padded name, which has to match the user's ID document.
*Proposal:* one option is to trim at the call site (`useKycFlow({ firstName: firstName.trim(), lastName: lastName.trim() })`), which also covers the hook's other callers.

**2. Re-entry in AWAITING_INPUT re-submits kycFlowStart with no resume path (blocking).**
A user already in `AWAITING_INPUT` who returns to the screen gets the name form again and issues a second `kycFlowStart`; if the server rejects a duplicate start, the hook's error alert calls `goBack()` and the same dead end repeats on every attempt.
*Example:* the user backgrounds the app mid-KYC, re-enters via Settings > Upgrade, re-types their names, and lands in error alert → back, indefinitely.
*Proposal:* confirming the duplicate-start behavior on staging (the check the PR body defers to) before merge would settle whether the deferral holds; if the server rejects duplicates, a resume branch is needed rather than deferred.

**3. The regression-trap mock is pinned to an exact payload.**
The trap that fails an unrequested `kycFlowStart` only matches `{ firstName: "", lastName: "" }`, so any payload drift — such as the deferred `type: KycFlowType.UpgradeLevelTwo` — silently disarms it and the negative test passes for the wrong reason.
*Example:* the follow-up adds `type` to the input; a reintroduced auto-start then misses both mocks, the rejection is swallowed by the hook's catch, and "does not start on its own" stays green against the regressed screen.
*Proposal:* additionally asserting that the alert spy, `goBack`, and `console.error` were not called would catch an unrequested start regardless of payload shape.

## Test Coverage Gaps

- No test enters whitespace-padded names, so the untrimmed-submission path from finding 1 is uncovered — the specs only type clean `John`/`Doe`.

esaugomez31 (changes requested):
**Blocker**

- `full-onboarding-flow.tsx`: dropping the effect leaves `AWAITING_INPUT` with no handler at all. It's not in the status branch either (there's no `AWAITING_INPUT` entry under `LL.FullOnboarding`), so it now falls through to the empty name form: nothing tells the user a verification is already pending, and Next fires `kycFlowStart` again. That effect has been the resume path since #2699. The real bug is the retrigger, not the resume: `startKyc` has `firstName`/`lastName` in its `useCallback` deps and was in the effect's dep array, so under `AWAITING_INPUT` every keystroke re-fired it. Can we fix the retrigger and keep a resume path? The copy already exists (`kycAwaitingInputTitle` / `kycAwaitingInputDescription`, used by the card banner).

**Non-blockers**

- The confirm alert renders raw state while the hook submits trimmed, so `"John "` shows as `John  Doe` and sends `John Doe`. Better to trim once in the screen, so the alert, the `disabled` guard and the payload read the same value.
- This screen still calls `useKycFlow` without `type`. `KycFlowType.UpgradeLevelTwo` exists in the schema and is used nowhere; the card flow passes `Card`. Pre-existing, but worth confirming with backend whether the discriminator matters for level 2.
- `disabled={!firstName.trim() || !lastName.trim()}`: pull it into a named const.
- `use-kyc-flow.spec.ts`: good catch on the enum values. The old ones (`Full`/`Basic`) aren't in the schema, so `KycFlowType.Card` was `undefined` on both sides and the "passes type" assertion proved nothing. Spreading `jest.requireActual("@app/graphql/generated")` for the enum would stop the next rename doing the same silently.
- No test for the Cancel branch of the alert — that's the other half of "doesn't start without the user".
- `await act(async () => {})` in the guard test: `flushEffects()` in `__tests__/helpers` does the same thing with an explicit `setImmediate`, and it's what the rest of the suite uses.
- The `jest.spyOn(Alert, "alert")` in `submitNames` and in the guard test is never restored, and `beforeEach` only clears, so it leaks into the next describe. Create it in `beforeEach` and pair it with `restoreAllMocks`.
- `confirmButton?.onPress?.()` isn't awaited inside `act`. No warning today, but awaiting it removes the timing dependence.

esaugomez31 (approved):

PR #4226, opened 2026-08-29
test(onboarding): cover the level 2 KYC flow's branches and repair its vacuous tests

kngako:
Follow-up to https://github.com/blinkbitcoin/blink-mobile/pull/4225, which fixed the auto-advance bug but left most of the already existing feature untested and two existing tests asserting nothing at all.

Screen spec: 6 tests -> 19. Hook spec: 9 -> 15.

Repaired
--------
The APPROVED and PROCESSING tests used

  await waitFor(() => expect(mockNavigate).not.toHaveBeenCalled(), { timeout: 500 })

waitFor resolves on its FIRST non-throwing evaluation, which lands before the Apollo query has resolved -- so the assertion ran against a screen that had not rendered the branch yet. Proven vacuous: the same idiom passes against the pre-fix screen, which definitely navigates, and after it returns queryByText("Your onboarding status is: Approved.") is still null. Both are now rows in a table covering all six terminal statuses, each asserting the rendered status line, the ContactSupportButton, and the absence of the name form.

The NOT_STARTED test asserted only that getByTestId("RNE_BUTTON_WRAPPER") was truthy -- a branch discriminator, not a behaviour. Replaced by a table over NOT_STARTED / AWAITING_INPUT / no-status that asserts the form renders with Next disabled.

Added, screen
-------------
- The per-keystroke re-fire regression. This was the half of the original bug with no guard at all: startKyc is re-memoised on every keystroke, so an effect watching it fired a fresh mutation per character. The test types one character at a time and asserts nothing is sent. Against the pre-fix screen this leaks partial names like { firstName: "Joh", lastName: "" } to the provider.
- Next starts disabled, and whitespace-only input does not enable it. The .trim() guard shipped untested in 599b92a6d.
- The confirmation alert's title, its interpolated "John Doe" body, and its button order -- the last because the Cancel and Yes tests address buttons by position.
- Cancel does not start KYC, does not navigate, and leaves the typed names intact.
- A second press while a start is in flight is swallowed (RNE gates onPress on `loading`). Uses delay: Number.POSITIVE_INFINITY so the button stays mid-flight with no wall-clock race.
- The form still renders when the status query errors. The screen destructures only { data, loading } and discards `error`, so this is the realistic way onboardingStatus arrives undefined.
- The spinner branch, asserted synchronously so it also pins that the form does not paint before the status is known.

Added, hook
-----------
- loading returns to false after a REJECTED mutation (the finally block). A stuck spinner is user-visible and nothing covered it.
- The hook does not navigate when the mutation rejects.
- The error alert's Ok handler calls goBack. Previously the alert was asserted but its handler was never invoked.
- An empty workflowRunId omits &workflow_run_id from the URL.
- A resolve with no data still navigates, with token= empty. This asserts current behaviour and is marked as such -- it is a latent bug, not a contract.
- A non-Error rejection produces the alert with an empty message tail.

Two traps this spec now defends against
---------------------------------------
Both were found by mutation-testing the tests themselves, and both make a "did not start" assertion pass against broken code:

1. An unmatched MockedResponse is swallowed. If the screen fires a mutation whose variables match no mock, MockLink errors, useKycFlow's catch absorbs it, and navigate is never called. 599b92a6d added a trap mock for { "", "" }, but that is too narrow for the keystroke test: against the pre-fix screen, typing John/Doe fires 8 unrequested mutations and only the last coincidentally matches the happy-path mock -- the test would have been red by luck. The trap is now a catch-all (variableMatcher: () => true, maxUsageCount: Infinity) listed LAST, since MockLink takes the first match and the John/Doe mock must still win. All 8 now surface as workflow-trap navigations.

2. Anchoring on a render can win the race against a request already in flight. Waiting for the form to paint proves the status query resolved, but not that a mutation the screen already sent has come back. Every negative assertion now runs after settleNetwork(), two macrotask flushes -- one to land MockLink's reply, one to land whatever it triggered.

Verification
------------
Every new and repaired test was mutation-tested against the real sources; each mutant is caught by exactly the test that claims to protect it:

  reinstate the auto-start effect      -> 6 red (both no-start guards included)
  drop .trim() from the disabled guard -> 1 red
  Cancel calls startKyc                -> 1 red
  drop loading={loadingKyc}            -> 1 red
  drop <ContactSupportButton />        -> 6 red (the whole status table)
  remove the loading branch            -> 1 red
  never clear loading in finally       -> 2 red
  error alert Ok no longer goes back   -> 1 red
  always append workflow_run_id        -> 2 red

tsc --noEmit clean; eslint --no-ignore and prettier clean on both files; __tests__/screens + __tests__/hooks green at 351 suites / 4494 tests; no order dependence (--maxWorkers=1 and --randomize across several seeds).

Deliberately not added
----------------------
- A get-kyc-banner-config spec. All nine of its branches are already exercised through the rendered banner title by card-personal-details-screen.spec.tsx. Only the icon and colour slots are unasserted, and that is card-flow surface rather than the level-2 upgrade.
- Requirements copy, alert body wording beyond the interpolated names, and styles -- product copy across 28 locales, with no branching to regress.
- An end-to-end settings -> upgrade -> names -> webview test. Nothing in this repo drives a user across screens through a live navigator; it would be out of idiom as well as out of proportion.
- Anything pinning the /canceled/i branch further. It was written for the native Onfido SDK's cancel error; the try block now wraps only an Apollo mutation and a navigate(), so it is unreachable. It should be deleted, not tested.
- The redundant setLoading(false) inside that branch: proven unkillable. Deleting it leaves all tests green because the finally already runs. No test can distinguish it.


grimen (dismissed):
## Findings

**1. Negative navigation assertions depend on a fixed two-tick flush window (blocking).**
All four "did not navigate" checks rely on `settleNetwork` flushing exactly two macrotasks, so an unrequested KYC start that reaches `navigate` on a later tick would pass every guard unnoticed.
*Example:* A future regression that auto-starts KYC behind a `setTimeout` debounce fires after the two flushes, and `expect(mockNavigate).not.toHaveBeenCalled()` stays green while partial names leak to the KYC provider.
*Proposal:* One option is to make `kycFlowStartTrap.result` a recording function (`result: jest.fn(() => ({ data: … }))`) and assert it was never invoked — the request itself is then caught one tick after it fires, independent of navigation timing.

**2. Comment describes a disabled-guard bypass that doesn't happen.**
The comment says pressing the RNE wrapper's pressable parent bypasses the button's own disabled guard, but the code presses by testID, which lands on the RNE touchable itself and routes through the `!loading && !disabled` guard in `handleOnPress` — nothing is bypassed.
*Example:* A maintainer trusting the comment misreads what the `alertSpy`-not-called assertion exercises, or "fixes" the press target to match the comment.
*Proposal:* Updating the comment to say the press goes through the guard (or dropping it) would keep the documentation truthful.

**3. `console.error` spy restore leaks when an assertion fails.**
The `consoleErrorSpy` setup plus trailing `mockRestore()` pair is repeated in six tests, and any assertion failing before the restore leaves `console.error` silenced for every later test in the file — the `beforeEach` `jest.clearAllMocks()` clears calls but not the `mockImplementation`.
*Example:* One red assertion hides real errors and React `act()` warnings in all subsequent tests, exactly when the suite is already failing.
*Proposal:* An `afterEach(() => consoleErrorSpy.mockRestore())` would be leak-proof and collapse the six duplicated two-line blocks to one.

**4. `pressAlertButton` can silently press nothing (blocking).**
The helper hardcodes `alertSpy.mock.calls[0]` and presses via `button?.onPress?.()`, so a missing button, handler, or a second alert turns the press into a no-op instead of a failure.
*Example:* Dropping Cancel's `onPress` in the screen leaves the cancel test green, because "nothing pressed" and pressing an empty handler are observationally identical.
*Proposal:* One option is to assert `expect(button?.onPress).toBeDefined()` before invoking it (or read `calls.at(-1)` to target the alert currently on screen).

esaugomez31:
@kngako ah, you got there first — I'll just do a quick pass then.

One thing from pulling it down: `full-onboarding-flow-screen.spec.tsx` does not compile on `a6d3ee66`, so the suite fails to run rather than fails a test.

```
TS2451: Cannot redeclare block-scoped variable 'submitNames'.  (206, 304)
TS1005: ',' expected.  (304)
```

The merge left a fragment of the old helpers inside the body of `it("should navigate to WebView with correct params when the user submits their name")` — lines 294-301 are the tail of the pre-merge `renderScreen` (the "form only paints once" comment, its `waitFor`, and `return screen`), which closes that test early and is followed by a second `submitNames` declaration.

Deleting that stray block should be the whole fix; the first `submitNames` at 206 is the one the tests use.

esaugomez31 (approved):
LGTM!

PR #4231, opened 2026-08-31
fix(settings): say "Create Lightning address" on the row with no address yet

designsats:
## Problem

In **Settings**, an account without a Lightning address shows a row titled **"Create address"**. The label doesn't say what kind of address it is, and every other place in the app calls this a **Lightning address**.

## Fix

`SettingsScreen.createAddress` now reads **"Create Lightning address"**.

This follows the review decision on the first version of this PR ("Set receive address"). "Receive address" was a new name that no other screen uses. The row opens `SetAddressModal` ("Choose Lightning Address"), and once the address is set, Account Information labels it "Lightning address". The row now uses that same name.

The key keeps its name because `createAddress` still describes the value. `app/i18n/i18n-types.ts` and `app/i18n/raw-i18n/source/en.json` are regenerated by `yarn update-translations`.

## Translations

All **28 locale files** in `app/i18n/raw-i18n/translations/` are updated in this PR. These are hand-written, not Crowdin output, so reviewers who speak these languages should check them.

Each value combines two strings that locale already has: its earlier "Create address" verb, and the wording it already uses for "Lightning address" in `SetAddressModal` / `AccountInformation`. No new terms were introduced. Every locale keeps the verb, so the row reads as an action and not as a value. `qu` is now `Lightning tinkunata ruray`, built from that file's own `tinkunata`, which also removes the unaccented "direccion".

| loc | value | chars |
|---|---|---:|
| **en** | **Create Lightning address** | 24 |
| `el` | Δημιουργία διεύθυνσης Lightning | 31 |
| `lg` | Kola endagiriro ya Lightning | 28 |
| `de` | Lightning-Adresse erstellen | 27 |
| `fr` | Créer une adresse Lightning | 27 |
| `xh` | Yenza idilesi ye-Lightning | 26 |
| `ca` | Crea una adreça Lightning | 25 |
| `cs` | Vytvořit Lightning adresu | 25 |
| `es` | Crear dirección Lightning | 25 |
| `hu` | Lightning cím létrehozása | 25 |
| `qu` | Lightning tinkunata ruray | 25 |
| `sk` | Vytvoriť Lightning adresu | 25 |
| `it` | Crea indirizzo Lightning | 24 |
| `nl` | Lightning-adres aanmaken | 24 |
| `pt` | Criar endereço Lightning | 24 |
| `ro` | Creează adresă Lightning | 24 |
| `sr` | Направи Lightning адресу | 24 |
| `sw` | Unda anwani ya Lightning | 24 |
| `tr` | Lightning adresi oluştur | 24 |
| `da` | Opret Lightning-adresse | 23 |
| `hr` | Stvori Lightning adresu | 23 |
| `hy` | Ստեղծել Lightning հասցե | 23 |
| `ms` | Cipta alamat Lightning | 22 |
| `th` | สร้างแอดเดรส Lightning | 22 |
| `ar` | إنشاء عنوان Lightning | 21 |
| `id` | Buat alamat Lightning | 21 |
| `vi` | Tạo địa chỉ Lightning | 21 |
| `af` | Skep Lightning-adres | 20 |
| `ja` | Lightningアドレスを作成 | 16 |

**Truncation is not verified on device.** The title is `numberOfLines={1}` with tail ellipsis and scales with the OS font size. Character count is not rendered width, so the longest values (`el` 31, `lg` 28, `de`/`fr` 27) still need a look on a 360dp screen at the largest font scale.

## Acceptance requirements

- [x] Settings row displays "Create Lightning address" for accounts without a Lightning address
- [x] The row uses the same name for the address as `SetAddressModal` and Account Information
- [x] All 28 locale files updated in the same PR, each keeping the verb and the locale's existing "Lightning address" wording
- [x] `yarn update-translations` run; generated files committed
- [x] Copy pinned by a test that reads the real locale files, not the i18n mock
- [ ] Longest locales (`el`, `lg`, `de`, `fr`) checked for truncation at max font scale on a 360dp screen

## Testing specs

- [x] `yarn jest __tests__/i18n` passes (9 suites, 185 tests)
- [x] `yarn jest __tests__/screens/settings-screen` passes (33 suites, 282 tests)
- [x] **Revert check:** with `en/index.ts` set back to the old string, `__tests__/i18n/create-lightning-address-label.spec.ts` fails
- [x] `yarn tsc:check` is clean
- [x] `yarn check:translations` exits 0 with the regenerated files committed
- [x] `npx eslint` on touched specs is clean; `npx prettier --check` on touched `.ts`/`.json` is clean
- [x] Each locale diff is one line; all files re-parse as valid JSON

`__tests__/i18n/create-lightning-address-label.spec.ts` loads the real `en` and `es` locales and checks that every locale file translates the value (it must differ from English) and keeps the word "Lightning". The component spec still tests how the row is wired, with its mock updated to the new copy.

## Screenshots

Closes https://github.com/blinkbitcoin/blink-wip/issues/1206

### DIFF
--- a/__tests__/i18n/create-lightning-address-label.spec.ts
+++ b/__tests__/i18n/create-lightning-address-label.spec.ts
@@ -1,0 +1,55 @@
+import fs from "fs"
+import path from "path"
+
+import { i18nObject } from "@app/i18n/i18n-util"
+import { loadLocale } from "@app/i18n/i18n-util.sync"
+
+const TRANSLATIONS_DIR = path.resolve(
+  __dirname,
+  "..",
+  "..",
+  "app",
+  "i18n",
+  "raw-i18n",
+  "translations",
+)
+
+const ENGLISH = "Create Lightning address"
+
+const locales = fs
+  .readdirSync(TRANSLATIONS_DIR)
+  .filter((file) => file.endsWith(".json"))
+  .map((file) => file.replace(/\.json$/, ""))
+
+const createAddressFor = (locale: string): unknown => {
+  const raw = JSON.parse(
+    fs.readFileSync(path.join(TRANSLATIONS_DIR, `${locale}.json`), "utf8"),
+  )
+  return raw.SettingsScreen?.createAddress
+}
+
+describe("SettingsScreen.createAddress", () => {
+  it("reads 'Create Lightning address' in English", () => {
+    loadLocale("en")
+
+    expect(i18nObject("en").SettingsScreen.createAddress()).toBe(ENGLISH)
+  })
+
+  it("reads 'Crear dirección Lightning' in Spanish", () => {
+    loadLocale("es")
+
+    expect(i18nObject("es").SettingsScreen.createAddress()).toBe(
+      "Crear dirección Lightning",
+    )
+  })
+
+  locales.forEach((locale) => {
+    it(`is translated and names Lightning in ${locale}`, () => {
+      const value = createAddressFor(locale)
+
+      expect(typeof value).toBe("string")
+      expect(value).not.toBe(ENGLISH)
+      expect(value).toContain("Lightning")
+    })
+  })
+})
--- a/__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx
+++ b/__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx
@@ -73,7 +73,7 @@ jest.mock("@app/i18n/i18n-react", () => ({
   useI18nContext: () => ({
     LL: {
       SettingsScreen: {
-        createAddress: () => "Create address",
+        createAddress: () => "Set receive address",
         addressDisabled: () => "(disabled)",
       },
       GaloyAddressScreen: { copiedLightningAddressToClipboard: () => "Copied" },
@@ -105,7 +105,7 @@ describe("AccountLNAddress (self-custodial)", () => {
 
     render(<AccountLNAddress />)
 
-    expect(lastRowProps().title).toBe("Create address")
+    expect(lastRowProps().title).toBe("Set receive address")
     expect(lastRowProps().rightIcon).toBeUndefined()
     expect(mockScModal.mock.calls.at(-1)?.[0]?.isVisible).toBe(false)
 
@@ -147,7 +147,7 @@ describe("AccountLNAddress (self-custodial)", () => {
 
     render(<AccountLNAddress />)
 
-    expect(lastRowProps().title).toBe("Create address")
+    expect(lastRowProps().title).toBe("Set receive address")
   })
 
   /** Registering an address is the very thing Incognito withholds: publishing one would
@@ -186,7 +186,7 @@ describe("AccountLNAddress (self-custodial)", () => {
 
     render(<AccountLNAddress />)
 
-    expect(lastRowProps().title).toBe("Create address")
+    expect(lastRowProps().title).toBe("Set receive address")
     expect(mockBackupRequiredModal.mock.calls.at(-1)?.[0]?.isVisible).toBe(false)
 
     act(() => (lastRowProps().action as () => void)())
@@ -269,7 +269,7 @@ describe("AccountLNAddress (custodial)", () => {
 
     render(<AccountLNAddress />)
 
-    expect(lastRowProps().title).toBe("Create address")
+    expect(lastRowProps().title).toBe("Set receive address")
     expect(lastRowProps().rightIcon).toBeUndefined()
     expect(mockCustodialModal.mock.calls.at(-1)?.[0]?.isVisible).toBe(false)
 
--- a/__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx
+++ b/__tests__/screens/settings-screen/settings/account-ln-address.spec.tsx
@@ -73,7 +73,7 @@ jest.mock("@app/i18n/i18n-react", () => ({
   useI18nContext: () => ({
     LL: {
       SettingsScreen: {
-        createAddress: () => "Set receive address",
+        createAddress: () => "Create Lightning address",
         addressDisabled: () => "(disabled)",
       },
       GaloyAddressScreen: { copiedLightningAddressToClipboard: () => "Copied" },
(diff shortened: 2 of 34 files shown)


kngako (changes requested):
2 blocking findings above.

kngako (approved):
No blocking findings are found. I'll leave it to @designsats to decide if he addresses any of the test coverage issues in this PR or in a future/stacked PR.

--- 

**Overall assessment:** The copy change is complete and internally consistent: one key, one call site, generated files in sync, all 28 locales moved together with wording that matches what each locale already calls the feature in Account Information, and a guard that makes the English string un-revertable by accident. **No finding is blocking.** Both findings are about the shape of what surrounds the change — the guard encodes a capitalisation the locale files do not uniformly follow, and the first commit is a complete, contradicted draft that the second commit overwrites hunk for hunk.

---

## Findings

### Finding 1 — The locale sweep is case-sensitive on "Lightning" while the locale files are not [nitpick non-blocking]

Fifteen of the 28 files already spell the feature with a lowercase "lightning" in `ReceiveScreen.lightningAddress` (`ca` "Adreça lightning", `es` "Dirección lightning", `fr` "Adresse lightning", `pt`, `it`, `ro`, `th`, `vi`, …). A Transifex translator who harmonises the new row title with that neighbouring string — "Crea una adreça lightning" — produces a correct, fully translated value that `expect(value).toContain("Lightning")` rejects (reproduced: `af` set to `"Skep lightning-adres"` fails the sweep). The sync PR goes red over a capital letter, and whoever reviews it has to either revert the translator or edit the test, which is the opposite of what a copy guard is for.

**Proposal:** Match case-insensitively — `expect(value).toMatch(/lightning/i)` — which still fails every regression the guard exists for (the English leak, the revert to `"Skep adres"`, and `""` all lack the word entirely, so all three still fail). Leave the `en` and `es` exact pins as they are; exact pinning is right for the two locales the repo's convention pins. Note the same guard also freezes the Latin-script spelling for `ja`, `th`, `sr`, `ar`, `el`, `hy`; every one of those files uses Latin "Lightning" today, so that is consistent, but if a locale ever moves to a transliteration (`ライトニング`) this sweep is the one test that will need a per-locale exception.

**Proposal:** Squash `1915f6d14` into `474c7af1a` (`git rebase -i 4ccf56872`, mark the second as `squash`), keeping commit 2's message since it is the one that describes the shipped copy and the per-locale method. While editing the message, fix the platform name: commit 1 says the locales were translated "rather than left to Crowdin", but the repo syncs through Transifex (`transifex.yml`, `tx_translations_*` PR branches). Commit 3 can stay as a separate `test(i18n)` commit — it reads well as a follow-on and its message is accurate. If Finding 1's regex change is made, put it in commit 3 so the guard lands in its final shape.

---

## Test Coverage Gaps

1. **The one spec that renders this row through the real `LL` asserts on a string that has not existed since July.** `__tests__/screens/settings-screen/settings-screen.spec.tsx:587` ("hides phone ln address when phone is missing") does not mock `@app/i18n/i18n-react` and checks `expect(screen.queryByText("Set your lightning address")).toBeNull()`. That copy left the codebase in `88effd4ee` (2026-07-22) when the key became `createAddress: "Create address"`; the assertion has passed vacuously through that rename and now through this one. Replace the literal with `LL.SettingsScreen.createAddress()` (or the new English string) and pair it with a positive anchor — the `username@domain` row is present — so the negative means something. Pre-existing, but this branch touched every other reference to this copy and is the natural moment to fix it.
2. **The sweep still accepts two degenerate values.** A value that keeps the brand word but loses the verb (`"Lightning"`) and an English leak padded with whitespace (`"Create Lightning address "`) both pass. Whitespace slop is real in these files — `tr.json`'s `SetAddressModal.setLightningAddress` carries a leading space today. `value.trim()` before the `not.toBe` and a `trim().length > "Lightning".length` check close both.
3. **Only `en` and `es` are asserted through the real loader.** The 26 other locales are checked on raw JSON; nothing exercises `loadLocale(<locale>)` + the `lodash.merge` over `en` for them. `id` and `xh` have no loader directory at all and never reach users (pre-existing, noted in the previous review), so for those two the sweep is asserting on files the app cannot load.
4. **The stated design intent is not asserted.** Commit 2's reason for the wording is that the row, `SetAddressModal.mainTitle` and `AccountInformation.lightningAddressLabel` should agree per locale; the sweep only checks the proxy "contains Lightning". A generic containment check fails on inflected locales (`cs` "adresa"/"adresu", `qu` "chaskiyninkuna"/"tinkunata"), but a spot-check for the three or four locales where the forms coincide (`en`, `de`, `hu`, `af`) would pin the intent rather than the proxy.
5. **`skip-queries-when-unauthed.spec.tsx`'s stub was updated but nothing reads it.** Harmless here — the stub only needs to exist — but it is the same hand-rolled-`LL` drift the previous review's Finding 1 described: value changes in these stubs are invisible to every test and every type check.

## Code Smells

1. **Fifth copy of the `TRANSLATIONS_DIR` boilerplate.** `brand-name`, `limit-copy`, `locale-parity`, `map-distance-label` and now `create-lightning-address-label` each declare the same nine-line `path.resolve(__dirname, "..", "..", "app", "i18n", "raw-i18n", "translations")` and the same `readdirSync().filter(".json")` listing — and the new file is the only one that does not `.sort()` it, so its test order follows `readdir`. A `__tests__/helpers/locales.ts` exporting `TRANSLATIONS_DIR`, a sorted `localeFiles`, and `readLocale(name)` would collapse all five to one import.
2. **Commit-message drift, all in commit 1.** "Crowdin" (the repo uses Transifex), "Every translation stays within six characters of the English" (true of the discarded strings, not the shipped ones — `el` is +7), and "matching how the feature is named elsewhere in the app" (commit 2 says it was not). All of it folds away with the squash in Finding 2.
3. **`SUBTITLE_SHORTER_LENGTH` now does nothing on the row that computes it.** `account-ln-address.tsx:84` passes `subtitleShorter={displayedTitle.length > 22}`, and this branch is what pushed the English empty-state title over that threshold (24 chars) — but `LightningAddressRow` never passes a `subtitle`, and `SettingsRow` only reads `subtitleShorter` to pick the subtitle's font. The computation has been dead since `d199097fe` introduced it; dropping the prop from this row (separate one-line PR) stops the next reader wondering what changed.
4. **Commit 2's message slightly overstates the source of the wording.** It says each value follows "SetAddressModal and AccountInformation"; in `sw` and `qu` those two disagree and the branch picked AccountInformation's word in one and SetAddressModal's in the other. The values are fine — naming AccountInformation as the tie-breaker in the squashed message would make the claim exact.

PR #4232, opened 2026-09-01
fix(i18n): call the cloud backup encryption secret a passphrase

designsats:
Closes blinkbitcoin/blink-wip#1214

## Problem

The cloud backup flow called its encryption secret a "password". The checkbox read **"Encrypt with additional password"**, the fields it revealed were **"Password"** / **"Confirm password"**, and the restore side asked for **"Enter your encryption password"**.

Two things are wrong with that. "Encrypt with…" describes the mechanism rather than the action the user takes, and it was the longest label on the screen. And "password" implies an account credential, when this secret is a BIP-39-style extra phrase layered on top of the backup phrase — "passphrase" is how it is described elsewhere in bitcoin wallets.

## Fix

Copy only, no behaviour change. The checkbox becomes **"Add additional passphrase"**, which states the action, and every other occurrence of "password" in this flow becomes "passphrase", on both the backup and restore side.

### `BackupScreen.CloudBackup`

| Key | Before | After |
| --- | --- | --- |
| `encryptCheckbox` | Encrypt with additional password | Add additional passphrase |
| `password` | Password | Passphrase |
| `confirmPassword` | Confirm password | Confirm passphrase |
| `confirmPasswordPlaceholder` | Enter password again | Enter passphrase again |
| `importantMessage` | Blink does not store this password. If you lose it, {bold} to recover your wallet. | Blink does not store this passphrase. If you lose it, {bold} to recover your wallet. |
| `passwordMismatch` | Passwords do not match | Passphrases do not match |

### `RestoreScreen`

| Key | Before | After |
| --- | --- | --- |
| `enterPassword` | Enter your encryption password | Enter your encryption passphrase |
| `wrongPassword` | Incorrect password. Please try again. | Incorrect passphrase. Please try again. |

`passwordPlaceholder` ("(12 characters minimum)") and `passwordTooShort` ("Minimum 12 characters") never named the secret, so they are unchanged.

## Notes

**The key names are deliberately unchanged — the diff touches string values only.** Renaming `password` → `passphrase` would ripple into `cloud-backup-screen.tsx`, `cloud-restore-screen.tsx`, `hooks/`, `cloud-backup-validation.ts` and the regenerated types for no user-visible gain, and would bury a copy change in a mechanical rename.

No screen file needed to change: every label already reads from these keys. The restore screen reuses `BackupScreen.CloudBackup.password()` for its own field label, so it picks the rename up from the backup-side key.

`BackupScreen.BackupMethod.passwordManager*` is untouched. It refers to a real password manager (iOS Keychain / Android autofill) offered as a backup destination, not to this secret, and is the only "password" left in the flow.

Generated files (`raw-i18n/source/en.json`, `i18n-types.ts`) were regenerated with `yarn update-translations`, not hand-edited. These keys exist only in `en`, so this is not a locale-parity change and no other locale file is modified.

No migration needed — no stored data, format or validation rule changes, and an existing encrypted backup still restores with the same secret.

### On the test

Because the keys kept their old names, the existing screen specs assert labels *through* `LL.BackupScreen.CloudBackup.password()` — they would still pass if the copy regressed. The new spec asserts the values instead, following the `recovery-phrase-label` precedent. Reverting `app/i18n/en/index.ts` fails its six tests covering the renamed strings; the two guarding unchanged copy pass either way by design. It also guards the flow as a whole: no user-facing value under `BackupScreen.CloudBackup` or `RestoreScreen` may contain "password" again.

## Acceptance requirements

- [x] Every string in the two tables matches its "After" value exactly.
- [x] The word "password" no longer appears anywhere on the cloud backup screen (either step) or in the cloud restore flow.
- [x] The `passwordManager*` strings on the backup-method screen are unchanged.
- [x] `grep -n "assword" app/i18n/en/index.ts` returns only the `passwordManager*` lines plus the untouched `passwordPlaceholder` / `passwordTooShort` key names.
- [x] Key names are unchanged; the `app/i18n/en/index.ts` diff touches string values only.
- [x] `raw-i18n/source/en.json` and `i18n-types.ts` regenerated via `yarn update-translations`.
- [x] No other locale files are modified.
- [x] Toggling the checkbox still shows/hides the fields; the 12-character minimum, the match validation and the decrypt-on-restore path are unchanged.
- [x] `yarn check:translations`, `yarn tsc:check` and `yarn eslint:check` pass.
- [x] The checkbox label, both field labels and the restore title fit without truncation on a small device (e.g. Pixel 4a / iPhone SE).

## Testing specs

1. Fresh self-custodial wallet → Back up your wallet → Google Drive (Android) or Apple iCloud (iOS).
2. **Step 1:** the checkbox above the fields reads "Add additional passphrase".
3. Tap it. **Step 2:** the fields read "Passphrase" and "Confirm passphrase", the second placeholder reads "Enter passphrase again", and the "Important" banner reads "Blink does not store this passphrase…".
4. Tap the checkbox again: the fields and banner collapse.
5. With it checked, enter fewer than 12 characters → "Minimum 12 characters". Enter two different values → "Passphrases do not match".
6. Complete an encrypted backup and confirm the upload still succeeds ("Backup saved to {provider}").
7. Reinstall / restore from that backup: the screen reads "Enter your encryption passphrase" over a field labelled "Passphrase".
8. Enter the wrong value → "Incorrect passphrase. Please try again." Enter the right one → the wallet restores.
9. Back on the backup-method screen, the "Password manager" option and its messages still say "password manager".
10. Repeat steps 2–3 and 7 on a small-screen device and confirm no label is truncated.

Verified locally: `yarn tsc:check`, `yarn eslint:check` and `yarn check:translations` pass, and `yarn jest __tests__/i18n __tests__/screens/self-custodial/onboarding` is green (42 suites, 664 tests). Built and installed on an Android emulator.

## Screenshots

| Before | After |
| --- | --- |
| Image | Image |
| "Encrypt with additional password" | "Add additional passphrase" |


kngako (changes requested):
3 blocking findings above.

PR #4234, opened 2026-09-01
feat(card): sign the investment agreement on an embedded DocuSign envelope

esaugomez31:
Stacked on #3707. Review that one first.

The Term Sheet's "Proceed to sign" button now leads to a real signing step instead of jumping straight to the transfer screen. The agreement is signed on a DocuSign envelope embedded in the app: the signer lands on the documents themselves, with every figure already written on them and locked.

## How it works

The signing UI comes from `@blinkbitcoin/esign-react-native`. We import the `/webform` entry, which is Apollo-free, so the app stays on Apollo 3.

The library is driven headless, through `useESignature` and a WebView of our own, rather than through its `ESignature` component. That component opens on a "Sign Document" screen with its own button, and the signer has already answered that question on the Term Sheet, so the step would ask it twice. The session starts as the screen paints and the document is what the signer sees.

`SignInvestScreen` owns the routing around it. Signing replaces the screen with the transfer step, since an agreement cannot be unsigned and a back swipe should not land on a finished session. Cancelling or declining goes back to the Term Sheet, and a declined session is not reopened on the way out. A failure stays put and is logged, so the retry the screen draws stays reachable; an expired session is retried like any other, since the hosted form mints a fresh envelope on every start. The one failure worded by the app is a signer the host has not named: it says, translated, to restart the app, since the host's fields are read at launch.

## Why the envelope is minted rather than opened

The first version embedded a published Web Form and wrote the investment's figures into its address. That works, but the signer can edit them before signing, and DocuSign offers no way to lock a field filled that way. Made read only, the field cannot be populated from a URL at all; hidden by a rule, the value never reaches the signed document. A Web Form also always walks the signer through its question pages, even when every answer is already filled in.

An envelope created from the templates does not have either problem: each value travels with its own lock, and the signer opens the agreement directly. So the app asks the e-sign service for an envelope (`POST /envelope/instance`) and opens the signing URL it answers with.

What goes on the document is decided in `investment-agreement.ts`. The seven figures come from the amount the investor chose and the price the app has just read, one bitcoin priced in whole cents so the rate keeps the cents the feed gave: units, price per unit, valuation, total, BTC/USD rate, the bitcoin it settles to and the moment the rate held. The rest, the signer's legal name and email, country of residence and the membership fields, comes from the `cardInvestmentAgreementPrefill` remote-config key, a dictionary from template label to value. Every value is sent locked; the figures are written last so a remote-config entry can never override them.

The satoshis the document names are kept from the mint and handed to the transfer step, so the payment is billed at the rate the agreement states rather than a later conversion.

Verified against DocuSign's demo account: the three documents arrive in one envelope, in order, with every value filled and not editable, and the signer only signs. One condition worth recording: every locked field has to be of type **Text**. A read only Number or Date makes DocuSign refuse the submission.

## Where the service runs

`GaloyInstance` gains `esignMintUrl`, the origin of the e-sign service for that environment. Staging points at the deployed service; Main and Local are empty until one exists there. With it unset a debug build falls back to a service on the developer's own machine, so the flow can still be exercised. A release build does not fall back, since that address is the developer's device.

The request carries the app's session token as the bearer. The staging deployment takes the bearer as the caller's id without verifying it; putting verification in front of the service is a deployment decision, not the app's.

## What the signer sees while the page loads

The signing page is DocuSign's own app. It lands as an empty shell, spins on an indicator of its own for a few seconds after the WebView reports the page loaded, then draws its interface. Left alone that is two spinners in a row: this step's while the envelope is minted, then the page's. So the step keeps its own spinner over the WebView until the page has drawn, which a small injected script reports the moment the page has visible text; it polls every 250 ms and gives up after 20 s, and the step uncovers the page at 20 s regardless, so nothing can stay hidden. The WebView's own loading indicator is switched off, since it ends seconds before the page has drawn, and the page is kept out of the accessibility tree while covered. Measured against the page as it is today: the shell and the redirects it arrives through carry no text until the interface is drawn.

A cold open waits for the price feed before minting, and that wait now ends: after 15 s without a price the step shows the lost-connection state with a retry, instead of a spinner with no end and no button. The session still starts on its own the moment the price arrives, whichever state is showing.

## The location prompt

The signing page asks the browser for the signer's location, which Android's WebView turns into a system permission prompt on every visit. The location is DocuSign's optional audit extra, not something the signature needs, so the WebView declines it before it reaches the signer. iOS has no such switch; there the request stops only once the account's signing resource file sets `DocuSign_DisableLocationAwareness`, which needs resource-file branding enabled on the production account.

## Also in here

`esign-form-url.ts`, the `cardInvestmentEsignFormUrl` remote-config key and the allowed-origin constant are gone: the mint replaces all three, and the session is not told an origin either, since a WebView message carries none and the library cannot check it on this platform.

The Term Sheet's lines are in sentence case as designed, in every language. The investment options are named as what they are, and the balance checks use the shared unit helpers.

The e-sign packages move to 0.3.0, for the Apollo-free `/webform` entry and the headless hook.

`react-native-webview` goes from 13.16 to 16.0. The library needs 14 or newer, but 14 and 15 ship a broken type declaration (`WebView<P = undefined>` collapses `WebViewProps & undefined` to `never`, so TypeScript rejects every prop). 16 has it fixed. The only breaking change that touches us is the Android API 24 floor, and we are on 26.

The `@blinkbitcoin` scope now resolves from GitHub Packages. The token is deliberately not in the repo's `.npmrc`: yarn 1 aborts every install, not just that scope, when it finds an unset `${ENV_VAR}` there. Put it in your own `~/.npmrc` instead. CI gets it through `setup-node`.

## Testing

Device tested on Android against the staging service: the step opens the envelope with no gate in front of it and one spinner from the tap to the document, no location prompt, the figures and the host fields arrive filled and locked, the three documents are signed in order, and the flow lands on the transfer step with the satoshis the document names.

The specs cover what the app owns: that the session starts once and only once, what the mint is asked to write and to lock, that a missing price or an unnamed signer is reported rather than minted around, that the document opened is the one the service answered with, what the signer is shown in each state, that a decline does not open another session, that the page stays covered until it reports or the timeout passes and is covered again for a retried session, that the page's report never reaches the library, that the location request is declined, where every outcome navigates, and that a release build offers no local fallback.

## Known limitations

The copy for a failure comes from the library and is English only. The title and the button are the app's.

The remote-config fields are global: the same signer name and email reach every user of the app at once. That matches how the round is run for now, one personal invitation at a time with the values updated before each one is sent, and it is not a substitute for per-user data.


kngako (changes requested):
Review of the 42 latest commits on feat--esign-screen-docusign — 9 findings (5 blocking), 9 test-coverage gaps, 9 code smells.

---

## Findings

### Finding 1 — The mint route ships in no released e-sign server **[BLOCKING]**

`esign-mint.ts:22` posts to `/envelope/instance`, but `package.json` pins `@blinkbitcoin/esign-react-native@0.3.0`, and at tag `v0.3.0` the matching server package (`packages/esign-server/src/express.ts`) serves only `/webform/instance` and `/webhook/esign` — `/envelope/instance` appears nowhere in that tag. The route exists only on `main` (`packages/esign-service`, `packages/esign-node`), and even there `mint.ts` selects it only when the deployment sets `ESIGN_MINT_MODE=envelope`; unset, blank or whitespace defaults to `webform`. Commit `58ec6c341` points staging at `https://esign.staging.blinkbtc.com` with no recorded check of either the service revision or that variable.

Concretely: a tester on staging taps "Proceed to sign", the POST 404s, `failureOf` files it as `ENVELOPE_CREATION_FAILED` with message `HTTP 404`, and `getErrorMessage` discards that message for "Unable to create signing session. Please try again." The retry button re-404s forever, and neither the screen nor `logError`'s context names the route or the status — the log carries only the code.

**Proposal:** Pin the client to a release that actually ships `/envelope/instance`, or vendor the service revision and record it, and confirm the staging deployment's `ESIGN_MINT_MODE` before this merges. Keep the path in one constant asserted in a test against the library's own exported `ENVELOPE_INSTANCE_PATH`, so a client bump that renames or drops the route fails the build rather than the signer. Separately, give 404 and 405 their own non-retryable copy — a route that does not exist is not something a signer can cure by tapping again — and put `response.status` into the `logError` context beside the code.

### Finding 2 — The signer is billed a figure the service is free to replace **[BLOCKING]**

`investment-agreement.ts:106` computes `settlementSats` from the app's own terms, and `transfer-invest-screen.tsx:57` bills exactly that. But the envelope mint on `main` runs the caller's request through a terms hook whose contract is explicit — `handlers.ts:456`: "The host's chance to decide who signs and what the envelope locks, from its own data … Client values are input, never trusted for the signer or for a locked value." That is the same hook commit `951785abe` now depends on to name the signer, so this branch newly requires the exact mechanism that may rewrite the figures. The mint reply is `{ url, envelopeId }` only, so the app has nothing to reconcile against.

Concretely: the service's terms hook re-reads its own price feed a second after the app read the app's, and locks `settlement_amount_btc` at 0.24980000. The document the investor signs says 0.24980000 BTC; the invoice the transfer step writes is for 25,000,000 sats. Nothing in the app notices, and the discrepancy only surfaces in reconciliation.

**Proposal:** Have the mint answer with the prefill it actually minted (or at minimum the settlement figure), derive `settlementSats` from that reply rather than from the local computation, and fail the step when the two disagree by more than a satoshi. Until the service returns it, the alternative is to make the app the sole authority and record "no terms hook on this deployment" as a deployment constraint with a test or a runbook entry — but then Finding 1's signer resolution has to come from somewhere else. A third option, if the e-sign GraphQL envelope is reachable, is to read the envelope's locked values back before writing the invoice.

### Finding 3 — A timed-out mint keeps running and can overwrite the settlement figure **[BLOCKING]**

`withTimeout` in `esign-core` races the mint against a 30s timer but never aborts the underlying `fetch` — the abandoned promise runs to completion. `sign-invest-screen.tsx:128` stores the figure in a ref written inside `createInstance`, so a late mint still writes it, and `goToTransfer` reads whatever the ref holds at completion time. The library hands `onComplete` an `ESignatureResult` carrying the `envelopeId` of the session that actually completed; the app ignores the argument entirely.

Concretely: the first mint stalls on a bad connection; at 30s the library's watchdog fails the session; the signer taps "Try again"; the second mint returns at 35s, writes its figure and opens the document; the abandoned first fetch lands at 40s and overwrites the ref with the older price's figure. The signer signs envelope B and is invoiced envelope A's satoshis — the one outcome the whole commit series exists to prevent.

**Proposal:** Store `{ envelopeId, settlementSats }` rather than a bare number, and in `goToTransfer` read `result.envelopeId` from the `onComplete` argument, passing the figure only when the ids match and failing the step when they do not. A generation counter checked inside `createInstance` before the write is an equally small fix if the ids ever come back undefined. Either way, add a test that a late mint resolving after a second one does not change what the transfer step is handed.

### Finding 4 — Any page in the WebView can complete the signing **[BLOCKING]**

`sign-invest-screen.tsx:182` deliberately passes no `allowedOrigin`, on the stated grounds that "a WebView message carries no origin, so the library cannot check it on this platform." Half of that is right — `useESignature`'s message handler never reads the origin, and `esign-core/signing/origin.ts` says outright it is "Web only in practice" — but the premise is wrong: `WebViewMessageEvent.nativeEvent.url` is part of `WebViewNativeEvent` in react-native-webview 16 (`lib/WebViewTypes.d.ts:48`). The app's own `handleWebViewMessage` already intercepts every message before the library sees it, and it checks nothing. There is also no `onShouldStartLoadWithRequest`, so in-place navigation away from the DocuSign instance is unrestricted.

Concretely: the agreement's own text contains a hyperlink, the signer taps it, the WebView navigates in place, and that page posts `{"event":"signing_complete"}`. `interpretBridgeEvent` maps it to a completion, the machine fires `onComplete`, and the app replaces to the transfer step and asks the backend for an invoice against an agreement nobody signed.

**Proposal:** In `handleWebViewMessage`, drop any message whose `event.nativeEvent.url` does not share the mint origin — the same service serves the return bridge, so one comparison covers it — before delegating to `webViewProps.onMessage`. Add an `onShouldStartLoadWithRequest` that keeps off-origin navigation out of this WebView and opens it externally instead. Correct the comment so the next reader does not conclude the check is impossible. Longer term the transfer step should confirm the envelope's status server-side (the service has `/webhook/esign` and an envelope store) rather than take a page's word before asking for money.

### Finding 5 — iOS pods were never updated for the two new native modules **[BLOCKING]**

`ios/Podfile.lock:1896` still pins `react-native-webview (13.16.0)` and contains no `react-native-netinfo` pod at all; no file under `ios/` or `android/` changed anywhere on the branch, not just in these 42 commits. `update_pods.yml` would regenerate it, but that workflow triggers only on `dependabot/npm_and_yarn/**` branches, so nothing on this branch will.

Concretely: a developer or CI job that builds iOS from this branch without remembering to re-run `pod install` links webview 13's native code behind webview 16's JavaScript and has no NetInfo module whatsoever — and `useESignature`'s very first act is `checkConnectivity()` → `NetInfo.fetch()`, so the signing step rejects before it ever mints. The new global Jest stub for netinfo means no test can catch this.

**Proposal:** Run `pod install` and commit the `Podfile.lock` diff with the branch. While there, check webview 16's iOS deployment target against the Podfile and whether the signing page needs any new `Info.plist` usage keys. Best of all, add a CI step (or widen `update_pods.yml`'s trigger) that fails when `Podfile.lock` disagrees with the native dependencies in `yarn.lock` — a lockfile this easy to forget will be forgotten again.

### Finding 6 — Nothing pins the agreement's field labels to the template's tab types [NON-BLOCKING]

`AGREEMENT_LABELS` names seven template fields, and `esign-node/providers/docusign/types.ts:48` warns that every key must name a **Text** tab on the signer's role, because "DocuSign matches a template-role value by tab type + label, so a label naming a Number, Date or List tab is ignored with a 200 and an empty, editable field." The app treats a 200 with a url as success and has no way to learn otherwise.

Concretely: whoever maintains the template types `rate_timestamp` as a Date tab and the money fields as Number tabs, because that is what they hold. DocuSign answers 200, the app opens the form, and the investor signs an agreement whose rate, rate timestamp and settlement amount are blank and editable — then gets invoiced the figure the app computed for the fields that never landed.

**Proposal:** Record the required tab type per label next to `AGREEMENT_LABELS` and pin the template id alongside it, so the contract lives in the repo rather than in a DocuSign console. If Finding 2's proposal lands — the mint answering with the prefill it minted — a missing tab becomes detectable from the reply, which is the only way the app can ever catch this. Until then a manual pre-release check of the rendered document against the seven labels is the honest mitigation, and worth naming in the PR.

### Finding 7 — The signing step's error copy is English in all 28 locales [NON-BLOCKING]

Commits `569a3acb2` and `ea09a2534` removed this step's i18n keys, and `sign-invest-screen.tsx:374`/`387` now render `getErrorMessage` from the library — a hard-coded English `switch` in `esign-core/signing/messages.ts`, with no locale parameter and no override hook on the headless path. Only the title (`LL.common.error()`) and the button (`LL.common.tryAgain()`) remain localised.

Concretely: a Spanish-speaking investor loses connection at the signing step and reads "Error", then "Connection lost. Please check your network and try again.", then "Intentar de nuevo" — two localised strings bracketing an English one, on the screen where they commit to an investment.

**Proposal:** Add a small `Record<string, () => LocalizedString>` mapping the library's codes (`ENVELOPE_CREATION_FAILED`, `NETWORK_ERROR`, `UNAUTHORIZED`, `SESSION_EXPIRED`, `PROVIDER_UNAVAILABLE`, plus a default) to the app's own keys, keeping the server-supplied reason only for `VALIDATION_ERROR` the way the library does. Add those keys to all 28 locale files, editing the lines rather than reserialising the JSON, since the indent width is not uniform across them. A guard test asserting no locale falls back to the English string would keep the next code the library adds from slipping through silently.

### Finding 8 — Production dead-ends on a retryable error, and the endpoint needs a release to change [NON-BLOCKING]

`GALOY_INSTANCES` gives `Main` an `esignMintUrl` of `""`, so `resolveMintOrigin` answers `""` in a release build and `mintSigningInstance` throws `signingFailure("no e-sign service is configured for this environment")` — which the library words as the ordinary retryable "Unable to create signing session. Please try again." The value is compiled into the binary; `d332a852e`'s message advertises remote config, but that change was reverted inside this range and `feature-flags-context.tsx` is untouched at the head.

Concretely: someone opens `blink://card/onboarding/investment` on a production build (the flow is deep-link and developer-shortcut only, which limits the blast radius), walks through to the signing step, and is handed a retry button that can never succeed. And when the service does go live on production, turning it on — or rolling it back after an incident — needs an App Store release.

**Proposal:** Gate the step on a configured origin and say "not available yet" rather than offering a retry, or gate the flow's deep link itself so the dead end is unreachable. Move `esignMintUrl` to the remote config the repo already uses for the sibling `cardESignConsentUrl`, keeping the per-instance value as the default, so the endpoint can be switched or withdrawn without shipping a build.

### Finding 9 — `e2e.yml` installs without the GitHub Packages token [NON-BLOCKING]

`check-code.yml`, `test.yml`, `audit.yml` and `update_pods.yml` all gained the `registry-url` / `scope` / `NODE_AUTH_TOKEN` trio and the `packages: read` permission. `e2e.yml:35` and `:169` still run a bare `nix develop -c yarn install`, and that workflow triggers on every `pull_request` to `main`.

Concretely: the lockfile hash changed, so the workflow's `node_modules` cache misses; yarn then resolves `@blinkbitcoin/esign-*` from `npm.pkg.github.com` with no credentials and fails the install step, so the E2E job goes red on this PR and every PR after it.

**Proposal:** Write the runner-local npmrc in that job the way `update_pods.yml` does (`echo "//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}" >> ~/.npmrc`) and add `packages: read` to its permissions. If E2E is on its way out — there is a `ci/remove-e2e-tests` branch — delete the workflow instead, rather than landing a required check that cannot pass.

---

## Test Coverage Gaps

Two of these were confirmed by mutation, not inferred: the mutant was applied, the suite still passed, and the source was restored.

1. **The double-start guard is unpinned — mutation-verified.** Removing `hasStartedFromIdle.current` from the start effect's condition (`sign-invest-screen.tsx:271`) leaves all 39 tests in `sign-invest-screen.spec.tsx` passing. The guard exists precisely because "`sign` is rebuilt whenever the source is, and starting twice would open a second session on top of the first" — i.e. it protects against a double *mint*, the most expensive mistake on this screen. The neighbouring `isLeaving` guard, by contrast, is caught (removing it fails 1 test), as is every other mutation I tried on this file: the price-read ref (fails 1), `accessibilityElementsHidden` (2), `importantForAccessibility` (1), the carried `settlementSats` (1), and the give-up timer (2). A test that rebuilds the source while the status stays `idle` and asserts `sign` was called once would close it.
2. **`Math.round` on the settlement sats is unpinned — mutation-verified.** Dropping it from `resolveSettlementSats` leaves all 50 tests across the three unit specs passing, because `it("names whole satoshis")` picks `totalUsd: 1000, usdCentsPerBtc: 7_000_000`, whose product is exact either way. A case that actually exercises the guard: `totalUsd: 1000, usdCentsPerBtc: 1_192_093` → `settlementBtc` 0.08388607 → `× 1e8` = 8388606.999999999. Without the round, a fractional satoshi count reaches the invoice mutation's amount.
3. **The injected readiness script is never executed.** `__mocks__/react-native-webview.tsx` is a root-level auto-mock, so every suite — including this one — renders a plain `View`, and `injectedJavaScript` is only ever read as a prop. `REPORT_PAGE_READY_SCRIPT` is an interpolated template string; it parses today (I checked it through `vm.Script`), but nothing in CI would catch a typo, and nothing asserts it posts the shape `isPageReadyMessage` expects.
4. **The library's real state machine is never exercised.** `sign-invest-screen.spec.tsx` replaces `useESignature` with a hand-written stand-in, so every app behaviour that depends on library semantics — decline landing in `idle` and firing `onCancel`, `retry` clearing the session, offline not calling `onError` — is asserted against the app's own model of the library rather than against the library. When `esign-react-native` changes one of those, these tests still pass. One integration spec driving the real hook with the real `createHostedFormSource` and a stubbed fetch would anchor the contract.
5. **No coverage of the timed-out-then-late mint** (Finding 3), or of two mints in flight at once.
6. **Nothing asserts the mint path against the library's own constant** (Finding 1), so a route rename in a client bump is silent at build time and at test time.
7. **The offline path is unreachable in tests.** The global `__mocks__/@react-native-community/netinfo.ts` reports `isConnected: true` for every suite. The screen's offline branch is only ever reached by forcing `status = "offline"` on the stand-in, never by the connectivity probe that produces it in production.
8. **The cover's accessibility is only half pinned.** Tests assert the WebView is hidden while covered, but not that anything replaces it: the spinner carries no `accessibilityLabel` or role, and the cover is not `accessibilityViewIsModal`, so a screen reader on this screen finds the close button and silence. Similarly, `select-invest-screen`'s options are individually `accessibilityRole="radio"` with no `radiogroup` container and no test for one.
9. **`esign-mint` has no test for a malformed configured origin** — a trailing slash yields `…//envelope/instance`, and a value that is not an http(s) origin is passed to `fetch` unchecked.

## Code Smells

1. **`jest.config.js:84` un-ignores the whole `@blinkbitcoin` scope.** The comment explains the e-sign packages' `react-native`-condition TypeScript entry, which is correct, but the pattern also pulls the precompiled `@blinkbitcoin/blink-client` and `react-native-geetest-module` through the transform on every run. `|@blinkbitcoin/esign-` would match exactly what the comment describes.
2. **`toMinorUnit(String(totalUsd))`** in `use-investment-funding.ts:86` round-trips a number through a helper whose string signature exists for text input (`Math.round(parseFloat(dollars) * 100)`). The paired `toMajorUnit` above it reads cleanly; this one reads as a type error worked around.
3. **`investment-terms.ts` is now four modules in one file:** the terms and quote maths, `INVESTMENT_OPTIONS`, the funding rule, the equity derivation, and two locale-pinned formatters. It is the natural home for the first; the formatters in particular would be easier to find in the flow's own formatting module.
4. **`esignMintUrl` is compiled in while a sibling e-sign URL is not.** `cardESignConsentUrl` already comes from remote config in `feature-flags-context.tsx`; the new endpoint does not, and `d332a852e`'s message claims otherwise for a change that was reverted before the head. See Finding 8.
5. **`settlementSats` as a mutable ref crossing an async boundary** (`sign-invest-screen.tsx:128`). The comment defends the choice against state — rightly, a state update there would restart the session — but a ref is the wrong shape for a value that must stay tied to one envelope. See Finding 3.
6. **`scriptHostname` newly exported and called at module load** for `LOCAL_MINT_ORIGIN`, a constant only a debug build ever uses. `galoy-instances` already does this for the Local instance's `fiatUrl`, so it adds no new risk, but computing it lazily inside `resolveMintOrigin`'s `__DEV__` branch would keep a dev-only turbo-module lookup out of the release import graph.
7. **The agreement's rate timestamp names no zone.** `formatAgreementTime` produces `"2026-09-07 09:09"` — correct Honduras wall-clock time, pinned by a good test — but the string written onto a legally binding document carries no offset or zone abbreviation, so whoever reads the signed agreement cannot tell which clock fixed the rate. `America/Tegucigalpa` lives only in a source constant.
8. **`resolveMintOrigin`'s comment and `mintSigningInstance`'s behaviour disagree.** The comment says an unconfigured release build means "the step reports a mint it cannot reach, which is the truth", but the code throws `signingFailure` (`ENVELOPE_CREATION_FAILED`), not `signingUnreachable` (`NETWORK_ERROR`) — the signer is told the mint failed, not that the service is unreachable.
9. **`reportSigningError` discards the original error.** `logError({ error: new Error(error.message) })` throws away the stack and the error identity, keeping only the code in `context`. Passing the error through and letting `logError` read `.code` would keep both.

PR #4236, opened 2026-09-02
fix(copy): Fix lightning and onchain send rate copy

### LINKED ISSUE #4230: fix(onchain): Correct fee tiers and support Blink wallet destinations
## Problem

The on-chain send flow has several fee-estimation and fee-tier issues:

1. All three priority tiers currently show the same fee.
2. The tier names should be **Priority**, **Standard**, and **Economy**.
3. The app cannot estimate an on-chain fee when the destination is another Blink wallet address. This prevents Blink users from reliably estimating fees when paying other Blink users, since senders cannot be expected to know which wallet the recipient uses.
4. The fee-rates table currently renders `0.2% + ~0% routing fee` for Lightning. The routing-fee mention is incorrect and should not be displayed.

## Expected behavior

- Each fee tier reflects its corresponding fee rate and estimated confirmation target.
- The tiers are labeled **Priority**, **Standard**, and **Economy**.
- Fee estimation works for valid on-chain destinations belonging to Blink wallets, just as it does for external wallet addresses.
- The Lightning row shows only `0.2%`, with no routing-fee suffix.

## Acceptance criteria

- [ ] Priority, Standard, and Economy display distinct fees whenever the underlying fee rates differ.
- [ ] The three tiers are named Priority, Standard, and Economy, in that order.
- [ ] Fee estimation succeeds for a valid on-chain address associated with a Blink wallet.
- [ ] Sending to Blink wallet addresses is not treated differently from sending to external wallet addresses.
- [ ] The fee-rates table displays Lightning payments as `0.2%` only; it does not mention a routing fee.
- [ ] The fee-rates table displays Standard on-chain payments at approximately `0.7%`.
- [ ] The fee-rates table displays Economy on-chain payments at approximately `0.5%`.
- [ ] Regression tests cover tier labels, tier fee calculation, Blink-wallet destination estimation, and omission of the Lightning routing-fee suffix.

## Priority

High (P1)

## Source

Requested by (`ihaveadifferentname`) in [Slack](https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1788157136829089).

Follow-up screenshot and correction: [Slack thread message](https://blinkbtc.slack.com/archives/C0AEYAALSCQ/p1788332924472329?thread_ts=1788187243.180829&cid=C0AEYAALSCQ).

kngako:
The Send section rendered the lightning rate as "0.2% + ~0% routing fee",
pairing a rate the account is actually charged with an estimate of a
network cost that Blink does not bill. The design quotes the rate alone,
so the row now reads "0.2%".

Drop lightningRoutingBps from FeeRatesConfig
--------------------------------------------
It was read in exactly one place — the string this commit deletes — so
with the routing fee off the screen it became a config key nothing
consumes, which invites someone to set it in Firebase and wonder why the
UI never moves.

fix(fee-rates): quote all three onchain payout speeds, not just Priority
----------------------------------------------------------------------------------------------------

The Send section listed a single "Onchain Priority (~10m)" row, so the
screen read as though onchain had one flat rate. It does not: a custodial
onchain send is priced by the payout queue it is put on, and the send
screen's own tier selector offers three of them. Standard and Economy
sends are cheaper than the one number the fee page quoted, so the page
overstated what they cost.

### DIFF
--- a/__tests__/screens/settings-screen/fee-rates-screen.spec.tsx
+++ b/__tests__/screens/settings-screen/fee-rates-screen.spec.tsx
@@ -51,7 +51,7 @@ describe("FeeRatesScreen", () => {
   })
 
   it("renders Send, Receive and Transfer sections with remote-config default rates", async () => {
-    const { getByText, getAllByText, queryByText, findByText } = render(
+    const { getByText, getAllByText, findByText } = render(
       <ContextForScreen>
         <FeeRatesScreen />
       </ContextForScreen>,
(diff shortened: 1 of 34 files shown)


esaugomez31 (approved):
LGTM

PR #4238, opened 2026-09-03
security(storage): purge the legacy key store items and gate the dependency drop

esaugomez31:
Closes blinkbitcoin/blink-wip#1163 (the purge half). Stacked on #4224, which has to merge first.

Step 4 of the migration off `react-native-secure-key-store` moves in two releases, and this is the first one. The library stays installed, because it is the only thing that can read and delete what it wrote. Dropping the dependency is the release after, and it deletes nothing on its own.

## What it does

On boot, after the mnemonic migration sweep, the app erases what it still owns in the legacy store: the six session keys, plus `mnemonic:<id>` and `mnemonic_network:<id>` for every account in the index. Once per install, behind an AsyncStorage flag that a reinstall clears, since a reinstall can leave items behind for the purge to reach again.

## Migrate, then verify, then erase

The ordering is the whole point, and the obvious implementation gets it wrong.

A read-through reports `found` for a value it read out of the legacy store even when the migrating write failed, so "this slot has been read" is not evidence the new store holds anything. Erasing on that evidence deletes the only copy: a re-login for a session slot, a seed for a mnemonic. So the new store has to answer `found` itself before anything is deleted, and every step runs inside the per-slot queue, because a read-through migrating the same slot sits between its own miss and its legacy read.

An empty legacy read is not the proof it looks like either. On iOS the native module discards the OSStatus and rejects every failed lookup with the not-found code, so a read taken before first unlock reports an empty store rather than an unreadable one. Where getting that wrong costs a re-login it is accepted, since a user who never set a PIN must not be retried forever. Where it costs a seed, the new store has to hold the value first.

The flag is set only when every key is provably gone. A partial purge leaves it unset and runs again on the next boot, so there is no failure state to recover from.

## Testing

Covered at 100%, and tested on the emulator with a self-custodial account. The purge runs, completes and records itself, and the account survives it.

## Not in this release

The legacy read-through fallback for the six session slots stays. The issue puts its removal here, but dropping it in the same release as the purge is strictly worse for someone upgrading from an older build: the PIN is read at boot, before the purge runs, so with the fallback in place that user migrates cleanly and the purge then finds the copy already gone. Without it they lose the lock and the session for nothing, since the purge deletes the data either way. It belongs with the dependency removal.

## Two things for #4224

Both are in the parent branch, and this change makes the first one matter more.

`rememberMnemonicAccount` discards its write result, so a failed tracking write is invisible. The sweep still reports `ok`, this purge erases the legacy copy, and the seed ends up in the new store un-named by the account list, where a later reinstall neither wipes it nor catches it.

`sweepMnemonicMigration` scores `absent` as a success and only `failed` as a failure, so its `ok` does not mean what its docstring says. The purge no longer depends on that, since it verifies each slot itself, but the contract and the code still disagree.

### DIFF
--- a/__tests__/self-custodial/storage/account-index.spec.ts
+++ b/__tests__/self-custodial/storage/account-index.spec.ts
@@ -4,6 +4,7 @@ const mockGetMnemonicForAccount = jest.fn()
 const mockReadMnemonicWithStatus = jest.fn()
 const mockGetMnemonicNetworkForAccount = jest.fn()
 const mockRememberMnemonicAccount = jest.fn()
+const mockPurgeLegacyKeyStore = jest.fn()
 
 jest.mock("@react-native-async-storage/async-storage", () => ({
   __esModule: true,
@@ -21,6 +22,7 @@ jest.mock("@app/utils/storage/secureStorage", () => ({
     getMnemonicNetworkForAccount: (...args: unknown[]) =>
       mockGetMnemonicNetworkForAccount(...args),
     rememberMnemonicAccount: (...args: unknown[]) => mockRememberMnemonicAccount(...args),
+    purgeLegacyKeyStore: (...args: unknown[]) => mockPurgeLegacyKeyStore(...args),
   },
 }))
 
@@ -38,11 +40,15 @@ import {
   removeSelfCustodialAccountId,
   setSelfCustodialLightningAddress,
   sweepMnemonicMigration,
+  purgeLegacyKeyStoreOnce,
   type SelfCustodialAccountEntry,
 } from "@app/self-custodial/storage/account-index"
 
 const ACCOUNT_INDEX_KEY = "selfCustodialAccountIndex"
 const LEGACY_ID_LIST_KEY = "selfCustodialAccountIds"
+const LEGACY_PURGE_DONE_KEY = "legacyKeyStorePurged"
+
+const SWEEP_OK = { status: "ok", migrated: 1 } as const
 
 const setIndex = (entries: SelfCustodialAccountEntry[]) => {
   mockGetItem.mockImplementation((key: string) =>
@@ -68,6 +74,7 @@ describe("self-custodial account-index", () => {
     mockReadMnemonicWithStatus.mockResolvedValue({ status: "absent" })
     mockGetMnemonicNetworkForAccount.mockResolvedValue(null)
     mockRememberMnemonicAccount.mockResolvedValue(undefined)
+    mockPurgeLegacyKeyStore.mockResolvedValue(true)
   })
 
   describe("listSelfCustodialAccounts", () => {
@@ -474,6 +481,93 @@ describe("self-custodial account-index", () => {
       const second = await sweepMnemonicMigration()
 
       expect(second).toEqual(first)
+    })
+  })
+
+  describe("purgeLegacyKeyStoreOnce", () => {
+    /** The index is read for the account ids, and the done-flag from its own key. */
+    const setIndexAndFlag = (
+      entries: SelfCustodialAccountEntry[],
+      flag: string | null,
+    ) => {
+      mockGetItem.mockImplementation((key: string) => {
+        if (key === ACCOUNT_INDEX_KEY) return Promise.resolve(JSON.stringify(entries))
+        if (key === LEGACY_PURGE_DONE_KEY) return Promise.resolve(flag)
+        return Promise.resolve(null)
+      })
+    }
+
+    it("purges every account in the index and records itself as done", async () => {
+      setIndexAndFlag([{ id: "a1", lightningAddress: null }], null)
+
+      const result = await purgeLegacyKeyStoreOnce(SWEEP_OK)
+
+      expect(result).toEqual({ status: "done" })
+      expect(mockPurgeLegacyKeyStore).toHaveBeenCalledWith(["a1"])
+      expect(mockSetItem).toHaveBeenCalledWith(LEGACY_PURGE_DONE_KEY, "true")
+    })
+
+    it("purges the fixed keys when the index holds no accounts", async () => {
+      setIndexAndFlag([], null)
+
+      const result = await purgeLegacyKeyStoreOnce(SWEEP_OK)
+
+      expect(result).toEqual({ status: "done" })
+      expect(mockPurgeLegacyKeyStore).toHaveBeenCalledWith([])
+    })
+
+    it("does not purge while the sweep left an account unread", async () => {
+      setIndexAndFlag([{ id: "a1", lightningAddress: null }], null)
+
+      const result = await purgeLegacyKeyStoreOnce({ status: "incomplete", failures: 1 })
+
+      expect(result).toEqual({ status: "skipped", reason: "sweep-incomplete" })
+      expect(mockPurgeLegacyKeyStore).not.toHaveBeenCalled()
+    })
+
+    it("does not purge twice on the same install", async () => {
+      setIndexAndFlag([{ id: "a1", lightningAddress: null }], "true")
+
+      const result = await purgeLegacyKeyStoreOnce(SWEEP_OK)
+
+      expect(result).toEqual({ status: "skipped", reason: "already-done" })
+      expect(mockPurgeLegacyKeyStore).not.toHaveBeenCalled()
+    })
+
+    it("purges anyway when the done-flag cannot be read", async () => {
+      mockGetItem.mockImplementation((key: string) => {
+        if (key === LEGACY_PURGE_DONE_KEY) {
+          return Promise.reject(new Error("AsyncStorage unavailable"))
+        }
+        return Promise.resolve(JSON.stringify([{ id: "a1", lightningAddress: null }]))
+      })
+
+      const result = await purgeLegacyKeyStoreOnce(SWEEP_OK)
+
+      expect(result).toEqual({ status: "done" })
+      expect(mockPurgeLegacyKeyStore).toHaveBeenCalledWith(["a1"])
+    })
+
+    it("does not purge when the index cannot be read", async () => {
+      mockGetItem.mockImplementation((key: string) => {
+        if (key === LEGACY_PURGE_DONE_KEY) return Promise.resolve(null)
+        return Promise.reject(new Error("AsyncStorage unavailable"))
+      })
+
+      const result = await purgeLegacyKeyStoreOnce(SWEEP_OK)
+
+      expect(result).toEqual({ status: "skipped", reason: "index-unreadable" })
+      expect(mockPurgeLegacyKeyStore).not.toHaveBeenCalled()
+    })
+
+    it("leaves the flag unset when a key could not be erased", async () => {
+      setIndexAndFlag([{ id: "a1", lightningAddress: null }], null)
+      mockPurgeLegacyKeyStore.mockResolvedValue(false)
+
+      const result = await purgeLegacyKeyStoreOnce(SWEEP_OK)
+
+      expect(result).toEqual({ status: "incomplete" })
+      expect(mockSetItem).not.toHaveBeenCalledWith(LEGACY_PURGE_DONE_KEY, "true")
     })
   })
 })
--- a/__tests__/store/persistent-state/index.spec.tsx
+++ b/__tests__/store/persistent-state/index.spec.tsx
@@ -23,10 +23,12 @@ jest.mock("@app/utils/storage", () => ({
 }))
 
 const mockSweepMnemonicMigration = jest.fn()
-// Scheduled off the boot path once the state has loaded; its own spec covers
-// what it does, and here it must not add reports to the ones under assertion.
+const mockPurgeLegacyKeyStoreOnce = jest.fn()
+// Scheduled off the boot path once the state has loaded; their own specs cover
+// what they do, and here they must not add reports to the ones under assertion.
 jest.mock("@app/self-custodial/storage/account-index", () => ({
   sweepMnemonicMigration: (...args: unknown[]) => mockSweepMnemonicMigration(...args),
+  purgeLegacyKeyStoreOnce: (...args: unknown[]) => mockPurgeLegacyKeyStoreOnce(...args),
 }))
 
 const mockGetActiveToken = jest.fn()
@@ -111,6 +113,7 @@ const setupStorageMockDefaults = () => {
   jest.clearAllMocks()
   storedStrings.clear()
   mockSweepMnemonicMigration.mockResolvedValue({ status: "ok", migrated: 0 })
+  mockPurgeLegacyKeyStoreOnce.mockResolvedValue({ status: "done" })
   mockSaveJson.mockResolvedValue(undefined)
   mockSaveString.mockResolvedValue(true)
   mockLoadString.mockImplementation(async (key: string) => storedStrings.get(key) ?? null)
--- a/__tests__/utils/key-store-api-stability.spec.ts
+++ b/__tests__/utils/key-store-api-stability.spec.ts
@@ -45,6 +45,8 @@ const PINNED_SURFACE = [
   "name",
   "parsePinFailureState",
   "prototype",
+  "purgeLegacyKeyStore",
+  "purgeSlot",
   "readActiveToken",
   "readIsBiometricsEnabled",
   "readIsPinEnabled",
--- a/__tests__/utils/secure-storage.spec.ts
+++ b/__tests__/utils/secure-storage.spec.ts
@@ -1733,3 +1733,220 @@ describe("KeyStoreWrapper clearUninstallSurvivingCredentials", () => {
     ])
   })
 })
+
+describe("KeyStoreWrapper.purgeLegacyKeyStore", () => {
+  /** Every key the purge is expected to name, in the order it names them. */
+  const FIXED_KEYS = [
+    "isBiometricsEnabled",
+    "PIN",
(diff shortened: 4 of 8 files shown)


kngako (changes requested):
## Findings

### Finding 1 — The purge names per-account keys from the AsyncStorage index, not the keychain list this subsystem trusts [BLOCKING]

`purgeLegacyKeyStore(accountIds)` builds its mnemonic slot list from the ids the caller passes, and `purgeLegacyKeyStoreOnce` gets those from `readIndex()` in AsyncStorage ([account-index.ts:251](app/self-custodial/storage/account-index.ts:251)). The reinstall wipe in the same class does not: it reads `KeyStoreWrapper.readMnemonicAccounts()` ([secureStorage.ts:523](app/utils/storage/secureStorage.ts:523)), the keychain-resident list that exists *because* the AsyncStorage one drifts, and it refuses to proceed when that list cannot be read or parsed.

**When it comes up:** A user deletes a self-custodial account. [use-delete-account.ts:83](app/self-custodial/hooks/use-delete-account.ts:83) calls `deleteMnemonicForAccount` and ignores its result, then removes the id from the AsyncStorage index at :87 regardless; because `deleteMnemonicForAccount` only calls `untrackMnemonicAccount` when the removal succeeded, a failed legacy erase leaves the seed under `mnemonic:<id>` in the legacy store, the id gone from the index and still present in the keychain list. The next boot's purge is never handed that id, so it never names that key — and returns `true` anyway, which writes the one-shot done flag. Probe confirmed: with `mnemonic:unknown-old-account` sitting in the legacy store and `purgeLegacyKeyStore([])`, the call returns `true` and the key is still there. On iOS that seed then stays in the keychain across an uninstall, at the legacy protection class, with the code that could have removed it about to be deleted by `#1163`. The docstring's "True only when every key is provably gone, so that a caller recording this as done cannot record it over a store that still holds something" is not what the function does.

**Proposal:** Inside `purgeLegacyKeyStore`, read `KeyStoreWrapper.readMnemonicAccounts()` and purge the union of that list and the caller's `accountIds` — both are private statics on the same class, so this is a local change and the parameter still earns its place for an index entry the keychain list never recorded. Return `false` when that read comes back `failed` or `malformed`, matching the conservative branch `clearUninstallSurvivingCredentials` already takes for the same list, so an unreadable list cannot be recorded as a completed purge. Then narrow the docstring to what remains true: every key this app can *name* is gone, and a key nothing names is out of scope.

### Finding 2 — A legacy read that only looks empty records the session slots as purged, and the flag is one-shot [BLOCKING]

`purgeSlot` reads `migratedFirst` before `legacyRead`, and on `legacy.status === "absent"` returns `requireMigrated ? migratedFirst.status === "found" : true` ([secureStorage.ts:795](app/utils/storage/secureStorage.ts:795)). The comment directly above it explains why that `absent` is not trustworthy on iOS — the native module discards the `OSStatus` and rejects every failed lookup with the not-found code, so an unreadable store and an empty store are the same answer — and then guards only the mnemonic half.

**When it comes up:** Any transient iOS keychain failure during the purge, not just the pre-first-unlock case the comment names: `errSecInteractionNotAllowed`, `errSecNotAvailable`, or a busy `securityd` at boot all come back through `legacyRead` as `absent`. Take a user with a PIN and no self-custodial accounts: every slot in the list is a session slot, each one answers `absent`, `allGone` stays `true`, and `purgeLegacyKeyStoreOnce` writes `legacyKeyStorePurged`. The legacy PIN and auth token are still in the keychain — surviving uninstall on iOS, at the old protection class — and the purge will never run again for that install. The committed spec pins this behaviour as intended ("accepts an empty legacy read for the session slots"), and the intent is defensible per slot; what is not defensible is that the same boolean also feeds a permanent, install-wide "done".

**Proposal:** Require the new store to have answered before treating an empty legacy read as proof: `if (legacy.status === "absent") return requireMigrated ? migratedFirst.status === "found" : migratedFirst.status !== "failed"`. That closes the case where the device is locked or the keychain is refusing everything — the reads fail together — while still letting a genuine fresh install finish on the first boot, because there `secureRead` answers `absent`, not `failed`. It will not catch a failure isolated to the legacy module, so pair it with the docstring narrowing from Finding 1 and with the telemetry in Finding 5, and say in the comment that "assumed empty" is an accepted residue rather than a proof.

### Finding 3 — An account in the index whose mnemonic exists nowhere stops the purge completing, for the life of the install [NON-BLOCKING]

For a mnemonic slot, `requireMigrated` is `true`, so an `absent` legacy read with an `absent` new store returns `false` ([secureStorage.ts:795](app/utils/storage/secureStorage.ts:795)). There is nothing to erase on that path and nothing left to retry, but `allGone` goes false, `purgeLegacyKeyStoreOnce` returns `incomplete`, and the flag stays unset. Probe confirmed: three consecutive `purgeLegacyKeyStore(["ghost"])` runs over two empty stores return `false, false, false`.

**When it comes up:** An iOS device restored from an iCloud or device-to-device backup. AsyncStorage's files come across, so the account index still lists every account; the keychain items do not, because both stores write `*_THIS_DEVICE_ONLY`. The sweep reports `ok` (an `absent` mnemonic read is not a failure — [account-index.ts:196](app/self-custodial/storage/account-index.ts:196)), the purge then runs, returns `false`, and repeats on every subsequent boot — roughly 31 keychain round trips per launch, forever, scheduled right after the sweep. The same state is reachable without a restore whenever an index entry outlives its key material.

**Proposal:** The `requireMigrated` rule exists to stop an *erase* on weak evidence, and this branch performs no erase — nothing is at risk from finishing it. The fix from Finding 2 already produces the right behaviour for both: with `migratedFirst.status !== "failed"` as the condition for the non-strict case and a matching relaxation here, "both stores genuinely say nothing" stops retrying while "the store could not answer" still does. If the strict `found` requirement is wanted for the mnemonic, bound the retry instead — count attempts beside the flag and stop after a handful — so an install in this state is not paying for it on every launch.

### Finding 4 — The boot purge contends with the lock screen on the PIN slots [NON-BLOCKING]

`purgeSlot` takes `onSlot(args.slot)` twice per slot (once inside `readThrough`, once for the verify-and-erase), and the fixed list includes `PIN`, `pinFailureState` and `pinAttempts`. Those are exactly the slots the lock path reads: [app-state.tsx:74](app/navigation/app-state.tsx:74) and [authentication-check-screen.tsx:44](app/screens/authentication-screen/authentication-check-screen.tsx:44) probe `PIN`, and [pin-verification.ts:45](app/screens/authentication-screen/pin-verification.ts:45) and :94 read the failure state and the PIN itself.

**When it comes up:** The purge is scheduled by `runAfterInteractions` at boot ([index.tsx:413](app/store/persistent-state/index.tsx:413)), which is the same few seconds in which the lock screen appears and the user starts typing. The per-slot queue is FIFO and `SLOT_OPERATION_TIMEOUT_MS` is 30 000 ms, so a native keychain call that hangs inside the purge — the exact scenario that timeout exists for — leaves the user's PIN verification sitting behind it for up to thirty seconds before it is even attempted. Before these commits nothing touched the PIN slot at boot except the lock path itself, so the queue had one user.

**Proposal:** Move the purge off the cold-boot path and onto the authenticated one — after the lock gate resolves there is no contention left to have, and the purge is explicitly allowed to wait for a later boot. Failing that, give the purge its own, much shorter per-slot budget (a few seconds) rather than inheriting the 30 s one: it has nothing to lose by giving up, and every slot it abandons is retried on the next launch anyway.

### Finding 5 — `PurgeResult` is computed and thrown away, so the last step of the migration has no field signal [NON-BLOCKING]

`purgeLegacyKeyStoreOnce` returns a four-way result — `done`, `incomplete`, and three `skipped` reasons — and the only caller drops it on the floor ([index.tsx:417](app/store/persistent-state/index.tsx:417)). Nothing reports, nothing counts.

**When it comes up:** Deciding whether `#1163` can actually delete the dependency. The read path was given `legacy_key_store_hit` precisely so that decision would have evidence ([secure-store-migration.ts:55](app/utils/storage/secure-store-migration.ts:55)), and the sweep reports its failures through `recordAppError`; the step that actually empties the store reports nothing at all. An install stuck in Finding 3's loop, or one that skipped for `index-unreadable` on every boot, is indistinguishable from one that finished on the first launch. Worse, the purge's own eager `readThrough` calls fire `legacy_key_store_hit` for every slot it migrates, so the one metric that does exist gets a burst from the purge itself on the release that ships it.

**Proposal:** Report `incomplete` through `recordAppError` with a fixed `dedupKey` (no ids, no key names beyond the key class the helper already strips to), and emit a one-shot analytics event on `done` so completion rate is measurable. Keep `skipped: "already-done"` silent — it is the steady state — but count `index-unreadable` and `sweep-incomplete`, since both mean the purge is being deferred by something worth knowing about. That also gives Findings 2 and 3 a way to be seen in the field rather than reasoned about.

---

## Test Coverage Gaps

1. **Commit `773b79442` has no test for what it does.** The spec change is the mock and its default return value; there is not one assertion on `mockPurgeLegacyKeyStoreOnce`. Verified by mutation: removing `.then(purgeLegacyKeyStoreOnce)` from the boot chain entirely leaves all 44 tests in `__tests__/store/persistent-state/index.spec.tsx` green. The commit's whole claim — that the purge is *chained onto* the sweep rather than scheduled beside it, and sees its result — needs one test that asserts `purgeLegacyKeyStoreOnce` was called with the sweep's result, and one that it is not called when the sweep rejects.
2. **No test for an index account whose mnemonic exists nowhere** (Finding 3). Neither the current behaviour (permanent `false`) nor the intended one is pinned, so whichever way it is resolved it will be resolved silently.
3. **Nothing pins the purge's scope.** The docstring says an id the caller does not name keeps its key; no test says so, and — the surprising half — no test says the purge still returns `true` in that case. A test with a stray `mnemonic:<id>` in the legacy store and `purgeLegacyKeyStore([])` would have made Finding 1 visible at review time.
4. **`purgeLegacyKeyStoreOnce`'s flag write is untested on both edges.** `saveString`'s boolean is discarded, so a failed flag write silently means "purge again next boot" — correct, but unasserted. And while `leaves the flag unset when a key could not be erased` covers `incomplete`, no test covers that a `skipped` result also leaves it unset.
5. **`keeps the legacy copy when the new store cannot be read back` cannot fail for the reason its name gives.** It sets `mockGetInternet.mockRejectedValue` globally, so the *first* read fails too and the migration never happens — the verify read is never reached. The test that would pin the ordering is one where the migrating write succeeds and only the second read fails; as written, moving the verify before the migration would keep it green.
6. **The iOS "empty legacy read" behaviour is pinned in only one direction.** `accepts an empty legacy read for the session slots` asserts the optimistic path; nothing asserts what happens when the new store is *also* unreadable, which is the case Finding 2 turns on and the one a fix has to keep from regressing.

## Code Smells

1. **`purgeSlot`'s second `secureRead` is dead, and its comment is wrong.** "Re-read, because the read-through above may have migrated it since" ([secureStorage.ts:803](app/utils/storage/secureStorage.ts:803)) — the `readThrough` is fully awaited *before* `onSlot` is entered, and nothing else can touch the slot while this task holds it, so `migrated` can only ever equal `migratedFirst`. Either drop the second read and use `migratedFirst` throughout, or keep it and say the real reason (belt-and-braces against a flaky adapter), because the stated one cannot happen.
2. **Two-store reasoning now lives in two modules.** `secureStorage.ts` imports `legacyRead` directly and re-implements proof-of-absence beside `eraseLegacyCopy`, whose home is `secure-store-migration.ts` — the module whose docstring describes itself as where this reasoning lives. A `purgeThrough(args, { requireMigrated })` exported from the migration module, next to `readThrough` / `removeThrough` / `writeThrough`, would keep the two-store rules in one file and leave `secureStorage.ts` with just the slot list.
3. **The session-slot list is a sixth hand-maintained copy of the key constants** ([secureStorage.ts:843](app/utils/storage/secureStorage.ts:843)). A slot added to `KeyStoreWrapper` tomorrow will be read, written, removed and wiped, and silently not purged. One `private static readonly SESSION_KEYS` array consumed by the purge (and referenced from `clearUninstallSurvivingCredentials`'s comment) removes the drift.
4. **`purgeSlot` is `private static` and is now pinned in `key-store-api-stability.spec.ts`'s `PINNED_SURFACE`.** That spec exists to keep the *public* surface from moving; adding an implementation detail to it means a harmless rename of a private helper now fails a stability test. If the pin is wanted, a comment saying why; otherwise only `purgeLegacyKeyStore` belongs there.
5. **Three commits, zero bodies** — the same standing smell as the previous two branch reviews. `0e86e84f7` is the one that hurts: the ordering argument, the `requireMigrated` split and the iOS `absent` caveat are all in code comments, which is the right place for them, but `git log` says nothing about why a purge appears now or what `#1163` is waiting on.
6. **Author string is still `Esaú Gómez)`** on all three commits — stray closing paren, carried from the earlier branches. `--amend --reset-author` or a squash-merge fixes it.
7. **`saveString`'s result is discarded without a marker** ([account-index.ts:262](app/self-custodial/storage/account-index.ts:262)). Deliberate and correct — a failed flag write just means the purge repeats — but every other ignored boolean in this subsystem carries a one-line comment saying so, and this one does not.

grimen (changes requested):
## Findings

**1. None of the previous review's findings are addressed at 60fb753 (blocking).**
The force-push since @kngako's review (773b794 → 60fb753) only rebases the three commits onto #4224's new head. `git range-diff` shows no change beyond the context it now sits in. The two blocking findings are still in the code: `purgeLegacyKeyStore` still names mnemonics only from the ids the AsyncStorage index hands it (`account-index.ts:277`), and an empty legacy read still counts as gone for every session slot (`secureStorage.ts:819-821`). The boot chain is also still unasserted.

*Example:* a user deletes a self-custodial account and the legacy erase fails. The id leaves the index but the seed stays under `mnemonic:<id>`. The next boot's purge never names it, returns `true` and writes `legacyKeyStorePurged`. That is the case the earlier review traced, and it still reproduces.

*Proposal:* one option is to reply per finding, either with a fix or with the reason it is accepted, before the next round, so the re-review can diff against a stated position.

**2. If this ships in the same release as #4224, the legacy mnemonic copy #4224 keeps for downgrades lasts only a few seconds.**
`mnemonicSlotFor` sets `deleteLegacyOnMigrate: false`, and its docstring gives the reason: "What retaining it buys is a downgrade and rollback that still find the value. The copies go in the explicit purge" (`secureStorage.ts:155-159`). #4224 targets `main` and has not shipped yet. 3.0.41 still reads mnemonics straight from `RNSecureKeyStore`. This PR is stacked on #4224. If both land before the next cut, the first boot of that build runs the sweep and then the purge in the same `runAfterInteractions` callback (`persistent-state/index.tsx:413-419`). The purge erases every legacy mnemonic the sweep has just migrated, so no build ever carries the retained copy long enough for a rollback to use it.

*Example:* a tester or sideloading user goes from the new build back to 3.0.41. `use-sdk-lifecycle.ts:268-270` gets `null` for the mnemonic and parks the wallet as `Unavailable`. Deleting that account then probes `NoMnemonic`, which `profile-row.tsx:128` answers with the plain confirm dialog and not the has-funds warning. After that, a user with funds and no backup phrase has only the new-store copy, and nothing in 3.0.41 can read it.

*Proposal:* one option is to hold this PR for the release after #4224 and say so in the description, next to the two-release plan it already sets out for the dependency drop. If shipping both together is the intent, the `mnemonicSlotFor` docstring and #4224's description should stop promising a rollback path.

**3. The proposed fix for Finding 2 of the previous review does not close the case that finding describes.**
The six session slots were written `ALWAYS_THIS_DEVICE_ONLY`, which `secureStorage.ts:115-119` notes was readable while the device was locked. So before first unlock a legacy session read comes back `found`, not `absent`, and the purge already returns `false` there, because the new store's `AFTER_FIRST_UNLOCK` read fails. That leaves one way for a session slot to answer `absent` while still holding a value: a failure in the legacy lookup alone, which `RNSecureKeyStore.m:183-187` turns into a 404 because it drops the `OSStatus`. In that case the new store answers normally, `absent` for a slot that never migrated, so `migratedFirst.status !== "failed"` still passes and the flag is still written. It takes two such failures in a row, since the `readThrough` at the top of `purgeSlot` also reads legacy first.

*Example:* on the first boot of this build, a slot the boot path has not read yet is still only in the legacy store, and `securityd` is busy. Both legacy lookups return nil, `purgeSlot` scores the slot gone, and the flag is written. The copy stays behind, still readable while the device is locked and surviving uninstall, and no later boot looks again.

*Proposal:* one option is to state this plainly as accepted residue: the docstring narrowing and the `incomplete` telemetry from the earlier review carry the weight, and the `requireMigrated` tweak can be dropped. If it needs closing, two consecutive clean boots before the flag is written would do it, at the cost of one extra purge pass per install.

## Test Coverage Gaps

- No spec covers a legacy store that holds a session value whose `get` rejects with the not-found code while the new store answers `absent`. `accepts an empty legacy read for the session slots` runs over a store that really is empty, and finding 3 hides in the untested case.
- None of the `purgeLegacyKeyStoreOnce` specs feeds `readIndex` a malformed index, meaning an element that fails `isEntry` or a payload that is not an array (`account-index.ts:63`). So a purge that returns `done` while a legacy `mnemonic:<id>` under a dropped id survives is unasserted.
- Removing `.then(purgeLegacyKeyStoreOnce)` from `persistent-state/index.tsx` still leaves all 1107 related tests green at 60fb753. It is the only one of six mutations tried that survives, and it is the gap finding 1 carries over from the previous review.

PR #4242, opened 2026-09-04
fix(migration): let a refused lightning-address re-point continue the migration

esaugomez31:
Closes the client-side half of [blink-wip#1211](https://github.com/blinkbitcoin/blink-wip/issues/1211).

## Why

When a migration stalls, support clears the flow so the user can start again. That restart provisions a **new** pubkey, and the backend then tries to move the Lightning address across using the same username. It cannot: the address already belongs to the pubkey of the failed attempt, so `transferLnAddressesToSpark` hits a conflict and — since the pubkeys differ — answers `FAILED` (`resolveConflict` in `core/api/src/app/migration-flow/transfer-ln-address.ts`, which never re-points to a different pubkey by design).

The app treated that `FAILED` as terminal and sent the user to Contact support. So every rescue attempt ended exactly where it started, with the account locked server-side and the funds out of reach.

## What changed

A refused re-point is now a settled outcome, not a failure:

- **Approve stays enabled.** The address is a convenience, the funds are not, and only the funds are irreversible.
- **A warning banner says the address could not be moved**, so approving is an informed choice rather than a silent loss.
- `MigrationSupportReason.LnAddressTransferFailed` is removed — nothing raises it any more.

A missing device key still hands over to support, unchanged: that cause breaks the commit too, so letting it through would offer a transfer that cannot complete.

## Screenshots

| Balances Overview, re-point refused |
| --- |
|  |

## Testing

Reproduced end to end on a locked staging account by mocking the mutation payload: the screen used to redirect to Contact support with `Reason: ln-address-transfer-failed` after ~3s, and now stays put with Approve enabled and the banner shown.

Unit coverage on the screen is at 100% branches and lines, with four tests replacing the one that assumed the rejection was terminal. Copy is translated across all 28 locales.


kngako (approved):
Review addressed. And finding two accepted left as is.

PR #4243, opened 2026-09-04
fix(migration): stop reading an unreadable store as a device with no checkpoint

esaugomez31:
Stacked on #4242. Answers the other half of [blink-wip#1211](https://github.com/blinkbitcoin/blink-wip/issues/1211): why the app loses its migration checkpoint.

## Why

`loadJson` swallows every storage error and returns `null`, so **"I could not read the store" and "there is nothing stored" arrive identical**. Three layers of handling were built on top of that — `loadCheckpoint` rethrows, the hooks raise `hasError`, the gate shows a retry instead of handing over — and none of them could ever fire, because the exception died one layer below. Their tests passed by mocking the wrapper to reject, a behaviour the real one cannot produce.

The consequences, both reproduced:

- The gate read a device it could not query as a wiped one and sent the user to Contact support with `locked-without-checkpoint`, which is terminal.
- Worse, the next save merged against that phantom "nothing" and **overwrote the real record without its `accountId` or `expectedReceiveSats`** — the one genuinely irreversible step. A dump of the emulator's own store after reproducing the original bug shows exactly that: a `balancesOverview` checkpoint stripped of its account id.

## What changed

- **`loadJsonOrThrow`** — the read counterpart of the existing `saveJson`. Only the read propagates; content that will not parse still reads as absent, so a corrupt value stays replaceable instead of stranding the key.
- **The checkpoint and pending-wallet reads use it.** A failed read no longer deletes the record, and a save refuses rather than overwriting a record it could not read.
- **`classifyStorageFailure`** — mapped against async-storage 2.2.0's native error surface. It names the one failure a user can act on (out of space) and refuses to conclude anything from the rest, since Android answers an unopenable database with the literal string `Database Error`. A spec pins the library version so an upgrade forces the map to be re-derived.
- **The gate explains itself.** It says what failed instead of "there was an error", tells the user to free space when that is the cause, and after three failed retries offers a way out with a new reason, `storage-unreadable` — which says the store could not be read rather than asserting the records are gone.

Accepting the terms now surfaces a failed checkpoint write instead of leaving Continue looking broken.

`locked-without-checkpoint` is not removed — it still fires, but now only once the reads have actually answered and confirmed both resume sources are gone: a reinstalled or wiped device, a checkpoint past its 48h expiry with no reusable wallet behind it, or a provisioned wallet whose key is no longer on the device. What it can no longer do is claim any of that about a device it simply could not read.

The 48h expiry itself is untouched here and remains a cause of its own.

## Screenshots

| Store unreadable | Device out of space | Escape after 3 retries | Support, new reason |
| --- | --- | --- | --- |
|  |  |  |  |

For reference, what the same failure used to produce:

| Before: terminal handover |
| --- |
|  |

## Testing

Every path above was reproduced on a locked staging account by simulating an unreadable store, including the original terminal handover the fix removes.

A new integration spec guards the seam that hid this: unlike the spec beside it, it mocks AsyncStorage rather than the storage wrapper, so the real read helper takes part. `migration-gate.tsx`, `migration-checkpoint-storage.ts` and `storage-failure.ts` are at 100% across the board, and the copy is translated across all 28 locales.

The six commits are ordered so each one builds on its own, and the second already stops the data loss even without the rest.


kngako (changes requested):
## Verdict

| # | Finding | Blocking |
|:--|:--|:--|
| 1 | The commit-point retry blames the network for the device's failure | **Yes** |
| 2 | A latched flag lets a stale expected-receive be approved with no retry | No |
| 3 | Every error leaving `ensureAccount` is classified as a storage failure | No |
| G1 | The Approve gate is unpinned — mutation-proven | **Yes** |
| G2 | The out-of-space toast in `use-migration-account.ts` is unpinned | No |
| G3 | The stale-figure path has no test | No |

---

## Blocking findings

### 1. The commit-point retry tells a full-disk user to check their internet

`4760eaaf0` folds `isCommitPointWriteRetryable` into `isRetryable`, which is right. But `isRetryable` also drives the banner above the button, and that banner is hardcoded at [balances-overview-screen.tsx:404](app/screens/account-migration/to-non-custodial/balances-overview-screen.tsx:404):

```
{isRetryable ? <Text …>{LL.errors.network.connection()}</Text> : null}
```

`errors.network.connection` is *"Connection issue.\nVerify your internet connection"*. So a user standing at the commit point whose store refused the write — the exact scenario this stack was built for — is sent to check their wifi. They are on wifi. They tap retry, the store refuses again, and the screen repeats the same wrong instruction; the hardware back is swallowed here, so that is the whole of what they can do.

This is the defect the branch exists to prevent, one screen over. `hasServerDataError` was introduced in the gate specifically so a device failure is never reported as the network's, and `44aef1e9e` rewrote a comment to keep that line honest. Here the inverse slips through unremarked.

> **Fix.** Everything needed is already in hand: `recordCommitPoint` receives `failure` from `saveCheckpoint`, and `storageUnavailable.outOfSpaceBody` / `unreadableBody` exist in all 28 locales. Keep the failure kind in state next to `hasCommitPointWriteFailed`, and pick the banner from it — the storage copy when the commit-point write is what failed, `errors.network.connection()` for the preview, start and re-point retries that genuinely are connection issues. That also makes the retry actionable for the out-of-space case, which is the only one the user can clear.

### G1. The Approve gate is unpinned, and the three tests named for it do not reach it

Deleting `!isCommitPointRecorded` from `isApproveDisabled` leaves **all 1401 tests green**.

The three tests `4760eaaf0` added are all satisfied by a different mechanism. `isRetryable` swaps the Approve button out for Retry entirely ([balances-overview-screen.tsx:407-418](app/screens/account-migration/to-non-custodial/balances-overview-screen.tsx:407)), so on a refused write Approve is *absent*, not disabled — `queryByTestId(...)` returns null either way, with or without the term under test. And in *"opens Approve once that record lands"* the flag is true, so `!isCommitPointRecorded` contributes nothing there either.

The state the term actually guards is neither of those. It is `!isCommitPointRecorded && !hasCommitPointWriteFailed` — the window where the write is still **in flight**: preview ready, start settled, re-point settled, and the AsyncStorage round-trip not yet back. Without the term, Approve is live and tappable in that window. On a slow or contended store, which is this stack's whole subject, that window is not narrow.

> **Fix.** Hold the write open and assert the button is disabled rather than absent: resolve `saveCheckpoint` from a promise you control, render with everything else settled, and check `getByTestId("migration-balances-overview-approve")` is `toBeDisabled()` before resolving it — then enabled after. That is the only assertion that fails when the term is removed, because it is the only one that constructs the state the term is for.

---

## Non-blocking findings

### 2. A latched flag lets a stale expected-receive be approved, with no retry offered

`setIsCommitPointRecorded((wasRecorded) => wasRecorded || isSaved)` never unlatches, and `isCommitPointWriteRetryable` is `hasCommitPointWriteFailed && !isCommitPointRecorded` — so once any write has landed, a later refusal is invisible.

The figure can change after that first write: `expectedReceiveSats` comes from the preview, which refetches on focus, and the fee and receive move with routing. Second visit, new figure, store now refusing: `isCommitPointRecorded` stays true, `isCommitPointWriteRetryable` is false, Approve is enabled, and the disk still holds the old number. The transferring screen's receive gate then waits on a figure that no longer matches what is coming, which ends at the `ReceiveDelayed` handover.

> **Fix.** Latch the *figure*, not the fact. Keep the `expectedReceiveSats` that was actually persisted in state and treat the commit point as recorded only while it equals the current one — `isCommitPointRecorded = recordedReceiveSats === expectedReceiveSats`. A changed figure then re-opens both the Approve gate and the retry, which is the behaviour the latch comment is reaching for ("a later failure cannot un-record a write that already landed") without also covering the case where what landed is no longer the right answer.

### 3. Every error leaving `ensureAccount` is classified as a storage failure

At [use-migration-account.ts:71](app/screens/account-migration/hooks/use-migration-account.ts:71), `const failure = checkpointFailure ?? classifyStorageFailure(err)` sits in a catch that also receives provisioning errors, Breez errors and anything else `guard.run` throws. Today it is harmless: those messages do not contain the markers, so they classify as `Unknown` and get the generic copy.

It is right by luck rather than by construction. `OUT_OF_SPACE_MARKERS` contains the bare fragment `"out of space"`, and the day a provisioning or key-generation error phrases itself that way, a user who is not out of space is told to free up storage on the screen where they create their wallet.

> **Fix.** Only classify the error that came from storage. `savePendingAccount` is the one other throwing storage call in that block — catch it at its own call site and set `checkpointFailure` there, the way the checkpoint write already does, then let the outer catch use `checkpointFailure` alone and fall back to the generic copy when it is null. The comment above the `let` already describes exactly this split; the code just doesn't hold it.

---

## Test coverage gaps

G1 is above. The rest:

- **G2 — the out-of-space toast on the provisioning path is unpinned.** Forcing `isOutOfSpace` to `false` in `use-migration-account.ts` leaves all 870 account-migration and accept-t-and-c tests green. `accept-t-and-c.spec.tsx:139` pins the identical branch properly; this second call site added by `204e95a8d` got none. One case resolving `saveCheckpoint` as `{ isSaved: false, failure: StorageFailure.OutOfSpace }` and asserting the toast message closes it.
- **G3 — the stale-figure path (finding 2) has no test.** No case changes `expectedReceiveSats` between two `recordCommitPoint` runs. A test that lands the first write, moves the preview figure, refuses the second, and asserts Approve is withheld would fail today.

---

## Code smells

1. **`hasGateDataError` restates `hasServerDataError`.** [migration-gate.tsx:185-188](app/screens/account-migration/to-non-custodial/migration-gate.tsx:185) spells out `apiKeysError || balancesError || lockError` again, fifty lines after naming that exact expression. Two copies that must stay identical, and the named one carries a comment warning readers not to widen it. `hasServerDataError || (isMigrationLocked && hasResumeDataError)` is the same value and cannot drift.
2. **The failure kind is smuggled across a `throw` in a mutable `let`.** `checkpointFailure` is declared in `ensureAccount`, assigned inside a nested closure, and read in the catch. It works, but the value travels by side effect through a control-flow edge. Returning a discriminated result from the inner function and throwing at the call site keeps it on the value path — and makes finding 3's fix natural rather than another assignment.
3. **The module doc comment in `storage-failure.ts` is orphaned.** `44aef1e9e` inserted `PINNED_ASYNC_STORAGE_VERSION` with its own JSDoc directly between the file-level block and `StorageFailure`, with no blank line before the const. The file's overview now reads as documentation for the version string. Move the pin below `StorageFailure`, or separate the blocks.
4. **The commit point writes twice on every mount.** `saveCheckpoint`'s identity depends on `expectedReceiveSats`, which is read from `stored`, which the successful write updates — so the effect re-runs once and writes the same record again before the deps settle. Harmless and pre-existing, but it doubles the writes on the screen most likely to be talking to a struggling store.

---

## What changed since the previous review

All six findings, three coverage gaps and four smells from the last pass are addressed. Two are worth noting:

- **The handover fix is better than what I proposed.** I suggested adding `StorageUnreadable` to `RESTART_RESOLVABLE_REASONS`; `38fc221ea` adds a distinct `MigrationSupportOrigin.GateStorage` instead, which restores the back door without swapping in the self-help copy or resetting the nav stack.
- **`8d1a24978` caught a defect I missed** — a stray space in the Japanese `outOfSpaceBody`. I read that string while checking locale parity and did not flag it.

### Mutation results

Nine reverts of the stack's load-bearing logic, run against all 92 suites:

| Mutation | Caught |
|:--|:--|
| Reset the retry counter when the branch goes quiet (reverts `1e614c3ef`) | yes |
| Drop `!isRetrying` from `shouldOfferHandover` | yes |
| Drop `isMigrationLocked` from `isStorageReadFailure` | yes |
| Count retries made while the server was what failed | yes |
| Make the storage handover terminal again (reverts `38fc221ea`) | yes |
| Report the checkpoint step optimistically (reverts `7e724610e`) | yes |
| Stop classifying the pending-wallet read failure | yes |
| Read an unreadable pending record as an empty one (reverts the core fix) | yes (3 tests) |
| **Drop `!isCommitPointRecorded` from `isApproveDisabled`** | **no** |

The three tests that were hollow last time now fail when their logic is removed.

---

## What was verified

| Check | Result |
|:--|:--|
| `jest` — account-migration, self-custodial, accept-t-and-c, storage | 92 suites, 1401 passed |
| `tsc --noEmit` | clean |
| `eslint --no-ignore` on all 12 changed app files | clean |
| Locale parity for `AccountMigration.storageUnavailable` | 3 keys × 28 locales, none missing, whitespace clean |

Mutations were applied one at a time and reverted; the worktree is clean and the `node_modules` symlink used for the runs has been removed.

PR #4245, opened 2026-09-05
fix(migration): let the server, not a clock, decide when a checkpoint expires

esaugomez31:
Stacked on #4243. Closes the last cause of [blink-wip#1211](https://github.com/blinkbitcoin/blink-wip/issues/1211): the checkpoint expiring on a clock the server does not share.

## Why

The server keeps a migration locked for as long as it needs to. The device forgot it after 48 hours. Anyone stuck longer than that was guaranteed to reach `locked-without-checkpoint`, which is what the reported account did — its commit failed on Aug 31 and the record was gone by Sep 2.

Raising the number only moves the cliff. As @grimen put it on the ticket: *"a state machine (~lock) that randomly resets makes it a flawed state machine — so yes state should not randomly change/reset."* The lock is already the authority — `useMigrationLock`'s own docstring says the checkpoint is "demoted to remembering WHICH screen to resume on" — and the clock contradicted that.

## What changed

- **The expiry is gone, and no deletion replaces it.** The only paths that clear a checkpoint remain the two real ones: the migration completed, or the flow walked past it. A record the server still backs survives indefinitely; one it does not is simply never acted on.
- **Resuming requires a live server flow.** The entry dispatcher and the flow's own skip-ahead both ask before jumping to the commit point. Once support clears a stuck flow the account reads unlocked, and the user starts over instead of landing on a commit screen for a migration that no longer exists. An unanswered server never resumes either — walking the flow costs a few taps, resuming a dead one costs the funds.
- **A restart drops the previous run's expected receive figure.** The write-once rule that guards against the post-drain zero (#4102) used to be bounded by the expiry; it is now bounded by the step falling back behind the commit point, which only a restart does. The figure is refused from the update as well, since the caller re-sends what it already holds to heal a failed write.

The provisioned wallet is untouched by all of this: a restarted flow still reuses it rather than provisioning an orphan, which also means the Lightning address is already on that pubkey and comes back `ALREADY_TRANSFERRED`.

## Screenshots

| Server reset: starts over | And walks the whole flow |
| --- | --- |
|  |  |

## Testing

Verified on device against the emulator's own AsyncStorage, reading the record back out of SQLite at each step:

| Scenario | Result |
| --- | --- |
| 30-day-old checkpoint, server holds the flow | Resumes — the expiry is gone |
| Same checkpoint, server reset | Does not resume: starts over |
| Continuing from there | Walks the flow, no skip back to the commit point |
| After accepting the terms | Receive figure absent, provisioned wallet kept |
| The record on disk | Survives every case |

The same checkpoint drives both of the first two rows; only the server's answer differs.

Device testing is also what found the two bugs the unit tests could not: the flow skipping back onto the commit point one tap after the entry screen refused to, and the receive figure riding back in through the caller's echo. Both now have regressions covering them.

`migration-checkpoint-storage.ts`, `use-migration-next-step.ts` and `migration-entry-screen.tsx` are at 100% across the board.

### DIFF
--- a/__tests__/screens/account-migration/hooks/use-migration-next-step.spec.ts
+++ b/__tests__/screens/account-migration/hooks/use-migration-next-step.spec.ts
@@ -10,6 +10,18 @@ let mockIsAtCommitPoint = false
 let mockCheckpointLoading = false
 let mockHasTransactions = false
 let mockTransactionsLoading = false
+let mockIsLocked = true
+let mockLockLoading = false
+let mockLockError = false
+
+jest.mock("@app/screens/account-migration/hooks/use-migration-lock", () => ({
+  useMigrationLock: () => ({
+    isLocked: mockIsLocked,
+    loading: mockLockLoading,
+    hasError: mockLockError,
+    refetch: jest.fn(),
+  }),
+}))
 
 jest.mock("@react-navigation/native", () => ({
   ...jest.requireActual("@react-navigation/native"),
@@ -39,6 +51,9 @@ describe("useMigrationNextStep", () => {
     mockCheckpointLoading = false
     mockHasTransactions = false
     mockTransactionsLoading = false
+    mockIsLocked = true
+    mockLockLoading = false
+    mockLockError = false
   })
 
   it("offers the history download to a fresh migration with history", () => {
@@ -143,5 +158,70 @@ describe("useMigrationNextStep", () => {
     const { result } = renderHook(() => useMigrationNextStep())
 
     expect(result.current.loading).toBe(true)
+  })
+
+  describe("a commit point the server no longer holds open", () => {
+    /** The device left off on the commit screen, but support has since cleared the flow. */
+    const arriveWithStaleCommitPoint = (): void => {
+      mockIsAtCommitPoint = true
+      mockIsLocked = false
+    }
+
+    it("starts the flow over instead of returning to the commit screen", () => {
+      arriveWithStaleCommitPoint()
+
+      const { result } = renderHook(() => useMigrationNextStep())
+      act(() => {
+        result.current.goToNextStep()
+      })
+
+      expect(mockNavigate).toHaveBeenCalledWith("accountMigrationExplainer")
+      expect(mockNavigateToCheckpoint).not.toHaveBeenCalled()
+    })
+
+    it("does the same for a skip guard, replacing rather than pushing", () => {
+      arriveWithStaleCommitPoint()
+
+      const { result } = renderHook(() => useMigrationNextStep())
+      act(() => {
+        result.current.replaceToNextStep()
+      })
+
+      expect(mockReplace).toHaveBeenCalledWith("accountMigrationExplainer")
+      expect(mockReplaceToCheckpoint).not.toHaveBeenCalled()
+    })
+
+    it("offers the history download again, since this run starts from the top", () => {
+      arriveWithStaleCommitPoint()
+      mockHasTransactions = true
+
+      const { result } = renderHook(() => useMigrationNextStep())
+      act(() => {
+        result.current.goToNextStep()
+      })
+
+      expect(mockNavigate).toHaveBeenCalledWith("accountMigrationDownloadHistory")
+    })
+
+    it("never skips ahead on an unanswered server, only on a confirmed one", () => {
+      mockIsAtCommitPoint = true
+      mockLockError = true
+
+      const { result } = renderHook(() => useMigrationNextStep())
+      act(() => {
+        result.current.goToNextStep()
+      })
+
+      expect(mockNavigateToCheckpoint).not.toHaveBeenCalled()
+    })
+
+    it("waits while the server is still being asked", () => {
+      mockIsAtCommitPoint = true
+      mockLockLoading = true
+
+      const { result } = renderHook(() => useMigrationNextStep())
+
+      expect(result.current.loading).toBe(true)
+    })
   })
 })
--- a/__tests__/screens/account-migration/to-non-custodial/migration-entry-screen.spec.tsx
+++ b/__tests__/screens/account-migration/to-non-custodial/migration-entry-screen.spec.tsx
@@ -34,6 +34,18 @@ let mockIsAtCommitPoint = false
 let mockHasResumableCheckpoint = false
 let mockCheckpointLoading = false
 let mockSelfCustodialDisabled = false
+let mockIsLocked = true
+let mockLockLoading = false
+let mockLockError = false
+
+jest.mock("@app/screens/account-migration/hooks/use-migration-lock", () => ({
+  useMigrationLock: () => ({
+    isLocked: mockIsLocked,
+    loading: mockLockLoading,
+    hasError: mockLockError,
+    refetch: jest.fn(),
+  }),
+}))
 
 jest.mock("@app/screens/account-migration/hooks", () => ({
   useMigrationCheckpoint: () => ({
@@ -64,6 +76,9 @@ describe("MigrationEntryScreen", () => {
     mockHasResumableCheckpoint = false
     mockCheckpointLoading = false
     mockSelfCustodialDisabled = false
+    mockIsLocked = true
+    mockLockLoading = false
+    mockLockError = false
     mockRemoteConfigReady = true
     mockRegistryLoading = false
   })
@@ -152,5 +167,52 @@ describe("MigrationEntryScreen", () => {
 
     expect(mockReplace).toHaveBeenCalledWith("Primary")
     expect(mockGoBack).not.toHaveBeenCalled()
+  })
+
+  describe("the server owns whether a migration is still open", () => {
+    /** A device that left off on the commit screen: everything local says resume. */
+    const arriveWithCommitPointCheckpoint = (): void => {
+      mockIsAtCommitPoint = true
+      mockHasResumableCheckpoint = true
+    }
+
+    it("resumes while the server still holds the flow", () => {
+      arriveWithCommitPointCheckpoint()
+
+      render(<MigrationEntryScreen />)
+
+      expect(mockReplaceToCheckpoint).toHaveBeenCalledTimes(1)
+      expect(mockReplace).not.toHaveBeenCalled()
+    })
+
+    it("starts over once support has cleared the flow, whatever the device remembers", () => {
+      arriveWithCommitPointCheckpoint()
+      mockIsLocked = false
+
+      render(<MigrationEntryScreen />)
+
+      expect(mockReplace).toHaveBeenCalledWith("accountMigrationStart")
+      expect(mockReplaceToCheckpoint).not.toHaveBeenCalled()
+    })
+
+    it("waits for the server rather than resuming on a stale record", () => {
+      arriveWithCommitPointCheckpoint()
+      mockLockLoading = true
+
+      render(<MigrationEntryScreen />)
+
+      expect(mockReplaceToCheckpoint).not.toHaveBeenCalled()
+      expect(mockReplace).not.toHaveBeenCalled()
+    })
+
+    it("starts over when the server could not be asked, never resumes on a guess", () => {
+      arriveWithCommitPointCheckpoint()
+      mockLockError = true
+
+      render(<MigrationEntryScreen />)
+
+      expect(mockReplace).toHaveBeenCalledWith("accountMigrationStart")
+      expect(mockReplaceToCheckpoint).not.toHaveBeenCalled()
+    })
   })
 })
--- a/__tests__/screens/account-migration/utils/migration-checkpoint-storage.spec.ts
+++ b/__tests__/screens/account-migration/utils/migration-checkpoint-storage.spec.ts
@@ -3,7 +3,6 @@ import {
   clearCheckpointFromStorage,
   getStorageKey,
   isCommitPointCheckpoint,
-  isExpired,
   loadCheckpoint,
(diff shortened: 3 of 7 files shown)


kngako (changes requested):
Here's the claude review. 

I would say finding 3 is not blocking because transient lock-read failures are very rare to come across.

---

# Code review — migration checkpoint expiry & server-lock resume

**Branch:** `claude/review-recent-commits-21e920`
**Commits reviewed (4 most recent):**

| SHA | Subject |
| --- | --- |
| `f1f42fc4d` | fix(migration): stop expiring the checkpoint on a clock the server does not share |
| `7a2d58a3f` | fix(migration): resume only while the server still holds the migration open |
| `9e47edece` | fix(migration): drop a previous run's receive figure even when the caller echoes it |
| `076bd3429` | fix(migration): stop the flow skipping back onto a commit point the server closed |

**Verification run:** `npx jest __tests__/screens/account-migration --maxWorkers=4` → 55 suites / 871 tests pass. `npx tsc --noEmit -p .` clean. ESLint clean on all six changed files (`--no-ignore`).

**Verdict:** the direction is right — the wall-clock expiry was never the right authority, and moving "is this flow still open?" to the server is the correct call. But the new guard is applied at two of the four places that can reach the commit screen, and "unlocked" is being read as one thing when it is two. Three blocking findings below.

---

## Summary

| # | Finding | Blocking |
| --- | --- | --- |
| 1 | [History-download detour walks straight past the stale-commit-point guard](#1-history-download-detour-walks-straight-past-the-stale-commit-point-guard) | **Yes** |
| 2 | [`isLocked === false` conflates "support cleared it" with "completed, swap still pending"](#2-islocked--false-conflates-support-cleared-it-with-completed-swap-still-pending) | **Yes** |
| 3 | [A transient lock-read failure irreversibly destroys `expectedReceiveSats`](#3-a-transient-lock-read-failure-irreversibly-destroys-expectedreceivesats) | **Yes** |
| 4 | [Three independent no-cache reads of the same lock on one screen](#4-three-independent-no-cache-reads-of-the-same-lock-on-one-screen) | No |
| 5 | [The restart signal is inferred from a step regression any background save can forge](#5-the-restart-signal-is-inferred-from-a-step-regression-any-background-save-can-forge) | No |

---

## Findings

### 1. History-download detour walks straight past the stale-commit-point guard

**Blocking.**

In [use-migration-next-step.ts:41](app/screens/account-migration/hooks/use-migration-next-step.ts:41) the history branch is evaluated before the new `isStaleCommitPoint` branch, and `shouldOfferHistoryDownload` was widened from `!isAtCommitPoint` to `!isResumable` — so a closed flow now *re-enables* it. A user with transaction history whose migration support has cleared taps Migrate, is correctly offered the CSV export, taps Skip, and [download-history-screen.tsx:37](app/screens/account-migration/to-non-custodial/download-history-screen.tsx:37) calls `navigateToCheckpoint()` directly rather than going through `useMigrationNextStep`. That resolver still reads the stored step, which is still `BalancesOverview`, so it lands them on `accountMigrationBalancesOverview` — the exact commit screen this commit exists to keep them off, and the screen whose own docblock says landing there with figures "declares the migration started server-side".

The new test `"offers the history download again, since this run starts from the top"` pins the first hop of this path and stops there, so the suite is green while the guard is bypassed.

**Proposal.** Make the guard live where the routing decision lives instead of at one caller: teach `useMigrationCheckpoint`'s `resolveDestination` to fall back to `resolveCheckpointRoute(null)` when the lock says the flow is closed, so every consumer of `navigateToCheckpoint`/`replaceToCheckpoint` inherits it. If you would rather not put a network read inside that hook, the narrower fix is to route `MigrationDownloadHistoryScreen`'s secondary action through `useMigrationNextStep().goToNextStep` like every other step screen does, and add an ordering assertion that a stale commit point with history ends at the explainer, not at the commit screen.

---

### 2. `isLocked === false` conflates "support cleared it" with "completed, swap still pending"

**Blocking.**

[use-migration-lock.ts:44](app/screens/account-migration/hooks/use-migration-lock.ts:44) locks on `InProgress | Transferring | Failed`, so `Completed` reads unlocked — pinned by `use-migration-lock.spec.ts:62` ("releases a completed migration"). But `useResumeCompletedMigration` exists specifically for the window where the server has said COMPLETED and this device has not swapped yet, and that window is open-ended: the receive gate "never gives up", and the checkpoint is only cleared inside `finishOnDevice`, which runs after `isReceiveConfirmed`. For that entire window the stored step is still `BalancesOverview` while the lock reads false.

Concretely: app killed mid-transfer, relaunched, funds still in flight. The user opens Settings → Migrate. The entry screen now computes `hasLiveServerFlow = false` and replaces to the gate; the gate is not post-deadline so it renders the voluntary "Time to upgrade" pitch; they tap Migrate and are walked into a brand-new migration for an account that already migrated. Step one of that run saves `BackupMethod` over the commit point, which under `9e47edece` drops `expectedReceiveSats` — the figure the background `useResumeCompletedMigration` is at that moment waiting on.

**Proposal.** "Not locked" is not "closed"; the two need separating at the source. Add a distinct signal to `useMigrationLock` — e.g. `isSettledClosed`, true only for `NotStarted` — and have `isStaleCommitPoint` and `hasLiveServerFlow` key off that rather than off `!isLocked`. `Completed` should continue to resume (or, better, hold) so the resume hook can finish its swap, since a completed-but-unswapped migration is the one case where the local commit point is still the truth. Add a case to `use-migration-next-step.spec.ts` fixing `Completed` as a non-restarting status.

---

### 3. A transient lock-read failure irreversibly destroys `expectedReceiveSats`

**Blocking.**

`isResumable = isAtCommitPoint && isLocked && !lockError` makes `isStaleCommitPoint` true whenever the read merely *failed*, and the hook's only two outcomes are resume or restart — there is no third "hold" branch. The entry screen gets away with the same rule because its error path lands on the gate, whose `hasGateDataError` blocks with a retry; `useMigrationNextStep` has no such surface, so one dropped `migrationStatus` request behind a CTA tap restarts a live migration. The restart's first checkpoint save regresses the step, `hasRestarted` fires, and `9e47edece` refuses the figure from both the record and the echoed update. That figure is knowable only at the commit point before the drain, so it is gone for good — and the balances screen re-entered after the drain will save `0` in its place, which `useMigrationReceiveConfirmation` reads as `isConfirmedWithoutWaiting` and swaps the session on. That is #4102, reintroduced by a network blip.

**Proposal.** Give the hook the third outcome. `lockError` should keep the user where they are rather than picking a direction: fold it into the hook's `loading` (or return an explicit `isBlocked` the CTA can render a retry against) so `isStaleCommitPoint` requires a *confirmed* closed answer — `isAtCommitPoint && !lockLoading && !lockError && isSettledClosed`. Independently, make the drop in `mergeCheckpoint` conditional on more than a step regression (see finding 5), so that even a wrong restart decision cannot destroy the figure. Then change the `"never skips ahead on an unanswered server"` test to assert the positive destination too, not just the absent one.

---

### 4. Three independent no-cache reads of the same lock on one screen

Not blocking.

`MigrationBlockerProvider` was built around a context precisely because "separate no-cache reads could disagree and leave a deeplinked screen operating over a locked account". `migration-gate.tsx:83` already deep-imports its own `useMigrationLock`, and `076bd3429` now adds a third instance inside `MigrationRequiredScreen` — the screen the gate itself renders at `migration-gate.tsx:367`. Each is a separate `fetchPolicy: "no-cache"` `migrationStatus` round trip, so one gate render fires three, and the CTA can act on a different answer than the chrome around it: the gate's read succeeds and renders the locked screen while the CTA's read fails and restarts the flow (finding 3).

**Proposal.** Lift the lock the way the blocker already did: expose the resolved `MigrationLock` through a provider (or extend `MigrationBlockerContext` to carry `isLocked`/`loading`/`hasError`) and have the gate, the required screen and `useMigrationNextStep` read the shared value. That removes the disagreement window and two of the three network reads at the same time. `useMigrationStatus` keeps `no-cache` for the account-switch reason it documents; sharing happens at the React layer, not Apollo's.

---

### 5. The restart signal is inferred from a step regression any background save can forge

Not blocking.

`hasRestarted` in [migration-checkpoint-storage.ts:141](app/screens/account-migration/utils/migration-checkpoint-storage.ts:141) is a proxy: it fires on *any* non-commit step written over a commit-point record, on the stated premise that "the flow never walks backwards on its own". Findings 1–3 are all cases where the flow now does exactly that. There is a fourth path that does not need a routing bug at all: `useMigrationBackupCheckpoint` saves from a plain `useEffect`, and the three screens that use it sit mounted beneath the commit screen for the rest of the run. `balances-overview-screen.tsx` explicitly gates its own save on `isFocused` with a comment about background instances re-saving; the backup hook has no such gate, so any re-run of that effect past the commit point would now wipe the figure as well as regress the step. I could not prove a re-entry trigger in the current dependency arrays, so this is hardening rather than a confirmed defect — but the blast radius grew with this commit.

**Proposal.** Make the run boundary explicit instead of inferred. Mint a `runId` when the explainer provisions or reuses an account, carry it in `CheckpointUpdate`, and drop `expectedReceiveSats` when the incoming `runId` differs from the stored one — a background re-save carries the same id and is then harmless, while a genuine restart is unambiguous. Keep the step-regression check as a belt-and-braces assertion if you like, but it should not be the sole authority. Separately, focus-gate `useMigrationBackupCheckpoint` to match the commit screen's precedent.

---

## Test Coverage Gaps

1. **The combination that reaches finding 1 is untested.** `use-migration-next-step.spec.ts` covers `hasTransactions` with a stale commit point, but only asserts the hop to `accountMigrationDownloadHistory`; and `download-history-screen.spec.tsx` mocks `navigateToCheckpoint` and asserts only that it was called. No test in either file can observe that Continue lands on the commit screen. A test that drives the real resolver with a `BalancesOverview` checkpoint and an unlocked server would fail today.

2. **The `lockError` test asserts only a negative.** `"never skips ahead on an unanswered server, only on a confirmed one"` checks `mockNavigateToCheckpoint` was not called and nothing else, so it passes identically whether the error path restarts, holds, or does nothing at all. The entry-screen spec in the same changeset pairs every negative with a positive (`expect(mockReplace).toHaveBeenCalledWith(...)`) — this one should too, and it is the assertion that would pin whichever behaviour finding 3 settles on.

3. **No test distinguishes `Completed` from `NotStarted` at the consumer level.** `use-migration-lock.spec.ts` pins that both read unlocked, but neither `use-migration-next-step.spec.ts` nor `migration-entry-screen.spec.tsx` exercises the completed-but-unswapped state, so nothing in the suite records that a migration awaiting its session swap must not be restarted.

4. **`mergeCheckpoint` has no test for a save that regresses the step *without* being a restart.** The four new cases in `migration-checkpoint-storage.spec.ts` all frame a step regression as a deliberate restart, which is the assumption under review. A case named for the hazard — a backgrounded backup screen re-saving `BackupMethod` over a live commit point — would document that the current design cannot tell the two apart.

5. **The entry-screen spec defaults `mockIsLocked = true` for every pre-existing test,** including `"renders nothing and starts the flow for a fresh migration"`, where a fresh migration would really be `NotStarted`/unlocked. The assertions happen not to depend on it, but the default encodes an impossible state and will mislead the next person adding a case. Set the default to unlocked and opt individual tests into the lock.

6. **`use-migration-lock.ts`'s own docblock is the only record of the removed expiry,** and no test covers the new "a record from a month ago still routes by its stored step" behaviour end to end — `migration-checkpoint-storage.spec.ts` covers the `accountId` half (`"lends an old record's account id to the fresh save"`) but not that the old record's *step* is still honoured by `resolveCheckpointRoute`.

---

## Code Smells

1. **`savedAt` is now write-only but still a hard validation gate.** With `isExpired` gone, nothing in the app reads the field, yet `validateStoredCheckpoint` still rejects the whole record when it is missing or non-numeric — discarding a resumable `accountId` over a field no decision consumes. Either drop it to advisory (the treatment `expectedReceiveSats` already gets, with the same rationale) or keep it and say in the type that it is diagnostic only.

2. **The docblock edit in `use-migration-lock.ts` left a 131-character line** ([use-migration-lock.ts:23](app/screens/account-migration/hooks/use-migration-lock.ts:23)) in a file where every other line sits at or under 94, against a `printWidth` of 90. Prettier does not reflow prose so lint stays green; it needs a manual rewrap.

3. **`restartDestination` is a constant dressed as a lookup.** `resolveCheckpointRoute(null).name` can only ever return `accountMigrationExplainer`, but it is recomputed every render and threaded through two `useCallback` dependency arrays. Its return type is the full `CheckpointDestination` union, so if `DEFAULT_DESTINATION` ever changed the "restart" would silently be allowed to navigate to `accountMigrationBalancesOverview` — the one destination this branch exists to avoid. Hoist it to a module constant and narrow the type to the restart route.

4. **The two entry points disagree on where a restart begins.** The entry screen restarts at `accountMigrationStart` (the gate, with its API-key warning, dollar-drain check and kill-switch); `useMigrationNextStep` restarts at `accountMigrationExplainer`, skipping all three. Each is internally justified in its own comment, but a reader cannot tell from either that "restart" means two different things. Worth naming the distinction explicitly, or routing both through the gate.

5. **Comment drift in `mergeCheckpoint`.** Two adjacent docblocks now explain the same decision — the `hasRestarted` rationale from `f1f42fc4d` and the "dropped outright" rationale from `9e47edece` — and between them they take 16 lines to describe eight lines of code. They also still frame the rule in terms of "the flow never walks backwards on its own", a premise findings 1–3 contradict. Worth collapsing into one block that states the invariant and its known exceptions.

PR #4248, opened 2026-09-06
feat(telemetry): measure enhanced-mode self-custodial activity behind…

kngako:
… a privacy boundary

Self-custodial funds never touch Blink's ledger, so activity that used to appear as ledger rows now appears as nothing at all. On every existing dashboard a custodial user migrating to Spark looks exactly like one who churned, which makes a successful product transition indistinguishable from an exodus.

This restores that visibility for Enhanced accounts only, without reintroducing the surveillance self-custody exists to eliminate. Scope is the client half of the telemetry PRD's P2 and P3: the boundary, the suppression lifecycle and the settlement events. The LNURL aggregates, the warehouse and the board are other teams' work and are untouched here.

## The boundary

app/self-custodial/measurement/ implements the pipeline the PRD specifies:

    domain event -> classifier -> privacy policy -> delivery adapter

Feature code calls the emitters in index.ts and nothing else. It never sees Firebase, never builds a payload and cannot reach the delivery adapter directly, which is what makes the policy stage unbypassable rather than merely conventional.

  contract.ts    the closed allowlist; each field is an enumeration rather than
                 a string, so a prohibited value fails to type-check
  gate.ts        the default-deny collection gate and its mode state machine
  policy.ts      drops any payload carrying an unknown field or an out-of-domain
                 value, and counts the drop locally for diagnostics
  classifier.ts  SDK settlement record -> coarse rail and direction
  event-id.ts    the random per-payment id that deduplication depends on
  delivery.ts    the Firebase adapter, behind a seam the outbox will take over

## Anonymity

Events carry no per-user identifier of any kind: no pubkey, account id, payment hash, amount, destination or Lightning Address, and nothing derived from one. A hashed pubkey is a pubkey for the purposes of this contract, so hashing was considered and rejected rather than used.

Enforcement is structural rather than conventional. Each field's value domain is an enumeration, so an allowed field cannot carry a smuggled value: rail_type accepts exactly three literals, and telemetry_event_id is shape-checked against a v4 UUID so that a 64-hex payment hash, a 66-hex compressed pubkey or a base64 digest cannot pass as one. That shape check is what stops the single string field in the payload from becoming the hole the derivation rule warns about.

The SDK payment id is held in memory solely as a Map key, so a repeated settlement callback reuses its original random id rather than minting a second one. It is never transmitted, and a test pins that.

One documented deviation: event_version is not in the PRD's approved field list but is required by the schema-versioning requirement. It is a constant, identical on every device, so it describes the schema rather than the user and leaves the linkage rule untouched. The reasoning is recorded in contract.ts.

## Suppression

Collection is off at process start and stays off until mode resolves as custodial or Enhanced. Unknown mode, unsynchronised mode, offline start, timeout and failed LNURL recovery all mean off. Failure never falls back to enabled.

A self-custodial account holding no mode resolves as unresolved, not Enhanced. This deliberately diverges from the settings row, which reads an unset mode as Enhanced so it has something to display: the contract asks for a positive resolution, and "we never asked, so we assumed consent" is not one. Such an account stays silent until useAccountModeSync recovers its real mode from the LNURL server, at which point the gate resolves for real.

Switching into incognito discards before disabling, never flushes then discards. The outbox does not exist yet, so onTelemetrySuppressed is the seam it will plug into; the ordering is the gate's to enforce rather than the outbox's to remember.

firebase.json disables analytics auto-collection natively. Without it, session_start and first_open fire before any JS runs and the gate cannot hold on a fresh install. initializeTelemetryGate() then re-asserts the closed state at module scope in app.tsx, because setAnalyticsCollectionEnabled persists across launches: a device that was Enhanced last run starts this one collecting, and the window before the mode resolves is precisely when a user who has since switched to incognito would leak.

## Events

payment_settled fires from the SDK event listener rather than from the send and receive call sites. The SDK is the only place that observes settlement itself: a send returns before it is final, and receives arrive with no call site at all, whether over the Lightning Address, a plain BOLT11 invoice or a direct Spark transfer. Direction is read off the settlement record, so both directions are counted from one place.

conversion_settled fires from the convert bridge, where the payment response is in hand. It carries direction and no volume; swap amounts are gated on a policy decision rather than an engineering task.

referral_completed is declared but has no call site. The app has no self-custodial referral surface, and the invite flow behind Circles is custodial and needs an authenticated me. The contract lives here so the event is specified when that flow is built, and so the board reads its tile as unbuilt rather than as zero.

Rails are coarse -- lightning, spark, onchain -- with no fallback rail. A record the classifier cannot vouch for is dropped, because guessing would put a real count in the wrong bucket, and mislabelling is a correctness defect of the same severity as a leak. Conversion legs are skipped on the payment path so a swap is not counted once as a swap and again as a payment.

## Deliberately not built

The persisted outbox is blocked until the transport decision is signed; its retry state machine, cleanup behaviour and dedup location all differ by option, and the decision record still carries blank signature fields. Consequence today: events created offline rely on Firebase's own buffering, and the event-id map is in memory, so a settlement re-delivered after a restart counts twice.

The 26 existing direct logEvent call sites were left in place. The collection gate makes them provably silent in incognito without touching them, which is the load-bearing half of the requirement. The residual gap: a self-custodial conversion still emits the custodial-shaped conversion_attempt and conversion_result alongside the correct conversion_settled, so an Enhanced account's swap is mode-mislabelled in the legacy events. Migrating those is the obvious follow-up.

## Verification

67 new tests, including the full suppression matrix -- cold start, restore from backup, account switching in both directions, offline start, failed LNURL recovery, and the race between mode resolution and the first event.

Every negative assertion is anchored against a positive control on the same call path, so "nothing was emitted" cannot pass merely because the call site was silently broken. Seven guards were mutation-tested: removing the gate check, regenerating the event id per callback, guessing an unknown rail as lightning, dropping the conversion-leg skip, reversing the discard/disable order, weakening the event-id shape check, and treating an unresolved mode as permitted. Each one turns tests red.

Emission is wrapped so a telemetry fault cannot escape into the flow that emitted it. The settlement listener also drives the wallet refresh, and a throw there would cost the user their balance update to save a metric. Two tests pin that, one for a throwing transport and one for an unreadable settlement record.

The shared Firebase analytics mock now returns promises and exposes the collection and user-property methods, because callers attach .catch() to them and an undefined return made a fire-and-forget log throw.

Full suite green: 724 suites, 9293 tests. tsc and eslint clean.

## Known assumption

Sends are counted from the SDK's PaymentSucceeded event on the strength of that event carrying a full Payment record with a paymentType field. That was inferred from the generated bindings rather than confirmed on a device. If the SDK only fires it for receives, sends undercount silently and the emission point needs to move to the send adapter.


esaugomez31 (changes requested):
## Blockers

1. **iOS outbox never drains** (`app/telemetry/outbox/store.ts:179`). `writeAtomically` relies on `RNFS.moveFile` overwriting an existing destination. On iOS that is `NSFileManager moveItemAtPath:toPath:`, which fails when the destination exists (Android's `renameTo` overwrites). Only the first write of a record succeeds; the `submitted` transition and the second `loss.json` / `acked.json` writes all throw. The drain starts with `markSubmitted`, so on iOS it fails on the first record every time and nothing ever leaves the outbox. The jest mock overwrites on move, so no test sees it.

2. **Custodial GA4 identity is lost for the whole session** (`app/telemetry/platform-analytics.ts:91`). `setCustodialAnalyticsIdentity` drops calls while `identityPermitted` is false and nothing replays them. The flag only flips inside the async `applyMode`, so every identity effect of `AnalyticsContainer`'s first commit is refused. `userId` (already in the persisted Apollo cache) and `galoyInstance` never change, so they never re-fire, and the gate cleared both at launch. Custodial sessions ship GA4 events without `user_id`.

## Decisions needed before merge

3. **The rollout flag also gates the four pre-existing custodial events** (`app/telemetry/mode.ts:278`). `isEventPermitted` ANDs `isTelemetryEnabled()` for every mode, and `telemetryEnabled` defaults to false, so `backup_completed`, `restore_completed`, `stable_balance_activated` and `rollout_exposed` stop everywhere until the ramp reaches a device. `rollout_exposed` carries `nonCustodialEnabled`, so the self-custody rollout's control group stops reporting. Is that acceptable during the ramp?

4. **A null server mode now leaves the account mode-less indefinitely** (`app/store/persistent-state/self-custodial-server-account-mode.ts:49`). Main wrote Enhanced on a null `/recover` answer, which pushed it and activated the Lightning Address. Now nothing is pushed, the address never activates, `/recover` is re-called on every `sdk` change, and the settings row keeps displaying Enhanced. The default-deny is documented as intended; how does a mode-less account reach a mode now, and should the settings row stop defaulting to Enhanced?

## Non-blockers

5. **Custodial-only devices now fetch `/telemetry-config`** (`app/self-custodial/providers/telemetry.tsx:179`). `mayReport` includes Custodial, so a device with no self-custodial account calls the LNURL server on activation and on every foreground. Is that intended?

6. **Duplicate reads per drain and per enqueue** (`app/telemetry/outbox/drain.ts:120`). Each drain calls `store.pending()` twice (once for the loss-report check, once for the records), and `enqueue` re-reads `acked.json` right after `sweep` loaded it. Read the queue once and have `sweep` return the tombstones it already has.

7. **`isSuppressedMode` is duplicated** (`app/self-custodial/providers/telemetry.tsx:201`). It is a copy of the private helper in `app/telemetry/mode.ts:136`. Exporting the original keeps one definition.

8. **`telemetry_enabled` parsing is duplicated** (`app/self-custodial/lnurl-telemetry-config.ts:45`). The body type and the "apply if boolean" check exist here and in `applyLnurlTelemetryFlag` in `app/self-custodial/lnurl-server-mode.ts`. When backend names the real field, both must change.

9. **Backoff never decays with time** (`app/telemetry/outbox/drain.ts:175`). A retryable always doubles from the stored `delayMs`, even if the previous window expired long ago; only an acknowledge resets it. One transient failure after a day offline waits the full 15 minute cap.

10. **Deleting an account can recreate its outbox directory** (`app/self-custodial/hooks/use-delete-account.ts:93`). The raw unlink does not bump the store generation and every store write runs `mkdir`, so a drain in flight during delete writes the directory back for an account no store will mount again. `createOutboxStore(dir).discardAll()` instead of the unlink closes it.

kngako:
Here's a PR comment for the fifth review, in its own structure. All of it is in `03b0b5875`.

---

Addressed in `03b0b5875` (on top of `d306ffb53`).

## Blockers

**1. iOS outbox never drains — fixed.** Confirmed in the react-native-fs sources: `moveFile` is `NSFileManager moveItemAtPath:toPath:` on iOS, which refuses an existing destination, while Android's `renameTo` overwrites. Two changes:

- The jest mock (`__mocks__/react-native-fs.js`) now refuses an existing destination like iOS. That alone turned 36 outbox tests red — the bug, reproduced — and means anything that passes jest from here on works on both platforms.
- `writeAtomically` removes the destination before the move, and readers cover the window so a process death inside it loses nothing: a `.tmp` whose final file is missing is read *as* the file and promoted; a `.tmp` beside a present final is a write that never finished and is removed; a torn `.tmp` counts as `parse_failed`. Tests for each, plus an explicit regression named for the platform fact (`store.ts:198`).

**2. Custodial GA4 events without `user_id` — fixed.** As described: the container's effects fire on the first commit and don't re-fire while their values stand, and permission arrives only after the mode's queued side effects. `setCustodialAnalyticsIdentity` now remembers what was asked for, and `setCustodialIdentityPermitted(true)` replays it — including after a round trip through a self-custodial account (`platform-analytics.ts:66`). Tests for the replay, the round trip, and that a grant with nothing requested touches nothing.

## Decisions

**3. Rollout flag gating the four custodial events.** Resolved by scoping: the rollout flag and the kill switch now gate the **Enhanced pipeline only** (`mode.ts:283`). P2 is what ships behind the flag and what NFR-O4 rolls back; `backup_completed`, `restore_completed`, `stable_balance_activated` and `rollout_exposed` predate it and ride GA4 exactly as on `main`, so the self-custody rollout's control group keeps reporting through the ramp. Test pins that the four flow under Custodial with the flag off and the switch engaged. Say so if the other reading was intended.

**4. Null server mode leaves the account mode-less.** The default-deny stands — AD-25, no assumed consent — and the restore flow already routes a null `/recover` answer to the choose-experience screen. What was actually wrong and is now fixed:

- The settings row read "Enhanced" for an account that behaves as nothing of the kind; it now reads **"Not set"** (untranslated, like the existing mode names) and remains the path into choosing (`account-mode.tsx:31`).
- `/recover` was re-asked on every SDK reconnect; a null answer is now remembered for the session and asked once per launch (`use-account-mode-sync.ts:38`).

Remaining product call, narrower than posed: legacy accounts that never chose get no prompt beyond that row. If they should be prompted, that's a UX flow to add.

## Non-blockers

- **5.** Only an Enhanced device fetches `/telemetry-config`; a custodial-only device has nothing the switch could roll back and no business calling the LNURL server on every foreground. Test added.
- **6.** The drain reads the queue once and re-reads only when it just filed a loss report (so the report rides the same drain — tested); `sweep` hands its tombstones to `enqueue` instead of a second read of `acked.json`.
- **7.** `isSuppressedMode` is exported from `mode.ts`; the provider's copy is gone.
- **8.** One `applyLnurlTelemetryFlag` in `lnurl-telemetry-flag.ts`, shared by `/recover`, `/mode` and `/telemetry-config`. When backend names the field, one constant changes.
- **9.** Backoff doubles only while the previous refusal is still fresh (its window hasn't been over for longer than it lasted); a refusal after a day offline is a first refusal at 5 s, not the 15 min cap. Tested.
- **10.** Already closed at `d306ffb53`: deletion goes through the store's `retire()`, which bumps the generation and serialises with any in-flight drain. This review looked at an older head.

## Verification

- `tsc`, `eslint` (app, tests, mocks, scripts), contract artifact in sync.
- Both CI shards green as the workflow runs them: 729 suites / 9,547 tests.
- 12 targeted mutations, all caught — including reverting `writeAtomically` to the old rename-over-existing, which the iOS-faithful mock now fails.
- Not verified here: an iOS build (the fix is at the RNFS call level; the mock now models the platform, but a device run of the drain on iOS is the real confirmation).

esaugomez31 (changes requested):
@kngako I went through the telemetry work carefully. Two findings I would like resolved before merge, and a set of smaller ones that can be follow-ups. I traced each one against the code rather than reading the diff alone, and three things that looked like regressions turned out to be deliberate; I listed those at the bottom so the reasoning is on the record.

## Blockers

### 1. A device with no active account stays `Unresolved` for the whole session

`app/telemetry/mode.ts:133`

The trace:

- `app/hooks/use-account-registry.tsx:130-153` builds an empty `accounts` list while unauthed with no self-custodial entry, so `activeAccount` is `undefined`.
- `toActiveAccountKind` (`app/self-custodial/providers/telemetry.tsx:216`) maps `undefined` to `ActiveAccountKind.None`.
- `deriveTelemetryMode` has no branch for `None`, so it falls through to `Unresolved` and stays there, because nothing later in a pre-auth session can resolve it.

Two consequences on every fresh install and every logged-out launch:

- Platform collection stays off, so the legacy `analytics().logEvent` calls across the acquisition funnel are silently dropped: `phone-login-validation`, `email-login-validate`, `use-create-device-account`, `telegram-auth`, `totp-login-validate`, and the screen views from `navigation-container-wrapper`.
- `DiagnosticsDisposition.Unresolved` tells the sink to *hold* errors, bounded, until an answer arrives. The answer never arrives, so signup and login crashes never reach Crashlytics.

The `Custodial` branch already carries the `|| !hasSelfCustodialAccount` escape hatch for a device that has nothing to have been rolled back from. A device with no account at all looks like it wants the same treatment, but I did not want to assume your intent here, since it is a policy call rather than a typo: what should a pre-auth session resolve to?

### 2. `initializeTelemetryGate()` runs at module scope with unguarded native calls

`app/app.tsx:74`

`initializeTelemetryGate` calls `closeSynchronously`, which calls `setPlatformCollectionEnabled` (`app/telemetry/platform-analytics.ts:40`) and `clearCustodialAnalyticsIdentity` (`:137`). Both invoke `analytics()` directly and catch only the returned promise, never the constructor. RNFB's `analytics()` throws synchronously when the native module is not linked, and at module scope that is an unhandled throw during bundle evaluation: the app does not start at all, white screen, no user recovery.

The queued half has the same root cause. If `applyMode` rejects, the module-level `applying` chain at `app/telemetry/mode.ts:186` stays rejected for the rest of the process, so every later `whenModeSettled()` rejects and `setPlatformCollectionEnabled(true)` is never reached again.

This needs a broken native linkage to fire, so it should not reach a correctly configured release build, and I want to be honest that the probability is low. I am raising it anyway because the blast radius is the whole app and the fix is a try/catch around the initialisation, which is the same defensive treatment `captureTelemetryFact` already gets for exactly this reason.

## Non-blocking

Rough priority order:

- **`app/telemetry/outbox/store.ts:500`** Every `enqueue` calls `sweep()`, which does a `readDir` plus one `readFile` per record plus the tombstone read. Near the 500-record cap, which a device offline for part of the 72h TTL will reach, that is 500+ bridge round-trips on the same tick as a settled payment, and a drain repeats the full read up to three times through `pending()` and `depth()`. An in-memory index of `{dedupKey, queuedAt, sdkPaymentId}` rebuilt once per store, or a single manifest file, would make enqueue O(1) writes instead of O(n) reads.
- **`jest.setup-after-env.ts:115`** The diagnostics gate is opened once per file at setup, but `resetDiagnosticsForTesting` closes it again through `resetTransmissibilityForTesting`. In a suite with that in a `beforeEach`, every test after the first runs against a closed gate. Tests asserting "nothing was reported" then pass for the wrong reason. Moving the open into a `beforeEach` would make the default hold for the whole file.
- **`package.json:17`** The `test` script lost `--runInBand` while `coverage` kept it. CI re-adds it on the command line, so CI is fine, but a local `yarn test` now spawns one worker per core, each carrying a full ts-jest type program that the workflow comment itself measures at roughly 2.3 GB.
- **`package.json:36`** `check:telemetry-contract` greps `git diff --name-only`, which never lists untracked files. A `CONTRACT_VERSION` bump writes a brand-new `telemetry-contract.v2.json`, the grep finds nothing, the check exits 0, and the PR merges with only the stale v1 artifact committed, so the relay allowlist is generated from the wrong version. `git status --porcelain` would catch the new-file case.
- **`app/telemetry/index.ts:187`** `drainActiveOutbox` snapshots the store into a local, but `reportLoss` routes through `captureTelemetryFact`, which re-reads the module-global `activeOutbox` at line 117. An account switch during an in-flight drain files account A's loss counters into account B's queue, or drops them via `countUnroutedEvent()`. Either way A never settles its loss and re-reports the same counts on every future drain. The window is narrow, but it corrupts the counter the FR-29 loss budget reads.
- **`app/telemetry/outbox/record.ts:60`** `dedupKeyFor` sanitises and truncates to 120 characters. Two ids sharing a normalised prefix pass the raw-id dedup check in `enqueue` and then collide on one filename, so `write()` overwrites the first record with no counter incremented, neither `deduplicated` nor `evicted`. Real SDK payment ids are far shorter, so this is theoretical today; hashing the id would close it for good.
- **`app/types/account.ts:32`** `ACCOUNT_MODE_UNSET_NAME = "Not set"` is rendered straight into the settings row beside a translated label, so a non-English user sees a mixed-language row such as "Modo: Not set". Unlike the Enhanced and Incognito brand names, this string has a reason to go through the i18n pipeline.
- **`.eslintrc.json:231` and `:267`** Two separate `overrides` entries target the identical `files` array for `platform-analytics.ts`, one setting `no-restricted-imports` and the other turning `no-restricted-syntax` off. Merging them keeps that file's rules in one place.
- **`scripts/telemetry-contract.ts:28`** `Object.fromEntries(Object.entries(row.params).map(([name, domain]) => [name, domain]))` is an identity copy of `row.params`. It reads as a normalisation step, so the next reader will go looking for one that is not there.
- **`app/utils/error-reporting.ts:212`** The `if (!__DEV__) return` guard in `crashForTesting` cannot be reached: its only call site is already inside the `{__DEV__ && (...)}` branch at `developer-screen.tsx:252`. The comment justifies the guard by the import ban, but the ban is on the crashlytics import, which that screen no longer has.

## Verified as intentional, not raising

- **`app/telemetry/contract.ts:204`** The four migrated legacy events pinned to `Custodial` under `PENDING_AD24_REVIEW` are a deliberate park, and the annotation says so. One thing worth confirming: because every call site runs with a self-custodial account active, the effect today is zero emission rather than the custodial-only emission the annotation describes. If the intent was to keep some signal alive until the privacy review lands, that is not what happens.
- **`app/store/persistent-state/self-custodial-server-account-mode.ts:49`** A null server answer settling nothing is AD-25, and the comment lays out the fail-open it removes. I withdrew my "no path back" concern after checking `AccountModeSetting`, which gives a mode-less account a visible "Not set" row and the one path into the selection screen.
- **`app/self-custodial/lnurl-telemetry-config.ts:29`** `lastFetchedAt` stamps before the fetch, so one offline failure blocks the kill switch for 15 minutes and a restored connection cannot bypass it. The file documents that the endpoint is not served yet, so the channel is inert and this is not urgent. Worth fixing before backend ships `/telemetry-config`, since a kill switch that cannot be retried promptly is the failure AD-28 exists to prevent.

kngako:
## What was addressed

**Blocker 1 — pre-account device stuck `Unresolved`.** Confirmed by the same trace. The policy call: a device with no account *and* no self-custodial account on it is the custodial product's visitor — the funnel is custodial analytics, a login-screen crash is a custodial crash, and nothing on the device could have been rolled back from — so it now resolves `Custodial`, the same escape hatch the custodial branch already takes. A device that *does* hold a self-custodial account (possibly an incognito one) stays `Unresolved` between accounts. Derivation table and provider tests for both. [[mode.ts:133](https://claude.ai/epitaxy/app/telemetry/mode.ts:133)](app/telemetry/mode.ts:133)

**Blocker 2 — unguarded native calls at module scope.** Confirmed in RNFB (`nativeModule.js:111` throws synchronously). Every SDK call in the platform seam now goes through one guard that reports a fault and carries on; the sink guards its Crashlytics calls the same way (the boundary's own fault reporter runs through it, so it had to); `initializeTelemetryGate` cannot throw; and the transition chain `.catch`es a rejected step so `whenModeSettled()` never stays rejected. Tests spy the mock's factory to throw like an unlinked module, through initialisation, resolution, and a later reopen. [[platform-analytics.ts:36](https://claude.ai/epitaxy/app/telemetry/platform-analytics.ts:36)](app/telemetry/platform-analytics.ts:36), [[mode.ts:243](https://claude.ai/epitaxy/app/telemetry/mode.ts:243)](app/telemetry/mode.ts:243)

**Non-blocking, done now:** loss report filed into the *draining* store, and only while its drain is still permitted — the test switches to a second Enhanced account mid-drain (a custodial switch was already caught by the mislabel guard, which is why my first test shape didn't discriminate); dedup filename is a `sha256` of the payment id (`js-sha256` was already a dependency); kill-switch interval starts on an answer, not an attempt, and concurrent triggers share one request; "Not set" is an i18n key across source, generated types and all 28 locales (the pipeline runs `en/index.ts → en.json`, which I learned by having my JSON edit overwritten); jest default gate re-opened in a `beforeEach`; `yarn test` in band again; `check:telemetry-contract` uses `git status --porcelain`; one ESLint override for `platform-analytics.ts`; the identity copy and the stale comment fixed.

**Follow-up, deliberately not done:** the O(n) sweep per enqueue. An in-memory index per directory is the right fix, but it touches the store's consistency model (discard, retire, leases, the parent sweep) and deserves its own change with its own tests rather than a tail on this one.

**Confirmed intentional:** the four parked events — `rollout_exposed` fires from the feature-flags context on every device, so it is emitted on custodial and (now) pre-account devices; the other three fire only inside self-custodial flows with that account already active, so they are emitted from *nowhere* until the AD-24 review admits their rows. That's the park, and the contract annotation now says so explicitly rather than implying custodial-only emission.

**Verified:** tsc, eslint, contract artifact and i18n types in sync, both shards green under `CI=true` (9,562 tests), 8 targeted mutations caught, one equivalent mutant documented.

esaugomez31 (dismissed):
LGTM

esaugomez31 (approved):

grimen (changes requested):
## Findings

**1. A device holding an incognito wallet resolves Custodial on cold start, because the "no account" rule from dd03d9785 reads the account registry before it has hydrated (blocking).**
`SelfCustodialTelemetryMount` computes `hasSelfCustodialAccount = selfCustodialEntries.length > 0` (`providers/telemetry.tsx:81`) and never looks at `useAccountRegistry().loading`. The registry seeds `selfCustodialEntries` with `[]` whenever `activeAccountId` is unset or `Custodial` (`use-account-registry.tsx:75-81`). The real list only arrives after the async `listSelfCustodialAccounts()`. Until then `deriveTelemetryMode` takes the `!hasSelfCustodialAccount` branch (`mode.ts:130`, `:145`). `closeSynchronously(Custodial)` then releases the held start-up errors to Crashlytics, writes native crash provenance `permitted`, and turns GA4 collection on, persisted natively. If the index read fails, `ReadFailed` leaves the entries at `[]` for the whole session. A probe reproduces both cases.

*Example:* a user with an incognito wallet on the device logs out of custodial. `resetState()` (`use-logout.ts:117`) leaves `activeAccountId` unset. On the next cold start the device runs as a custodial visitor until the index read lands, and the crash provenance it wrote survives the rest of the session (finding 3).

*Proposal:* one option is a `registryHydrated` input to `TelemetryModeInputs` that forces `Unresolved` until the registry has loaded, and reading `ReadFailed` as "holds a self-custodial account".

**2. A crash in an Unresolved session that follows a permitted session is uploaded on both platforms (blocking).**
`CrashCollectionPolicy.atLaunch("permitted")` returns `collect = true` for the session now starting (`CrashCollectionPolicy.kt:42-45`, mirrored in `AppDelegate.mm`). The design relies on the next launch deleting what that session recorded, but the SDK decides before the app gets a say. On iOS, Crashlytics persists the last `setCrashlyticsCollectionEnabled` as a sticky setting that overrides the plist (`FIRCLSDataCollectionArbiter.m`). `[FIRApp configure]` (`AppDelegate.mm:69`) therefore takes the enabled branch in `FIRCLSReportManager.m:323-338`, which queues the upload and says it will not "respect Send / DeleteUnsentReports". `applyCrashCollectionAtLaunch()` runs after it, at `AppDelegate.mm:83`. On Android, a fatal crash with collection on is uploaded inside the crashing process, so the next launch never sees it. The comments in `CrashCollection.kt:9-13` and `AppDelegate.mm` assume the upload waits on a settings fetch, but settings only gate the transfer, not the send-or-delete decision.

*Example:* a device holding a custodial and a self-custodial account ended its last session custodial. On relaunch remote config is untrusted, so the mode stays Unresolved all session (`mode.ts:129-131`). A crash in that session is uploaded, on Android immediately and on iOS at the next launch.

*Proposal:* one option is to keep the SDK's automatic collection off for the whole process, and at launch call `sendUnsentReports()` when provenance is `permitted` and `deleteUnsentReports()` otherwise. Both work with collection off. The cost is Android's in-process fatal upload. Either way it needs an on-device check with Crashlytics debug logging, since nothing in CI runs natively.

**3. A runtime transition into Unresolved still leaves crash collection on with provenance `permitted` (blocking).**
This is the bot's HIGH on d306ffb53, and it is still open. `applyCrashCollection` returns early for `DiagnosticsDisposition.Unresolved` (`error-reporting.ts:181-182`), and the native module only knows "permitted" and "denied-and-delete". Neither 03b0b5875 nor dd03d9785 touches this path. `error-reporting.spec.ts:395` now pins the no-op as intended. A probe reproduces it: Custodial → Unresolved produces the native call `[true]` and nothing else.

*Example:* a custodial user restores a wallet whose server mode is unknown and switches to it (`use-restore-wallet.ts:111-123`). Until a mode is chosen, crashes are collected, and the next launch reads `permitted`.

*Proposal:* one option is a third native call for Unresolved: collection off, unsent reports kept, provenance written `unresolved`, so a later Permitted can still send them.

**4. Rollback no longer happens when this launch's Remote Config fetch fails, even if the rollback was activated in an earlier session (blocking).**
`setRemoteConfigTrusted(true)` only runs inside the `try` after `fetchAndActivate()` (`feature-flags-context.tsx:551`), and nothing retries. A failed fetch leaves `remoteConfigTrusted` false for the session, so `use-rollback.ts:34` returns early and `isLockedOutOfSelfCustodial` stays false. Main bounced users on defaults, which was fail-closed by accident. This PR makes the rollback, a safety control, fail open. A probe with a rejected fetch and an activated remote `false` yields `ready=true, trusted=false` and no rollback.

*Example:* self-custody is pulled for a cohort. A device that already activated `nonCustodialEnabled=false` keeps using the self-custodial wallet in every session whose fetch fails: offline starts, throttled fetches, networks that block Firebase.

*Proposal:* one option is to trust per value rather than per fetch, reading `getValue(key).getSource() === "remote"` as trusted even in the `catch`. That keeps the fix for defaults being read as answers.

**5. SDK log lines from a non-active account are forwarded under the active account's mode.**
`initLogging` installs one process-wide listener. `logging.ts:52-60` forwards Info and Warn lines as breadcrumbs and Error lines as non-fatals, gated only by the global disposition, and the file's own comment says these lines "routinely carry payment ids and amounts". Three paths run an SDK for an account that is not active: `probeSelfCustodialAccountWallets` when deleting an inactive account, `migration-transfer-request.ts:108`, and `selfCustodialRestoreWallet`, which on a fresh install runs entirely under the pre-account Custodial from finding 1.

*Example:* Enhanced account A is active and the user deletes incognito account B. B's sync lines reach Crashlytics as breadcrumbs.

*Proposal:* one option is to forward SDK lines only while no auxiliary SDK instance is alive, through a ref-count held by probe, migration and restore.

**6. `self_custodial_restore_completed` reaches GA4 with the ledger `user_id`, because it is logged before the account switch applies.**
`activateAccount` only queues a state update, and `resolveTelemetryMode` runs later in the mount's effect. `logSelfCustodialRestoreCompleted()` runs synchronously straight after it (`use-restore-wallet.ts:94-96`, `:111-113`). `currentWalletProvider()` still says custodial, the mislabel guard passes because the label is just as stale, and `logPlatformEvent` sends the event to GA4. That contradicts `contract.ts:164-169`, which says the parked events are "emitted from nowhere".

*Example:* a custodial user re-enters the phrase of an incognito wallet already on the device. GA4 records the restore against their ledger id.

*Proposal:* one option is to emit after `whenModeSettled()` for the new account, or to drop the event unless the target account resolves Enhanced.

**7. `measurement.ts` counts swap legs as payments, and counts a USD-funded send only as a swap.**
`logPaymentSettled` recognises a swap only through `payment.conversionDetails` (`measurement.ts:124`). The app's own `hasConversion` (`mappers/transaction.ts:163`) also treats `details.inner.conversionInfo` as a conversion, which is how legs are modelled, so each leg becomes a Spark send plus a Spark receive. SDK 0.22 documents `conversionDetails` as "involved a conversion before the payment", so a Lightning send paid from USDB emits `conversion_settled` and returns before `payment_settled`. Probes reproduce both against the code path. Whether the SDK fires `PaymentSucceeded` for child legs is not verified on a device.

*Example:* a user pays a Lightning invoice from their USD balance. The board counts one swap and no send.

*Proposal:* one option is to reuse `hasConversion` for the leg skip, always emit `payment_settled` for a real send or receive, and add `conversion_settled` when the conversion completed.

**8. A mode-less account still offers its Lightning Address, and none of its crashes reach Crashlytics.**
`withSelfCustodialModeFromServer` now leaves state unchanged on a null answer, so these accounts stay mode-less. `use-lightning-address-gate.ts:13-15` withholds the address only when `isAnonMode`, so it still shows on receive and home. The PR's own comment says it "stays dormant" (`account-mode.tsx:32`). The account also derives `Unresolved` (`mode.ts:126`), so its breadcrumbs are dropped, its errors are held and never released, and crash collection stays off. On main these accounts were set to Enhanced. Now the only way out is Settings → Mode → "Not set", with no prompt.

*Example:* a restored wallet whose owner left the choose-experience screen shows an address the server may not serve, and a crash in that wallet is never reported.

*Proposal:* one option is to also gate the address on `accountMode === null` and prompt mode-less accounts once. If showing the address is deliberate, the comment in `account-mode.tsx` should say so.

## Test Coverage Gaps

- No spec runs the registry and `SelfCustodialTelemetryMount` together, so the hydration window and a `ReadFailed` index are unasserted. Finding 1 hides there.
- `error-reporting.spec.ts:395` asserts the Permitted → Unresolved no-op as correct, so finding 3 is pinned rather than caught.
- `CrashCollectionPolicyTest` covers the pure table only. Its "never resolved → deletes" row passes while the SDK ignores that delete once collection is on (finding 2). The ObjC copy of the policy has no test.
- `self-custodial/analytics.spec.ts:57-60,66,95,109` replaced `toHaveBeenCalledTimes(1)` with `paramsSentFor`, which takes the first match, so a double emission now passes.
- `measurement.ts` has no fixture for a leg payment or a USD-funded send (finding 7).
- No feature-flags or rollback spec covers a failed fetch over an already-activated rollback (finding 4).

## Code Smells

- The PR is `DIRTY` against main. The one conflict is `__tests__/screens/send-confirmation.spec.tsx`, where #4284 landed the proper version of the same flake fix. Taking main's file loses nothing.
- Unrelated changes ride along: the jest `.claude` worktree ignore, CI sharding, the AD-25 mode-less change, and the native crash-collection modules. The native modules especially would review better on their own, since they need an on-device check this PR cannot get from CI.
- `platform-analytics.ts` never clears the remembered identity on logout, because `analytics.tsx` only sets `userId` when it is truthy. A later Custodial grant replays the previous ledger id until the next `me` resolves.
- The v22 field is optional, so the schema bump turns a downgrade into `handleUnusableBlob`, which wipes `activeAccountId` and every per-account mode. Staying on v21 would keep downgrades lossless.

PR #4249, opened 2026-09-07
fix(map): give the place locator's buttons the pairing the rest of the app uses

kngako:
The bar for aiming a new place's pin drew Cancel and Next side by side, and
both were wrong. Cancel had no surface at all: a 20pt orange label sitting
straight on the map imagery, which is illegible over anything but the palest
tiles. The row split its width one part to two, which left Cancel far too
narrow for its own label, so it wrapped to "Cance / l" and collided with the
Next pill next to it. Nothing else in the app pairs a primary and a secondary
this way.

Why the surface went missing

The locator asked for one. It passed buttonStyle={{ backgroundColor:
colors.white }} to GaloySecondaryButton with a comment explaining why a
transparent button needs its own ground over a map. The atom never applied it:
it spreads the caller's props and then sets buttonStyle={styles.buttonStyle}
afterwards, so its own transparent fill wins and the caller's is dropped
without a warning. Rendering the atom and flattening its styles confirms it —
the fill comes out rgba(0, 0, 0, 0) whatever the caller passes.

That could have been fixed in the atom, by merging the incoming buttonStyle
rather than discarding it. It is deliberately not, here. The atom is shared by
some thirty screens and its transparency is the definition of the secondary
button, not an oversight; a change there is a change to every one of them. The
locator was the only caller trying to override it, and it was only trying
because it had put the button somewhere the design system does not put one.

What it looks like now

Stacked full width, primary first and the secondary beneath it. That is what
onboarding-layout, custom-modal, get-started-screen and the locator's own
sibling place-sheet all do; there is no side-by-side pair of these two
anywhere in the tree. Neither label can wrap now, and the visual weight lands
where it should, on Next.

The pair sits on the panel the map's other overlays already paint —
colors.white, radius 20, a grey4 hairline — the same treatment as the place
sheet, the search bar, the add-place pill and the status pill. The panel is
what makes the transparent secondary legible, so the atoms are both used
exactly as shipped and follow the theme into dark mode on their own. The three
style overrides the locator carried are gone with it.

The height it costs

BAR_HEIGHT is not decoration: locatorBarTop feeds it to the locate-me button
so that button stays reachable, above the bar rather than inside it, while the
pin is being aimed. Stacking takes the bar from 50pt to 122pt, and the
constant is now derived from the parts that make it up rather than restated as
a literal. The two component heights in it were read off a real render rather
than reasoned about: the primary's 50 is its own minHeight, and the
secondary's 42 is a 24pt line box plus the 1pt the library pads its title by,
inside the 8pt of vertical padding the app's Button theme sets.

The bar covering ~72pt more of the map is the real cost of this change, and it
is a judgement call rather than a free win. Consistency and legibility are
worth it; the pin and its hint sit at the map's centre and are untouched. The
narrower alternative — keeping the row and giving Cancel a white pill through
containerStyle, which does survive the atom — was left on the table.

Tests

The two assertions that pin these numbers are updated with their reasoning:
locatorBarTop's own, and the LocationButtonCopy case that offsets from it. Both
were written to catch exactly this kind of drift, so they are meant to move
deliberately. The testIDs are unchanged, so the 49 map-component tests that
drive this flow needed no edits. tsc, eslint and prettier are clean, and all
187 tests under __tests__/components/map-component/ pass.

Screenshots
----

image

image


designsats (approved):
## Review: #4249 — place locator button pairing

**Verdict: approve.** The core change is correct and fixes two pre-existing brand violations rather than just reshuffling layout. Comments below are non-blocking and I will do them in follow-up pr

### What's right

Two deletions matter more than the additions:

- **`cancelButton: { backgroundColor: colors.white }` removed.** That override painted a fill onto the secondary button. The spec is explicit that secondary is transparent — no fill, no border, it's a text button. The old code contradicted the atom; the new code lets the atom be itself and moves the surface up to the panel where it belongs.
- **`flex: 1` / `flex: 2` removed.** A 1:2 side-by-side ratio isn't a pattern anywhere in the system. Primary is specified as "fills its container width."

The stacked primary-then-secondary layout is genuinely the app's convention — verified across all 31 `GaloySecondaryButton` call sites. `backup-method-screen.tsx`, `keep-receiving-screen.tsx` and the map's own `place-sheet.tsx` all stack full-width in that order. The PR comment claiming this is accurate.

It builds from the existing atoms and adds no new component — correct call, since a primary/secondary pair is exactly what these two are specified for.

### Off-scale values introduced

The system has three radii (`8 / 16 / 999`) and seven spacings (`3 · 5 · 8 · 10 · 14 · 20 · 30`). Three new values sit outside both:

| Value | In PR | System | Note |
|---|---|---|---|
| `borderRadius: 20` | panel | `16` (`radius.lg`, cards/panels) | matches map's local overlays |
| `BAR_PADDING = 12` | panel | `10` or `14` | card padding is specified as `20` |
| `BUTTON_GAP = 4` | between buttons | `3` or `5` | — |

The `borderRadius: 20` + `colors.white` pairing has a fair defence: it's verbatim what `index.tsx:661-662` (status pill) and `index.tsx:685-686` (add-place button) already use, so the panel matches its neighbours. But the brand doc is pointed that this drift is recorded "so you recognise it in existing screens, not so you reproduce it" — and this is new code being written.

**Token note:** `colors.white` inverts in dark mode to pure `#000000` (`colors.ts:74`), not the `grey7` `#0F0F0F` specified for a static card surface. Pure-black chrome is a known open question in the brand docs so this isn't new, but the panel is a static card by every other definition and `grey7` is the token for it. `borderColor: colors.grey4` is correct — grey4 is the border role. Note the map's other two overlays carry no border at all, so the panel borrows `place-sheet.tsx:630`'s treatment rather than the pill's.

### Main concern: `BAR_HEIGHT` is hand-computed

    BAR_HEIGHT = 2 + BAR_PADDING * 2 + PRIMARY_HEIGHT + BUTTON_GAP + SECONDARY_HEIGHT

The arithmetic is currently correct — verified `PRIMARY_HEIGHT = 50` against the atom's `minHeight` (`galoy-primary-button.tsx:42`), and `SECONDARY_HEIGHT = 42` as `24` lineHeight + `2` title padding + `2 × 8` RNE `spacing.md` (confirmed `md: 8` in `@rn-vui/base/dist/helpers/index.js:23`). Total 122, so `locatorBarTop(0) = 156`.

The weakness is that the test pins the **computed constant**, not the **rendered height**. If someone changes the primary's `minHeight` or the RNE theme's spacing, the constant and its test stay `156` together and both are silently wrong — the location button would then overlap the bar with nothing failing. An `onLayout` measurement would make this self-correcting.

Not a blocker, but the comment "The atoms' own heights" understates how far this reaches into another component's private business.

### Scope

The `__DEV__` gate bypass in `index.tsx:110` is unrelated to a PR titled "give the place locator's buttons the pairing the rest of the app uses."

The reasoning is sound and `__DEV__` gating is idiomatic here (11 other uses, including `root-navigator.tsx:187`), but those all **add a dev-only affordance** — this one **bypasses a production entitlement gate**, so a debug build shows a button whose backend call will be refused for an unentitled account. The inline comment acknowledges this. Defensible, but belongs in its own commit.

### Tests

All 188 pass across the 10 map-component suites. (`act()` warnings in the output are pre-existing noise from `use-account-registry`, unrelated to this PR.)

The `__DEV__` save/restore in the spec is careful work: Jest's RN preset runs `__DEV__` true, so without pinning it false every existing gate assertion would have been silently vacuous. That's a real bug the PR caught in its own test setup.

### Requested changes, ranked

1. `borderRadius: 20` → `16`
2. `BUTTON_GAP: 4` → `5`
3. `BAR_PADDING: 12` → `14`
4. Split the `__DEV__` change into its own commit
5. Consider `colors.grey7` over `colors.white` for the panel surface
6. Optional: measure `BAR_HEIGHT` via `onLayout` instead of deriving it

PR #4254, opened 2026-09-08
refactor(map): call the add-place form a panel, since that is what it is

designsats:
Addresses https://github.com/blinkbitcoin/blink-wip/issues/1254, and the sheet half of [#1255](https://github.com/blinkbitcoin/blink-wip/issues/1255).

**Builds on #4253** and consumes two things added there, so that one merges first. GitHub will not take a base branch that lives on the fork, so this targets `main` and its diff shows #4253's commits too until #4253 lands. The commits that belong to this PR are the ones after #4253's, including `5e52425d64` and `368e6eca02` below.

### The form is a sheet now, without becoming a window

Review of the running build asked for this form to animate in and dismiss on swipe down exactly as the merchant sheet does, and to be synchronised with it. It now is: same entry, same pull handle, same drag-to-dismiss, same spring, same dismiss distance — taken from the shared primitive rather than reproduced, so the two cannot drift apart.

What it does not take is the window and the scrim. The pin is aimed by **panning the map above this form while the fields are being typed** — the hint on the map says so in as many words — and a modal window would swallow every one of those touches. `presentation="inline"` (added in #4253) is exactly that distinction: everything that makes a sheet feel like a sheet, none of what makes it take the screen.

This is also why the form keeps its layout slot rather than floating over the map. The map's own comment is explicit that the coordinate submitted is the centre of the map *view*, whatever height it currently has — so a sheet that covered the map instead of shortening it would put the pin's centre behind the sheet and submit the wrong place.

### The category is chips again

Restored at review's request: they worked better.

They were replaced by a dropdown in #4249 for a real reason — fourteen chips wrapped taller than the rest of the form put together and pushed the send button off the bottom. **That is fixed at the cause rather than reverted around:** the button is now held below the scroll by the sheet's `footer`, so the chips can be as tall as they need and take the scroll with them instead of taking the button with them.

Reading fourteen short, familiar words is faster than opening a list of the same fourteen, which hides thirteen behind a tap and shows nothing that was not already visible.

**Scheme copied from `percentage-selector`** — the transfer screen's % buttons — which is where the app already draws a one-of-many choice as chips:

| | Value |
|---|---|
| Ground | `colors.grey5` |
| Selected fill | `colors.primary` |
| Label | `colors.primary`, bold |
| Selected label | `colors.white` |
| Radius | `100` |
| Padding | `8` × `16` |

Copied rather than shared, per direction — a chip primitive is its own ticket. Two things deliberately do not come across: its row (four chips share a row evenly; fourteen have to wrap) and its `minWidth: 64`, which would pad the short labels into ragged columns.

Tapping the selected chip clears it. A list did not need that — there is no empty row to pick in a set of chips.

### Two steps, and a new title

Added on review of the running build.

The form asked both its questions at once and sent on one button. It now asks **where the place is and what it is called** first, on **Continue**, and **what kind of place it is** second, on **Submit request**.

> When the user has selected the location and name of the business to add to the map
> And clicks "Continue"
> Then the contents of the sheet change to category selection

Splitting them puts each question on the surface that suits it. The location and the name belong with the map — the pin is aimed by panning it while they are typed, and both are on screen together. The category is a fourteen-way choice that wants the sheet to itself: the chips were always the tallest thing on this form, and now they are the only thing on their step rather than something the rest of the form has to make room for.

**Title is now "Suggest business"**, not "Add a place". Nothing here puts a pin on the map — it goes to BTC Map, whose community decides. The old title promised an outcome this form does not have, which is the same thing the review note beneath it exists to walk back.

**Stepping back is a first-class way through**, not just an escape: the header grows an arrow on the second step, and nothing is cleared on the way. Coming back to fix a typo in the name would be a poor trade for the category already chosen. The first step has no arrow, having nothing behind it.

#### `sentLocation` removed with the split

It froze the coordinate row while a send was in flight, so the row could not follow the map away from the place being sent. That row is on the first step and the send is on the second, so there is no longer anything on screen for it to freeze. **What gets submitted is unchanged** — the request has always carried the pin as it stood when the button was tapped.

A failure is now drawn only on the step that sends, and stepping back off that step takes it down: it described a send made from a form the user has since left.

#### i18n

Both strings are translated into **all 28 locales**, not left carrying English. `submitPlaceRequest` had to exist everywhere for CI key parity — Crowdin does not backfill a key it has never seen — and `addPlaceTitle` had changed in English only, so the other 27 were still offering to "add a place", the outcome the new title exists to stop promising.

The wording follows what each locale already uses rather than being translated in isolation. The noun is taken from the merchant term each one had settled on (`payMerchant`, `CardFlow.merchant`) so the map keeps one word for the thing it is about — comercio in es, commerce in fr, Geschäft in de, işletme in tr, biashara in sw, ishishini in xh, 店舗 in ja. The verb reuses the locale's own `common.submit` where it fits, which is why sw sends with *Wasilisha* and xh with *Thumela*.

**Worth a native reader before release: `lg` and `qu`.** Both had the thinnest existing material to follow, and neither has a settled word for a business as a *place* rather than a *trader*. Every other locale is built on a term already shipping in the app.

Not fixed here: `ar`, `qu`, `sr`, `th` and `vi` carry an untranslated `common.submit` already. That predates this PR and is not its to fix — the two keys added here are translated in all five regardless.

`i18n-types.ts` is unchanged, since no key was added or removed. Values only: exactly two lines per file, 56 in and 56 out across the 28. `yarn check:translations` is clean and the 155 locale-parity tests pass.

### CTA spacing

The send button now sits `20` off the foot of the sheet rather than `12`, per the app's standing gap, and stays pinned there while the content scrolls.

### Naming

`add-place-sheet.tsx` → `add-place-panel.tsx`. It was named for a surface it did not have, and two tickets were filed against it asking for the scrim and drag-to-dismiss its siblings had. It now has the drag-to-dismiss; it still has no scrim, and still is not a window, so it is still not a sheet in the sense the other two are.

### Left behind

`MapScreen.placeCategoryHint` is now unused — it was the dropdown's placeholder. Removing it means all 28 locales plus a regenerated `i18n-types`, which is worth its own commit rather than riding along here.

### A sheet that doesn't expand

Added for Send 3.0's error message sheet (blink-wip#1275), which will be this primitive's first caller outside the map. That sheet rests on its header, and the header is all it holds. Dragging it up opened an empty scroll under its call to action.

`expandable={false}` (default `true`) holds a `restsOnHeader` sheet at rest. A drag up stops at the resting offset, and a release never springs it to full height. Dragging down still dismisses it. The place sheet and category filter don't pass it, so they behave as before. Inline sheets don't accept it, the same way they don't accept `restsOnHeader`.

### System bars: the bottom inset counted once

The modal presentation now passes `statusBarTranslucent` and `navigationBarTranslucent`. Without them, the `Modal` window on Android stops above the navigation bar, but callers pad by the root `insets.bottom` (the app's convention for a bottom CTA), so the bar was counted twice. Measured on the Medium Phone emulator: an error sheet's CTA sat 69dp above the screen edge, where screen CTAs sit at 44.6dp. With the flags it lands on the same pixel row. The place and category sheets pad by `insets.bottom` the same way, so they should drop by the same amount (not measured separately). The scrim now also covers both bars.

## Acceptance requirements

- [x] The form animates in and dismisses on swipe down, the same way the merchant sheet does
- [x] That behaviour is shared with the merchant sheet, not reimplemented, so the two stay synchronised
- [x] The map above stays pannable while the form is open, and the submitted coordinate stays the map view's centre
- [x] Category is chips again, dropdown gone
- [x] Chips use the `percentage-selector` scheme
- [x] CTA sits 20 off the bottom and stays reachable when content overflows
- [x] Component named for what it is
- [x] First step takes location and name, CTA reads "Continue"
- [x] Continue reveals category selection, CTA reads "Submit request"
- [x] Title reads "Suggest business"
- [x] Both new strings translated in all 28 locales, not shipped as English placeholders
- [x] A shared chip primitive — **out of scope**, its own ticket per direction

## Testing specs

- **Category chips** (`add-place-panel.spec.tsx`, 21 tests, 6 new): only one chip reads as selected after a swap; tapping the selected chip takes the category back off and disables the send with it. The catch-all `other` is still not offered, now asserted against the chips rather than an opened list.
- **Regression**: the remaining 15 assertions on this form — submission, in-flight guarding, error handling, pin-follows-map — pass against the reworked shell with only the category interaction updated from two taps to one.
- **Sheet behaviour** is covered in #4253 on the primitive itself: inline puts up no scrim, still drags away past the threshold, takes no height of its own, and draws nothing when hidden.
- **Two-step flow** (6 of those new): neither an empty name nor one of only spaces gets past the first step; the category survives a step back; the first step offers no way back; the pin is not on the step that sends; each button says which step it is on.
- **i18n**: all 28 locales translated for both keys; `yarn check:translations` clean; 155 locale-parity tests pass across 8 suites; `i18n-types.ts` unchanged.
- **Non-expandable sheet** (`bottom-sheet.spec.tsx`, 2 new, 27 in the suite): a hard throw up past halfway leaves it at its resting offset with the list still locked, and fails when the change is reverted. A drag down still dismisses it. `yarn jest --maxWorkers=2 __tests__/components/bottom-sheet/bottom-sheet.spec.tsx`
- **System bars** (`bottom-sheet.spec.tsx`, 1 new, 28 in the suite): the modal is translucent under both bars; fails when the flags are removed. Map suites: 235 tests pass across 11 suites.
- **Manual, both platforms, still to do**: keyboard open with the chips scrolled; drag-to-dismiss while the map is mid-pan; system font size at max, where fourteen chips wrap furthest.

Local: `tsc`, `eslint` and `prettier` clean. 207 tests pass across the 11 suites.

## Screenshots


kngako (dismissed):
The "blocking" finding seems like reasonable UX for me. 

I'll let Andre use his discretition on this one. But from my test it's good to go.

---

## Findings

### 1. A flick that only scrolls the list dismisses the sheet — [BLOCKING]

`onUpdate` correctly refuses to move the sheet while the list is scrolling back up ([index.tsx:242](app/components/bottom-sheet/index.tsx:242)), but `onEnd` never learns that the drag was consumed: it projects `offset.value + velocityY * 0.15` and dismisses on anything past `restOffset + 80` ([index.tsx:248](app/components/bottom-sheet/index.tsx:248)). Concretely: a user reads to the bottom of the category filter's fourteen rows, flicks down to get back to the top, and the sheet closes under them — the flick's release velocity alone (1,500–4,000 px/s is ordinary) projects to 225–600px while `offset` is still 0. The slow version is worse: drag down on a list scrolled to 300, and at the frame `scrollOffset` reaches 0 the guard stops firing and `offset` jumps straight to the whole accumulated `translationY`, teleporting the sheet down and past the dismiss threshold in one frame. This came over verbatim from `place-sheet`, but the branch now also puts it under the filter list (14 rows in a 60% sheet) and the add-place chips.

**Proposal:** Give the gesture a `suppressed` shared value: set it to `event.translationY` on every frame the guard returns, reset it to 0 in `onBegin`, and subtract it in the live update — `offset.value = Math.max(0, dragStart.value + event.translationY - suppressed.value)`. That alone fixes the teleport and makes the continued drag start from where the finger actually was when the list ran out. For the flick, add a second shared flag set the first time the guard fires and cleared once the sheet actually moves, and in `onEnd` snap back to `restOffset` without consulting `velocityY` while it is still set — a gesture the scroll view ate should not be able to decide the sheet's fate.

### 2. The Map tab's bar keeps the theme it had when add-place last closed — [BLOCKING]

The hide/restore effect early-returns whenever `isAddingPlace` is false ([index.tsx:371](app/components/map-component/index.tsx:371)), so the last `setOptions({ tabBarStyle })` the Map screen ever receives is the cleanup's, carrying whatever `useBottomTabBarStyle()` returned at that instant. `setOptions` beats `screenOptions` and persists for the life of the screen, so: a user adds a place in light mode, closes the panel, then switches to dark in Settings — every other tab's bar repaints from the navigator's fresh `screenOptions`, and the Map tab's stays white with a light `grey4` hairline. The effect *does* re-run when `tabBarStyle` changes, so it has the new value in hand and throws it away at the guard; `GaloyThemeProvider` sits above `NavigationContainer` and the navigator is not keyed on the mode, so nothing remounts the screen to clear the override.

**Proposal:** Drop the early return and let the effect own the option unconditionally: `navigation.setOptions({ tabBarStyle: isAddingPlace ? HIDDEN : tabBarStyle })`, with `HIDDEN` hoisted to a module constant so the object identity is stable. That keeps the restore in sync with the theme for free, and removes the cleanup entirely — which also removes the `setOptions`-after-unmount the current shape does when the screen is torn down mid-add. Worth pulling `{ display: "none" }` out of the render path either way; a fresh object each run makes the effect's own dependency reasoning harder to read than it needs to be.

### 3. Android's back button abandons the add-place panel instead of closing it — [NON-BLOCKING]

The modal presentation hands `onRequestClose` to RN's `Modal` ([index.tsx:344](app/components/bottom-sheet/index.tsx:344)), which is what makes the hardware back close the place sheet and the filter. The inline branch returns the bare sheet ([index.tsx:341](app/components/bottom-sheet/index.tsx:341)) and registers nothing, so on Android a user halfway through typing a shop name presses back — the reflex for "get out of this form", and now the only visible exit besides the X, since finding the tab bar gone is the whole point of commit `f621c20ca` — and the navigator leaves the Map tab, unmounting the panel and throwing away the name, the category and the minted `submissionId` with no prompt.

**Proposal:** Register a `BackHandler.addEventListener("hardwareBackPress", …)` in the inline branch that calls `onClose` and returns `true` while `isVisible`, mirroring what `onRequestClose` gives the modal branch — that keeps the two presentations honest about differing "only in the window and the scrim". Put it in `BottomSheet` rather than in `AddPlacePanel` so the next inline caller inherits it. It also wants to be the same `onClose` the X uses, so the panel closes rather than the tab changing under it.

### 4. The second step sends a pin the panel stopped showing on the first — [NON-BLOCKING]

The coordinate row lives on the details step and the send button lives on the category step ([add-place-panel.tsx:188](app/components/map-component/add-place-panel.tsx:188)), while `submission` is rebuilt from the live `location` on every render ([add-place-panel.tsx:96](app/components/map-component/add-place-panel.tsx:96)) — and the map above stays pannable throughout, which is the reason this is a panel at all. So a user reads "13.496743, −89.439462", taps Continue, brushes the map while reaching for a chip, and submits a different coordinate than the one they checked, with nothing on the panel disagreeing. The old single-step form always had the row beside the button, and even froze it during the round trip; what keeps this from being blocking is that the pin itself is still on the map, so the drift is visible if the user happens to look up.

**Proposal:** Keep a one-line coordinate readout in the header on both steps — it is 20 characters and the header already has a `space-between` row that would carry it under the title. If the design will not take it, freeze `location` into state when Continue is pressed and submit the frozen value, with the back arrow releasing it; that makes Continue mean "this is the place", which is what the two-step split implies to the user anyway. Either way the retry path deserves a second look: a retry re-sends under the same `submissionIdRef`, so a moved pin on retry is a different place arriving under an idempotency key the backend will dedupe against.

### 5. Closing from the X skips the exit the drag gets — [NON-BLOCKING]

Inline, `isVisible` false unmounts the sheet in the same commit that starts the `withTiming` slide-out ([index.tsx:341](app/components/bottom-sheet/index.tsx:341) vs [index.tsx:201](app/components/bottom-sheet/index.tsx:201)), and `AddPlacePanel` passes `onClose` straight through to the X and hard-codes `isVisible` — so `stopAddingPlace` unmounts it immediately. The result is two different exits from the same panel: drag it away and it slides out (the parent only unmounts after the timing callback fires), tap the X and it vanishes while the map snaps back to full height. The class's own doc promises the two presentations give up "only the window and the scrim".

**Proposal:** Hold the sheet mounted through its own exit: keep an `isLeaving` state that the `!isVisible` effect sets, clear it in the `withTiming` callback via `runOnJS`, and render inline while `isVisible || isLeaving`. That makes both exits identical and costs one state flag. If that is more machinery than the panel is worth, the cheaper honest fix is to say in the prop doc that inline is unmount-on-hide and that callers wanting the slide-out should drive it from the drag.

### 6. The shared sheet quietly restyles both surfaces it replaced — [NON-BLOCKING]

`e9fd0c8a5` reads as a pure "give the map's two sheets one sheet to share", but the shared surface is not either of the two it replaced: the corner radius drops 20 → 16, the handle goes from 40×4 at radius 2 to 26×3 at radius 999, and the filter sheet — which had no border at all — picks up the place sheet's 1px `grey4` outline ([index.tsx:379-400](app/components/bottom-sheet/index.tsx:379)). A designer reviewing the branch by its commit subjects has no reason to open the style block, and the map's two most-used surfaces ship a different silhouette than the one that was signed off.

**Proposal:** Decide the three values deliberately and say so in the commit body, or restore 20 / 40×4 and make the border opt-in so the refactor is genuinely behaviour-preserving. If the new values are the intent, they are worth a line in the component doc explaining that this is now the app's one bottom-sheet silhouette — the existing comments explain every other number in that file, and these three are the ones a future caller is most likely to want to override.

### 7. The seam fix is applied only to the modal presentation — [NON-BLOCKING]

`BOTTOM_OVERHANG` exists because Android antialiases the straight bottom edge of the sheet's rounded outer path and leaves a half-lit hairline, and it is applied in the modal branch only ([index.tsx:295](app/components/bottom-sheet/index.tsx:295)); `sheetInline` is a bare `flex: 1` ([index.tsx:390](app/components/bottom-sheet/index.tsx:390)). The inline panel sits on the screen's bottom edge with the same rounded shape, the same `borderBottomWidth: 0`, and — since `f621c20ca` took the tab bar away — nothing beneath it, so the same hairline should appear along the bottom of the add-place panel on Android. It reads as a seam against the root background rather than against a scrim, which is why it has not been noticed.

**Proposal:** Add `marginBottom: -BOTTOM_OVERHANG` to `sheetInline`; a flex child with a negative bottom margin overflows its slot by that much, which is the same trick the modal branch plays with an explicit height. Worth confirming on an Android device at a non-integer density before committing to it, since this is exactly the class of bug that only shows on hardware. If it does not reproduce inline, the constant's comment should say why the modal case is different — it currently reads as a property of the sheet rather than of the presentation.

---

## Test Coverage Gaps

1. **The scroll-versus-drag guard is dead code under test.** Every run logs `[Reanimated] animatedRef is not initialized in useScrollOffset`, so `scrollOffset.value` is 0 in every one of them and the `scrollOffset.value > 0` branch at [index.tsx:242](app/components/bottom-sheet/index.tsx:242) is never entered — Finding 1 is invisible to the suite as written. Stub the offset (mock `useScrollViewOffset` to return a shared value the test can set) and add two cases: a flick with a non-zero `velocityY` while the list is scrolled must *not* close the sheet, and a drag continued past the list's top must move the sheet by the post-guard travel rather than the whole translation.

2. **`VELOCITY_PROJECTION` is never exercised.** All four `fireGestureHandler` sequences in the suite — three in `bottom-sheet.spec`, one in `category-filter-sheet.spec` — pass `velocityY: 0`, so the projection could be deleted and the suite would stay green — including the `toFull = projected < restOffset / 2` snap-up decision, which is the only thing that flips `isExpanded` and therefore the only thing that unlocks the scroll view. A short drag with a high upward velocity that must expand, and a short drag with a high downward velocity that must dismiss, would pin both halves.

3. **"takes the last failure off when the place itself is edited" cannot see the `goBack` half it asserts.** The final two lines at [add-place-panel.spec.tsx:300](__tests__/components/map-component/add-place-panel.spec.tsx:300) press back and expect `place-submission-error` to be gone, but the error is rendered only when `step === "category"` — deleting `setError(null)` from `goBack` keeps all 21 tests passing (verified by mutation). Assert it by returning to the category step and checking the error is still absent, which is the behaviour the comment actually describes.

4. **"locks the list until the sheet is dragged up" only tests the lock.** [place-sheet.spec.tsx:303](__tests__/components/map-component/place-sheet.spec.tsx:303) asserts `scrollEnabled` is false at rest and stops there; nothing drags up, so the `restsOnHeader ? isExpanded : true` flip and the `setExpanded` call that drives it are untested. The test name promises the second half.

5. **The "a growing header must not drop an expanded sheet" guard has no test.** It is the subtlest condition in the component ([index.tsx:207](app/components/bottom-sheet/index.tsx:207)) and the one most likely to be lost in a future edit: expand the sheet by drag, then fire a taller `layout` on the header and assert `translateY` stays 0. `place-sheet` hits this every time a `requires-app-card` lands after the details resolve.

6. **`heightRatio` is asserted nowhere.** No test checks that a `0.6` sheet is 60% of the window, so the `Math.round(windowHeight * (heightRatio ?? 1))` line and the `?? 1` default are both unpinned — `bottom-sheet.spec` reads the height back off the rendered style rather than against an expectation.

7. **The tab-bar effect is tested only on the happy in/out path.** Neither of the two new `map-component` cases covers unmounting while `isAddingPlace` is true (the cleanup still fires `setOptions` on a torn-down screen) nor a `tabBarStyle` identity change while the panel is open or after it closed — which is Finding 2. A test that rerenders with a different style and asserts the last `setOptions` carries the new one would pin the fix.

---

## Code Smells

1. **`add-place-panel.spec` still mocks `react-native-modal` for a dropdown that no longer exists.** [Lines 10–16](__tests__/components/map-component/add-place-panel.spec.tsx:10) carry a four-line comment about "the category dropdown puts its list in a react-native-modal"; the dropdown became chips in `086267634`. The mock is inert, and the comment now misdescribes the component under test.

2. **`getByText: jest.fn() as never` is passed as a query.** [Lines 127 and 138](__tests__/components/map-component/add-place-panel.spec.tsx:127) fake half the `Queries` object to satisfy a helper that does not use it. Narrow `continueToCategories` to `Pick<Queries, "getByTestId">` — it only ever touches `getByTestId`.

3. **`place-sheet.spec` re-declares what `bottom-sheet` now exports.** `const BOTTOM_OVERHANG = 1` and `animatedHeightOf` are duplicated in both specs, with the place-sheet copy carrying a comment pointing at "the component" that no longer holds it. Import the exported constant and lift the helper into a shared test util.

4. **`heightRatio` and `restsOnHeader` are silently inert under `presentation="inline"`.** Both are documented on the modal path and neither is rejected or warned about inline; `restsOnHeader` in particular would compute a rest offset against a flex-sized sheet and leave a gap in the layout. A discriminated union on `presentation` would make the two shapes unmisusable, which is worth doing before a third caller arrives.

5. **`header !== undefined` renders an empty padded box for `header={null}`.** [Line 307](app/components/bottom-sheet/index.tsx:307) — `null` is a legitimate "no header" from a conditional expression, and it currently produces a `View` carrying the caller's `headerStyle` padding. `Boolean(header)` reads the same and behaves the way callers expect; the same applies to `footer` on line 334.

6. **`accessibilityLiveRegion="polite"` is Android-only.** The submission error at [add-place-panel.tsx:162](app/components/map-component/add-place-panel.tsx:162) is never announced to VoiceOver, so an iOS screen-reader user taps Submit and hears nothing at all — the button just stays where it was. `AccessibilityInfo.announceForAccessibility` on the transition to a non-null `error` covers both platforms.

7. **`CategoryFilterSheet`'s `onClose` is an inline arrow.** `onClose={() => setFilterOpen(false)}` at [index.tsx:650](app/components/map-component/index.tsx:650) is a new identity every `MapComponent` render, and `onClose` is in the `pan` memo's dependency list — so the whole `Gesture.Pan()` is rebuilt on every map settle and every places refresh, which gesture-handler asks callers to avoid. `useCallback` it alongside `stopAddingPlace` and `closeSheet`, which already are.

8. **`useStyles({ bottomInset })` defeats `makeStyles`' memo in all three sheets.** `makeStyles` memoizes on `[props, theme]` by identity, and every caller passes a fresh object literal, so `StyleSheet.create` re-runs on every render. Pre-existing, and cheap — but it is now in three files that render on every map settle, and a `useMemo` on the props object would settle it.

9. **The filter spec's reanimated mock now spreads the real module to stub one hook.** The comment at [category-filter-sheet.spec.tsx:14](__tests__/components/map-component/category-filter-sheet.spec.tsx:14) explains the change well, but the underlying reason is that `Switch` still writes `useAnimatedStyle` without explicit dependencies. Giving `Switch` the explicit-deps treatment the sheet already uses would let both specs drop the mock entirely.

kngako (dismissed):
**All seven findings from the first pass are genuinely addressed**, and the four new tests I asked for are real rather than decorative — I reverted each fix in turn and watched the matching test fail. The two blocking issues are closed.

One new issue arrived with the fix for finding 5: adding `leave` to the visibility effect's dependency list couples the sheet's animation to the parent's render cadence, and for the one caller with an unstable `onClose` — the category filter — a re-render landing inside the 200ms slide-out cancels the dismissal outright, so the sheet springs back open and `onClose` is never called. I reproduced it and pinned the cause. Three smaller new items follow it.


Claude considers this a blocking finding but I think it's nit picking. @designsats can apply his own discretion on whether he things these issues should be fixed.

---

### 1. A parent re-render during the slide-out swallows the dismissal — [nit-pick blocking]

`leave` joined the visibility effect's dependency list ([index.tsx:273](app/components/bottom-sheet/index.tsx:273)), and `leave` is rebuilt whenever `onClose` changes identity. `CategoryFilterSheet` is passed `onClose={() => setFilterOpen(false)}` ([map-component/index.tsx:658](app/components/map-component/index.tsx:658)) — a fresh arrow on every `MapComponent` render — so any re-render inside the 200ms slide-out re-runs the effect, which for a non-resting sheet unconditionally fires `offset.value = withSpring(restingOffset, SPRING)`. That cancels the in-flight `withTiming`, its completion callback receives `finished: false`, and the `runOnJS(onClose)()` behind it never fires: the user drags the category filter away, a places refresh or a region settle lands, and the sheet springs back open as though the drag never happened.

I reproduced this against the current head — a `BottomSheet` given a new `onClose` arrow and re-rendered once after the release ends with `onClose` called 0 times, while the identical test with a stable `onClose` passes. Removing `leave` from the dependency array alone makes the repro pass, which pins the cause.

**Proposal:** Stop letting the caller's render cadence reach the animation: hold `onClose` in a ref that is updated each render and have `leave` read `onCloseRef.current`, which makes `leave` stable and takes `onClose` out of both this effect's dependencies and the `pan` memo's — the gesture is currently rebuilt on every parent render for the same reason. Wrap the filter sheet's `onClose` in `useCallback` as well, alongside `stopAddingPlace` and `closeSheet` which already are; that fixes today's caller, but the ref is what stops the next one re-introducing it. Worth also guarding the effect so it does not re-spring a sheet that is already leaving — `if (isLeaving) return` — since nothing else in that body has anything to say while the sheet is on its way out.

---


### 2. The map still offers a pin — and the instruction to move it — that the second step ignores — [NON-BLOCKING]

Freezing the location at Continue is the right call, but nothing told the map: `{isAddingPlace && <PlaceLocator />}` ([map-component/index.tsx:597](app/components/map-component/index.tsx:597)) has no idea which step the panel is on, and `PlaceLocator` draws both the movable pin and the hint *"Move the map to put the pin on the place"*. So on the category step the user reads an instruction that is no longer true, pans the map, watches the pin move to a new door, taps Submit — and the place is filed at the coordinate they left behind on step 1, with no way to tell from the screen. The previous behaviour submitted the drifted pin silently; this one ignores it silently, which is safer but no less confusing.

**Proposal:** Lift `step` out of the panel (or hand the panel a callback) so `MapComponent` can stop drawing `PlaceLocator` — or draw it in a settled, non-instructional state — once Continue has been pressed. If the pin should stay visible as a reminder of where the place is, the honest version is a fixed marker at `pinnedLocation` rather than a crosshair at the map's centre, since the two stop being the same thing at that point. Cheapest partial fix if the map is not to be touched: put the frozen coordinate line back on the category step, so at least the panel says which place is about to be sent.

### 3. A successful submit is the one exit that still does not slide out — [NON-BLOCKING]

`c243bf499` routed the X, the back button and the drag through `leave()`, but the success path never goes near it: `handlePlaceSubmit` calls `setAddingPlace(false)` the moment the mutation resolves, and the panel is behind `{isAddingPlace && <AddPlacePanel …/>}`, so it is unmounted from the outside before it can ask. Concretely: a user finishes the form, taps Submit, and on success the whole panel vanishes in one frame while the map snaps back to full height — the exact behaviour the commit set out to remove, on the one exit where the user most wants to see the form leave deliberately.

**Proposal:** Let the panel drive this exit like the others — on a `null` reason from `onSubmit`, call `setOpen(false)` and let `BottomSheet` slide out and then call `onClose`, with `handlePlaceSubmit` no longer flipping `isAddingPlace` itself on success. The map still needs its own `setAddingPlace(false)` for the abandoned-attempt case (`addSessionRef.current !== attempt`), where there is no panel on screen to animate. The success toast can stay where it is; it already fires independently of the panel.

### 4. Back during the slide-out falls through to the navigator — [NON-BLOCKING]

The back handler is registered only while `isVisible` ([index.tsx:284](app/components/bottom-sheet/index.tsx:284)), but the sheet stays on screen for `CLOSE_DURATION_MS` after `isVisible` goes false — that is the whole point of `isLeaving`. So in the 200ms after the X is tapped, the panel is still visibly on screen and a back press is no longer being caught: it reaches the navigator and leaves the Map tab. It is a narrow window and the panel was on its way out anyway, so the cost is landing on a different tab rather than losing work.

**Proposal:** Register the handler on `isVisible || isLeaving` so the sheet keeps answering back for as long as it is drawn, and have it no-op (return `true` without calling `leave` again) while already leaving — that also stops a second back press restarting the slide-out from the top. One condition, and it makes "the sheet owns back for exactly as long as it is on screen" true rather than approximately true.

---

## Test Coverage Gaps

1. **Nothing exercises a parent re-render during any of the sheet's animations.** Every test renders once and drives the sheet from there, so the whole class that finding 1 belongs to — an effect re-running mid-animation and cancelling it — is invisible. A test that re-renders with a fresh `onClose` between the release and the assertion, asserting `onClose` still lands, would pin the fix and is close to the repro I used.

2. **The one caller with an unstable `onClose` cannot see its own sheet move.** `category-filter-sheet.spec` stubs `useAnimatedStyle: () => ({})` ([category-filter-sheet.spec.tsx:22](__tests__/components/map-component/category-filter-sheet.spec.tsx:22)) so `Switch` will mount, which also blanks the sheet's transform — so no test at that level can observe position, and the sheet most exposed to finding 1 is the least testable. Stubbing the `Switch`'s hook rather than the module-wide export (or giving `Switch` explicit dependencies, per the standing smell) would give the filter sheet the same observability the other two have.

3. **The success exit is untested from either side.** `add-place-panel.spec` asserts `onSubmit` is called and the error clears, and `map-component.spec` asserts the panel unmounts, but nothing asserts *how* it leaves — which is why finding 3 slipped through a commit specifically about exits. Whichever way finding 3 is resolved, the assertion is that a successful submit produces the same slide-out the X does.

4. **`isLeaving` has no test for a sheet re-shown during its own exit.** If `isVisible` returns to true inside the 200ms, the pending `withTiming` is cancelled, its callback gets `finished: false`, and `setLeaving(false)` never runs — leaving the flag stuck true until the next real close. Unreachable for today's callers, and cheap to pin now rather than after someone adds a sheet that can be re-opened quickly.

5. **The back-handler test unmounts rather than hiding.** *"gives the back button up once it is no longer visible"* ([bottom-sheet.spec.tsx:448](__tests__/components/bottom-sheet/bottom-sheet.spec.tsx:448)) proves the unmount cleanup, not the `!isVisible` branch the name describes — and that branch is exactly what finding 4 is about. Re-render with `isVisible: false` and assert `remove` was called.

---

## Code Smells

### New

1. **The scroll-offset stub is coupled to how many times the implementation reads it.** `mockScrollOffset` is a getter over a consume-once queue ([bottom-sheet.spec.tsx:31](__tests__/components/bottom-sheet/bottom-sheet.spec.tsx:31)), so `scrollReads(300, 0)` means "the first read, then everything after". Reordering the operands of `dragStart.value === 0 && scrollOffset.value > 0 && …` — a refactor with no behavioural meaning — changes how many reads happen per frame and silently re-aims these tests. A settable value the test writes between `act` blocks would say what it means and survive the guard being rewritten.

2. **`CLOSE_DURATION_MS` is re-declared in the spec** with the comment *"Matches the sheet's own, which it does not export"*, while `BOTTOM_OVERHANG` and `PAN_TEST_ID` are exported from the component for exactly this reason. Export it too and delete the copy; a duration the tests must match is a duration that will drift.

3. **`settle()` burns real time.** `CLOSE_DURATION_MS + 100` of wall clock per call, twice, is most of why `bottom-sheet.spec` is now the slowest suite in the map set at 9.3s against 6s for its neighbours. Reanimated's jest helpers or fake timers would give the same guarantee without the wait.

4. **`useHardwareBackGuard` already exists** at [use-hardware-back-guard.ts](app/screens/account-migration/hooks/use-hardware-back-guard.ts) and is the same effect, wrapped in `useFocusEffect` so it releases back when the screen is not focused. The sheet hand-rolls it instead. Either reuse it (moving it somewhere less migration-specific), or note in the sheet why the focus-scoped version was not wanted — a second copy of a back handler is how two screens end up both claiming back.

5. **`isLeaving` is dead state for the modal presentation.** Only the inline return reads it, but `leave()` sets it for both, so every modal close now costs an extra render on a component that is about to be hidden. Setting it only when `isInline` would make the flag mean what its comment says.

### Carried over from the first pass, still open

6. **`header !== undefined` renders an empty padded box for `header={null}`** ([index.tsx:410](app/components/bottom-sheet/index.tsx:410)); same for `footer` on line 437. `Boolean(…)` reads identically and behaves the way callers expect.

7. **`accessibilityLiveRegion="polite"` is Android-only** ([add-place-panel.tsx:183](app/components/map-component/add-place-panel.tsx:183)), so a submission failure is never announced to VoiceOver — the button simply stays put. `AccessibilityInfo.announceForAccessibility` on the transition to a non-null `error` covers both.

8. **`CategoryFilterSheet`'s `onClose` is still an inline arrow** — now the trigger for finding 1 rather than just a wasted `Gesture.Pan()` rebuild, so this one has been promoted from hygiene to a real fix.

9. **`useStyles({ bottomInset })` still defeats `makeStyles`' memo** in all three sheets: it memoizes on `[props, theme]` by identity and every caller passes a fresh literal, so `StyleSheet.create` re-runs every render.

10. **`category-filter-sheet.spec`'s module-wide reanimated mock** is still there because `Switch` writes `useAnimatedStyle` without explicit dependencies. Fixing `Switch` would let the spec drop the mock and close gap 2 above at the same time.

kngako (dismissed):
All blocking findings have been addressed. 

@designsats can choose to address the remaining findings in this PR or an a future/stacked PR.


---

**All four findings from the second pass are closed**, including the blocking one, and closed the way that was asked for: `onClose` is read through a ref so `leave` and the pan no longer follow the caller's render cadence, `isDismissing` keeps the visibility effect out of an exit in progress, back is held for `isVisible || isLeaving`, a sent place slides out through the same `setOpen(false)` the X uses, and the crosshair is swapped for a real `Marker` at the fixed coordinate once Continue has taken it. The `renderedVisible` render-phase update in `25378b0a3` also quietly fixes something the second pass missed: the effect-only `setLeaving(true)` left one committed frame with the inline sheet unmounted — a blink of the map at full height, and a fresh `ScrollView`/`TextInput` subtree on the way back — before the exit played.

One new non-blocking finding, in the seam between the gesture and the new exit state. No blocking issues remain.

---

## What was fixed

| Second-pass finding | Commit | Verified by |
| --- | --- | --- |
| 1. Parent re-render swallows the dismissal — *blocking* | `25378b0a3` | `onCloseRef` + `reportClose` take `onClose` out of `leave`'s and the pan's deps; `closeFilter` is now a `useCallback`. The repro from the second pass passes and is in the suite |
| 2. Step 2 still offered a movable pin and its hint | `776ff0e18` | `onPin` lifts the fixed coordinate to the map, which draws `PinnedPlaceMarker` in place of `PlaceLocator`; `map-component.spec` asserts the swap both ways and the hint's absence |
| 3. Success was the exit that did not slide out | `48ca6dfb0` | The panel calls `setOpen(false)` on a `null` reason and the map no longer unmounts it; reverting that fails *"slides out once the place has gone…"* and *"does not send the place again while it slides out"* |
| 4. Back fell through during the slide-out | `25378b0a3` | Handler held on `isVisible \|\| isLeaving`, no-op while `isDismissing`; both cases tested |

Second-pass test gaps 1, 3, 4 and 5 are closed. Gap 2 (the filter spec cannot observe its own sheet's position) is still open. Second-pass smells 2, 4, 5, 6, 7 and 8 are closed — `CLOSE_DURATION_MS` exported, `useHardwareBackGuard` not reused *with the reason written down*, `isLeaving` inline-only, `Boolean(header)`, `announceForAccessibility`, `closeFilter`. Smells 1, 3, 9 and 10 remain and are restated at the end.

---

## Findings

### 1. Grabbed during its own slide-out and pulled back up, the inline sheet is stranded — [NON-BLOCKING]

`isDismissing` guards the visibility effect and the back handler, but not the pan: `onBegin` ([index.tsx:358](app/components/bottom-sheet/index.tsx:358)) takes `dragStart` from wherever the exit animation has got to, and the first `onUpdate` writes `offset.value` directly, which cancels the `withTiming` — its callback sees `finished: false`, so `reportClose` never fires. A user who taps the X and, inside the 200ms, catches the panel and flicks it back up ends with the sheet drawn fully open while `isVisible` is false, `isLeaving` is true and `isDismissing` is true. From there the X is inert (`isOpen` is already false, so `setOpen(false)` is not a transition), back is swallowed by the `isDismissing` check, and a successful Submit cannot close it either for the same reason the X cannot — the panel's only way out is another drag down. I reproduced this against the head: after X + an upward flick the sheet sits at `translateY` 0 with `onClose` called 0 times.

**Proposal:** Make the pan inert for the whole of an exit — `if (isDismissing.value) return` at the top of `onBegin`, `onUpdate` and `onEnd`, which is the same sentence the effect already says about itself ("on its way out, nothing has anything to say"). If catching a departing sheet is meant to be allowed, the alternative is for `onEnd` to notice `isDismissing` was set when the drag began and resume the exit regardless of where the finger let go — but a 200ms slide-out is not really a thing the user is meant to be able to argue with, and the inert version is three lines.

---

## Test Coverage Gaps

1. **The `isDismissing` guard in the visibility effect is untested — the suite stays green without it.** Removing `if (isDismissing.value) return` at [index.tsx:298](app/components/bottom-sheet/index.tsx:298) fails nothing, because the re-render test is now caught one layer earlier by `onCloseRef`. The guard still earns its place: a `restsOnHeader` sheet whose header re-lays out during a drag-dismiss — details landing on the place sheet — re-runs the effect through `headerBottom`, and without the guard the spring cancels the exit and `onClose` never fires. I confirmed that with a scratch test (fails without the guard, passes with it); it is worth adding as written: rest on the header, drag past the dismiss distance, fire a taller `layout` on the header before `settle()`, assert `onClose` was called once.

2. **Nothing sends a gesture at a sheet that is already leaving.** Finding 1 is the case; the same scaffolding that drives `dragAway()` can drive an upward flick after `rerender(isVisible: false)` and assert the sheet still reports — or, if the pan is made inert, that `translateYOf` keeps heading for `sheetHeight`.

3. **The pinned marker's coordinate is never asserted.** *"fixes the pin to the place once the form has taken its location"* checks that `pinned-place-pin` appears and `place-pin` goes, but passes the region centre to `onPin` and never reads the marker's `coordinate` back — so a marker drawn at the map centre rather than at what `onPin` received would pass. One `expect(getByTestId("pinned-place-pin").props.coordinate).toEqual(…)` with a coordinate that is *not* the centre closes it.

4. **`isSent = !reason` has no test for the empty string.** The `onSubmit` contract is `string | null`, and `""` is a string that this line reads as success. A test resolving `""` and asserting the panel stays (or a tightened check — see smell 2) would settle which is intended.

5. **Carried over: the filter sheet still cannot observe its own position.** `category-filter-sheet.spec` blanks `useAnimatedStyle` module-wide for `Switch`'s sake, so the one caller that provoked the second pass's blocking finding is still the one whose spec could not have caught it.

---

## Code Smells

### New

1. **`pinnedLocation` lives in two places.** The panel keeps its own ([add-place-panel.tsx:98](app/components/map-component/add-place-panel.tsx:98)) to decide what is *sent*, and tells the map through `onPin` to keep a second copy ([map-component/index.tsx:148](app/components/map-component/index.tsx:148)) to decide what is *drawn* — and `onPin` is optional ([add-place-panel.tsx:40](app/components/map-component/add-place-panel.tsx:40)), so a caller that omits it gets exactly the drawn-here-sent-there mismatch `776ff0e18` was written to remove. Either make `onPin` required, or lift the state to the map and pass `location={pinnedLocation ?? center}` down so the panel has one input and no copy.

2. **`isSent = !reason` treats `""` as a sent place** ([add-place-panel.tsx:152](app/components/map-component/add-place-panel.tsx:152)). Nothing returns an empty string today, but the type allows it and the consequence is the panel sliding out on a failure. `reason === null` says what is meant.

3. **`tappable={false}` is inert on both providers this app ships** ([pinned-place-marker.tsx:43](app/components/map-component/pinned-place-marker.tsx:43)). react-native-maps documents it as "iOS: Google Maps only; Android: not supported", and the map runs Apple Maps on iOS. It is harmless — the marker has no `onPress` and the map has `moveOnMarkerPress={false}` — but it reads as a guarantee it is not.

4. **`renderedVisible` and `wasVisible` both track the same transition** — one as state for the render phase ([index.tsx:209](app/components/bottom-sheet/index.tsx:209)), one as a ref for the effect ([index.tsx:216](app/components/bottom-sheet/index.tsx:216)). Both are needed as written, since the effect wants the *committed* previous value, but a reader has to hold two "did `isVisible` just change" mechanisms; a comment on `wasVisible` saying why it is not `renderedVisible` would save the next person the derivation.

5. **`settle()` is now six real 300ms waits.** That is 1.8s of pure clock in a suite that runs 25 tests in 15s, against `place-sheet.spec`'s 32 in 13s on the same machine and run — modest today, but it grows by 300ms per test that touches an exit, and every finding this pass has been about exits. Fake timers, or reanimated's jest helpers, would give the same guarantee without the wait.

### Carried over, still open

6. **The scroll-offset stub is coupled to read count** (second pass, smell 1): `scrollReads(300, 0)` is "first read, then everything after", and reordering the guard's `&&` operands would silently re-aim the two tests that use it.

7. **`useStyles({ bottomInset })` still defeats `makeStyles`' memo** in all three sheets.

8. **`category-filter-sheet.spec`'s module-wide reanimated mock** is still what stands between the filter sheet and gap 5 above; fixing `Switch`'s dependencies closes both.

PR #4257, opened 2026-09-08
fix(spark): pay fixed BOLT11 invoices with fractional satoshi amounts

### LINKED ISSUE #4250: bug: Self-custodial fee quote fails for fractional-satoshi BOLT11 invoices
## Problem

A fixed-amount BOLT11 invoice whose amount is not a whole number of satoshis cannot be prepared from a Blink self-custodial account.

In Blink Mobile 3.0.40:

1. Paying from a custodial Blink account finds a route and displays a fee.
2. Paying the same invoice from a self-custodial account fails during fee calculation. The confirmation screen shows “The payment could not be completed” and the fee slider remains unavailable.

The reported invoice amount is `1,267,644 msat` (`1,267.644 sat`).

## Root cause

Blink Mobile 3.0.40 pins `@breeztech/breez-sdk-spark-react-native` to `0.22.0`.

In Breez Spark 0.22.0, fixed BOLT11 amounts are converted from millisatoshis using floor division:

```rust
.map(|msat| u128::from(msat).saturating_div(1000))
```

This turns `1,267,644 msat` into `1,267 sat`. The lower-level fee estimator rounds the invoice requirement up to `1,268 sat` and rejects the supplied amount because it is below the invoice amount.

Upstream fixed this by using ceiling division. The first stable release containing the fix is `0.23.0`:

- https://github.com/breez/spark-sdk/commit/a5b204508b7a0656d90696699605d09ce2a3fc2b
- https://github.com/breez/spark-sdk/blob/0.22.0/crates/breez-sdk/core/src/sdk/payments/prepare/bolt11.rs#L137-L141

## Expected behavior

Self-custodial accounts should calculate a fee and pay valid fixed-amount BOLT11 invoices containing fractional-satoshi amounts.

## Acceptance criteria

- Upgrade Breez SDK Spark to a release containing the ceiling-rounding fix (`>=0.23.0`), or apply an equivalent verified fix.
- A fixed invoice for `1,267,644 msat` reaches the confirmation state with a valid fee quote.
- The payment sends successfully when a route is available.
- Whole-satoshi and amountless BOLT11 invoice behavior remains unchanged.
- Add regression coverage for fixed invoice amounts with a non-zero millisatoshi remainder.

## Environment

- Blink Mobile: `3.0.40`
- Breez SDK Spark React Native: `0.22.0`
- Account type: self-custodial
- Network: Bitcoin mainnet / Lightning

## Evidence

**Custodial account fee quote**

Custodial account fee quote

**Self-custodial account failure**

Self-custodial account failure

**Decoded invoice details**

Decoded invoice details 1

Decoded invoice details 2

## Source

Requested by `@kemal` in the Blink Slack workspace. The exact Slack permalink was unavailable to the filing agent; this issue includes the complete supplied report and screenshots.

esaugomez31:
Closes #4250.

## Do not merge before the LNURL server is adapted

0.23 changes how the SDK talks to our own Lightning address server (`blink-lnurl-server`, serving `blink.sv`), and its client speaks only the new form: there is no option to keep the old one. Three things moved:

- availability is now `POST /lnurlpay/{pubkey}/available`, signed, instead of the unsigned `GET /lnurlpay/available/{username}`
- every call signs a `breez-lnurl:v2\n<route>\n<domain>\n…` envelope instead of `{message}-{timestamp}`
- the metadata signature travels in `X-Breez-*` headers instead of the query string

Our server serves only the old route and verifies only the old signature. Probed on 2026-09-17: the new route answers 404 on both `staging.blink.sv` and `blink.sv`, and a v2 signature answers 400 "invalid signature". So on 0.23.1 nobody can check or claim a Lightning address, and a fresh install or a restore never recovers the address the wallet already owns. The app's own `/mode` and `/recover` calls are unaffected: they are hand-signed in the legacy format, which the server still accepts.

The server change is additive and backwards compatible, and it is what Breez already did upstream: serve the new route alongside the old one, verify both signature formats, and read the metadata signature from either place. This PR should merge only once that ships, and after one run of register, availability and restore against staging on a device.

## The problem

A fixed-amount BOLT11 invoice whose amount is not a whole number of satoshis cannot be paid from a self-custodial account. The confirmation screen shows "The payment could not be completed" and the fee never appears, so the flow dead-ends. The same invoice pays fine from a custodial account, which does not go through the Spark SDK.

Amounts like this come from anything that prices in fiat: the conversion from cents lands between two satoshis, for example 1,267,644 msat, which is 1,267.644 sat.

## Why it happens

Breez Spark 0.22.0 converts a fixed BOLT11 amount from millisatoshis with floor division, so 1,267,644 msat becomes 1,267 sat. The fee estimator then rounds the same invoice requirement up, to 1,268 sat, and rejects the prepared amount for being below what the invoice asks. The two halves of the same flow round in opposite directions.

## The fix

Upgrade `@breeztech/breez-sdk-spark-react-native` from 0.22.0 to 0.23.1. Breez 0.23.0 rounds up when preparing as well, so both halves agree. The upstream change is "Round up invoice amount in millisat to sat" (breez/spark-sdk#1056).

## Also in here

0.23 makes `receiverIdentityPublicKey` a required field when building a BOLT11 receive request. It is passed as `undefined` in the two places that create an invoice: our wallets always receive for themselves, and an invoice naming another receiver is a feature we do not use.

A fixed invoice's amount is now rounded up where the destination is wrapped, because that is what the SDK debits. The screen used to round it to the nearest sat instead, so an invoice with a remainder of 1 to 499 msat showed one sat less than left the wallet (1,001 msat showed "1 SAT" against 2 debited). Self-custodial only; the custodial path is untouched.

The `currentActivity` patch on the SDK is deleted. 0.23.0 carries the same fix upstream ("fix(react-native): resolve currentActivity on RN 0.80+"), and the installed package already reads `reactContext.currentActivity` in all three call sites, so keeping the patch would duplicate it.

## Testing

TypeScript is clean and 189 suites (2,526 tests) covering self-custodial, the send flow, payment destinations, payment details, hooks and the vendored patches pass.

### Test coverage

The rounding itself lives in the SDK's Rust, which no Jest test can run. What the app owns is everything around it, and each part is now pinned:

- **The send path, end to end.** `self-custodial-fixed-invoice.integration.spec.ts` runs the real invoice parser, destination wrapper, payment details, fee quote and send, with only the SDK mocked. The invoices are signed inside the test with a synthetic key, at 1,267,644, 1,001, 999, 500 and 1 msat, with 1,268,000 msat as the whole-satoshi control. For each one the amount resolves to the whole sats the SDK debits, the formatted string the screen shows is checked against that same figure through the real `useDisplayCurrency`, and both the fee quote and the send hand the SDK the invoice verbatim with no amount of the app's own. It also covers a USD wallet paying the fractional invoice, and a refused quote: the confirmation screen gets a classified error, a non-fatal is recorded, and nothing is sent.
- **Each decision point on its own.** `wrap-destination.spec.ts` rounds a fractional amount up to the sats the SDK debits and treats an invoice under one satoshi as fixed rather than amountless. `lightning.spec.ts` never gives a fixed invoice an amount of its own, including below one satoshi and from a USD wallet. `bridge/send.spec.ts` pins the exact request `prepareSend` builds.
- **The SDK the fix depends on.** `breez-sdk-spark-upstream-fixes.spec.ts` fails if the installed SDK drops below 0.23.0, if a local Breez patch comes back, or if the Android passkey module reads `currentActivity` other than through `reactContext`.
- **The new receive field.** The receive and migration specs now compare the request with `toStrictEqual`, which, unlike `toEqual`, fails when `receiverIdentityPublicKey` names a receiver other than this wallet.

Each of these was checked against the bug it guards: reintroducing the regression in the source (rounding the invoice amount to the nearest sat, overriding it with an amount of the app's own, treating sub-satoshi invoices as amountless, naming another receiver, leaving a Breez patch in place) makes the matching tests fail.

### On a device

Verified on an Android device against a self-custodial mainnet account holding 1,623 sats. Getting there took a detour worth recording: nothing in reach issues an invoice with a millisatoshi remainder. Blink rounds in the app, the point of sale, the cash register and its LNURL endpoint, which rejects millisatoshi amounts outright, and so do Alby, Stacker News, coinos and LNbits. `moneyer.dev` does not: its LNURL callback takes the amount in millisatoshis and verifies the decoded invoice matches, so asking it for 1,267,644 msat returns `lnbc12676440p...`, exactly the shape the issue describes.

Four runs, same account, same balance, same emulator:

| Invoice | Amount | 0.22.0 | 0.23.1 |
|---|---|---|---|
| `lnbc12676440p` (1,267.644 sat) | 1,268 SAT | fee quote fails | fee $0.01 (7 SAT) |
| `lnbc12680n` (1,268 sat exactly) | 1,268 SAT | fee $0.01 (7 SAT) | not re-run |

The whole-satoshi invoice is the control, and it matters: it comes from the same service, the same node and the same route, and asks for the same 1,268 sats. It quotes a fee on 0.22.0 while the fractional one does not, which rules out routing, liquidity and balance and leaves only the millisatoshi remainder.

A real payment then went through on 0.23.1: 115 SAT ($0.09) against a 114,587 msat invoice, fee 4 SAT.

### Screenshots

The same fractional invoice, before and after the upgrade:

| Before, 0.22.0 | After, 0.23.1 |
|---|---|
| The fee quote fails on 0.22.0 | The fee quote succeeds on 0.23.1 |
| The fee quote fails | The fee quote succeeds |

The control on 0.22.0, and the payment going through on 0.23.1:

| Control, 0.22.0 | Payment, 0.23.1 |
|---|---|
| Whole satoshis quote a fee on 0.22.0 | The payment completes on 0.23.1 |
| Whole satoshis, same amount, quotes a fee | 115 SAT sent, fee 4 SAT |


grimen (changes requested):
I see no test coverage for the change?

kngako (changes requested):
## Findings

### Finding 1 — The bump changes the LNURL-server wire protocol under Blink's own server, with no evidence the server accepts it [BLOCKING — until verified on staging]

0.22's client asked `GET /lnurlpay/available/{username}` unsigned and signed `{username}-{timestamp}` on register (`crates/breez-sdk/core/src/lnurl.rs` @ 0.22.0:183, :240, :150). 0.23.1's client (`fa4d8c887`, "Scope LNURL signed messages to their route and domain") sends `POST /lnurlpay/{pubkey}/available` and signs a `breez-lnurl:v2\n<route>\n<domain>\n…` envelope for every call — available, register, unregister, recover, metadata (`lnurl.rs` @ 0.23.1:264–440; format pinned as "the protocol… a wire break rather than a test to update" in `lnurl-models/src/signed_message.rs:186`). The app does not talk to Breez's server: `MAINNET_LNURL_DOMAIN = "blink.sv"` (`app/self-custodial/config.ts:37–38`), a fork with its own `/lnurlpay/{pubkey}/mode` route (`app/self-custodial/lnurl-server-mode.ts:50`).

**When it comes up:** A new self-custodial user finishes their backup and enters a username; `useRegisterLightningAddress` calls `checkLightningAddressAvailable` (`app/self-custodial/hooks/use-register-lightning-address.ts:56`). Against a server built for 0.22 the new route answers 404, the hook's bare `catch` maps it to `SetUsernameError.UNKNOWN_ERROR` (`:64–65`), and **no user can claim a Lightning address** — with a generic error that says nothing about why. If the server has instead already moved to the v2 `lnurl` crate, the app's own hand-signed `recover` (`lnurl-server-mode.ts:76–85` signs `${pubkey}-${timestamp}`, the 0.22 format) now depends on that fork keeping a pre-v2 verifier on `/recover`; restore survives (`lifecycle.ts:175–184` catches and reports), but every restored wallet lands with `isServerModeKnown: false` and a crashlytics non-fatal.

**Proposal:** Before merge, run register → availability → restore-recover on a 0.23.1 build against `staging.blink.sv`, and write the server version/commit it was verified against into the upgrade commit's body (the way #3765 records its send matrix). Add "LNURL-server wire compatibility" to the Self-custodial section of `docs/pr-review.md` so every SDK bump repeats the check — upstream's own header says the format changes are wire breaks. For the app's hand-signed calls: the SDK's `signMessage` now refuses the `breez-lnurl:` namespace (`crates/breez-sdk/core/src/sdk/api.rs:200–208`), so `recoverLnurlServerMode` cannot be moved to v2 by hand; either the server's legacy verifier for `/recover` and `/mode` is kept deliberately and documented next to `lnurl-server-mode.ts:76`, or those two routes get a Blink-owned message format that is not tied to the Breez version.

### Finding 2 — The screen rounds a fractional-sat invoice to nearest while the SDK debits the ceiling, and the new tests pin the fractional value as the intended display [NON-BLOCKING — small fix]

`buildLightningDetail` hands the parsed `millisatoshis / 1000` through unrounded (`app/self-custodial/payment-details/wrap-destination.ts:58–59`; blink-client `parsing/index.js:886`), it becomes `destinationSpecifiedAmount` and, for a BTC wallet, `settlementAmount` (`lightning.ts:43–47, :110`), and `formatMoneyAmount` renders SAT with `fractionDigits: 0` (`app/hooks/use-display-currency.ts:204–207, :272–274`), i.e. round-half-up. The SDK quotes and pays `div_ceil(1000)`. The branch's headline vector, 1 267 644 msat, lands on 1 268 either way — which is why the tests didn't notice.

**When it comes up:** A user pays a 1 001 msat invoice (fiat-priced merchants and LNURL services quote in msat); the confirmation screen shows "1 SAT" and Total "1 + fee" while 2 + fee leave the wallet. A 1 msat invoice shows "0 SAT". Any remainder of 1–499 msat is off by one. The balance gate is unaffected (`send-bitcoin-confirmation-screen.tsx:397–416`: ceil(x + fee) = ceil(x) + fee against an integer balance), so this is display-only and ≤ 1 sat — but `lightning.spec.ts:317–330` ("keeps the millisatoshi remainder on the amounts the screen shows") and `wrap-destination.spec.ts:137–173` ("unrounded") now assert it is the intended behaviour.

**Proposal:** Make the wrap boundary produce what the SDK will debit — `Math.ceil(destination.amount)` at `wrap-destination.ts:58–59` (self-custodial only; it is the SDK's rule, and the same value 0.23's `validate_request` accepts as a matching explicit amount). The more durable option is to surface `prepared.amount` from the fee quote — the SDK's authoritative "amount to be sent, in satoshis" (`PrepareSendPaymentResponse.amount`), which the onchain adapter already returns as `recipientAmount` (`app/self-custodial/adapters/payment.ts:64`) — and let the confirmation screen prefer it. Either way, flip the four "unrounded / keeps the remainder" expectations to the ceiled value and add 1 001 msat (shows 2) and 1 msat (shows 1) as vectors, since 1 267 644 cannot tell the two roundings apart. Note the custodial path shows the same fractional value (`payment-destination/lightning.ts:49`); whether Galoy also ceils is worth one question before the two modes are made to differ.

---

## Test Coverage Gaps

1. **No evidence for the protocol change (Finding 1).** Nothing on the branch — test, doc, or commit body — touches `checkLightningAddressAvailable` / `registerLightningAddress` / `recoverLnurlServerMode` against the server they now speak to differently. The bridge tests are pass-throughs and cannot see it; this is a staging run plus a written result, not a unit test.
2. **The refusal path is tested with a stand-in that can only classify as `Generic`.** `self-custodial-fixed-invoice.integration.spec.ts:30` stubs `SdkError.instanceOf: () => false` and `:239–241` rejects with a plain `Error`, so `classifySdkError` (`app/self-custodial/sdk-error.ts:70`) short-circuits. The SDK's real refusal is `SdkError::InvalidInput("Requested amount (5 sats) does not match invoice amount (1 sats)")` → `sc_invalid_input` (`:29`). Shape the rejection as `{ tag: "InvalidInput", inner: ["Requested amount…"] }` under the shared mock's `instanceOf` and pin the code the screen would actually show.
3. **The display layer is never driven with a fractional `MoneyAmount`.** The integration spec stops at the `PaymentDetail`; `formatMoneyAmount` and the confirmation screen's Amount/Total lines are untested for `1.001` / `0.001` sats. One `formatMoneyAmount` unit case per vector, or a confirmation-screen render with a fractional settlement amount, would have surfaced Finding 2.
4. **Upgrade-in-place is not described.** A wallet created on 0.22.0 opened on 0.23.1 now gets `brz_*` tree tables created in the existing storage DB and a persisted leaf tree instead of one rebuilt from the operators on each connect (`9b360e875`). Additive by construction, but it belongs on the manual matrix for this release alongside send/receive from a pre-existing wallet.
5. **The upstream-fixes spec cannot fail for the reason its name gives.** `breez-sdk-spark-upstream-fixes.spec.ts:33–37` "installs a release that rounds… up" asserts `version ≥ 0.23.0`; the rounding is only observable in Rust. Fine as a pin, but see Smell 4 — cite the commit so the next bump can re-verify it the way this review did.

## Code Smells

1. **Three commits, zero bodies.** The upgrade commit silently inherits a new rejection (`#1057`: an explicit amount that disagrees with a fixed invoice is now `InvalidInput`), a new `Config` default (`exitChainAutoFetchEnabled: true`, background exit-chain collection on every wallet — the checklist's "don't rely on SDK defaults silently", `docs/pr-review.md:349–350`), a new opt-in (`maxInstantDepositClaimFeeBps`, left unset — the safe value), a wire-protocol change (Finding 1), and drops a patch. The 0.22 branch had a commit and a body per concern (`39647d5e4`); this one has a subject line.
2. **Parallel jest hides every BigInt assertion failure in these specs.** On Jest 29.7 a failing `toStrictEqual` whose diff contains `5000n` / `1268n` (`receive.spec.ts:95–101`, `migration-transfer-request.spec.ts:123`, the integration spec's `lightningQuote`) cannot be serialised by `jest-worker`, so under `--maxWorkers=N` the suite reports "Test suite failed to run — Do not know how to serialize a BigInt" and the real diff is gone (reproduced with M4: 112/112 "passed", suite missing). CI's `--runInBand` is unaffected; `--workerThreads` fixes it locally. Worth a line in `docs/dev.md`, or `toNumber()` on the BigInt fields in expectations.
3. **A 25th inline SDK mock.** `self-custodial-fixed-invoice.integration.spec.ts:11–45` re-declares ~35 lines of `__mocks__/@breeztech/breez-sdk-spark-react-native.js` and adds the two entries the shared manual mock lacks (`SendPaymentOptions.Bolt11Invoice`, `SendPaymentMethod_Tags`). Adding those two to the shared mock lets this spec — and the next send-path spec — drop the factory; the 0.22 branch already paid for this pattern twice (`af635dfe1`, `ee8f1eb71`).
4. **The version pin has no provenance.** `FIRST_VERSION_ROUNDING_INVOICE_MILLISATS_UP = "0.23.0"` (`breez-sdk-spark-upstream-fixes.spec.ts:13`) is right (`a5b204508`, PR #1056, in `0.23.0` = `eb8821f7a`), but nothing in the file says how it was established. A comment citing the upstream commit and `prepare/bolt11.rs::invoice_amount_sats` turns the next bump's re-check into a diff read instead of a re-derivation.
5. **`syncSelfCustodialWallet` is exported and unused, and just got heavier.** `app/self-custodial/bridge/wallet.ts:57–58` has no caller; under 0.23.1 the public `sync_wallet` also awaits `collect_exit_chains` (`crates/breez-sdk/core/src/sdk/sync.rs:663–668`), which needs the operators reachable. Delete it, or document that it is not the SDK's background sync before someone wires it to a pull-to-refresh.
6. **Commit `3b6bd86c5` is authored by `Esaú Gómez)`** — a stray `)` in `user.name`, so `git shortlog` will list the contributor twice. `git commit --amend --reset-author` on that commit (or a squash-merge) fixes it.
7. **The "receiver is always us" invariant lives in two places.** `receive.ts:29–31` carries the comment; `migration-transfer-request.ts:181` gets the same `receiverIdentityPublicKey: undefined` with none, and the two five-field `Bolt11Invoice` constructions now drift independently. A `bolt11ReceiveMethod({ description, amountSats, expirySecs })` in the bridge would keep the field — and the reason — in one place.

esaugomez31:
Applied in 8dd3f795a..a8a590b07.

- **Finding 1, LNURL wire change:** confirmed. Probed: the new `POST /lnurlpay/{pubkey}/available` answers 404 on staging and production, and a v2 signature answers 400. It also breaks the SDK's own recover, so a fresh install or restore shows no Lightning address, and the metadata sync fails. Server work in `blink-lnurl-server`, now stated in the description. Your other half is right: `/mode` and `/recover` are unaffected, hand-signed in the legacy format, and cannot move to v2 since `signMessage` refuses that namespace.
- **Finding 2, display rounding:** fixed. Measured one sat short for remainders of 1 to 499 msat. `wrap-destination` now rounds up, self-custodial only; 1,499, 1,001 and 1 msat added as vectors, since 1,267,644 cannot tell the two roundings apart.
- **Gap 1:** same as Finding 1; the staging run waits on the server fix.
- **Gap 2:** the refusal is now a real `SdkError` under the shared mock's `instanceOf`, pinning `sc_invalid_input` instead of `Generic`.
- **Gap 3:** the integration spec drives the real `useDisplayCurrency` and asserts the formatted string per vector. That is what catches Finding 2.
- **Gap 4:** the device runs went 0.22.0 then 0.23.1 on the same wallet. Added: the storage migration is one-way, a 0.22 build fails with `DatabaseTooFarAhead`.
- **Gap 5 and Smell 4:** the version test is renamed to what it asserts and cites upstream commit a5b204508 and `prepare/bolt11.rs::invoice_amount_sats`.
- **Smell 1:** the inherited SDK changes are documented in the description; commits stay single-line by convention.
- **Smell 2:** unchanged. CI runs jest with `--runInBand`, so it is local-only and predates this branch.
- **Smell 3:** the two missing entries moved to the shared manual mock; the spec's own factory is gone.
- **Smell 5:** `syncSelfCustodialWallet` deleted.
- **Smell 6:** real. The squash merge fixes the author, or a rebase if you prefer.
- **Smell 7:** both invoice builders go through `bolt11ReceiveMethod` in the bridge.

This must not merge until `blink-lnurl-server` serves the new route and verifies v2 signatures alongside the old ones, deployed to staging.

kngako (approved):
I'll approve and leave it to @esaugomez31 to coordinate the changes needed on the blink-lnurl-server side for this to be supported.

---

## Findings

### Finding 1 (carried, confirmed) — Blink's LNURL server does not speak 0.23's wire protocol; the app side is ready, the server is not [BLOCKING — on `blink-lnurl-server` shipping v2 support]

Round 1 established from the Rust that 0.23.1 changes every LNURL-server call the SDK makes — `checkLightningAddressAvailable` moves from an unsigned `GET /lnurlpay/available/{username}` to a signed `POST /lnurlpay/{pubkey}/available`, and register/unregister/recover/metadata sign a `breez-lnurl:v2` envelope (`crates/breez-sdk/core/src/lnurl.rs` @ 0.23.1:264–440). **Confirmed by probe (reported 2026-09-17):** against both staging and production, the new `/available` route answers 404 and a v2-signed request answers 400. Commit `49e52414b` documents the constraint in code (`lnurl-server-mode.ts:104–109`) and in the SDK-bump checklist (`docs/pr-review.md:351–354`); the PR description now states the server dependency.

**When it comes up:** Three places, not one. (1) Registration: a new self-custodial user enters a username after backup; `useRegisterLightningAddress` (`app/self-custodial/hooks/use-register-lightning-address.ts:56–65`) gets the 404, and the user sees `UNKNOWN_ERROR` with no way to claim an address. (2) Every fresh install and restore: the SDK's own `get_lightning_address()` falls through to a v2-signed `recover` whenever its cache is empty (`crates/breez-sdk/core/src/sdk/lightning_address.rs` @ 0.23.1:101–107 → `lnurl.rs:300–318`), the server answers 400, and the app's `getLightningAddress(sdk).catch(() => undefined)` (`app/self-custodial/hooks/use-self-custodial-account-info.ts:42`) shows the account as having no Lightning address — an existing address is not lost on the server, but the phone cannot see it. (3) Metadata: `sync_lnurl_metadata` signs its `list_metadata` call the same way, so it fails on every sync cycle — logged and skipped in the sync loop (`sync.rs:250–256`) and on each received-payment event (`sync.rs:125–128`), which means payments received via the address lose their sender comment/metadata until the server is fixed. The app's own hand-signed `/recover` (mode recovery) and `/mode` calls are **unaffected**: they use the legacy `{message}-{timestamp}` format the server still verifies, and they cannot move to v2 because the SDK's `signMessage` refuses the `breez-lnurl:` namespace (`crates/breez-sdk/core/src/sdk/api.rs:200–208`).

**Proposal:** Gate the merge on `blink-lnurl-server` shipping the v2 verifier for available / register / unregister / recover / metadata **while keeping the legacy verifier on `/recover` and `/mode`** for the app's hand-signed calls — the server needs both, and the second requirement is the one a server-side reader will not find in upstream. Once the server is deployed, re-run the three flows above on staging with a 0.23.1 build (registration, fresh restore showing the existing address, a received payment carrying its metadata) and record the server version and date in the upgrade commit's body or the PR description. Because the 0.22 client and the 0.23 client cannot both be served by a server that only knows one format, the server must ship v2 support *alongside* the legacy format before this app version reaches any user, and keep it until every 0.22 build is gone — worth stating in the server PR.

---

## Test Coverage Gaps

1. **The post-server-fix verification (Finding 1).** Not a unit-testable seam; the coverage is a recorded staging run of the three affected flows — registration, fresh restore showing the existing address, and a received payment carrying its metadata — against the deployed `blink-lnurl-server`, with version and date.
2. **Upgrade-in-place (carried).** A wallet created on 0.22.0 and opened on 0.23.1 gets `brz_*` tree tables created in its existing storage DB and a persisted leaf tree (`9b360e875`). Additive by construction, but still not on any written matrix for this release.
3. **Custodial parity is now a real divergence and nothing documents it.** With the ceil in place, a 1 001 msat invoice shows "2 SAT" in self-custodial and "1 SAT" in custodial (`app/screens/send-bitcoin-screen/payment-destination/lightning.ts:49` still passes the raw `1.001`). Whether that is correct depends on what Galoy debits for the same invoice — unknown from this repo. One line in the PR ("custodial shows X because the backend debits Y") or a follow-up issue closes it; a test pinning the custodial display for the same vector would make the divergence deliberate rather than accidental.
4. **The Spark and memo tests added to `wrap-destination.spec.ts`** (`:127–140, :260–277`) are welcome but ride in a `fix(spark)` commit whose subject says nothing about them; they belong in the test commit, or in the body of this one.

## Code Smells

1. **Eight commits, zero bodies (carried).** `8dd3f795a` is the clearest case: a behaviour change to what every fixed-amount Lightning send displays, explained only in a code comment. The comment is good; the history should say why the change was made *now* (SDK 0.23 started ceiling) so `git log -S Math.ceil` tells the story without the review doc.
2. **`#4257` in the new checklist item vs. `4250` in the branch name** (`docs/pr-review.md:354`). One of them is the PR, the other may be an issue; worth a glance so the citation points where a reader will look.
3. **The `postSigned` comment slightly over-generalises** (`lnurl-server-mode.ts:105–108`): "`/recover` and `/mode` … work only while blink-lnurl-server keeps its legacy verifier on those routes." `/mode` is Blink's own route with Blink's own `mode:` message; it never had a Breez verifier to keep. Only `/recover` shares a route — and a message shape — with the upstream server. Saying so stops the next reader from hunting upstream for `/mode`.
4. **The migration spec now loads the entire real bridge to reach one helper** (`migration-transfer-request.spec.ts:20`: `...jest.requireActual("@app/self-custodial/bridge")`). It works because the shared SDK mock absorbs the imports, but `jest.requireActual("@app/self-custodial/bridge/receive").bolt11ReceiveMethod` would take just the function and keep the spec's mock surface explicit.
5. **`shownOnScreen` fakes four internal seams with partial shapes** (`…integration.spec.ts:22–51`): the `usePriceConversion` stub omits `convertMoneyAmount`, which `useDisplayCurrency` tolerates today. Fine for a SAT string; if the hook grows another dependency the failure will read as a hook bug. A one-line `renderDisplayCurrency()` helper next to the other hook tests would give future specs the same scaffold without copying the four mocks.
6. **Author string on `3b6bd86c5` (carried).** `Esaú Gómez)` — fix on merge.

grimen (changes requested):
My earlier coverage concern is resolved. The integration spec exercises the real parser, wrapper, details, fee quote and send, and reverting the `Math.ceil` in `wrap-destination.ts` fails 16 of the new tests (checked locally at a8a590b07, alongside a clean `tsc` and 23 self-custodial suites / 367 tests green). What follows is about what surrounds the merge.

## Findings

**1. Nothing mechanical stops this merging before `blink-lnurl-server` ships v2 support.**
The description and the reply to kngako both say this must wait until the server serves `POST /lnurlpay/{pubkey}/available` and verifies v2 signatures on staging. The 2026-09-17 probe got 404/400 on staging and production. But the PR is open, not a draft, unlabelled, and already approved. The only thing holding it now is this review's state. The gate depends on someone remembering.

*Example:* the stale review is dismissed while the server PR is in flight, and the PR is squash-merged. The next release ships a build where:
- no self-custodial user can claim an address;
- every fresh install and restore shows no Lightning address;
- received-payment metadata sync fails on every cycle.

Per Finding 3, users cannot then be rolled back to a 0.22 build.

*Proposal:* one option is converting to draft, or a `do-not-merge` label, until the server change is on staging and the three flows are recorded. I'll keep this review at changes requested until then.

**2. The app's `/recover` signature is byte-for-byte the legacy metadata credential, so a "verify both formats" server change keeps that credential alive as long as the app sends it.**
`recoverLnurlServerMode` signs `${identityPubkey}-${timestamp}` (`lnurl-server-mode.ts:77-80`).
- At 0.22.0 the SDK signs exactly that string for both `recover` and `list_metadata` (`crates/breez-sdk/core/src/lnurl.rs` @ 0.22.0:200, :337).
- Upstream's legacy verifier checks only timestamp freshness (±600 s) on those two routes. Neither claims the statement, so there is no replay protection (`lnurl/src/routes.rs` @ 0.23.1:1459-1466, :1484-1489, :1598-1602). The 0.23.1 source itself calls the recover candidate "byte-identical to the legacy metadata message".

v2 exists to scope each signature to one route and one domain. The description asks the server to "verify both signature formats, and read the metadata signature from either place", with no end condition. Because this app keeps sending the legacy `/recover` message, usage-based retirement never triggers for it: upstream's cutoff is 30 days of zero legacy verifies (`routes.rs:1705-1754`).

*Example:* a `/recover` request body is captured on its way somewhere: a request-logging proxy, an APM trace, a support debug capture. For ten minutes it then reads the wallet's LNURL payment metadata (sender comments, payer identities, zap requests) through `GET /lnurlpay/{pubkey}/metadata`. This is traced against upstream only; if `blink-lnurl-server` has diverged, the answer may differ.

*Proposal:* one option is to move the app's hand-signed recover off the shared legacy shape as part of the server work. `signMessage` refuses only messages that start with `breez-lnurl:`: case-insensitive, prefix-only (`crates/breez-sdk/core/src/sdk/api.rs` @ 0.23.1:668-686). So a Blink-owned, domain-bound message in the style of the existing `mode:` prefix is still signable, such as `recover:<domain>:<pubkey>-<ts>`. The server can then accept it on `/recover` alone, and let the legacy `{pk}-{ts}` verifier on `/recover` and `/metadata` sunset with the last 0.22 build. If keeping the legacy format indefinitely is deliberate, stating the sunset condition in the server PR would help. (`mode:{mode}:{pk}-{ts}` does not collide with anything: `:` is not a legal username character.)

**3. A wallet opened on 0.23.1 can no longer be opened by a 0.22 build, and nothing on the branch or in the description says so.**
0.23.1 appends a migration to the SDK's SQLite store: `ALTER TABLE unclaimed_deposits ADD COLUMN instant_claim_status` (`crates/breez-sdk/core/src/persist/sqlite.rs` @ 0.23.1:382). 0.22.0 runs `to_latest` in `SqliteStorage::new` (`sqlite.rs` @ 0.22.0:83), which returns `DatabaseTooFarAhead` once `user_version` is ahead, so `connect` fails. The reply to kngako says "Added: the storage migration is one-way, a 0.22 build fails with `DatabaseTooFarAhead`". That sentence is not in the description or any commit, and `git grep DatabaseTooFarAhead` is empty at a8a590b07.

*Example:* the release regresses somewhere unrelated, and the rollback candidate is still on 0.22. Or a QA device side-loads an older build. Every self-custodial wallet then fails to connect until the app is upgraded again. The seed keeps funds safe, but the wallet is unusable.

*Proposal:* one option is a line in the description and in the release notes, so any rollback decision knows it must stay on 0.23 or later.

**4. 0.23 turns on background exit-chain fetching for every wallet through `defaultConfig`, and neither new config default is named anywhere.**
0.23 adds `exitChainAutoFetchEnabled`, default `true` (`models/mod.rs` @ 0.23.1:611-623). It starts a background downloader that pulls exit chains from the Spark operators, and each `sync` waits for one collection pass (`runtime/client.rs:73-75`). `createSdkConfig` starts from `defaultConfig(network)` (`app/self-custodial/bridge/lifecycle.ts:47`). It sets neither this field nor `maxInstantDepositClaimFeeBps`, the new opt-in that is off when unset, which is the safe value.

This sits badly with two things already in the repo:
- `sdk-error.ts:36` describes unilateral-exit flows as ones "the app never starts".
- The checklist this PR edits says "don't rely on SDK defaults silently" (`docs/pr-review.md:349-350`).

The reply says the inherited SDK changes are "documented in the description". Neither field appears there, nor does the new `InvalidInput` for an explicit amount that disagrees with a fixed invoice (#1057).

*Proposal:* one option is setting both fields explicitly in `createSdkConfig`, with a line each on why, and asserting them in `lifecycle.spec.ts`. Whether auto-fetch stays on is a product call: it is what keeps a unilateral exit possible later, at the cost of background traffic to the operators. Making it explicit keeps that call visible.

## Test Coverage Gaps

- No CI build of the new native binaries on either platform. Both E2E jobs (run 35260746815) died at `nix: command not found` before building. That runner problem hits every branch, not just this PR, but it means the device evidence here is Android only. An iOS build that quotes and pays the fractional invoice would close this.
- Nothing asserts the SDK config fields the app now inherits (Finding 4). `lifecycle.spec.ts` pins `lnurlDomain` and `maxDepositClaimFee`, but not `exitChainAutoFetchEnabled` or `maxInstantDepositClaimFeeBps`.
- Upgrade-in-place is covered by the device runs (the same wallet opened on 0.22.0, then 0.23.1). The reverse direction is untested and unwritten (Finding 3).

## Code Smells

- `lnurl-server-mode.ts:105-106`: "the legacy `{message}-{timestamp}` format, which the SDK can no longer produce for us" reads as if the SDK can no longer sign legacy messages. It can: what it refuses is its own `breez-lnurl:` namespace, so these routes cannot move to v2 by hand. Rewording along those lines stops the next reader from concluding that `signMessage` changed.
- In the shared mock, the new `SendPaymentOptions.Bolt11Invoice` builds `{ tag, inner }`, while the existing `BitcoinAddress` right above it spreads the args at the top level (`__mocks__/@breeztech/breez-sdk-spark-react-native.js:83-92`). The spec-local mocks (`send.spec.ts:32`, `receive.spec.ts:14`) and the uniffi classes use `{ tag, inner }` for both. Aligning `BitcoinAddress` means no spec asserts a shape the SDK never produces.

PR #3884, opened 2026-06-26
chore(i18n): refine Dollar Balance popup copy (USD wording and shorter non-custodial text)


esaugomez31:
## What

Refines the Dollar Balance info popup copy, following product feedback:

- **Custodial body**: "you hold Bitcoin, not actual dollars" becomes "not actual USD".
- **Non-custodial body**: shortened to a single sentence ("In Non-Custodial Mode, your Dollar Balance uses supported stablecoins, designed to maintain the same value in USD terms.").

Both changes were propagated to English and all 28 locales.

### Custodial popup
image

### Non-custodial popup

image

PR #4262, opened 2026-09-09
fix(send): name the on-chain fee tiers Priority, Standard and Economy

### LINKED ISSUE #4260: bug(onchain): Correct fee-tier labels and confirmation times
## Problem

The on-chain payment flow still displays incorrect fee-tier names and/or confirmation-time estimates.

## Expected behavior

The fee options must be shown exactly as:

- **Priority** — `~ 10m`
- **Standard** — `~ 4h`
- **Economy** — `~ 24h`

## Acceptance criteria

- [ ] The payment flow displays the tier names **Priority**, **Standard**, and **Economy**, in that order.
- [ ] Priority displays `~ 10m`.
- [ ] Standard displays `~ 4h`.
- [ ] Economy displays `~ 24h`.
- [ ] No previous or alternative tier labels remain in the payment flow.
- [ ] Regression coverage verifies the exact labels and time estimates.

## Related work

This is a follow-up regression/product correction after https://github.com/blinkbitcoin/blink-mobile/issues/4230.

## Source

Requested by Slack user `U097VQ50JHJ`. The exact Slack message permalink was unavailable in the filing integration.

designsats:
Closes #4260.

The on-chain fee selector labelled its three tiers **Fast**, **Medium** and **Slow**. Those words describe the queue rather than what the sender is choosing between, and they did not match the naming used for the same payout speeds elsewhere. This renames the copy to **Priority**, **Standard** and **Economy**, in English and in all 28 locales.

The confirmation windows needed no change: `buildFeeTierOptions` already renders the tiers in Fast → Medium → Slow order, and `CUSTODIAL_PAYOUT_ETA_MINUTES` is already `{10, 240, 1440}`, so the custodial rail was already reading `~ 10m`, `~ 4h` and `~ 24h`. What was missing was any test holding it there.

The tier keys and the `FeeTierOption` values stay as they are. They name the backend payout queue the payment is put on, not the words on screen, and `PAYOUT_SPEED_BY_FEE_TIER` maps them onto `PayoutSpeed.Fast/Medium/Slow` — renaming them would be a schema change wearing a copy change's clothes.

## Copy that changed alongside the tiers

- **A marker comment on the shared keys.** `fast` / `medium` / `slow` are read by two rails whose windows differ by 24× — custodial payout queues at 10m / 4h / 24h, self-custodial mempool rates at 10m / 30m / 60m. `docs/pr-review.md` §2.7 asks for the marker so a future translator tuning "Economy" for the 24-hour queue doesn't silently re-word a 60-minute fee-rate picker.
- **Three dead keys deleted.** `UnclaimedDeposit.feeRateFastest` / `feeRateHalfHour` / `feeRateHour` ("Fastest (~{sats} sats) ~10 min", "Normal (…) ~30 min", "Economy (…) ~60 min") are the refund flow's pre-`buildFeeTierOptions` tier copy. They carry a third and fourth naming scheme for the same three tiers and are referenced nowhere in `app/`, `__tests__/` or `e2e/`. Removed from `en` and all 28 locales.

## Scope note

The self-custodial rail keeps its own `FEE_TIER_ETA_MINUTES` of 10 / 30 / 60. Those sends broadcast straight from the SDK against mempool's `fastest` / `halfHour` / `economy` rates rather than going through bria's payout queues, so `~ 4h` and `~ 24h` there would misstate what the sender is actually paying for. The names are shared; the windows are each rail's own, and both are now pinned by tests.

## Acceptance requirements

- [x] The payment flow displays the tier names **Priority**, **Standard** and **Economy**, in that order.
- [x] Priority displays `~ 10m`.
- [x] Standard displays `~ 4h`.
- [x] Economy displays `~ 24h`.
- [x] No previous or alternative tier label remains in the payment flow, in any locale, or in the specs around it — including a locale's own *translation* of a retired name, not just the English word.
- [x] Regression coverage verifies the exact labels and time estimates.

## Testing specs

**`__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts`** (new). Every existing spec around the fee selector mocks both the copy and the ETAs, so the tier names could have been renamed, or the payout windows re-timed, without a single test noticing. This one takes its labels from the shipped `app/i18n/en` strings and its windows from the shipped constants:

- custodial payout queues: `Priority ~ 10m`, `Standard ~ 4h`, `Economy ~ 24h`, in that order
- self-custodial mempool rates: `Priority ~ 10m`, `Standard ~ 30m`, `Economy ~ 60m`, read from `FEE_TIER_ETA_MINUTES` — which nothing in the suite read from source before, so re-timing `medium` to 45 minutes was invisible
- the `fast` / `medium` / `slow` ids the payout speeds are keyed on
- a sweep of `SendBitcoinScreen` and `UnclaimedDeposit` for every word these tiers have been called

**`__tests__/i18n/locale-parity.spec.ts`** (extended) and **`__tests__/i18n/retired-tier-copy.json`** (new). The first version of this case compared each locale's value against a list of **English** words, which only ever caught a locale that had never translated the tier at all — reverting all 28 files turned two of them red, `af` and `da`, the two carrying an untranslated `"Medium"`. The regression it is named for does not look like that: `es` held `Rápido / Medio / Lento`, `de` held `Schnell / Mittel / Langsam`, `ja` held `高速 / 中速 / 低速`.

The invariant the tier keys need is cross-version — when `en`'s tier copy changes, every locale's changes with it — and a spec that only sees the working tree cannot observe a change. `retired-tier-copy.json` approximates it with state: it records, per locale file, the tier strings that locale has already retired, seeded at `ef449120d` with the Fast / Medium / Slow generation. The case fails when a locale's current value is one of **its own** retired strings, as well as when it is a retired English word — the English list stays, since it independently catches untranslated fallbacks.

The ledger is append-only and grows by three entries per locale per rename, which is the point: a rename that skips a locale shows up as a missing line in the diff rather than as 28 files a reviewer has to read. A second case asserts every translation file appears in the ledger, so a locale added later cannot opt out of the check by never being listed.

The case that compares the selector's tier word with the Settings row's reduces each Settings label to its bare tier word, which means dropping the "Onchain" marker (`I-Onchain` in `xh`, オンチェーン in `ja`, trailing in `ar`). That started as a global regex replace and CodeQL flagged it as incomplete multi-character sanitization ([alert 8](https://github.com/blinkbitcoin/blink-mobile/security/code-scanning/8)): removing a word globally can rejoin the remainder into the same word. The marker only ever sits at one end of the label, so it is now sliced off that end once. Test-only code over translation files, so nothing was exploitable, but the alert is a real one about the pattern. Revert check: making the helper a no-op reddens the case for all 28 locales.

Existing key parity covers the deleted `feeRate*` keys across all 28 files.

**Revert checks.** Each case has a mutation it alone catches:

| Mutation | Reddens |
| --- | --- |
| `slow: "Economy"` → `"Slow"` in `en/index.ts` | both window cases + the sweep |
| Reinstating `UnclaimedDeposit.feeRateFastest` | the sweep alone |
| `de.json` `slow` left as the untranslated `"Slow"` | the locale-parity case alone |
| `es.json` `slow` reverted to its translated `"Lento"` | the locale-parity case alone (green before the ledger) |
| A 29th locale file added without a ledger entry | the ledger-coverage case alone |

The ids case correctly survives all of them, since the ids are not what changed.

Updated fixtures, so no superseded name is left standing in the suite: `fee-tier-options.spec.ts`, `fee-tier-selector.spec.tsx`, `use-onchain-fee-tier-options.spec.ts`, `unclaimed-deposits-screen.spec.tsx`.

Local:

- `yarn jest` — 9261 passed, 720 suites. One unrelated flake in `send-confirmation.spec.tsx` ("hide balance › leaves the amount being sent readable", a `₦0 ($0.00)` formatting assertion); passes standalone, and that spec has no fee-tier references.
- `yarn tsc:check` — clean
- `yarn eslint:check` — clean
- `yarn update-translations` — `i18n-types.ts` and `raw-i18n/source/en.json` regenerated and committed, so `check:translations` is a no-op

## Translation note

The 28 locale values were produced here rather than through a Crowdin round-trip, so they are a best-effort pass and want a native-speaker check before merge — particularly `lg`, `qu` and `xh`, where "Priority / Standard / Economy" have no settled fee-tier convention to borrow. The alternative the review offered — deleting the keys so the merge falls through to English — is available if any of these are wrong, but a wrong-but-plausible translation seemed worse to ship than a reviewable one.

## Screenshots


kngako (changes requested):
## 1. Findings

### F1 — [BLOCKING] The rename ships English-only

`app/i18n/en/index.ts` is the only dictionary that moved. Every locale is
`merge({}, en, rawTranslated)` ([es/index.ts:6](app/i18n/es/index.ts:6)), so a translated
value always wins over English — and all 26 wired non-English locales still hold their
translation of the *old* words. A Spanish sender on this build opens Send → on-chain and
reads `Rápido / Medio / Lento`; German reads `Schnell / Mittel / Langsam`; Japanese reads
`高速 / 中速 / 低速`. That is precisely the queue-describing naming the commit message
argues is wrong, still shipping to everyone outside `en`, and now inconsistent with the
`Onchain Priority (~10m)` row those same users see in Settings → Fee rates.

This is the repo's own gate, not a nicety: `docs/pr-review.md` §3 i18n requires "a copy
change regenerates **all ~28 locales in the same PR** (expect a wide but mechanical diff,
~30 files for two strings — #3884)", and every recent copy commit does exactly that —
#4228 touched 33 files, #4205 touched 31, #4236 touched 34. **Proposal:** land the
translated strings for all 26 locales in this PR via the usual translation pass
(`merge-translations.py` against `en-key-order.json`). If translations cannot be sourced
in time, the acceptable interim is to *delete* the three `SendBitcoinScreen.fast/medium/slow`
keys from each locale JSON so the merge falls through to the new English — a deliberate
English fallback beats a stale correct-looking translation, and it makes the debt visible
in the diff rather than invisible in production.

### F2 — [BLOCKING] The payout-queue comment now misnames the backend's own tier

[use-custodial-onchain-fee-tiers.ts:34](app/screens/send-bitcoin-screen/hooks/use-custodial-onchain-fee-tiers.ts:34)
documents what *the backend* advertises — "Mirrors the broadcast windows the backend
advertises for each payout queue (`payoutSpeeds` in galoy-values)" — and the diff rewrote
`Flexible ~24 hours` to `Economy ~24 hours` to match the new app copy. The comment was
introduced in `c8af618e2` with `Flexible` deliberately, at a point when the app already
said "economy" elsewhere, because `Flexible` is galoy-values' word, not Blink's. The
failure mode is concrete: the next engineer re-checking the 24-hour window opens
galoy-values, greps `payoutSpeeds` for `Economy`, finds nothing, and either assumes the
comment is stale or assumes the queue was renamed server-side.

**Proposal:** restore `Flexible` and let the comment keep doing its one job — pointing at
the upstream contract. If the two vocabularies are worth reconciling in-comment, say so
explicitly rather than collapsing them, e.g. "Priority ~10 minutes, Standard ~4 hours and
Flexible ~24 hours (the tier this app calls Economy)". Same treatment applies to the
`Standard` in that sentence, which happens to match both vocabularies today and should not
be assumed to keep doing so.

### F3 — The "Transaction priority" heading now collides with the "Priority" option

`FeeTierSelector` renders its `title` immediately above the selected option's label
([fee-tier-selector.tsx:46](app/screens/send-bitcoin-screen/fee-tier-selector.tsx:46)), and
the title is `LL.SendBitcoinScreen.feeTier()` = `"Transaction priority"`
([send-bitcoin-details-screen.tsx:658](app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx:658)).
The default custodial tier is `Fast`, so the field on first render now reads
"Transaction priority" with "Priority (500 sats)" directly beneath it — the same word
twice, one as the question and one as the answer. Pick the cheapest option and it reads
"Transaction priority: Economy", which parses as a contradiction on a quick scan.

**Proposal:** rename the heading rather than the options, since the options are the ones
that now match Settings → Fee rates. `"Payout speed"` or `"Confirmation speed"` both
describe the axis without borrowing an option name, and neither is used elsewhere in
`SendBitcoinScreen`. This is a copy call rather than a code call, so it wants design
sign-off — but it should be made in this PR, because it is a direct consequence of the
rename and will otherwise ship as-is. Note the heading key is shared with the refund flow's
selector, so whatever lands applies to both.

### F4 — Three tier keys, two domains, windows that differ by 24×, no marker comment

`SendBitcoinScreen.fast/medium/slow` have two consumers:
[use-onchain-fee-tier-options.ts:90](app/screens/send-bitcoin-screen/hooks/use-onchain-fee-tier-options.ts:90)
and [unclaimed-deposits-screen.tsx:52](app/screens/unclaimed-deposits/unclaimed-deposits-screen.tsx:52).
The custodial rail pairs them with `10 / 240 / 1440` minutes; the self-custodial refund
rail pairs the same three strings with `FEE_TIER_ETA_MINUTES` = `10 / 30 / 60`
([payment.ts:37](app/types/payment.ts:37)). Before this diff "Slow" was vague enough to
straddle both. Now "Economy" is the word Settings → Fee rates attaches specifically to
`(~24h)`, so the refund screen ships a tier labelled `Economy ~ 60m`. A future translator
or copy editor tuning "Economy" for the 24-hour custodial queue silently re-words a
60-minute network-fee-rate picker in another domain.

**Proposal:** add the marker comment `docs/pr-review.md` §2.7 asks for (#3864) above the
three keys in [en/index.ts:2479](app/i18n/en/index.ts:2479), naming both consumers and the
fact that the windows differ per rail — or duplicate the keys per domain now, which is the
option that guide offers when the sharing is not actually intentional. Given the refund
flow already owns superseded per-domain copy for exactly this (see G5), duplicating is
probably the honest fix. Low urgency, but it is cheap and the guide names it explicitly.

### F5 — `leaves no superseded tier name in the on-chain copy` cannot fail on its own

[fee-tier-labels.spec.ts:65](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:65)
asserts the three label values are not `"Fast"` / `"Medium"` / `"Slow"`, but `labels` is
built from the same three keys the first case already asserts equal exactly
`Priority / Standard / Economy`. Mutation-tested both ways: renaming `slow` back to
`"Slow"` fails cases 1, 2 **and** 4; renaming it to `"Frugal"` fails cases 1 and 2 and
leaves 4 green. There is no mutation that case 4 catches and case 1 misses — it is strictly
weaker, which is the repo's Tier-2 "false-confidence test" shape (`docs/pr-review.md` §2.1).
Its name also claims a sweep of "the on-chain copy" that it never performs.

**Proposal:** either delete the case, or make it earn its name by actually sweeping — walk
`en.SendBitcoinScreen` and `en.UnclaimedDeposit` for the superseded words and assert none
survive. The sweep version would immediately go red on the three orphaned
`UnclaimedDeposit.feeRate*` keys (G5), which is the point: that is the check the commit
message claims to have made. If neither is wanted, drop it — three cases that each catch a
distinct mutation is a stronger spec than four where one is decorative.

---

## 2. Test Coverage Gaps

**G1 — The self-custodial windows are still unpinned.** The new spec pins
`CUSTODIAL_PAYOUT_ETA_MINUTES` only. `FEE_TIER_ETA_MINUTES` (`10 / 30 / 60`) drives both
the self-custodial send rail
([use-onchain-fee-tiers.ts:20](app/screens/send-bitcoin-screen/hooks/use-onchain-fee-tiers.ts:20))
and the refund rail, and no test reads it from source — `use-onchain-fee-tier-options.spec.ts`
mocks the ETAs into its fixtures. Re-timing `medium` to 45 minutes there is invisible to the
whole suite. Extend the new spec with a second `describe` over the self-custodial constant,
asserting `Priority ~ 10m / Standard ~ 30m / Economy ~ 60m`.

**G2 — The second consumer of these keys is still fixture-mocked.**
[unclaimed-deposits-screen.spec.tsx:73](__tests__/screens/unclaimed-deposits/unclaimed-deposits-screen.spec.tsx:73)
hand-writes `fast: () => "Priority"` into a mocked `i18n-react`. The diff updated the
fixture, which keeps the suite green but re-creates the exact blind spot the new spec was
written to close: rename the shipped copy again and this file stays green while the refund
screen ships the old words. One assertion in the new spec covering the refund flow's label
map — or a shared helper both specs read the real `en` through — would close it.

**G3 — Nothing tests any locale but English.** F1 is invisible to CI: `check:translations`
only asserts `i18n-types.ts` is regenerated, and no spec inspects the locale JSONs. A small
parity spec over `app/i18n/raw-i18n/translations/*.json` — asserting each locale either
carries a `SendBitcoinScreen.fast/medium/slow` value or omits the key entirely, and that
none of them equals a known-superseded English word — would have caught this diff before
review and would keep catching the next copy change.

**G4 — The heading/option pairing is untested.** No spec asserts what `feeTier` renders
alongside the option labels, so F3 was introduced without a red test. `fee-tier-selector.spec.tsx`
passes a `title` fixture rather than the shipped string. Reading `en.SendBitcoinScreen.feeTier`
in the new spec and asserting it does not equal, contain, or be contained by any tier label
would pin the collision cheaply.

**G5 — Superseded on-chain tier copy is still standing, contrary to the commit message.**
`UnclaimedDeposit.feeRateFastest` / `feeRateHalfHour` / `feeRateHour`
([en/index.ts:4238](app/i18n/en/index.ts:4238)) read `"Fastest (~{sats} sats) ~10 min"`,
`"Normal (…) ~30 min"`, `"Economy (…) ~60 min"` and are referenced nowhere in `app/`,
`__tests__/` or `e2e/` — confirmed by grep. They are the refund flow's pre-`buildFeeTierOptions`
tier copy, they carry a fourth and fifth naming scheme for the same three tiers, and they
survive in all 28 locale files. The commit says "no superseded tier name is left standing in
the suite"; these are left standing in the dictionary. Deleting them (all locales) is a
one-line-per-file cleanup and would make G3's sweep meaningful.

---

## 3. Code Smells

**S1 — Dead `jest.mock("react-native-fs", …)`.**
[fee-tier-labels.spec.ts:12](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:12).
Verified removable: deleting the block leaves all four cases passing. Nothing in the import
graph (`@app/i18n/en`, `fee-tier-options`, `fee-tiers.types`, `use-custodial-onchain-fee-tiers`)
reaches `react-native-fs`. It reads as copied scaffolding and will outlive its own
explanation. Drop it.

**S2 — `enBase as unknown as Translations` double cast.**
[fee-tier-labels.spec.ts:17](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:17).
The comment explains *why*, which is the right instinct, but a double cast through `unknown`
opts the whole dictionary out of type-checking — including the three property reads directly
below it. Narrower and just as honest: type the three values individually, e.g.
`const fast: string = enBase.SendBitcoinScreen.fast as string`, or cast only
`enBase.SendBitcoinScreen`.

**S3 — Case 2 subsumes case 1.** Case 2 asserts `{label, detail}` pairs for all three tiers;
case 1 asserts the labels alone. Every mutation that reddens case 1 also reddens case 2 —
observed in both mutation runs above. Keeping both is defensible for failure-message clarity
(case 1 fails with a clean label diff), but it is duplication worth a deliberate call rather
than an accident.

**S4 — `describe("on-chain fee tier presentation")` and the filename overpromise scope.**
The spec covers the custodial rail only, while "on-chain fee tier presentation" and
`fee-tier-labels.spec.ts` both read as covering every on-chain rail — which is what makes G1
easy to miss. Either narrow the name (`custodial on-chain fee tier presentation`) or widen
the spec to match the name.

**S5 — `labels` and `buildOptions()` at `describe` scope, rebuilt per case.** Minor: the
`labels` object is module-level shared mutable state that all four cases read, and
`buildOptions()` re-runs the builder for each assertion rather than once. Harmless here
because nothing mutates, but the surrounding specs in this directory build their fixtures
inside the `it` blocks; matching that would be more consistent with the neighbourhood.

kngako (changes requested):
Round 1's F1–F5 are all addressed; one new blocking finding (F6) plus F7–F11 and refreshed gaps/smells.

---

## 1. Findings

### F6 — [BLOCKING] The locale guard catches 2 of the 28 locales it was written for

`2e846231b` says "`locale-parity.spec.ts` gains a case asserting no locale carries a
superseded tier name, so the next copy change cannot ship `en`-only unnoticed." The case
compares each locale's tier value against a list of **English** words —
`name.toLowerCase() === value.trim().toLowerCase()` at
[locale-parity.spec.ts:158](__tests__/i18n/locale-parity.spec.ts:158) — but the regression
it names never looked like that. Before `0df8394a8`, `es.json` held `Rápido / Medio /
Lento`, `de.json` held `Schnell / Mittel / Langsam`, `ja.json` held `高速 / 中速 / 低速`.
None of those equals `"fast"`, so none is caught.

Mutation-proven twice. Restoring exactly the pre-fix Spanish values leaves the whole file
green — **84 passed, 0 failed**. Reverting all 28 locale files to `ef449120d` fails the new
case in **`af.json` and `da.json` only** — the two that happened to carry the untranslated
English word `"Medium"`. The other 26 pass. The 28 further failures in that run are the
*key-parity* case reacting to the deleted `UnclaimedDeposit.feeRate*` keys, not to the
rename; unbundle that cleanup and reverting the entire i18n commit turns two files red out
of twenty-eight. By `docs/pr-review.md` §2.1 — "if someone reverted the core line of this
diff, which test goes red?" — the answer for `0df8394a8` is "two locales' worth", and by
the same guide's taxonomy this is a Tier-2 false-confidence test: it passes while the exact
bug it is named for is alive.

**Proposal.** The real invariant is cross-version ("when `en`'s tier copy changes, every
locale's must too"), which a stateless spec cannot see, so approximate it with state: check
in a `__tests__/i18n/retired-tier-copy.json` mapping locale → the strings that locale has
retired (`es: ["Rápido","Medio","Lento"]`, `de: ["Schnell","Mittel","Langsam"]`, …, seeded
from `git show ef449120d:app/i18n/raw-i18n/translations/*.json`) and assert no locale's
current value appears in its own retired list. It grows by three entries per locale per
rename and the reviewer sees that growth in the diff, which is exactly the visibility that
is missing. Keep the English-word list as well — it independently catches untranslated
fallbacks, which is what it caught here. If a repo-wide guard is preferred, the more
general form is a CI check: a changed value under `SendBitcoinScreen.fast|medium|slow` in
`en/index.ts` requires the 28 locale files to have changed in the same commit. Until one of
these exists, the honest edit is to soften the commit-message claim to what the case
actually does — catch untranslated English left behind in a locale.

### F7 — The comment fix changed two more names than it corrected, unsourced

The round-1 finding was one word: the comment documents `payoutSpeeds` in galoy-values, and
`Flexible` had been rewritten to the app's `Economy`. `ef449120d` restores `Flexible` — and
also rewrites `Priority ~10 minutes, Standard ~4 hours` to `Fast ~10 minutes, Medium ~4
hours`, plus `Standard moved off bria's half-hour queue` to `Medium moved off…`
([use-custodial-onchain-fee-tiers.ts:34](app/screens/send-bitcoin-screen/hooks/use-custodial-onchain-fee-tiers.ts:34)).
Those two names were not in question, and the resulting triple is internally odd: `FAST`
and `MEDIUM` are the GraphQL enum constants
([supergraph.graphql:1697](supergraph.graphql:1697) — `enum PayoutSpeed { FAST MEDIUM SLOW }`),
so if the sentence is quoting the enum the third should be `SLOW`, and if it is quoting
`payoutSpeeds`' display names then the first two presumably are not the enum words. The
original author of `c8af618e2` wrote `Priority / Standard / Flexible`, which at least reads
as one coherent set of display names. Grepping the repo, the only occurrence of `Flexible`
as a payout speed is this comment itself — nothing here can settle it.

**Proposal.** Paste the actual `payoutSpeeds` entry (galoy-values path plus the three
`displayName` values) into the PR description; if `Priority / Standard` were right, restore
them. If the names were taken from the GraphQL enum rather than from galoy-values, say
`FAST / MEDIUM / SLOW` and drop the galoy-values citation, because the two are then
different sources. Worth a follow-up ticket either way: the schema exposes
`PayoutSpeeds { description displayName speed }`
([supergraph.graphql:1705](supergraph.graphql:1705)), so the windows and names could be
read from the backend instead of mirrored in a comment that goes stale silently — that is
the durable fix, and too big for this PR.

### F8 — The sweep false-positives on ordinary English

`survivingTierNames` matches `Fast`, `Medium`, `Slow`, `Fastest`, `Normal`, `Flexible`,
`Half hour` case-insensitively as whole words against every string value in
`SendBitcoinScreen` and `UnclaimedDeposit` — 72 strings today
([fee-tier-labels.spec.ts:45](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:45)).
Those are ordinary words, not reserved ones. Verified: adding
`networkStatus: "Network conditions are normal right now."` to `SendBitcoinScreen` — copy
with nothing to do with fee tiers — turns the case red with a message saying a superseded
tier name is standing in the on-chain copy. `Medium` and `Fast` are equally exposed; the
next person to write "medium confidence" or "fast enough" into either section gets a
failure that points at the wrong thing.

**Proposal.** Scope the sweep to the copy that can actually render a tier name rather than
two whole sections: the three tier keys plus any key matching `/feeRate|tier/i`, which is
where every superseded name in this branch's history lived. That keeps the mutation the
case exists for — reinstating `feeRateFastest` still reddens it, since the key name matches
— while removing the collision surface. If sweeping broadly is preferred, invert it to an
allowlist of keys permitted to contain those words, so a false positive is resolved by a
one-line, review-visible addition rather than by rewording innocent copy.

### F9 — The sweep silently drops two strings

`Object.values({ ...sendScreen, ...en.UnclaimedDeposit })`
([fee-tier-labels.spec.ts:113](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:113))
merges two objects that share a key: both sections define `title`, so `UnclaimedDeposit`'s
overwrites `SendBitcoinScreen.title` (`"Send Bitcoin"`) and that string is never swept. The
`typeof value === "string"` filter then drops `SendBitcoinScreen.confirmFeesModal`
entirely — a nested object holding three more strings. Both are harmless today (I checked
their contents; no tier name in either), but the sweep quietly covers less than it looks
like it does, and the next nested section added to `SendBitcoinScreen` will be skipped
without a word.

**Proposal.** Concatenate instead of merging — `[...Object.values(sendScreen),
...Object.values(en.UnclaimedDeposit)]` fixes the collision in one character's worth of
change. For the nesting, `locale-parity.spec.ts` already exports exactly the right helper:
`collectLeaves` flattens a translation object to dotted leaf paths, which is what this
sweep wants and what would also give failures a key name instead of a bare string. Reusing
it satisfies §2.4 and removes the hand-rolled traversal.

### F10 — The marker comment cannot reach the people §2.7 wrote it for

The marker at [en/index.ts:2480](app/i18n/en/index.ts:2480) is good and says the right
things. But §2.7's stated purpose is "so a future **translator** doesn't diverge them
accidentally", and translators work from `raw-i18n/source/en.json`, which
`utils/export-language.mjs` produces through `JSON.stringify` — verified, the shipped
`en.json` carries no trace of the comment. So it protects the engineer editing
`en/index.ts` and nobody downstream of the export, which is where the divergence risk
actually sits.

**Proposal.** Keep it — it is still the cheaper half of §2.7's two options and it did not
exist before. But note the trade in the PR description: the branch chose the comment over
the split, and the same commit deleted the only per-domain tier keys the refund flow ever
had. If the two rails' windows diverge further (they are already 24× apart at the slow
tier), the split is the option that survives export, and the deleted keys are the seed for
it. A follow-up ticket is a fine home for that.

### F11 — German drops the word this codebase had already settled on

`0df8394a8` deletes `de.feeRateHour: "Sparsam (~{sats} sats) ~60 Min"` and, in the same
commit, translates the new `slow` tier as `Ökonomisch`. `Sparsam` was this repo's own
already-shipped German for exactly this tier; `Ökonomisch` is a calque that reads closer to
"economic" in the academic sense than to "cheap option". Spot-checking the rest, the batch
is good — `es: Prioritario / Estándar / Económico`, `fr: Prioritaire / Standard /
Économique`, `ja: 優先 / 標準 / 節約` all read naturally, and the deletion cleared the
untranslated `"Economy (~{sats} sats) ~60 min"` string that had been sitting in about
twenty locales (#3862).

**Proposal.** Use `Sparsam`, or `Günstig` if a shorter word is wanted for the field. It is
a one-word change and it keeps the German consistent with what the app shipped last week.
Per `docs/pr-review.md` §3, this is the kind of thing a reviewer who speaks the language
should call rather than wave through on mechanics — so treat it as a prompt for whoever
speaks German on the team, not as a veto from me.

---

## 2. Test Coverage Gaps

**G6 — The tier → payout-queue mapping is pinned by nothing.**
`PAYOUT_SPEED_BY_FEE_TIER`
([use-custodial-onchain-fee-tiers.ts:27](app/screens/send-bitcoin-screen/hooks/use-custodial-onchain-fee-tiers.ts:27))
is what turns the user's pick into `PayoutSpeed.Fast|Medium|Slow` on the mutation, and it
is referenced nowhere under `__tests__/` — confirmed by grep. The new spec pins the option
*ids* (`["fast","medium","slow"]`), which is one step short: swap two entries in that map
and a sender who picks Economy is charged for and queued on Priority, with all 472 tests
still green. This is money movement, so it is the one gap here worth closing in this PR —
three `expect` lines next to the existing id case.

**G7 — The refund flow's labels are still fixture-mocked** (round-1 G2, unresolved).
[unclaimed-deposits-screen.spec.tsx:76](__tests__/screens/unclaimed-deposits/unclaimed-deposits-screen.spec.tsx:76)
still hand-writes `fast: () => "Priority"` into a mocked `i18n-react`. The new spec covers
the *self-custodial window set* that flow uses, which is real progress, but the screen
itself remains pinned to a fixture: rename the shipped copy again and that file stays green
while the refund screen ships the old words. One assertion reading the real `en` through
the same label map would close it.

**G8 — The heading collision is checked in `en` only.** The `does not head the selector
with one of the option names` case reads `sendScreen.feeTier`, and
`locale-parity.spec.ts`'s superseded check covers only the three tier keys — not `feeTier`.
All 28 locales did change the heading in this branch, but nothing stops a future
translation rendering `feeTier` as `Prioridad` next to a `Prioritario` option and
reproducing F3 in Spanish. The collision check is per-locale-shaped already; running it
inside the existing `describe.each` over locale files would generalize it for free.

**G9 — Nothing detects an unused i18n key.** The three `UnclaimedDeposit.feeRate*` keys sat
unreferenced across 28 locale files until this branch removed them by hand. The deletion is
correct — I re-verified no reference exists in `app/`, `__tests__/` or `e2e/` — but the
next set will accumulate the same way. A spec that walks `en.json` leaves and greps `app/`
for each key path would be a cheap standing guard; it belongs in a separate PR, not this
one.

---

## 3. Code Smells

**S6 — `SUPERSEDED_TIER_NAMES` is defined twice, differently.**
[fee-tier-labels.spec.ts:34](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:34)
lists seven names; [locale-parity.spec.ts:143](__tests__/i18n/locale-parity.spec.ts:143)
lists six — the same list minus `"Half hour"`. Two copies of one fact, already diverged in
the commit that introduced them, in files that will be edited at different times. Extract
one fixture and import it into both.

**S7 — `"Half hour"` was never shipped copy.** It is the Breez SDK's field name
(`rates.halfHour`), not a string any screen displayed — the key that existed was
`feeRateHalfHour`, whose *value* read `"Normal (…) ~30 min"`. Its presence makes the list
look assembled from identifiers rather than from retired user-facing strings, which is what
the sweep is for. Drop it, or add a word saying why an identifier is in a copy list.

**S8 — The sweep hand-rolls a traversal that already exists.** See F9: `collectLeaves` in
`locale-parity.spec.ts` does flattening properly, including nesting. §2.4.

**S9 — `presentTiers()` rebuilds the tier record on every call.** Three cases call it, each
re-running `buildZeroTiers`. Harmless, and arguably clearer than sharing state — noting it
only because round 1 flagged the same shape and it is unchanged.

**S10 — `describe("the copy around the tiers")` covers only the send screen's heading.**
The refund flow's heading (`UnclaimedDeposit.feeRate`) sits directly above the same three
labels at
[unclaimed-deposits-screen.tsx:164](app/screens/unclaimed-deposits/unclaimed-deposits-screen.tsx:164)
and is not checked for the same collision. `"Network fee"` cannot collide today, so this is
scope drift in the name rather than a bug — either check both headings or name the case for
the one it checks.

kngako (changes requested):
One blocking issue, and it is a one-line fix. The rest concern the test design the branch introduces, which is more elaborate than what it enforces.

---

## Findings

### 1. The payout-queue comment now states the wrong backend vocabulary — [BLOCKING]

`ef449120d` rewrites [use-custodial-onchain-fee-tiers.ts:35](app/screens/send-bitcoin-screen/hooks/use-custodial-onchain-fee-tiers.ts:35) to say galoy-values names the queues "Fast ~10 minutes, Medium ~4 hours and Flexible ~24 hours — galoy-values' names for the queues". The backend's own config schema has, under `payoutSpeeds`, `displayName: "Priority"` for `fast`, `"Standard"` for `medium` and `"Flexible"` for `slow` — which is exactly what the comment said on `main`, written in #4083 by someone citing galoy-values and blink-deployments#10052 directly. "Fast" and "Medium" are the config keys and the GraphQL enum values (`PayoutSpeed.Fast` is two lines above), not display names; the sentence now mixes two enum keys with one display name as though they were one vocabulary. This is the only substantive content of that commit, and its stated purpose was to stop the next engineer being misled about the backend's names.

**When it comes up:** the next engineer reconciling a queue change reads "galoy-values names them Fast, Medium", greps galoy-values for a `displayName` of "Fast", finds only the key, and concludes the comment is stale — the precise failure the commit message says it was written to prevent. The muddle has already propagated once: both new spec files list "Flexible" as a *superseded app tier name*, but "Flexible" was never app copy; it is the backend's current display name for SLOW.

**Proposal:** Restore `main`'s wording for the backend half and keep the branch's useful addition about the app's words, e.g. *"`payoutSpeeds` displayNames in galoy-values: Priority ~10m, Standard ~4h, Flexible ~24h. This app shows the third as Economy to match the Fee rates screen (#4228)."* If galoy-values genuinely overrides `displayName` away from the schema defaults, cite the file and line in the comment so the claim is checkable next time. Drop "Flexible" from both `SUPERSEDED_TIER_NAMES` lists (Finding 4). Separately worth a follow-up ticket, not this PR: the backend already serves `PayoutSpeeds { displayName description }` and nothing in the app queries it, yet `description` is the field that has already drifted once (blink-deployments#10052) and needed an app release to catch up.

### 2. Eight locales now say one thing in Settings → Fee rates and another in the send selector — [NON-BLOCKING]

The commit message for `0df8394a8` gives the motivation as the `Onchain Priority (~10m)` row in Settings disagreeing with the tier above it. #4228 already translated that row's three tier words in every locale, but the new selector words were chosen without consulting them. Comparing `FeeRatesScreen.onchain{Priority,Standard,Economy}` with `SendBitcoinScreen.{fast,medium,slow}` per locale — ignoring case, grammatical gender and the "Onchain" prefix — the words differ in `ar` (Standard: عادي vs قياسي), `de` (Economy: "Economy" vs "Ökonomisch"), `el` (Standard: κανονικό vs Τυπική), `id` (Economy: ekonomis vs Ekonomi), `sr` (Priority: приоритетно vs Приоритет), and for all three tiers in `lg`, `qu` and `xh`. Fourteen tier words in total.

**When it comes up:** a German user opens Settings → Fee rates and reads "Onchain Economy (~24 Std.)", then goes to send on-chain and is offered "Ökonomisch (…) ~ 24h" — the same tier, two different words, on the two screens this PR exists to bring into agreement. ("Ökonomisch" is also the academic sense of *economic* in German; a fee tier would normally say "Günstig" or "Sparsam". The Settings row kept the English word, so neither screen has a natural German term yet.)

**Proposal:** Pick one word per tier per locale and apply it to both keys in this PR — the branch already touches all 28 files and it is 14 edits. Where the Settings row's word is the more natural one (`ar` عادي, `el` κανονικό, the `lg`/`qu`/`xh` rows) use it in the selector; where neither is good (`de` Economy) fix both. Then add the comparison as a case in `locale-parity.spec.ts` — strip the "Onchain " prefix and the parenthetical, normalize with `toLocaleLowerCase`, keep an explicit per-locale allowlist for grammatical agreement such as `el` οικονομικό/Οικονομική — so the two screens cannot diverge again. That is a stronger guard than the ledger, and it needs no state.

### 3. The retired-copy ledger cannot see the rename it was built to catch — [NON-BLOCKING]

`a38fd0897` adds [retired-tier-copy.json](__tests__/i18n/retired-tier-copy.json) and `RETIRED_TIER_COPY` ([locale-parity.spec.ts:164](__tests__/i18n/locale-parity.spec.ts:164)) so that "a rename that skips a locale shows up as a missing line in the diff". But the ledger only knows generations that someone has already written into it. Mutation A: rename `en` to Express / Regular / Saver and touch nothing else. Result: `fee-tier-labels.spec.ts` fails its two English pins; **all 28 `locale-parity` cases stay green**. The natural fix for two red English pins is to update the two expectation arrays — and the rename ships `en`-only, which is precisely the regression `0df8394a8` fixed. The branch's own mutation check ("reverting all 28 locales fails 28 cases") exercised the case where the ledger had *already* been appended to, which presupposes the discipline the ledger exists to replace.

**When it comes up:** the next copy tweak to these three words. The engineer sees two red cases naming Priority / Standard / Economy, updates them, sees green, ships. Spanish senders read "Prioritario" under an English "Express", and the ledger's 60 lines of comment describe a guarantee that was never in force.

**Proposal:** Make the ledger red on an `en` change with no locale work: add an `"en"` entry recording the generation the locales were translated against (`"en": ["Priority", "Standard", "Economy"]`) and assert it equals the live `en` values in `locale-parity`. An `en`-only rename then reddens *that* case with the ledger's path in the failure — which is where the append-only instructions live — and the 28 per-locale entries keep doing what they do now. A leaner alternative is to store per locale only the English triple it was last translated against and assert equality with `en`: an `en` rename reddens 28 cases, each fix touches that locale's own line, and there is no foreign-word list to maintain or to false-positive on (Finding 4). It does admit a lazy "bump the line, keep the old word" fix, but that is visible in the diff as a ledger line changing without its locale file, which is the review signal the commit wanted anyway.

### 4. The global English ban list rejects correct translations, and bans words that were never tier copy — [NON-BLOCKING]

`SUPERSEDED_TIER_NAMES` at [locale-parity.spec.ts:149](__tests__/i18n/locale-parity.spec.ts:149) is compared against every locale's *value* for the three tier keys. Only Fast / Medium / Slow were ever values of those keys; "Fastest", "Normal" and "Flexible" come from the deleted `UnclaimedDeposit.feeRate*` strings and from the hook comment (Finding 1), so they guard nothing. And the list is English spelling, not meaning. Mutation B: set `es.json` medium to "Normal" — a perfectly good Spanish rendering of Standard — and `locale-parity` fails with *carries no superseded on-chain fee tier name*. The same is true of "Medium" in Danish, Dutch and Afrikaans, and "Flexible" in Spanish and French; `hu` already skirted the rule with "Normál".

**When it comes up:** the next translator, or the fix for Finding 2, picks the natural word and gets a red case whose message accuses them of leaving a superseded name behind. They pick a worse word to make the test pass, and the locale ships it.

**Proposal:** In `locale-parity`, drop the global list and compare each locale only against its own ledger entries plus the words `en` itself retired for these keys (the `"en": ["Fast", "Medium", "Slow"]` entry from Finding 3's proposal). That keeps the af/da untranslated-fallback catch — the actual #3862 bug — while a locale can use "Normal" freely. In `fee-tier-labels.spec.ts`, remove "Flexible" and "Half hour" (neither was ever copy); Finding 5 covers the rest of that sweep.

### 5. The copy sweep bans ordinary English words across 73 unrelated strings — [NON-BLOCKING]

[fee-tier-labels.spec.ts:112](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:112) regex-scans every string in `SendBitcoinScreen` (49) and `UnclaimedDeposit` (24) for `\bfast\b`, `\bslow\b`, `\bnormal\b`, `\bmedium\b`, `\bflexible\b`, case-insensitively. Mutation C: add `feeHint: "Higher tiers confirm fast; lower ones can be slow when the mempool is busy."` — a sentence with nothing to do with tier names — and the case fails. The commit message says the sweep's job is to stop `feeRateFastest` & co. being reinstated as a second naming scheme; what it actually asserts is that two large sections of copy never use five common adjectives.

**When it comes up:** any future on-chain copy in either section — a fee warning, a tooltip, a "taking too long?" prompt. The failure says *leaves no superseded tier name standing*, and the author goes looking for a tier they never touched.

**Proposal:** Scope the sweep to keys that name tiers — the three tier keys plus any key matching `/feeRate|tier/i` — or replace it with the exact assertion it stands in for: `en.UnclaimedDeposit` has no key matching `/^feeRate(Fastest|HalfHour|Hour)$/`. The first case in the file already pins the three labels exactly, so the sweep's only unique job is the reinstatement guard; make it say that. Escape the names before interpolating them into `new RegExp` while there.

### 6. The heading now shares a word with the default option, and the branch's own reasoning about it is lost — [NON-BLOCKING]

After the rename, `FeeTierSelector` renders "Transaction priority" ([en/index.ts:2478](app/i18n/en/index.ts:2478), passed at [send-bitcoin-details-screen.tsx:658](app/screens/send-bitcoin-screen/send-bitcoin-details-screen.tsx:658)) directly above the selected label ([fee-tier-selector.tsx:47](app/screens/send-bitcoin-screen/fee-tier-selector.tsx:47) → [:54](app/screens/send-bitcoin-screen/fee-tier-selector.tsx:54)), and the custodial default is `FeeTierOption.Fast` ([use-onchain-fee-tier-options.ts:63](app/screens/send-bitcoin-screen/hooks/use-onchain-fee-tier-options.ts:63)). `0df8394a8` changed the heading to "Confirmation speed" for this reason and `4c92c77cd` reverted it as outside the ticket. The net diff carries neither the change nor a note.

**When it comes up:** every custodial on-chain send at first render reads "Transaction priority / Priority (1,200 sats) ~ 10m". After picking a tier it reads fine ("Transaction priority / Economy" is ordinary English), so this is an awkwardness rather than a contradiction — the commit message overstated it — but the next person to notice will re-derive the whole change-and-revert.

**Proposal:** This is a product call, not a defect — the Settings screen already uses "Priority" as a tier name, so the vocabulary is established. Make the call once: raise it in the PR description as a question for design, and if the heading stays, add one line to the marker comment at [en/index.ts:2480](app/i18n/en/index.ts:2480) recording that "Transaction priority" over "Priority" was considered and kept. If design prefers a different heading, that is a one-key, 28-locale change in its own PR.

---

## Test Coverage Gaps

1. **No test reproduces the regression the ledger is for.** Mutation A (an `en`-only rename to a new generation) leaves all 28 locale cases green. Adding the `"en"` generation line and asserting it against live `en` (Finding 3) closes this with one case.

2. **"is covered by the retired tier copy ledger" is satisfied by an empty entry.** [locale-parity.spec.ts:223](__tests__/i18n/locale-parity.spec.ts:223) checks `Object.keys(...).toContain(localeFile)`, which `"xx.json": []` passes — so a newly added locale gets a green "covered" case and no per-locale guard. It also forces a meaningless entry for a locale that has never been through a rename. Assert on content, or (with Finding 3's leaner design) on the translated-against triple, which a new locale *would* have.

3. **No cross-screen vocabulary test.** The Settings tier words and the selector tier words are asserted independently and nowhere together; Finding 2's fourteen mismatches were found with a shell loop. A per-locale comparison in `locale-parity` (normalize with `toLocaleLowerCase`, strip the "Onchain " prefix and the parenthetical, explicit allowlist for grammatical agreement) would have caught them and needs no state.

4. **Ledger well-formedness is unchecked.** Nothing asserts entries come in multiples of three, contain no duplicates, or that every ledger key is a locale file that still exists — a stale `"xy.json"` entry would sit forever. One `describe("retired tier copy ledger")` with three `it`s.

5. **`normalize` is ASCII-only in intent but used on 28 scripts.** `value.trim().toLowerCase()` ([locale-parity.spec.ts:172](__tests__/i18n/locale-parity.spec.ts:172)) is fine for the current entries, but the ledger is meant to hold whatever each locale writes; a Turkish "İ" or a title-cased Greek entry compares differently than a reader expects. `toLocaleLowerCase()` plus a case in the ledger block that round-trips each entry through `normalize` costs nothing now and avoids a confusing red later.

---

## Code Smells

1. **`SUPERSEDED_TIER_NAMES` is defined twice with different contents** — seven entries at [fee-tier-labels.spec.ts:34](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:34) (with "Half hour"), six at [locale-parity.spec.ts:149](__tests__/i18n/locale-parity.spec.ts:149). They will drift. Keep one, ideally as the `"en"` entry in the ledger JSON (Findings 3–4) so English is just another row.

2. **A commit SHA in a source comment.** [locale-parity.spec.ts:161](__tests__/i18n/locale-parity.spec.ts:161) says "Seeded at `ef449120d`". `main` is squash-merged (every recent commit carries a `(#NNNN)`), so that SHA will not exist there. Use the PR number, or drop it.

3. **84 hand-typed foreign words in a spell-checked path.** `retired-tier-copy.json` lives under `__tests__/i18n/`, outside the `typos.toml` exclusion for `app/i18n/raw-i18n`. It is clean today (verified), but every future generation appends 84 more words. If the ledger stays in any form, add the file to `extend-exclude` now rather than after the first CI red.

4. **The marker comment points at the wrong files.** [en/index.ts:2481](app/i18n/en/index.ts:2481) attributes the 10m / 4h / 24h windows to `use-onchain-fee-tier-options.ts`; they live in `use-custodial-onchain-fee-tiers.ts` (`CUSTODIAL_PAYOUT_ETA_MINUTES`). It attributes 10m / 30m / 60m to the refund screen alone; the self-custodial *send* rail uses the same table (`use-onchain-fee-tiers.ts` → `FEE_TIER_ETA_MINUTES` in `app/types/payment.ts`) through the same label map. Name the two constants rather than the screens — constants are what a rename of the copy actually has to stay consistent with, and they do not move when a screen does.

5. **Unescaped interpolation into `new RegExp`** ([fee-tier-labels.spec.ts:47](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:47)). Harmless for the current list; a future entry with `(` or `.` either matches the wrong thing or throws at describe time.

6. **The ledger's comment block describes the invariant it approximates, not the one it enforces.** Sixty lines across two files explain "when `en`'s tier copy changes, every locale's changes with it"; what the test checks is "a locale's value is not one it already retired". Whatever design lands after Finding 3, write the comment for the check that runs.

7. **`enBase as unknown as Pick<Translations, …>`** ([fee-tier-labels.spec.ts:17](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:17)). The sibling i18n specs (`brand-name.spec.ts`, `limit-copy.spec.ts`) import `en` and read it directly. The double cast is explained and it is a test file, so this is cosmetic — but if the reads need to be typed, `en.SendBitcoinScreen.fast as string` at the three use sites is less machinery than narrowing the whole tree.

designsats:
@kngako should be ready to go

kngako (dismissed):
Yo @designsats gonna approve as there's no blocking findings. But please prepare a follow up PR for the other findings.

---
Partly — it addresses the one **blocking** finding, fully and exactly as proposed, and nothing else.

**What `6efacf4` fixes — Finding 1 [BLOCKING]:** the hook comment now reads "`payoutSpeeds` displayNames in galoy-values: Priority ~10 minutes, Standard ~4 hours and Flexible ~24 hours. This app shows the third as Economy to match the Fee rates screen (#4228)", and "Flexible" is dropped from both `SUPERSEDED_TIER_NAMES` lists. I checked out the three files at that commit: prettier and `eslint --no-ignore` clean, `locale-parity` + `fee-tier-labels` green (116 tests). The branch is no longer blocked.

**Still open:**
- **Finding 2** — the 14 tier words in 8 locales (`ar`, `de`, `el`, `id`, `sr`, `lg`, `qu`, `xh`) that disagree between Settings → Fee rates and the send selector. This is the one I'd still push for in this PR: it's 14 edits in files the branch already touches, and it goes to the PR's own purpose.
- **Finding 3** — the ledger still can't see an `en`-only rename (no `"en"` generation entry).
- **Finding 4** — only partly: "Flexible" is gone, but "Fastest"/"Normal" remain and the list is still compared against locale values, so the Spanish-"Normal" false positive still reproduces.
- **Findings 5–6**, all five test coverage gaps, all seven code smells — untouched. Smell 2 is slightly worse: the `ef449120d` SHA in the comment is now two commits back on a branch that will be squashed.

I appended a per-item status table to [[.idea/docs/review-fee-tier-rename-2026-09-14.md](https://claude.ai/epitaxy/.idea/docs/review-fee-tier-rename-2026-09-14.md)](.idea/docs/review-fee-tier-rename-2026-09-14.md). One note: the commit is on `fix/onchain-fee-tier-labels` (parented on the head I reviewed); this worktree's branch is still at `4c92c77cd`, and the tree is clean again.

designsats:
@kngako thanks for the approval. Finding 2 is fixed in 03ec3bb3b7, since it's in files this PR already touches.

The send selector and Settings → Fee rates now use the same word for each tier in all 28 locales:

- `ar`: Standard is عادي
- `el`: Standard is Κανονική (feminine, to match Settings' κανονικό)
- `id`: Economy is Ekonomis
- `sr`: Priority is Приоритетно
- `lg`, `qu`, `xh`: all three tiers now use the Settings words
- `de`: **Sparsam** on both screens, replacing "Economy" in Settings and "Ökonomisch" in the selector. It's the German this app already shipped for this tier. It would still be good for a German speaker to confirm it.

`locale-parity.spec.ts` gains a case that compares each locale's selector word with its Settings row. It ignores the "Onchain" marker and the time window, and Greek gender agreement is on an explicit allowlist. Restoring the old translations fails exactly those 8 locales. `locale-parity` and `fee-tier-labels` pass (144 tests).

Findings 3–6, the test coverage gaps and the code smells will go in a separate follow-up PR.

kngako (approved):
All the findings are expected in a future PR.


---

## Findings

### 1. The retired-copy ledger still cannot see an `en`-only rename — [NON-BLOCKING] *(carried: was Finding 3)*

Unchanged. Mutation A re-run at this head: rename `en` to Express / Regular / Saver, touch nothing else — `fee-tier-labels.spec.ts` fails its two English pins, **all 28 `locale-parity` cases stay green** (144 tests, 2 failed). The new cross-screen case does not help here, because it compares a locale with itself; nothing compares a locale with `en`. So the ledger at [locale-parity.spec.ts:164](__tests__/i18n/locale-parity.spec.ts:164) still only catches a skipped locale if the renamer already appended 28 × 3 outgoing strings — the discipline it was built to replace.

**When it comes up:** the next tweak to these three English words. Two red English pins, two arrays updated, green, shipped `en`-only — the regression `0df8394a8` fixed.

**Proposal:** As before: add an `"en": ["Priority", "Standard", "Economy"]` entry and assert it equals live `en` in `locale-parity`, so an `en`-only rename reddens the ledger with the ledger's own path in the failure. With the cross-screen case now in place, the leaner alternative from the first review — store only the English triple each locale was translated against and drop the foreign-word lists — is more attractive than it was, since the cross-screen case already catches a lazy "bump the line" fix whenever Settings and the selector fall out of step.

### 2. The global English ban list still rejects a correct translation, now with no other guard doing so — [NON-BLOCKING] *(carried: was Finding 4)*

`6efacf466` removed "Flexible"; "Fastest" and "Normal" remain at [locale-parity.spec.ts:149](__tests__/i18n/locale-parity.spec.ts:149), and neither was ever a value of the three keys the list is checked against (they came from the deleted `feeRate*` strings). Mutation B′, the fair version now that the cross-screen case exists: set `es.json` medium to "Normal" **and** its Settings row to "Onchain normal (~4h)" — a consistent, correct Spanish rendering of Standard. The cross-screen case passes; *carries no superseded on-chain fee tier name* fails. The ban list is the only thing rejecting it.

**When it comes up:** a Spanish, Portuguese, French, German, Catalan or Romanian translator picks "Normal" — the most natural word for Standard in all six — and is pushed to a worse one by a test whose message accuses them of leaving old copy behind. "Medium" in Danish/Dutch/Afrikaans and "Flexible" in Spanish/French are the same story.

**Proposal:** Drop the global list from the locale check and compare a locale only against its own ledger entries plus the English words `en` itself retired for these keys (`Fast`, `Medium`, `Slow`) — that keeps the af/da untranslated-fallback catch, which is the #3862 bug. Remove "Normal" and "Fastest" from both lists; remove "Half hour" from the labels-spec list (it was a key-name fragment, never copy).

### 3. The copy sweep still bans five adjectives across 73 unrelated strings — [NON-BLOCKING] *(carried: was Finding 5)*

Code unchanged apart from "Flexible" leaving the list ([fee-tier-labels.spec.ts:34](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:34), case at [:104](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:104)). Mutation C from the first review — a harmless "confirm fast; … can be slow" sentence in `SendBitcoinScreen` — still fails it.

**When it comes up:** any future on-chain copy in either section that uses "fast", "slow", "normal" or "medium" as an ordinary word.

**Proposal:** Scope the sweep to keys whose names mention a tier or a rate (`/feeRate|tier/i` plus the three tier keys), or replace it with the exact assertion it stands in for — `en.UnclaimedDeposit` has no `feeRate(Fastest|HalfHour|Hour)` key. Escape the names before interpolating into `new RegExp` ([:39](__tests__/screens/send-bitcoin-screen/fee-tier-labels.spec.ts:39)).

### 4. The heading/option word collision is still undocumented — [NON-BLOCKING] *(carried: was Finding 6)*

Unchanged: "Transaction priority" renders directly above the custodial default "Priority (…)" and the branch's own change-and-revert on this is recorded nowhere in the tree. Product call; make it once — a line in the PR description and, if the heading stays, one sentence in the marker comment at [en/index.ts:2480](app/i18n/en/index.ts:2480).

### 5. The cross-screen case is coupled to three hard-coded spellings of "Onchain" — [NON-BLOCKING, low] *(new)*

`ONCHAIN_MARKERS` at [locale-parity.spec.ts:211](__tests__/i18n/locale-parity.spec.ts:211) lists `I-Onchain`, `Onchain` and `オンチェーン`, and `withoutOnchainMarker` ([:215](__tests__/i18n/locale-parity.spec.ts:215)) strips only those, only from an end. Mutation F: change `cs.json`'s Settings row from "Onchain prioritní" to "On-chain prioritní" — same tier word, different marker spelling — and the case fails: `SendBitcoinScreen.fast "Prioritní" vs FeeRatesScreen.onchainPriority "On-chain prioritní (~10 min)"`. Today all 28 rows happen to use one of the three spellings, so it is green; a locale that translates "Onchain" natively (which several plausibly should — `cs`, `es`, `fr` all keep the English word today) will red this case for a copy change that does not touch the tier.

**When it comes up:** the next Settings → Fee rates copy pass in any locale. The failure is loud and prints both strings, so the reader can see the marker is the difference — this is a false positive, not a false pass, which is the better failure to have.

**Proposal:** Two cheap options. Keep the list but say so in the failure: when no marker matches, return a distinct message (`no Onchain marker recognised in "…"`) so the fix is obvious. Or drop the list: after removing the parenthetical, split the Settings label on whitespace and assert the selector word appears as a contiguous run of tokens — that handles `I-Onchain ephambili`, `أولوية Onchain`, and multi-word `Ey'omuwendo omutono`, and needs no spelling list. It does not handle `ja` (no whitespace), which would keep a one-entry exception. Either is fine; the current code is acceptable as long as its comment says the list must grow with new spellings.

---

## Test Coverage Gaps

1. **No test reproduces the `en`-only rename** (Mutation A). Unchanged; Finding 1.

2. **"is covered by the retired tier copy ledger" is satisfied by an empty entry.** [locale-parity.spec.ts:283](__tests__/i18n/locale-parity.spec.ts:283) still checks only `Object.keys(...).toContain(localeFile)`. Unchanged.

3. ~~No cross-screen vocabulary test.~~ **Resolved** by `03ec3bb3b` ([locale-parity.spec.ts:277](__tests__/i18n/locale-parity.spec.ts:277)); Mutations D and E confirm it catches drift in both directions.

4. **Ledger well-formedness is unchecked.** Unchanged.

5. **Two normalisers in one file.** The superseded check still uses `value.trim().toLowerCase()` ([:172](__tests__/i18n/locale-parity.spec.ts:172)) while the new case uses `toLocaleLowerCase(locale)` ([:223](__tests__/i18n/locale-parity.spec.ts:223)). Same file, same kind of string, two casing rules — a Turkish "İ" in a ledger entry would be handled differently by the two cases. Use the locale-aware one in both.

6. **The `TIER_WORD_AGREEMENT` allowlist has no self-check** ([:205](__tests__/i18n/locale-parity.spec.ts:205)). An entry keyed with a capital (`Κανονική`) silently never matches and the case goes red with a mismatch message that does not mention the allowlist. One `it` asserting every allowlist key equals its own `toLocaleLowerCase` — or lowercasing on lookup — closes it.

---

## Code Smells

1. **`SUPERSEDED_TIER_NAMES` is still defined twice** — five entries in `locale-parity`, six in `fee-tier-labels`. Unchanged.

2. **`ef449120d` is still cited in a comment** ([locale-parity.spec.ts:161](__tests__/i18n/locale-parity.spec.ts:161)); it is now three commits back on a branch that will be squashed.

3. **84 foreign words in a `typos`-checked path.** Unchanged; add `__tests__/i18n/retired-tier-copy.json` to `typos.toml` `extend-exclude` if the ledger stays.

4. **The marker comment points at the wrong files** for both ETA tables ([en/index.ts:2481](app/i18n/en/index.ts:2481)). Unchanged; name `CUSTODIAL_PAYOUT_ETA_MINUTES` and `FEE_TIER_ETA_MINUTES` instead of screens.

5. **Unescaped `new RegExp` interpolation.** Unchanged.

6. **The ledger comment still describes the invariant it approximates.** Unchanged.

7. **The double cast in `fee-tier-labels.spec.ts`.** Unchanged; cosmetic.

8. **`locale-parity.spec.ts` is becoming the home for every fee-tier copy invariant** *(new)*. Of its 288 lines, roughly 140 — the ban list, the ledger, the cross-screen pairs, the agreement allowlist, the marker strip — are about three keys. The sibling specs (`limit-copy`, `brand-name`, `map-distance-label`, `recovery-phrase-label`) show the house pattern: one spec per copy invariant. A `__tests__/i18n/fee-tier-copy.spec.ts` that reads the same 28 files would leave `locale-parity` to key and placeholder parity, which is what its name says.

9. **The CodeQL rewrite traded generality for scanner silence** *(new)*. `js/incomplete-multi-character-sanitization` fires on any global `.replace()` of a multi-character pattern; it exists for HTML/path sanitisers where a second pass could reveal a payload. A string comparison in a spec is not a sanitiser, and the slice at [:215](__tests__/i18n/locale-parity.spec.ts:215) is strictly less general than the replace it replaced (marker must sit at an end). Dismissing the alert as a false positive would have been an equally valid resolution. Not wrong as done — just worth knowing the alert did not point at a defect.

Issue #4268, opened 2026-09-11
Fix very long success/error/warning messages


designsats:
## Summary

Ten English toast messages wrap to **three lines** in the new notification toast,
and most translations of them to four or five. Shorten the copy so every toast
message fits in **two lines** on a 360pt-wide screen.

This is a **copy-only** ticket. The toast component does not change here.

## Why

#4267 moves the toast to the top of the screen and restyles it: bold 16px text,
and **no line limit**, so a message is never cut off. The old toast was 14px regular
text truncated at two lines, which hid how long these strings are. With the new row
they show in full, and a three-to-five-line toast covers the screen header and the
content right below it for its full 4 seconds.

There is room for about 38 bold characters per line on a 360pt screen (a 264pt text
column: 360 − 2×20 inset − 2×12 padding − 2×1 border − 18 icon − 12 gap).

## How this was measured

- Advance widths from `SourceSansPro-Bold.ttf` (the face the toast uses) at 16px,
  greedy word wrap, hard breaks on `\n`.
- Text width 264pt (360pt screen) and 316pt (412pt screen, e.g. Pixel 7a).
- Placeholders filled with realistic values: `{requiredFee}` → `1,234`,
  `{network}` → `Mainnet`, `{provider}` → `Google Drive`,
  `{identifier}` → `satoshi@blink.sv`.
- Locales: the 22 Latin-script translations in `app/i18n/raw-i18n/translations`.
  Arabic, Greek, Armenian, Japanese, Serbian and Thai were skipped because the Bold
  face has no glyphs for them and the OS substitutes a fallback font.

Estimates, not device captures. Real wrapping can differ by a word at a line edge.

## Outliers — three lines or more in English

| # | Type | Message (en) | Key | 360pt | 412pt | Locales at 3+ (of 22) | Worst locale |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | error | Network fees are too high to claim ({requiredFee} sats required). Try again later or refund to an on-chain address. | `UnclaimedDeposit.feeExceeded` | 3 | 3 | 22 | 5 (de, pt) |
| 2 | error | Unable to export transactions to csv. Something went wrong. If issue persists please contact support. | `SettingsScreen.csvTransactionsError` | 3 | 3 | 22 | 10 (qu) |
[...]

PR #4269, opened 2026-09-11
feat(toast): move notifications to the top and restyle them as a row

### LINKED ISSUE #4267: Change position and styling of notification toast
## Summary

Move the notification toast from the bottom of the screen to the top, and replace
its white card (coloured left bar, bold "Success" / "Error" / "Warning" title) with
a single-message row on a `grey7` surface: a coloured icon, bold message text, and a 1px border in
the same colour.

Change the shared `GaloyToast`, not the call sites. All 98 `toastShow()` call sites
pick it up unchanged.

## Why

At the bottom the toast is often covered by the user's hand, or by Android's own
system overlays near the navigation bar. Confirmations like "address copied" and
errors are easy to miss there.

## Design

[Figma screens ](https://www.figma.com/design/9MQuQi8ZhXVvDibWSI3C4c/Blink?m=auto&node-id=37-4624&t=UgVNSbSfo5qycvfM-1)

Mockups are dark mode only. Light mode is derived from the tokens below, not drawn.

| Success | Error |
| --- | --- |
| Image | Image |

The mockup's error copy ("Internet connection issue") is illustrative; message
strings do not change. The real string, `LL.errors.network.connection()`, is
"Connection issue.\nVerify your internet connection", so it renders on two lines.

## Position

| Property | Value |
| --- | --- |
| `position` | `"top"` (was `"bottom"`, `bottomOffset: 80`) |
| `topOffset` | `useSafeAreaInsets().top + 40` |
| Horizontal inset | 20 each side, row stretches between (`TOAST_HORIZONTAL_INSET`) |

Set `topOffset` as a prop on `<Toast>` inside `GaloyToast`, not in `Toast.show()`.
The toast is mounted four times (`app.tsx`, plus the invite, May challenge and June
challenge modals), and the prop keeps them all in step from one place.

The offset is relative to the safe area so it clears the status bar on Android and
the notch / Dynamic Island on iOS. A fixed number would sit under the island on
recent iPhones.

## Geometry

| Element | Value |
| --- | --- |
| Row | `flexDirection: "row"`, `alignItems: "center"`, `alignSelf: "stretch"` |
| Padding | 14 vertical, 12 horizontal |
| Border | 1px, per-variant colour (below) |
| Corner radius | 8 |
| Icon | `GaloyIcon`, size 18 |
| Icon → text gap | 12 |
| Text | 16 / line height 22, no line limit, `flex: 1` |
| Height | 52 for one line (14 + 22 + 14 + 2), +22 per extra line |

## Colours — themed tokens only

| Role | Token | Dark | Light |
| --- | --- | --- | --- |
| Background (all variants) | `colors.grey7` | `#0F0F0F` | `#F9F9F9` |
| Message text | `colors.black` | `#FFFFFF` | `#000000` |
| Success border + icon | `colors._green` | `#00A700` | `#00A700` |
| Error border + icon | `colors.red` | `#DC2626` | `#DC2626` |
| Warning border + icon | `colors.warning` | `#FFC563` | `#E18E02` |

The border and icon share one accent colour per variant. No filled backgrounds.

## Variants

| `type` | Icon | Accent |
| --- | --- | --- |
| `success` | `check-circle` | `_green` |
| `error` (default) | `warning-circle` | `red` |
| `warning` | `warning` | `warning` |

Error is the negative notification ("Error loading…", network failures, etc.).
It is the default `type`, so roughly 70 of the 98 call sites use it without saying so.

## Typography — always bold

The message is **always** bold, not a caller option. There is no title line: the
icon and colour carry the type.

Set the Bold face directly, with no `fontWeight`:

```ts
fontFamily: "SourceSansPro-Bold",
fontSize: 16,
lineHeight: 22,
includeFontPadding: false,
```

### Two things that will bite

1. **`<Text type="p2" bold>` is a fake bold on Android.** The theme sets
   `fontFamily: "SourceSansPro-Regular"` with `fontWeight: "600"`, and Android
   synthesises that weight from the Regular file rather than loading a real face.
2. **`fontFamily: "Source Sans Pro"` renders Roboto on Android.**
   `ReactFontManager` resolves the family to `assets/fonts/<name>{_bold}.ttf`, and
   no file has that name, so it falls back to `Typeface.create`, which is the system
   font. It works on iOS, which is why it looks fine there. `"SourceSansPro-Bold"`
   matches the asset file on Android and the PostScript name on iOS.

## Handoff notes

- **Files:** `app/components/galoy-toast/galoy-toast.tsx` (row, position, styles)
  and `app/utils/toast.ts` (`position: "top"`, drop `bottomOffset`).
- Replace the library's `SuccessToast` / `ErrorToast` renderers in `toastConfig`
  with one `ToastRow` that takes `type` and `text2`.
- Keep passing `text1` (the translated title) to `Toast.show()`. The UI no longer
  renders it, but the existing `toast.spec.ts` assertions read it, and leaving it
  keeps `toast.ts` to a position-only diff. Analytics (`logToastShown`) logs the
  English message and type, not `text1`, so it is unaffected.
- Style with `makeStyles` / `useTheme`. The current file imports `light` directly,
  so today the toast ignores dark mode entirely.
- `GaloyToast` in `app.tsx` already sits inside `SafeAreaProvider` and
  `GaloyThemeProvider`, so the hooks work without new providers.
- Don't set `numberOfLines`. At bold 16px only ~38 characters fit per line on a
  360pt screen, so 10 English messages wrap to three lines (CSV export, network
  fees too high, biometrics not enrolled…), plus the `StatusCode` network dump.
  The old toast's `text2NumberOfLines={2}` would cut them off. Shortening them is
  #4268, not this ticket.
- The existing FIXME in `toast.ts` still applies: a toast fired from `app.tsx` can
  be hidden behind an open RN `Modal`. Moving it to the top does not change that.

## Out of scope — follow-up PRs

**Message copy.** This ticket changes no text: no strings, no translation keys, no
locale files. Messages that run past two lines in the new row are tracked in #4268.

**`GaloyInfo` boxes.** The same row should replace the inline `GaloyInfo` boxes (left-bar style) on
10 screens: phone login input/validation, phone registration input/validation,
Telegram login validation, theme, default wallet, transaction detail, and the map
place sheet. Tracked separately to keep this diff small.

## Acceptance requirements

- [ ] Toasts appear at the top: safe-area top inset + 40, on Android and iOS
- [ ] On iOS the toast clears the notch / Dynamic Island
- [ ] Every toast sits 20 from both screen edges, including inside modals
- [ ] Success, error and warning each render as a single row with their own icon,
      1px border and icon colour, on a `grey7` background
- [ ] No "Success" / "Error" / "Warning" title line
- [ ] Message text is Source Sans Pro **Bold** on both platforms (not Roboto, not
      synthesised)
- [ ] Messages show in full at any length — the row grows, nothing is truncated
- [ ] Row follows the theme in both dark and light mode
- [ ] Toasts shown inside the invite and May/June challenge modals use the same
      position and style
- [ ] `logToastShown` still receives the English message and type
- [ ] No changes under `app/i18n/`

## Testing specs

Automated:

| Spec | Covers |
| --- | --- |
| `__tests__/utils/toast.spec.ts` | `position: "top"`, no `bottomOffset` |
| `__tests__/components/galoy-toast.spec.tsx` | `topOffset` = inset + 40; per variant: row renders message without title and with no `numberOfLines`, `grey7` background, 1px accent border, 20 horizontal inset, 14 vertical padding, `SourceSansPro-Bold` with no `fontWeight` |

```
yarn jest __tests__/components/galoy-toast.spec.tsx __tests__/utils/toast.spec.ts --maxWorkers=2
```

Each spec must fail against the current `GaloyToast` / `toastShow`.

> The icon colour is not asserted in a spec. Cover it in manual QA.

Manual QA, Android and iOS, dark and light:

1. Receive → **Copy** (any `useClipboard` copy): a success toast appears at the
   top, below the status bar, green border and icon, bold message.
2. Restore a self-custodial wallet with an invalid recovery phrase
   (`RestoreScreen.invalidMnemonic`): an error toast appears, red border and icon.
3. Trigger a warning toast, e.g. a card screen whose data query fails
   (`use-statements-data.ts`): orange border and icon.
4. Trigger a long message (e.g. Settings → export CSV failing,
   `SettingsScreen.csvTransactionsError`, ~3 lines): the row grows to show the whole
   text, and the icon stays vertically centred.
5. On an iPhone with a Dynamic Island, confirm the row sits clear of it.
6. Open the invite modal and tap its copy action: the toast inside the modal
   matches the app-level one.
7. Switch the theme between dark and light and repeat step 1.

designsats:
## Summary

Moves the notification toast from the bottom of the screen to the top, and restyles it
from the library's white card into a themed row: an accent-coloured icon, an always-bold
message, and a 1px border in the same accent on a `grey7` surface.

At the bottom the toast was often covered by the user's hand, or by Android's own
overlays near the navigation bar.

Closes #4267

| Commit | What it does |
| --- | --- |
| `48dad5cf6` | `toastShow` shows toasts at `position: "top"` and drops `bottomOffset: 80` |
| `843f2630e` | `GaloyToast` renders the new row and sets `topOffset` to the safe-area inset + 40 |
| `f33a8f72e` | One `TOAST_HORIZONTAL_INSET = 20` sets the row's horizontal margin for every toast |

Only 4 files change: `app/utils/toast.ts`, `app/components/galoy-toast/galoy-toast.tsx`
and their two specs. No call site, string or translation changes.

### Implementation notes

- **Offset is on `<Toast>`, not in `Toast.show()`.** `GaloyToast` is mounted four times
  (`app.tsx`, plus the invite, May challenge and June challenge modals); setting
  `topOffset={insets.top + TOAST_TOP_OFFSET}` on the component keeps all of them in step
  from one place, and clears the status bar on Android and the notch / Dynamic Island on
  iOS. The 20pt horizontal inset is the same kind of single constant,
  `TOAST_HORIZONTAL_INSET`, so every toast in the app sits 20 from both screen edges.
- **One `ToastRow` for all three types.** Success, error and warning share the layout and
  differ only in icon and accent: `check-circle` / `_green`, `warning-circle` / `red`,
  `warning` / `warning`. The row renders `text2` only; `toastShow` still passes the
  translated title as `text1`, which is no longer drawn.
- **Bold is the font file, not a weight.** The message sets
  `fontFamily: "SourceSansPro-Bold"` with no `fontWeight`. Android's `ReactFontManager`
  resolves `fontFamily` to `assets/fonts/<name>{_bold}.ttf`, so `"Source Sans Pro"` falls
  back to Roboto there. The theme's `<Text bold>` gives `SourceSansPro-Regular` plus
  weight 600, which Android fakes by synthesising the weight.
- **No line limit.** The old config set `text2NumberOfLines={2}`; the row sets no
  `numberOfLines`, so long messages show in full and the row grows instead.
- **Themed colours.** The old config imported the `light` palette directly, so the toast
  ignored dark mode. The row reads everything from `useTheme` / `makeStyles`.

### Out of scope

- **Message copy** — #4268. Ten English messages wrap to three lines in the new row
  (CSV export, network fees too high, biometrics not enrolled…). This PR changes no text.
- **Inline `GaloyInfo` boxes** on 10 screens move to the same row in a follow-up PR.
- **Enter/exit animation** is unchanged: the library's default spring.

## Acceptance requirements

- [ ] Toasts appear at the top: safe-area top inset + 40, on Android and iOS
- [ ] On iOS the toast clears the notch / Dynamic Island
- [ ] Every toast sits 20 from both screen edges, including inside modals
- [ ] Success, error and warning each render as a single row with their own icon, a 1px
      border and icon in the accent colour, on a `grey7` background
- [ ] No "Success" / "Error" / "Warning" title line
- [ ] Message text is Source Sans Pro **Bold** on both platforms (not Roboto, not faked)
- [ ] Messages show in full at any length; the row grows, nothing is cut off
- [ ] The row follows the theme in dark and light mode
- [ ] Toasts inside the invite and May/June challenge modals use the same position and
      style
- [ ] No changes under `app/i18n/`

## How it should look and behave

[toast.webm](https://github.com/user-attachments/assets/41f79e16-77b4-48a5-9209-457df242fcd5)

image

image

## Testing specs

Automated:

| Spec | Covers |
| --- | --- |
| `__tests__/utils/toast.spec.ts` | `position: "top"`, no `bottomOffset` |
| `__tests__/components/galoy-toast.spec.tsx` | `topOffset` = inset + 40 at insets 0 / 24 / 59; renderers registered for all three types; per type: message without title, no `numberOfLines`, icon name + size 18 + accent colour, row geometry (row, gap 12, horizontal inset 20, padding 12/14, radius 8), 1px accent border on `grey7`, `SourceSansPro-Bold` 16/22 with no `fontWeight`, message in `colors.black` not the accent, light theme |

```
yarn jest __tests__/components/galoy-toast.spec.tsx __tests__/utils/toast.spec.ts --maxWorkers=2
```

The 26 component tests were mutation-checked: a fixed `topOffset={64}`, a white error
icon, accent-coloured text, a hard-coded background, and the theme's Regular face plus
weight 600 each fail 2–6 tests. All 56 suites that touch `toastShow`, `GaloyToast` or
`react-native-toast-message` pass (872 tests).

Reviewed on the Android emulator (`Medium_Phone_API_36.1`, debug build of `f33a8f72e`). **iOS is not yet verified.**

Manual QA, Android and iOS, dark and light:

- [x] Receive → **Copy**: a success toast appears at the top, below the status bar, with
      a green border and icon and a bold message
- [ ] Restore a self-custodial wallet with an invalid recovery phrase
      (`RestoreScreen.invalidMnemonic`): an error toast with a red border and icon
- [x] A card screen whose data query fails (e.g. statements): a warning toast with an
      orange border and icon
- [ ] A long message (e.g. `SettingsScreen.csvTransactionsError`, about 3 lines) shows in
      full, and the icon stays vertically centred
- [ ] On an iPhone with a Dynamic Island, the row sits clear of it
- [ ] Invite modal → copy: the toast inside the modal matches the app-level one
- [ ] Switch the theme between dark and light and repeat the first check

## Notes / behavior changes

Existing users will see, after updating:

- Toasts come in from the top of the screen instead of the bottom, and cover the top of
  the header while they show.
- There is no "Success" / "Error" / "Warning" title anymore; the icon and colour carry
  the type.
- In dark mode the toast is now dark (`grey7`) rather than a white card.
- Long messages that used to be cut off after two lines now show in full, so some toasts
  are three lines tall until #4268 shortens them.


designsats:
@kngako can you review? Especially on iOS

kngako (changes requested):
Add test coverage gaps and this is ready to go. 

---

**Overall assessment:** The change does what #4267 asks, in the shape it asks for, and the specs pin it tightly enough that five of the eight regressions I tried fail loudly. **No finding is blocking.** The three findings are: the modal mounts do not get the geometry the commit message says they get; the translated title is still built and sent on every show but the row throws it away, which costs screen-reader users the toast type; and commit 1 on its own ships a toast under the Dynamic Island.

---

## Findings

### Finding 1 — Inside the invite and challenge modals the toast is neither 20 from the edges nor `inset + 40` from the top [NON-BLOCKING]

**When it comes up:** Open the invite modal and tap copy: `useClipboard` fires a success toast that renders through the modal's own `<GaloyToast />`, whose nearest native parent is `react-native-modal`'s content view, and that view carries `margin: deviceWidth * 0.05` (`react-native-modal@13.0.2`, `dist/modal.js:508`), so on a 393pt iPhone the row sits about 40pt from each edge and about 20pt lower than the app-level toast. On Android the RN `Modal` window also starts below the status bar (`statusBarTranslucent` defaults to false, which sets `fitsSystemWindows` — `ReactModalHostView.kt:357`) while `useSafeAreaInsets().top` still reports the edge-to-edge root's inset, so on Android 15+ the row lands a status-bar height plus that margin lower than at the root. Commit 3's "every GaloyToast mount … sits 20 from each edge" and the issue's "including inside modals" acceptance line are not met.

**Proposal:** Give the three modals `style={styles.modal}` with `margin: 0` (react-native-modal applies `style` to exactly that content view) and move the 5% gutter onto `modalCard` as `marginHorizontal`, then pass `statusBarTranslucent` so the Android modal window shares the root's frame — which also dims the status bar behind the modal, something it does not do today. If that is left for a follow-up, correct commit 3's message now so history does not claim a geometry the modals do not have, and leave the issue's checkbox open. A layout assertion is not feasible in RNTL, so this stays on manual QA step 6 either way.

### Finding 2 — The translated title is still built and sent on every show, but the row discards it, so screen readers lose the toast type and an empty message yields a blank row [NON-BLOCKING]

**When it comes up:** A VoiceOver or TalkBack user copies an address. The old card's root was a `TouchableOpacity` (accessible by default, `TouchableOpacity.js:303`) that read "Success, Address copied" as one element; the new root is a plain `View` whose only readable child is the message, and `GaloyIcon` carries no label, so once colour is taken away the one thing that separates an error from a success is never announced. The same gap shows at the eleven call sites that pass `error.message` / `err.message` straight through (`all-contacts.tsx:97`, `contacts-card.tsx:131`, `use-card-freeze.ts:46`, …): an `Error` with an empty message used to at least show the "Error" title, now it shows a red-bordered row with an icon and nothing else.

**Proposal:** Pass `text1` through to `ToastRow` and spend it where the eye no longer needs it: `accessible accessibilityRole="alert" accessibilityLabel={[text1, text2].filter(Boolean).join(". ")}` on the row (which also restores the single-element grouping the touchable gave), and `{text2 || text1}` as the visible text so an empty message still says "Error". That gives the `text1` pins in `toast.spec.ts:41,55,69` something real to guard; add one assertion per variant that the row's label starts with the title. `toastShow` already computes all three titles per show (`toast.ts:26-30`), so nothing new is built.

### Finding 3 — Commit 1 on its own ships the toast under the notch [NON-BLOCKING — history hygiene]

**When it comes up:** At `48dad5cf6`, `toastShow` already says `position: "top"` while `GaloyToast` still passes no `topOffset`, so the library default of 40 applies and on any iPhone with a notch or Dynamic Island the toast draws into the status-bar area; the message says the safe-area offset "is applied by GaloyToast", which only becomes true one commit later. Both toast suites are green there, so a bisect for "toast under the island" stops at a commit the branch never meant to ship alone.

**Proposal:** Squash `48dad5cf6` into `843f2630e` (`git rebase -i 15f907695`, mark the second commit `squash`), keeping commit 2's message with commit 1's "why" paragraph folded in, and fold commit 3's 14→20 correction in as well, since it adjusts a value commit 2 introduced — three commits become one that is true at every point. If the author prefers to keep the move and the restyle separate, reorder so the restyle lands first and every commit in the range is shippable.

---

## Test Coverage Gaps

1. **The spec renders under a theme without the production `Text` config, so "no synthesised weight" is asserted against a theme that never sets one.** `ContextForScreenWithTheme` (`__tests__/screens/helper.tsx:55`) builds `createTheme({ lightColors, darkColors, mode })` and drops `theme.components`; `galoy-toast.spec.tsx:156` then checks `fontWeight` is undefined. Reproduced: adding `type="p2"` to the toast `<Text>` — the obvious "use the theme's scale" edit — leaves all 26 tests green, while under the real theme the style becomes `fontFamily: "SourceSansPro-Bold", fontWeight: "400"`, which iOS resolves to the *Regular* face (closest weight in the family) and Android keeps Bold (weight < 700 still maps to the unstyled file): the platforms silently diverge. Build the test theme from the real one — `createTheme({ ...theme, mode })` in `createThemeWithMode` (six specs use it; re-run them) or locally in this spec — and that mutation fails on the `fontWeight` line.
2. **`flex: 1` on the message is not pinned.** Removing it keeps 26/26 green; without it a message longer than one line does not wrap inside the row but runs past the border (the row is `flexDirection: "row"` and the text has nothing to shrink against). The "shows the whole message without a line limit" test at `:114` checks only `numberOfLines`; add `flex: 1` to the font `toMatchObject` at `:151` or a dedicated "wraps inside the row" assertion, since wrapping is the property "no line limit" depends on.
3. **`includeFontPadding: false` is not pinned.** Removing it is invisible to the spec; it is the Android fix that keeps the glyphs centred on the icon (the same line the warning card documents at `warning-card.tsx:125`). One more key in the existing `toMatchObject`.
4. **`toast.spec.ts` pins a payload field nothing consumes.** The three "correct title" tests (`:33`, `:47`, `:61`) assert `text1: "Error" / "Success" / "Warning"` on `Toast.show`, but no renderer reads `text1` any more, so the tests describe a title that does not exist on screen. Either give `text1` a consumer (Finding 2) or rename the tests to say what they now guard (that the type is forwarded).
5. **Nothing exercises the modal mounts.** The one behaviour the branch cannot pin in jest is the one that is off (Finding 1). The modals have no specs today; if one is added, render the modal open and assert a `toast` test id is present under it, so a future removal of the per-modal mount — which would leave the toast hidden behind the modal, per the FIXME in `toast.ts:38` — is noticed.

## Code Smells

1. **`ToastType` is declared twice.** `galoy-toast.tsx:13` and `toast.ts:18` (`type?: "error" | "success" | "warning"`) spell the same union independently; a fourth type added to one and not the other compiles until a toast is shown. Export the union from one place — the component is the natural owner since it holds the icon map — and reuse it in `toastShow`.
2. **The three config entries are copy-paste.** `:41-51` differ only in the literal, so adding a type means touching four places (`ToastType`, `icons`, `accent`, `toastConfig`). Deriving them from the `icons` keys (`Object.fromEntries((Object.keys(icons) as ToastType[]).map(...))`) or accepting the duplication are both fine; optional.
3. **`accent` is the one style that is not in `makeStyles`.** `:27-31` build a lookup on every render and splice `borderColor` inline at `:34`; `makeStyles` takes props (`useStyles({ accent })`, as `slide-up-handle/index.tsx:140` does), which keeps every style in one place and drops the inline array.
4. **Two exported constants with no importer.** `TOAST_TOP_OFFSET` and `TOAST_HORIZONTAL_INSET` are `export`ed (`:10-11`) but nothing imports them, and the spec deliberately pins the literals 40 and 20 — which is right, importing the constant would make the pin tautological. Drop `export` unless another mount is meant to read them.
5. **The spec reaches the theme through the full screen context.** `renderToast` (`galoy-toast.spec.tsx:51`) wraps a two-element row in `NavigationContainer` + `MockedProvider` + `AccountRegistryProvider` to obtain a `ThemeProvider`, which is where the `act()` warnings in the run come from (`use-account-registry.tsx:109-122`). A bare `<ThemeProvider theme={createTheme({ ...theme, mode })}>` is enough, and fixes Gap 1 in the same move.
6. **`config` is captured through a module-level mutable filled by the mock.** `mockToastProps` is populated by rendering `<GaloyToast />` and then read to call the renderer by hand (`:51-57`). Exporting `toastConfig` (or `ToastRow`) from the component and rendering it directly says the same thing without the side channel — a trade-off between test ergonomics and a public surface, not a defect.
7. **The font comment is right, and contradicts its neighbour.** `galoy-toast.tsx:73-75` correctly describes RN's Android resolution; `warning-card.tsx:114-119` documents the opposite rule and uses `fontFamily: "Source Sans Pro"` + `fontWeight: "700"`, which by the same `ReactFontManager` path (no `fonts/Source Sans Pro_bold.ttf`, no `addCustomFont` registration) falls back to the system typeface on Android — as do the 64 `fontFamily: "Source Sans Pro"` declarations across 31 files, mostly under `card-screen/`. Not this branch's bug, but this branch is now the precedent a follow-up should point at.

designsats:
@kngako should be rdy to ship

PR #3862, opened 2026-06-16
feat(balances): rename "Account" to "Balance" and rework Default Balance screen

designsats:
## Summary

This PR does two related things:

1. **Renames "Account / Accounts" wording to "Balance / Balances"** across the app's English copy, to align the UI with the "Balance" terminology.
2. **Reworks the Default Balance screen** so the description text reflects the selected balance, and disables the Dollar default for non-custodial accounts.

All 28 locale files are updated for the new i18n keys.

---

## 1. "Account" → "Balance" terminology rename

Updated English strings throughout (`app/i18n/en/index.ts`, regenerated `i18n-types.ts` and `raw-i18n/source/en.json`), e.g.:

- `My Accounts` → `Balances`
- `Default Account` → `Default Balance`
- `Bitcoin Account` / `Dollar Account` → `Bitcoin Balance` / `Dollar Balance`
- `Receiving Account` / `Sending Account` → `Receiving Balance` / `Sending Balance`
- Stablesats modal, transfer history, and region-restriction copy updated to match.

## 2. Default Balance screen rework

**Copy is now selection-aware** (`app/screens/settings-screen/default-wallet.tsx`). The previous single combined paragraph was split so each selection shows its own description:

- `infoBtc` — *"Your Bitcoin Balance is for stacking sats, where the value moves with the market."*
- `infoUsd` — *"Your Dollar Balance is powered by Stablesats — it holds Bitcoin that is hedged to stay stable in USD terms. You're always holding Bitcoin, but the USD value doesn't change."*
- `infoUsdSelfCustodial` — *"Due to user experience concerns we decided to not activate this feature in the non-custodial accounts. Only Bitcoin is available as your default receive setting."*

**Behavior:**
- **Custodial:** info switches between `infoBtc` / `infoUsd` based on the selected wallet.
- **Self-custodial:** the Dollar option is rendered **disabled** and can no longer be set as default. Tapping it still surfaces its description via a separate `descriptionCurrency` state, while the actual default currency stays on Bitcoin.

**Navigation:** the screen header now uses `DefaultWalletScreen.title` (*"Default Balance"*) instead of `SettingsScreen.receiveCurrency` (*"Receive currency"*), matching the settings-list row in both modes.

**i18n cleanup:** removed the now-unused `DefaultWalletScreen.info` / `infoSelfCustodial` keys.

## 3. Translations

[...]

PR #4270, opened 2026-09-14
fix(i18n): call an outgoing payment "You sent" instead of "You spent"

designsats:
## Problem

The transaction detail screen headed every outgoing payment "You spent". Not every send is a spend: a transfer to yourself moves money without paying anyone.

Refs blinkbitcoin/blink-wip#426

## Fix

| Commit | What it does |
| --- | --- |
| `34ccabe84` | Renames `TransactionDetailScreen.spent` → `sent` ("You sent") in `en` and all 28 locales, regenerates `i18n-types.ts`, updates the screen, the heading spec and the 5 detox assertions. Replaces the three hand-typed `LL` stubs (row-actions, fallback, seen-state) with a shared Proxy-backed `mockI18n()` in `transaction-detail-mocks.tsx` |
| `99eddb633` | Pins the heading's copy in every locale, and adds heading cases for a self-transfer and a broadcast on-chain send |

- **The key is renamed, not just the value.** A key called `spent` holding "You sent" would mislead the next reader.
- **Every locale gets a real translation.** Locales are `merge({}, en, raw)`, so a changed `en` alone leaves all 28 locales on the old wording. Each locale now uses the verb it already uses for "send" elsewhere in the app (e.g. es *Enviaste*, de *Du sendetest*, fr *Vous avez envoyé*, lg *Osindise*, qu *Apachirqanki*).
- **ja is unchanged.** 送金完了 ("transfer complete") never said "spent".
- **No more hand-typed `LL` stubs in the transaction detail specs.** `mockI18n()` returns `""` for any key unless the spec names it, so renaming a key can't leave a stale stub that crashes with "sent is not a function".
- The comment above the heading no longer names the longest locale. It only describes how the heading works (`numberOfLines={1} adjustsFontSizeToFit`), since a named example goes stale the next time a string changes.

## Acceptance requirements

- [ ] A completed outgoing payment (Lightning, on-chain, intraledger) is headed "You sent" in English
- [ ] A queued on-chain send still reads "Sending"; a receive still reads "You received"
- [ ] In any non-English locale except ja, the heading uses a "send" verb, not a "spend" verb
- [ ] In ja the heading still reads 送金完了
- [ ] The heading stays on one line in every locale

## Testing specs

| Spec | Covers |
| --- | --- |
| `__tests__/i18n/send-heading-copy.spec.ts` | Each shipped locale's `sent`, pinned through `i18nObject` (the same merge over `en` the app renders from), `ja` included. It is typed `Record<Locales, string>`, so a new locale won't compile until it's pinned. `id` and `xh` are translated but not in `Locales`, so they're pinned on the raw JSON |
| `__tests__/screens/transaction-detail-heading.spec.tsx` | Heading copy and single-line layout for a Lightning send, an intraledger self-transfer, a receive, a queued on-chain send ("Sending") and a broadcast on-chain send ("You sent") |
| `__tests__/screens/transaction-detail-{row-actions,fallback,seen-state}.spec.tsx` | Unchanged assertions, now on the shared `mockI18n()` |
| `__tests__/i18n/locale-parity.spec.ts` | Every locale carries the renamed `sent` key |

```
yarn jest __tests__/i18n/send-heading-copy.spec.ts __tests__/screens/transaction-detail- __tests__/i18n/locale-parity.spec.ts
```

Setting ca's `sent` to a spend verb ("Vas gastar") fails the locale pin. So does reverting it, replacing it with English, or deleting the key.

Manual QA:
1. Send a payment, open it from the transaction list → heading reads "You sent".
2. Settings → Language → Español → reopen it → "Enviaste".
3. Switch to 日本語 → still 送金完了.

> Translations were not reviewed by native speakers; lg, qu, xh and hy are the least certain.

## Notes / behavior changes

Users in every locale except ja will see new wording on the transaction detail heading for outgoing payments.

**Follow-up (future ticket, not this PR):** the heading ignores payment status. It checks direction and on-chain broadcast state only, so a Lightning send with `status: "FAILURE"` or `"PENDING"` also reads "You sent". A failed payment shows nothing else on the screen saying it failed. This was just as wrong with "You spent"; the rename doesn't cause it or fix it. The fix is to check `status` before `direction`, add a `TransactionDetailScreen.failed` key in all 28 locales, and add `FAILURE`/`PENDING` rows to the heading spec. If past tense is fine for a pending payment, add a test case that pins that decision.


kngako (changes requested):
Finding 3 is quite interesting. Should the message still say "You sent" for failed transaction?


---

## 1. Findings

### F1 — Two more `LL` stubs still key `spent`

`edc8d4958` fixes the row-actions spec's hand-rolled `useI18nContext` mock, which crashed
with "sent is not a function" once the screen called `LL.TransactionDetailScreen.sent()`.
Two sibling specs carry the identical stale stub and were not touched:
[transaction-detail-fallback.spec.tsx:100](__tests__/screens/transaction-detail-fallback.spec.tsx:100)
and
[transaction-detail-seen-state.spec.tsx:157](__tests__/screens/transaction-detail-seen-state.spec.tsx:157).
They are green today only by accident of their fixtures: the fallback spec renders with
`data: {}` (the resolver fallback returns before the heading is built), and the seen-state
spec has zero `direction: "SEND"` fixtures, so the heading always takes the `received()`
branch. The first person to add a send case to either — "marks a sent transaction as
seen" is the obvious next seen-state test — hits the exact crash this branch just
debugged in row-actions, and will spend time working out why a mock that lists
`TransactionDetailScreen` keys is missing the one the screen wants.

**Proposal.** Rename both stubs to `sent` in this PR, and fold them together with
`edc8d4958` into `d38657551`, so the rename commit is self-contained and no commit on the
stack leaves a suite red (see [S3]). The durable fix is [S1]: stop hand-typing the `LL`
shape in each spec. Either use the real `en` bundle, as
[transaction-detail-heading.spec.tsx:72](__tests__/screens/transaction-detail-heading.spec.tsx:72)
already does, or add a `mockI18n()` to
[transaction-detail-mocks.tsx](__tests__/helpers/transaction-detail-mocks.tsx), which
already exports every other mock these three specs share. A renamed key cannot go stale in
a stub that is not typed by hand.

### F2 — The locale guard only recognises the verbatim retired string

`1c424a075` adds a case per locale asserting
`parsed.TransactionDetailScreen.sent !== <the old "You spent" wording>`
([transaction-detail-heading.spec.tsx:213](__tests__/screens/transaction-detail-heading.spec.tsx:213)).
The claim in its commit message — "fail if it comes back" — holds: reverting `ca.json` to
`"Has gastat"` turns it red. But its *name*, "no longer calls a send a spend", promises more
than that. Set `ca.json`'s `sent` to `"Vas gastar"` (still a spend) and it stays green; set
it to `"You sent"` — the untranslated-English fallback that the i18n checklist singles out
([pr-review.md §3 i18n](docs/pr-review.md:317), "no stale English fallbacks left in
individual locales, #3862") — and it stays green too. It is a tripwire for one exact string
per locale, carried forever in a 27-entry tombstone table, plus a comment explaining why
`ja` is not in it.

**Proposal.** Use the positive pin the repo already uses for copy it cares about
([footer-headline.spec.ts](__tests__/i18n/footer-headline.spec.ts),
[recovery-phrase-label.spec.ts](__tests__/i18n/recovery-phrase-label.spec.ts)): a
`Record<Locales, string>` of the 28 *new* values and, per locale,
`loadLocale(l); expect(i18nObject(l).TransactionDetailScreen.sent()).toBe(expected[l])`.
Verified on `ca`: red on the verbatim revert, on `"You sent"`, on `"Vas gastar"`, and on the
key being deleted (the merge falls back to English and the pin sees it). It drops the
tombstone table, includes `ja` with no special-casing, goes through the same
`lodash.merge` path the app renders from, and — because `Locales` is the generated union in
[i18n-types.ts:8](app/i18n/i18n-types.ts:8) — a 29th locale is a compile error until it is
pinned. Put it in `__tests__/i18n/` next to its siblings ([S2]).

### F3 — The heading is status-blind: a failed or pending send reads "You sent"

Pre-existing and non-blocking; noted because this branch's rationale is the accuracy of
exactly this verb. The branch at
[transaction-detail-screen.tsx:397](app/screens/transaction-detail-screen/transaction-detail-screen.tsx:397)
picks the heading from `direction` and on-chain broadcast state only, so a lightning
payment with `status: "FAILURE"` or `"PENDING"` falls through to `sent()`. For a pending
one the date row at least prints `PENDING`; for a failed one nothing on the screen says
so, and a user opening a payment that never went through reads "You sent" over a negative
amount. It was equally wrong as "You spent"; the rename neither introduces nor fixes it.

**Proposal.** Out of scope here — file a follow-up on #426's thread rather than growing
this PR. The shape of the fix is a `status` check ahead of the direction one, a new
`TransactionDetailScreen.failed` key across the 28 locales, and `FAILURE` / `PENDING` rows
in the heading spec's `cases` table (§2.1: "enumerate all enum states"). If the team
decides past tense is acceptable for a pending payment, pin that decision with a case so
it is deliberate rather than accidental.

---

## 2. Test Coverage Gaps

**G1 — The motivating scenario has no case.** `d38657551`'s message justifies the rename
with "a transfer to yourself moves money without paying anyone", but the heading spec's
`cases` table
([transaction-detail-heading.spec.tsx:131](__tests__/screens/transaction-detail-heading.spec.tsx:131))
covers a lightning send, a lightning receive, and a queued on-chain send. Add an
intraledger self-transfer (`settlementVia: { __typename: "SettlementViaIntraLedger", … }`,
`direction: "SEND"`) expecting "You sent", so the case that motivated the copy is the one
that pins it.

**G2 — The `Sending` → `You sent` flip is pinned on one side only.** The only branch with
two inputs is `onChainTxNotBroadcasted`; the spec has the queued side
(`transactionHash: null` → "Sending") but no broadcast on-chain send (`transactionHash`
set → "You sent"). The detox suite covers it end-to-end with a 30× timeout
([03-payment-send.test.ts:223](e2e/detox/03-payment-send.test.ts:223)); the unit spec is
where it belongs.

**G3 — `FAILURE` and `PENDING` sends.** From [F3]: neither status appears in any heading
case, so today's behaviour is neither pinned nor decided.

**G4 — `ja`'s value is pinned by nothing.** `locale-parity` checks the key exists and the
new guard skips `ja` by design, so 送金完了 could become anything without a test noticing.
The positive pin in [F2] closes this for free.

---

## 3. Code Smells

**S1 — Three specs hand-roll the same `LL` stub; the shared helper has every mock but
i18n.**
[transaction-detail-mocks.tsx](__tests__/helpers/transaction-detail-mocks.tsx) exports
`mockThemedText`, `mockScreen`, `mockNavigation`, `mockAppHooks`, … and is imported by
the heading, fallback and row-actions specs — yet each of fallback, row-actions and
seen-state types out its own `TransactionDetailScreen: { received, sending, sent|spent,
… }` literal, and the heading spec takes a fourth route (the real `en` bundle). §2.4. One
`mockI18n()` export, or the real bundle everywhere, and [F1] cannot recur.

**S2 — The locale guard lives in a component render spec.** It needs `fs` and `path`; it
sits inside a file that mocks fifteen modules to render a screen, so anyone running
"just the copy check" pays for the render harness, and anyone looking for locale checks in
`__tests__/i18n/` — where `brand-name`, `map-distance-label` and `locale-parity` read the
same JSONs the same way — will not find it. It also adds a fourth copy of the
`TRANSLATIONS_DIR` resolver
([transaction-detail-heading.spec.tsx:196](__tests__/screens/transaction-detail-heading.spec.tsx:196)).
Move it to `__tests__/i18n/send-heading-copy.spec.ts`; with [F2]'s shape it needs neither
`fs` nor the resolver.

**S3 — `edc8d4958` is a fixup, not a commit.** By its own message, `d38657551` left all
five row-actions tests crashing, and `1c424a075` sits on top of that red state; the
"test assertions updated alongside copy changes" line in the i18n checklist is the guide's
"most common `Tests` failure" precisely because of this. Squash-merge hides it on `main`,
but the stack is what gets reviewed. `d38657551` also appends a stray blank line to the
heading spec (a Prettier violation at that commit) that `1c424a075` then fills in — the two
were split after the fact. Squash all three, plus [F1]'s two lines.

**S4 — The screen comment names a "longest locale" that is already only half true.** The
comment above the heading
([transaction-detail-screen.tsx:431](app/screens/transaction-detail-screen/transaction-detail-screen.tsx:431))
was rewritten in this PR from `ms "Anda dah belanjakan"` to `nl "Je hebt verzonden"`; the
same `<Text>` also renders `sending`, and `af`'s "Besig om te stuur" ties `nl` at 17
characters today. It will rot silently the next time any of the 84 strings changes, and
nothing checks it. Say "a long locale" and drop the example — the mechanism
(`numberOfLines={1} adjustsFontSizeToFit`) is what the comment is protecting, not the
record holder.

**S5 — `en-key-order.json` is an orphan that now names a dead key.** A tracked root-level
file listing every `en` key path, last touched by #3601, referenced by nothing in the repo
(the only grep hit is itself). After this branch it lists `TransactionDetailScreen.spent`.
Not this PR's doing; it is just the first change to make the file visibly wrong. A
separate `chore:` should delete it, or regenerate it inside `update-translations` if it is
meant to be an inventory.

PR #4271, opened 2026-09-14
fix(self-custodial): keep the removed account on screen until removal finishes

### LINKED ISSUE #4264: Wrong wallet shown during account deletion animation
## Describe the bug

When deleting one Blink account while multiple accounts are configured, Blink switches to another remaining wallet while the deletion animation is still visible.

This makes it look as if the newly displayed wallet is the one being deleted.

In my case, I deleted an empty wallet. During the deletion animation, Blink suddenly switched to another wallet that still contained funds. For a moment, it looked like I had accidentally deleted the wallet containing funds instead of the empty one.

Even though the correct wallet was apparently deleted, the UI transition is extremely confusing and alarming.

## To Reproduce

1. Have multiple Blink accounts configured in the app.
2. Select one of the accounts.
3. Delete that account.
4. Observe the screen while the deletion animation is still running.
5. Blink switches to another remaining wallet before the deletion animation has fully finished.
6. The deletion animation is visually shown together with the newly selected wallet.

## Expected behavior

The wallet being deleted should remain clearly identifiable until the deletion process and animation are completely finished.

Blink should only switch to another wallet after the deletion animation has ended.

Ideally, the app could also briefly confirm which wallet/account was deleted, so there is no ambiguity.

## Screenshots

Unfortunately, I could not take a screenshot during the issue. Screenshot capture also appeared to be blocked/not available in this situation.

## Smartphone

* Device: Honor (exact model unknown)
* OS: Android
* Version: Unknown

## Additional context

This is particularly concerning in a Bitcoin wallet because the next wallet displayed may still contain funds.

Showing that wallet while a deletion animation is still active can make the user believe they deleted the wrong wallet, even if the underlying deletion worked correctly.

The transition between "wallet deleted" and "next wallet selected" should therefore be visually unambiguous.

designsats:
## Problem

Removing the active self-custodial account showed the **next** account while "Please Wait." was still on screen. In the repro (Anon user, $0, removed; satoshin21, $3.58, remaining) the green check moved to satoshin21 during the spinner, and Home opened on the funded wallet with nothing saying which account was removed. For a moment it looks like the wrong wallet was deleted.

Fixes #4264

## Cause

`useDeleteAccount` switches the active account **before** it disconnects the SDK and wipes the removed one. That order is deliberate: `useSdkLifecycle` then shuts the SDK down itself, and a spec pins it. But the "Please Wait" `Overlay` in `ProfileRow` and in the Danger zone `DeleteAccount` was transparent, so the screen underneath kept re-rendering for the new active account in plain view. In the Danger zone it was worse: if the next account was custodial, `DangerZoneSettings` swapped out `<DeleteAccount />` partway through, and the overlay disappeared with it.

## Fix

| Commit | What it does |
| --- | --- |
| `6ca41a0ee` | Covers the screen with an opaque lock naming the removed account, starts removal only after the confirm modal has hidden, and shows a toast saying which account was removed |

- **The deletion order is unchanged.** Only what's visible while it runs changes.
- **`AccountDeleteContextProvider` now covers the screen instead of replacing it.** The screen stays mounted under an opaque cover, so the component doing the removal can finish it. The Accounts screen now uses the provider too, and the cover takes an `identifier` so it can name the account.
- **New `useAccountRemoval` hook**, shared by `ProfileRow` and `DeleteAccount`. Confirming only records the removal; it starts in the confirm modal's `onModalHide`, so the modal fading out never overlaps the lock. On success it navigates and then shows `ProfileScreen.removedAccount` ("You removed account {identifier}."). That string already exists, so there are no locale changes. On failure it lifts the lock.
- **The name is saved when you confirm.** After the switch, the row's title and the wallet's live lightning address describe the next account.
- `CustomModal` and `DeleteAccountConfirmModal` now pass `onModalHide` through.
- The transparent per-component overlays are removed.

## Acceptance requirements

- [ ] Removing the active self-custodial account from Settings → Account → Switch account never shows the next account (check mark, name or balance) before the removal finishes
- [ ] While removing, the screen is covered and shows the removed account's name
- [ ] The removal lock only appears after the confirm modal has fully closed
- [ ] After removal, Home shows "You removed account {name}." naming the removed account
- [ ] Removing an inactive account from the list behaves the same (lock, Home, toast)
- [ ] Settings → Account → Danger zone → Delete shows the same lock and toast, including when the remaining account is custodial
- [ ] Cancelling the confirm modal removes nothing and shows no lock
- [ ] Custodial account deletion (Danger zone → Delete account) still shows its "being closed" screen and completes as before

## How it should behave 

[delete-acc.webm](https://github.com/user-attachments/assets/d000dab3-5e51-4177-ac79-05ec06a68f1d)

## Testing specs

| Spec | Covers |
| --- | --- |
| `__tests__/screens/settings-screen/account/account-delete-context.spec.tsx` | Cover names the account; screen stays mounted under it; cover stays up when the component that raised it leaves; release lifts it |
| `__tests__/screens/settings-screen/self-custodial/use-account-removal.spec.ts` | Nothing runs without a confirmed request; lock is raised with the name before `deleteWallet`; success navigates and shows the toast without dropping the lock; failure releases it; a removal runs once |
| `__tests__/screens/settings-screen/self-custodial/profile-row.spec.tsx` | Removal waits for `onModalHide`; cancel + hide removes nothing; lock and toast name the removed account even after the row re-renders for the switched one |
| `__tests__/screens/settings-screen/self-custodial/delete-account.spec.tsx` | Danger zone waits for `onModalHide`; lock is named from the address read before the switch |
| `__tests__/components/custom-modal.spec.tsx`, `__tests__/screens/settings-screen/delete-account-confirm-modal.spec.tsx` | `onModalHide` is passed through |

```
yarn jest __tests__/screens/settings-screen/account __tests__/screens/settings-screen/self-custodial __tests__/components/custom-modal.spec.tsx __tests__/screens/settings-screen/delete-account-confirm-modal.spec.tsx __tests__/screens/settings-screen/multi-account.spec.tsx __tests__/screens/settings-screen/account-screen.spec.tsx __tests__/self-custodial/hooks/use-delete-account.spec.ts
```

Reverting each part fails its spec: going back to replacing the screen (2 fail), starting removal on confirm (1), dropping the lock before navigating (1).

Manual QA (debug build, emulator):
1. Create two self-custodial accounts; fund one.
2. With the empty one active, Settings → Account → Switch account → ✕ on it → type "delete" → Confirm.
3. The modal closes, then a covered screen shows the removed account's name → Home opens on the other account with "You removed account …".
4. Repeat from Settings → Account → Danger zone.

## Notes / behavior changes

- The self-custodial removal wait screen is now opaque and named, replacing the transparent "Please Wait." spinner.
- Removal starts after the confirm modal's closing animation, a few hundred ms later than before.
- A new success toast appears after self-custodial removal.
- The custodial deletion wait screen now covers the account screen instead of replacing it; it looks the same.


designsats:
@kngako can you review? Not urgent.

PR #4272, opened 2026-09-15
fix(self-custodial): send from the dollar wallet to a lightning address without an idempotency key

esaugomez31:
## Summary

Self-custodial sends from the Dollar wallet to a Lightning address or an LNURL-pay destination fail every time with "The payment details look invalid. Please review and try again." Sends from the Bitcoin wallet are not affected.

Closes blinkbitcoin/blink-wip#1281

## Cause

`createSelfCustodialLnurlPaymentDetails` always mints an idempotency key and hands it to `lnurlPay`. A Dollar send also carries a `tokenIdentifier` and a `UsdToBtc` conversion, and the Breez SDK refuses that combination in `orchestrate_send`:

> Idempotency key is not supported for payments with a token transfer leg (direct token send or AMM conversion).

Token transfers have no idempotency hook on the Spark side, so the key cannot protect the retry it promises; the SDK rejects it instead of accepting it silently. The error surfaces as `InvalidInput`, which the app maps to `sc_invalid_input` and the generic copy above.

Versions:

- The check exists in `breez/spark-sdk` from 0.17.0 on. It is not in 0.15.0 or 0.16.0.
- The app was on 0.15.0 until #4155 moved it to 0.22.0. That upgrade is where Dollar sends to a Lightning address started failing.
- The Dollar path was verified on device under 0.13.2 when LNURL pay landed (#3765), so it did work before.

The spec did not catch it because `executeLnurl` is mocked, and one test asserted that the key is forwarded on every send.

## Fix

The key is passed to `executeLnurl` only for a Bitcoin send. It is still minted and carried on the payment detail, so a Bitcoin send keeps its retry protection and switching the wallet back to Bitcoin reuses the same key. A Dollar send has no idempotency on retry; that was already the case, the SDK just says so now.

The second commit closes coverage gaps in `lnurl.spec.ts` that predate this change (the setters, `decipher` on message and url success actions, a zero amount, a missing domain). `lnurl.ts` is at 100% on every column.

## Retry safety on the keyless rail

Without a key the SDK cannot refuse a duplicate, and its own guidance for a send that throws after dispatch is to look for the payment before sending again. So a dollar LNURL send that throws after `lnurlPay` was dispatched now looks its payment up in the wallet (outgoing payments since the attempt, matched by Lightning address or LNURL domain, three looks two seconds apart) and reports it as sent or pending when found; only a wallet that shows nothing lets the slider re-arm. A retry made later, once the connection is back, looks for the earlier attempt's payment before it sends anything. The lookup lives in `send-helpers.ts` so the other self-custodial rails can share it.

The detail no longer mints the key at construction: it carries an `idempotencyKeyRef` holder that `useSendPayment` mints into inside its own try, as the custodial details do, and drops it on a new amount or wallet.

## Verification

- `yarn tsc:check`, eslint, `lnurl.spec.ts` (47 tests) and the self-custodial payment-details and bridge suites (162 tests) pass.
- Not covered by the emulator run: a dollar send forced to throw after dispatch (connection dropped mid-`lnurlPay`). It needs a device with a real network cut, so it is left for the manual matrix.
- On an emulator with a self-custodial account, sending $0.50 from the Dollar wallet to a `@blink.sv` Lightning address, on `main` and on this branch:

| Before (`main`) | After (this branch) |
| --- | --- | 
| 01-before-usd-lnurl-send-fails | 02-after-usd-lnurl-confirmation |

<details>
<summary>After: payment result and receipt</summary>
<p>
03-after-usd-lnurl-payment-successful
04-after-usd-lnurl-receipt
</p>
</details>


kngako (changes requested):
Claude found no blocking findings. But I am upgrading finding 1 to blocking.

---

## Findings

### Finding 1 — A dollar LNURL send that fails ambiguously is re-armed for a blind retry that can pay twice [NON-BLOCKING (upgraded to Blocking)]

`sendPaymentMutation` (`app/self-custodial/payment-details/lnurl.ts:219–242`) wraps both `prepareLnurl` and `executeLnurl` in one `catch` that maps every throw to `PaymentSendResult.Failure`; `useSendPayment` treats `Failure` as retryable and unlocks the slider (`app/screens/send-bitcoin-screen/use-send-payment.ts:338–340`), on the documented assumption that the retry carries the same idempotency key (`:242–247`). After this branch a USD send carries no key, so the assumption is false for exactly this rail. Concretely: a user sends $20 from the dollar wallet to `friend@walletofsatoshi.com`, the SDK converts USDB and dispatches the Lightning payment, and the connection drops before `lnurlPay` resolves — the JS side gets a `NetworkError`, the screen shows "Network connection problem. Check your connection and try again." (`app/i18n/en/index.ts:4208`), the user slides again, and the SDK — which has no way to recognise the second request as the first — pays $20 twice. The SDK's own docs name this case and the remedy: "Retrying after a failure that leaves the outcome unknown may pay twice … Look for the [payment] … before sending it again" (`breez_sdk_spark.d.ts:20684–20687`). The custodial path already does precisely this for its 409 conflict (`send-bitcoin-confirmation-screen.tsx:322–360` via `useVerifyPaymentSettled`); the self-custodial USD path has no equivalent.

**Proposal:** Add a verify-before-retry step for the keyless rail, mirroring `useVerifyPaymentSettled`: when `executeLnurl` (not `prepareLnurl`, which is safe to retry) rejects on a USD send, poll the bridge's existing `listPayments` (`app/self-custodial/bridge/wallet.ts:60`) for outgoing Lightning payments with `fromTimestamp` = the attempt's start and `lnurlPayInfo.lnAddress`/`domain` matching the destination — the mapper already reads that field (`app/self-custodial/mappers/transaction.ts:120–123`) — and return Success/Pending with that payment instead of `Failure`; only an empty result should re-arm the slider. Until that lands, at minimum split the catch so an `executeLnurl` rejection on a USD send does not surface the "try again" copy, and extend the new source comment with the consequence and the SDK doc reference, so the next reader sees that the dropped key is a trade-off rather than a free change. This is the same exposure every other self-custodial rail already has (Smell 3), so the guard belongs somewhere `send-helpers.ts` can share.

### Finding 2 — The newly covered setters are not pinned to carry the idempotency key [NON-BLOCKING — land in this PR]

The second commit's tests for `setInvoice`, `setSuccessAction` and `setConvertMoneyAmount` (`lnurl.spec.ts:154–197`) execute those three lines and assert a field each, but the only reason all three spread `paramsWithKey` rather than `params` is to keep the minted key alive across recreations — and no test observes that. Mutants M5/M6/M7 (spread `params`) and M8 (`idempotencyKey: undefined` in `setInvoice`) all survive with 47/47 green. None of the three runs between an attempt and its retry on the confirmation screen today, so the mutants are harmless in the straight-line flow — but the details screen calls `setConvertMoneyAmount` on every converter refresh (`send-bitcoin-details-screen.tsx:202–210`), and `useSendPayment` explicitly promises that "backing out of the confirmation screen and re-entering retries under the original key" (`use-send-payment.ts:275–279`). Under M5 a fixed-amount destination (`min === max`, whose root detail is never replaced by `setAmount`) that is backed out of, catches a price tick, and is re-entered would send under a fresh key, and the suite — and the coverage number the commit was written for — would stay perfect. The existing preservation test (`:756`) covers `setMemo`/`setAmount`/`setSendingWalletDescriptor` for exactly this reason and simply was not extended.

**Proposal:** Extend "preserves the idempotency key across … recreations" to chain `setConvertMoneyAmount`, `setInvoice` and `setSuccessAction` as well, asserting `mockExecuteLnurl.mock.calls[n][2]` equals the original key after each (the `randomUUID` counter spy in that `describe` already makes a fresh key detectable). Then the three new smoke tests can stay as they are, or fold the `setInvoice` one into the chain since "comes back unchanged" is best expressed as "still sends under the same key". Re-run M5–M8 afterwards; each should fail exactly one test.

---

## Test Coverage Gaps

1. **The discriminator is not pinned (M3 survives).** Both idempotency tests set the wallet currency and the typed amount's currency to the same value (`lnurl.spec.ts:694–708` BTC/BTC, `:715–737` USD/USD), so `unitOfAccountAmount.currency === Usd` passes in place of `isUsdSend`. That mutant would strip the key from the most common bitcoin send there is — a USD-display user paying from the BTC wallet — because the details screen seeds the amount in the display currency (`send-bitcoin-details-screen.tsx:234–237`) and `chooseWallet` re-denominates it again after a wallet switch (`:322–336`). Add a BTC-wallet/USD-amount case asserting a string key (probe P3 passes on HEAD) and a USD-wallet/BTC-amount case asserting `undefined`.
2. **The wallet-switch carry the new comment justifies is untested.** "The key is still generated and carried above so a bitcoin send's retries stay deduplicated" (`lnurl.ts:201–202`) is the reason a dollar detail mints a key it cannot use, yet no test creates a USD detail, switches to BTC, and asserts the key appears and repeats on retry — nor the reverse (BTC → USD drops it). Both hold on HEAD (probes P1/P2). The existing switch test only goes BTC → BTC because `SetSendingWalletDescriptor<T>` pins the generic; the screen crosses it at runtime with `wallet.walletCurrency`, so the test needs the same cast.
3. **Key carry through `setConvertMoneyAmount` / `setInvoice` / `setSuccessAction`** (Finding 2).
4. **The ambiguous-failure retry is undocumented in the spec.** Nothing shows that a rejected `executeLnurl` on a dollar send returns `Failure` and that the next `sendPaymentMutation` goes out with no key (probe P4). Pinning it now makes the trade-off in Finding 1 visible and gives the follow-up a red test to turn green.
5. **SDK contract drift is invisible to the suite.** Every spec mocks `@app/self-custodial/bridge`, so the rejection introduced by an SDK bump was undetectable for three-plus weeks; the bridge spec still only checks that `executeLnurl` forwards whatever key it is given (`__tests__/self-custodial/bridge/send.spec.ts:392–418`). There is no self-custodial send in detox. Short of a device test, add a checklist item to SDK-upgrade PRs (#4155 has an empty body) to re-run the manual matrix from #3765 — USD→LN address, BTC→LN address, bech32 LNURL — and record it.

## Code Smells

1. **`Crypto.randomUUID()` runs at construction, outside any `catch`, for a key a dollar detail can never use.** `lnurl.ts:161` mints the key when the detail is built — inside the details screen's seeding effect (`send-bitcoin-details-screen.tsx:215–240`), pre-existing since #3765 — whereas the custodial hook deliberately defers minting to send time inside a `try` because the CSPRNG is "the first synchronous native (Nitro JSI) call" and a throw outside the send handler reaches the app-wide ErrorBoundary (`use-send-payment.ts:271–288`). The new comment re-affirms the construction-time design and gives it a reason (the wallet switch); the mint happens at most once per lineage, since `paramsWithKey` carries it through every rebuild. A lazy mint on the first `sendPaymentMutation`, stored on the same carried field, would match the custodial precedent and put the native call where its failure can be reported.
2. **Commit subject narrows the change.** "send from the dollar wallet to a lightning address" — the fix applies to every LNURL-pay destination the detail handles, bech32 LNURL included (`destination: lnurlParams.identifier || lnurl`). Message-only; worth widening before merge so `git log -S` on a raw-LNURL report finds it.
3. **Pre-existing: the other self-custodial rails never had a key.** `executeSend` (`app/self-custodial/bridge/send.ts:100–110`) has no idempotency parameter and `send-helpers.ts:136,154` never pass one, although `SendPaymentRequest.idempotencyKey` supports non-token Spark/Bolt11/on-chain sends. After this branch the only self-custodial send with replay protection is a bitcoin-wallet LNURL send. Out of scope here, but it is the frame for Finding 1 — the exposure is rail-wide — and the natural home for one shared verify-before-retry helper.
4. **New top-level tests sit outside the spec's thematic groups.** "can neither quote nor send while the amount is zero" (`lnurl.spec.ts:199`) belongs under `describe("amount handling")`, and the empty-domain case (`:216`) under the `prepareLnurl options` group; the file is now 855 lines and the groups are what make it navigable.
5. **`setInvoice({} as never)`** (`lnurl.spec.ts:158`) reads as if the setter accepts an empty invoice. Passing the real shape (`{ paymentRequest: "lnbc…", paymentRequestAmount: … }`) costs two lines and documents that the self-custodial detail ignores it on purpose.

esaugomez31:
> Claude found no blocking findings. But I am upgrading finding 1 to blocking.
> ## Findings
> ### Finding 1 — A dollar LNURL send that fails ambiguously is re-armed for a blind retry that can pay twice [NON-BLOCKING (upgraded to Blocking)]
> 
> `sendPaymentMutation` (`app/self-custodial/payment-details/lnurl.ts:219–242`) wraps both `prepareLnurl` and `executeLnurl` in one `catch` that maps every throw to `PaymentSendResult.Failure`; `useSendPayment` treats `Failure` as retryable and unlocks the slider (`app/screens/send-bitcoin-screen/use-send-payment.ts:338–340`), on the documented assumption that the retry carries the same idempotency key (`:242–247`). After this branch a USD send carries no key, so the assumption is false for exactly this rail. Concretely: a user sends $20 from the dollar wallet to `friend@walletofsatoshi.com`, the SDK converts USDB and dispatches the Lightning payment, and the connection drops before `lnurlPay` resolves — the JS side gets a `NetworkError`, the screen shows "Network connection problem. Check your connection and try again." (`app/i18n/en/index.ts:4208`), the user slides again, and the SDK — which has no way to recognise the second request as the first — pays $20 twice. The SDK's own docs name this case and the remedy: "Retrying after a failure that leaves the outcome unknown may pay twice … Look for the [payment] … before sending it again" (`breez_sdk_spark.d.ts:20684–20687`). The custodial path already does precisely this for its 409 conflict (`send-bitcoin-confirmation-screen.tsx:322–360` via `useVerifyPaymentSettled`); the self-custodial USD path has no equivalent.
> 
> **Proposal:** Add a verify-before-retry step for the keyless rail, mirroring `useVerifyPaymentSettled`: when `executeLnurl` (not `prepareLnurl`, which is safe to retry) rejects on a USD send, poll the bridge's existing `listPayments` (`app/self-custodial/bridge/wallet.ts:60`) for outgoing Lightning payments with `fromTimestamp` = the attempt's start and `lnurlPayInfo.lnAddress`/`domain` matching the destination — the mapper already reads that field (`app/self-custodial/mappers/transaction.ts:120–123`) — and return Success/Pending with that payment instead of `Failure`; only an empty result should re-arm the slider. Until that lands, at minimum split the catch so an `executeLnurl` rejection on a USD send does not surface the "try again" copy, and extend the new source comment with the consequence and the SDK doc reference, so the next reader sees that the dropped key is a trade-off rather than a free change. This is the same exposure every other self-custodial rail already has (Smell 3), so the guard belongs somewhere `send-helpers.ts` can share.
> ### Finding 2 — The newly covered setters are not pinned to carry the idempotency key [NON-BLOCKING — land in this PR]
> 
> The second commit's tests for `setInvoice`, `setSuccessAction` and `setConvertMoneyAmount` (`lnurl.spec.ts:154–197`) execute those three lines and assert a field each, but the only reason all three spread `paramsWithKey` rather than `params` is to keep the minted key alive across recreations — and no test observes that. Mutants M5/M6/M7 (spread `params`) and M8 (`idempotencyKey: undefined` in `setInvoice`) all survive with 47/47 green. None of the three runs between an attempt and its retry on the confirmation screen today, so the mutants are harmless in the straight-line flow — but the details screen calls `setConvertMoneyAmount` on every converter refresh (`send-bitcoin-details-screen.tsx:202–210`), and `useSendPayment` explicitly promises that "backing out of the confirmation screen and re-entering retries under the original key" (`use-send-payment.ts:275–279`). Under M5 a fixed-amount destination (`min === max`, whose root detail is never replaced by `setAmount`) that is backed out of, catches a price tick, and is re-entered would send under a fresh key, and the suite — and the coverage number the commit was written for — would stay perfect. The existing preservation test (`:756`) covers `setMemo`/`setAmount`/`setSendingWalletDescriptor` for exactly this reason and simply was not extended.
> 
> **Proposal:** Extend "preserves the idempotency key across … recreations" to chain `setConvertMoneyAmount`, `setInvoice` and `setSuccessAction` as well, asserting `mockExecuteLnurl.mock.calls[n][2]` equals the original key after each (the `randomUUID` counter spy in that `describe` already makes a fresh key detectable). Then the three new smoke tests can stay as they are, or fold the `setInvoice` one into the chain since "comes back unchanged" is best expressed as "still sends under the same key". Re-run M5–M8 afterwards; each should fail exactly one test.
> ## Test Coverage Gaps
> 
>     1. **The discriminator is not pinned (M3 survives).** Both idempotency tests set the wallet currency and the typed amount's currency to the same value (`lnurl.spec.ts:694–708` BTC/BTC, `:715–737` USD/USD), so `unitOfAccountAmount.currency === Usd` passes in place of `isUsdSend`. That mutant would strip the key from the most common bitcoin send there is — a USD-display user paying from the BTC wallet — because the details screen seeds the amount in the display currency (`send-bitcoin-details-screen.tsx:234–237`) and `chooseWallet` re-denominates it again after a wallet switch (`:322–336`). Add a BTC-wallet/USD-amount case asserting a string key (probe P3 passes on HEAD) and a USD-wallet/BTC-amount case asserting `undefined`.
> 
.
.
.
.
.
Thanks @kngako, all applied:

Finding 1: a dollar LNURL send that throws after dispatch now looks its payment up in the wallet before reporting anything (findLostSend in send-helpers.ts, shared for the other rails). Found means Success/Pending with that payment; only nothing found re-arms the slider. A retry made later, once the connection is back, looks for the earlier attempt's payment before it sends.
Smell 1: no more randomUUID at construction. The detail carries an idempotencyKeyRef that useSendPayment mints into, same as the custodial details, and it's dropped on a new amount or wallet.
Finding 2 + gaps 1-4: pinned in lnurl.spec.ts (key across every recreation, BTC wallet with USD amount and the reverse, wallet switch both ways, the ambiguous failure path).
Gap 5: checklist item in docs/pr-review.md for SDK upgrades.
Smells 4/5: tests moved into their groups, setInvoice gets the real shape.
Smell 2: fix commit reworded to "LNURL destination".

kngako:
All issues from previous review have been addressed. New blocking finding has been found.
---

## Findings

### Finding 1 — The lost-send lookup can claim a neighbouring payment to the same address and report a failed send as sent [BLOCKING — small fix]

`isPaymentTo` (`app/self-custodial/payment-details/lnurl.ts:99–107`) matches any outgoing Lightning payment whose `lnurlPayInfo.lnAddress` equals the destination (or whose `domain` matches, for raw LNURL), and `findLostSend` opens its window 5 s before the attempt (`send-helpers.ts:180,202`) and, on the retry path, at the *first* attempt's clock (`lnurl.ts:336`). Nothing ties the match to the invoice this attempt paid — even though `prepareLnurl` hands back `invoiceDetails.paymentHash` (`PrepareLnurlPayResponse`, SDK `.d.ts`) and the wallet records the same hash on every Lightning send at `details.inner.htlcDetails.paymentHash`, which `transaction-csv.ts:110` already reads. Concretely (probe P5): a user at a till pays `merchant@shop.example` $5, then a fixed-amount $20 request from the same merchant a few seconds later; the $20 `lnurlPay` is rejected outright with `InsufficientFunds`, the catch runs the lookup regardless of error class, the $5 payment (created 3 s earlier, inside the lookback) matches by address, and the screen shows the $20 as **sent** — with the $5 payment's preimage as proof and its success action — while the merchant received nothing further. The same address-only match also lets a second device on the same seed, or any legitimate repeat payment to the same address, be claimed on the retry path, and it can never find a raw-LNURL send whose params carry no `domain` (`hasDomain` false → no match → retry allowed → double-pay). The custodial precedent for this exact mechanism, `useVerifyPaymentSettled`, matches by payment hash for these reasons.

**Proposal:** Match on the hash and keep the address as a fallback only. Take `prepared.invoiceDetails.paymentHash` (or `.invoice.bolt11`) after `prepareLnurl`, and make `matches` check `details.inner.htlcDetails?.paymentHash === hash` first, falling back to the address/domain comparison only when a hash is unavailable. For the retry path, store the hash(es) on the holder — `lostSendRef: { startedAtMs?: number; paymentHashes?: string[] }`, appending each lost attempt's hash, since a retry that also throws leaves two invoices either of which may have landed — and match any of them. With an exact key the 5 s lookback becomes harmless and the lookup can safely be skipped for definitive error classes (`InsufficientFunds`, `InvalidInput`) to spare the user the 4 s wait; add P5 as the regression test (fails before, green after), plus one where the hash matches but the address does not (the SDK lower-casing or normalising the recorded address must not hide a landed payment).

---

## Test Coverage Gaps

1. **No test drives `findLostSend` and `isPaymentTo` together.** `lnurl.spec.ts` mocks `findLostSend`; `send-helpers.spec.ts` uses an abstract `toFriend` predicate. The false-success in Finding 1 lives in the seam between them and neither spec can see it. One integration-style case with the real helper and a mocked `listSentPaymentsSince` (P5's shape) closes it.
2. **The definitive-failure path is not distinguished.** Nothing asserts what happens when `executeLnurl` rejects with a class that cannot have moved money (`InsufficientFunds`, `InvalidInput`); today it takes the same 4 s lookup as a timeout. Whether that is intended deserves a pinned test either way (see Smell 1).
3. **Raw LNURL with no domain.** "hands the SDK an empty domain when the lnurl params carry none" (`lnurl.spec.ts:354`) shows the case exists, but no test shows the lost-send lookup can never match it. Resolved by the hash; until then it deserves a documenting test.
4. **Spark-detailed results.** "carries no preimage or success action for a found payment without them" feeds a `Spark`-tagged found payment through `foundOutcome`, but `isPaymentTo` returns `false` for any non-Lightning details — so if the SDK ever records a pay-request send under `PaymentDetails.Spark`, it is unfindable. Worth a one-line confirmation from the SDK docs/binary and a test pinning the chosen behaviour.

## Code Smells

1. **Every keyless throw pays the 4 s lookup, definitive or not.** `sendPaymentMutation` catches `executeLnurl` and runs `findThisSend` for any error (`lnurl.ts:345–355`), so an `InsufficientFunds` — nothing moved — shows its error ~4 s late while the slider says "Confirming…". Cheap to gate on `classifySdkError(err)` ∈ {NetworkError, Generic} once the match is exact.
2. **`f856b9a5a` has no commit body.** A 931-line change that swaps the key model to the custodial holder, changes reset semantics on `setAmount`/`setSendingWalletDescriptor`, and adds a polling lookup is described only by its subject; the in-code comments are excellent, but the round-1 doc noted `#4155`'s empty body as how the original regression went unexplained. The house format (see #3765) is a body per design decision.
3. **`lostSendRef` and `idempotencyKeyRef` travel as two holders with identical lifecycles.** Both are created together, carried by the same setters, reset by the same setters, and read by the same mutation. One `attemptRef: { idempotencyKey?, lostSend? }` would make the "same holder as the key" comments (`lnurl.ts:51–55`, `lnurl.spec.ts:1194`) literally true and remove the second `?? {}` / `: undefined` pair per setter.
4. **`LostSendRef` is a private type but `IdempotencyKeyRef` comes from the shared `index.types`.** If Smell 3 is not taken, the lost-send holder belongs next to `IdempotencyKeyRef` so the on-chain/Bolt11 rails can adopt it without re-declaring it.

grimen (changes requested):
## Findings

**1. The lost-send lookup cannot match a send that was actually lost, so the dollar LNURL rail still pays twice on a retry.**
`isPaymentTo` (`app/self-custodial/payment-details/lnurl.ts:99–107`) only accepts a Lightning payment whose `lnurlPayInfo` carries the destination's `lnAddress` or `domain`, and returns `false` when that field is missing (`:103`). In SDK 0.22.0, `lnurlPayInfo` is written only by `finalize_lnurl_pay`, which runs after `orchestrate_send(...).await?` has returned successfully (`crates/breez-sdk/core/src/sdk/lnurl/pay.rs:293–343`, persisted at `:386–398`). Nothing is stored before the network call either: `pay_and_persist_lightning_invoice` builds the `Payment` only after `pay_lightning_invoice` returns (`sdk/lightning_sender.rs:65–85`). So when `lnurlPay` throws after dispatch (the only case this lookup exists for), the payment reaches the wallet later through sync as a bare Lightning send. That record has `htlcDetails.paymentHash` but no `lnurlPayInfo`, and it can never satisfy `isPaymentTo`. The same happens when the payment completes but `finalize_lnurl_pay` itself throws, for example an AES success action on a Spark-routed payment (`sdk/helpers.rs:109–113`). The spec encodes the wrong record shape: the `paymentTo` fixture (`lnurl.spec.ts:973`) gives a lost send a populated `lnurlPayInfo`, and `:1058` pins `lnurlPayInfo: undefined → false` as intended behaviour.

*Example:* a user sends $20 from the dollar wallet to `friend@walletofsatoshi.com`, and the connection drops before `lnurlPay` resolves. `findThisSend` asks three times over about 4 s and finds nothing, because the payment either has not synced yet or has synced without `lnurlPayInfo`. The slider re-arms. On the retry the pre-check at `:320` runs again, still cannot match the now-synced record, and a second $20 goes out. A probe using the real `findLostSend` and a `listSentPaymentsSince` that returns the SDK-shaped record (Pending, `htlcDetails.paymentHash` = the prepared invoice's hash, `lnurlPayInfo: undefined`) fails both checks: the first attempt returns `FAILURE`, and `executeLnurl` is called twice.

*Proposal:* this moves kngako's round-2 hash proposal from "match the hash first" to "match on the hash only". Keep `prepared.invoiceDetails.paymentHash` (`PrepareLnurlPayResponse.invoiceDetails: Bolt11InvoiceDetails`) and match `details.inner.htlcDetails.paymentHash` against it. The address fallback can only match a send that finished normally, and such a send never reaches this catch, so it adds only the neighbour false positive kngako described. On the retry path, store the hash or hashes on `lostSendRef` instead of `startedAtMs`, because the retry's `prepareLnurl` fetches a new invoice. kngako's round-2 Finding 1 (the neighbouring payment claimed as sent) is still open at `0032045` and is closed by the same change.

**2. The lookup window compares the device clock with the SSP's clock, allowing 5 s of skew.**
`findLostSend` asks for payments with `timestamp >= floor(startedAtMs / 1000) − 5` (`send-helpers.ts:202`), where `startedAtMs` is the phone's `Date.now()`. The SDK filters with `p.timestamp >= ?` (`persist/sqlite.rs:605`), and a Lightning send's `timestamp` is the SSP's `created_at` (`models/adaptors.rs:381`). On a phone whose clock runs more than 5 s ahead of the SSP, the lookup excludes the payment it is looking for, even after Finding 1 is fixed.

*Example:* a phone set manually 30 s fast (common after travel, or with automatic time turned off) re-arms the slider for every lost dollar send.

*Proposal:* once the match is on the payment hash, a wide window costs nothing, since no other payment can match. One option is a lookback of several minutes plus a page size that covers it. If the 5 s is deliberate, a comment saying which clock the timestamp comes from would help.

## Test Coverage Gaps

- No test drives the real `findLostSend` together with `isPaymentTo` using a record shaped the way the SDK writes a thrown send (kngako's round-2 gap 1). That seam is where Finding 1 lives. The probe above is that test: it fails on HEAD and should pass after the fix.
- `paymentTo` (`lnurl.spec.ts:973`) and the `lnurlPayInfo: undefined` assertion (`:1058`) should be rewritten to the SDK's shape, keyed by payment hash. As written, they turn the bug into the specification.
- No test covers a retry whose `prepareLnurl` returns a different invoice from the lost attempt. Under hash matching, that case is what shows the retry looks for the earlier hash and not the new one.

## Code Smells

- The direct path reports `PaymentSendResult.Success` for whatever `lnurlPay` returns (`lnurl.ts:340–344`). But `executeLnurl` passes no completion timeout, so the SDK returns the payment right after submission, usually still Pending (`lightning_sender.rs:112–113`). `foundOutcome` maps the same Pending payment to `Pending`, so one payment reports two different outcomes depending on the path. The direct-path mapping predates this PR, but the PR adds the second mapping next to it.
- The `isPaymentTo` doc comment ("both as the wallet records them on a pay-request send") is only true for a send that finished. Once the match moves to the hash, the comment should say why the address is not used.

<details>
<summary>Probe spec (fails 2/2 on 0032045)</summary>

```ts
/* eslint-disable camelcase */
import { LnUrlPayServiceResponse, Satoshis } from "lnurl-pay"

import { PaymentSendResult, WalletCurrency } from "@app/graphql/generated"
import { createSelfCustodialLnurlPaymentDetails } from "@app/self-custodial/payment-details/lnurl"

const mockPrepareLnurl = jest.fn()
const mockExecuteLnurl = jest.fn()
const mockListSentPaymentsSince = jest.fn()

jest.mock("@app/self-custodial/bridge", () => ({
  prepareLnurl: (...args: unknown[]) => mockPrepareLnurl(...args),
  executeLnurl: (...args: unknown[]) => mockExecuteLnurl(...args),
  extractLnurlFee: jest.fn(),
  listSentPaymentsSince: (...args: unknown[]) => mockListSentPaymentsSince(...args),
  buildConversionType: jest.fn().mockReturnValue({ tag: "ToBitcoin" }),
  resolveSendTokenIdentifier: (currency: WalletCurrency) =>
    currency === WalletCurrency.Usd ? "usdb-token-id" : undefined,
  toSdkSendAmount: (amount: number) => BigInt(amount * 10000),
}))
jest.mock("@app/utils/sleep", () => ({ sleep: () => Promise.resolve() }))
jest.mock("@app/utils/error-logging", () => ({ reportError: jest.fn() }))
jest.mock("@app/self-custodial/sdk-error", () => ({
  SelfCustodialErrorCode: {
    InsufficientFunds: "sc_insufficient_funds",
    BelowMinimum: "sc_below_minimum",
  },
  classifySdkError: () => "sc_network_error",
}))
jest.mock("@breeztech/breez-sdk-spark-react-native", () => ({
  AesSuccessActionDataResult_Tags: { Decrypted: "Decrypted", ErrorStatus: "ErrorStatus" },
  FeePolicy: { FeesExcluded: 0, FeesIncluded: 1 },
  PaymentStatus: { Completed: 0, Pending: 1, Failed: 2 },
  OnchainConfirmationSpeed: { Fast: 0, Medium: 1, Slow: 2 },
  SuccessActionProcessed_Tags: { Aes: "Aes", Message: "Message", Url: "Url" },
  PaymentDetails: {
    Lightning: { instanceOf: (o: { tag?: string } | undefined) => o?.tag === "Lightning" },
    Spark: { instanceOf: (o: { tag?: string } | undefined) => o?.tag === "Spark" },
  },
}))

const HASH = "a".repeat(64)

const params = () => ({
  sdk: {} as never,
  lnurl: "lnurl1abc",
  lnurlParams: {
    callback: "https://example.com/cb",
    fixed: false,
    min: 1 as Satoshis,
    max: 100000 as Satoshis,
    domain: "walletofsatoshi.com",
    metadata: [["text/plain", "Test"]],
    metadataHash: "",
    identifier: "friend@walletofsatoshi.com",
    description: "",
    image: "",
    commentAllowed: 0,
    rawData: { metadata: '[["text/plain","Test"]]' },
  } as LnUrlPayServiceResponse,
  unitOfAccountAmount: { amount: 2000, currency: WalletCurrency.Usd, currencyCode: "USD" },
  isMerchant: false,
  convertMoneyAmount: jest.fn((a, t) => ({ amount: a.amount, currency: t, currencyCode: t })),
  sendingWalletDescriptor: { id: "w-usd", currency: WalletCurrency.Usd },
})

/**
 * The wallet's record of a send whose lnurlPay threw after dispatch, as the SDK 0.22.0
 * writes it: lnurlPayInfo is only attached by finalize_lnurl_pay, which runs after
 * orchestrate_send returned (pay.rs:331-398); a record that arrives through sync is a
 * bare Lightning send carrying the invoice's payment hash.
 */
const lostSendRecord = () => ({
  id: "transfer-1",
  status: 1, // Pending
  timestamp: BigInt(Math.floor(Date.now() / 1000)),
  details: {
    tag: "Lightning",
    inner: {
      htlcDetails: { paymentHash: HASH, preimage: undefined },
      lnurlPayInfo: undefined,
    },
  },
})

describe("probe: lost dollar LNURL send as the SDK records it", () => {
  beforeEach(() => {
    jest.clearAllMocks()
    mockPrepareLnurl.mockResolvedValue({
      feeSats: BigInt(0),
      invoiceDetails: { paymentHash: HASH, invoice: { bolt11: "lnbc1..." } },
    })
    mockExecuteLnurl.mockRejectedValue({ tag: "NetworkError" })
    mockListSentPaymentsSince.mockResolvedValue({ payments: [lostSendRecord()] })
  })

  it("finds the dispatched payment instead of re-arming the slider", async () => {
    const detail = createSelfCustodialLnurlPaymentDetails(params())
    if (!detail.canSendPayment) throw new Error("expected canSendPayment")

    const first = await detail.sendPaymentMutation({} as never)
    expect(first.status).not.toBe(PaymentSendResult.Failure)
  })

  it("does not dispatch a second payment on the retry", async () => {
    const detail = createSelfCustodialLnurlPaymentDetails(params())
    if (!detail.canSendPayment) throw new Error("expected canSendPayment")

    await detail.sendPaymentMutation({} as never)
    mockExecuteLnurl.mockResolvedValue({
      payment: { timestamp: BigInt(1), details: undefined },
    })
    await detail.sendPaymentMutation({} as never)

    expect(mockExecuteLnurl).toHaveBeenCalledTimes(1)
  })
})
```

</details>

PR #4274, opened 2026-09-15
feat(send): slide to send with a busy border and rotating progress labels

designsats:
Closes blinkbitcoin/blink-wip#1135 · Parent blinkbitcoin/blink-wip#1134 · Failure-mode inventory blinkbitcoin/blink-wip#1273

The slide-to-send control on the review screen had a greyed-out disabled state that gave no reason, and a small spinner while a send was in flight that was easy to miss. A long wait looked like a frozen app. This PR rebuilds `GaloySliderButton` so the busy state is obvious and shows progress, following the [Lovable prototype](https://swipe-and-glow.lovable.app/) ("v2, rotating solid") and the Figma `galoy-slider-button` component (`2377:17720`).

**UI only.** The send, fee, dust and 409-verify logic is untouched. The screen only passes new props.

## Spec

**Track**
- 60 high, full width of its container, radius 30, `grey5` fill, 1px `grey4` border.
  - Figma draws it at 50. On device that read smaller than the old control, so it stays 60.
- The handle is a 60 circle with a 30 white arrow. Its colour follows the sending wallet: `primary` for Bitcoin, `_green` for Dollar (the Dollar pill's green).

**States**
| State | When | Label | Handle |
|---|---|---|---|
| Calculating | fee loading, fee error without an amount, or dust check `pending` | **Calculating fee…** | 50% opacity, can't be dragged |
| Disabled | any other disable (over balance, `sendPayment` withheld, dust check `blocked`) | **Slide to send** | 50% opacity, can't be dragged |
| Ready | send allowed | **Slide to send** (`grey2`), fades as the handle moves | draggable |
| Busy | from the moment the send commits, through `sendPayment` loading and the 409 verify | rotating progress labels | slides to the end and fades out |
| Failure | the send settles without leaving the screen | back to Ready | springs back to rest |

**Drag**
- The send commits as soon as the handle covers **70%** of its travel, while you're still dragging. Letting go earlier springs it back, and nothing is sent.
- On commit the handle eases to the end over 300 ms, fades past 90% of the travel, and a medium haptic fires.

**Busy**
- **Border:** a 1px foreground (`black`) arc runs clockwise round the track, with a gap that breathes between 40° and 130° of the loop. It starts at 0.6 rev/s and eases to 0.3 rev/s over 10 s, so long waits feel calmer.
[...]

PR #4273, opened 2026-09-15
feat(send): amount entry with an in-screen keypad for free and fixed amounts

designsats:
Closes blinkbitcoin/blink-wip#1274 · Parent blinkbitcoin/blink-wip#1134 · Failure-mode inventory blinkbitcoin/blink-wip#1273

Slices 1 and 2 of Send 3.0. The send details screen had an amount field that opened a modal keypad, a dropdown wallet picker and a "Max" text button. This PR replaces them with one amount-entry screen that has two variants:

- **Free amount:** the in-screen keypad, the percent chips and a wallet card you can switch.
- **Fixed amount:** the same screen, with the keypad and chips removed.

**UI only.** Validation (`isValidAmount`), limits, fee tiers, the LNURL invoice fetch and the high-fee modal are wired in unchanged.

## Spec

The screen, top to bottom:

1. **Nav bar:** back arrow only, no title (it used to say "Send"). The screen leads with its own icon and destination.
2. **Hero** — the shared `SendHero`, in its **active** mode. Review (blinkbitcoin/blink-wip#1278) renders the same component settled, so the block does not move between the two screens.
   - `send` icon in `primary`, 32 in a 44 frame, no background. It's `IconHero`, which now takes a bare icon frame, a caption line and a compact spacing mode.
   - Destination on one line:
     - Raw on-chain addresses, invoices and Spark addresses keep their first 10 and last 8 characters (`bc1q6pwejx...mg9fq4aw`).
     - A Blink username gets the host appended (`name@blink.sv`).
     - A Lightning address or LNURL stays whole.
     - Long-press copies the full destination. The whole destination line is the copy trigger, not a separate icon. Review (blinkbitcoin/blink-wip#1278) follows the same rule, where the entire Destination field copies.
   - **Main amount:** SourceSansPro-Bold 36/38 for the first five digits (`12,345`). From the sixth digit it steps down evenly per digit to 22/24 at ten digits (`1,000,000,000`) and holds there, and it still shrinks to fit one line. Review's settled amount is unchanged.
   - **Second amount:** SourceSansPro-Bold 18/24, shown only when the second currency differs from the first.
   - **Colours**
     | State | Main amount | Second amount |
     |---|---|---|
     | Empty (free amount, still zero) | `grey2` | `grey3` |
     | Active (typed, chip-set or fixed) | `black` (foreground) | `grey2` |

     A fixed amount always uses the active colours.
[...]

PR #4275, opened 2026-09-15
feat(send): review screen with a Sending hero, Destination field and Details card

designsats:
Part of blinkbitcoin/blink-wip#1278 · Parent blinkbitcoin/blink-wip#1134 · Failure-mode inventory blinkbitcoin/blink-wip#1273

> [!NOTE]
> **Stacked on #4273 (amount entry) and #4274 (slide to send).** The branch merges both, so their commits show here until they land. The review work is the two commits `feat(send): review screen…` and `test(send): cover the review screen…`, plus the follow-ups from device testing listed under **Round 2** and **Round 3**. Merge commits bring in later changes from #4273 and #4274 (currently #4274's 60-high slider).

Slice 6 of Send 3.0. The confirmation screen was five stacked grey fields (Destination, From, Amount, Note, Fee), with every error as the same centred red paragraph above the slider. This PR rebuilds it as the **review** screen: the amount leads, then Destination, From Balance and one Details card, in the same order on every rail.

**UI only.** `useFee`, `useSendDustWarning`, `useSendPayment`, `useVerifyPaymentSettled`, the balance check, `handleSendPayment` and the `sendBitcoinCompleted` params are unchanged. The one data addition is read-only: self-custodial on-chain payment details now expose `feeTier`.

## Spec

The screen, top to bottom. The content scrolls and the slider stays pinned under it.

1. **Nav bar:** back arrow only, no title (it used to say "Send"). The arrow is always tinted with the theme foreground (`black`), so it is visible in dark and light mode even where the screen draws its own header row.
2. **Hero** (Figma `20362:111251`) — #4273's `SendHero`, settled rather than active, so the block does not move when the sender crosses from amount entry to review.
   - `send` icon, 32 in a 44 frame, `primary`, **no background** (`IconHero` with a bare frame).
   - "Sending", 12/18.
   - The leading amount, SourceSansPro-Bold 20/24, shrinks to fit one line. It is in the currency the sender last swapped amount entry's keypad to (`preferredAmountCurrency`): the display currency by default, or the sending wallet's unit. A fixed amount always leads with the display currency.
   - The other amount, 14/20, only when the display currency differs from the wallet's. The pair is display ↔ the **sending wallet's** currency (sats for Bitcoin, dollars for Dollar), the same pair the keypad swaps between.
   - Review has no tap-to-swap: the choice is made on amount entry.
3. **Destination - {type}** (Lightning for lightning and lnurl, Onchain, Intraledger, Spark)
   - The card flow's `ActionField`: a labelled `grey5` field with one action. Raw addresses and invoices go in whole and shorten in the middle on one line at whatever width the field has, as the old confirmation screen did. Usernames get their host from #4273's `formatDestination` (`name@blink.sv`); Lightning addresses stay whole.
   - It keeps `ActionField`'s radius 8 and 16/22 value rather than Figma's radius 12 / `p3`, for the same reason the Details card keeps its own radius: a global pass aligns both.
   - **The whole field is the copy button**, not only the icon.
4. **From Balance:** #4273's `SendWalletSummary`, read-only (no switch), on the `grey7` static surface.
   - **Hidden balance:** when balances are hidden (the global setting, or hidden from home), the card shows the placeholder. Tapping the card reveals the balance **for this screen only**. The setting and home's state don't change.
5. **Details:** `InfoSection` / `InfoRow` from the card flow, on `grey7`.
   | Row | When | Value |
   |---|---|---|
   | Transaction priority | On-chain only | `{tier} ~ {eta}`. Custodial reads `payoutSpeed` (custodial windows), self-custodial reads `feeTier` (SDK windows) |
   | Fee | Always | Loading: a spinner that keeps the row height. Set: `$0.10 (11 sats)`, or `11 sats ($0.10)` when the hero leads with the wallet unit; the bracketed amount is in the sending wallet's currency. testID `Successful Fee`. Max fee: value + ` *`. Failed with no amount: `—`, with the reason under the card |
   | Note | Only when the payment has a note | The note, right-aligned. The label takes only its own width, so an address in the note wraps whole onto the next line rather than breaking mid-word |

   No error text is ever drawn inside the card. Every error goes to the notice slot below it.
6. **Notice slot** under the card, in this order. Errors use the unfilled `GaloyErrorBox`, left-aligned with the icon on the first line. When the error is one a different amount can fix, a centred **Change amount** `Chip` sits 12 below it and goes back to amount entry.
   | Case | Card outline | Message | Change amount | Slider |
   |---|---|---|---|---|
   | Send failure | none | The failure text | When amount-fixable | As today |
   | Total exceeds balance (R4) | `error` | "Total exceeds your balance of …" | Always | Disabled |
   | Fee quote failed, no amount (R2) | `error` | The translated self-custodial code, or the generic fee error | When amount-fixable | Disabled |
   | High on-chain fee (R6), when nothing above blocks | `warning` | `WarningBanner` "High fee. We recommend Lightning.", wraps | — | Enabled |
   | Max fee (R3) | none | `*` footnote in `grey2` | — | Enabled |
   | USD remainder sweep (R5 `visible`) | none | `WarningBanner` | — | As today |

   **Amount-fixable failures** (`isAmountFixableError`), checked against the raw error before it is translated:
   | Kind | Self-custodial code | Custodial backend text (blinkbitcoin/blink source) |
   |---|---|---|
   | Low funds | `sc_insufficient_funds` | "… exceeds balance …" (`payment-flow.ts`), "No balance left to send." (`send-on-chain.ts`) |
   | Too small | `sc_below_minimum` | "Use lightning to send amounts less than N sats" (`onchain-payment-flow-builder.ts`), "Amount sent was too low for recipient's usd wallet." (`error-map.ts`) |
   | Over a limit | — | "Cannot transfer more than $X in 24 hours" (`limits-checker.ts`: intraledger, withdrawal, trade) |

   Custodial errors carry no code, so these are matched on wording: if the backend rewords one, the chip stops appearing for it and the error still shows. Network, routing, invoice and generic failures never get the chip.
7. **The outline never moves anything.** It's a 1px layer drawn over the card (`pointerEvents="none"`), not the card's border, so turning it on or changing its colour leaves every row where it was. Notices only appear below the card.
8. **Slider:** #4274's, pinned at the bottom with 10 above and 20 below. `disabled` and `isLoading` inputs are unchanged.

**Spacing** (Figma body column): 20 at the sides, 10 under the nav bar, 14 between blocks, 7 between a title and its field.

### Rulings from testing that override the ticket

| Ticket said | Ruled during testing |
|---|---|
| Hero icon in a 44px circle | No background: the 32 glyph centred in the 44 frame, Figma's type sizes |
| Destination keeps `PaymentDestinationDisplay` + a copy icon | The shared `ActionField`, with the whole field as the copy trigger. #4273's body notes the same rule for its destination line. `PaymentDestinationDisplay` had no callers left and is deleted |
| Hidden balance: From Balance shows the placeholder | Kept, plus a tap on the card reveals it for this screen only (no change to the setting or session) |
| R2 and S3–S10 open the #1275 problem sheet | **Deferred:** the sheet doesn't exist yet, so both show inline in the notice slot. No "Try again" and no `useFee` retry trigger until then |
| Tier names Priority / Standard / Economy | Waits for #4262. Review reads the same `SendBitcoinScreen.fast/medium/slow` strings, so it picks up the rename when that lands |
| Details card radius 14, padding 14, title gap 7 (Figma) | `InfoSection` keeps its own radius 8 / padding 10·14 / gap 3 for now; a future global pass aligns it |
| Service fee row (Figma) | Left out. One Fee row, as ruled in the ticket |

### Round 2: rulings from device testing

| Was | Ruled |
|---|---|
| Change amount as a text link (`GaloySecondaryButton`) | A centred `Chip`, the same pill as amount entry's 25/50/75/100% |
| Change amount only for "Total exceeds your balance" | Every failure a different amount can fix: low funds, below minimum / dust, 24h limits, from the fee quote or the send |
| Fee quote failure in red inside the Fee row | `—` in the row, the reason under the card. No errors inside the card |
| Unfilled error box text centred | Left-aligned, icon on the first line, everywhere it is used |
| Note value squeezed to half the row, so `satoshi@cake.cash` broke as `…@cake` / `.cash` | `InfoRow` label takes its own width (max 50%), the value gets the rest |

### Round 3: the amount keeps the currency the sender typed in

| Was | Ruled |
|---|---|
| Review always led with the display currency, even after the sender swapped to sats on amount entry | The hero and the Fee row lead with the saved `preferredAmountCurrency`, the setting amount entry already writes on each swap. It is app-wide, so it also holds for the next send |
| Review's second line was always sats, even for a Dollar-wallet send | The sending wallet's currency: dollars for Dollar, so review shows the same pair as amount entry |
| Fixed amount led with the invoice's own denomination on amount entry | Display currency first on amount entry and on review, whatever the saved choice |
| — | No tap-to-swap on review |

### Round 4: the destination fills its field

| Was | Ruled |
|---|---|
| Raw addresses and invoices cut to a fixed 10…8, leaving most of the field empty | The whole destination, shortened in the middle at the field's width, as the old confirmation screen did. Usernames keep their host. Amount entry's caption and the receipt keep 10…8 |

### Reuse pass (self-review)

A pass over the branch for anything it rebuilt that the app already had. Behaviour is unchanged except where noted.

| Was | Now |
|---|---|
| `SendReviewHero` | #4273's `SendHero`, settled mode |
| `SendReviewDestination` drawing its own field | `ActionField`, which gains `numberOfLines` (middle truncation), `iconSize` and `valueTestID` |
| `PaymentDestinationDisplay` | Deleted — nothing referenced it once entry and review moved on |
| The tier name map and a hand-rolled `payoutSpeed` reverse lookup | `useFeeTierLabels()` and `feeTierFromPayoutSpeed()` |
| The completed screen shortening the destination at 13…8 | The same `formatDestination` (10…8) as amount entry |
| The wallet card reading `grey7` from "no switch" | The explicit `inactive` prop |
| Change amount chip styled on its own | New `Chip` atom, extracted from `PercentageSelector`, which now draws its chips with it |

### Round 5: the back arrow is visible outside a native header

| Was | Ruled |
|---|---|
| On a device build with the #1136 Sent transition, where review is headerless and draws its own header row, the back arrow could be tapped but wasn't visible | The shared `HeaderBackButtonWithTheme` always tints the arrow: the tint a header passes, otherwise the theme foreground (`black`) |

**Cause.** `@react-navigation/elements`' `HeaderBackButton` passes `tintColor={tintColor}` to `HeaderIcon`, and `HeaderIcon` spreads its props after its own `tintColor={colors.text}` default. With no tint given, the explicit `undefined` replaces that default and the icon draws its raw asset: a black glyph, invisible on the dark background but still tappable. The native header never hit this because it passes `headerTintColor`. Review's own header row renders the button bare, so nothing set the tint. The fix is in the shared button rather than at the call site, so no future bare render can hit it again.

On this branch review still uses the native header, so the arrow already showed here. The fix is what lets the headerless review keep a visible arrow.

## Acceptance requirements

**Every rail** (Blink username, Lightning invoice, Lightning address / LNURL, on-chain, Spark), **custodial and self-custodial**
- [x] Hero, Destination, From Balance and Details render in that order, with no nav title.
- [x] The back arrow is visible in dark and light mode and returns to amount entry with the amount, wallet and note as they were.
- [x] The back arrow is tinted with the theme foreground even when rendered outside a native header (review's own header row).
- [x] With the display currency chosen (the default), the hero amounts equal the `currencyAmount` / `satAmount` passed to `sendBitcoinCompleted`.
- [x] After swapping to sats on amount entry, review's hero and Fee row lead with sats; swapping back leads with the display currency again. The choice holds for the next send.
- [x] A Dollar-wallet send pairs the display currency with dollars, not sats.
- [x] A fixed amount leads with the display currency on amount entry and on review, whatever was last chosen.
- [x] Pressing anywhere on the Destination field copies the full destination.
- [x] A raw on-chain address or invoice fills the Destination field and is shortened in the middle, not at a fixed 10…8.
- [x] From Balance can't switch wallets and sits on `grey7`, like the Details card.
- [x] Fee row: spinner while loading, the fee once set (`Successful Fee`), ` *` plus the grey footnote for a max fee, `—` when the quote fails with no amount.
- [x] No error text inside the Details card; every error sits under it, left-aligned.
- [x] Note row only when a note is set; an address in it wraps whole, not mid-word.
- [x] Over balance with the fee: red outline, "Total exceeds your balance of …", the **Change amount** chip goes back, slider disabled.
- [x] Fee quote failed with no amount: red outline, the reason under the card, slider disabled.
- [x] Fee quote failed below the minimum or for insufficient funds: the Change amount chip is offered. Failed on the network: no chip.
- [x] A send failure shows its text under the card with no outline. Change amount appears only when the failure is amount-fixable (e.g. the 24h limit), not for e.g. "route not found".
- [x] Drawing or recolouring the outline doesn't shift the rows.

**On-chain**
- [x] Custodial: Transaction priority reads the payout speed with the custodial window (e.g. `Slow ~ 24h`).
- [x] Self-custodial: Transaction priority reads the SDK tier with its window (e.g. `Medium ~ 30m`).
- [x] Fee over 2% of the total: warning outline and the advice wraps, slider enabled. Low funds wins over it.

**Off-chain**
- [x] No Transaction priority row.

**Hidden balance**
- [x] With balances hidden, From Balance shows the placeholder; hero amount and fee stay visible.
- [x] Tapping the card shows the balance; leaving and reopening review hides it again; the global setting is untouched.

## Notes / behavior changes

- **The confirmation screen looks completely different.** QA should expect the hero + Details layout and the Destination field as a copy button.
- **Changes to shared components that other screens will see:**
  - `InfoRow`: new optional `loading` (spinner holding the row height) and `valueTestId`. Card screens and account migration don't pass them. **Layout change:** the label now sizes to its text (max 50%) with a 12 gap, and the value takes the rest of the row, where before each took half. Short labels give their values more room on card screens too; long labels still wrap at half.
  - `Chip` (new, `components/atomic/chip`): label, `selected`, `disabled`, `loading`. `PercentageSelector` renders it with the same styles, test IDs and accessibility labels, so amount entry and conversion look and test the same.
  - `GaloyErrorBox` unfilled variant: left-aligned text, icon top-aligned with the first line. Also seen in amount entry's error lines (`send-bitcoin-details-extra-info.tsx`, #4273). The filled variant is unchanged.
  - `InfoSection`: new optional `inactive` (`grey7`) and `outlineColor`. Existing card screens don't pass them.
  - `SendWalletSummary`: new optional `inactive` (the `grey7` static surface and no swap icon), `isBalanceHidden` and `onReveal`. Review passes them; amount entry passes none, so a region-gated card there keeps the active surface and always shows the balance (#4127).
  - `ActionField`: new optional `numberOfLines`, `iconSize` and `valueTestID`, and its field grows past 50 instead of being fixed at it. Card details passes none of them. Its `fontFamily` was `"Source Sans Pro"`, which Android resolves to Roboto; it now uses the `fonts` token.
  - `formatEta` is now exported from `fee-tier-options.ts`.
  - `HeaderBackButtonWithTheme` (`header-back-control.tsx`) is exported and falls back to `colors.black` when it's given no `tintColor`. Every native header already passes `headerTintColor` (`colors.black`), so no existing back arrow changes.
- **Amount currency (round 3):** review reads `preferredAmountCurrency`, which amount entry already wrote but nothing downstream read. The completed screen still receives display-first `currencyAmount` / `satAmount`, so after a swap to sats review and the receipt show the two amounts in different orders. The receipt is out of scope; flagged if it should follow.
- **Fixed-amount entry (round 3):** the amount-entry hero now leads a fixed amount with the display currency instead of the invoice's denomination. This is a #4273 surface changed here.
- **The completed screen's Recipient line** is now shortened by `formatDestination`, like amount entry: raw destinations keep 10…8 instead of 13…8, and an intraledger username arrives with its host (`name@blink.sv`). Receipt/Share is otherwise out of scope for Send 3.0.
- **Payment detail type:** `feeTier?: FeeTierOption` on `PaymentDetail`, set only by `createSelfCustodialOnchainPaymentDetails`. It doesn't change how fees are quoted or payments sent.
- **Nav title** on `sendBitcoinConfirmation` is now empty.
- **i18n:** new `SendBitcoinConfirmationScreen.sending`, `fromBalance`, `details`, `changeAmount`; `lightningRecommended` reads "High fee." instead of "High fee!". All in `en` and all 28 locales.
- **Locale parity still fails on this branch** for #4274's slider keys (`slideToSend`, `calculatingFee`, `sendProgress.*`), whose translations were deferred in that PR. This PR's keys are complete.
- **No migration.**

### Flagged, not fixed (senior dev, per blinkbitcoin/blink-wip#1134's UI-only rule)
- **D22 / N13:** send max shows the full balance in the hero while the fee comes out of it.
- **G4:** custodial `contactCreate` is awaited without a catch after SUCCESS.
- **No high-fee modal on self-custodial on-chain.** `ConfirmFeesModal` (fee over 50% of the amount, on amount entry's Next) is gated on custodial by `shouldWarnAboutHighFee` (`!isSelfCustodial`), on upstream `main` as well as here. It was not dropped by Send 3.0, but a self-custodial sender only gets review's 2% banner. The SDK already quotes the picked tier's fee, so removing the gate would be enough.

## Testing specs

```
yarn jest __tests__/screens/send-confirmation.spec.tsx __tests__/components/header-back-control __tests__/screens/send-bitcoin-screen __tests__/components/card-screen __tests__/screens/card-screen __tests__/self-custodial/payment-details/onchain.spec.ts --maxWorkers=2
```
Local, after round 2: the command above plus `__tests__/components/percentage-selector.spec.tsx` runs 106 suites, 1,567 tests, all passing. The conversion suites (18 suites, 259 tests) also pass. After round 3: `send-confirmation`, `send-details`, `send-amount-entry` and `send-amount-input-screen` run 116 tests, all passing; reverting the two screens fails the three new specs that pin a behaviour change (sats-led review, dollar pairing, fixed amount on entry). `tsc`, eslint and prettier are clean. Revert checks: removing the tap-to-reveal, the `grey7` Details surface or the self-custodial `feeTier` each fails its specs; the new below-minimum and 24h-limit Change amount specs fail against the old condition.

| Spec | Covers |
|---|---|
| `__tests__/screens/send-confirmation.spec.tsx` (updated) | Section order; hero equals the completed-screen params; Destination field copies; Details on `grey7`; Fee spinner / failed quote as `—` with the error under the card + red outline / max-fee `*` + grey footnote; Change amount for a below-minimum fee quote and a 24h-limit send failure, none for a network fee error; Note row on/off; Transaction priority for custodial `payoutSpeed`, self-custodial `feeTier`, and absent off-chain; R4 red outline + message + Change amount → back + disabled slider; R6 warning outline + untruncated advice + enabled slider; R4 hides R6; send failure without outline or Change amount; tap-to-reveal without touching `toggleHideAmount`; no tap offered while the balance is visible. The LNURL test now reads the hero instead of the old Amount field |
| `__tests__/screens/send-confirmation.spec.tsx` → `amount currency` (round 3) | Hero + Fee row lead with the display currency by default and with sats after a swap to sats; a Dollar wallet pairs with dollars; a fixed amount leads with the display currency whatever was chosen |
| `__tests__/screens/send-amount-entry.spec.tsx` → `amount currency` (round 3) | The keypad opens in the saved currency; each hero tap saves the new choice (`default` ↔ `display`); a fixed amount leads with the display currency whatever was chosen |
| `__tests__/screens/send-bitcoin-screen/amount-fixable-error.spec.ts` (new) | Both self-custodial codes and each backend message count as amount-fixable; network, invalid input, generic, routing and ledger errors don't |
| `__tests__/screens/send-bitcoin-screen/send-hero.spec.tsx` (new, shared with #4273) | `SendHero` in both modes: caption and both amounts; no second line without a second currency; the 44 frame with no background; settled vs entry type sizes; muted empty state; press-to-swap; long-press copy |
| `__tests__/screens/send-bitcoin-screen/review/send-review-destination.spec.tsx` (new) | Title per payment type; username host; a raw address reaches the field whole with one-line middle truncation; foreground colour; the whole field copies |
| `__tests__/screens/send-bitcoin-screen/send-wallet-summary.spec.tsx` (updated) | `grey7` when `inactive`, `grey5` when active — with or without a switch; hidden placeholder replaces both lines; no balance in the accessibility label; tap calls `onReveal` |
| `__tests__/components/card-screen/info-row.spec.tsx` (updated) | Spinner replaces the value; hidden spacer line holds the height; `valueTestId` |
| `__tests__/components/card-screen/info-section.spec.tsx` (updated) | `inactive` surface; no outline by default; outline colour; outline is an absolute overlay and the card has no border |
| `__tests__/components/header-back-control/header-back-control.spec.tsx` (updated, round 5) | A bare `HeaderBackButtonWithTheme` tints its icon with the theme `black` (fails on revert with `undefined`, the invisible-arrow bug); a tint supplied by the header is kept |
| `__tests__/self-custodial/payment-details/onchain.spec.ts` (updated) | `feeTier` defaults to Medium, follows the build param and survives `setAmount` |

> **Not pinned by a test:**
> - How many characters of an address show depends on the field width and the OS's native ellipsis. Check on device.
> - Spacing (20 sides, 10 top, 14 between blocks, 7 title gaps, slider 10/20) is style values only. Check on device.
> - The empty nav title is a navigator option with no spec.
> - Tier and ETA strings in the priority tests are asserted in `en` only.
> - The `InfoRow` label/value widths, the chip's centring and the error box's left alignment are style values; the row and error-box specs mock `makeStyles`. Check on device.

### Manual QA
1. **Blink username, Bitcoin wallet:** review shows the send icon with no circle, "Sending", the amount and sats line, "Destination - Intraledger", From Balance, and Details with Fee. No nav title.
2. Tap anywhere on the Destination field (not the icon): the toast confirms the copy and the clipboard holds the full destination.
   On-chain: the address fills the field up to the copy icon and is shortened in the middle; rotate or use a narrow device and it still fits one line.
3. Add a note on amount entry: a Note row appears. Remove it: the row is gone.
4. **On-chain, custodial:** pick Slow on amount entry. Review shows `Transaction priority · Slow ~ 24h` above Fee. **Self-custodial:** the default reads `Medium ~ 30m`.
5. On-chain with a small amount so the fee is over 2%: the card is outlined in the warning colour and the advice wraps under it; the slider still works.
6. Send close to the whole balance so amount + fee exceeds it: red outline, "Total exceeds your balance of …" left-aligned under the card, and a centred **Change amount** chip that returns to amount entry; slider disabled. Watch the rows: they don't move when the outline appears.
7. Block the API host at the network layer while online and open review: the Fee row shows the spinner at the same height, then `—`; the fee error sits under the card with a red outline and **no** Change amount chip; the slider stays disabled.
8. Make a send fail (e.g. an unroutable invoice on staging): the failure text appears under the card, no outline, no chip.
9. **Self-custodial**, send an amount below the SDK minimum: "Amount is below the minimum…" under the card, not in the Fee row, with the Change amount chip.
10. **Custodial**, send over the 24h limit (or on-chain below the dust threshold): the backend message under the card with the Change amount chip.
11. Turn on **Hide balance**: From Balance shows the dots, the hero amount and fee stay visible. Tap the card: the balance shows. Go back and forward: hidden again. Home is still hidden.
12. Repeat 1–11 in light mode, at the largest OS text size, and on a 360dp device with a long Lightning address and a long note. Nothing clips and the slider stays reachable. A note like "Pay to satoshi@cake.cash" stays on one line or drops the whole address to line two.
13. **Amount currency:** on amount entry tap the amount to type in sats, then Next: review's hero and Fee lead with sats. Go back, tap to swap back to the display currency, Next: review leads with it again. Start a new send: amount entry opens in the last choice. Repeat from the Dollar wallet: the second line is dollars, not sats. Scan a fixed-amount invoice after choosing sats: both screens lead with the display currency.
14. Card flow → card details and replace/order confirm steps: info cards keep their content; short labels now leave their values more width.
15. Amount entry and Convert: the 25/50/75/100% chips look and behave as before. Amount entry's error lines are left-aligned.
16. **Back arrow:** in dark mode, open review. The arrow is visible at top left. Tap it: amount entry shows the same amount, wallet and note. Repeat in light mode. With the Sent transition (#1136) build, where review draws its own header row, the arrow is visible there too.
17. Any other screen with a back arrow (e.g. Settings → Language): unchanged.


kngako:
@designsats seems something broke the tests. Also can we get screenshots of what the new changes look like?

PR #4277, opened 2026-09-16
fix(ui): render every bold at 700, never 600

designsats:
## Summary

All bold text now uses one weight, 700. Before this, bold was 600 in some places and 700 in others.

Refs blinkbitcoin/blink-wip#1304

## Problem

The theme's `<Text bold>` and 30 style rules set `fontWeight: "600"`. Many other rules use `"700"`, and Blink's type scale has Bold 700 and no SemiBold. As a result:

- **iOS** mixes two bolds on the same screen: 600 loads `SourceSansPro-SemiBold`, 700 loads `SourceSansPro-Bold`.
- **Android** never loads the Bold file for a `fontWeight`. RN looks for `SourceSansPro-Regular_bold.ttf`, doesn't find it, and fakes bold from the regular face. The fake bold is thinner than the real one.

## Fix

- **Theme** (`app/rne-theme/theme.ts`): `bold` picks the `SourceSansPro-Bold` face and sets no weight, so both platforms draw the real Bold face. This works with or without a `type`. The face names are tokens in the new `app/rne-theme/fonts.ts`.
- **Style rules:** every `fontWeight: "600"` is now `"700"`: 30 rules in 23 files. `GaloyTertiaryButton` was `clear ? "bold" : "600"` and `CustomModal`'s title was `ios ? "600" : "700"`; both are now just `"700"`. That leaves `Platform` unused in `custom-modal.tsx`, so its import is removed.
- **Guard:** `__tests__/rne-theme/font-weight.spec.ts` fails if a 600 weight or `SemiBold` shows up again anywhere in `app/`.

Which strings *should* be bold is out of scope. Code marks some text bold that Figma draws regular, like "Balances" on Home. The ticket tracks those as future fixes.

## Acceptance requirements

- [ ] No `fontWeight: "600"` and no `SemiBold` anywhere in `app/`
- [ ] `<Text bold>` renders the `SourceSansPro-Bold` face on iOS and Android, for `h1`–`p4` and with no `type`
- [ ] Text without `bold` still renders `SourceSansPro-Regular`
- [ ] On iOS, text that was SemiBold is now Bold: button titles (primary, secondary, tertiary), `CustomModal` titles, the fee tier options, map place sheet (name, open/closed/boosted badges), filter sheet title, search retry, card info row labels, Blink card, balance header
- [ ] On Android, `<Text bold>` (e.g. "Balances" on Home, the restricted-region amounts) draws the real Bold face, not a fake bold
- [ ] Nothing that rendered regular before turns bold, except `<Text bold>` with no `type` (see Notes)

### Automated

| Spec | Covers |
|---|---|
| `__tests__/rne-theme/theme.spec.ts` (new) | `bold` gives the bold face with no weight for every `type` and with none; plain text gives the regular face |
| `__tests__/rne-theme/font-weight.spec.ts` (new) | No 600 weight or `SemiBold` anywhere in `app/` (reports `file:line` for each hit) |
| `__tests__/custodial/components/restricted-region-screen.spec.tsx` | Bold amounts use the bold face and the nested fiat text uses the regular face (it now checks `fontFamily` instead of `fontWeight`) |
| `__tests__/components/card-screen/info-row.spec.tsx` | The label's default bold is `"700"` |

```sh
yarn jest __tests__/rne-theme __tests__/custodial/components/restricted-region-screen.spec.tsx __tests__/components/card-screen/info-row.spec.tsx --runInBand --forceExit
```

Revert check: with `theme.ts` and `balance-header.tsx` put back to `main`, both new specs fail (9/9), and the guard names `balance-header.tsx:132`.

`jest --findRelatedTests` over the changed files: 295 suites pass. `yarn tsc:check` and eslint on the changed files pass.

> `__tests__/screens/send-confirmation.spec.tsx` is flaky on `main` without this change: one of two clean runs failed on a different test ("IncreasedToAvoidDust"). With this change it passed 3/3.

### Manual QA

1. iOS, Home: "Balances" and the wallet balances look bold, the same weight as the primary button title on Send.
2. iOS, open any `CustomModal` (e.g. the backup nudge): the title is Bold, not SemiBold.
3. Android, Home: "Balances" is the real Bold face (compare with the `SourceSansPro-Bold` wallet balance next to it: same weight).
4. Map: open a place. The name and the open/closed badge are Bold. Open the category filter sheet: its title is Bold.
5. Send → on-chain → fee tier selector: the title, the selected label and the option labels are Bold.
6. Settings → Transaction limits: the increase-limits link is Bold. Stable balance settings: the title and row titles are Bold.

## Notes / behavior changes

- **iOS:** everything that was SemiBold (600) is now Bold (700), so it looks slightly heavier. That is intended.
- **Android:** `<Text bold>` switches from fake bold to the real Bold face. Style rules that go from 600 to 700 look the same on Android, because both weights fake bold there. Moving them to `fonts.bold` is a follow-up in the ticket.
- `<Text bold>` with no `type` was ignored before and now renders bold. There's one use: the warning text in `set-lightning-address-modal`.
- The Send 3.0 review stack (#4275) has the same `fonts.ts` and theme change, so whichever merges second resolves a trivial conflict.
- No migration needed.


kngako (changes requested):
---

## Findings

### Finding 1 — The bold face is named, but the Xcode project does not bundle it; a missing face with no weight resolves to the system font at *Regular* [BLOCKING — confirm with an iOS build before merge]

`ios/GaloyApp/Info.plist:96-101` lists the six `SourceSansPro-*.ttf` under `UIAppFonts`, and `ios/link-assets-manifest.json` records them as linked, but `ios/GaloyApp.xcodeproj/project.pbxproj` contains no `PBXFileReference` for any of them — the only app font it references is `DMSans-Bold.ttf` (`project.pbxproj:40`). The six `PBXBuildFile` entries at `project.pbxproj:14-22` (`{isa = PBXBuildFile; }`, no `fileRef`) are the stubs #1161 left behind when it stripped the references, and they are what sits in "Copy Bundle Resources" at `:232-238`. `yarn fonts` cannot repair this: the manifest's sha1s still match the files, so `react-native-asset` skips all six (`filterFileByFilesWhichNotExists`), which is also why `DMSans-Bold.ttf` was re-linked later and Source Sans Pro was not. No pod, script phase or synchronized group copies them either. Unless an iOS build shows otherwise, the app has been rendering San Francisco on iOS since 2023 — which is consistent with what the commit observed there (SF Semibold at 600 vs SF Bold at 700).

*When it comes up:* on iOS the theme now emits `fontFamily: "SourceSansPro-Bold"` with no `fontWeight`. `RCTFontWithFontProperties` resolves the weight to `UIFontWeightRegular` first (`RCTFontUtils.mm:28,43`), tries the string as a family (empty), then as a PostScript name — `[UIFont fontWithName:]` returns nil for a face that is not in the bundle — and falls back to `RCTDefaultFontWithFontProperties(fontProperties)` with that Regular weight (`:355-367`). The balance in `BalanceHeader`, every `type="h1" bold` title, every `bold` amount: all render SF Pro Regular. Before this commit the same elements carried `fontWeight: "600"`, which the same fallback honours as SF Semibold, so the visible effect of merging is that bold text on iOS becomes lighter than it was, app-wide.

**Proposal:** First confirm: on a Mac, build once and run `ls "$(xcodebuild -workspace ios/GaloyApp.xcworkspace -scheme GaloyApp -sdk iphonesimulator -showBuildSettings | awk '/ TARGET_BUILD_DIR/{print $3}')/Blink.app" | grep -c SourceSansPro` — the project file predicts `0`. Then fix the linking rather than the theme: delete the six `SourceSansPro` entries from `ios/link-assets-manifest.json`, delete the six dangling `PBXBuildFile` lines, run `yarn fonts`, and check the pbxproj gains a `PBXFileReference` + build-file pair per face under the `Resources` group (and add the token-to-bundle spec from Coverage Gap 1 so it cannot silently regress again). Independently of that, make the theme degrade the way it used to: add `...Platform.select({ ios: { fontWeight: "700" } })` to the bold branch of `universalStyle` — on iOS a PostScript name plus a weight still resolves to the named family's closest face (`RCTFontUtils.mm:364,372-385`), and a missing face then falls back to SF *Bold* instead of Regular; on Android the weight must stay absent (see Finding 2), which the theme probe should assert per platform.

### Finding 2 — On Android the 26 rewritten `Text` rules draw Roboto Bold, not the Source Sans Pro bold face, and the commit's explanation of why is wrong [NON-BLOCKING — fix the comments in this PR, plan the rules]

The commit message says "on Android neither weight loads a file: RN fakes bold from the regular face", and `app/rne-theme/fonts.ts:4-6` / `theme.ts:24-28` repeat it. RN 0.85's asset path does something else. `TypefaceStyle.nearestStyle` (`ReactFontManager.kt:138-144`) maps a weight below 700 to `Typeface.NORMAL` and 700+ to `Typeface.BOLD`; `createAssetTypeface` (`:162-188`) then looks for `fonts/<fontFamily><ext>.ttf` with `ext` = `""` or `"_bold"`, and when that file is missing falls back to `Typeface.create(fontFamilyName, style)`, which for a name Android does not know is `Typeface.DEFAULT` — Roboto — in the requested style (`Typeface.java:1052-1054,1436-1439`). So `"600"` on `SourceSansPro-Regular` loaded the *Regular* file (no emphasis at all — the actual bug on Android), and `"700"` on it looks for `SourceSansPro-Regular_bold.ttf`, finds nothing, and draws real Roboto Bold. Nothing is synthesised; the typeface is swapped.

*When it comes up:* 26 of the 30 rewritten rules land on the themed `Text` from `@rn-vui/themed`, whose theme style contributes `fontFamily: "SourceSansPro-Regular"` (`withTheme.js` concatenates theme style and prop style, so both survive flattening). `FeeTierSelector`'s title and option labels, `PlaceSheet`'s name and badges, `PaymentOfflineNotice`'s `type="h2"` title, `ConversionFeeRow`'s label and the rest go from Source Sans Pro Regular to Roboto Bold on Android — bolder, as intended, but in a different typeface from the line above and below them, which is the same state the 60 pre-existing `"700"` rules and 48 `"bold"` rules have been in. The commit message's "draws the real Bold face on both platforms" is true only of the `bold` prop. The other four rules (three RNE `Button` `titleStyle`s and `BalanceHeader`'s raw RN `Text`) have no Source Sans Pro family at all and render the system font on both platforms, before and after.

**Proposal:** Correct the three comments and the commit message to the real mechanism — "Android maps a `fontFamily` to `fonts/<name>.ttf` or `<name>_bold.ttf`; a weight of 700+ on a face file we do not ship falls through to Roboto" — because the current wording invites the next author to keep writing `fontWeight: "700"` on Source Sans Pro text believing it merely synthesises. For the 26 themed rules, the fix the theme already provides is the `bold` prop (or `fontFamily: fonts.bold`) with the weight removed; that can be this PR or a follow-up, but the message should stop claiming it is done. The structural alternative that makes `fontWeight` honest everywhere — including the 64 rules that hard-code `fontFamily: "Source Sans Pro"` and are Roboto on Android today — is to register the family once in `MainApplication` via `ReactFontManager.getInstance().addCustomFont(this, "Source Sans Pro", R.font.source_sans_pro)` with a `res/font` XML declaring the 400/700 (and italic) members; `minSdkVersion = 26` is enough for that path, and it would let iOS and Android share the family-name-plus-weight convention that half the app already uses.

### Finding 3 — `RestrictedRegionScreen` combines `bold` with `fontWeight: "700"`, which on Android defeats the bold face on the very screen whose spec was rewritten to prove the face is used [NON-BLOCKING — land in this PR]

`app/custodial/components/restricted-region-screen.tsx:68` (`<Text type="h2" bold style={styles.title}>`) and `:83,91,100` (`<Text type="p3" bold style={styles.balanceValueBold}>`) pair the `bold` prop with rules that still say `fontWeight: "700"` (`:155,181`). It is the only place in `app/` that does this (scanned all 41 `bold` `Text` elements), and it is the screen whose spec now says "Bold is the bold face, not a weight on the regular one" (`restricted-region-screen.spec.tsx:88-89`).

*When it comes up:* the flattened style is `{ fontFamily: "SourceSansPro-Bold", fontWeight: "700" }`. On Android that is `nearestStyle = BOLD`, so `ReactFontManager` looks for `fonts/SourceSansPro-Bold_bold.ttf`, does not find it, and falls through to Roboto Bold — the sat amount and the fee on the restricted-region screen are the one place the new theme *cannot* draw the bold face. `toHaveStyle({ fontFamily: fonts.bold })` is a subset match, so the spec is green while the Android result is the opposite of its comment; the probe I ran (`expect(flatten(style).fontWeight).toBeUndefined()`) fails today with `Received: "700"`.

**Proposal:** Delete `fontWeight: "700"` from `title` and `balanceValueBold`; the `bold` prop now carries the emphasis on both platforms. Strengthen the spec so it can see the combination: after the `toHaveStyle` checks, flatten each element's style and assert `fontWeight` is either undefined or below 700 — that is the actual contract with Android's `nearestStyle`, and it is the assertion that would have caught these four. If the theme is ever meant to be robust against callers doing this, it could strip `fontWeight` from `props.style` when `bold` is set, but a spec plus the comment fix in Finding 2 is the proportionate answer.

### Finding 4 — The source guard forbids the number, not the mechanism [NON-BLOCKING]

`__tests__/rne-theme/font-weight.spec.ts:12` fails the build on `600` or the substring `SemiBold` anywhere under `app/**/*.ts(x)`. That pins the commit's title, but the invariant the theme now depends on is different: *Source Sans Pro text must never carry a weight of 700+*, because on Android that weight is what routes the lookup to `_bold.ttf` and then to Roboto (Finding 2), and a `bold` `Text` must never carry a weight at all (Finding 3). The guard permits both. It also reads any `SemiBold` in a comment as an offence, and misses `fontWeight: "semibold"` — a value RN's `TextStyle` type accepts (`StyleSheetTypes.d.ts:598-628`) — as well as a `600` reached through a constant or a template literal.

*When it comes up:* someone adds `title: { fontWeight: "700" }` to a `makeStyles` block on a themed `Text` — the guard is green, iOS is right, Android draws Roboto. Someone else writes `// was SemiBold before #1304` above a style — the guard fails the build for a comment. A designer tries `fontWeight: "semibold"` because the type allows it — the guard is green; Fabric's `fromRawValue(FontWeight)` does not parse the string and logs `Unsupported FontWeight value` (`conversions.h:231-234`), so the text renders Regular on iOS with an error in the console.

**Proposal:** Keep the grep (it is cheap and it did its job for the migration) but point it at the mechanism: flag `fontWeight` of `700|800|900|bold` in any rule whose block, or whose file's theme, names a Source Sans Pro face — a first cut is "any `fontWeight:` in a file that imports `Text` from `@rn-vui/themed`, unless the rule also sets a non-Source-Sans `fontFamily`", which the 26 rules in Finding 2 would fail until migrated. Restrict the `SemiBold` match to string literals (`["']…SemiBold…["']`) so comments cannot trip it, and either add `semibold` to the pattern or lean on the type instead: a `satisfies TextStyle["fontWeight"]`-style union in `fonts.ts` (`weights = { regular: "400", bold: "700" } as const`) plus an ESLint `no-restricted-syntax` on string-literal `fontWeight` would move this from a regex to the compiler.

---

## Test Coverage Gaps

1. **Nothing ties `fonts.ts` to what the platforms bundle.** `theme.spec.ts:21,27,33` compare the emitted `fontFamily` with the token; the guard compares source with a regex; nobody checks that `SourceSansPro-Bold` is a file iOS or Android will load. A spec that, for each value in `fonts`, asserts `app/assets/fonts/<face>.ttf` and `android/app/src/main/assets/fonts/<face>.ttf` exist, `Info.plist` lists `<face>.ttf`, and `project.pbxproj` references it, fails today on the last check for both tokens — it is the test that would have caught Finding 1, and it keeps the token from ever naming a face that is not shipped (renaming `bold` to `SourceSansPro-Black` leaves every current spec green).

2. **The guard's mutation results.** Fed 15 candidate lines: it catches `"600"`, `'600'`, numeric `600`, `"600" as const`, both ternary forms and `SourceSansPro-SemiBold`; it misses `fontWeight: "semibold"`, a template-literal `` `600` ``, and any indirection (`fontWeight: weights.semi`); it false-positives on `SemiBold` inside a comment. More important than any of those: it does not fail on `fontWeight: "700"` on Source Sans Pro text or on `bold` + `fontWeight`, the two shapes that actually break Android (Findings 2, 3).

3. **`restricted-region-screen.spec.tsx:93-95` cannot see the weight it says is gone.** `toHaveStyle` is a subset match; the probe assertion `expect(StyleSheet.flatten(el.props.style).fontWeight).toBeUndefined()` fails today with `"700"` on both bold elements. Add it (or `expect(Number(fontWeight ?? 0)).toBeLessThan(700)` to state the Android rule exactly) so the spec and its comment agree.

4. **`theme.spec.ts` never reaches the style the native side sees.** It calls the factory directly with a hand-built `{ colors: light }`; the Android default `@rn-vui/base` prepends (`fontWeight: "normal"`, `Text.js:21-23`) and the prop-style concatenation in `withTheme.js` are outside it. One render-level case per platform (`Platform.OS` mocked) asserting the flattened `{ fontFamily, fontWeight }` on a `bold` `Text` — `"normal"`/undefined on Android, `"700"` on iOS once Finding 1's fallback lands — would pin the contract where it is consumed, and it is where the iOS-only weight from Finding 1 needs to be asserted anyway.

5. **`info-row.spec.tsx:174` asserts the value the mock returns.** `makeStyles` is stubbed to `label: { fontWeight: "700" }` (`:23`), so "renders label bold" proves the component picks `styles.label` over `styles.regularLabel`, not that the label is bold; the 600→700 edit there changes nothing the test can observe. Worth a one-line rename to say what it checks, or a render through the real `makeStyles` so the rule at `info-row.tsx:55-56` (`fontFamily: "Source Sans Pro"` + `"700"`, Roboto on Android) is the thing under test.

6. **No before/after from a device.** The commit describes what iOS and Android do, and the code disagrees with it on both platforms. A screenshot per platform of one `bold` title and one rewritten rule — attached to the PR — is the cheapest evidence for a change whose whole effect is visual, and it would have surfaced Finding 1 immediately.

---

## Code Smells

1. **Four comments carry the wrong Android model.** `fonts.ts:4-6`, `theme.ts:24-28`, `font-weight.spec.ts:5-6` and `restricted-region-screen.spec.tsx:88-89` all say Android "synthesises" the weight. The path is file lookup → Roboto fallback (Finding 2). Comments at the point of decision are the right place for this knowledge — which is exactly why they need to be true.

2. **`fonts.ts` is the token module for one caller.** The theme uses it; 64 rules under `app/` still hard-code `fontFamily: "Source Sans Pro"` (every `card-screen` component, `blink-card.tsx:117-213`, `warning-card.tsx:119`, `year-selector.tsx`, …), two say `"SourceSansPro-Regular"` (`mnemonic-word-input.tsx:100`, `backup-phrase-confirm-screen.tsx:235`), and `wallet-overview.tsx:376` says `"SourceSansPro-Bold"` by hand. None of these go through the guard, and the family-name form is Roboto on Android. Either the tokens are the convention — then a lint or the guard should say so — or they are a theme-internal detail and the file comment should not call them "the app's two type faces".

3. **Assets the app can no longer reach are still shipped.** `SourceSansPro-SemiBold.ttf` / `-SemiBoldItalic.ttf` remain in `app/assets/fonts`, `android/app/src/main/assets/fonts` (352 KB), `Info.plist:100-101` and both link manifests, though the guard now guarantees nothing references them. Pre-existing but worth the same cleanup: 19 `Montserrat-*.ttf` files (4.6 MB) sit in the Android assets with zero references in `app/`.

4. **An orphaned rule was edited rather than deleted.** `transaction-limits-screen.tsx:79-84` `increaseLimitsText` has had no consumer since #2699; the commit changed its weight. Remove it.

5. **`theme.spec.ts:7-14` fakes the theme with a double cast.** `(theme.components?.Text as unknown as TextStyleFactory)(props, { colors: light })` works today because the factory reads only `colors`; the first read of `mode` or `spacing` breaks the probe silently. Rendering through `ThemeProvider` (as the restricted-region spec does) or typing the second argument as `Parameters<typeof createTheme>[0]`'s resolved theme keeps it honest for free.

6. **`custom-modal.tsx:187` lost its platform split without a note.** `Platform.OS === "ios" ? "600" : "700"` was a deliberate iOS-lighter choice by someone; collapsing it to `"700"` is right under this commit's rule, but a modal title at 24px is one of the "clips at large sizes" cases the theme comment mentions, so a one-line note in the PR that the title was checked on both platforms would close the loop.

7. **The RNE `Button` titles were never Source Sans Pro.** `galoy-primary-button.tsx:35`, `galoy-secondary-button.tsx:69` and `telegram-button.tsx:48` style `@rn-vui/base`'s `Button`, which renders its title with RN's own `Text` (`Button.js:14`), so there is no theme family on it and the 700 there is SF/Roboto Bold. Not this commit's doing, but if the design intent is Source Sans Pro on buttons, the fix is a `fontFamily` on those `titleStyle`s, not a weight.

PR #4279, opened 2026-09-16
feat(card): steer a signed investor back to the payment from the home

esaugomez31:
Stacked on #4234. Review that one first.

Part of blinkbitcoin/blink-wip#901.

Once the agreement is signed the investor owes a payment, and until now the app forgot about it the moment they left the transfer step. This adds a card on the home that keeps pointing at that payment until it is made, and welcomes them once it is.

## What the home shows

One card, mutually exclusive states, in the order the flow moves through them:

| State | Card | Tap |
|---|---|---|
| Invited, not signed yet | Welcome to become part of Blink | Back to the first screen |
| Signed, money not there | Insufficient balance | Deposit, via the shortfall screen |
| Signed, money spread over both wallets | Your funds are split | Convert, via the same screen |
| Signed, still short, deposit on its way | Hold on | Waits; nothing to do |
| Signed and covered | Complete your investment | Back to the transfer step |
| Paid | Welcome as Blink shareholder! | Close; the record is gone |

Only the last state can be closed. A signed agreement cannot be unsigned, so the others offer no way to hide the card. It sits above the server bulletins and has no priority over them.

The first row exists because the server's invitation bulletin is acknowledged the moment it is tapped, so it works once. Opening the flow's first screen now records the invitation, and this card restates it until the agreement is signed. Whoever leaves before signing keeps a way back. It cannot be closed: an invitation is signed or left to lapse.

The "your card is ready" state that follows is the server's; it arrives as an ordinary bulletin and is not in here.

## How it is recorded

No backend records the investment yet, so it lives in the persisted state as `cardInvestmentByAccountId` (schema 22). Keyed by the custodial account's server id, never the store's active-account slot: every custodial profile on a device shares that slot, and one person's signed agreement must not surface on another's home.

Custodial only. The investment is paid from the investor's own Blink balance, so a self-custodial account has no record, no card, and is sent back to the home from the flow's first screen and from the transfer step, the two places a link can land on. The transfer step also leaves for the home when there is no signed record to pay for, at render and again at the tap, so a record that lapsed while the step was open cannot have an invoice minted and paid on it. Logout keeps this map for the same reason it is safe to.

A record lives a day from its latest moment, the invitation, the signature, the invoice issued to pay it or the payment, the same life the agreement and its payment link are given; after that the home reads it as nothing and the flow starts over.

Once signed, the flow cannot be walked again: signing rebuilds the stack as the home and the transfer step, so nothing that could sign is left underneath, and the first screen, which every way in lands on, sends a signed investor straight to the payment and a paid one back to the home.

Three moments write it. The sign step records the amount and the sats the agreement settles at when the form completes, and does not open the form until the account id is known. The transfer step records the invoice it issues. The receipt marks it paid.

The first screen waits for the account id the same way, since it records the invitation under it. The id is usually in the cache; when it does not come within fifteen seconds the screen says the connection was lost and offers to ask again, the same wait and copy the signing step uses.

The receipt does not know it is settling an investment; the generic send flow is opened on an invoice like any other. So the confirmation passes the invoice it settled to the receipt, and the receipt marks the investment paid only when that invoice is the one the record names, and only once. A payment to anyone else made from the send flow left open is not the investment. Reading the record, rather than a flag set on the way in, holds through an app killed with the payment in flight.

## What "enough" means once signed

Before signing, the balance is measured against the dollars the investor chose. Once signed, the debt is the satoshis the agreement names, fixed at that moment's rate, and the dollars no longer describe it: bitcoin down since, a wallet holding exactly those satoshis is worth less than the chosen dollars and still pays the invoice in full; bitcoin up, a wallet holding exactly the chosen dollars no longer buys them and the payment fails. So the transfer step, the shortfall screen and the home card all measure against the signed satoshis priced at today's rate, the same figure the invoice is written for.

## Paying the same invoice twice

If the app dies with the payment in flight, the receipt never records it and the home keeps asking for the money. Minting a fresh invoice on the next tap would charge the investor twice. So the transfer step pays the invoice already recorded while it still can be paid (25 of its 30 minutes), which meets a claim the recipient has already settled instead of a new one. When the recipient answers that the invoice is already paid, the ledger is asked for the settlement first, as it is for an idempotency conflict, so the receipt carries the real status and moment; when it cannot answer, the receipt is still reached.

## Also in here

The conversion flow can now carry where to return to, so an investor sent to convert lands back on the investment step rather than on the home. The region gates on Convert are waived for that return the way they are for a drain, or a restricted investor with split funds would be passed between the home and Convert with no way to pay; such an investor is held to converting dollars into bitcoin, with the amount still theirs to choose, and waits for the region verdict before the direction is set.

The invoice's external id now carries the paying account and the amount signed for, stamped with the mint time, so the receiving ledger can tie each payment to its subscriber without reading memos. The stamp is there because the ledger keeps the id unique per receiving account and never drops an unpaid invoice, so a second mint for the same amount would otherwise be refused. No invoice is minted while the account is unknown.

## Decisions worth a look

A receipt in `PENDING` counts as paid. A pending payment can still fail and the record has no way back, but holding it would leave the home asking for money that is in flight, and a second tap mints a second invoice.

## Follow-ups, not in this PR

- The recorded invoice is not verified by payment hash before a new one is minted. A custodial payment that settled without being recorded is caught on the next attempt, since the recipient answers "already paid" and the receipt records it; a self-custodial one is not, because that adapter reports a settled invoice as a plain failure, and past the reuse window a second invoice is minted either way. Fix is to look the invoice up before showing the card.
- "Hold on" is shown for any pending deposit, whatever its amount.

## Testing

Specs cover the record and its migration, the resolver's precedence, each card's copy and where it navigates, that the sign step waits for the account id, that the transfer step reuses the invoice recorded while it can still be paid and leaves for the home when there is nothing signed to pay for, that the receipt records only the invoice the record names and only once, that a restricted investor is let into Convert one way only, that the first screen gives up waiting for the account and asks again, and that logout keeps the map.

Device tested on Android against the staging e-sign service: invitation card, signing, transfer and the shareholder card.


kngako:
Review of the latest 32 commits (ea09a2534..HEAD) on feat--invest-pending-home-state — 7 findings (2 blocking), 7 coverage gaps, 9 smells.

---

## Findings

### Finding 1 — The invoice is remembered for a day but only checked for 25 minutes, so a payment whose receipt never ran is paid again [BLOCKING]

**When it comes up:** An investor signs for $25,000 and pays the invoice; the network drops as the mutation goes out, the confirmation screen shows the error, and they close the app rather than retrying — but the payment settled server-side. An hour later the home still says "Complete your investment" (nothing set `paidAt`; only [send-bitcoin-completed-screen.tsx:321](app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx:321) does). They tap Continue: `isInvoiceReusable` is false past 25 minutes ([use-investment-invoice.ts:38](app/screens/card-screen/onboarding/investment-flow/use-investment-invoice.ts:38)), so [transfer-invest-screen.tsx:150-158](app/screens/card-screen/onboarding/investment-flow/transfer-invest-screen.tsx:150) mints a **fresh** invoice — which the `ALREADY_PAID` branch cannot recognise, because it is genuinely unpaid — and $25,000 leaves the account a second time. The record still holds the first invoice for the full 24 hours; nothing asks the ledger about it. The same range deliberately removed the one server-side guard that would have refused the second mint: `resolveInvoiceExternalId` stamps `Date.now()` precisely so the ledger's per-account uniqueness does not bite ([use-investment-invoice.ts:74-78](app/screens/card-screen/onboarding/investment-flow/use-investment-invoice.ts:74)).

**Proposal:** Before minting a replacement, ask the ledger about the invoice already on record. `useVerifyPaymentSettled` ([use-verify-payment-settled.ts:120](app/screens/send-bitcoin-screen/hooks/use-verify-payment-settled.ts:120)) already does exactly this — decode the hash, poll `transactionsByPaymentHash` — and needs only a `walletId`. In `resolvePaymentRequest`, when `progress.invoice` exists but is past reuse, verify it first; if it settled, call `markPaid()` and send the investor to the receipt (or straight to the Shareholder card) instead of minting. Keep the mint as the answer only when the ledger says the old invoice went unpaid, and pin both branches with a transfer-step spec.

### Finding 2 — The home tells a region-restricted investor they are "Ready" with money the send flow will not let them spend [BLOCKING]

**When it comes up:** A custodial investor in a restricted region holds $30,000 in the Dollar balance and $200 in bitcoin, and signs for $25,000. `useInvestmentFunding` reads every wallet off `useActiveWallet()` ([use-investment-funding.ts:42](app/screens/card-screen/onboarding/investment-flow/use-investment-funding.ts:42)), names the Dollar wallet the fullest and answers "covered", so the bulletin renders **Ready** with a Continue button. Continue mints an invoice and opens the send flow, where `useSendWallets` applies `restrictToBitcoinWhenBlocked` ([use-send-wallets.ts:40-50](app/screens/send-bitcoin-screen/hooks/use-send-wallets.ts:40)) and offers only the $200 bitcoin wallet — `isValidAmount` answers `InsufficientBalance` and Next is dead. The shortfall screen is not offered (the app believes the balance covers it), the Convert CTA is not offered (nothing is "split"), and the home goes on saying Ready until the record lapses. This branch fixed the neighbouring case — `066545d3c` lets exactly this investor through the convert gates when their funds *are* split — so the gap is now inconsistent inside the feature: the restriction is honoured one screen later and ignored in the measurement that decides what the home says.

**Proposal:** Measure against the wallets the send flow will actually offer. `useSendWallets()` already applies the gate, waits on the region (`loading` covers `isRegionPending`) and handles both custody shapes; swapping it in for `useActiveWallet()` and folding its `loading` into `isLoading` makes a gated investor read as bitcoin-only, so the split case — and its working Convert path — is what they get instead of a dead Ready. Its `WalletBalance` carries `balance: number` per `walletCurrency`, so the conversion becomes `toBtcMoneyAmount`/`toUsdMoneyAmount` by currency. Note that the current specs pin the gate-blind reading as correct (Gap 2), so they change with the fix.

### Finding 3 — The split-funds Convert opens in the direction the payment does not want, on a combined balance the spread will eat into [NON-BLOCKING]

**When it comes up:** An unrestricted investor holds $13,000 in dollars and $12,000 in bitcoin against a $25,000 investment, so the card says "your funds are split" and offers Convert. `isUsdToBtcOnly` is false for them, so `resolveInitialConvertWallets(btc, usd, false)` opens the screen **BTC → USD** ([migration-convert-wallets.ts:23-25](app/screens/conversion-flow/migration-convert-wallets.ts:23)) and `shouldPrefillFullBalance` fills 100 % of the *from* wallet ([conversion-details-screen.tsx:200](app/screens/conversion-flow/conversion-details-screen.tsx:200)) — "convert your whole bitcoin balance into dollars", when the invoice waiting for them is denominated in satoshis. Worse, `isSplitAcrossWallets` is `combinedUsd >= totalUsd` ([investment-terms.ts:124](app/screens/card-screen/onboarding/investment-flow/investment-terms.ts:124)), an equality the conversion spread cannot survive: an investor holding exactly $25,000 across the two converts, lands short, and the card they were following silently changes from "Convert" to "deposit the remaining amount".

**Proposal:** Open the investment arm USD → BTC for everyone, not only for the restricted investor — the payment is a sats invoice, and that is the direction that leaves it payable without a second spread at send time. Then give `isSplitAcrossWallets` a margin (`combinedUsd >= totalUsd * (1 + SPREAD)`, with the constant named and documented beside `INVOICE_EXPIRY_MINUTES`) so Convert is only offered when it can actually finish the job, and let the DepositPending/Insufficient branch take the boundary case. One `resolveInvestmentFunding` case at the new boundary and one convert-screen case for the unrestricted direction pin both halves.

### Finding 4 — The 24-hour record is the only thing preventing a second agreement, and it lapses on the signed-but-unpaid too [NON-BLOCKING]

**When it comes up:** `CARD_INVESTMENT_LIFETIME_MS` is 24 h from the latest moment ([card-investment.ts:10](app/store/persistent-state/card-investment.ts:10)), and no backend records the investment. An investor who signs on Friday evening and does not pay has, by Saturday evening, a record that reads as nothing: the home bulletin disappears, and the transfer step — reached from a notification or `blink://card/onboarding/investment` — sends them home as having "nothing to pay for" ([transfer-invest-screen.tsx:86-89](app/screens/card-screen/onboarding/investment-flow/transfer-invest-screen.tsx:86)). Their only way on is to walk the flow again and sign a **second** subscription agreement. The mirror case is worse: a paid investor who dismisses the Shareholder card (`clear()`) or simply waits out the day can re-enter and sign again, because `start` overwrites unconditionally and the `paidAt` guard on the welcome step ([welcome-invest-screen.tsx:57](app/screens/card-screen/onboarding/investment-flow/welcome-invest-screen.tsx:57)) has no record left to read.

**Proposal:** Split the lifetime from the duplicate guard. Keep 24 h for what it is documented to cover — the invitation, and the payability of the invoice — but keep a small terminal marker per account (the signed moment, or an `agreementId` once the mint returns one) that never lapses, and let the welcome step refuse to re-enter the flow when one exists. Until then, at minimum stop counting a *signed* record's life down at all, so an agreement stays payable for as long as the envelope is; and put the e-sign envelope id on the record so the eventual server-side check has something to reconcile against.

### Finding 5 — Any pending incoming transaction, of any size, replaces the actionable shortfall card with "Hold on" [NON-BLOCKING]

**When it comes up:** `hasPendingCustodialReceive` is `(pendingIncomingTransactions?.length ?? 0) > 0` ([home-screen.tsx:368](app/screens/home-screen/home-screen.tsx:368)) — any pending on-chain receive, into either wallet, for any amount. An investor who is $20,000 short of a $25,000 investment and happens to have a $5 on-chain receive confirming sees the **DepositPending** card: "Hold on — We are waiting for your deposit to settle", with no button and no figure, outranking the Insufficient card that would have told them the shortfall and offered Deposit ([use-card-investment-bulletin.ts:51](app/screens/card-screen/onboarding/investment-flow/use-card-investment-bulletin.ts:51)). The card they are told to wait on will not cover anything, and tapping it opens a screen whose only control is "Okay".

**Proposal:** Give the home the amount, not just the count, and only claim a deposit is on its way when it would close the gap. The home already resolves the pending amount for the balance row (`usePendingReceiveAmount`), so pass that figure into `useCardInvestmentBulletin` alongside the flag and let `resolveCardInvestmentBulletin` take the DepositPending branch only when `balanceUsd + pendingUsd >= totalUsd`. Anything smaller should still read as Insufficient, which is the card that can actually be acted on.

### Finding 6 — A payment that is only pending is recorded as paid, and nothing ever walks it back [NON-BLOCKING]

**When it comes up:** The receipt marks the investment paid on any status that reaches it, `PENDING` included ([send-bitcoin-completed-screen.tsx:311-321](app/screens/send-bitcoin-screen/send-bitcoin-completed-screen.tsx:311)), and `markPaid` is first-mark-wins with no inverse. A lightning payment that sits pending and then fails — a hold invoice, an HTLC that times out — leaves the investor with "Welcome as Blink shareholder!" on their home, the money still in their wallet, and no route back into the flow: every entry point now resumes on `paidAt` and sends them to the Shareholder card. The comment at the call site names the trade-off honestly, but the branch chose it against a nag, not against a wrong welcome.

**Proposal:** Record the two states separately — a `payingAt` stamp that silences the nag (which is all the comment's rationale actually requires) and a `paidAt` that only a settled transaction sets. The bulletin then renders a quiet "payment on its way" for `payingAt` without `paidAt`, and the Shareholder card waits for the real thing. Reconciling `payingAt` is one `verifyPaymentSettled` call against the recorded invoice on the next home mount, which Finding 1 wants in the tree anyway.

### Finding 7 — `cardInvestmentByAccountId` now outlives logout and is never pruned [NON-BLOCKING]

**When it comes up:** `resetState` — the logout path, and its only caller ([use-logout.ts:117](app/hooks/use-logout.ts:117)) — now carries `cardInvestmentByAccountId` through ([index.tsx:382-390](app/store/persistent-state/index.tsx:382)), while everything else starts over. Nothing removes an entry except the investor dismissing the Shareholder card: `getCardInvestment` filters a lapsed record at read time but leaves it on disk, so every account that ever opened the invitation on the device leaves a permanent row holding a bolt11 invoice, a signed amount and timestamps. On a shared or resold phone that survives an explicit logout and an account deletion, and the map only grows.

**Proposal:** Keep the preservation — the reasoning for it is sound — but prune on the way through. `resetState` is the natural place: carry forward only entries `isCardInvestmentCurrent` still accepts, which drops everything lapsed at exactly the moment the rest of the state is being cleared, and makes the retained set as small as the feature needs. An entry whose `paidAt` is older than the lifetime should go too, since the Shareholder card it existed for can no longer render.

---

## Test Coverage Gaps

1. **Nothing pins the investment bulletin above the server bulletins.** `47cab1394` exists for that placement, but `home.spec.tsx` mocks both `useCardInvestmentBulletin` and `CardInvestmentBulletin` (returning `null`), so swapping the card below `<BulletinsCard>` leaves the suite **113/113 green** — the one mutation of fourteen that survived. Rendering a real marker from the mock and asserting its index among the home's children closes it.

2. **No spec drives the funding hook with the dollar-balance gate closed — and the existing ones pin the gate-blind reading as correct.** `use-investment-funding.spec.ts` mocks `useActiveWallet` and never a gate; narrowing the hook to the bitcoin wallet fails 6 cases across two suites, which means Finding 2's fix cannot land without rewriting them. A case with the gate closed and the Dollar wallet the fuller one (expect `balanceCurrency: Btc`, `isSplitAcrossWallets: false`) is what should be pinning it instead.

3. **The replacement-mint path is untested against an invoice that was already paid.** `transfer-invest-screen.spec.tsx` covers reuse inside the window and minting outside it, but nothing drives "a recorded invoice exists, reuse has expired, and it settled" — the whole of Finding 1. Once the verify call is added, the three outcomes (settled → receipt, unpaid → mint, ledger silent → mint) each need a case.

4. **`home.spec.tsx` pins the behaviour Finding 5 questions.** "tells the hook a deposit is pending when a custodial receive is confirming" asserts `hasPendingDeposit: true` from a single `pendingOnchainReceiveTx` with no reference to its amount, so the spec would have to change for the shortfall-aware version — worth knowing before the fix, not after.

5. **The `isInvestmentInvoice` seam is mocked on both sides in the send flow.** The hook's own spec covers the case-insensitive compare, but `send-confirmation.spec.tsx` and `send-bitcoin-completed-screen.spec.tsx` each stub `useCardInvestmentProgress` with strict `===`, so nothing proves the real pairing survives whatever casing the destination parser hands back — which is the one reason the `toLowerCase` exists.

6. **Nothing covers the conversion round trip end to end.** `conversion-success-screen.spec.tsx` pins the `Investment` return and `insufficient-balance-screen.spec.tsx` pins the arming, but no spec walks shortfall → convert → success → transfer step on a real stack the way `insufficient-balance-screen.stack.spec.tsx` does for the deposit path, so the post-convert landing (and Finding 3's short-by-the-spread outcome) is unobserved.

7. **`refetchAccount` is never exercised on a query that is skipped.** The welcome step's "try again" calls `refetch()` on a query whose `skip` may still be true (no session, or not yet eligible); the spec asserts the call and that a rejection is swallowed, but not what a refetch on a skipped query does to the wait it is supposed to end.

---

## Code Smells

1. **"Keep the reset to the home in one place" left two of the three in place.** `eab3553c1` extracts `RESET_TO_HOME` ([reset-to-home.ts:4](app/navigation/reset-to-home.ts:4)), but both region guards still build the identical action inline — [use-dollar-balance-restriction-guard.ts:38](app/hooks/use-dollar-balance-restriction-guard.ts:38) and [use-transfer-blocked-guard.ts:38](app/hooks/use-transfer-blocked-guard.ts:38) — in files this very range edits. They are the two most likely to drift, since they are the ones that fire without the user asking.

2. **The generic send flow now knows what a card investment is.** Both `send-bitcoin-confirmation-screen.tsx` and `send-bitcoin-completed-screen.tsx` mount `useCardInvestmentProgress()` and branch on it, so a card-onboarding concern sits inside the screen every payment in the app passes through, and its Apollo query is mounted on both. A small "payment observers" registration — a hook the send flow calls with `{paymentRequest, status}` and the investment supplies one implementation of — would keep the coupling out of the screens.

3. **The welcome step imports the e-sign library to word a timeout that has nothing to do with signing.** [welcome-invest-screen.tsx:6](app/screens/card-screen/onboarding/investment-flow/welcome-invest-screen.tsx:6) pulls `getErrorMessage` from `@blinkbitcoin/esign-react-native/webform` to render `LOST_CONNECTION_CODE` while it waits for the *account query*. The app has its own copy for a lost connection; borrowing a signing library's string ties an unrelated screen to a package version bump.

4. **Two of the six bulletins have an action and no button.** `Invited` and `DepositPending` pass `action` without `buttonLabel`, so `NotificationCardUI` renders a tappable card with no visible control, while Insufficient, SplitFunds and Ready all get a `GaloyPrimaryButton` ([card-investment-bulletin.tsx:49](app/components/card-investment-bulletin/card-investment-bulletin.tsx:49) and `:80`). The specs describe it as "nothing to press but the card", which reads as deliberate, but the inconsistency is what a user notices.

5. **`useCardInvestmentProgress` reads the clock during render.** `getCardInvestment(persistentState, accountId, Date.now())` at [use-card-investment-progress.ts:118-120](app/hooks/use-card-investment-progress.ts:118) makes the hook impure and means a lapse only takes effect on the next render for whatever reason. The transfer step already compensates by re-reading the clock in `handleNext`; the home has no such compensation, so a home left open past the boundary keeps pointing at a step that will bounce it.

6. **The transfer step draws its figures from the route param while validating the record.** `resolveInvestmentTerms(selectedAmountUsd)` and the invoice's `amountUsd` both come from the route, while `owedSats` and the lapse check come from `progress` ([transfer-invest-screen.tsx:61-62](app/screens/card-screen/onboarding/investment-flow/transfer-invest-screen.tsx:61)). Every current caller passes the record's own amount and the screen is not deep-linkable, so the two cannot diverge today — which is exactly why an assertion (or simply reading `progress.selectedAmountUsd` once the record is in hand) costs nothing now and prevents a mismatched external id later.

7. **`resolveInvoiceExternalId` defeats the ledger's uniqueness on purpose.** The `mintedAt` stamp is documented as avoiding a permanent refusal for a repeat mint ([use-investment-invoice.ts:74-78](app/screens/card-screen/onboarding/investment-flow/use-investment-invoice.ts:74)), which is a real problem — but the fix removes the only server-side duplicate guard in the feature (Finding 1). If the id must stay unique per mint, the reconciliation has to live on the client; the comment should say which side owns it.

8. **`isSound`'s invitation branch optional-chains a value its own type says is an object.** `Number.isFinite(record?.invitedAt)` at [card-investment.ts:25](app/store/persistent-state/card-investment.ts:25) guards against a non-object off disk, which is right — but `isSignedCardInvestment` already does that check two lines up and the parameter is typed `CardInvestmentRecord`. A single `typeof record !== "object" || record === null` at the top of `isSound` would say what is actually being defended against.

9. **The debt is compared in dollars after a round trip through the price feed.** `owedUsd` converts the signed satoshis back to USD and compares against balances converted the same way ([use-investment-funding.ts:46-53](app/screens/card-screen/onboarding/investment-flow/use-investment-funding.ts:46)) — the rate cancels, so for a bitcoin wallet it is a sats-vs-sats comparison taking a detour through cents. The doc explains why the comparison is *in dollars*; it does not explain why the bitcoin side is not simply `walletSats >= settlementSats`, which is exact and needs no feed. Either way the routing fee is still unaccounted for, so an investor holding exactly the settlement amount passes here and is refused at the send step.

kngako (changes requested):
Review above.

PR #4280, opened 2026-09-16
fix(region): only block the Dollar Balance when the server confirms the region is restricted

### LINKED ISSUE #4278: fix(region): users in allowed countries are told Dollar Balance is not available in their region
## Problem

An eligible custodial account is intermittently shown the modal **“Dollar Balance is not available in your region”**, even though the account's verified phone country (`BI`, as reported) is not in the Dollar Balance restricted-region list.

The affected user reported seeing this for two days. A separate report says the modal appeared once and disappeared after restarting the app, suggesting the failure may be intermittent or session-dependent.

This incorrectly blocks or prompts a transfer from Dollar Balance to Bitcoin Balance for an account that should retain Dollar Balance access.

## Expected behavior

For custodial accounts, Dollar Balance eligibility should be derived from the verified phone country. If that country is not restricted, the regional-restriction modal must not appear and Dollar Balance must remain available.

## Investigation notes

- Check the affected device/app logs to determine whether the region/eligibility call to the backend succeeded.
- Confirm what region value and source the app used when it displayed the modal.
- Check whether startup, session refresh, stale local state, or a failed backend response can incorrectly activate the restriction.

## Acceptance criteria

- [ ] An account whose verified phone country is not in the Dollar Balance restricted-region list does not see the restriction modal.
- [ ] The app handles region/eligibility backend failures without incorrectly classifying an eligible account as restricted.
- [ ] Restarting or refreshing a session does not change Dollar Balance eligibility unless the authoritative eligibility data changes.
- [ ] Add regression coverage for this failure mode.

## Evidence

The supplied screenshot was cropped and account balances and identifying discussion were removed before publication.

Dollar Balance regional restriction shown to an eligible account

## Related work

- https://github.com/blinkbitcoin/blink-mobile/issues/3924 — broader region-determination implementation; this ticket tracks the specific incorrect restriction observed after that work.

## Source

Requested by `@Esaú` in Slack. The exact message permalink was unavailable in the supplied context. Original report and discussion were provided as the attached screenshot.

esaugomez31:
Closes #4278

When the `custodialRestrictions` query failed, the app took the missing answer as a restricted region and opened the "Dollar Balance is not available in your region" modal. That modal has no close button, so a user in an allowed country could only get past it by converting their dollar balance to bitcoin. The retries were counted as they were sent, so the app gave up while the last one was still in flight, and it did not ask again until the app came back from the background or was restarted.

The backend answers correctly for the reported account: BI is not on the production block list.

### Changes

- The query moves into `app/custodial/providers/restrictions.tsx`, shared by every screen and mounted beside the custodial wallet provider. Home alone was mounting four copies of it, each with its own retry timers.
- A failed request is retried after 1, 2 and 4 seconds, on top of the transport's own retries, and each failure only counts once its retry has come back. Once the retries are spent the verdict reads Unknown and the app polls once a minute, in the foreground only, until an answer lands; the return to the foreground and a pull to refresh ask again as well. A verdict the server already served is kept when a later refetch fails, and is not retried.
- The verdict keeps a served answer apart from a failed request. Without an answer the dollar balance is still gated, per the unknown region policy in the PRD, but the forced conversion only opens when the server said the account is restricted.
- The two availability gates expose why they are closed (`reason`: Anon, region, or unknown region), so no surface derives it on its own. An unanswered query no longer reads as a restriction: the dollar row says "couldn't check availability, pull down to retry" and the Transfer explanation becomes "Couldn't check Dollar Balance availability", with a body that says how to ask again. A served restriction, Anon and a sanctioned region keep the copy they had. Three new keys, translated in every locale.
- Pull to refresh on home asks for the restrictions again, without holding the spinner on the answer.
- A query that stays unanswered is logged once per outage, with a copy of the error so the one Apollo keeps is left untouched. A network failure only leaves a Crashlytics breadcrumb, while other failures are recorded as non-fatal errors.
- `useBackoffRetry` moves from the self-custodial hooks to `app/hooks` so the provider can reuse it. `schedule` now returns whether a retry was scheduled.
- `useDollarBalanceRestricted` is removed, since home was its last consumer.

### Notes

- **New network pattern.** While the verdict is Unknown the app sends one `custodialRestrictions` request a minute, foreground only, `no-cache`, and stops on the first answer. Nothing else in the app polls this query. No migration.
- **Pull to refresh no longer waits on the verdict.** The spinner retracts with the balances; the verdict lands on its own. A pull still asks.
- **New copy for the unanswered state.** Three keys under `StablesatsRestriction` and `DollarBalanceRestriction`, translated in all 28 locales. The served-restricted, Anon and sanctioned-region copy is unchanged, and nothing new shows while the verdict is pending.
- **Gate API.** `useDollarBalanceGate` and `useTransferGate` now return `reason` (`anon` / `region` / `unknownRegion`, or null). Existing consumers that read only `isGated` and `isRegionPending` are unaffected.
- **Upgrade behaviour for QA.** A device that could not reach the server used to show "not available in your region" after the retries; it now shows the retry label and recovers within a minute of connectivity returning, without a pull.

### Test plan

- [x] Provider tests against a real Apollo client with a scripted link: the last retry still in flight when the old limit would have run out, the `no-cache` policy against a verdict a previous session cached, a new client under the same provider after Unknown and after a served answer, pulls that succeed or fail during the backoff, the hand-over from the backoff to the poll, the poll stopping on an answer, in standby and in the background, and a foreground re-ask folding into a poll in flight.
- [x] Each gate asserted per reason; the wallet overview, the restriction modal and the home Transfer tap rendered for the Unknown verdict; the three copy keys pinned per locale.
- [x] iOS simulator, with a temporary link forcing the query to fail or to answer restricted: the failing query no longer opens the modal and leaves the dollar balance gated, the connection toast stops once the retries are spent, a restricted answer still opens the modal, and a pull to refresh brings the balance back once the query passes.
- [ ] Airplane-mode launch: after the retries the dollar row reads "couldn't check availability, pull down to retry" and Transfer opens "Couldn't check Dollar Balance availability".
- [ ] Connectivity back while sitting on home: the row recovers within a minute without a pull.
- [ ] Pull to refresh with the query hanging: the spinner retracts with the balances.

### Screenshots

| Before: query failing | After: query failing | After: server answers restricted | After: pull to refresh once the query passes |
|---|---|---|---|
| before-query-failing | 1-query-failing-no-modal | 2-server-restricted-modal-shown | 3-recovered-after-pull-to-refresh |


kngako (changes requested):
Claude deemed finding 1 non-blocking but I am upgrading it to blocking as the UX would be inconsistent without it being addressed.

---

## Findings

### Finding 1 — An Unknown verdict still reads "not available in your region" everywhere but the forced conversion [non-blocking (upgraded to blocking) — recommend closing in this PR]

The branch establishes, and names, that `Unknown` "never [means] that the server restricted anything" (`app/types/account.ts`), and threads `isRegionDetermined` to exactly one consumer (`home-screen.tsx:497`). Every other surface maps `Unknown` to `RESTRICTED_UNKNOWN_REGION` (`use-account-restrictions.ts:82`) and then reads `isGated` with no way to tell the two apart: the wallet overview labels the dollar row `LL.StablesatsRestriction.walletLabel()` = "not available in your region" (`wallet-overview.tsx:93`), the disabled Transfer button opens `DollarBalanceRestriction.modalTitle` = "Dollar Balance is not available in your region" (`home-screen.tsx:744-745`), and the migration preview mutes the current dollar row the same way (`balances-overview-screen.tsx:291`). **When it comes up:** a user opens the app on a train or behind a captive Wi-Fi portal; the four attempts fail within roughly 25 s (see Smell 3) while their BTC balance renders fine from the persisted cache. They now read that Blink has decided their region is restricted, which is false, with no hint that pulling down would ask again — and the same copy is what a genuinely restricted user sees, so support cannot tell the two reports apart either. Gating is the PRD's FAIL_CLOSED call and is right; the copy is the branch's own stated target ("stop treating an unanswered restrictions query as a restricted region") not yet reached.

**Proposal:** Carry `isRegionDetermined` through the two gates (`useDollarBalanceGate` at `use-dollar-balance-restricted.ts:38`, `useTransferGate` at `use-transfer-blocked.ts:19`) — a `reason: "anon" | "region" | "unknown"` on the gate would also retire the `isAnonMode` re-derivation the wallet overview does on its own. Then give the Unknown case its own copy in `WalletOverview`'s `unavailableLabel` and the restriction modal, something like "couldn't check availability — pull down to retry", added as a new key to `en/index.ts` and all 28 locale JSONs (the parity spec will hold the line). The home `refetch` already re-asks, so the copy only needs to say so. If product prefers one label for both, at least drop "in your region" for the undetermined case.

### Finding 2 — Once the burst is spent nothing re-asks while the app stays in the foreground [non-blocking — a design decision worth stating in the commit]

`281267dcb` replaces the first version's endless 60 s-capped loop with three retries, after which `haveRetriesRunOut` pins the verdict at Unknown and only a foreground transition (`app-state.tsx:120`) or a pull (`home-screen.tsx:609`) asks again; a successful answer is the only thing that resets the budget (`restrictions.tsx:118-123`). The app has no connectivity listener and the restrictions query never polls, so there is no third path. **When it comes up:** the same user as Finding 1 gets signal back thirty seconds later while sitting on the home screen. `HomeAuthed` and the price feed recover through their own refetch paths, so the screen visibly comes back to life — but the dollar row stays "not available in your region" until they happen to pull or switch apps, which nothing prompts. Compared with `main` this is not a regression (the old hook also stopped after three), but the branch first added self-healing and then removed it without recording why.

**Proposal:** Keep the bounded burst — it is the right answer to a server that is down — and add one slow lane for the Unknown state only: `startPolling(60_000)` while `verdict.status === Unknown`, `stopPolling()` on anything else (Apollo polls a `no-cache` query without touching the cache, and the log effect already reports once per episode so a poll adds no noise). If polling is unwanted, a `NetInfo` `isInternetReachable` false→true transition is the other natural trigger and could live next to the foreground re-ask in `AppStateWrapper`. Either way, say in the constant's doc comment (`restrictions.tsx:32-33`) whether "asks again on foreground and on pull" is the whole recovery story on purpose, and pair it with Finding 1's copy so the user knows how to trigger it.

### Finding 3 — Every re-ask folds into a request that has not returned, so a hung first request cannot be rescued [non-blocking]

Both re-asks the branch relies on go through `ObservableQuery.refetch`, and Apollo's default `queryDeduplication` joins a refetch to any in-flight request for the same document instead of sending a new one; nothing in the client (`app/graphql/client.tsx`) gives the HTTP link a timeout, and React Native's Android OkHttp client is built with none. Reproduced in the scratch spec: with the first request held open, a context `refetch()` plus a `client.refetchQueries` re-ask leave `requestCount` at 1, the verdict at `Pending`, and the pull's promise unsettled. **When it comes up:** a Wi-Fi→cellular handover during launch leaves the first request on a half-open socket. Every dollar surface then holds on `isRegionPending` (spinner, Transfer disabled with no tap handler), and because the home `refetch` now `Promise.all`s the restrictions refetch, the pull-to-refresh spinner is pinned as well — until the OS gives up on the socket, which on Android can be minutes. The spec's own "does not give up while the last retry is still on its way" case is the same shape with the held request as the last retry.

**Proposal:** Give the re-ask a way to bypass the stuck request: `refetch` accepts no context, but `client.query({ query: CustodialRestrictionsDocument, fetchPolicy: "no-cache", context: { queryDeduplication: false } })` does, and its result can be fed to the observable via `updateQuery`, or more simply the provider can own an `AbortController` per attempt and pass `context: { fetchOptions: { signal } }` with a timeout (`apollo-link-timeout` is the packaged form) so a hung request errors into the normal retry path instead of blocking it. In the home `refetch`, don't let the verdict hold the spinner: `void refetchCustodialRestrictions()` alongside the awaited balance refetches, since the verdict lands whenever it lands. This is pre-existing transport behaviour, so a follow-up is fine — but it is the branch's recovery story that depends on it.

---

## Test Coverage Gaps

1. **The `no-cache` policy lost its guard.** `main`'s `use-account-restrictions.spec.ts:159` asserted `fetchPolicy: "no-cache"`; the provider spec drives a real client and never checks it. With `CachePersistor` restoring the cache on launch, a `cache-first` regression would serve last session's verdict before the network answers. One case: `client.writeQuery` a restricted verdict into the cache, script the link to answer unrestricted, assert the verdict follows the link and `requestCount === 1`.
2. **The foreground re-ask has only a negative assertion.** `app-state.spec.tsx:380` checks a signed-out user does *not* refetch `custodialRestrictions`; nothing asserts a signed-in one does, although `restrictions.tsx:32-33` and Finding 2 rest on that line (`app-state.tsx:120`). The existing `refetchedOperations()` helper makes it a three-line case.
3. **Login/logout is modelled as `skip` on one client, which the app never does.** `GaloyClient` rebuilds the `ApolloClient` per token, so the real path is a new client under the same mounted provider; that path resets the backoff and must not leak the previous client's `data` into a failed first request. My scratch cases (swap after Unknown, swap after Served with the next first request failing) pass and would drop straight into `restrictions.spec.tsx` beside "ignores a retry that returns after the question was dropped" (`:411`).
4. **A pull during the backoff is untested.** Neither "a pull that succeeds mid-backoff settles Served and cancels the remaining retries" nor "a pull that fails mid-backoff neither spends nor cancels a retry" is in the spec; both pass today (scratch), and both are exactly what a user does when the first attempt fails.
5. **The reason for the copied error is not pinned.** `restrictions.spec.tsx` mocks `logError`, so nothing shows the query's own `ApolloError` keeps its message after reporting. A case mocking `recordAppError` instead and asserting `error.message` on the hook result is still "Network request failed" would stop someone "simplifying" `toReportedError` away.
6. **Unknown is indistinguishable from Served-restricted in every surface spec** (`wallet-overview.spec.tsx`, `home.spec.tsx`'s transfer and modal cases). Whichever way Finding 1 is resolved, a case that renders the Unknown verdict and asserts the label is the place to record the decision.

## Code Smells

1. **`refetchVerdict` is a `.then(() => undefined, () => undefined)` inside an `async` that returns nothing** (`restrictions.tsx:191-197`). `await refetch().catch(() => undefined)` says the same thing in the repo's usual idiom (`home-screen.tsx:631-642` swallows the pull the same way).
2. **`toReportedError` copies an error to dodge `logError` mutating it** (`restrictions.tsx:67-72` working around `log-error.ts:19`, which prefixes `formatted.message` on the instance `toError` hands back unchanged). Every caller that passes a live error it still uses has the same exposure; a one-line fix in `logError` (`new Error` with the prefixed message, name carried over) would let this branch drop the helper and its test. Worth a separate PR.
3. **Two backoffs are stacked and the comment credits only one.** The link chain already retries each request up to 5 times (`client.tsx:163`, `RetryLink` default 300 ms jittered doubling), so the provider's three retries mean roughly 20 HTTP attempts and ~25 s before Unknown, while `restrictions.tsx:32` says the provider's delays are what keep a down server from being hammered. Either name the link-level retry in the comment or shorten the provider's list now that the transport does the fast retries.
4. **The home screen reads three hooks for one concern** — `useDollarBalanceRestriction()`, `useDollarBalanceGate()` and `useCustodialRestrictions()` (`home-screen.tsx:430-436`), the first two each running `useAccountRestrictions` and its device/IP/remote-config hooks. `isRegionDetermined` on the gate (Finding 1) and `refetch` on the gate would leave one.
5. **The screen test harness has no provider, so a real hook waits forever.** `ContextForScreen` (`__tests__/screens/helper.tsx:71-76`) mirrors `IsAuthedContextProvider` and `AccountRegistryProvider` but not `CustodialRestrictionsProvider`, and `app/graphql/mocks.ts` has no `custodialRestrictions` mock, so any screen reaching `useAccountRestrictions` unmocked sits at `Pending` with a silent no-op `refetch`. The branch worked around it per spec (`conversion-details-screen.spec.tsx:55`, `home.spec.tsx:175`); adding the provider plus a served mock to the harness would stop the next screen spec rediscovering it.
6. **The first commit's title promises more than it delivers.** "stop treating an unanswered restrictions query as a restricted region" — the query still gates every feature as `RESTRICTED_UNKNOWN_REGION`; what stops is the forced conversion. "…as grounds for the forced conversion" is what the diff does, and it is a good change; the mismatch is what Finding 1 grows out of.
7. **`CustodialRestrictionsProvider` is nested under `SelfCustodialWalletProvider`** (`app.tsx:76-77`) though it reads only the registry, auth and Apollo; it could sit beside `CustodialWalletProvider`, which is where its file lives. Cosmetic.

esaugomez31:
**Findings**

1. Done. Both gates expose `reason` (`anon` / `region` / `unknownRegion`) via one shared `toGateReason`; the wallet overview and the restriction modal read it. Unknown shows "couldn't check availability, pull down to retry" and "Couldn't check Dollar Balance availability" with a body. Served, Anon and the sanctions block keep today's copy, and nothing new shows while Pending. Three keys, 28 locales.
2. Done. Once the retries are spent the provider polls once a minute while the verdict is Unknown, `no-cache`, foreground only (`skipPollAttempt`), and stops on an answer, on a lost account and in the background. The constant's comment now states the whole recovery story.
3. The pull no longer waits on the verdict. The link timeout is pre-existing transport behaviour, as you note; follow-up.

**Test coverage gaps**

1. Done: a cached restricted verdict is ignored and the link's answer wins with one request.
2. Done: the foreground re-ask is asserted for a signed-in user.
3. Done: a new client under the same provider, after Unknown and after a served answer whose next first request fails.
4. Done: a pull that succeeds mid-backoff settles Served and cancels the retries; one that fails neither spends nor cancels one.
5. The context exposes no `error`, so the assertion cannot be written as proposed. The existing case at 344 already fails if the copy is removed, since it asserts the reported error is not the `ApolloError`.
6. Done: the wallet overview, the modal and the home transfer tap are rendered for the Unknown verdict.

**Code smells**

1. Done.
2. Agreed, separate PR.
3. Done: the comment names the `RetryLink` bursts and the real budget (five attempts per request, about 25 s on average, 45 s at worst).
4. `reason` is on the gate now; folding the three home hooks is a follow-up so this PR does not touch the forced-conversion logic.
5. Follow-up: the harness is shared by 114 specs.
6. The title stays: what the diff stops is the forced conversion, and the copy change now reaches the rest.
7. Done.

kngako (dismissed):
No more blocking findings. I'll leave it to @esaugomez31 discretion on whether he wants to address the remaining issues in this PR or a future one.

---

## Findings

### Finding 1 — A request that never returns still strands the dollar surfaces without bound; the poll cannot reach it [non-blocking — follow-up]

`8404a37e8` fixes the visible half of the earlier Finding 3: the pull spinner retracts with the balances. The other half stands. The restrictions request goes through an `HttpLink` with no timeout (`app/graphql/client.tsx`), React Native's Android OkHttp client is built without one, Apollo's `queryDeduplication` folds every re-ask (`refetch`, `client.refetchQueries`, and now the poll — `maybeFetch` returns early while `queryInfo.networkStatus` is in flight, `ObservableQuery.js:561`) into the request already out, and the poll only begins after the backoff has *failed*, which a hung request never does. **When it comes up:** a Wi-Fi→cellular handover during launch leaves the first request on a half-open socket; the verdict stays `Pending`, so the dollar row shows its loader and Transfer is disabled with no tap handler (`home-screen.tsx:755-756`) until the OS gives up on the socket — minutes on Android. `docs/pr-review.md` §2.3 names exactly this: "surface a distinct error immediately or after a bounded wait — never strand the user in loading". This predates the branch and is shared by every query in the app, which is why it is a follow-up rather than a gate; but this branch is the one that built the recovery lanes, and none of them can fire here.

**Proposal:** Give each attempt a bound so a hung attempt turns into the failure the backoff already knows how to handle: an `AbortController` per fetch passed as `context: { fetchOptions: { signal } }` with a 15–20 s timer (cleared on settle), or `apollo-link-timeout` in the chain for this operation. Once a hung attempt errors, the existing path takes over — retry, Unknown, the "pull down to retry" label, the 60 s poll — and nothing else needs to change. Pin it with a case that holds the first request, advances past the bound, and asserts the verdict moves to `Pending`-with-retry and then `Unknown`, i.e. the current "does not give up while the last retry is still on its way" becomes "gives up on a retry that has been on its way for N seconds".

---

## Test Coverage Gaps

1. **"stops polling when the account it asked about is gone" passes without the `stopPolling` cleanup.** With the cleanup mutated to `return undefined`, that case (`restrictions.spec.tsx:500`) stays green; the mutant is caught only by "keeps polling while Unknown and stops the moment an answer lands" (`:424`). The reason is Apollo, not the provider: a query moved to `standby` keeps `queryInfo.networkStatus` in flight, so `maybeFetch` skips every tick whether or not the interval was cleared. The case is true but decorative for its stated purpose; its docstring should say it documents the standby behaviour, or it should be dropped in favour of the one that actually guards the cleanup.
2. **The three new copy keys have no value guard.** `locale-parity` checks keys only; nothing pins the 27 translations or rejects an English leak, and the repo's convention (`__tests__/i18n/*-label.spec.ts`, as the 2026-09-14 review recorded) is to pin copy positively per locale. A `StablesatsRestriction.unknownRegionWalletLabel` table in `__tests__/i18n/` would be the same shape as `anonymous-user-label.spec.ts`. Low priority: the translations are present and non-English today (checked above).
3. **The `toReportedError` reason is still unpinned** (carried from the first review): `logError` is mocked in `restrictions.spec.tsx`, so nothing shows the live `ApolloError`'s message survives reporting.
4. **Poll and foreground re-ask overlapping is untested.** The two can coincide on a resume (the poll timer fires as `AppStateWrapper` refetches); dedup makes it one request, but nothing asserts `requestCount` stays at +1 for the pair.

## Code Smells

1. **The `RetryLink` sentence in the provider's doc comment misstates the per-resend delay** (`restrictions.tsx:37`): "the four resends spaced by a jittered, doubling delay of up to 600 ms" — 600 ms is the *first* resend's ceiling; the four are drawn from `[0, 300·2ⁿ]` = up to 600, 1200, 2400 and 4800 ms (`delayFunction.js`). The totals that follow (≈25 s average, ≈45 s worst) are computed correctly from the real ceilings, so only the clause needs rewording ("starting at up to 600 ms and doubling").
2. **A test title outlived the behaviour it names.** "reads Unknown, without retrying, when the server answers with no verdict" (`restrictions.spec.tsx:289`) now advances `59_999` ms precisely to stop short of the poll that does retry; "without a backoff retry" is what it checks.
3. **`useTransferGate().reason` is computed and read by nothing** (`use-transfer-blocked.ts:26`). The transfer guard resets navigation rather than rendering copy, so the field exists for symmetry with the dollar gate. Fine to keep, but a one-line comment saying so would stop the next reader hunting for its consumer.
4. **The gated-tap ordering comment is now half wrong.** `home-screen.tsx:553` justifies letting the migrate-now nudge outrank the compliance modal because "that modal is a dead end (a title and a Close)" — the Unknown variant now carries a body and a remedy. In the receive-disabled phase (`canReopen` = `isReceiveDisabled && !isSelfCustodialDisabled`) the row says "pull down to retry" while the tap opens the migrate nudge; the label still carries the remedy, so the behaviour is defensible, but the comment should say the nudge wins *even over* the retry explanation, and why.
5. **The home screen still reads three hooks for one concern** (`home-screen.tsx:430-442`; carried). Now that the gate publishes `reason`, `useDollarBalanceRestriction()` is only there for the forced conversion's `isRestricted && isRegionDetermined`, which `reason === GateReason.Region` expresses on its own.
6. **`ContextForScreen` still has no `CustodialRestrictionsProvider`** and `app/graphql/mocks.ts` no `custodialRestrictions` mock (carried): an unmocked screen sits at `Pending` forever in tests.
7. **`toReportedError` still works around `logError` mutating its argument** (carried, `log-error.ts:19`); the fix belongs in `logError`, as a separate PR.
8. **For the PR description** (per `docs/pr-review.md`, which I cannot check from here): a "Notes" entry for each behaviour change — the 60 s poll while Unknown (new network pattern), the pull no longer awaiting the verdict, the new copy — plus before/after screenshots of the row label and modal in the Unknown state, and a test plan that includes the airplane-mode launch.

esaugomez31:
@kngako the PR was already approved, but I decided to apply your remaining recommendations where they could be done safely.

**Finding 1**: follow-up, as you suggest. A per-attempt bound interacts with the `RetryLink` resends and the deduplication, so it deserves its own PR.

**Test coverage gaps**

1. Done: the standby case now says what it documents (Apollo parks the query, so no tick fires) and names the case that guards the cleanup, which fails without it.
2. Done: `__tests__/i18n/unknown-region-copy.spec.ts` pins the three keys per locale with literals, read from the raw files so `id` and `xh` are covered although they are not compiled. It also fails if a translation file is not in the table, or if any locale repeats the English copy.
3. Not written: the context exposes no `error`, and pinning the live message needs the real `logError`, which prints in tests. The existing case still fails if the copy is removed.
4. Done: a foreground re-ask while a poll is on its way stays one request, and the verdict lands when it settles.

**Code smells**

1. Done: the delay clause now reads "starts at up to 600 ms and doubles each time".
2. Done: the title reads "without a backoff retry".
3. Done: `reason` on the transfer gate says it has no copy consumer yet and is there so both gates answer alike.
4. Done: the tap-order comment says the wind-down nudge wins even over the unknown-region explanation, and why.
5. Follow-up: folding the three home hooks touches the forced-conversion logic.
6. Follow-up: the harness is shared by 114 specs.
7. Separate PR, as agreed.
8. Done: the description has a "Notes" entry per behaviour change and a test plan; the three on-device items and the Unknown-state screenshots are pending.

kngako (approved):

Issue #2799, opened 2023-11-20
High onchain fee alert when mistakenly sending onchain


designsats:
Blink needs an alert preventing users to mistakenly send a small onchain transaction which should have been via LN.

**Trigger**: Popup would trigger if users scans or pastes onchain address, fills the amount (which is smaller than 2x of fee) and presses **Continue**. Then they should be alerted and aknowledge they are sure to proceed. **Cancel** shoud be primary button - we assume this was a mistake.

Upon pressing **Cancel** - user should be redirected to first step of send flow with empty destination field.

Pressing **I am sure** would allow them to **Confirm** sending.

![dark theme - high fee alert](https://github.com/GaloyMoney/galoy-mobile/assets/17384787/9bce7ce9-e5fe-4f39-b9c7-f8bfa691984b)

nicolasburtey:
I would not think `Cancel` should be the default. I think "I am sure" should be the default.

designsats:
@nicolasburtey But assuming it is an error, main action should be correcting that error. Instinctive behavior is to go for main call to action. How often have you clicked call to action button in haste?

PR #4282, opened 2026-09-17
feat(send): error message bottom sheet over review when the fee can't be quoted

designsats:
Part of blinkbitcoin/blink-wip#1275 and blinkbitcoin/blink-wip#1323 (high-fee sheet) · Parent blinkbitcoin/blink-wip#1134 · Failure-mode inventory blinkbitcoin/blink-wip#1273

> [!NOTE]
> **Stacked on #4281 (Sent), which is stacked on #4275, #4273 and #4274.** It also carries a copy of `BottomSheet` from #4254, which hasn't merged. GitHub won't take a base branch on the fork, so this targets `main` and shows the stack's commits until they land. This PR's own work is two commits: `feat(ui): bring the shared bottom sheet over from #4254` and `feat(send): open the error message sheet over review when the fee can't be quoted`.

[send-failures-before-after.pdf](https://github.com/user-attachments/files/32407542/send-failures-before-after.pdf)

First slice of the error message sheet. Figma's `error-msg-bottom-sheet` is built once, as a shared component, and wired to one failure so the pattern can be checked on device before the rest of the send flow's errors move onto it. The long-term aim is that an edge case never stops the user's flow: every message ends with an action they can take.

**UI only.** `useFee`, the SDK error classification and `translateSdkError` are unchanged.

Figma: [`error-msg-bottom-sheet` `17226:68949`](https://www.figma.com/design/9MQuQi8ZhXVvDibWSI3C4c/Blink?node-id=17226-68949)

### `ErrorMsgBottomSheet` (`app/components/error-msg-bottom-sheet`)

- Built on the shared `BottomSheet` (modal presentation, `restsOnHeader`), per the #1273 ruling that sheets use that primitive.
- Content is the existing `IconHero` (the `warning` glyph in `colors.warning` on the `grey5` circle, title, body), then one `GaloyPrimaryButton`.
- Props: `isVisible`, `onClose`, `title`, `body`, `primaryLabel`, `onPrimaryPress`, `testID`, plus an optional `secondaryLabel` / `onSecondaryPress` (`GaloySecondaryButton`, 10 under the primary). Without them every sheet renders as before.
- **Rests on its content** and **can't be dragged up** (`expandable={false}`). The messages are short, so a full-height state would only be empty space.
- Closes on a swipe down, a scrim tap or Android back. Closing doesn't run the primary action.
- **CTA gap:** 20 above the bottom safe area. The inset comes from a `SafeAreaView` inside the sheet rather than `useSafeAreaInsets()` from the app root. With the root value, the CTA sat about 25dp higher on the Android emulator: inside the sheet's `Modal` window, the root inset overstates the space it needs.

### `BottomSheet` copy

`app/components/bottom-sheet/index.tsx` is byte-identical to #4254's head, `368e6eca02`. Its `5e52425d64` adds `expandable` with two specs, one of which fails when the change is reverted, and `368e6eca02` makes the modal translucent under both system bars. When #4254 merges, the copy drops out of this diff on the next merge from `main`, since both sides add the same content.

### First use: R2 on review

- When the **self-custodial** fee quote fails with a classified SDK code (`isSelfCustodialErrorCode`), the sheet opens over review:
  - Title: **"A small problem"** (new key `SendBitcoinScreen.problemSheetTitle`, the default title from #1275)
  - Body: the translated SDK message, **word for word**, e.g. `SelfCustodialError.generic` "The payment could not be completed. Try again in a moment, or use a smaller amount."
  - Action: **"Change amount"** (existing `SendBitcoinConfirmationScreen.changeAmount`), which calls `navigation.goBack()` to amount entry
- **The inline error stays.** The red outline on Details, the inline message and the slider state are exactly as on #4281. The sheet opens on top of them.
- `useDismissibleErrorMsg` keeps a dismissed sheet closed while the same failure holds. Once the failure clears, the next one opens it again.
- Custodial fee failures carry raw GraphQL text, so they don't open the sheet and keep today's inline "Unable to calculate fee".

### High-fee sheet (blink-wip#1323)

Replaces the custodial on-chain `ConfirmFeesModal` popup with the error message sheet, in **both modes**. Three commits: `feat(ui): let the error message sheet offer a secondary action`, `feat(send): high-fee error message sheet on every rail, both modes` and `feat(send): "Accept fee" and no high-fee warning under 100 sats`.

- **Rule** (`isHighFee` in `hooks/high-fee-alert.ts`, renamed from `onchain-fee-alert.ts`): the fee is **50% or more** of the amount sent (fee × 2 ≥ amount), both in sats. A free send never warns, and **neither does a send under 100 sats**. The old check was strict, so exactly 50% now warns.
- **On-chain** (custodial and self-custodial): opens on **Next** from amount entry, once the picked tier's fee is quoted. Self-custodial on-chain is new.
- **Lightning** (invoice, Lightning address, LNURL; custodial and self-custodial): opens on **review** once the fee quote is `set`, since that's the first place its fee is known (`review/review-high-fee-sheet.tsx`). On-chain isn't warned again there. A blocking error, the error sheet or an attempted send wins over it.
- **Copy:** title "High fee alert!" (existing translations kept), body "Your fee is 50% or more of the amount you're sending. Are you sure you want to proceed?", primary **Cancel payment**, secondary **Accept fee**. Icon is the shared `warning` glyph (Figma's "alert").
- **Cancel payment** goes to step 1 with nothing entered (`startOver`, as in #2799). **Accept fee** continues to review, or closes the sheet on review. Swiping it down on amount entry stays there.
- `ConfirmFeesModal` and its `confirmFeesModal` strings are deleted. The new `SendBitcoinScreen.highFeeSheet` keys are in **all 28 locales**.
- Figma [`highfee-error-msg-bottom-sheet` `24793:141099`](https://www.figma.com/design/9MQuQi8ZhXVvDibWSI3C4c/Blink?node-id=24793-141099) differs: its primary is "Send via Lightning" and its secondary "Accept $2.89 fee". Both were overridden (ruled 2026-09-18) and are listed in #1323's *Figma changes needed*.

## Rulings

| Ruling | Shipped |
|---|---|
| Component name follows Figma | `ErrorMsgBottomSheet` / `error-msg-bottom-sheet`, not "problem sheet". Ticket text still says "problem sheet". The string key keeps the ticket's name |
| **Inline error stays, the sheet opens on top** | Reverses #1275's "inline error vs problem sheet" split and #1273's "review shows no failure text". Dismissing the sheet leaves the error visible |
| No swipe up, since messages are short | `expandable={false}` on `BottomSheet` (pushed to #4254) |
| CTA 20 from the bottom, like every other screen | 20 above the safe area, measured inside the sheet's window |
| **Error wording is out of scope** | The body is the current message, unchanged. #1275 updated (2026-09-17): the S10 `generic` row no longer drops "use a smaller amount", and a ruling says every message keeps today's copy |
| Test case: the S10 `generic` message with a "Try again"-style action that goes back one step | Wired at the **fee quote (R2)**, not the send failure (S10). The same message reaches both, but after a self-custodial send "Change amount" is blocked (N20) and "Try again" is only safe on some routes (N18). Before a send nothing has been paid, so going back is safe on every rail |
| Stack on the newest Send 3.0 PR | Cut from `feat/send-3-sent-animation` (#4281) |

### Off ticket / deferred

| Ticket | This PR |
|---|---|
| R2: Try again (fetches the fee again) · Change amount | Change amount only. A retry needs a `useFee` refetch trigger |
| S3–S10 send failures open the sheet | Not yet. They stay inline, as on #4281 |
| L1–L3, A3 daily limit, N2–N8 server messages | Not yet |
| Secondary action (Close / Home) | The component supports one now (used by the high-fee sheet). No failure sheet uses it yet |
| Body scrolls when taller than the screen | Not supported: the sheet can't expand and the header doesn't scroll. Fine for today's messages, not for raw server text at large OS font sizes |
| Translations | `problemSheetTitle` is **en only**, so locale parity fails CI until the other 27 locales land |

## Acceptance requirements

**Self-custodial, Lightning invoice, Lightning address or Spark**

- [ ] When the fee quote on review fails with a classified SDK error, a sheet opens titled "A small problem", with that error's current message as the body and a "Change amount" button
- [ ] The inline error, the red Details outline and the disabled slider are still on review underneath the sheet
- [ ] "Change amount" closes the sheet and returns to amount entry with the destination kept
- [ ] Swiping down, tapping the scrim or pressing Android back closes the sheet and stays on review with the inline error
- [ ] Dragging the sheet up doesn't move it past its resting position
- [ ] A closed sheet doesn't reopen while the same failure holds, and opens again for the next one after returning from amount entry
- [ ] The CTA sits 20 above the bottom safe area on Android (gesture and 3-button navigation) and iOS

**Custodial**

- [ ] A failed fee quote opens no sheet and shows "Unable to calculate fee" inline, as on #4281

**High fee, both modes** (blink-wip#1323)

- [ ] On-chain, 100 sats or more, fee 50% or more of the amount: Next opens "High fee alert!" instead of going to review
- [ ] Lightning (invoice, Lightning address, LNURL), same fee share: the sheet opens on review once the fee shows
- [ ] Fee just under 50%, or an amount under 100 sats: no sheet, on any rail
- [ ] Intraledger sends never show it, and an on-chain send accepted on amount entry isn't warned again on review
- [ ] **Cancel payment** returns to step 1 with an empty destination
- [ ] **Accept fee** continues to review (on-chain) or closes the sheet with the slider enabled (Lightning)
- [ ] A send failure, a failed fee quote or low funds on review shows instead of the high-fee sheet

## Testing specs

> [!IMPORTANT]
> **POC: no specs for `ErrorMsgBottomSheet`, `useDismissibleErrorMsg` or the review wiring yet.** `expandable` itself is covered on #4254 (`bottom-sheet.spec.tsx`, 27 tests, 2 new). `send-confirmation.spec.tsx` still carries the 7 failures from #4281.

Local: `tsc` and `eslint` clean on the changed files.

`yarn jest --maxWorkers=2 __tests__/components/bottom-sheet/bottom-sheet.spec.tsx`

**High-fee sheet specs** (blink-wip#1323):

| Spec | Covers |
|---|---|
| `__tests__/screens/send-bitcoin-screen/high-fee-alert.spec.ts` | 50% boundary (49 / 50 / 60), 99 vs 100 sats, free send, cents fee weighed in sats, on-chain only on amount entry |
| `__tests__/components/error-msg-bottom-sheet/error-msg-bottom-sheet.spec.tsx` | Secondary action is absent unless given, and runs on its own |
| `__tests__/screens/send-details.spec.tsx` › high-fee sheet | Opens on Next, Accept continues to review, Cancel payment starts over, under 50% goes straight to review |
| `__tests__/screens/send-confirmation.spec.tsx` › high-fee sheet | Custodial and self-custodial Lightning open it, under 50% and on-chain don't, Accept frees the slider, Cancel payment resets to step 1, low funds wins |

`yarn jest --maxWorkers=2 __tests__/screens/send-bitcoin-screen/high-fee-alert.spec.ts __tests__/components/error-msg-bottom-sheet __tests__/screens/send-details.spec.tsx __tests__/screens/send-confirmation.spec.tsx`

The 7 failures in `send-confirmation.spec.tsx` are the ones from #4281. There's no screen-level spec for self-custodial on-chain on amount entry: its quote comes from the SDK hook, and the rule itself has no mode input.

**Manual QA** (a failed fee quote can't be reproduced on demand, so force one locally):

1. In `app/self-custodial/payment-details/lnurl.ts`, add `throw new Error("forced")` as the first line inside the `getFee` `try`. For an invoice or Spark destination, add it inside `createGetFee` in `send-helpers.ts`. Don't commit it.
2. On a self-custodial account: Send → a saved Lightning address → pick 25% → Next.
3. Review opens with the sheet up: "A small problem", "The payment could not be completed. Try again in a moment, or use a smaller amount.", **Change amount**.
4. Drag the sheet up: it stays put. Drag it down: it closes, and the inline error and red outline are still on review.
5. Go back and Next again: the sheet opens again.
6. Tap **Change amount**: amount entry opens with the destination kept.
7. Repeat on iOS and on Android with 3-button navigation, and check the 20 gap under the CTA.

**Manual QA, high-fee sheet:**

1. Custodial: Send → an on-chain address → the smallest amount on-chain accepts → Next (on mainnet fees that's usually enough). "High fee alert!" opens with **Cancel payment** and **Accept fee**.
2. **Accept fee** → review. Go back and Next again → **Cancel payment** → step 1, destination empty.
3. Same with 99 sats: if the amount is accepted, Next goes to review with no sheet.
4. Self-custodial: repeat step 1 on-chain.
5. Self-custodial: Send → a Lightning address → a few sats. Once the fee shows on review the sheet opens. **Accept fee** closes it and the slider sends.


designsats:
## Self-review at `aa10f12d1b`

### Real strengths first
- The sheet is built once on the shared `BottomSheet` and `IconHero`, with no forked sheet or hero (`app/components/error-msg-bottom-sheet/error-msg-bottom-sheet.tsx:44-66`).
- `expandable` is an opt-in prop, so every existing caller is untouched. Its spec fails when the change is reverted (#4254, `5e52425d64`).
- Hook order is safe: `useDismissibleErrorMsg` runs unconditionally, with no early return before it (`send-bitcoin-confirmation-screen.tsx:608`).
- Wording and SDK classification are untouched. Only where a message appears has changed.

### Important

**I1. "Change amount" is offered for errors that changing the amount can't fix.** `send-bitcoin-confirmation-screen.tsx:608` opens the sheet for every classified SDK code (`Generic`, `NetworkError`, `InvalidInput`). The inline chip on the same screen only offers "Change amount" for amount errors (`isAmountFixableError`, `:639-642`). A fixed-amount invoice (`paymentDetail.canSetAmount === false`) also sends the user back to a screen where the amount can't be edited.

**I2. The same message shows twice on review.** The inline red text and the sheet body are identical, and the sheet's top edge clips the inline text's second line. This is by ruling (the inline error stays), but it needs a decision on whether the inline slot should change while the sheet is up.

**I3. The Android bottom-gap fix hasn't been checked on iOS or in Android light mode.**
- **iOS:** the gap comes from a `SafeAreaView` inside a view that slides in from off-screen (`error-msg-bottom-sheet.tsx:54`). Its inset may read 0 while off-screen and 34 once the sheet arrives, which changes the measured header and can re-settle the rest offset mid-entry.
- **Android:** the numbers suggest the modal window stops above the navigation bar. In light mode that could leave a scrim-coloured strip under a white sheet.

**I4. #4254's map sheets probably have the same doubled inset.** `place-sheet` and `category-filter-sheet` add the root `insets.bottom` inside the same `Modal`. If I3's reading holds, both sit about 24dp too high on Android. The commit message also states "counted twice" as the cause, which was only inferred from two measurements (CTA 69dp from the bottom edge with the root inset, 44.6dp with `SafeAreaView`).

**I5. No specs** for `ErrorMsgBottomSheet`, `useDismissibleErrorMsg` or the review wiring (POC for now). `send-confirmation.spec.tsx` still has #4281's 7 failures.

**I6. The tickets contradict the PR.** blinkbitcoin/blink-wip#1273 says "review shows no failure text", and blinkbitcoin/blink-wip#1275 still describes inline error and problem sheet as either/or.

### Minor
- **M1.** `useDismissibleErrorMsg` only resets when the message clears, so a dismissed sheet stays closed if the error changes to a different message (`use-dismissible-error-msg.ts:12`).
- **M2.** The prettier fix on `sentReceiptRef` (`:305`) is #4281's code riding along in this diff.
- **M3.** The CTA renders at weight 600 until #4277 lands (Figma says 700).
- **M4.** Locale parity will fail: `problemSheetTitle` is `en` only.
- **M5.** The `BottomSheet` copy has to be kept in sync with #4254 by hand until #4254 merges.
- **M6.** Merge order relative to `feat/send-3-receipt-style` (blinkbitcoin/blink-wip#1309) is undecided.
- **M7.** "Change amount" navigates while the `Modal` is still closing. That's fine on Android, but on iOS it can warn or flicker.

### Landing order
I1 first: gate the action on `isAmountFixableError(feeErrorCode) && paymentDetail.canSetAmount` and decide the fallback. Then I3 on iOS and Android light mode, which decides I4.

PR #4284, opened 2026-09-18
test(send): wait on the confirmation screen's late rows instead of reading them early

esaugomez31:
`send-confirmation.spec.tsx` has four cases that assert straight after the first `flushEffects()`, before the fee row and the display currency have settled, so each goes red on its own about one run in four on CI.

This waits on the assertion instead, the fee amount first and its three siblings the same way. Test-only, no app code.


kngako (changes requested):
## Findings

### 1. `waitFor` around the placeholder counts removes the guard's upper bound — [BLOCKING]

Both hide-balance count assertions now sit inside `waitFor` ([send-confirmation.spec.tsx:1335](__tests__/screens/send-confirmation.spec.tsx:1335), [:1356](__tests__/screens/send-confirmation.spec.tsx:1356)). Probes A and B show neither ever waits: the placeholder is rendered synchronously from context, so `toHaveLength(1)` is satisfied before `flushEffects()` is even called, and `toHaveLength(0)` is satisfied by an empty tree at any instant. Probe C shows what the wrapper costs — a component that renders one placeholder and a second one 100 ms later passes `toHaveLength(1)`, and one that leaks a placeholder in after the check passes `toHaveLength(0)`.

**When it comes up:** the test's own comment says the point of "exactly one" is that "a second would mean the amount or fee field had been masked too". The plausible way that breaks is someone masking the fee row off the same display-currency resolution the sibling test waits for — i.e. a second placeholder that appears *late*. That is precisely the case `waitFor` now lets through, so the assertion has been disarmed against the only regression it was written to catch.

**Proposal:** Keep the wait, but put it on a positive settled signal and assert the count synchronously afterwards — `await waitFor(() => expect(screen.getByLabelText("Successful Fee").children).toEqual(["₦0 ($0.00)"]))`, then `expect(screen.queryAllByTestId("hidden-balance-placeholder")).toHaveLength(1)`. The anchor proves the display currency has resolved, and the bare assertion then reads a settled tree, so both directions of "exactly one" hold. Lifting that into a small `renderSettled(hideAmount)` helper in this describe would keep all three cases on one anchor. The same rewrite applies verbatim to the `toHaveLength(0)` case.

### 2. The four `does NOT render the warning` cases read during the window this branch just documented — [BLOCKING]

`236397db6` correctly identifies that the sweep warning lands after the first `flushEffects()`, and fixes the one case that says the warning *should* be there. The four cases that say it should *not* be there still assert `queryByText(...).toBeNull()` immediately after `flushEffects()` ([:814](__tests__/screens/send-confirmation.spec.tsx:814), [:832](__tests__/screens/send-confirmation.spec.tsx:832), [:850](__tests__/screens/send-confirmation.spec.tsx:850), [:880](__tests__/screens/send-confirmation.spec.tsx:880)) — the exact moment at which `useSendDustWarning` is still `PENDING` and the banner is absent for every input. Mutation D confirms they are not reliable guards: with the dust condition widened to every USD send, run 2 of three let "no amountAdjustment" and "draining the full USD balance" pass.

**When it comes up:** someone loosens `resolveDustAdjustment` — say, treating a `null` adjustment as dust, or dropping the drain check — and CI goes green on the run that matters. The screen then tells a user sending $2.00 of a $10.00 balance that their remainder will be converted to Bitcoin when it will not, which is the false positive the BTC-source case was explicitly written to prevent.

**Proposal:** Give each negative the same anchor the positive now has, then assert absence synchronously: wait for a signal that proves the price conversion resolved (`getByLabelText("Successful Fee")` text, as in the hide-balance case), and only then `expect(screen.queryByText(usdRemainderSweepMatcher)).toBeNull()`. A shared `renderSweepCase()` in that describe would apply it in one place to all five. Re-run Mutation D afterwards — all four USD-source negatives should fail on every run, which is the check that the anchor is doing its job.

### 3. The stated cause does not fit the two placeholder cases — [NON-BLOCKING]

`236397db6`'s message attributes ~1-in-4 reds to reading "before the fee row and the display currency have settled", and the new comment at [:1331](__tests__/screens/send-confirmation.spec.tsx:1331) repeats it for the placeholder count. Probe B contradicts this: the count is 1 synchronously, before any flush, because `hideAmount` is a plain context value. And `toHaveLength(0)` cannot go red from reading too early at all — zero is what an unsettled tree returns.

**When it comes up:** the next person debugging this file reads the comment, believes the placeholder rows are display-currency-dependent, and reasons about an ordering that does not exist. If those two cases really did flake at 1-in-4, the cause is still in the tree and will resurface somewhere with no `waitFor` on it.

**Proposal:** Pull the CI failure output for those two cases before keeping the hunks. If the failures were the fee-amount case in the same describe (the one with a genuine async read), fix that one — done in `1012c2f8a` — and revert the two placeholder hunks to the plain assertions, which is also what Finding 1 asks for. If they genuinely failed on the count, the suspect is shared `beforeEach` state or a render-time throw, not settling, and the comment should say what was actually found.

### 4. `.children` deep-equality is a brittle way to read the fee — [NON-BLOCKING]

[:1348](__tests__/screens/send-confirmation.spec.tsx:1348) asserts `screen.getByLabelText("Successful Fee").children` equals `["₦0 ($0.00)"]`, carried over from the pre-existing style at [:328](__tests__/screens/send-confirmation.spec.tsx:328). This reaches into the host node's child array and depends on the fee `Text` containing exactly one string child.

**When it comes up:** the fee cell grows any nested element — an approximate-symbol `Text`, a tooltip, the `*` max-fee marker that already exists one branch over at [send-bitcoin-confirmation-screen.tsx:586](app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx:586). The visible text is still "₦0 ($0.00)", but `children` becomes an array of nodes and the case fails with a diff of fiber objects rather than of strings.

**Proposal:** Assert on the rendered text instead — `expect(screen.getByLabelText("Successful Fee")).toHaveTextContent("₦0 ($0.00)")` — which survives nesting and prints readable failures. As a `waitFor` anchor it behaves identically, since it throws until the conversion lands. Worth applying to [:328](__tests__/screens/send-confirmation.spec.tsx:328) at the same time so the file has one idiom.

---

## Test Coverage Gaps

1. **Nothing pins that the amount and fee rows are unmasked at the settled state.** This is the direct cost of Finding 1: after the change, no assertion in the file would fail if hide-balance leaked into the fee cell one tick after render. A positive assertion — the fee text is readable while `hideAmount` is on — already exists in the middle case and is the honest guard; the count assertion should complement it, not stand in for it.

2. **The `PENDING` window is only ever observed, never asserted.** Mutation D and Probe A both show the screen spends a real interval with `dustWarning.status === "pending"`, and that state disables the slider via `dustNotEvaluable` ([send-bitcoin-confirmation-screen.tsx:148](app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx:148), used at [:629](app/screens/send-bitcoin-screen/send-bitcoin-confirmation-screen.tsx:629)). No case renders with the price conversion unresolved and asserts the slider is disabled, so the "cannot sweep unwarned" property holds only for the fee-error path that is tested.

3. **`BLOCKED` has no case at all.** `useNonCustodialConversionLimits` is mocked once at module scope with `loading: false, error: null` ([send-confirmation.spec.tsx:211](__tests__/screens/send-confirmation.spec.tsx:211)), so the `error || !limits?.minFromAmount` branch — which also disables the slider — is unreachable from this spec. A `jest.fn()` mock that individual cases can re-point would open both this and gap 2.

4. **No case proves the warning is absent *after* settling for a non-dust input.** Even with Finding 2 fixed by anchoring, it is worth one case that waits for the warning's own inputs to resolve and then asserts absence, so the suite distinguishes "not yet" from "never".

---

## Code Smells

1. **`flushEffects()` immediately followed by `waitFor` contradicts the helper's own contract.** [flush-effects.ts](__tests__/helpers/flush-effects.ts) says in its docblock: "Tests that already await `waitFor`/`findBy*` on the resolved state do not need this — the awaited query already settles the effects inside `act()`." All four new sites do both. Harmless at runtime, but it is now the pattern four cases demonstrate.

2. **The file carries four waiting idioms.** `act(() => new Promise(r => setTimeout(r, 10)))` ([:319](__tests__/screens/send-confirmation.spec.tsx:319)), `await findByLabelText` ([:328](__tests__/screens/send-confirmation.spec.tsx:328)), `flushEffects()` alone, and now `flushEffects()` + `waitFor`. A reader cannot tell which one a new case should use. One `renderSettled` helper per describe, anchored on the fee text, would collapse all four.

3. **Four copies of the same three-line `waitFor` block.** Each new site repeats the wrapper inline rather than sharing it, which is why Finding 2's four siblings were easy to overlook — there was no single place that defined what "settled" means for this screen.

4. **The new comments assert a mechanism rather than an intent.** "the rows settle after the first flush" ([:1331](__tests__/screens/send-confirmation.spec.tsx:1331)) and "the row renders before the display currency resolves" ([:1345](__tests__/screens/send-confirmation.spec.tsx:1345)) are causal claims; the first is not true of the block it annotates (Finding 3). Comments that name the anchor ("waits for the converted amount, which is the last thing to land") age better than ones that describe an ordering.

5. **Block-comment style is new to this file.** The branch introduces `/** … */` inside test bodies where every other comment here is `//`. Cosmetic, but it is a diff-visible break with the surrounding file.

6. **`1012c2f8a` has no message body.** It makes the same class of change as `236397db6`, whose body is genuinely useful (flake rate, suspected cause). Squashing the two, or giving the first the same two lines, would leave one legible record of why these waits exist.

esaugomez31:
@kngako Thanks — all of it is in `13e21d7c3`, with the two commits squashed into one. Where each point landed:

## Findings

1. **`waitFor` removes the count's upper bound** — Fixed. Both count assertions are synchronous again, after a shared `renderSettled` anchor.
2. **The four negatives read during the pending window** — Fixed. All five sweep cases render through `renderSettled` and then assert synchronously. Re-ran your Mutation D: the three USD-source negatives now fail on **every** run, where before two of three passed. The BTC-source one stays green, as it should — it is guarded by `isUsdSource`, not by the dust rule.
3. **The stated cause does not fit the placeholder cases** — You were right. A probe confirmed the placeholder count is `1` synchronously, before any flush, so both hunks were reverted to plain assertions. Your other hypothesis was also correct: the real fragility was shared `beforeEach` state. Every describe but the first was running on mocks leaked from it, and none could run in isolation. The shared defaults moved to a root `beforeEach`, and each describe now passes standalone.
4. **`.children` deep-equality is brittle** — Fixed at both sites, but with `within(row).getByText(...)` rather than `toHaveTextContent`. That matcher does not typecheck here: RNTL 13 and `jest-native` both augment `jest.Matchers` with incompatible signatures for it, so TS drops the member. `within` survives nesting the same way and is already used in a sibling spec.

**On the anchor itself**, since everything above rests on it: `renderSettled` waits for the fee to read as a whole-number NGN amount. `displayCurrencyHasSignificantMinorUnits` returns `true` while `usePriceConversion().convertMoneyAmount` is undefined, so NGN prints `₦0.00` until the realtime price lands and `₦0` after. `useSendDustWarning` gates `pending` on that same value, so the whole-number form is what proves the dust verdict is final. Waiting on the `₦` symbol alone would only have proven the currency list had loaded.

## Test Coverage Gaps

1. **Nothing pins the rows unmasked at the settled state** — Covered. The hide-balance case now asserts both the amount row and the fee row on the settled tree.
2. **The `PENDING` window is never asserted** — Covered. A new case reads the slider before settling (disabled), then after (enabled), so the gate is shown to be that state and not a permanent condition of the input.
3. **`BLOCKED` has no case at all** — Covered. The conversion-limits mock is now a `jest.fn()` cases can re-point, as you suggested, and a new case pins slider-disabled plus no warning when the limits fail.
4. **No case proves absence after settling** — Covered by the anchored negatives in Finding 2.

## Code Smells

1. **`flushEffects()` + `waitFor`** — Gone; those sites use `renderSettled` alone.
2. **Four waiting idioms** — `act(() => new Promise(r => setTimeout(r, 10)))` is now gone from the file entirely and `renderSettled` is the single settle idiom. `flushEffects` alone remains at four sites where `fee.status` is `error` or `loading`, so the fee row never renders and there is nothing to anchor on. Open to a better anchor for those if you have one.
3. **Four copies of the same `waitFor` block** — One helper.
4. **Comments assert a mechanism** — Rewritten to name the anchor and the intent rather than an ordering.
5. **Block-comment style** — Not applied, and flagging it rather than leaving it silent: the convention I work to on this repo is `//` for a single line and `/** … */` once a comment runs to two or more. Happy to switch if you would rather the file stay uniform.
6. **`1012c2f8a` has no message body** — Squashed, per your suggestion, into one commit whose body records the flake and what the waits are for.

One extra, found while applying the above: the `GaloySliderButton` mock set `disabled` only on `accessibilityState`, never on the prop, so `fireEvent.press` fired `onSwipe` on a slider the real component would refuse. Now wired to both. Forcing the slider permanently disabled used to fail 6 cases; it fails 16.

39/39 green across three consecutive runs, lint and `tsc` clean.

kngako (approved):

